PR #83152, opened 2022-01-26
Fix and optimize "Add XContentFieldFilter (#81970)"

mushao999:
In "Add XContentFieldFilter (#81970)"，I introduced `XContentFieldFilter` to improve filtering performance for `SourceFieldMapper` and `ShardGetService`. But no tests is added. Something unexpected was then found in issue #82891 .So changes in that PR was reverted in #82897 .
- This PR fix the bug reported in @82891 and add some tests that similiar to `XContentMapValuesTests` to make new filter behave in the same way with the old map filter.
-  We found that new filter will not preserve empty array and object. @nik9000 had fixed this in `jackson-core repository` : https://github.com/FasterXML/jackson-core/pull/729. We marked the related tests to `@AwaitsFix` to wait for this patch to be added into ES.
- We found that dot in fieldName is not supported in the new filter. So we made some changes to support this feature.(moved to #83178 )
- We also add some null and empty check for the new filter.

PR #85382, opened 2022-03-28
Do not fail on duplicated content field filters

arteam:
Source includes and excludes, both configured in the mappings as well as provided as part of a search and get request, may include duplicate entries. This caused no problems until #83152 which makes Elasticsearch throw an error when we are processing the request. While users can update their request for get and search, this prevents users from upgrading to 8.1 if they have duplicate [source includes/excludes](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-source-field.html#include-exclude) configured in their existing mapping.

`Set.of()` throws a runtime exception if it encounters a duplicated element, `Set.copyOf`, on the contrary, doesn't do that.
We shouldn't fail if the user configure multiple include or exclude filters for `_source` fields.


javanna:
hey @arteam do you have any background on what bug this PR is fixing? Is it something that you encountered, or is there an issue filed for it? Would it be possible to add a description around what it is addressing from a user perspective? Thanks!

javanna:
This seems to have been introduced with #83152 ,  @nik9000 @romseygeek would you mind having a look as you were involved in getting that one change in?

romseygeek (approved):
Huh, TIL we shouldn't use `Set.of()` with user-provided input. LGTM!

server/src/main/java/org/elasticsearch/common/xcontent/XContentFieldFilter.java
```diff
@@ -73,8 +73,8 @@ static XContentFieldFilter newFieldFilter(String[] includes, String[] excludes)
             };
         } else {
             final XContentParserConfiguration parserConfig = XContentParserConfiguration.EMPTY.withFiltering(
-                Set.of(includes),
-                Set.of(excludes),
+                Set.copyOf(Arrays.asList(includes)),
```
javanna, line 76: Given that we can have duplicated entries, although whether they are allowed or not depends on the implementation, I wonder if we should change the signature of withFiltering here, and take a Collection instead.

javanna:
Would it make sense to add a test to SourceFieldMapperTests that is affected by this change? The other caller that is affected is the fetch phase, but I haven't found a great place to add a contained test. Maybe a yaml test would do too.

nik9000 (approved):
Huh. Today I learned.

nik9000:
> Would it make sense to add a test to SourceFieldMapperTests that is affected by this change? The other caller that is affected is the fetch phase, but I haven't found a great place to add a contained test. Maybe a yaml test would do too.

I probably would have added a yaml test myself but I'm not sure we super need one for this. OTOH, I'm paranoid and paranoia is good.

arteam:
I've added a couple of test to `SourceFieldMapperTests` :+1:

arteam:
@elasticmachine update branch

arteam:
Thanks Alan, Luca, and Nik!

server/src/main/java/org/elasticsearch/common/xcontent/XContentFieldFilter.java
```diff
@@ -73,8 +73,8 @@ static XContentFieldFilter newFieldFilter(String[] includes, String[] excludes)
...
+                Set.copyOf(Arrays.asList(excludes)),
```
idegtiarenko, line 77: I had a chance to review it only after the pr was merged, may be we could consider this for a followup or a future change:

May be we should have a dedicated method for this in CollectionUtils so that intention is clear and nobody revert/simplify code to the shorter `Set.of` version. May be something like `CollectionUtils.setOfCollectionWithPossibleDuplicates(Collection<T> c)`

javanna: not advocating against your suggestion, but we do have tests for this now so reverting the change would cause test failures

nik9000: This PR introduced a test for duplicate filters. I think that's what you want, right?

idegtiarenko: Correct, the test would catch the revert. I was suggesting to have a dedicated set factory that would display the intention (create set from collection that may contain duplicates)

Issue #84046, opened 2022-02-16
Optimize tsdb data stream timestamp parsing if ingest pipeline is used


martijnvg:
For tsdb data streams, the `@timestamp` field is used to determine to which backing index a data stream resolves to.

Currently in `IndexAbstraction.DataStream#getWriteIndex(...)` the source of the document is parsed in order to extract the `@timestamp` field. If an ingest pipeline is active then this parsing is redundant, because ingest parse the source of a document to into a map of maps in order for a pipeline to process it.

After the final pipeline has executed, ingest can extract the `@timestamp` field and attach it as String/Long to the `IndexRequest` class as a field. Then `IndexAbstraction.DataStream#getWriteIndex(...)` can check for whether that field has been set and if so use that parse it into a `Instant` instance.

martijnvg:
Implemented via #85387

PR #85387, opened 2022-03-28
Optimize tsdb data stream timestamp parsing if ingest pipeline is used

weizijun:
This PR is to implement https://github.com/elastic/elasticsearch/issues/84046.
When setting a mapped source in IndexRequest, it cache in the sourceAsMap.
The ingest pipeline will set the mapped source, so data stream get the timestamp from sourceAsMap.


server/src/main/java/org/elasticsearch/action/index/IndexRequest.java
```diff
@@ -83,6 +83,7 @@
     private String routing;
 
     private BytesReference source;
+    private Map<String, Object> sourceAsMap;
```
martijnvg, line 86: The timestamp is always going to be in the `@timestamp` field. So maybe instead of caching the entire source just cache the timestamp itself?

martijnvg (commented):
Thanks @weizijun! I took a quick look and I think this looks good. I left one comment and I will take a better look later.

server/src/main/java/org/elasticsearch/action/index/IndexRequest.java
weizijun: I think the cached sourceAsMap may be needed elsewhere as well. Time series index will parse routing fields from the source, it can reuse the sourceAsMap.

martijnvg: > Time series index will parse routing fields from the source, it can reuse the sourceAsMap.

I rather not keep a map of maps copy of the source around here. Let's focus on routing in a followup change? I think we can both timestamp and routing fields post ingest time without carrying around a bloated map of maps.

server/src/main/java/org/elasticsearch/action/index/IndexRequest.java
```diff
@@ -363,11 +374,14 @@ public IndexRequest source(Map<String, ?> source) throws ElasticsearchGeneration
...
      */
+    @SuppressWarnings("unchecked")
     public IndexRequest source(Map<String, ?> source, XContentType contentType) throws ElasticsearchGenerationException {
         try {
             XContentBuilder builder = XContentFactory.contentBuilder(contentType);
             builder.map(source);
-            return source(builder);
+            IndexRequest indexRequest = source(builder);
+            this.sourceAsMap = (Map<String, Object>) source;
+            return indexRequest;
```
martijnvg, line 384: Maybe make a to `IngestService` instead of modifying this method (which does get invoked in other places too)?
I think we can add the following to `IngestService#innerExecute(...)` method at line 873:

```
Object rawTimestamp = ingestDocument.getSourceAndMetadata().get("@timestamp");
                if (rawTimestamp != null) {
                    indexRequest.setRawTimestamp(rawTimestamp);
                }
```

and then add a `rawTimestamp `field to `IndexRequest`.

This feels more targeted and cleaner to me.

weizijun: ok, I will modify the code as you suggest.

server/src/main/java/org/elasticsearch/action/index/IndexRequest.java
```diff
@@ -112,6 +112,8 @@
 
     private Map<String, String> dynamicTemplates = Map.of();
 
+    private Object rawTimestamp;
```
weizijun, line 115: hi, @martijnvg , will the rawTimestamp be cleaned up after a new source set? If so, it will cause the setRawTimestamp and getRawTimestamp not the same, when a new source set. It's a little strange.

martijnvg: No, it will not be cleaned up. I don't think a new source will be set after `setRawTimestamp(...)` is invoked.
I think to be sure we can maybe add an assertion to `IndexRequest#source(...)` setter and if `rawTimestamp` is not null then fail.

The `rawTimestamp` field is only needed on coordinating node when resolving the data stream defined in `index`  field to a backing index.

martijnvg:

weizijun:
@elasticmachine update branch

martijnvg:

martijnvg:
@elasticmachine update branch

martijnvg:

martijnvg:
@elasticmachine update branch.

martijnvg:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/ingest/IngestService.java
```diff
@@ -880,6 +881,15 @@ private void innerExecute(
                 handler.accept(null);
             }
         });
+        postExecute(ingestDocument, indexRequest);
```
martijnvg, line 884: I think `postExecute(...)` should be invoked from callback before invoking the  `handler` (line 881).
If pipeline execution is asynchronous then modification to `ingestDocument` may still be made here.

server/src/main/java/org/elasticsearch/action/index/IndexRequest.java
```diff
@@ -761,4 +763,12 @@ public IndexRequest setDynamicTemplates(Map<String, String> dynamicTemplates) {
     public Map<String, String> getDynamicTemplates() {
         return dynamicTemplates;
     }
+
+    public Object getRawTimestamp() {
+        return rawTimestamp;
+    }
+
+    public void setRawTimestamp(Object rawTimestamp) {
+        this.rawTimestamp = rawTimestamp;
```
martijnvg, line 772: Maybe add `assert this.rawTimestamp == null` before setting `this.rawTimestamp`?

server/src/main/java/org/elasticsearch/action/index/IndexRequest.java
```diff
@@ -112,6 +112,8 @@ public class IndexRequest extends ReplicatedWriteRequest<IndexRequest> implement
 
     private Map<String, String> dynamicTemplates = Map.of();
```
martijnvg, line 114: maybe add a comment here that `rawTimestamp` field doesn't need to be serialised, because it is only used on the node the coordinates the request.

martijnvg (commented):
@weizijun I left a few comments.

Maybe also change `DataStreamGetWriteIndexTests#getWriteIndex(...)` method
to randomly set source with timestamp or the timestamp via the `setRawTimestamp(...)` field.

weizijun:
> @weizijun I left a few comments.

@martijnvg , Thanks a lot, I will improve them.

server/src/main/java/org/elasticsearch/action/index/IndexRequest.java
weizijun: done

server/src/main/java/org/elasticsearch/ingest/IngestService.java
weizijun: Now pipeline execution is synchronous, but it's more suitable to place as you suggest. And I rename it to `postIngest`

server/src/main/java/org/elasticsearch/action/index/IndexRequest.java
martijnvg: 👍

server/src/main/java/org/elasticsearch/ingest/IngestService.java
martijnvg: In the case a processor implementation goes async then it isn't. This is the case for example with the `enrich` and `inference` processors. These processor invoke a (remote) action.

In any case the method is no invoked from the callback. So I think it is good now 👍

weizijun:
@elasticmachine update branch

weizijun:
> Maybe also change `DataStreamGetWriteIndexTests#getWriteIndex(...)` method
> to randomly set source with timestamp or the timestamp via the `setRawTimestamp(...)` field.

done

martijnvg:

weizijun:
@elasticmachine update branch

martijnvg:

martijnvg (approved):
LGTM

Issue #57918, opened 2020-06-10
Elasticsearch build and Gradle plugins should support Gradle configuration cache


breskeby:
The gradle configuration cache is an incubating feature that significantly improves gradle configuration time by applying cache mechanisms to the gradle build configuration. 
The functionality is described in detail at https://docs.gradle.org/nightly/userguide/configuration_cache.html 
In order to support this configuration cache some requirements must be met by plugins and build scripts that are described here: https://docs.gradle.org/nightly/userguide/configuration_cache.html#requirements

PR #85388, opened 2022-03-28
Shut down test cluster in a configuration cache compliant way

breskeby:
TaskExecutionListener are deprecated in Gradle and we want to move away from them
to be able to take advantage of the gradle configuration cache (see #57918)

This removes the listener used for shutting down test clusters. We want to remove more
but replacing logic in beforeActions is not as straight forward yet with Gradle 7.4


breskeby:
@elasticmachine update branch

PR #84230, opened 2022-02-22
SQL: Implement binary format support for SQL clear cursor

luigidellaquila:
- support binary_format parameter on _sql/close
- make cursor close response consistent with the query protocol
  (ie. in ODBC/JDBC/CLI return CBOR response for cursor close - as for query)

Fixes #53359

PR #83833, opened 2022-02-11
SQL: Fix issues with format=txt when paging through result sets and in mixed node environments

Luegg:
Resolves #83581
Resolves #83788

SQL REST requests using `format=txt` stand out from the other formats because the cursor needs to carry the formatter state from the initial request to subsequent scroll requests. The state is needed to be able to format the subsequent pages with column widths that match the widths in the first page.

Currently, this is solved by wrapping cursor objects together with the formatter state in `TextFormatterCursor`. Hence, a query might return a `ListCursor` and `TextFormat.PLAIN_TEXT` adds the necessary state when formatting the output. This approach is handy because a `TextFormatterCursor` is a `Cursor` and delegates calls to the wrapped cursor. Unfortunately, it also has some downsides that have been revealed when looking into #83581 and #83788:

- MediaType formatting is a concern of the REST layer of the plugin that does not have access to the serialization logic usually accessible through the `NamedWriteablesRegistry`. Because the formatting required to (de)serialize the cursors for the wrapping, this meant that the `Cursor` deserialization could only use a subset of `NamedWriteable`s specific to Cursors (namely the ones returned by `Cursors.getNamedWriteables`). This subset of writeables is not enough to deserialize `CompositeAggCursor` which lead to the surprising design that `CompositeAggCursor` has a `nextQuery` member consisting of the serialized `SearchSourceBuilder` whose deserialization is suspended until the call to `Cursor.nextPage`.
- `TextFormat.PLAIN_TEXT` had to deserialize the complete cursor to do it's work, not just the information relevant to the formatter. Because of the `min_compatible_version` redirects that occur during rolling upgrades, it can happen though that an upgraded node needs to deserialize a cursor from a redirected SQL query written by an older node. Because we do not offer any form of bwc for cursors, this causes an error and leads to #83581.

In this PR, I propose to change the design such that the REST layer can add state to the cursor strings without having to read or write instances of `Cursor`. This allows to address the issues described above and also made fixing #83788 very straightforward.

Attaching state to the cursor is achieved by extending the format of encoded cursors. Previously, an base64 encoded SQL cursor had the following wire format `<version><zoneId><cursor>`. Now, there are two different cursor types:

* `<version><zoneId><CursorType.NO_STATE><cursor>` for most cases
[...]

PR #83591, opened 2022-02-07
SQL: Compress Cursors

Luegg:
Partially addresses #83284

It looks like compressing cursors is a low hanging fruit to address issues around cursor size that seems to pay off pretty nicely (thanks @costin for the pointer). This is especially relevant for queries using the `ListCursor` (e.g. `SHOW COLUMNS` or queries sorting on aggregated columns) but might also benefit queries that store `SearchSourceBuilder` in the cursor.

Some anecdata for the potential savings in cursor size with `big` being an index with 1000 long fields `field1`, `field2`, ...:

| Request | Cursor size before | after | % |
|---------|--------------------|-----|----|
| `http post "localhost:9200/_sql?format=txt" query="SHOW COLUMNS IN big" fetch_size=1` | 34528b | 3208b | 9% |
| `http post "localhost:9200/_sql?format=csv" query="SHOW COLUMNS IN big" fetch_size=1` | 34508b | 3188b | 9% |
| `http post "localhost:9200/_sql?format=txt" query="SHOW COLUMNS IN test_emp" fetch_size=1` | 720b | 336b | 47% |
| `http post "localhost:9200/_sql?format=txt" query="SELECT emp_no, COUNT(*) FROM test_emp GROUP BY 1 ORDER BY 2" fetch_size=1` | 2672b | 352b | 13% |
| `http post "localhost:9201/_sql?format=csv" query="select max(languages), gender from test_emp group by 2" fetch_size=1` | 268b | 224b | 84% |

The format only has an impact on cursor size for the `txt` format. Hence, I'd also expect similar savings for the binary protocols.

PR #85399, opened 2022-03-28
SQL: Fix pagination during rolling upgrade

### LINKED ISSUE #83726: SQL: Scroll cursors issued by upgraded nodes during rolling upgrade are invalid
### Elasticsearch Version

8.2.0-SNAPSHOT

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

any

### Problem Description

If a SQL query is made against an upgraded node during a rolling upgrade from any version < 8.2 to any version >= 8.2, the returned scroll cursor cannot be used to fetch subsequent pages.

The problem is caused by SQL redirecting queries to the older nodes in the cluster. Hence, when querying an 8.2 node it will redirect the request to the old node (let's say 7.17) and respond with a 7.17 cursor. A subsequent scroll request with the 7.17 cursor to the 8.2 fails with:

```
{
    "error": {
        "reason": "Unsupported cursor version [7.17.1], expected [8.2.0]",
        "root_cause": [
            {
                "reason": "Unsupported cursor version [7.17.1], expected [8.2.0]",
                "type": "sql_illegal_argument_exception"
            }
        ],
        "type": "sql_illegal_argument_exception"
    },
    "status": 500
}
```

A way to work around this issue is to ensure that SQL query and scroll requests always go to the node in the cluster with the earliest version.

This issue is related to https://github.com/elastic/elasticsearch/issues/83581 but not quite identical.

### Steps to Reproduce

1. Create a two node cluster with ES 7.17.1
2. Create any index
3. Upgrade one node to 8.2.0-SNAPSHOT
4. Run any SQL query that returns a scroll cursor against the 8.2 node
5. Take the cursor from the response and issue a scroll request to the same 8.2 node

### Logs (if relevant)

_No response_

Luegg:
Resolves #83726 by trying to redirect requests with a cursor with the wrong version to a node with the matching version.

The most frequent scenario that this PR addresses is the following situation during a rolling upgrade:

1. An SQL query either hits a not-yet-upgraded node or it is redirected to the older node by the mechanism implemented in `TransportActionUtils.executeRequestWithRetryAttempt`. The returned cursor will only work on the nodes with the old version.
2. A follow up scroll (or clean cursor) request hits a node that has already been upgraded. Previously, this failed during deserialization of the cursor. With this change, the request will be forwarded to a node with the cursor's version.

Especially in larger clusters, one might also encounter an additional error mode that is addressed by this PR:

1. An index is allocated on nodes that have already been upgraded.
2. A search request hits an upgraded node but won't be redirected because all shards fulfill the version requirement.
3. The follow up scroll (or clean cursor) requests hits an old node.

The fix should also work with requests with `format=txt` but only if all nodes support the redirect. This case is more complicated because the cursor wraps another cursor together with the formatter and inner and outer cursor might not have the same version.


x-pack/plugin/sql/qa/mixed-node/src/test/java/org/elasticsearch/xpack/sql/qa/mixed_node/SqlCompatIT.java
```diff
@@ -154,42 +168,109 @@ public static String sqlQueryEntityWithOptionalMode(String query, Version bwcVer
...
-            Matchers.containsString("Unsupported cursor version [" + version1 + "], expected [" + version2 + "]")
-        );
+        assertNotNull(scrollJson.get("rows"));
+    }
+
+    public void testTextCursorFromOldNodeWorksOnNewNode() throws IOException {
+        assertTextCursorCompatibleAcrossVersions(bwcVersion, oldNodesClient, newNodesClient);
+    }
+
+    @AwaitsFix(bugUrl = "tbd")
```
Luegg, line 249: See last remark in the PR description. I'll create an according issue once this PR is approved and ready to merge.

Luegg: Never mind about this comment. The `@AwaitsFix` referred to a previous attempt to address the issue. With the current attempt, text format based cursors should also be usable across versions if all nodes support the cursor redirect from this PR.

x-pack/plugin/sql/qa/mixed-node/src/test/java/org/elasticsearch/xpack/sql/qa/mixed_node/SqlCompatIT.java
```diff
@@ -154,42 +168,112 @@ public static String sqlQueryEntityWithOptionalMode(String query, Version bwcVer
...
+
+        Request scrollReq = new Request("POST", "_sql/close");
+        scrollReq.addParameter("error_trace", "true");
+        scrollReq.setJsonEntity("{\"cursor\": \"%s\"}".formatted(cursor));
+        Map<String, Object> scrollJson = performRequestAndReadBodyAsJson(client2, scrollReq);
+
+        assertEquals(scrollJson.get("succeeded"), true);
+    }
+
+    public void testCursorForIndexAllocatedOnNewNodeWorkOnOldNode() throws IOException {
```
Luegg, line 215: Unfortunately, this test and `testTextCursorFromNewNodeWorksOnOldNode` can only be tested once we have a release including the cursor redirect. Any ideas on how to run the test anyway would be highly appreciated :)

x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/plugin/TransportActionUtils.java
```diff
@@ -78,4 +78,56 @@ public static void executeRequestWithRetryAttempt(
...
+            }
+            onFailure.accept(exception);
+        }
+
+    }
+
+    private static DiscoveryNode findNodeWithVersionOtherThanLocalNode(ClusterService clusterService, Version expectedVersion) {
+        DiscoveryNode localNode = clusterService.state().nodes().getLocalNode();
+        if (expectedVersion.equals(localNode.getVersion())) {
+            return null;
```
luigidellaquila, line 122: Not sure I understand the logic here: if current node has the right version, it should be a good candidate itself... and a QlVersionMismatchException shouldn't have been thrown in the first place. Am I missing something?

x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/plugin/TransportActionUtils.java
```diff
@@ -78,4 +78,56 @@ public static void executeRequestWithRetryAttempt(
...
     }
+
+    public static void retryOnNodeWithMatchingVersion(
+        ClusterService clusterService,
+        QlVersionMismatchException exception,
+        Consumer<DiscoveryNode> retryRequest,
+        Consumer<Exception> onFailure,
+        Logger log
+    ) {
+        if (log.isDebugEnabled()) {
```
luigidellaquila, line 89: (very minor note) IMHO this log does not belong here. This method is responsible of retrying the request, not of checking the version.

luigidellaquila (commented):
In general it seems pretty good. 
Just leaving a couple of comments

x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/plugin/TransportActionUtils.java
Luegg: Yes, that's right. This condition should not happen and I don't think you can end up in this situation. But if it is the case, you would want to avoid the retry because you could end up in an endless loop otherwise. Throwing an exception here is probably even better.

luigidellaquila: Right, it makes sense. 
The exception makes the intention much more clear 👍

x-pack/plugin/sql/qa/mixed-node/src/test/java/org/elasticsearch/xpack/sql/qa/mixed_node/SqlCompatIT.java
Luegg: Since 8.2 has now been branched out, I can backport this PR to 8.2 when it's ready to merge, see how the tests are doing and then merge to master.

Luegg:
Running the tests against the 8.2 branch revealed some more issues. I'll have to have another look at the implementation to get it right in all cases.

costin (changes requested):
It's best we avoid doing retries based on the cursor version since they can get out of sync with the underlying data ones.
It introduces another layer of complexity and can lead to an unsolvable case: 
the last old node in the cluster creates an "old" cursor, then gets upgraded - there are no more nodes to redirect the "old" cursor to.

An alternative solution is to create old cursors and use those as is, even on the new data nodes, for the entire duration of the query.

x-pack/plugin/sql/sql-action/src/main/java/org/elasticsearch/xpack/sql/action/SqlClearCursorRequest.java
```diff
@@ -103,7 +108,9 @@ public Boolean binaryCommunication() {
     public void writeTo(StreamOutput out) throws IOException {
         super.writeTo(out);
         out.writeString(cursor);
-        out.writeOptionalBoolean(binaryCommunication);
+        if (out.getVersion().onOrAfter(Version.V_8_3_0)) {
```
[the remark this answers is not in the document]
Luegg: the change has been introduced in https://github.com/elastic/elasticsearch/pull/84230 which will go into 8.3

Luegg:
> It's best we avoid doing retries based on the cursor version since they can get out of sync with the underlying data ones. It introduces another layer of complexity and can lead to an unsolvable case: the last old node in the cluster creates an "old" cursor, then gets upgraded - there are no more nodes to redirect the "old" cursor to.
> 
> An alternative solution is to create old cursors and use those as is, even on the new data nodes, for the entire duration of the query.

Looking into this approach it seems like creating old cursors on new nodes could potentially cause quite some difficulties when introducing new features. E.g. when introducing a new `Extractor` that's also encoded in the `Cursor`, SQL could only use the new extractor once all the nodes have been upgraded. As a result, the bwc logic would leak into the remaining QL code (e.g. in form of a compatibility level passed into the planner).

The so far unresolvable case can also be addressed by only forwarding new cursors from old nodes to new nodes but old cursors are always executed on the local node. This would "only" require that old cursor can be read on new nodes. This is currently not the case but should be doable. It would mean that we have to be more careful with some changes that did not fall under bwc so far though (e.g. changing extractors). Thankfully, the `SearchSourceBuilder`, which is probably the most complex structure in the cursors, already falls under bwc checks.

Altogether, this would also make the reasoning about the cursor redirection more straightforward.

Wdyt @costin ?

Luegg:
@astefan pointed out that there is also a way to run the tests that depend on the PR being backported to 8.2.0. I cherry picked some of the changes to https://github.com/Luegg/elasticsearch/tree/cursorCompat82 that branches from 8.2 and set
```
    private static final Version INTRODUCE_REDIRECTS_FOR_CURSORS_WITH_VERSION_MISMATCH_VERSION = Version.V_8_2_0;
```

The PR can now be tested against that version of 8.2.0 by using `gw :x-pack:plugin:sql:qa:mixed-node:v8.2.0#mixedClusterTest --tests "org.elasticsearch.xpack.sql.qa.mixed_node.SqlCompatIT" -Dbwc.remote=Luegg -Dbwc.refspec.8.2=cursorCompat82 -Dtests.bwc.git_fetch_latest=false`

Luegg:
Some notes for a discussion with @costin:

There seems to be 3.5 different ways to approach the problem, all with some other trade offs:

### Always redirect cursor to node with matching version

That's the approach currently implemented in this PR: A query might hit a new or an old node (or it might also be forwarded to an old node) and produce an according cursor. If the scroll requests hits a node with a non-matching version, the request is forwarded to a node with matching versions.

Pros:
* We don't have to care about backwards compatibility of cursors and can easily make breaking changes in Cursors, Extractors, the data stored in ListCursors etc.

Cons:
* Scroll requests with old cursors can still fail as soon as all nodes have been upgraded.
* An additional routing mechanism.
* Only works between versions that both support the redirection mechanism (doesn't work with past releases)

### Only redirect new cursors from old to new nodes

New nodes should always be able to deserialize and execute cursors from old nodes (within the bwc guarantees of rolling upgrades). That eliminates all the issues from old nodes producing cursors that are executed on new nodes. The remaining cases (which are probably less common) will be handled by redirecting new cursors on old nodes to new nodes.

Pros:
* As long as the rolling upgrade progresses as planned, no scroll request should fail because of compatibility issues.
* Redirects only go in one direction (old to new)

Cons:
* Requires Cursors (and all dependent structures) to be bwc compliant.
* An additional routing mechanism.
* Only works between versions that both support the redirection mechanism (doesn't work with past releases)

### Only create old cursors as long as old nodes are in the cluster

There are two variations of how this could be achieved:

#### Redirect queries to old node

Use the cluster state to determine which one is the oldest version in the cluster and redirect all queries to nodes of that version. New nodes would need to be able to understand scroll requests from the old nodes.

Pros:
* Just a modification of the existing query redirection mechanism
* Would work with all past releases.

Cons:
* Increasing load imbalance during the upgrade. Eventually, only one old node would be responsible for planning and executing all incoming queries. This node would also become a single-point-of-failure for the SQL feature.
* Requires Cursors (and all dependent structures) to be bwc compliant.

#### Have new nodes create old cursors

Use the cluster state to determine which one is the oldest version in the cluster and set this as a minimum compatibility level during query planing. The planner would then produce queries that can be executed on nodes of at least this version and the resulting cursor would also be of this version and can be processed on any nodes.

Pros:
* No more redirections (not even the current query redirection to old data nodes)
* Would work with all past releases.

Cons:
* It's unclear how this could be implemented and what impact it would have on future changes. Chances are high that it would significantly complicate various parts in the planner once we want to introduce new features.

Luegg:
We also have to consider the validity of the PIT encoded in the cursors: A PIT can no longer be used as soon as one node hosting a relevant shard has been restarted.

This is illustrated in the following diagram:

image

Query 1 creates PIT 1 on node1 and 2 before they are upgraded from version v1 to v2. Because of the limitations inherent to PIT, subsequent scroll requests can only succeed before the restart of node1. Or in other words, the probability of the scroll requests succeeding goes from 1 before the restart to 0 afterwards because the PIT 1 resource on Node 1 is no longer available.

With 2 nodes, this will result in 5 relevant query scenarios:

* Query 1: PIT is created on node 1 and 2 before any node is upgraded
* Query 2: PIT is created on node 2 before node 1 is upgraded
* Query 3: PIT is created on node 1 and 2 after node 1 is upgraded
* Query 4: PIT is created on node 1 after node 1 is upgraded
* Query 5: PIT is created on node 2 after node 1 is upgraded

This gives the following success rates of scroll requests with the current implementations (master branch):

image

For most queries, the success rate goes down to 0.5 after upgrade of node 1 because of the possibility of having the scroll query hitting node 1 (the cursor will be of version v1). After upgrade of node 2, all cursors except the one issued for query 4 will be invalid. The cursor for Query 4 might still be valid if it has been issued by node 1.

The goal of this PR is to turn as many of the 0.5 into a 1. This should be possible by any of the options discussed above except of "Always redirect cursor to node with matching version" for which the probabilities will look like:

image

The Query 4 cursor might have been issued by node 2 with a probability of 0.5. After the final upgrade there is no node of v1 left to redirect the cursor to.

With more nodes, the pagination success rate for a query hitting data on all nodes is `{numberOfNotYetUpgradedNodes} / {numberOfNodes}`. Hence, it gets worse the more nodes have been upgraded. Averaging the possibilities over the whole time of the cluster upgrade should also give an overall probability of 0.5. If the data is unevenly distributed across nodes, the calculation becomes more cumbersome.

Luegg:
Regarding backwards compatibility: There are three recent changes that broke bwc for the SQL cursors (because it was not a necessity back then):

* https://github.com/elastic/elasticsearch/pull/83591 (8.1.0 and 8.2.0)
* https://github.com/elastic/elasticsearch/pull/83833 (8.2.0)
* https://github.com/elastic/elasticsearch/pull/83381 (8.2.0)

The combined delta of these changes is quite substantial and includes changes to the encoding of the cursors itself (not just the payload). This means that achieving backwards compatibility with versions before 8.2 will be quite unattainable (without spending too much effort).

This affects all the options discussed above that require some sort of cross version compatibility (all but the cursor redirection to nodes with matching version).

Issue #84735, opened 2022-03-07
Remove direct uses of hppc


rjernst:
This is a meta issue for removing most uses of hppc from Elasticsearch.

Like any complex software project, Elasticsearch uses various data structures from computer science: lists, maps, sets, etc. Java provides interfaces for these structures (Java Collections Framework, JCF), and concrete implementations using well known techniques, like a hashtable based Map implementation, HashMap. The Java provided implementations are used throughout Elasticsearch.

Additionally, Elasticsearch server depends on the HPPC library. It provides alternatives to Java’s builtin collections that attempt to provide efficient, fast and “open” implementations (open here means easily hackable). The library is mostly wrapped with our own classes, namely ImmutableOpenMap and ImmutableOpenIntMap, but it is also used sporadically throughout the codebase. But when to use the HPPC based collections vs the Java ones has never been well defined within the team.

HPPC has been around for more than a decade, and while it once provided major performance benefits over JCF, Java has changed a lot in that time, especially in hotspot. It is unclear if HPPC actually provides advantages over JCF anymore. On top of this, the HPPC based collections are more difficult to use, and don’t easily interoperate with JCF interfaces, requiring conversion, or more often, forcing a developer to use HPPC because an upstream object used it, like when dealing with cluster state.

We have decided to try to move away from using hppc. Doing so is tricky since swapping out JCF classes are not trivial in most uses. This meta issue lists direct uses of hppc that should be either removed entirely, or evaluated for performance differences with JCF. Whether performance tests are needed would depend on the use case, and whether it is expected to be sensitive to memory footprint, and the objects lifetime.

- [x] `server`
  - [x] `o.e.action.admin.cluster.stats` (#85639)
  - [x] `o.e.action` Multi* request/response (https://github.com/elastic/elasticsearch/pull/85888)
  - [x] `o.e.action.admin.indices` (https://github.com/elastic/elasticsearch/pull/85911)
  - [x] `o.e.common` (https://github.com/elastic/elasticsearch/pull/85957)
  - [x] `o.e.http` (https://github.com/elastic/elasticsearch/pull/86024)
  - [x] `o.e.index.engine` (https://github.com/elastic/elasticsearch/pull/86072)
  - [x] `o.e.index.mapper.BinaryFieldMapper` (#85352)
  - [x] `o.e.index.mapper` (#85417, #85475)
  - [x] `o.e.index.seqno` (https://github.com/elastic/elasticsearch/pull/86073)
[...]

PR #85406, opened 2022-03-28
Remove hppc from o.e.c.util tests

rjernst:
These tests have no need to use hppc since they do not interact with any
hppc types from server.

relates #84735


original-brownbear (approved):
LGTM

PR #85360, opened 2022-03-25
Fix calculation of max parallel forks for test execution

mark-vieira:
We were inadvertently always using the "fallback" calculation of half
the reported number of processors instead of the more sophisticated
platform-specific implementations. On an M1 Pro this resulted in only
5 parallel test workers instead of the more appropriate 10. Also, even
though Gradle will do this for us, let explicitly make sure we don't
set it higher than the current max worker count.

This commit also removes a duplicate and unused copy of this same logic
from the GlobalBuildInfoPlugin.

PR #85410, opened 2022-03-28
Fall back to naive parallel forks calculation in non-standard cases

mark-vieira:
Follow up to #85360 to fix some issues with parallelism calculation on non x86 systems.


mark-vieira:

PR #85413, opened 2022-03-28
[ML] correctly validate permissions when retention policy is configured

### LINKED ISSUE #85409: [ML] Transform does not validate permissions appropriately when using a retention policy
### Elasticsearch Version

8.0.0

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

n/a

### Problem Description

When creating a transform with a `retention_policy` configured, `delete` permissions are not validated against the destination index. Consequently, the transform may fail to run later once the retention policy attempts to run. 

It should be validated and fail on creation (if validations are not skipped), just like we do for our other permissions.

Additionally, the documentation needs to reflect this permission nuance as right now `delete` is not mentioned.

### Steps to Reproduce

 - create a transform with a retention policy with a user without delete permissions
 - notice its failure 

```
task encountered irrecoverable failure: org.elasticsearch.ElasticsearchSecurityException: action [indices:data/write/delete/byquery] is unauthorized for user [<username>] with roles [<roles>] on indices [<destination index>], this action is granted by the index privileges [delete,write,all]
```

Related issue: https://github.com/elastic/elasticsearch/pull/84941

### Logs (if relevant)

_No response_

benwtrent:
When a transform has a `retention_policy` it needs to be able to delete documents in the destination index. 

`create_index` does not necessitate that we can delete documents from it. So, even if we create the index, we need to verify that we can delete documents given the `retention_policy` definition.

This is not a crucial bug as the transform will simply fail later. Its nicer to fail sooner.

closes https://github.com/elastic/elasticsearch/issues/85409

### DIFF
--- a/docs/changelog/85413.yaml
+++ b/docs/changelog/85413.yaml
@@ -1,0 +1,6 @@
+pr: 85413
+summary: Correctly validate permissions when retention policy is configured
+area: Transform
+type: bug
+issues:
+ - 85409
--- a/docs/reference/transform/apis/put-transform.asciidoc
+++ b/docs/reference/transform/apis/put-transform.asciidoc
@@ -19,36 +19,37 @@ Instantiates a {transform}.
 
 Requires the following privileges:
 
-* cluster: `manage_transform` (the `transform_admin` built-in role grants this 
+* cluster: `manage_transform` (the `transform_admin` built-in role grants this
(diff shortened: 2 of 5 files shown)


przemekwitek (approved):
LGTM

Thanks for fixing this!

PR #85414, opened 2022-03-28
Add NOOP ActionListener Singleton

original-brownbear:
I know there was some reservation against adding this in the past because
noop listeners are hard to debug since they don't even have any logging in them.
Then again, given that it was possible to identify `103` uses of a noop listener
within minutes using Idea I think it's better to face reality here and at least save
some objects and make the code easier to read.
Also having a clean `toString` on this method will render nicer in logs IMO when combined with other listeners.
Last but not least it simply saves some memory and compiler work to have a
singleton for this.

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/ActionListener.java
+++ b/server/src/main/java/org/elasticsearch/action/ActionListener.java
@@ -34,6 +34,32 @@ public interface ActionListener<Response> {
      */
     void onFailure(Exception e);
 
+    @SuppressWarnings("rawtypes")
+    ActionListener NOOP = new ActionListener() {
+        @Override
+        public void onResponse(Object o) {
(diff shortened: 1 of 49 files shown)


rjernst (approved):
LGTM. Since this consolidates these cases to ignore the result of an action, could we consider having a debug or trace level log message output in the noop listener stating the results are being ignored?

original-brownbear:
Thanks Ryan!

> could we consider having a debug or trace level log message output in the noop listener stating the results are being ignored?

I think I have an idea for a follow-up that might make this unnecessary :) I'll try it out, but yea if it doesn't work I think that makes good sense and I'll open a PR for that. I'd still prefer to not have this thing for the most part if I can find a way :)

PR #85417, opened 2022-03-28
Use Set instead of LongSet in long script field

rjernst:
The long script field uses hppc's LongSet to store the unique values
to be queried. However, this set should be relatively small. This commit
changes the internals of the runtime long script field to use Set<Long>.

relates #84735

### DIFF
--- a/server/src/main/java/org/elasticsearch/common/Numbers.java
+++ b/server/src/main/java/org/elasticsearch/common/Numbers.java
@@ -126,9 +126,9 @@ public final class Numbers {
     /** Return the long that {@code stringValue} stores or throws an exception if the
      *  stored value cannot be converted to a long that stores the exact same
      *  value and {@code coerce} is false. */
-    public static Long toLong(String stringValue, boolean coerce) {
+    public static long toLong(String stringValue, boolean coerce) {
         try {
-            return Long.valueOf(stringValue);
+            return Long.parseLong(stringValue);
         } catch (NumberFormatException e) {
             // we will try again with BigDecimal
         }
--- a/server/src/main/java/org/elasticsearch/common/Numbers.java
+++ b/server/src/main/java/org/elasticsearch/common/Numbers.java
@@ -126,9 +126,9 @@ public final class Numbers {
     /** Return the long that {@code stringValue} stores or throws an exception if the
      *  stored value cannot be converted to a long that stores the exact same
      *  value and {@code coerce} is false. */
-    public static long toLong(String stringValue, boolean coerce) {
+    public static Long toLong(String stringValue, boolean coerce) {
         try {
-            return Long.parseLong(stringValue);
+            return Long.valueOf(stringValue);
         } catch (NumberFormatException e) {
             // we will try again with BigDecimal
         }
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/elasticsearch/common/Numbers.java
```diff
@@ -126,9 +126,9 @@ public static long toLongExact(Number n) {
     /** Return the long that {@code stringValue} stores or throws an exception if the
      *  stored value cannot be converted to a long that stores the exact same
      *  value and {@code coerce} is false. */
-    public static long toLong(String stringValue, boolean coerce) {
+    public static Long toLong(String stringValue, boolean coerce) {
         try {
-            return Long.parseLong(stringValue);
+            return Long.valueOf(stringValue);
```
nik9000, line 131: I think I'd keep the return type on this one.

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
```diff
@@ -1098,7 +1098,7 @@ public static double objectToDouble(Object value) {
          * Converts an Object to a {@code long} by checking it against known
          * types and checking its range.
          */
-        public static long objectToLong(Object value, boolean coerce) {
+        public static Long objectToLong(Object value, boolean coerce) {
```
nik9000, line 1101: I'd probably keep the return type on this one too.

server/src/test/java/org/elasticsearch/common/NumbersTests.java
```diff
@@ -16,20 +16,21 @@
...
-        assertEquals(9223372036854775807L, Numbers.toLong("9223372036854775807.99", true));
-        assertEquals(-9223372036854775808L, Numbers.toLong("-9223372036854775808.99", true));
+        assertThat(Numbers.toLong("3", false), equalTo(3L));
+        assertThat(Numbers.toLong("3.1", true), equalTo(3L));
+        assertThat(Numbers.toLong("9223372036854775807.00", false), equalTo(9223372036854775807L));
+        assertThat(Numbers.toLong("-9223372036854775808.00", false), equalTo(-9223372036854775808L));
+        assertThat(Numbers.toLong("9223372036854775807.00", true), equalTo(9223372036854775807L));
+        assertThat(Numbers.toLong("-9223372036854775808.00", true), equalTo(-9223372036854775808L));
+        assertThat(Numbers.toLong("9223372036854775807.99", true), equalTo(9223372036854775807L));
+        assertThat(Numbers.toLong("-9223372036854775808.99", true), equalTo(-9223372036854775808L));
```
nik9000, line 33: :+1:

server/src/main/java/org/elasticsearch/common/Numbers.java
rjernst: I moved this to return the boxed type so that we do not rely on the JIT to inline+elide this boxing. Otherwise we would parse into a primitive `long`, then need to box when actually inserting into the set. I don't feel strongly about this, though.

rjernst: I thought about it more, and I changed back to primitive.

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
rjernst: Changed back to primitive.

rjernst:
@nik9000 Can you take another look?

nik9000 (approved):
:boat:

PR #85423, opened 2022-03-28
Deprecate [indices.queries.cache.all_segments] setting

dnhatn:
The`indices.queries.cache.all_segments` setting should not be used in production environments. This change deprecates that setting.


jpountz (approved):
Thanks @dnhatn. I am not familiar enough with the settings infrastructure anymore to know whether this is the best way to do this, but I agree we should warn users when they do this.

dnhatn:
@jpountz Thanks for review. Are we okay to remove that setting and register it in a test plugin? I was worried that is a breaking change.

jtibshirani (approved):
This makes sense to me too. I'm wondering if we should mention "the setting will be removed in a future version", and if so we could make this a deprecation warning?

server/src/main/java/org/elasticsearch/indices/IndicesService.java
```diff
@@ -796,7 +796,7 @@ public synchronized void verifyIndexMetadata(IndexMetadata metadata, IndexMetada
             IndicesFieldDataCache indicesFieldDataCache = new IndicesFieldDataCache(settings, new IndexFieldDataCache.Listener() {
             });
             closeables.add(indicesFieldDataCache);
-            IndicesQueryCache indicesQueryCache = new IndicesQueryCache(settings);
+            IndicesQueryCache indicesQueryCache = new IndicesQueryCache(Settings.EMPTY);
```
jtibshirani, line 799: Did you mean to make this change?

jtibshirani (commented):
Expanding on my last comment -- what if we marked this setting as deprecated too, so we could remove it entirely in a future version? I think we do that by adding `Property.DEPRECATED` to the setting.

dnhatn:
>  what if we marked this setting as deprecated too, so we could remove it entirely in a future version

Yes, but that would require larger changes (i.e., handling deprecation logs in tests). I will make it in a follow-up as I prefer to have this warning in 8.2.0.

server/src/main/java/org/elasticsearch/indices/IndicesService.java
dnhatn: Yes, to avoid many warning/deprecation logs. The cache here is just a place holder.

dnhatn:
run elasticsearch-ci/part-1

jtibshirani (approved):
Thanks for the explanations, sounds good to get this warning into 8.2.

dnhatn:
We can leave the test setting as is.

PR #85433, opened 2022-03-29
fix running check command.

m-ostadi:
There is no config folder in /etc/elasticsearch so it was showing this error :
  curl: (77) error setting certificate verify locations:
  CAfile: /etc/elasticsearch/config/certs/http_ca.crt
  CApath: /etc/ssl/certs

changes:
correct cert address in command.
add required sudo.

### DIFF
--- a/docs/reference/setup/install/check-running.asciidoc
+++ b/docs/reference/setup/install/check-running.asciidoc
@@ -5,7 +5,7 @@ You can test that your {es} node is running by sending an HTTPS request to port
 
 ["source","sh",subs="attributes"]
 ----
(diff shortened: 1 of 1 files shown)


MOstadi:
i signed the Contributor Agreement

lockewritesdocs:
@MOstadi, thanks for pointing out this issue! We fixed the directory in #86390 so that the path is correct in the installation docs for [Debian](https://www.elastic.co/guide/en/elasticsearch/reference/current/deb.html#deb-check-running) and RPM. Regarding `sudo`, we intentionally do not include that in command examples because it doesn't apply to most environments.

PR #85332, opened 2022-03-24
[ML] Null out deployment stats fields that have not been used

davidkyle:
When inference has not been used on a node in a trained model deployment certain fields in the deployment stats response should not be set. 

Last access time was incorrectly being reported as Epoch 0 seconds and Average Inference time last period was 0.0 when it should be null if it has not been used. 

Non issue as this fixes a regression that has not been released yet.

PR #85436, opened 2022-03-29
[ML] Fix missing deployment stats peak throughput field

davidkyle:
In an edge case `peak_throughput_per_minute` was not being returned even if the stat could be calculated for the last bucket.

Follow up to #85332


dimitris-athanasiou (approved):
LGTM

PR #85058, opened 2022-03-17
fix: nested top metrics sort on keyword field

salvatore-campagna:
Using a double as a return value works only if the field we are
sorting on is a number. If the field is not a value we can convert
to a double, like a non-numeric keyword, converting it to a number
returns `NaN`. Without this patch, sorting takes place on the bucket
key, if the order field points to a non-numeric value. The additional
bucket key comparator is implicitly added as a tie breaker to avoid
non-deterministic sorting of buckets.

With this change we support sorting using any subclass of SortValue.
This means the bucket key will be used just in case of equal values
on the order field.

Issue: #78506

PR #85437, opened 2022-03-29
fix: make EmptySortValue instances sort after everything else

salvatore-campagna:
When comparing instances of SortValue we make sure that instances
of subclass EmptySortValue sort after everything else. This is
also to be consistent with the way Double.NaN values are sorted
with respect to Double values.

Additionally we make sure to create an instance of EmptySortValue
when creating an instance from a Double.NaN value.

This PR is intented to fix a potential issue introduced by #85058.

### DIFF
--- a/server/src/main/java/org/elasticsearch/search/sort/SortValue.java
+++ b/server/src/main/java/org/elasticsearch/search/sort/SortValue.java
@@ -125,7 +125,7 @@ public abstract class SortValue implements NamedWriteable, Comparable<SortValue>
     @Override
     public abstract String toString();
 
-    // Force implementations to override sortOrder and sort each subclass by type first
+    // Force implementations to override compareToDifferentType and sort each subclass by type first
     protected abstract int compareToDifferentType();
 
     /**
--- a/server/src/main/java/org/elasticsearch/search/sort/SortValue.java
+++ b/server/src/main/java/org/elasticsearch/search/sort/SortValue.java
@@ -123,7 +123,7 @@ public abstract class SortValue implements NamedWriteable, Comparable<SortValue>
     @Override
     public abstract String toString();
 
-    // Force implementations to override compareToDifferentType and sort each subclass by type first
+    // Force implementations to override typeComparisonKey and associate each subclass with an integer key
     protected abstract int typeComparisonKey();
 
     /**
--- a/server/src/main/java/org/elasticsearch/search/sort/SortValue.java
+++ b/server/src/main/java/org/elasticsearch/search/sort/SortValue.java
@@ -19,15 +19,12 @@ import org.elasticsearch.xcontent.XContentBuilder;
 
 import java.io.IOException;
 import java.util.Arrays;
-import java.util.Comparator;
 import java.util.List;
 
 /**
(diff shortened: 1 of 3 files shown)


salvatore-campagna:

salvatore-campagna:
#85058 was already merged so Ideally I would like this to land in `v8.2.0`.

nik9000 (approved):
LGTM.

I'd prefer to stop sorting by the name soon, but there isn't any incentive to rush that. Sorting by the name was totally my bad idea. It's amazing how long it takes for somethings to come back to haunt you. Well, in that case it came back to haunt you and not me. But still, you get the idea.

Also - I don't like the name "valueless". `Empty`, I think. Are you cool reverting the name change when we add the sort field?

server/src/main/java/org/elasticsearch/search/sort/SortValue.java
```diff
@@ -74,16 +80,7 @@ private SortValue() {
...
-         * It might make sense to try and compare doubles to longs
-         * *carefully* to get a real sort. but it might not. For now
-         * we sort all doubles before all longs.
-         */
-        int typeCompare = getWriteableName().compareTo(other.getWriteableName());
-        if (typeCompare != 0) {
-            return typeCompare;
-        }
-        return compareToSameType(other);
+        return VALUE_COMPARATOR.compare(this, other);
```
nik9000, line 83: I think I'd just do the comparisons directly here. I'm not sure what the comparators buys.

server/src/main/java/org/elasticsearch/search/sort/SortValue.java
```diff
@@ -128,6 +125,9 @@ public final XContentBuilder toXContent(XContentBuilder builder, DocValueFormat
     @Override
     public abstract String toString();
 
+    // Force implementations to override compareToDifferentType and sort each subclass by type first
+    protected abstract int compareToDifferentType();
```
nik9000, line 129: maybe `typeComparisonKey` is a better name.

server/src/main/java/org/elasticsearch/search/sort/SortValue.java
salvatore-campagna: I wanted to avoid creation of the comparator every time a comparison is made.

server/src/main/java/org/elasticsearch/search/sort/SortValue.java
salvatore-campagna: This name was meant to be the counterpart of `compareToSameType`. I thought it just reads nice like this:
```
private static final Comparator<SortValue> TYPE_COMPARATOR = Comparator.comparing(SortValue::compareToDifferentType);
    private static final Comparator<SortValue> VALUE_COMPARATOR = TYPE_COMPARATOR.thenComparing(SortValue::compareToSameType);
```

you can read it almost as "compare to different type and then compare to same type"

nik9000 (approved):
LGTM.

My main objection to that old method name was that it wasn't really doing a comparator - it was fetching a thing for comparators. It'd have confused me one day. Sorry for the trouble.

PR #85441, opened 2022-03-29
Quantize geo queries to remove  true negatives from search results

### LINKED ISSUE #40891: geo_shape 'intersects' query doesn't find neighbors sharing coordinates
As "discussed" [here](https://discuss.elastic.co/t/neighboring-touching-geo-shapes-not-found/175543) (thanks for the fast reply) geo_shape queries don't find neighbors.

This issue is related to #40085 but with direct neighbors not being found. Please see the screenshots below of an actual real-life dataset.

**Elasticsearch version**: 6.7.1
**JVM version**: 12
**OS version**: Debian

neighbors_1

neighbors_2

neighbors_3
neighbors_5

neighbors_4

iverase:
Latitudes and longitude values on a Elasticsearch index are encoded as integers. The way it works is it divides the latitude rand longitude value range by `1 << 32` which give us the scale of our encoded values. So for example all values between 0 (inclusive) and 4.190951585769653E-8 (exclusive) degrees are encoded to integer 0. When decoding the value (which is what we consider the quantize value) is always set to 0 degrees for that range. What it means in our encoding is that our latitudes are pulled south and our longitudes are pulled west.

On the other hand queries are not quantise and use the original values. Therefore because points are pulled south and west, it means that points on the south-west edges of a polygon might be pulled away from the polygon (true negatives), while points on the north-east edges might be pulled inside (false positives).  

This behaviour is wrong as results should be as accurate as possible, but in any case the should only have false positives and never true negatives as the former can be filter out later but true negatives are lost for ever. In addition the current behaviour might not meet the expectation of users, for example the current behaviour does not meet the expectation of intersection with neighbour polygons that shred coordinates (https://github.com/elastic/elasticsearch/issues/40891).

Therefore what we propose here is to quantise our query geometries so we match better the expectations of users and remove false negatives from our search result.

fixes #40891


iverase:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/index/mapper/GeoPointScriptFieldType.java
```diff
@@ -107,13 +102,8 @@ public Query existsQuery(SearchExecutionContext context) {
...
     @Override
-    public Query geoShapeQuery(Geometry shape, String fieldName, ShapeRelation relation, SearchExecutionContext context) {
-        final LatLonGeometry[] luceneGeometries = GeoShapeUtils.toLuceneGeometry(fieldName, context, shape, relation);
-        if (luceneGeometries.length == 0
-            || (relation == ShapeRelation.CONTAINS && Arrays.stream(luceneGeometries).anyMatch(g -> (g instanceof Point) == false))) {
-            return new MatchNoDocsQuery();
-        }
-        return new GeoPointScriptFieldGeoShapeQuery(script, leafFactory(context), fieldName, relation, luceneGeometries);
+    public Query geoShapeQuery(SearchExecutionContext context, String fieldName, ShapeRelation relation, LatLonGeometry... geometries) {
+        return new GeoPointScriptFieldGeoShapeQuery(script, leafFactory(context), fieldName, relation, geometries);
```
craigtaverner, line 106: Are we loosing an optimisation here by deleting the check for any geometries not being Points when we have CONTAINS?

craigtaverner: Or perhaps this is fixing a bug, because the previous condition of 'return no results if the relation is CONTAINS and there are any non-points in the search' seems a bit unreasonable to me.

server/src/main/java/org/elasticsearch/index/mapper/GeoShapeQueryable.java
```diff
@@ -8,28 +8,215 @@
...
+    default Query geoShapeQuery(SearchExecutionContext context, String fieldName, ShapeRelation relation, Geometry shape) {
+        final LatLonGeometry[] luceneGeometries = toQuantizeLuceneGeometry(fieldName, context, shape, relation);
+        if (luceneGeometries.length == 0) {
+            return new MatchNoDocsQuery();
+        }
+        return geoShapeQuery(context, fieldName, relation, luceneGeometries);
+    }
 
     @Deprecated
     default Query geoShapeQuery(
```
craigtaverner, line 55: This function can be deleted. It is the same as the function it calls, but with an extra field that is always ignored.

server/src/main/java/org/elasticsearch/index/query/GeoBoundingBoxQueryBuilder.java
```diff
@@ -303,7 +303,7 @@ public Query doToQuery(SearchExecutionContext context) {
             luceneTopLeft.getLat(),
             luceneBottomRight.getLat()
         );
-        return geoShapeQueryable.geoShapeQuery(rectangle, fieldType.name(), SpatialStrategy.RECURSIVE, ShapeRelation.INTERSECTS, context);
+        return geoShapeQueryable.geoShapeQuery(context, fieldType.name(), SpatialStrategy.RECURSIVE, ShapeRelation.INTERSECTS, rectangle);
```
craigtaverner, line 306: Just call the new API directly.

craigtaverner:
```suggestion
        return geoShapeQueryable.geoShapeQuery(context, fieldType.name(), ShapeRelation.INTERSECTS, rectangle);
```

server/src/main/java/org/elasticsearch/index/query/GeoDistanceQueryBuilder.java
```diff
@@ -244,7 +244,7 @@ protected Query doToQuery(SearchExecutionContext context) throws IOException {
 
         final GeoShapeQueryable geoShapeQueryable = (GeoShapeQueryable) fieldType;
         final Circle circle = new Circle(center.lon(), center.lat(), this.distance);
-        return geoShapeQueryable.geoShapeQuery(circle, fieldType.name(), SpatialStrategy.RECURSIVE, ShapeRelation.INTERSECTS, context);
+        return geoShapeQueryable.geoShapeQuery(context, fieldType.name(), SpatialStrategy.RECURSIVE, ShapeRelation.INTERSECTS, circle);
```
craigtaverner, line 247:
```suggestion
        return geoShapeQueryable.geoShapeQuery(context, fieldType.name(), ShapeRelation.INTERSECTS, circle);
```

server/src/main/java/org/elasticsearch/index/query/GeoShapeQueryBuilder.java
```diff
@@ -178,7 +178,7 @@ public Query buildShapeQuery(SearchExecutionContext context, MappedFieldType fie
             );
         }
         final GeoShapeQueryable ft = (GeoShapeQueryable) fieldType;
-        return new ConstantScoreQuery(ft.geoShapeQuery(shape, fieldType.name(), strategy, relation, context));
+        return new ConstantScoreQuery(ft.geoShapeQuery(context, fieldType.name(), strategy, relation, shape));
```
craigtaverner, line 181:
```suggestion
        return new ConstantScoreQuery(ft.geoShapeQuery(context, fieldType.name(), relation, shape));
```

craigtaverner (approved):
I do think we could remove the default deprecated method as it is not used anywhere that cannot be changed to use the new method. Or is this a public API?

server/src/main/java/org/elasticsearch/index/mapper/GeoShapeQueryable.java
iverase: See `LegacyGeoShapeFieldMapper` where the parameter SpatialStrategy is needed.  This is a legacy parameter that we need to carry over until we remove the old way of working with geo_shape field.

server/src/main/java/org/elasticsearch/index/query/GeoDistanceQueryBuilder.java
iverase: Needed for compatibility with `LegacyGeoShapeFieldMapper`

server/src/main/java/org/elasticsearch/index/query/GeoShapeQueryBuilder.java
iverase: Needed for compatibility with `LegacyGeoShapeFieldMapper`

server/src/main/java/org/elasticsearch/index/query/GeoBoundingBoxQueryBuilder.java
iverase: Needed for compatibility with `LegacyGeoShapeFieldMapper`

server/src/main/java/org/elasticsearch/index/mapper/GeoPointScriptFieldType.java
iverase: I recover the optimization. the only way for an indexed point to contain a query shape is that the query shape is a point or multi-point?

server/src/main/java/org/elasticsearch/index/mapper/GeoShapeQueryable.java
craigtaverner: Except the parameter is completely ignored in the implementation, so it can also be deleted from `LegacyGeoShapeFieldMapper`, right?

PR #85120, opened 2022-03-18
Support GeoJSON for geo_point

craigtaverner:
Support GeoJSON for points when the mapper specifies geo_point.

Closes #47815
Closes #84599

PR #85442, opened 2022-03-29
Support 'GeoJSON' in CartesianPoint for 'point'

craigtaverner:
This is done by using a common parser with GeoPoint (and 'geo_point').

In addition, we refactored away the pattern of creating the point high
in the stack and modifying it deep in the stack. Instead, we create the
points low in the stack. This leads to much simpler code. There are a
few cases of increased object creation, but they seem very slight:

- The GeoBoundingBox class uses it to extract the bounds
- The GeoPointFieldScript used a shared point for repeated use

This PR enhanced the work done in https://github.com/elastic/elasticsearch/pull/85120

### DIFF
--- a/server/src/main/java/org/elasticsearch/common/geo/GenericPointParser.java
+++ b/server/src/main/java/org/elasticsearch/common/geo/GenericPointParser.java
@@ -1,0 +1,271 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.common.geo;
+
+import org.elasticsearch.ElasticsearchParseException;
+import org.elasticsearch.xcontent.XContentParser;
+import org.elasticsearch.xcontent.XContentSubParser;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.function.Function;
+
+import static org.elasticsearch.common.Strings.collectionToDelimitedString;
+
+/**
+ * This class can parse points from XContentParser and supports several formats:
+ * @param <T> type of point to produce (either geographic or cartesian)
+ */
+public abstract class GenericPointParser<T> {
+    private static final String X_FIELD = "x";
+    private static final String Y_FIELD = "y";
+    private static final String Z_FIELD = "z";
+    private static final String GEOHASH = "geohash";
+    private static final String TYPE = "type";
+    private static final String COORDINATES = "coordinates";
+    private final Map<String, FieldParser<?>> fields;
+    private final String mapType;
+    private final String xField;
+    private final String yField;
+
+    private abstract static class FieldParser<T> {
+        final String name;
+
+        abstract T parseField(XContentSubParser subParser) throws IOException;
+
+        private FieldParser(String name) {
+            this.name = name;
+        }
+
+        @Override
+        public String toString() {
+            return this.name;
+        }
+    }
+
+    private static class StringFieldParser extends FieldParser<String> {
+
+        private StringFieldParser(String name) {
+            super(name);
+        }
+
+        @Override
+        String parseField(XContentSubParser subParser) throws IOException {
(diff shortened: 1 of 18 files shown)


server/src/main/java/org/elasticsearch/script/GeoPointFieldScript.java
```diff
@@ -80,8 +80,6 @@ public interface LeafFactory {
         GeoPointFieldScript newInstance(LeafReaderContext ctx);
     }
 
-    private final GeoPoint scratch = new GeoPoint();
-
```
iverase, line 84: I think we should not be creating a GeoPoint per document here. I wonder if this case invalidates the approach of creating the objects at the end of the stack.

server/src/main/java/org/elasticsearch/common/geo/GenericPointParser.java
```diff
@@ -0,0 +1,271 @@
...
+
+/**
+ * This class can parse points from XContentParser and supports several formats:
+ * @param <T> type of point to produce (either geographic or cartesian)
+ */
+public abstract class GenericPointParser<T> {
+    private static final String X_FIELD = "x";
+    private static final String Y_FIELD = "y";
+    private static final String Z_FIELD = "z";
+    private static final String GEOHASH = "geohash";
```
iverase, line 33: geohash only makes sense for geo which makes it difficult to have it as a generic?

server/src/main/java/org/elasticsearch/common/geo/GenericPointParser.java
```diff
@@ -0,0 +1,271 @@
...
+
+import static org.elasticsearch.common.Strings.collectionToDelimitedString;
+
+/**
+ * This class can parse points from XContentParser and supports several formats:
+ * @param <T> type of point to produce (either geographic or cartesian)
+ */
+public abstract class GenericPointParser<T> {
+    private static final String X_FIELD = "x";
+    private static final String Y_FIELD = "y";
```
iverase, line 31: I would expect this pameters to be`ParseField` objects, any reason why you prefer not to use them?

iverase (changes requested):
I am having troubles with this approach as I see two main issues:

~~1) The idea of a generic point parser is nice but it breaks on the geohash parameter. This only makes sense in Geo so we might need to add special handling for this.~~ I see we can handle this case easily

~~2) Creating a point per parse call simplifies the code but it is going to make the runtime field more inefficient in some cases. I am wondering if that invalidates the approach. Maybe we should still allow the current approach?~~ We spoke offline and in this case, we are parsing the point from source so the effect can be considered negligible.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/common/CartesianPoint.java
```diff
@@ -157,124 +156,33 @@ public String toString() {
...
-            String val = parser.text();
-            return point.resetFromString(val, ignoreZvalue);
-        } else {
-            throw new ElasticsearchParseException("point expected");
-        }
+        return cartesianPointParser.parsePoint(parser, ignoreZValue, value -> {
+            CartesianPoint point = new CartesianPoint();
+            point.resetFromString(value, ignoreZValue);
+            return point;
+        }, value -> null);
```
iverase, line 182: Instead of null, Should be throw an IllegalArgumentException here?

server/src/main/java/org/elasticsearch/common/geo/GenericPointParser.java
iverase: see https://github.com/elastic/elasticsearch/pull/85442#discussion_r855104508

server/src/main/java/org/elasticsearch/common/geo/GenericPointParser.java
iverase: fair anough

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/common/CartesianPoint.java
craigtaverner: It was already throwing an exception. The list of valid fields is checked much earlier. I've added a number of tests to assert on the actual exception that should be thrown for all the possible invalid formats. One thing the new test shows is that seemingly similar things (like test for NaN) are tested quite differently (and produce different error messages) for different code paths (geojson parser, compared to string parser, compared to array parser). I've not attempted to unify the error messages, for backwards compatibility, but just assert that the expected errors are thrown.

iverase (approved):
LGTM

PR #85018, opened 2022-03-16
Add test showing over allocation of partially/fully mounted indices on cold/frozen node

tlrx:
This pull request adds a basic test to show that partially/fully mounted indices can be overallocated to a frozen/cold data node if the disk usages cluster information are not recent enough.

PR #85445, opened 2022-03-29
Sync cluster service information when mounting a searchable snapshot

arteam:
Because of the delay in updating the cluster service info (happens every in 15s), we can have outdated information about the free space on disk in `DiskThresholdDecider` when we check whether it's possible to allocate an index. One possible solution for this would be to try to refresh the cluster service info during the preparation of a new searchable snapshot.

Fixes #85018


arteam:
@elasticmachine update branch

arteam:
@elasticmachine update branch

arteam:
@elasticmachine update branch

x-pack/plugin/searchable-snapshots/src/main/java/org/elasticsearch/xpack/searchablesnapshots/action/TransportMountSearchableSnapshotAction.java
```diff
@@ -250,6 +254,8 @@ protected void masterOperation(
                 dataTierAllocationSetting.get(indexSettings);
             }
 
+            clusterInfoService.syncClusterInfo();
```
fcofdez, line 257: I think this do not fixes the issue, at this point the index is not mounted yet. I think this pass the test as the implementation in `InternalClusterInfoService#syncClusterInfo` triggers an async operation and at the point it refreshes the information the index is already mounted I think?

PR #85446, opened 2022-03-29
[ML] fix bug where certain trained model deployments failed with "Object cannot be set twice!"

benwtrent:
When an allocation fails for a given model, we retry that allocation. We only do this in certain failure paths.

But, if an allocation makes it as far as setting the model config, we will attempt to set the model config again on the same task instance. This will throw a Lucene `Object cannot be set twice!` exception.

This commit addresses this bug by using a `trySet`, which is atomic and safe. And then starts tracking the model usage once the config is set.


davidkyle (approved):
LGTM

benwtrent:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.1|[](https://github.com/elastic/elasticsearch/pull/85451)|

### Questions ?

PR #85447, opened 2022-03-29
Disable dynamic mapping in Stack Monitoring index templates

miltonhultgren:
This aligns the new Metricbeat templates with the settings found in the older
internal collection templates. By disabling dynamic mapping we avoid the
possible field explosion that can happen for example when there are lots of
nodes in the Elasticsearch cluster stats response which could easily blow past
the 2 000 field limit.

This commit also adds the missing ILM policy to the Enterprise search mappings.

### DIFF
--- a/x-pack/plugin/core/src/main/resources/monitoring-beats-mb.json
+++ b/x-pack/plugin/core/src/main/resources/monitoring-beats-mb.json
@@ -3,6 +3,7 @@
   "version": ${xpack.stack.monitoring.template.release.version},
   "template": {
     "mappings": {
+      "dynamic": false,
       "properties": {
         "beat": {
           "properties": {
@@ -2237,6 +2238,7 @@
       }
     },
     "settings": {
+      "index.mapping.total_fields.limit": 2000,
       "index.lifecycle.name": "${xpack.stack.monitoring.policy.name}"
     }
   },
--- a/x-pack/plugin/core/src/main/resources/monitoring-ent-search-mb.json
+++ b/x-pack/plugin/core/src/main/resources/monitoring-ent-search-mb.json
@@ -7,6 +7,7 @@
   },
   "template": {
     "mappings": {
+      "dynamic": false,
       "properties": {
         "enterprisesearch": {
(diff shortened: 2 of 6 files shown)


miltonhultgren:
@elastic/infra-monitoring-ui

miltonhultgren:
Feedback/help on how to label this is very welcome!

carlosdelest:
LGTM on the enterprise-search side 👍

jbaiera (approved):
The change LGTM

Had a question regarding the root of the issue:

> By disabling dynamic mapping we avoid the
possible field explosion that can happen for example when there are lots of
nodes in the Elasticsearch cluster stats response which could easily blow past
the 2 000 field limit.

Should this possibility for field explosion also be addressed on the Metricbeat side of things?

miltonhultgren:
@jbaiera Indeed, I suspect Metricbeat has had this possibility for field explosion all along since the large template it uses has dynamic mapping enabled. It does use a 10 000 field limit (to allow for mapping multiple modules) which makes it less likely to notice, in addition users must enable the `cluster_stats` metricset and set `scope: cluster` in their config for this data to get ingested. 

We'll look into disabling dynamic mapping for our modules in the Metricbeat mapping!

dakrone:
@miltonhultgren since this is a bug fix, should this be backported to the 8.1 branch also?

miltonhultgren:
@jbaiera Can this still be backported to 8.1.X? I see in the release schedule that both feature freeze and build candidate dates have passed for 8.1.2, will there be a 8.1.3?

jbaiera:
From what I can tell there is no 8.1.3 planned before 8.2 is released, so I don't know if this is vital to get into the 8.1 line. That said, if there is an emergency release for some reason it would get picked up in it if it were present.

dakrone:
Yep you're right, no 8.1.3, so it doesn't make sense to backport this to 8.1, however, this is marked as `v8.2.0` though it targets master (which is 8.3), so it *should* be backported to the 8.2 branch, because there are more releases planned there.

PR #85449, opened 2022-03-29
Format runtime geo_points

### LINKED ISSUE #85245: Fields API not returning GeoJson for runtime geo_point field
### Elasticsearch Version

main

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

macos

### Problem Description

geo_point runtime fields are returning values as "lat,lon" strings  instead of GeoJSON. 

### Steps to Reproduce

Run the commands below in Kibana console. Notice the response how `test_location` is returned as `"-37.91000002529472, 144.6599999908358"` instead of GeoJSON

Create index
```
PUT test
{}

PUT test/_doc/1
{}
```

Request
```
GET test/_search
{
  "_source": false,
  "fields": [
    "test_location"
  ],
  "runtime_mappings": {
    "test_location": {
      "type": "geo_point",
      "script": {
        "source": "emit(-37.91,144.66)"
      }
    }
  }
}
```

Response
```
{
  "took" : 0,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 1,
      "relation" : "eq"
    },
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : "test",
        "_id" : "1",
        "_score" : 1.0,
        "fields" : {
          "test_location" : [
            "-37.91000002529472, 144.6599999908358"
          ]
        }
      }
    ]
  }
}
```

### Logs (if relevant)

_No response_

nik9000:
This formats the result of the `fields` section of the `_search` API for
runtime `geo_point` fields using the `format` parameter like we do for
non-runtime `geo_point` fields. This changes the default format for
those fields from `lat, lon` to `geojson` with the option to get `wkt`
or any other format we support.

The fix does so by introducing a new method to `DocValueFormat` that
formats `GeoPoint` objects. It doesn't implement a properly serializable
`DocValueFormat` for `GeoPoint`s because one isn't required at this
point, but I don't see why we couldn't do that later on when we need
one.

Closes #85245


rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search/330_fetch_fields.yml
```diff
@@ -1064,3 +1064,65 @@ test fetching metadata fields:
   - match: { hits.hits.0.fields._id.0: "1" }
   - match: { hits.hits.0.fields._index.0: "test" }
   - match: { hits.hits.0.fields._version.0: 1 }
+
+---
+fetch geo_point:
```
nik9000, line 1069: You can run this in two ways:
```
./gradlew -p rest-api-spec/ check -Dtests.method='*fetch_fields*'
./gradlew -p x-pack/qa/runtime-fields/core-with-mapped/ check -Dtests.method='*fetch_fields*'
```

The former runs against indexed fields and passes without this change. The latter runs against runtime fields that load from `_source` and fails without this change.

nik9000:
> The test failure is because we had a test for fetching runtime geo points in yaml testing the bad format. Actually, we have two! #85452 tracks that.

I opened #85453 which I'd like to resolve before we merge this.

nik9000:
> The precision loss you've linked sounds more like it's caused by us using `double` instead of arbitrary precision for the parsing. I think that's the precision that runtime fields have. I could be wrong though.

I just talked with @iverase. Going through doc values adds an encoding step which pitches even more precision than regular old double precision floating point. So I'll rework this not to go through doc values.

server/src/main/java/org/elasticsearch/index/mapper/GeoPointScriptFieldType.java
```diff
@@ -127,11 +135,75 @@ public Query distanceFeatureQuery(Object origin, String pivot, SearchExecutionCo
         double pivotDouble = DistanceUnit.DEFAULT.parse(pivot, DistanceUnit.DEFAULT);
         return new GeoPointScriptFieldDistanceFeatureQuery(
             script,
-            leafFactory(context)::newInstance,
+            valuesEncodedAsLong(context.lookup(), name(), leafFactory(context)::newInstance),
```
nik9000, line 138: I flipped all of the other queries to taking two doubles. I probably should flip this one as well, but it's a bit more complex so I'm waiting on the first review.

iverase (commented):
This is looking very cool. One thing that is worrying is that we are changing the default behaviour for the default format. Currently if you use the fields API on a geo_point runtime field without specifying a format, the output would have the format "lat,"lon". With this change the output format will be GeoJson. Can this be considered a breaking change?

nik9000:
> Can this be considered a breaking change?

I admit to being a bit hazy still, but I've been considering this a bug. We'd intended runtime geo_points to have the same defaults as regular geo_points.

nik9000:
But it totally could be considered breaking.

server/src/main/java/org/elasticsearch/index/mapper/GeoPointScriptFieldType.java
iverase: Yes, this seems more complicated and I think this solution might make more sense. 

We need to quantize the values anyway so it behaves as in a normal field so what you are doing here makes lot of sense as we are only encoding values once.

iverase (approved):
LGTM

I think it is ok to use the approach you use for the distance feature query. We can switch on a follow up PR but if you want to implement it is ok too. Just make sure you encode /decode values so they match the values on an index.

server/src/main/java/org/elasticsearch/index/mapper/GeoPointScriptFieldType.java
iverase: It would be ok to flip it too but we need to make sure we quantize the values to match the doc values of an index.

nik9000:
> LGTM

For those following along at home I had a conversation with a few folks directly how this breaks and convinced them to consider this a bug fix. It does change the default format for *runtime* geo_point fields. But it changes it to be the same as regular geo_point fields. Which feels like we're fixing a bug. To me.

> I think it is ok to use the approach you use for the distance feature query. We can switch on a follow up PR but if you want to implement it is ok too. Just make sure you encode /decode values so they match the values on an index.

I'd prefer if you grabbed that one, yeah. Thanks so much!

PR #84710, opened 2022-03-07
Added buffer pixels to vector tile spec parsing

craigtaverner:
Previously this was hard-coded to 5, but now is configurable using the
format z/x/y@extent:buffer, where both extent and buffer are optional
and default to 4096 and 5 pixels respectively.

Relates #84492

PR #85450, opened 2022-03-29
Support vector tile buffer in REST API

### LINKED ISSUE #84492: Make vector tile buffer size configurable
Vector tiles buffer is used to include data that is just outside of the tile. This data is used to prevent border effects when rendering the geometries. It is currently [hardcoded to 5 pixels](https://github.com/elastic/elasticsearch/blob/de2781f98696c1e6cfad2b5bf46fe752a5e3edbb/x-pack/plugin/vector-tile/src/main/java/org/elasticsearch/xpack/vectortile/feature/FeatureFactory.java#L63) but it should probably need to be user configurable by adding a new property on the vector tiles API which can change the value.

craigtaverner:
In a previous PR https://github.com/elastic/elasticsearch/pull/84710
we supported the configurable buffer for field formatting in the mvt() function, but not in the REST API.
This commit adds the missing piece.

Closes https://github.com/elastic/elasticsearch/issues/84492

### DIFF
--- a/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/vector-tile/10_basic.yml
+++ b/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/vector-tile/10_basic.yml
@@ -53,6 +53,18 @@ setup:
           extent: 256
 
 ---
+"buffer param":
+  - do:
+      search_mvt:
+        index: locations
+        field: location
+        x: 0
+        y: 0
+        zoom: 0
+        body:
+          buffer: 10
+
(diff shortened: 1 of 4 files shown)


iverase (commented):
I think we need to add a test in `https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/vector-tile/10_basic.yml`, this is use for the clients to check the API against them.

Once we have that test we are good to go

iverase (approved):
LGTM

Issue #85312, opened 2022-03-24
The `superuser` built-in role doesn't work with an unavailable .security index


albertzaharovits:
### Elasticsearch Version

`8.*`, ~`7.17.*`~

### Problem Description

We used to think and advise that a file-based superuser can rescue any cluster with a botched `.security-*` index.
Alas this is not true, in general.
If the `.security-7` index exists, but it is closed or hasn't got the primary assigned, the `superuser` role assigned to a file-based user will not work because it includes (all) application privileges, which themselves are stored in the `.security-*` index and resolving them would fail the authentication.

### Steps to Reproduce

Disable role-related caching in `elasticsearch.yml`:
```
xpack.security.authz.store.privileges.cache.ttl: 0
xpack.security.authz.store.roles.cache.max_size: 0
xpack.security.authz.store.roles.negative_lookup_cache.max_size: 0
```
Add a new super role to the `config/roles.yml` that can close the `.security-7` index:
```
supersuperuser:
  run_as: [ '*' ]
  cluster: [ 'all' ]
  indices:
    - names: [ '*' ]
      privileges: [ 'all' ]
      allow_restricted_indices: true
```
Create a new file-based user:
```
./bin/elasticsearch-users useradd test -p password -r superuser,supersuperuser
```
Close the `.security-7` (in order to make it unavailable):
```
curl -k -u test:password -X POST "https://localhost:9200/.security-7/_close"
```

In this state, all file-based users with the `superuser` role AND some other role (could be an empty role) will fail to authenticate:
```
curl -k -u test:password -X GET "https://localhost:9200/_security/_authenticate?pretty"
{
  "error" : {
    "root_cause" : [
      {
        "type" : "index_closed_exception",
        "reason" : "closed",
        "index_uuid" : "_na_",
        "index" : ".security-7"
      }
    ],
    "type" : "index_closed_exception",
    "reason" : "closed",
    "index_uuid" : "_na_",
    "index" : ".security-7"
  },
  "status" : 400
}
```
with a sample stack trace extract:
```
    "stack_trace" : "[.security-7] org.elasticsearch.indices.IndexClosedException: closed
at org.elasticsearch.xpack.security.support.SecurityIndexManager.getUnavailableReason(SecurityIndexManager.java:137)
at org.elasticsearch.xpack.security.authz.store.NativePrivilegeStore.innerGetPrivileges(NativePrivilegeStore.java:182)
[...]

PR #85455, opened 2022-03-29
App permissions with action patterns do not retrieve privileges

albertzaharovits:
In order for file realm users, with application privileges, to continue
to work (authenticate) when the `.security` index is unavailable,
Security must avoid loading the application privileges from the .security index
(which is unavailable, hence error out).
Roles (defined inside the `roles.yml` file, not in the index) that must work
when the `.security` index is unavailable must use inlined action patterns
for application permisions, rather than privilege names.

Related https://github.com/elastic/elasticsearch/issues/85312


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/privilege/ApplicationPrivilege.java
```diff
@@ -162,10 +162,17 @@ public static void validatePrivilegeName(String name) {
         }
     }
 
-    private static boolean isValidPrivilegeName(String name) {
+    public static boolean isValidPrivilegeName(String name) {
         return VALID_NAME.matcher(name).matches();
     }
 
+    public static void validateActionName(String action) {
```
justincr-elastic, line 169: Does it make sense for a new `validateActionName` method to call a new `isValidActionName` method?

For example, there is an existing `validatePrivilegeName` method which calls `isValidPrivilegeName`. Does it make sense to follow that pattern?

In other words, I found it confusing to have a method called `validateActionName` which does validation on action name, but then also calls through to `validatePrivilegeOrActionName`.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/privilege/PutPrivilegesRequest.java
```diff
@@ -63,14 +63,8 @@ public ActionRequestValidationException validate() {
...
                 for (String action : privilege.getActions()) {
-                    if (action.indexOf('/') == -1 && action.indexOf('*') == -1 && action.indexOf(':') == -1) {
-                        validationException = addValidationError(
-                            "action [" + action + "] must contain one of [ '/' , '*' , ':' ]",
-                            validationException
-                        );
-                    }
                     try {
-                        ApplicationPrivilege.validatePrivilegeOrActionName(action);
+                        ApplicationPrivilege.validateActionName(action);
```
justincr-elastic, line 67: I found it confusing to have a method called `validateActionName` which does validation on action name, but then also calls through to `validatePrivilegeOrActionName`.

justincr-elastic (changes requested):
I would need more information to be able to complete a review, but I did have a feedback point.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/privilege/ApplicationPrivilege.java
albertzaharovits: > Does it make sense for a new validateActionName method to call a new isValidActionName method?

I am unsure what that would achieve.
The `validateActionName` throws two different kinds of errors at the moment. How can the proposal, as I understand it, achieve the same thing when just calling `isValidPrivilegeName`?

I have pushed dea850d34b7d04d1e338001 . Not sure if this helps or not.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/privilege/PutPrivilegesRequest.java
albertzaharovits: Do you also have proposals to improve it?
Also keeping in mind this is a bug-fix PR.
I only slightly changed code in this part to make sure the action names I generate IN TESTS are valid.

albertzaharovits:
This PR fixes the bug from https://github.com/elastic/elasticsearch/issues/85312 by making any file-based role that contains application permissions that do not refer to named application privileges (which only mention inlined action patterns, like the `superuser` role does) not unnecessarily query the `.security` index. By avoiding querying the index, such roles can be used even when the index is unavailable.

Another option to fix the same bug would be to always return the statically built `superuser` role, after failing to resolve a role name list that includes it. We do something like that today, when failing to retrieve any role descriptors, but not when failing to retrieve privileges referred in those role descriptors.

I think both approaches have pros-cons, but I don't think we have to choose between them.
I think it is perfectly reasonable for file-based roles to not reach the index if they don't need to, but if they do, and fail, we should not fail the authentication if the user also has `superuser`.

tvernum:
Hey @albertzaharovits - this approach looks like a good direction to me. 
Are you still looking for a review?

justincr-elastic (approved):
LGTM

albertzaharovits:
> Hey @albertzaharovits - this approach looks like a good direction to me.
Are you still looking for a review?

Yes, please review. I have adapted the PR description and label to reflect that this is now `>enhancement` rather than a `>bug` fix after you have merged https://github.com/elastic/elasticsearch/pull/85519 .

tvernum:
Given this is now an enhancement, I don't think we should backport to `7.17`

tvernum (approved):
LGTM

albertzaharovits:
@elasticmachine update branch

PR #85457, opened 2022-03-29
Correct prefix in docs for ILM fully-mounted searchable snapshots

### LINKED ISSUE #85432: [Doc][Bug] Searchable snapshot doesn't have "recovered-" prefix but have "restored-" in 8.x
### Elasticsearch Version

8.x

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

supported major OS

### Problem Description

In manual, there's such description
https://www.elastic.co/guide/en/elasticsearch/reference/8.1/searchable-snapshots.html

> Indices managed by ILM are prefixed with `recovered-` when fully mounted.

Screenshot:

<details>

image

</details>

However, when testing the behavior, it's actually not using the prefix `recovered-`, but instead, it's using `restored-`.

image

---

The prefix `restored-` is the same as 7.x, which is mentioned below.
https://www.elastic.co/guide/en/elasticsearch/reference/7.17/ilm-searchable-snapshot.html

Also additionally, the following page 8.x version also shows the correct spelling.
https://www.elastic.co/guide/en/elasticsearch/reference/8.1/ilm-searchable-snapshot.html

---

Based on that, seemingly it's a simple typo when that line was added into the manual and it caused the confusion.
https://www.elastic.co/guide/en/elasticsearch/reference/8.1/searchable-snapshots.html#searchable-snapshot-mount-storage-options

### Steps to Reproduce

as mentioned above.

Can we have someone double check if this is a simple doc typo please?
I am happy to file a doc fix PR but want to check about the tech part at first.

### Logs (if relevant)

_No response_

dakrone:
This incorrectly had "recovered-" when "restored-" is what is actually used.

Resolves #85432

### DIFF
--- a/docs/reference/searchable-snapshots/index.asciidoc
+++ b/docs/reference/searchable-snapshots/index.asciidoc
@@ -133,7 +133,7 @@ retrieve the data needed to complete the search in parallel with the ongoing
 recovery. On-disk data is preserved across restarts, such that the node does
 not need to re-download data that is already stored on the node after a restart.
 +
-Indices managed by {ilm-init} are prefixed with `recovered-` when fully mounted.
+Indices managed by {ilm-init} are prefixed with `restored-` when fully mounted.
 
 [[partially-mounted]]
 Partially mounted index::


jbaiera (approved):
LGTM

PR #85458, opened 2022-03-29
Vector disk usage

ywelsch:
Adds disk usage information for knn vectors.


ywelsch:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/action/admin/indices/diskusage/IndexDiskUsageAnalyzer.java
```diff
@@ -526,6 +532,38 @@ void visitField(Fields vectors, String fieldName) throws IOException {
...
+                // do a couple of randomized searches to figure out min and max offsets of index file
+                VectorValues vectorValues = vectorReader.getVectorValues(field.name);
+                int numDocsToVisit = reader.maxDoc() < 1000 ? reader.maxDoc() : 1000 * (int) Math.log10(reader.maxDoc());
+                int skipFactor = reader.maxDoc() / numDocsToVisit;
+                for (int i = 0; i < reader.maxDoc(); i += skipFactor) {
+                    if ((i = vectorValues.advance(i)) == DocIdSetIterator.NO_MORE_DOCS) {
+                        break;
+                    }
+                    cancellationChecker.checkForCancellation();
+                    vectorReader.search(field.name, vectorValues.vectorValue(), 1, null, 20);
```
jtibshirani, line 558: A `visitedLimit` of 20 is quite low and might mean that you only end up visiting the upper layers to the graph over and over again, instead of exploring the bottom layer (the full nearest neighbor graph). Also setting `k=1` means the search could search ends really early.

Maybe something like `k=1000, visitedLimit=1000` would make this more stable? Or some function of the number of docs?

jtibshirani:
@ywelsch do you have time / interest to pick this up again? It'd be great to get it in, we've gotten questions from users about kNN vectors usage.

ywelsch:
@jtibshirani It's not high on my list at the moment. I addressed your round of feedback, but I think this needs more at-scale validation to see if it performs well enough on larger data sets. This is quite time-consuming, and I would be happy for someone else to pick the PR up from here.

jtibshirani:
Got it, we'll see if someone else has the time! It may less effort for me or Mayya, as we have local test set-ups on large vector datasets.

javanna:
heya @jdconrad would you like to finish this up? cc @benwtrent

jdconrad:
@jdconrad Sure, I'll see if I can get this up to date soon.

PR #85465, opened 2022-03-29
[ML] disallow new trained model deployments when nodes are different versions

benwtrent:
Since trained model deployments are experimental. Backwards compatibility is not always guaranteed between versions. 

To protect users, this commit disallows new deployments to be created in a mixed versioned cluster environment.

### DIFF
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportStartTrainedModelDeploymentAction.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportStartTrainedModelDeploymentAction.java
@@ -154,7 +154,7 @@ public class TransportStartTrainedModelDeploymentAction extends TransportMasterN
         if (state.nodes().getMaxNodeVersion().after(state.nodes().getMinNodeVersion())) {
             listener.onFailure(
                 new ElasticsearchStatusException(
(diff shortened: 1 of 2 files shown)


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportStartTrainedModelDeploymentAction.java
```diff
@@ -151,6 +151,17 @@ protected void masterOperation(
             return;
         }
 
+        if (state.nodes().getMaxNodeVersion().after(state.nodes().getMinNodeVersion())) {
+            listener.onFailure(
+                new ElasticsearchStatusException(
+                    "Cannot not start a new model deployment as not all nodes are on version {}. All nodes must be the same version",
```
przemekwitek, line 157:
```suggestion
                    "Cannot start a new model deployment as not all nodes are on version {}. All nodes must be the same version",
```

przemekwitek (approved):
One typo,
apart from that LGTM

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportStartTrainedModelDeploymentAction.java
benwtrent: lulz, double negative :D

PR #85470, opened 2022-03-29
Migrate all uses of hppc BitMixer to Lucene

rjernst:
Lucene has its own copy of BitMixer. Rather than giving Elasticsearch
yet another copy of these functions, this commit converts the uses to
Lucene's BitMixer.

relates #84735


benwtrent (approved):

rjernst:

PR #85473, opened 2022-03-29
Serialize interval in auto date histogram aggregation

javanna:
The auto interval is computed as part of the reduction phase, initialized to 1 and never serialized over the wire.
That works in most cases as the node that executes the reduction is also the node returning the response, hence the response with the right interval set will be serialized to xcontent and sent back to the user. In case the response goes again through the wire, the interval is lost though. This happens when using async search, in which case the response is saved in binary format to an internal index and later retrieved. The saved response will always hold 1 as the auto interval.

This commit serializes the interval to the wire, so that it does not get lost when the response goes to another node, or is saved to the async search index.


server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/InternalAutoDateHistogram.java
```diff
@@ -234,13 +239,19 @@ protected void doWriteTo(StreamOutput out) throws IOException {
...
-    public DateHistogramInterval getInterval() {
+    long getBucketInnerInterval() {
+        return bucketInnerInterval;
+    }
 
+    public DateHistogramInterval getInterval() {
         RoundingInfo roundingInfo = this.bucketInfo.roundingInfos[this.bucketInfo.roundingIdx];
         String unitAbbreviation = roundingInfo.unitAbbreviation;
-        return new DateHistogramInterval(Long.toString(bucketInnerInterval) + unitAbbreviation);
+        return new DateHistogramInterval(bucketInnerInterval + unitAbbreviation);
```
not-napoleon, line 254: I'm nitpicking, but I think explicitly calling `Long.toString` is more readable.

not-napoleon (approved):
LGTM

nik9000 (approved):

server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/InternalAutoDateHistogram.java
javanna: my IDE told me to remove it, so did I :)

not-napoleon: Java: "Operator overloading is terrible because C++ overloaded bit shift to mean print!"
Also Java: "`long` + `String` means concatenate, lol"

At any rate, I don't care enough to argue about it.

javanna:
Thanks @nik9000 & @not-napoleon ! I am delaying backporting it for now to prevent misalignment between the version this fix goes out with and the added version conditionals.

PR #85475, opened 2022-03-30
Use Set instead of LongSet in double script field

rjernst:
The double script field uses hppc's LongSet to store the unique values
to be queried. However, this set should be relatively small. This commit
changes the internals of the runtime double script field to use Set.

relates #84735


server/src/main/java/org/elasticsearch/search/runtime/DoubleScriptFieldTermsQuery.java
```diff
@@ -66,8 +64,8 @@ public boolean equals(Object obj) {
     double[] terms() {
         double[] result = new double[terms.size()];
         int i = 0;
-        for (LongCursor lc : terms) {
-            result[i++] = Double.longBitsToDouble(lc.value);
+        for (long l : terms) {
```
dnhatn, line 67: nit: maybe replace this with stream/mapToDouble.

dnhatn (approved):
LGTM. Thanks Ryan!

PR #85476, opened 2022-03-30
Cleanup settings preparer

rjernst:
The settings preparer builds the initial environemnt and settings, and
applies settings from several sources. This commit cleans up the method
to have a clear set of steps that are applied to produce the final
settings.


grcevski (approved):
LGTM!

PR #85479, opened 2022-03-30
remove suggest flag

anmironov:
we have a parameter suggest, it is removed since 7.0.0 https://github.com/elastic/elasticsearch/pull/29635. But I am unsure

### DIFF
--- a/docs/reference/rest-api/common-parms.asciidoc
+++ b/docs/reference/rest-api/common-parms.asciidoc
@@ -515,9 +515,6 @@ of each Lucene index file.
 `store`::
 Size of the index in <<byte-units, byte units>>.
 
-`suggest`::
-<<search-suggesters,Suggester>> statistics.
-
 `translog`::
 <<index-modules-translog,Translog>> statistics.
 


joegallo:
Okay, tracing this through backwards and writing up my thought process:

1. `suggest` where you've removed it is part of the `index-metrics` tag. 
2. And that's only used in one place, the indices stats page:
    ```
    joegallo@galactic:~/Code/elastic/elasticsearch $ git grep 'index-metric'
    docs/reference/indices/stats.asciidoc:`GET /<target>/_stats/<index-metric>`
   docs/reference/indices/stats.asciidoc:include::{es-repo-dir}/rest-api/common-parms.asciidoc[tag=index-metric]
   ```
3. Which is https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-stats.html, or `GET /{someindex}/_stats`
4. Testing `GET /myindex/_stats/suggest` does indeed result in an error, because we removed that in https://github.com/elastic/elasticsearch/pull/29635, as you correctly indicated.

So, yeah, +1, that makes sense to me.

While I was there, I noticed a couple of other things that were off, too, and I'll bring those up with the relevant stakeholders and see to fixing them. Thanks for bringing this to our attention and submitting a pull request!

joegallo (approved):

PR #85488, opened 2022-03-30
Make reading NamedWritable more efficient

original-brownbear:
Just a random find from many-shards benchmarks:

Reading a list of these for a given category shouldn't have to do the lookup
over and over. Also the cold exception paths are moved out of the read methods
and lookups use more compact and faster maps now.

Also this adds an assertion on hitting a broken `Writable.Reader` implementation that returns `null` because that's simply a bug.


server/src/main/java/org/elasticsearch/common/io/stream/NamedWriteableAwareStreamInput.java
```diff
@@ -28,6 +32,21 @@ public NamedWriteableAwareStreamInput(StreamInput delegate, NamedWriteableRegist
...
+    public <T extends NamedWriteable> List<T> readNamedWriteableList(Class<T> categoryClass) throws IOException {
+        int count = readArraySize();
+        if (count == 0) {
+            return Collections.emptyList();
+        }
+        final Map<String, Writeable.Reader<?>> readers = namedWriteableRegistry.getReaders(categoryClass);
+        List<T> builder = new ArrayList<>(count);
+        for (int i = 0; i < count; i++) {
+            final String name = readString();
+            builder.add(NamedWriteableRegistry.getReader(categoryClass, name, readers).read(this));
```
DaveCTurner, line 45: Would it save anything to check whether `name` matches the `name` from the previous iteration and re-use the previous reader instead of doing another lookup? I expect in practice we see the same `name` repeatedly.

original-brownbear: >  I expect in practice we see the same name repeatedly.

I'm not so sure about that. In queries for example (which is pretty the one thing where this gets really hot) that doesn't seem to be the case but could be I'm wrong there. The thing is, you trade a string comparison for a map lookup. That saves you hashing `name` (which is a freshly read string so you have to compute the hash code) if you're right, but if you're wrong you get hashing and 2 comparisons. Given how cheap the map lookup is for the immutable maps you'd have to be right a ton here to justify this optimization.
Also, by adding that branching you made the loop more complicated which might have some negative side effects as well in terms of JIT/inlining.

DaveCTurner: Fair enough, makes sense.

original-brownbear:
Jenkins run elasticsearch-ci/rest-compatibility

original-brownbear:
@elasticmachine update branch

original-brownbear:

DaveCTurner (approved):
LGTM

original-brownbear:
Thanks David!

Issue #68468, opened 2021-02-03
Generic threads deadlock related to ILMHistoryStore indexing


DaveCTurner:
**Elasticsearch version** (`bin/elasticsearch --version`): 7.10.1

**Plugins installed**: Cloud

**JVM version** (`java -version`): 15.0.1+9

**OS version** (`uname -a` if on a Unix-like system): Cloud

**Description of the problem including expected versus actual behavior**:

All generic threads are waiting with the following stack trace:

```
at java.util.concurrent.locks.LockSupport.park(Ljava/lang/Object;)V (LockSupport.java:211)                                                                                                             
at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(Ljava/util/concurrent/locks/AbstractQueuedSynchronizer$Node;IZZZJ)I (AbstractQueuedSynchronizer.java:714)                             
at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(I)V (AbstractQueuedSynchronizer.java:937)                                                                                             
at java.util.concurrent.locks.ReentrantLock$Sync.lock()V (ReentrantLock.java:153)                                                                                                                      
at java.util.concurrent.locks.ReentrantLock.lock()V (ReentrantLock.java:322)                                                                                                                           
at org.elasticsearch.action.bulk.BulkProcessor.internalAdd(Lorg/elasticsearch/action/DocWriteRequest;)V (BulkProcessor.java:379)                                                                       
at org.elasticsearch.action.bulk.BulkProcessor.add(Lorg/elasticsearch/action/DocWriteRequest;)Lorg/elasticsearch/action/bulk/BulkProcessor; (BulkProcessor.java:361)                                   
at org.elasticsearch.action.bulk.BulkProcessor.add(Lorg/elasticsearch/action/index/IndexRequest;)Lorg/elasticsearch/action/bulk/BulkProcessor; (BulkProcessor.java:347)                                
at org.elasticsearch.xpack.ilm.history.ILMHistoryStore.lambda$putAsync$0(Lorg/elasticsearch/action/index/IndexRequest;Lorg/elasticsearch/xpack/ilm/history/ILMHistoryItem;)V (ILMHistoryStore.java:150)
at org.elasticsearch.xpack.ilm.history.ILMHistoryStore$$Lambda$8091+0x0000000801eda370.run()V (Unknown Source)                                                                                         
[...]

PR #85490, opened 2022-03-30
Keep track of ingest load in DataStreams

### LINKED ISSUE #90102: Capture data streams write load

### LINKED ISSUE #90103: Store write load into a system data stream (including ILM policies for cleanup)

fcofdez:
This commit adds ingest load tracking for data streams.
In order to capture the ingest load a sampling process is
spawn in the background with a configurable frequency 
(`indices.write_load.collect.sampling_frequency` defaults to `1s`),
this takes a sample of the total time spent indexing 
(including parsing documents) and substracts that from the 
latest sample to get an approximate CPU usage. This process
keeps track of refreshes and merges too. Since storing all
the samples is too costly, the samples are recored in a 
histogram. Using a histogram allows us to know how the CPU
load is distributed. So far the load information is stored
in memory, but we're interested in inspecting historical ingest
load data to make better decisions around different cluster 
management operations, in order to do that, there's another
scheduled tasks with a configurable frequency 
(`indices.write_load.store.frequency` defaults to `1m`) that stores
part of the histogram for each active datastream write index
in the system datastream ``. This system datastream is attached
to an ILM policy that rollovers daily or when the shard reaches
`50GB` and keeps the data for 7 days.

`.indices-write-load` has the following mappings:

```
    "mappings": {
      "dynamic": "strict",
      "properties": {
        "@timestamp": {
          "type": "date",
          "format": "epoch_millis"
        },
        "data_stream": {
          "type": "keyword"
        },
        "index_name": {
          "type": "keyword"
        },
        "index_uuid": {
          "type": "keyword"
        },
        "shard": {
          "type": "integer"
        },
        "primary": {
          "type": "boolean"
        },
        "index_load_histogram": {
          "properties": {
            "50": {
              "type": "double"
            },
            "90": {
              "type": "double"
            },
            "95": {
              "type": "double"
            },
            "99": {
              "type": "double"
            },
            "max": {
              "type": "double"
            }
          }
        },
        "merge_load_histogram": {
          "properties": {
            "50": {
              "type": "double"
            },
            "90": {
              "type": "double"
            },
            "95": {
              "type": "double"
            },
            "99": {
              "type": "double"
            },
            "max": {
              "type": "double"
            }
          }
        },
        "refresh_load_histogram": {
          "properties": {
            "50": {
              "type": "double"
            },
            "90": {
              "type": "double"
            },
            "95": {
              "type": "double"
            },
            "99": {
              "type": "double"
            },
            "max": {
              "type": "double"
            }
          }
        }
      }
    }
```

There are a couple of things that need to be tackled:
- ~Move the logic to the x-pack plugin~
- Should we keep track of the node that hold the shard in the indexed document?
- Keep track of force-merges separately, this can be done in a follow up PR.

Closes https://github.com/elastic/elasticsearch/issues/90102
Closes https://github.com/elastic/elasticsearch/issues/90103


henningandersen:
Would you mind updating the description with a high-level overview of what we track, where we keep it etc?

fcofdez:
> Would you mind updating the description with a high-level overview of what we track, where we keep it etc?

Sure, once I fix the failing tests I'll provide a high-level overview 👍

server/src/main/java/org/elasticsearch/index/engine/ElasticsearchConcurrentMergeScheduler.java
```diff
@@ -117,7 +120,7 @@ protected void doMerge(MergeSource mergeSource, MergePolicy.OneMerge merge) thro
             beforeMerge(onGoingMerge);
             super.doMerge(mergeSource, merge);
         } finally {
-            long tookMS = TimeValue.nsecToMSec(System.nanoTime() - timeNS);
+            long tookMS = TimeValue.nsecToMSec(relativeTimeInNanosSupplier.getAsLong() - timeNS);
```
henningandersen, line 123: I think the relative time supplier we use is the cached one and thus this degrades the took time into 200ms granularity. I am not sure it is important but I think I would prefer to keep this using an outside supplied `System.nanoTIme` instead.

server/src/main/java/org/elasticsearch/index/engine/InternalEngine.java
```diff
@@ -1035,7 +1046,7 @@ public IndexResult index(Index index) throws IOException {
                     assert index.origin().isFromTranslog() || indexResult.getSeqNo() == SequenceNumbers.UNASSIGNED_SEQ_NO;
                     localCheckpointTracker.markSeqNoAsPersisted(indexResult.getSeqNo());
                 }
-                indexResult.setTook(System.nanoTime() - index.startTime());
+                indexResult.setTook(relativeTimeInNanosSupplier.getAsLong() - index.startTime());
```
henningandersen, line 1049: I think this also degrades into 200ms granularity? This could provoke slow logs if slow logging is setup to warn about indexing taking longer than 200ms.

server/src/main/java/org/elasticsearch/indices/IndicesWriteLoadStatsCollector.java
```diff
@@ -0,0 +1,242 @@
...
+    Iterable<IndexShard> getDataStreamsWriteIndicesShards() {
+        // Wait until the node has access to the cluster state
+        if (clusterStateReady.get() == false) {
+            return Collections.emptyList();
+        }
+
+        return () -> clusterService.state()
+            .metadata()
+            .getIndicesLookup()
+            .values()
```
henningandersen, line 80: This looks sort of expensive in a many-indices world? I wonder if iterating `IndicesService` instead is cheaper? Or perhaps maintaining a list of the relevant shards? We almost have that in `histograms` anyway.

server/src/main/java/org/elasticsearch/indices/IndicesWriteLoadStatsCollector.java
```diff
@@ -0,0 +1,242 @@
...
+            .filter(IndexAbstraction::isDataStreamRelated)
+            .map(dataStream -> indicesService.indexService(dataStream.getWriteIndex()))
+            .filter(Objects::nonNull)
+            .flatMap(f -> StreamSupport.stream(f.spliterator(), false))
+            .iterator();
+    }
+
+    private void cleanRolledOverIndices() {
+        final Metadata metadata = clusterService.state().metadata();
+        for (IndexAbstraction indexAbstraction : metadata.getIndicesLookup().values()) {
```
henningandersen, line 91: I wonder if we could do a `histograms.remove(predicate)` instead?

server/src/main/java/org/elasticsearch/indices/IndicesWriteLoadStatsCollector.java
```diff
@@ -0,0 +1,242 @@
...
+
+    @Nullable
+    String getParentDataStreamName(String indexName) {
+        IndexAbstraction indexAbstraction = clusterService.state().getMetadata().getIndicesLookup().get(indexName);
+        IndexAbstraction.DataStream parentDataStream = indexAbstraction.getParentDataStream();
+        return parentDataStream != null ? parentDataStream.getName() : null;
+    }
+
+    public List<ShardWriteLoadDistribution> getWriteLoadDistributionAndReset() {
+        cleanRolledOverIndices();
```
henningandersen, line 126: Should we prefer to collect the data from `histograms` first and then clean the rolled over indices? Just to get the last second of activity included.

server/src/main/java/org/elasticsearch/indices/IndicesWriteLoadStatsService.java
```diff
@@ -0,0 +1,201 @@
...
+            scheduledStore.cancel();
+        }
+    }
+
+    private void maybeScheduleSampling() {
+        if (enabled == false) {
+            return;
+        }
+
+        scheduledSampling = threadPool.schedule(this::collectWriteLoadSamples, samplingFrequency, ThreadPool.Names.GENERIC);
```
henningandersen, line 191: We can discuss this later too, no need to fix in this PR. But I wonder if we should just have a single special thread for this rather than using the generic pool?

server/src/main/java/org/elasticsearch/indices/IndicesWriteLoadStore.java
```diff
@@ -0,0 +1,190 @@
...
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
+    public static final Setting<Integer> MAX_RETRIES_SETTING = Setting.intSetting(
+        "indices.write_load.store.max_retries",
+        3,
+        1,
+        Setting.Property.NodeScope
+    );
+    public static final Setting<TimeValue> FLUSH_INTERVAL_SETTING = Setting.timeSetting(
```
henningandersen, line 61: I think I would prefer to set the bulk size in bytes rather than this time based condition.

server/src/main/java/org/elasticsearch/indices/IndicesWriteLoadStore.java
```diff
@@ -0,0 +1,190 @@
...
+
+            @Override
+            public void afterBulk(long executionId, BulkRequest request, Throwable failure) {
+                long items = request.numberOfActions();
+                logger.warn(new ParameterizedMessage("Failed to index {} items into indices write load index", items), failure);
+            }
+        }, threadPool, threadPool, () -> {})
+            .setBulkActions(-1)
+            .setFlushInterval(FLUSH_INTERVAL_SETTING.get(settings))
+            .setConcurrentRequests(1)
```
henningandersen, line 147: Perhaps this needs configuration?

henningandersen (commented):
Started going through this, here are my initial comments.

server/src/main/java/org/elasticsearch/indices/IndicesWriteLoadStore.java
```diff
@@ -0,0 +1,201 @@
...
+            @Override
+            public void afterBulk(long executionId, BulkRequest request, Throwable failure) {
+                long items = request.numberOfActions();
+                logger.warn(new ParameterizedMessage("Failed to index {} items into indices write load index", items), failure);
+            }
+        }, threadPool, threadPool, () -> {})
+            .setBulkActions(MAX_DOCUMENTS_PER_BULK_SETTING.get(settings))
+            .setBulkSize(MAX_BULK_SIZE_SETTING.get(settings))
+            .setConcurrentRequests(MAX_CONCURRENT_REQUESTS_SETTING.get(settings))
+            .setBackoffPolicy(BackoffPolicy.exponentialBackoff(TimeValue.timeValueMillis(1000), MAX_RETRIES_SETTING.get(settings)))
```
fcofdez, line 162: I was wondering if we should just drop the indexing ops when the cluster is overloaded, we should have enough information at that point to work with the past data more or less reliably. It would be nice to have some kind of QoS so operations like this one could have lower priority than operations issued by users.

server/src/main/java/org/elasticsearch/indices/IndicesWriteLoadStore.java
```diff
@@ -0,0 +1,201 @@
...
+    void putAsync(List<ShardWriteLoadDistribution> shardWriteLoadDistributions) {
+        if (enabled == false) {
+            return;
+        }
+
+        for (ShardWriteLoadDistribution shardWriteLoadDistribution : shardWriteLoadDistributions) {
+            try (XContentBuilder builder = XContentFactory.jsonBuilder()) {
+                shardWriteLoadDistribution.toXContent(builder, ToXContent.EMPTY_PARAMS);
+                final var request = new IndexRequest(INDICES_WRITE_LOAD_DATA_STREAM).source(builder).opType(DocWriteRequest.OpType.CREATE);
+                bulkProcessor.add(request);
```
dakrone, line 175: Will using the `BulkProcessor` here expose this at all to https://github.com/elastic/elasticsearch/issues/68468#issuecomment-772990141 ?

server/src/main/java/org/elasticsearch/indices/IndicesWriteLoadStatsCollector.java
```diff
@@ -0,0 +1,238 @@
...
+            .toList();
+
+        cleanRolledOverIndices();
+
+        return shardWriteLoadDistributions;
+    }
+
+    public List<ShardWriteLoadDistribution> getShardLoadDistributions() {
+        return histograms.values().stream().map(ShardWriteLoadHistogram::getWriteLoadDistribution).toList();
+    }
```
henningandersen, line 128: This method looks unused and once removed, we can make `getWriteLoadDistribution` private and then remove the synchronized inside it.

server/src/main/java/org/elasticsearch/indices/IndicesWriteLoadStatsCollector.java
```diff
@@ -0,0 +1,238 @@
...
+            );
+            long totalIndexingTime = Math.addExact(indexShard.getTotalIndexingTimeInNanos(), indexShard.getTotalDeleteTimeInNanos());
+            long totalMergeTimeInMillis = indexShard.getTotalMergeTimeInMillis();
+            long totalRefreshTimeInNanos = indexShard.getTotalRefreshTimeInNanos();
+            shardWriteLoadHistogram.recordSample(totalIndexingTime, totalMergeTimeInMillis, totalRefreshTimeInNanos);
+        }
+    }
+
+    List<IndexShard> getDataStreamsWriteIndicesShards() {
+        // Wait until the node has access to the cluster state
```
henningandersen, line 70: Why do we need this? I'd think we would simply get an empty list anyway from below?

server/src/main/java/org/elasticsearch/indices/IndicesWriteLoadStatsCollector.java
```diff
@@ -0,0 +1,238 @@
...
+
+        ShardWriteLoadDistribution getWriteLoadDistribution() {
+            synchronized (histogramsLock) {
+                return new ShardWriteLoadDistribution(
+                    System.currentTimeMillis(),
+                    parentDataStreamName,
+                    shardId,
+                    primary,
+                    indexingTimeHistogram.getValueAtPercentile(50),
+                    indexingTimeHistogram.getValueAtPercentile(90),
```
henningandersen, line 222: Can we also record 95th and 99th percentile from the beginning?

server/src/main/java/org/elasticsearch/indices/IndicesWriteLoadStore.java
```diff
@@ -0,0 +1,201 @@
...
+        ) {
+            ComposableIndexTemplate composableIndexTemplate = ComposableIndexTemplate.parse(parser);
+            return new SystemDataStreamDescriptor(
+                INDICES_WRITE_LOAD_DATA_STREAM,
+                "Stores indices write load distribution over time",
+                SystemDataStreamDescriptor.Type.INTERNAL,
+                composableIndexTemplate,
+                Map.of(),
+                Collections.emptyList(),
+                ExecutorNames.DEFAULT_SYSTEM_DATA_STREAM_THREAD_POOLS
```
henningandersen, line 101: I wonder if we should use the `DEFAULT_SYSTEM_INDEX_THREAD_POOLS` instead to reduce interference from user-invoked indexing?

server/src/main/java/org/elasticsearch/indices/cluster/IndicesClusterStateService.java
```diff
@@ -121,7 +122,8 @@ public IndicesClusterStateService(
         final SnapshotShardsService snapshotShardsService,
         final PrimaryReplicaSyncer primaryReplicaSyncer,
         final RetentionLeaseSyncer retentionLeaseSyncer,
-        final NodeClient client
+        final NodeClient client,
+        final IndicesWriteLoadStatsCollector indicesWriteLoadStatsCollector
```
henningandersen, line 126: I wonder if we should pass in the service instead here and let it delegate to the collector? Seems like we have the service orchestrating most things except this bit. That could also avoid registering the collector in Node.

henningandersen (commented):
Went through most of the production code, left comments.

fcofdez:
I moved all the relevant code to the new x-pack plugin. We might want to have `ShardWriteLoadHistogramSnapshot` in server as I think we would end up using that information there.

fcofdez:
I'm closing this as we won't follow this approach in the near term.

Issue #83423, opened 2022-02-02
Remove HLRC from ES codebase


dakrone:
### Description

With the launch of the [new Java client](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/8.0/introduction.html) the Elasticsearch High Level Rest Client (HLRC) is deprecated, no longer shipped as an artifact, and can be removed from the ES codebase. This meta issue is to track the the removal of the HLRC for 8.x.

Steps (subject to change):
- [x] Remove all HLRC tests #83425
- [x] Remove all usages of HLRC sub-clients by tests, replacing them with the low-level `RestClient`. These classes can be found inside of [`RestHighLevelClient.java`](https://github.com/elastic/elasticsearch/blob/master/client/rest-high-level/src/main/java/org/elasticsearch/client/RestHighLevelClient.java#L278)
  - [x] Remove usages of `IndicesClient` (`indices()`) (@martijnvg, #85492)
  - [x] Remove usages of `ClusterClient` (`cluster()`) (@gmarouli, https://github.com/elastic/elasticsearch/pull/83593) 
  - [x] Remove usages of `IngestClient` (`ingest()`) (@martijnvg, #85483)
  - [x] Remove usages of `SnapshotClient` (`snapshot()`) (@dakrone, https://github.com/elastic/elasticsearch/pull/85845)
  - [x] Remove usages of `TasksClient` (`tasks()`) (@danhermann, https://github.com/elastic/elasticsearch/pull/83649)
  - [x] Remove usages of `XPackClient` (`xpack()`) (@gmarouli, https://github.com/elastic/elasticsearch/pull/83593)
  - [x] Remove usages of `GraphClient` (`graph()`) (@danhermann, https://github.com/elastic/elasticsearch/pull/83471)
  - [x] Remove usages of `IndexLifecycleClient` (`indexLifecycle()`) (@dakrone, #83476)
  - [x] Remove usages of `SearchableSnapshotsClient` (`searchableSnapshots()`) (@dakrone, #85472)
  - [x] Remove usages of `FeaturesClient` (`features()`) (@danhermann, https://github.com/elastic/elasticsearch/pull/83651)
  - [x] Remove usages of `MigrationClient` (`migration()`) (@joegallo, https://github.com/elastic/elasticsearch/pull/83609)
  - [x] Remove usages of `MachineLearningClient` (`machineLearning()`) (#83885, #83912)
  - [x] Remove usages of `SecurityClient` (`security()`)
[...]

PR #85492, opened 2022-03-30
Remove HLRC IndicesClient and related classes.

martijnvg:
Relates to #83423


test/framework/src/main/java/org/elasticsearch/test/rest/ESRestTestCase.java
```diff
@@ -1141,6 +1149,15 @@ protected void refreshAllIndices() throws IOException {
         client().performRequest(refreshRequest);
     }
 
+    protected static RefreshResponse refresh(String index) throws IOException {
```
martijnvg, line 1152: Reusing server module's CreateIndexResponse, AcknowledgedResponse for RefreshResponse tests. 
It isn't part of the hlrc code base, but does provide us in tests the capabilities to easily parse the responses for the related apis.

test/framework/src/main/java/org/elasticsearch/test/rest/ESRestTestCase.java
```diff
@@ -2015,4 +2052,17 @@ protected FieldCapabilitiesResponse fieldCaps(
...
+        return XContentHelper.createParser(XContentParserConfiguration.EMPTY, toBytes(response), XContentType.JSON);
+    }
+
+    protected static BytesReference toBytes(Response response) throws IOException {
+        return new BytesArray(EntityUtils.toByteArray(response.getEntity()));
+    }
+
+    protected static Map<?, ?> toMap(Response response) throws IOException {
+        return XContentHelper.convertToMap(XContentType.JSON.xContent(), response.getEntity().getContent(), false);
+    }
```
dakrone, line 2066: Is it necessary to have the `Map<?, ?>` response for this? We already have the `responseAsMap` method in `ESRestTestCase` that returns a `Map<String, Object>`, do we need both?

dakrone (approved):
LGTM, I left one comment about duplicate helper methods

PR #85495, opened 2022-03-30
replaces BooleanSupplier with a Predicate

idegtiarenko:
This change replaces BooleanSupplier with a Predicate. This should simplify the code a little and also avoid capturing the variable from the lambda context.


DaveCTurner (approved):
LGTM

(removed the version label since this won't land in 8.2)

PR #85497, opened 2022-03-30
[TEST] Add readiness Docker tests

grcevski:
Add new readiness service test inside the DockerTests.


rest-api-spec/build.gradle
```diff
@@ -81,6 +81,8 @@ tasks.named("yamlRestTestV7CompatTransform").configure { task ->
   task.skipTest("search.aggregation/20_terms/string profiler via global ordinals native implementation", "The profiler results aren't backwards compatible.")
   task.skipTest("search.aggregation/20_terms/string profiler via map", "The profiler results aren't backwards compatible.")
   task.skipTest("search.aggregation/20_terms/numeric profiler", "The profiler results aren't backwards compatible.")
+  task.skipTest("search.aggregation/360_date_histogram/Multi-value date histogram", "The aggregation results aren't backwards compatible.")
```
rjernst, line 84: This should be in a separate PR, if we are going to mute it (though the message doesn't seem write, I don't see why these aggs would not be backward compatible, that seems like a bug).

rjernst (approved):
LGTM

rest-api-spec/build.gradle
rjernst: I muted this https://github.com/elastic/elasticsearch/commit/0ecaceb013ad3632a6e3f450b03c49e0b3f5bfeb

grcevski: Thanks.

grcevski:

PR #85500, opened 2022-03-30
Fix wildcard highlighting on match_only_text

### LINKED ISSUE #85493: Highlighting in Kibana is not working for match_only_text
### Elasticsearch Version

8.1.1.

### Installed Plugins

non

### Java Version

_bundled_

### OS Version

ESS

### Problem Description

When searching for any term in Kibana the highlighting works on any field type except `match_only_text` which has become the default field type in ~7.14 for all of the beats and Elastic Agent integrations.
E.g. here it does not work:
image
E.g. on 7.17 using text and keyword fields:
image

Discussed this with @romseygeek already. seems that the highlighter is not capeable of dealing with the specialized match_only_text field type.

### Steps to Reproduce

Index something using a `match_only_text` field.
Run a search on it. It should be highlighted.

### Logs (if relevant)

_No response_

ywelsch:
Fixes a bug where `match_only_text` fields were ignored during highlighting when a field name with wildcard was specified.

Closes #85493


javanna (approved):
LGTM thanks @ywelsch

ywelsch:
@elasticmachine update branch

romseygeek:
Should we apply this to `constant_keyword` as well?

ywelsch:
> Should we apply this to constant_keyword as well?

I will leave that to a follow-up (possibly by someone else). I think addressing match_only_text is the biggest issue, and I want to merge this PR before going on PTO.

dikshachauhan-qasource:
Hi @ywelsch 

We have attempted to validate the highlight working on discover tab with field values using wild card and observed that "data_steam.dataset" field is not responding as expected.

**Screenshots:**
**Working fine for host related fields:**
image

image

**Not working fine for agent ephemeral id and data_steam.dataset under both indexes Logs and metrics:**
image

image

Can you please confirm if we are on right path and isn't missing anything.

Thanks
QAS

ywelsch:
`data_stream.dataset ` is a `constant_keyword` field, not a `match_only_text` field (what this PR fixes). It's a related bug tracked here: https://github.com/elastic/elasticsearch/issues/85596

PR #85501, opened 2022-03-30
Better error message for run-as denials 

### LINKED ISSUE #72904: Improve the error message for run-as failure
Elasticsearch supports one user (`User_A`) impersonate another user (`User_B`) with the run-as feature. If `User_A` does _not_ have the privilege to run-as `User_B` when issuing a request with run-as, ES correctly rejects the request. But error message is along the line of:
```
action [...] is unauthorized for user [User_A] run as [User_B]
```
This is not accurate enough and can be misleading at times because when the authorization fails, we don't really know whether the action is unauthorized for either `User_A` or `User_B`. What we do know is that `User_A` cannot run-as `User_B`.  Hence the error message should be improved to be more precise about the actual failure.

n1v0lg:
Users can impersonate i.e. run an action on behalf of other users via
the run-as feature, provided they have the right privileges. For user A
running as user B, a run-as can be denied either because A does not
have permission to run as B or because B does not have permission to
perform the given action. For the former (missing run as permission) we
don't clearly identify that that was the cause of an auth failure. This
PR refactors how we construct denial messages, and provides a more
concrete error message when a run as is denied due to missing run as
permissions.

Closes #72904


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationDenialInfo.java
```diff
@@ -0,0 +1,157 @@
...
+                apiKeyId,
+                roles,
+                context,
+                grantingClusterPrivileges,
+                grantingIndexPrivileges
+            );
+        }
+
+        public Builder withRunAsUserPrincipal() {
+            User user = authentication.getUser();
```
n1v0lg, line 111: The bodies of these methods are copy&paste from what used to be in `AuthorizationService`

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java
```diff
@@ -862,66 +860,46 @@ private ElasticsearchSecurityException denialException(
...
+        if (isAnonymousEnabled && anonymousUser.equals(authentication.getUser().authenticatedUser())) {
+            if (anonymousAuthzExceptionEnabled == false) {
+                return authcFailureHandler.authenticationRequired(action, threadContext);
+            }
+        }
+
+        final AuthorizationDenialInfo.Builder denialInfoBuilder = AuthorizationDenialInfo.builder(authentication, action)
+            .withApiKeyId()
+            .withRunAsUserPrincipal()
+            .withContext("because user unauthorized to run as target user");
```
n1v0lg, line 874: Will add a better message

ywangd (commented):
The refactoring makes sense to me in general. 

Since this is a draft, I didn't spend much time into reading the details. For now just one comment on the new Builder class: The `withXxx` methods do not seem to be very useful because they cannot be freely turned on and off, i.e. there are inherent logic on how they should be used together (or not together). Having them as separate methods gives the impression that they are free to pick and choose, but they are not really and this may cause future usage errors.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java
```diff
@@ -853,75 +840,76 @@ private void putTransientIfNonExisting(String key, Object value) {
         }
     }
 
-    private ElasticsearchSecurityException denialException(
+    private ElasticsearchSecurityException runAsDenied(Authentication authentication, String action) {
```
n1v0lg, line 843: In principle, we could further refactor by pulling all the code below into a `ElasticsearchSecurityExceptionFactory` class that would look s.th. like this:

```java
class ElasticsearchSecurityExceptionFactory {
  private final boolean isAnonymousEnabled;
  private final boolean anonymousAuthzExceptionEnabled;
  private final AnonymousUser anonymousUser;
  private final AuthenticationFailureHandler authcFailureHandler;
  
  ElasticsearchSecurityException runAsDenied(Authentication authentication, String action) {
        ....
  }
  
  ElasticsearchSecurityException bulkItemActionDenied(
        String action,
        TransportRequest request,
        String resolvedIndex
    ) 

    ...
}
``` 

`denialException` and `AuthorizationDenialMessages` could be entirely encapsulated by that class.

However, that's a bigger undertaking - not sure if it's worth it.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java
```diff
@@ -853,75 +840,76 @@ private void putTransientIfNonExisting(String key, Object value) {
...
     ) {
-        return denialException(authentication, action, request, null, cause);
+        return actionDenied(
+            authentication,
+            action,
+            request,
+            IndexAuthorizationResult.getFailureDescription(List.of(resolvedIndex), restrictedIndices),
+            null
+        );
     }
```
ywangd, line 860: This method does not look necessary to me. It's a very thin wrapper for `actionDeined` and the only thing it really does is to calcualte the `context` String. But to do that it has to introduce a `String resovledIndex` argument which is an anomaly in the series of `xxxDenied` methods. I'd rather stick to the existing five arguments (which is already plenty) and directly invoke `actionDenied` at the caller and remove this indrection.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java
```diff
@@ -326,7 +325,7 @@ private void checkOperatorPrivileges(Authentication authentication, String actio
             threadContext
         );
         if (operatorException != null) {
-            throw denialException(authentication, action, originalRequest, "because it requires operator privileges", operatorException);
+            throw actionRequiresOperatorPrivileges(authentication, action, originalRequest, operatorException);
```
ywangd, line 328: Similarly, I don't think this extra `actionRequiresOperatorPrivileges` is necessary and would prefer to drop it for direct invocation of `actionDenied`.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java
```diff
@@ -853,75 +840,76 @@ private void putTransientIfNonExisting(String key, Object value) {
...
-            if (privileges != null && privileges.size() > 0) {
-                message = message
-                    + ", this action is granted by the index privileges ["
-                    + collectionToCommaDelimitedString(privileges)
-                    + "]";
-            }
+    private ElasticsearchSecurityException denialException(
+        Authentication authentication,
+        String action,
+        Supplier<String> messageSupplier,
```
ywangd, line 902: Do we really need `Supplier` here vs a simple `String`?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java
ywangd: We can stop at extracting out the string composing part in this PR. I am OK with it.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/AuthorizationServiceTests.java
```diff
@@ -922,26 +923,26 @@ public void testServiceAccountDenial() {
...
             verify(auditTrail).runAsDenied(eq(requestId), eq(authentication), eq(action), eq(request), authzInfoRoles(role.names()));
+            assertThat(securityException, throwableWithMessage(containsString("run as [" + finalUser.principal() + "]")));
         } else {
             if (canRunAs) {
                 verify(auditTrail).runAsGranted(eq(requestId), eq(authentication), eq(action), eq(request), authzInfoRoles(role.names()));
+                assertThat(securityException, throwableWithMessage(containsString("run as [" + finalUser.principal() + "] with roles [")));
+                assertThat(
+                    securityException,
+                    throwableWithMessage(containsString("this action is granted by the index privileges [read,all]"))
+                );
```
ywangd, line 938: I believe this assertion should be moved outside of this `if (canRunAs)` block and at the same level of the following `verify(...)` call.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/AuthorizationServiceTests.java
```diff
@@ -922,26 +923,26 @@ public void testServiceAccountDenial() {
...
-            )
+            throwableWithMessage(containsString("[" + action + "] is unauthorized for service account [" + serviceUser.principal() + "]"))
         );
-        if (isRunAs) {
-            assertThat(securityException, throwableWithMessage(containsString("run as [" + finalUser.principal() + "] with roles [")));
-        }
-        assertThat(securityException, throwableWithMessage(containsString("this action is granted by the index privileges [read,all]")));
         if (isRunAs && false == canRunAs) {
             verify(auditTrail).runAsDenied(eq(requestId), eq(authentication), eq(action), eq(request), authzInfoRoles(role.names()));
+            assertThat(securityException, throwableWithMessage(containsString("run as [" + finalUser.principal() + "]")));
```
ywangd, line 930:
```suggestion
            assertThat(securityException, throwableWithMessage(containsString("is unauthorized to run as [" + finalUser.principal() + "]")));
```
Otherwise, the erorr message for unauthorized action (as opposed to denied-run-as), i.e. "run as [...] with roles [" can also match the assertion.

ywangd (approved):
LGTM

I have a few comments and would appreciate if you could address them. But nothing critical to worth another round.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/AuthorizationServiceTests.java
n1v0lg: Currently, the run as denied message does not include context on index privileges since it's somewhat irrelevant context and distracts from the underlying issue (that the run as is not allowed, which is not related to index privs). Would prefer to keep it that way - wdyt?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java
n1v0lg: Makes sense! I was thinking about moving closer to hiding the details of how messages are created but the implementation falls flat of that so I'm for keeping it simple and less cluttered.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/AuthorizationServiceTests.java
ywangd: If I read the conditions correctly, outside this if block, the error message should not be about run-as, but just for an unauthorised action.

docs/changelog/85501.yaml
```diff
@@ -0,0 +1,6 @@
+pr: 85501
+summary: Authorization denial message specifies when run as is unauthorized
```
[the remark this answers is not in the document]
n1v0lg: 👍 going with "Better error message for run-as denials"

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/AuthorizationServiceTests.java
n1v0lg: Ahh you're totally right! Thanks

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java
n1v0lg: I went with a supplier because `denialException` doesn't always use the passed in message. When we're in anonymous mode and authz exceptions are disabled, we return an authc exception instead. The supplier signals that a little better than passing in a string. I'm considering a rename to `authzDenialMessageSupplier` to make it more explicit. It also preserves the flow we had before this PR: we return an authc exception in the anonymous case, and construct an authz exception otherwise. As such prefer the supplier, despite the added indirection. 


wouldn't get constructed at all if we're in anonymous mode and authz exceptions are disabled. In terms of control flow this makes sense: if we need an authentication exception, we construct that, otherwise we construct an authz exception. 

Though it adds indirection in comparison to passing in a String, I'd prefer to preserve the flow and use a supplier. Wdyt?

ywangd: No problem. I have no strong opinion.

n1v0lg:
@elasticmachine update branch

Issue #58820, opened 2020-07-01
SQL: ORDER BY a derived numerical field doesn't sort NULLs


bpintea:
Given the index: `SELECT long FROM idx`:
```
     long      
---------------
-10            
null           
-1             
2    
```
, ordering by an operation on field `long`, for instance `SELECT long FROM idx ORDER BY long+0` (or `SELECT long+0 AS x FROM idx ORDER BY x`) returns:
```
     long      
---------------
-10            
-1             
null           
2   
```
not sorting the `null`s correctly. And similarly for other operations (including applying functions) and numerical types.
Consequently the `NULLS FIRST`/`LAST` don't apply either.

bpintea:
When ordering by a function, the sorting is accomplished by a Painless script (vs. being pushed directly to ES) using `InternalQlScriptUtils` functions returning a value. Since this value can't be a `null` - as Painless would err on it -, the functions (`nullSafe*()`) will return a default value instead (like `false` for Boolean, `0` for `Number`s, "" for Strings). Which interferes with documents actually having the respective field values, these being sorted similarly to docs not having any value for respective field.
Inline Painless constructs of the type `if (doc['field'].size() != 0) doc['field'].value` would have the `nulls` grouped out at the bottom of the returned hits, but this is not controllable; i.e. can't `ORDER by some_func(field...) NULLS FIRST`.

So potential solutions might be to do the sorting locally, only support `NULLS LAST` for ordering by functions or add a "missing" parameter to filtering by script?

costin:
See https://github.com/elastic/elasticsearch/issues/34221 and in particular the discussion around https://github.com/elastic/elasticsearch/issues/34597
It might be worth revisiting this to see whether something has changed but the way around this is to do a sorting by a script with a composite key - a bool/int that tells whether something is null or not, the second the actual value.
This allows the null values to be sorted first or last followed/proceeded by the actual values.

Luegg:
With runtime fields, there is a new canonical way to support nulls in sort keys and use the `missing: _first/_last` option at the same time. A query like `SELECT languages - 2 AS a FROM test_emp ORDER BY a NULLS LAST` can be translated to something using:

```
{
  "runtime_mappings": {
    "a": {
       "type": "long",
       "script": """
           long res = InternalSqlScriptUtils.sub(InternalQlScriptUtils.docValue(doc,\"languages\"),2);
           if(res!=null) emit(res);
       """
    }
  },
  "sort": [
[...]

PR #85502, opened 2022-03-30
Add support for VERSION field type in SQL and EQL

### LINKED ISSUE #83375: EQL support for fields of type version
EQL doesn't support `version` type fields e.g. `sequence by agent.version` where `agent.version` is of type `version` fails with 

```
verification_exception: Found 1 problem\nline 2:38: Cannot use field [agent.version] with unsupported type [version]
```

The `version` field type was introduced in 7.10 https://github.com/elastic/elasticsearch/pull/59773

luigidellaquila:
Fixes  #83375


x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/type/DataTypeConverter.java
```diff
@@ -537,6 +548,10 @@ public static Object convert(Object value, DataType dataType) {
                 throw new QlIllegalArgumentException("[" + o + "] is not a valid IPv4 or IPv6 address");
             }
             return o;
+        }),
+        STRING_TO_VERSION(o -> {
+            // TODO validate...?
```
luigidellaquila, line 553: Not sure this is needed. The underlying implementation seems to do no validation on VERSION fields

luigidellaquila:
@elasticmachine update branch

luigidellaquila:
@elasticmachine update branch

luigidellaquila:
@elasticmachine update branch

luigidellaquila:
@elasticmachine update branch

x-pack/plugin/sql/qa/server/src/main/resources/version.csv-spec
```diff
@@ -0,0 +1,69 @@
+// To mute tests follow example in file: example.csv-spec
+
+//
+// Tests for VERSION fields
+//
```
Luegg, line 5: a couple more things that I would cover:

* group by version
* constant expression returning a version (cast)
* select a string field casted to a version (script returning a version value). like `select concat(str, '.0')::version from "versions"`
* order by script returning a version
* group by script returning a version
* cast a version to a string

Luegg (commented):
looks mostly good to me but I left some suggestions for more testing

luigidellaquila:

x-pack/plugin/sql/qa/server/src/main/resources/version.csv-spec
luigidellaquila: Thank you @Luegg, it revealed a problem:

```
select '1.2.3'::version < '1.11.0'::version
```

the comparison takes into consideration SemVer order, we will have to reimplement it in *QL (now it's just comparing strings, that is wrong)

luigidellaquila: Just for the record, the same happens with `ip` field type

Luegg: oh, I didn't see your response. I also created an issue for the IP case https://github.com/elastic/elasticsearch/issues/85563

luigidellaquila:
@elasticmachine update branch

luigidellaquila:
@elasticmachine update branch

x-pack/plugin/sql/qa/server/src/main/resources/version.csv-spec
```diff
@@ -0,0 +1,206 @@
+// To mute tests follow example in file: example.csv-spec
```
astefan, line 1: I suggest adding more tests with queries with:
- `HAVING`
- `GROUP BY` and `COUNT`
- `NULLS FIRST`
- other functions using `version`, including in `ORDER BY`. For example `SELECT CASE WHEN LEFT(version, 1) = '1' THEN 1 WHEN LEFT(version, 1) = '2' THEN 2 ELSE 0 AS v, COUNT(*) GROUP BY 1 ORDER BY 1 DESC`.
- `ORDER BY` with `DESC`
- `SELECT version, version as v, version as v2 GROUP BY 1 ORDER BY 2 LIMIT 3`

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/querydsl/agg/GroupByKey.java
```diff
@@ -62,6 +62,7 @@ public final CompositeValuesSourceBuilder<?> asValueSource() {
             } else if (script.outputType() == IP) {
                 builder.userValuetypeHint(ValueType.IP);
             }
+            // TODO hint for VERSION
```
astefan, line 65: ?

x-pack/plugin/sql/src/test/java/org/elasticsearch/xpack/sql/analysis/analyzer/FieldAttributeTests.java
```diff
@@ -182,7 +182,7 @@ public void testDottedFieldPathTypo() {
     public void testStarExpansionExcludesObjectAndUnsupportedTypes() {
```
astefan, line 182: I think you need to look at `testUnsignedLongVersionCompatibility` as a test example for `version`. What do you think @bpintea?

x-pack/plugin/sql/src/test/java/org/elasticsearch/xpack/sql/plan/logical/command/ShowColumnsTests.java
```diff
@@ -94,4 +97,16 @@ public void testUnsignedLongFiltering() {
...
+
+    public void testVersionFieldFiltering() {
+        List<?> rowSupported = List.of("version", "VARCHAR", "version");
+        List<?> rowUnsupported = List.of("version", "OTHER", "unsupported");
+        for (SqlVersion version : VERSION_FIELD_TEST_VERSIONS) {
+            List<List<?>> rows = new ArrayList<>();
+            // mapping's mutated by IndexCompatibility.compatible, needs to stay in the loop
+            Map<String, EsField> mapping = loadMapping("mapping-multi-field-variation.json", true);
+            ShowColumns.fillInRows(IndexCompatibility.compatible(mapping, Version.fromId(version.id)), null, rows);
+            assertTrue((supportsUnsignedLong(Version.fromId(version.id)) && rows.contains(rowSupported)) || rows.contains(rowUnsupported));
```
astefan, line 109: `supportsUnsignedLong`?

astefan (commented):
Great progress, but I think more tests are needed. Added some suggestions, feel free to check other already existent tests to think about other ideas, as well.

x-pack/plugin/sql/qa/jdbc/src/main/java/org/elasticsearch/xpack/sql/qa/jdbc/ResultSetTestCase.java
```diff
@@ -2470,6 +2497,48 @@ private static void indexTestFieldsDoc(String docId, Object... values) throws IO
...
+            builder.array("version", "1.3.0");
+            builder.array("name", "version 1.3.0");
+        });
+        index("test", "" + (docId++), builder -> {
+            builder.array("version", "1.11.0");
+            builder.array("name", "version 1.11.0");
+        });
+
+        // some higher versions
+        for (int i = 0; i < randomInt(10); i++) {
```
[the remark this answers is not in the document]
luigidellaquila: It was an attempt to randomize the test dataset as much as possible, adding valid and invalid version values outside of the range of the test queries so that the tests are still predictable. Eg. test that random values that do not match the test criteria are not returned by mistake.

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/querydsl/agg/GroupByKey.java
luigidellaquila: I see `ValueType` is  deprecated, so I was unsure if adding further content to that class would be acceptable and if there could be an alternative. Any suggestions?

luigidellaquila: After off-line discussions and investigations, it seems it's not needed.

x-pack/plugin/sql/qa/server/src/main/resources/version.csv-spec
luigidellaquila: HAVING unfortunately is supported only for a limited set of data types, and Version is not one of them.

astefan (approved):
LGTM. Thank you.

Luegg (commented):
Some additional comments also partially related to our offline discussion about extracting versions from search results as instances of `Version`.

x-pack/plugin/sql/qa/server/src/main/resources/version.csv-spec
```diff
@@ -0,0 +1,367 @@
...
+11    |kkkkk      |1.
+12    |aaaaa      |1.1.2.3.4
+13    |lllll      |1.
+14    |mmmmm      |1.5.2.9
+
+;
+
+
+castConstantToVersion
+SELECT '1.2.3'::version as v;
```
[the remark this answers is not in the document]
luigidellaquila: I tried it, but it seems to work fine. Are you testing it on the same dataset or on something different?
Anyway, I'm adding a specific test case to the csv specs.

x-pack/plugin/sql/qa/server/src/main/resources/version.csv-spec
```diff
@@ -0,0 +1,367 @@
...
+SELECT * FROM apps WHERE version::string RLIKE '2.*.4';
+
+  id:i  |  name:s     |  version:s
+--------+-------------+---------------
+3       |ccccc        | 2.3.4
+
+;
+
+orderByVersion
+SELECT * FROM apps ORDER BY version, id;
```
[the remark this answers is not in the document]
luigidellaquila: At first I thought it could depend on valid vs. invalid version numbers, but I tried it on different datasets and it seems to be completely independent from that.
I investigated it a bit more and it seems it depends on the same problem reported in #58820

luigidellaquila:
> There seems to be a problem with queries like select `first(version::string::version) from apps` though.

The reason for this error is that StreamOutput has no writer defined for Version type. The blocker for fixing it is that Version class is in x-pack, so I cannot reference it from inside Server module. Unfortunately I don't see a simple solution for this, apart from moving Version implementation to Server module, with all the implications on licensing. 
In my opinion, a fix with this impact (changes to multiple modules, licensing implications) is not justified by the uncommon use case, unless there is some request from the user base. My suggestion is to leave it as it is for now and track it with a new issue.

luigidellaquila:
@elasticmachine update branch

x-pack/plugin/sql/src/test/java/org/elasticsearch/xpack/sql/analysis/analyzer/FieldAttributeTests.java
```diff
@@ -349,6 +351,45 @@ public void testUnsignedLongVersionCompatibility() {
         }
     }
 
+    public void testVersionFieldVersionCompatibility() {
```
bpintea, line 354: `testVersionTypeVersionCompatibilty()`? `SELECT '1.2.3'::version'...` makes use of no field to test.

x-pack/plugin/sql/src/test/java/org/elasticsearch/xpack/sql/plan/logical/command/sys/SysColumnsTests.java
```diff
@@ -170,6 +179,30 @@ public void testUnsignedLongFiltering() {
         }
     }
 
+    public void testVersionFieldFiltering() {
```
bpintea, line 182: `testVersionTypeFiltering()`?

x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/type/DataTypes.java
```diff
@@ -44,6 +44,8 @@ public final class DataTypes {
     public static final DataType DATETIME         = new DataType("DATETIME", "date",        Long.BYTES,  false, false, true);
     // ip
     public static final DataType IP               = new DataType("ip",                45,                false, false, true);
+    // version
+    public static final DataType VERSION          = new DataType("version",           Integer.MAX_VALUE, false, false, true);
```
bpintea, line 48: Giving VERSION a size of Integer.MAX_VALUE seems wrong, though I guess that's the theoretical max. Not sure if ES type enforces any limit? It seems there's only a recommendation, but no actual limit: https://semver.org/#does-semver-have-a-size-limit-on-the-version-string

bpintea (approved):
Nice work, LGTM!

x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/type/DataTypes.java
luigidellaquila: Technically, any string can be indexed as a Version, so the limit is the same as a KEYWORD.

x-pack/plugin/sql/qa/jdbc/src/main/java/org/elasticsearch/xpack/sql/qa/jdbc/ResultSetTestCase.java
```diff
@@ -2244,6 +2245,30 @@ public void testResultSetConsumed() throws IOException {
...
+
+        // bad version value
+        query = "SELECT name, version from test where version = 'foo'";
+        doWithQuery(query, results -> {
+            results.next();
+            assertEquals("version foo", results.getString("name"));
+            assertEquals("foo", results.getString("version"));
+            assertFalse(results.next());
+        });
+    }
```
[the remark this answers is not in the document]
luigidellaquila: Since there is this distinction between valid and invalid versions, I thought it would be good to have both in the result set test. Anyway, I agree that knowing the implementation details this test is probably not crucial. I'll refactor it a bit to make it shorter.

Luegg (approved):
👍

x-pack/plugin/sql/jdbc/src/main/java/org/elasticsearch/xpack/sql/jdbc/EsType.java
```diff
@@ -32,6 +32,7 @@ public enum EsType implements SQLType {
     TIME(Types.TIME),
     DATETIME(Types.TIMESTAMP),
     IP(Types.VARCHAR),
+    VERSION(Types.VARCHAR),
```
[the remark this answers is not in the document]
luigidellaquila: Good point, fixed

x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/expression/gen/script/Scripts.java
```diff
@@ -80,7 +80,12 @@ public static ScriptTemplate nullSafeFilter(ScriptTemplate script) {
     }
 
     public static ScriptTemplate nullSafeSort(ScriptTemplate script) {
-        String methodName = script.outputType().isNumeric() ? "nullSafeSortNumeric" : "nullSafeSortString";
+        String methodName;
```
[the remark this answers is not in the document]
luigidellaquila: Just a leftover, fixed

luigidellaquila:
@elasticmachine update branch

costin (changes requested):

costin (approved):

costin (changes requested):
Requesting changes regarding the impact on mapper-version project.

x-pack/plugin/mapper-version/src/main/java/org/elasticsearch/xpack/versionfield/VersionEncoder.java
```diff
@@ -35,7 +35,7 @@
  *  lexically in ASCII sort order. Numeric identifiers always have lower precedence than non-numeric identifiers.
  * </ul>
  */
-class VersionEncoder {
+public class VersionEncoder {
```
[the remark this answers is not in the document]
luigidellaquila: Good point, actually these changes are a leftover from the previous attempts to support Version without impacting on the Search/Painless implementation, but since we went with a more complete solution (see https://github.com/elastic/elasticsearch/pull/85990), we don't need them anymore.

luigidellaquila:
@elasticmachine update branch

costin (approved):
Thanks for reviewing the feedback. LGTM!

luigidellaquila:
Had to revert due to the following:

```
va.lang.IllegalArgumentException: can not write type [class org.elasticsearch.xpack.versionfield.Version]
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.StreamOutput.writeGenericValue(StreamOutput.java:828)
        at org.elasticsearch.xpack.ql.expression.gen.processor.ConstantProcessor.writeTo(ConstantProcessor.java:68)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.StreamOutput.writeNamedWriteable(StreamOutput.java:1097)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.StreamOutput.writeNamedWriteableList(StreamOutput.java:1203)
        at org.elasticsearch.xpack.sql.expression.predicate.conditional.ConditionalProcessor.writeTo(ConditionalProcessor.java:73)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.StreamOutput.writeNamedWriteable(StreamOutput.java:1097)
        at org.elasticsearch.xpack.ql.execution.search.extractor.ComputingExtractor.writeTo(ComputingExtractor.java:56)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.StreamOutput.writeNamedWriteable(StreamOutput.java:1097)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.StreamOutput.writeNamedWriteableList(StreamOutput.java:1203)
        at org.elasticsearch.xpack.sql.execution.search.SearchHitCursor.writeTo(SearchHitCursor.java:80)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.StreamOutput.writeNamedWriteable(StreamOutput.java:1097)
        at org.elasticsearch.xpack.sql.session.Cursors.encodeToString(Cursors.java:82)
        at org.elasticsearch.xpack.sql.session.Cursors.encodeToString(Cursors.java:73)
        at org.elasticsearch.xpack.sql.plugin.TransportSqlQueryAction.createResponse(TransportSqlQueryAction.java:225)
```

The problem seems to be related to how Version objects are translated to strings to be serialized, so it should be deterministic, but strangely it's not (before merging the build was green and I didn't notice this error before during the tests)

Issue #85333, opened 2022-03-24
Pushing back on index stats requests can cause ILM rollover-ready checks to pile up


original-brownbear:
#83832 introduced a rate limit of 100 per coordinating node for certain stats requests. This causes issues once ILM starts executing more than 100 of them at a time which it can easily do during a rollover-ready check.
For one, if more than 100 indices need to be checked for rollover readiness during one periodic execution of ILM those in excess of the 100 will be marked as in ILM error.
More importantly though, this effectively limits us to checking only 100 indices per ILM trigger period (10 minutes by default) which will cause issues in larger deployments.

I think the easiest solution here might be to disable this new functionality by default to not introduce new functionality that might cause trouble elsewhere into 8.2 given how close the 8.2 FF is.

dakrone:
Okay, a few (non-exhaustive!) ideas for fixing this for us to discuss:

1. Disable the limiting for this API entirely.

This is the fast-and-easy fix, though we remove the pushback and we'll have to figure out if/how to re-add it in the future. This might be something we do for 8.2, but I don't think we should consider it a permanent fix.

2. Add a special flag that makes requests not subject to the 100-max limit.

This would allow the ILM-made requests to pass this flag through and not run into the limit. The downside for this is that if we still increment the number of concurrent requests, we could end up causing *other* stats requests to fail. We don't necessarily have to increment the request counter though, so it could still be an option.

3. Handle rejections in the ILM action itself.

We could handle these in the `onFailure` listener for the rollover check, and then do a retry (maybe with a timeout) inside the `WaitForRolloverReadyStep` class itself if it got the "too many concurrent stats requests" error. It won't really help too much though, because the first 100 will get in, and then all the others will be sitting there doing retries and we'll still end up not being able to get the request through.

4. Add the ability to "wait-for-an-empty-slot" for the stats request.

This would let us use the timeout to say, wait 30 seconds on the semaphore for a permit to be available. It essentially limits us to 100 concurrently, but they will all queue up behind one another, and we could end up in the state where the last rollover check waits longer than the timeout and we end up in an error anyway.

5. Implement check-rollover-ready within ILM differently.

Since we're triggering most of these from a cluster state changed event, we *could* grab the stats once at the beginning of the new state, and pass it in to all of these events (so that they use the same stats). This would entail probably not calling the rollover stuff itself, but factoring that logic out so that ILM could use it directly (passing in the stats info as necessary).

[...]

PR #83832, opened 2022-02-11
Push back excessive requests for stats

gmarouli:
### The issue we are addressing
We are trying to avoid overwhelming the coordinating node of a stats or a recovery request in the case where other nodes are having troubles. See #51992 :
 
> if a single node becomes unresponsive for a few minutes then we could see a couple hundred requests build up and in a decent sized cluster each could consume many MBs of heap on the coordinating node.

### Approach
We are introducing a limit on how many stats and recovery requests a node can coordinate concurrently. When we receive one too many of this requests the node rejects with a `409`. We are not concerned with further handling of this error because we assume that these requests would have timed out anyway. The limit is configurable as a cluster setting. 

The choice to have one limit that would apply to both requests was made for the sake of simplicity. This is seen as a protective mechanism that is triggered only to defend the node if the cluster is having troubles and not to rate limit requests when things are going well. We could alternatively create separate limits for each requests we want to bound, if there are objections to this approach. 

### Implementation details
We introduce `StatsRequestLimiter` which uses `AdjustableSemaphore` to bound the requests. Via the method `tryToExecute(Task task, Request request, ActionListener<Response> listener, TriConsumer<Task, Request, ActionListener<Response>> execute)`, this class orchestrates all aspects of pushing back, it checks the semaphore, it tracks metrics and handles exceptions.

### Affected actions:
- `IndicesStatsActions`
- `RecoveryAction`
- `IndicesSegmentsAction` 
-  `NodesStatsAction`
- `ClusterStatsAction`
- `NodesInfoAction`
- `NodesUsageAction`

Resolves #51992

Issue #51992, opened 2020-02-06
Push back on excessive requests for stats


DaveCTurner:
Today stats requests are processed by the `management` threadpool, which is also used for important internal management tasks such as syncing global checkpoints and retention leases, ILM and SLM. The `management` threadpool has an unbounded queue. Some stats take a nontrivial amount of effort to compute and it is certainly possible to request stats more frequently than the cluster can respond. We cannot control the behaviour of clients requesting stats, and I've seen more than a few situations where an errant monitoring system harms the cluster with its request rate (see links in #51915). Since we doggedly enqueue every request it can take a very long time to recover from this situation, and while working through the queue the well-behaved internal management tasks do not run in a timely fashion. The quickest recovery path may be to restart any affected nodes.

I think we should be pushing back against this kind of behaviour to protect the cluster from abusive monitoring clients. We could, for instance, use different threadpools for the internal (and well-behaved) actions from external (and possibly-abusive) ones, and use a bounded queue for the threadpool handling the external actions. Some users of the management threadpool are not clearly one or the other and we'll need to use some judgement to decide whether we need to protect them from abuse - e.g. license management, security cache management. I've yet to see such actions involved in struggling clusters, however, so perhaps either way would be ok.

Relates #51915

DaveCTurner:
We discussed this today and agreed that it's a valid concern; we could perhaps move the "well-behaved" actions into the `GENERIC` threadpool and bound the queue on the `MANAGEMENT` threadpool to push back on anything that's left.

However we don't see too much urgency on this front, and there are more general questions about how we assign work to the various threadpools that subsume this one, so we decided to proceed on the local improvement to stats performance in #51991 and close this issue to indicate that we won't be working on it in the near future.

DaveCTurner:
I'm reopening this as we've seen cases where stats requests can become overwhelming even after the fixes mentioned above. I think we should consider again some ideas for bounding the resources used by stats requests, for instance limiting the number of in-flight stats requests being coordinated by each node. If a single node becomes unresponsive for a few minutes then we could see a couple hundred requests build up and in a decent sized cluster each could consume many MBs of heap on the coordinating node.

Relates https://github.com/elastic/elasticsearch/issues/82337 which would fix one particular way to get into this mess.

Relates https://github.com/elastic/elasticsearch/issues/77466 since this particularly affects high-shard-count clusters.

burqen:
[...]

PR #85504, opened 2022-03-30
Revert "Push back excessive requests for stats" (#83832) from 8.2

joegallo:
Due to #85333, we want to have a little longer to figure out how to address #51992 -- for the moment, that looks like reverting #83832 from 8.2 so that we have another release cycle to figure out what changes (if any) to make to #83832 in order to address the issues raised on #85333.

Once this is merged, I'll update the version tag on #83832 to 8.3.0 rather than 8.2.0


original-brownbear (approved):
Assuming this is a straight revert without other changes (which it looks like it is) LGTM :)

joegallo:
Yeah, there was a trivial conflict that I resolved but besides that it's a straight revert.

dakrone (approved):
Assuming this is a straight revert like Armin said, this LGTM; but can you please change the title to have some keywords for what this is actually reverting (not just an issue number) as like 90% of my memory is based on title-matching things in my browser history? :D

PR #85510, opened 2022-03-30
Fix skip caching factor with indices.queries.cache.all_segments

dnhatn:
Spin off #85423

The `indices.queries.cache.all_segments` setting should not be used in production environments. And we should also use the default `skipCachingFactor` (i.e, 10) for LRUQueryCache when that setting is enabled.


server/src/test/java/org/elasticsearch/indices/IndicesQueryCacheTests.java
```diff
@@ -345,6 +353,79 @@ public void testStatsOnEviction() throws IOException {
         cache.close(); // this triggers some assertions
     }
 
+    public void testSkipCachingFactorAtLeast10() throws Exception {
```
jtibshirani, line 356: Small comment -- to me this test doesn't add much except guard against someone accidentally changing the skip factor. I think it'd accomplish the same to add a code comment like "the skipFactor must be 10 here to match the default in Lucene's LRUQueryCache".

jtibshirani (approved):
This looks good to me. I left a small comment for you to consider, but feel free to merge!

jpountz (approved):
Agreed with Julie's feedback, otherwise LGTM.

dnhatn:
@jtibshirani @jpountz Thanks for review.

PR #78825, opened 2021-10-07
Add a specialized PainlessExecuteIndexAction

jimczi:
This change refactors the painless execute action into two actions:
  * PainlessExecuteAction
  * PainlessExecuteIndexAction

Script context that require an index to run should use the newly added PainlessExecuteIndexAction.
This action checks if the user has the read permission on the index to run but unlike the existing
PainlessExecuteAction, no cluster privilege is required.
For scripts that run outside of an index context, the existing PainlessExecuteAction can still be used.
Note that passing the index in the body continues to work for backward compatibility purpose but in such
case it's the old action that requires the cluster privilege **and** the read privilege that will run.

Relates #48856

Issue #48856, opened 2019-11-05
Add privilege for script actions


spinscale:
**Describe the feature**: The `manage_watcher` privilege right now only contains the `"cluster:admin/xpack/watcher/*", "cluster:monitor/xpack/watcher/*"` actions.

I do think it makes sense to also add `cluster:admin/script/*` (get/put/delete scripts) in there as well, as a watcher admin may need to refer to external scripts.

If you think that makes sense, I'm happy to open a PR for this.

albertzaharovits:
Rather than adding new permitted actions to the `manage_watcher` privilege, I think a new privilege such as `manage_scripts` is more consistent with the rest of our privileges namespace.
In this case the role for the watcher user should have both `manage_watcher` and `manage_scripts` privilege.

P1llus:
@albertzaharovits 

Do you mean that manage_scripts should be a global permission, and not restricted to watcher?

Would you then also want to support more granular permission, so scripts can only be executed in single or multiple places (like watches)?

albertzaharovits:
@P1llus I originally meant to create a new [cluster privilege](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-privileges.html#privileges-list-cluster) to be able to use the GET/PUT/DELETE script APIs.

What you are suggesting, being able to confine the usage of scripts to certain contexts, is a different thing. But I think this is an interesting proposal. I envisage being able to define a list of context names when PUTing the script. I wonder what folks from @elastic/es-core-features think about this?

P1llus:
@albertzaharovits 

The reason the idea came up, is just like @spinscale mentioned, when the case is external scripts, just to minimize the possibility of some sort of vulnerability.
Now I have not fully set myself into the capabilities of the external scripts engine, what painless allows and not etc, but If there is any possibilities that the script can be any sort of malicious, being able to have more granular control would be beneficial.

While context level would be great, if that is a really big task, then we should at least be able to give access based on script type maybe?

albertzaharovits:
We discussed this in our weekly team meeting.

We have decided on adding a `manage_scripts` cluster privilege that would permit usage of the GET/PUT/DELETE script APIs. This means the `manage_watcher` privilege is not changed. It also means APIs can use all the scripts internally, without requiring any extra privilege (nothing changes in this regard).

[...]

PR #85512, opened 2022-03-30
Authorize painless execute as index action when an index is specified

### LINKED ISSUE #86428: Resolve permissions issue with `cluster:admin/scripts/painless/execute`
2nd attempt at https://github.com/elastic/elasticsearch/issues/84591

references https://github.com/elastic/elasticsearch/issues/48856#issuecomment-831144741

tldr; the permission structure for the endpoint isn't useful for most users. If users can create runtime fields they should have access to this api, they shouldn't need a specific permission set.

javanna:
Painless execute allows users to validate their scripts. Some of the supported script contexts
support providing a sample document as well as an index to pull the mappings from.

The painless execute API requires cluster admin privileges today and while that's ok for the contexts that
don't support providing an index, it is not ideal when an index is provided. In fact users can run scripts
as part of the search API, which requires only the indices/read privilege on the indices that the users
is reading from.

This commit maps the painless execute action to an indices/read action when an index is specified, so that in
that case the same privileges as a search action will be requested to run painless execute.

Note: a different approach that does not require handling the painless execute case in the security code has been tried before and required creating a separate index level action with the index in the request path instead of in the request body (see #78825). We'd like to keep things as they are though in the painless execute request, with the index in the definition of the context setup as part of the request body.

Relates to #48856
Closes #86428


javanna:
I opened this as a draft to get some early feedback. If we agree that this is the way forward, it will need more tests, at least an integration test to verify that it all works as expected end-to-end.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/SecurityActionMapper.java
```diff
@@ -41,6 +44,13 @@ public static String action(String action, TransportRequest request) {
                     return CLUSTER_PERMISSION_ANALYZE;
                 }
             }
+            case CLUSTER_PERMISSION_PAINLESS_EXECUTE -> {
+                assert request instanceof SingleShardRequest;
+                String index = ((SingleShardRequest<?>) request).index();
+                if (index != null) {
+                    return INDICES_PERMISSION_PAINLESS_EXECUTE;
```
javanna, line 51: while it is pretty clear that we want to make this case an indices read action, and the request is already an indices request (it will not contain wildcards hence there is no need for it to implement Replaceable), I wonder if we are still ok with the action without the index being a cluster admin level action.

tvernum: I think we're good like this.
If we made a no-index action an index privilege, then it would have to fail because we won't know how to authorize it (which index would we check?)
Leaving it as a cluster privilege will at least give it a chance to succeed.

tvernum (approved):
LGTM

mattkime:
addresses https://github.com/elastic/elasticsearch/issues/86428 - I wanted to make sure there was an associated issue.

javanna:
@tvernum thanks for the review. I added an integration test, which made me realize I also needed to handle the shard level action. Fixed that, this should be ready now.

PR #78050, opened 2021-09-21
[DOCS] Add composite runtime fields

lockewritesdocs:
Adds documentation for composite runtime fields that were added in #76287.

PR #85513, opened 2022-03-30
[DOCS] add composite field context to the painless execute docs

javanna:
The painless execute API supports the composite_field context since the composite field was added.
This commit adds docs for it where missing.

Relates to #78050


romseygeek (approved):
LGTM!

PR #85522, opened 2022-03-31
DOC CCR Disaster recovery does not handle Security configuration

albertzaharovits:
We do not support and don't plan to support disaster recovery arrangements
where Security configuration is replicated between the production and the
disaster recovery cluster because the cluster-local Security APIs assume
exclusive write on the .security system index.


lockewritesdocs:
@elasticmachine run @elasticsearch-ci/docs

lockewritesdocs:

docs/reference/ccr/index.asciidoc
```diff
@@ -54,6 +54,12 @@ https://www.elastic.co/webinars/replicate-elasticsearch-data-with-cross-cluster-
 Then, <<ccr-getting-started-tutorial,set up {ccr}>> on your local machine and work
 through the demo from the webinar.
 
+IMPORTANT: In all of these cases <<manually-configure-security, Security is to be configured>> independently for each cluster.
+In particular, no disaster recovery arrangement described bellow handles the replication of Security configuration,
+such as <<native-realm, native users>>, roles, or <<token-authentication-services, token credentials>>.
+Security configuration should be backed up and restored using the <<back-up-specific-feature-state, regular snapshot and restore procedure>>
+for the "security" feature.
+
```
lockewritesdocs, line 62:
```suggestion
IMPORTANT: In all of these use cases, you must
<<manually-configure-security,configure security>> independently on every
cluster. The security configuration is not replicated when configuring {ccr} for 
disaster recovery. To ensure that the {es} `security` feature state is backed up,
<<back-up-specific-feature-state,take snapshots>> regularly. You can then restore
the native users, roles, and tokens from your security configuration.
```

lockewritesdocs (commented):
Left a suggested rewording. I also removed some of the links (such as for native users and tokens) which I don't think are needed here.

albertzaharovits:
@lockewritesdocs Can you please drop an eye on this again?

lockewritesdocs (approved):
LGTM 🦖

PR #85525, opened 2022-03-31
Capture deprecation warnings in batched master tasks

### LINKED ISSUE #85506: Batched cluster state executors can leak deprecation info across response threads
Discovered this issue in some of our tests where indexing a new document that creates a new index results in a deprecation log meant for a completely unrelated index. This seems to be a wider problem with `TaskBatcher` and its management of thread contexts when calling executors.

#### AutoCreateIndex example:
When indexing a document to an index that does not exist, the `TransportBulkAction` forwards requests to create new indices to `AutoCreateAction.TransportAction`. This action maintains a cluster state executor instance that applies multiple create index requests in one operation. The create request is submitted to the master service where it is batched together with other index auto creation tasks using the executor instance as a batching key. When the state is meant to be applied, the executor is called from the cluster state update thread and iterates over the indices, creating each one. After each index is created, the executor calls the listener that is waiting on that index creation and moves on to the next create call in the batch.

The problem arises if one of the indices in a batch generates a deprecation log of some sort. An example would be using a dot-prefixed index name on a non-hidden/non-system index. This deprecation log is stored in the ThreadContext of the current thread so that it may be surfaced to remote clients through an HTTP header. After the index is created, the next index to be auto-created in the batch is processed but the ThreadContext is not cleared out from the previous task. When this unrelated index is created, the listener picks up the same deprecation log from the previous index in the batch, thus leaking the deprecation notice into a response that may be unrelated to the original deprecation message.

This problem is seen in a number of test failures where a document is being written but returns an unrelated deprecation log in the response, which fails the test as it was not expecting a deprecation warning. Examples include #83366, #84993, and #84885. I was able to reproduce the error with a client program that submits a number of index requests for indices that do not exist, half of which generate a deprecation log. Deprecation logs for unrelated indices were present on responses for valid index names:

<pre><code>WARNING: request [POST http://localhost:9200/<b>index-298753</b>/_doc] returned 2 warnings: [299 Elasticsearch-8.2.0-SNAPSHOT-d2d87578150e215d08a9a8ae896168f07b822895 "index name [<b>.index-384428</b>] starts with a dot '.', in the next major version, index names starting with a dot are reserved for hidden indices and system indices"],[299 Elasticsearch-8.2.0-SNAPSHOT-d2d87578150e215d08a9a8ae896168f07b822895 "index name [<b>.index-81812</b>] starts with a dot '.', in the next major version, index names starting with a dot are reserved for hidden indices and system indices"]</code></pre>

The request references `index-298753` but receives warnings for indices `.index-384428` and `.index-81812` which are not present in that request, but were submitted around the same time. This happens reliably even over a relatively small number (40) of asynchronously submitted requests (in my test, one thread using the new async java rest client).

Notably, the CreateIndexAction [does not submit cluster state updates in a batched manner][normal]. In my recreation attempts I was not able to reproduce the problem when creating the indices directly. The problem only arises when indexing a document that would create a new index.

[normal]: https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/cluster/metadata/MetadataCreateIndexService.java#L312-L315

DaveCTurner:
It's possible for a cluster state update task to emit deprecation
warnings, but if the task is executed in a batch then these warnings
will be exposed to the listener for every item in the batch. With this
commit we introduce a mechanism for tasks to capture just the warnings
relevant to them, along with assertions that warnings are not
inadvertently leaked back to the master service.

Closes #85506


DaveCTurner:
Not ready to merge, this needs testing in `MasterServiceTests` too, but before I start on that I want to see whether CI finds any other cases that need fixing and/or some more fundamental problems.

DaveCTurner:
The Windows packaging test failure is definitely unrelated and I'm pretty sure the DiscoveryDisruptionIT is too - opened #86974 to track it.

server/src/main/java/org/elasticsearch/cluster/action/shard/ShardStateAction.java
```diff
@@ -318,7 +320,8 @@ public ShardFailedClusterStateTaskExecutor(AllocationService allocationService,
         @Override
         public ClusterState execute(
             ClusterState currentState,
-            List<ClusterStateTaskExecutor.TaskContext<FailedShardUpdateTask>> taskContexts
+            List<TaskContext<FailedShardUpdateTask>> taskContexts,
+            Supplier<Releasable> dropHeadersContextSupplier
```
idegtiarenko, line 324: It looks like this is not required in every implementation. It would also be great to minimize the amount of parameters.
I wonder if there is a nice way to convert it to a method call within the interface?

DaveCTurner: Not really. You can do this:

```java
    default ClusterState execute(ClusterState currentState, List<TaskContext<T>> taskContexts, Supplier<Releasable> dropHeadersContextSupplier) throws Exception {
        return execute(currentState, taskContexts);
    }

    ClusterState execute(ClusterState currentState, List<TaskContext<T>> taskContexts) throws Exception;
```

I don't like this pattern because the implementations that _do_ need the supplier must also supply a dummy implementation of the 2-arg method which is never called. I'd rather just have the extra arg throughout.

idegtiarenko (approved):

server/src/main/java/org/elasticsearch/cluster/action/shard/ShardStateAction.java
henningandersen: I think it would be nice to introduce an object for the parameters, like `BatchContext`. That would have `state` and `taskContexts` methods as well as a `dropHeaders` method. This would remove the high visibility of this extra arg as well as make the signature nicer (no need for `Supplier`). It would allow future similar extensions with less effect across code base. And avoid the multi-line breaking of the formatter (minor point).

henningandersen (approved):
LGTM.

Left one optional comment.

server/src/main/java/org/elasticsearch/cluster/action/shard/ShardStateAction.java
DaveCTurner: :+1: makes sense. I'll not do that here but I'll open a follow-up.

DaveCTurner: Actually I'll do that first, that'll reduce the noise from this change.

PR #85526, opened 2022-03-31
TSDB: fix the time_series in order collect priority

weizijun:
Time series index is sort by timestamp desc. So the collect order will be order by timestamp desc. 
When _tsid are the same, the bigger timestamp segment will be pop.


romseygeek:
Good spot.  I think we're still experimenting with the order in which things are sorted, so maybe the best thing to do here is to read the index sort from the index reader when we build the searcher and determine whether we're sorting by asc or desc timestamp then?

weizijun:
> Good spot. I think we're still experimenting with the order in which things are sorted, so maybe the best thing to do here is to read the index sort from the index reader when we build the searcher and determine whether we're sorting by asc or desc timestamp then?

Yeah, @imotov can look at the demand? The index sort config of time_series index is set internally, is it need to read the index sort config for `TimeSeriesIndexSearcher`?

romseygeek:
> is it need to read the index sort config for TimeSeriesIndexSearcher?

Yes, you can read it from the IndexReader's metadata.

weizijun:
> Yes, you can read it from the IndexReader's metadata.

I can't find sort metadata from IndexReader, but I find the index sort in the SearchContext, which can be passed to the TimeSeriesIndexSearcher constructor.

weizijun:
> Good spot. I think we're still experimenting with the order in which things are sorted, so maybe the best thing to do here is to read the index sort from the index reader when we build the searcher and determine whether we're sorting by asc or desc timestamp then?

@romseygeek  I set the time_series sort to a static value, and `TimeSeriesIndexSearcher` read the value to decide the right order.

weizijun:
@elasticmachine update branch

csoulios:

csoulios (commented):
@weizijun can you please add the changelog file?

weizijun:
@elasticmachine update branch

csoulios (approved):
LGTM. Thank you!

weizijun:
> LGTM. Thank you!

Thanks @csoulios !

PR #85533, opened 2022-03-31
Apply precommit checks for non productive projects

breskeby:
Ensure projects with only yaml, java or cluster tests also apply precommit checks.
We only apply testingconventions now for projects with existing src test folder
as the TestingConventionsTask is incompatible with projects with no test sourceSet.
This is a prerequisite to port more projects away from using StandaloneRestTestPlugin
and RestTestPlugin in favor of yaml, java or cluster tests with dedicated sourceSets.

Also we fix deprecation warnings for forbiddenPattern and filePermissions tasks about
implicit non declared dependencies on resources tasks

### DIFF
--- a/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/InternalPluginBuildPlugin.java
+++ b/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/InternalPluginBuildPlugin.java
@@ -8,6 +8,7 @@
 
 package org.elasticsearch.gradle.internal;
 
+import org.elasticsearch.gradle.internal.precommit.TestingConventionsPrecommitPlugin;
 import org.elasticsearch.gradle.internal.precommit.TestingConventionsTasks;
 import org.gradle.api.Project;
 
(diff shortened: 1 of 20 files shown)


mark-vieira (commented):
> Ensure projects with only yaml, java or cluster tests also apply precommit checks.

Since we disable testing conventions when a test souceset is missing doesn't this effectively do nothing? For those QA projects that only have rest tests we still won't be running test conventions on them, right?

breskeby:

breskeby:
@mark-vieira yes we would not apply test conventions on them (test conventions is incompatible with non 'test' sourceSet tests, but we would apply any other precommit check (including checkstyle and friends)

breskeby:
> @mark-vieira yes we would not apply test conventions on them (test conventions is incompatible with non 'test' sourceSet tests, but we would apply any other precommit check (including checkstyle and friends). Once we do not have any test tasks sharing the test sourceSet anymore we can basically remove testingConventions or at least a huge part of it

@mark-vieira can I have your tick off here or should I understand your 👍 as "approved"? :)

mark-vieira (approved):
Yep, LGTM.

PR #85545, opened 2022-03-31
various updates related to minimum Java version 17

DJRickyB:
Since 8.0.0, we only support Java 17 and up. This change cleans up unsupported Java versions from jvm.options and updates documentation to reference supported Java versions, in case it is confusing to users.

### DIFF
--- a/distribution/src/config/jvm.options
+++ b/distribution/src/config/jvm.options
@@ -49,17 +49,7 @@
 ##
 ################################################################
 
-## GC configuration
-8-13:-XX:+UseConcMarkSweepGC
-8-13:-XX:CMSInitiatingOccupancyFraction=75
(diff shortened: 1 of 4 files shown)


distribution/src/config/jvm.options
```diff
@@ -71,9 +61,8 @@
 # specified
 -XX:+HeapDumpOnOutOfMemoryError
 
-# exit right after heap dump on out of memory error. Recommended to also use
-# on java 8 for supported versions (8u92+).
-9-:-XX:+ExitOnOutOfMemoryError
+# exit right after heap dump on out of memory error
```
DJRickyB, line 64: removed the full stop for symmetry with other comments ¯\_(ツ)_/¯

ChrisHegarty (approved):
LGTM.

Issue #72715, opened 2021-05-04
[Transform] Re-visit priviledge checks


hendrikmuhs:
put and update run privilege checks to warn the user early in case of lacking permissions.

Revisit:
`view_index_metadata` - this is today only checked if transform creates the dest index, however `view_index_metadata` is also used for checkpoints

`read_cross_cluster` - this is additionally required for CCS

hendrikmuhs:
problems found:

 - `view_index_metadata` should always be checked for the source index
 - CCS checks are done against the local cluster this has some problems:
   - no checks on the remote
   - the check on the local cluster can fail although the check is wrong
   - to workaround the issue its necessary to create privileges on the local cluster that mimic the remote privileges or use `?defer_validation=true`

Apart from that `read_cross_cluster` is not effective, this is tracked in #58476.

Given these new findings I think this is a bug.

valeriy42:
#142403 has partially fixed the symptoms of this issue.

## Before vs. After PR #142403

| API | Before (broken) | After (fixed) |
|-----|-----------------|---------------|
| `_preview` | Fails, but with misleading message: *"Source indices have been deleted or closed."* | Fails with actionable message: *"User lacks the required permissions to read source indices on remote cluster [X]."* |
| `PUT _transform` | **Silently succeeds** -- unauthorized transform is created | **Rejects** with a validation error naming the permission problem |
| `_start` | **Silently succeeds** -- transform starts, processes zero documents, user must debug why | **Rejects** with a validation error naming the permission problem |
| `PUT` with `defer_validation: true` then `_start` | PUT succeeds (expected), _start **silently succeeds**, produces nothing | PUT succeeds (expected), _start now **rejects** with clear error |

**Key nuance** (lines 146-158 of `AbstractCompositeAggFunction.java`): `validateQuery` only rejects when a security failure is **positively identified** (e.g., `ElasticsearchSecurityException` in cluster/shard failures). When null aggregations are returned but no security signal is found, it still passes -- this is intentional because a local wildcard pattern resolving to zero indices is a legitimate use case (integrations that start transforms before source data exists, per #95562).

The actual fix of the problems described in this issue 
- Up-front ` _has_privileges`  for remote indices
- `view_index_metadata` for remote source
is blocked by #67798 (open, High prio, Team:Security)

[...]

PR #85547, opened 2022-03-31
[Transform] Include missing privileges in the error message

przemekwitek:
This PR enriches privilege message by including the exact privileges that were missing.

Relates https://github.com/elastic/elasticsearch/issues/72715


hendrikmuhs (approved):
LGTM

PR #82457, opened 2022-01-12
Distinguish "missing repository" from "missing repository plugin"

idegtiarenko:
Currently it is possible to uninstall repository plugin before
deleting repository that uses it. In such case the repository is still
listed in the API however every interraction with it produces the
"no such repository" exception. This is very confusing for the users.
This change updates the error message to correctly point the root of the
 issue.

Closes: #81758

PR #83706, opened 2022-02-09
Remove irrelevant nodeHotThreads comment above nodeUsage method in ClusterAdminClient

mushao999:
This PR remove the irrelevant `nodeHotThreads` comment above `nodeUsage` mehotd which was introduce by #24169.

PR #85551, opened 2022-03-31
Distinguish missing and invalid repositories

### LINKED ISSUE #85550: NPE in RepositoriesService
### Elasticsearch Version

7.16.2

### Installed Plugins

no

### Java Version

_bundled_

### OS Version

Linux iZbp1dn0ktfed9mjoyuiajZ 5.10.84-10.2.al8.x86_64 #1 SMP Thu Feb 10 15:03:09 CST 2022 x86_64 x86_64 x86_64 GNU/Linux

### Problem Description

When put repository usint PutRespository API for an existing repository, NullPointerException was thrownd, short stack shows as below
```
1> java.lang.Exception: java.lang.NullPointerException: Cannot invoke "org.elasticsearch.repositories.Repository.getMetadata()" because "repository" is null
  1>   at org.elasticsearch.cluster.service.MasterService.innerExecuteTasks(MasterService.java:916) ~[main/:?]
  1>   at org.elasticsearch.cluster.service.MasterService.executeTasks(MasterService.java:871) ~[main/:?]
  1>   at org.elasticsearch.cluster.service.MasterService.runTasks(MasterService.java:253) [main/:?]
  1>   at org.elasticsearch.cluster.service.MasterService$Batcher.run(MasterService.java:153) [main/:?]
  1>   at org.elasticsearch.cluster.service.TaskBatcher.runIfNotProcessed(TaskBatcher.java:114) [main/:?]
  1>   at org.elasticsearch.cluster.service.TaskBatcher$BatchedTask.run(TaskBatcher.java:170) [main/:?]
  1>   at org.elasticsearch.common.util.concurrent.ThreadContext$ContextPreservingRunnable.run(ThreadContext.java:714) [main/:?]
  1>   at org.elasticsearch.common.util.concurrent.PrioritizedEsThreadPoolExecutor$TieBreakingPrioritizedRunnable.runAndClean(PrioritizedEsThreadPoolExecutor.java:260) [main/:?]
  1>   at org.elasticsearch.common.util.concurrent.PrioritizedEsThreadPoolExecutor$TieBreakingPrioritizedRunnable.run(PrioritizedEsThreadPoolExecutor.java:223) [main/:?]
  1>   at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) [?:?]
  1>   at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) [?:?]
  1>   at java.lang.Thread.run(Thread.java:833) [?:?]
```

### Steps to Reproduce

I will create an IT to reproduce this issue. The reproduce step is as following:
- put repository using `RestPutRepository` API and make sure non-master node failed to create repository.
- restart master let one of the previous non-master node become master
- put the same repository using the `RestPutRepository` API.

### Logs (if relevant)

_No response_

mushao999:
The `RepositoryService` on each node maintains a repository instance for
every repository defined in the cluster state. The master validates each
repository definition before inserting it into the cluster state, but in
some cases this validation is incomplete. For instance, there may be
node-local configuration of which the master is unaware which prevents
the repository from being instantiated on some other node in the
cluster.

Today if a repository cannot be instantiated then the node will log a
warning and continue as if the repository doesn't exist. This results in
a confusing `RepositoryMissingException` when trying to use the
repository, and various other surprises (e.g. #85550). With this commit
we create a placeholder `InvalidRepository` which reports a more
accurate exception when it is used.

Relates #82457 which did the same sort of thing for unknown plugins.
Closes #85550 since the repository in question is no longer `null`.


server/src/main/java/org/elasticsearch/cluster/service/MasterService.java
```diff
@@ -899,7 +899,7 @@ private static ClusterState innerExecuteTasks(
         final var taskContexts = castTaskContexts(executionResults);
         try {
             return executor.execute(previousClusterState, taskContexts);
-        } catch (Exception e) {
+        } catch (Throwable e) {
```
DaveCTurner, line 902: Catching `Throwable` is not acceptable, sorry, you'll need to revert this.

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
```diff
@@ -0,0 +1,121 @@
...
+            Environment environment,
+            NamedXContentRegistry namedXContentRegistry,
+            ClusterService clusterService,
+            BigArrays bigArrays,
+            RecoverySettings recoverySettings
+        ) {
+            super(metadata, environment, namedXContentRegistry, clusterService, bigArrays, recoverySettings);
+            List<String> unstableNodes = UNSTABLE_NODES.get(metadata.settings());
+            if (unstableNodes.contains(clusterService.getNodeName())) {
+                throw new RepositoryException(metadata.name(), "Failed to create repository: current node is not stable");
```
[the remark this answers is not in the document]
mushao999: I've created the `RepositoriesDuplicatedCreationIT` to reproduce this bug.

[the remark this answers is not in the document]
mushao999: And more general, repository creation logic is open to be implement in repository plugin which is uncontrolled, so if the implement is not good enough creation failed can happen.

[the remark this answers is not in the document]
DaveCTurner: This sounds like a bug in your plugin, it should not require a healthy network to create the repository. Compare for instance `S3Repository` which merely validates some settings in its constructor.

server/src/main/java/org/elasticsearch/cluster/service/MasterService.java
DaveCTurner: It's ok for a test to just hang if something throws an `Error`. Certainly not ideal, but definitely better than catching `Throwable`. There are timeouts which would terminate the test run eventually.

DaveCTurner (changes requested):
I would rather fix this using an approach closer to the one in #82457 which avoids the `null` by creating a dummy repository that returns a more specific exception message than the `RepositoryMissingException` we get today. The repository isn't really missing, it's invalid on some nodes, and we should distinguish these cases.

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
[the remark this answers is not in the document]
mushao999: > This sounds like a bug in your plugin, it should not require a healthy network to create the repository. Compare for instance `S3Repository` which merely validates some settings in its constructor.

I‘ve discuss with the plugin owner. It's true there are some network action during repository creation which can cause creation failure. 
However, I still think elasticsearch core should not lay it's robustness on plugin implement.

mushao999:
Thanks, I will update in that way.

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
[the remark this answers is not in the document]
DaveCTurner: > I still think elasticsearch core should not lay it's robustness on plugin implement.

Yes I agree, I just want to be clear that protecting against this NPE will not really address the problem in the plugin. If transient network issues can prevent the plugin from even being created then you will need to fix this by retrying the put-repository request once the transient issues are resolved. I don't see a great way to detect that a retry is needed here.

mushao999:
@DaveCTurner  hi, I've update this PR by adding a new DummyRepository, please help to reivew it.

mushao999:
@DaveCTurner  By the way , there is a small PR of mine which has been there for a long time , could you please help to review it as well: https://github.com/elastic/elasticsearch/pull/83706

server/src/main/java/org/elasticsearch/repositories/DummyRepository.java
```diff
@@ -0,0 +1,187 @@
...
+import java.io.IOException;
+import java.util.Collection;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * This class represents a repository that failed during creation.
+ * This could happen when repository plugin is not stable enough to guarantee repository creation success all the time
+ */
+public class DummyRepository extends AbstractLifecycleComponent implements Repository {
```
DaveCTurner, line 35: Naming suggestion:

```suggestion
public class InvalidRepository extends AbstractLifecycleComponent implements Repository {
```

server/src/main/java/org/elasticsearch/repositories/DummyRepository.java
```diff
@@ -0,0 +1,187 @@
...
+    private final RepositoryMetadata repositoryMetadata;
+
+    public DummyRepository(RepositoryMetadata repositoryMetadata) {
+        this.repositoryMetadata = repositoryMetadata;
+    }
+
+    private RepositoryException createCreationException() {
+        return new RepositoryException(
+            repositoryMetadata.name(),
+            "repository type [" + repositoryMetadata.type() + "] failed to create on current node"
```
DaveCTurner, line 46: I think we should keep hold of the creation exception and include it as the cause of this exception.

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
```diff
@@ -524,10 +524,11 @@ public void applyClusterState(ClusterChangedEvent event) {
                         logger.warn(() -> new ParameterizedMessage("failed to create repository [{}]", repositoryMetadata.name()), ex);
                     }
                 }
-                if (repository != null) {
-                    logger.debug("registering repository [{}]", repositoryMetadata.name());
-                    builder.put(repositoryMetadata.name(), repository);
+                if (repository == null) {
+                    repository = new DummyRepository(repositoryMetadata);
```
DaveCTurner, line 528: I would rather do this within the `catch` blocks above, using the exception that is thrown by the repository creation attempt. At the moment these blocks catch `RepositoryException` but we should instead catch `Exception` and then `assert ex instance RepositoryException` to fail tests on other kinds of exception.

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
```diff
@@ -0,0 +1,121 @@
...
+        );
+
+        // put repository for the first time: only let master node create repository successfully
+        createRepository(
+            repositoryName,
+            UnstableRepository.TYPE,
+            Settings.builder().put("location", randomRepoPath()).put(UnstableRepository.UNSTABLE_NODES.getKey(), unstableNode)
+        );
+
+        // restart master
```
DaveCTurner, line 103: Could we do something before restarting the master here to assert that attempts to use the repository will fail with the expected exception. For instance, the verify repository API should report the right kind of failure on the unstable node.

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
```diff
@@ -0,0 +1,121 @@
...
+        internalCluster().restartNode(internalCluster().getMasterName());
+        ensureGreen();
+
+        // put repository again: let all node can create repository successfully
+        createRepository(repositoryName, UnstableRepository.TYPE, Settings.builder().put("location", randomRepoPath()));
```
DaveCTurner, line 108: Similarly could we run the verify repository API here and check that it now works?

DaveCTurner (commented):
I like it, thanks @mushao999. I left a few small comments.

mushao999:
@DaveCTurner  Thanks for your suggestion, I've updated the code.

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
mushao999: I didn't change `RepositoryException` to `Exception` here ,because `RepositoriesService#createRepository` have done such thing and make sure exception is `RepositoryException`.

mushao999:
Should we add `fail("xxx");` here?https://github.com/elastic/elasticsearch/blob/caa0b618120004a8406470fead60eb3bf2b0cd5b/test/framework/src/main/java/org/elasticsearch/test/hamcrest/ThrowableAssertions.java#L27

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
DaveCTurner: Ah good point

DaveCTurner:
> Should we add fail("xxx"); after code.run();?

Yes, I think so.

mushao999:
> > Should we add fail("xxx"); after code.run();?
> 
> Yes, I think so.

so we add it in this PR or maybe open a new one to add it?

DaveCTurner:
Here would be fine with me.

mushao999:
> Here would be fine with me.

code added

DaveCTurner:
Looks like your branch is based off an old commit, you will need to merge latest master before CI will pass.

server/src/test/java/org/elasticsearch/repositories/InvalidRepositoryTests.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public void testShouldNotThrowWhenApplyingLifecycleChanges() {
+        repository.start();
+        repository.stop();
+    }
+
+    public void testShouldNotThrowWhenClosingToAllowRemovingRepo() {
+        repository.close();
+    }
```
DaveCTurner, line 34: I think the lifecycle things are covered adequately by the tests that use this repo so we can drop these.

server/src/test/java/org/elasticsearch/repositories/InvalidRepositoryTests.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class InvalidRepositoryTests extends ESTestCase {
+
+    private InvalidRepository repository = new InvalidRepository(
+        new RepositoryMetadata("name", "type", Settings.EMPTY),
+        new RepositoryException("name", "failed to create repository")
+    );
+
+    public void testShouldThrowWhenGettingMetadata() {
+        expectThrows(RepositoryException.class, () -> repository.getSnapshotGlobalMetadata(new SnapshotId("name", "uuid")));
```
DaveCTurner, line 24: Could we assert that the message/inner exception are as expected here?

test/framework/src/main/java/org/elasticsearch/test/hamcrest/ThrowableAssertions.java
```diff
@@ -25,6 +26,7 @@ public static void assertThatThrows(
     ) {
         try {
             code.run();
+            fail("Exception with type " + exceptionType.getName() + "should be thrown");
```
DaveCTurner, line 29: Ah, this won't totally work because we catch `Throwable` below. We'll need to set a `success` flag here and then assert that `success == false` outside the `catch` block.

x-pack/plugin/repository-encrypted/src/internalClusterTest/java/org/elasticsearch/repositories/encrypted/EncryptedRepositorySecretIntegTests.java
```diff
@@ -420,7 +420,16 @@ public void testSnapshotIsPartialForMissingPassword() throws Exception {
...
-                .allMatch(shardFailure -> shardFailure.reason().contains("[" + repositoryName + "] missing"))
+                .allMatch(
+                    shardFailure -> shardFailure.reason()
+                        .contains(
+                            "RepositoryException[["
+                                + repositoryName
+                                + "] failed to create repository]; nested: IllegalArgumentException[Secure setting [repository.encrypted."
+                                + repositoryName
+                                + ".password] must be set]"
+                        )
```
DaveCTurner, line 431: I think this over-constrains the message, really the only important bit is to report the problem about the setting:

```suggestion
                    shardFailure -> shardFailure.reason()
                        .contains("Secure setting [repository.encrypted." + repositoryName + ".password] must be set")
```

server/src/test/java/org/elasticsearch/repositories/RepositoriesServiceTests.java
```diff
@@ -212,6 +214,78 @@ public void onFailure(Exception e) {
...
+        repositoriesService.registerRepository(request, new ActionListener<>() {
+            @Override
+            public void onResponse(AcknowledgedResponse acknowledgedResponse) {
+                assertTrue(acknowledgedResponse.isAcknowledged());
+                assertThat(repositoriesService.repository(repoName), isA(TestRepository.class));
+            }
+
+            @Override
+            public void onFailure(Exception e) {
+                fail("[" + repoName + "] repository type [test] faield to create");
```
DaveCTurner, line 284: Suggest failing like this so we see the cause:

```suggestion
                assert false : e;
```

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
```diff
@@ -515,13 +515,15 @@ public void applyClusterState(ClusterChangedEvent event) {
...
                     }
                 } else {
                     try {
                         repository = createRepository(repositoryMetadata, typesRegistry, RepositoriesService::createUnknownTypeRepository);
                     } catch (RepositoryException ex) {
                         logger.warn(() -> new ParameterizedMessage("failed to create repository [{}]", repositoryMetadata.name()), ex);
+                        repository = new InvalidRepository(repositoryMetadata, ex);
                     }
                 }
                 if (repository != null) {
```
DaveCTurner, line 529: `repository` cannot be `null` here any more I think.

server/src/main/java/org/elasticsearch/repositories/InvalidRepository.java
```diff
@@ -0,0 +1,190 @@
...
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * This class represents a repository that failed during creation.
+ * This could happen when repository plugin is not stable enough to guarantee repository creation success all the time
+ */
```
DaveCTurner, line 34: Rewording suggestion:

```suggestion
/**
 * Represents a repository that exists in the cluster state but could not be instantiated on a node, typically due to invalid configuration.
 */
```

docs/changelog/85551.yaml
```diff
@@ -0,0 +1,5 @@
+pr: 85534
+summary: "Fix NPE in RepositoriesService"
```
DaveCTurner, line 2: Think this would be better for the release notes:

```suggestion
summary: "Distinguish missing and invalid repositories"
```

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.elasticsearch.test.hamcrest.ElasticsearchAssertions.assertAcked;
+import static org.elasticsearch.test.hamcrest.ThrowableAssertions.assertThatException;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.isA;
+
+@ESIntegTestCase.ClusterScope(supportsDedicatedMasters = false, numDataNodes = 3)
```
DaveCTurner, line 39: I would prefer to have the test rely less on the exact shape of the cluster (I'll add more comments about how to do that inline)...

```suggestion
```

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
```diff
@@ -0,0 +1,146 @@
...
+            }
+
+            @Override
+            public List<Setting<?>> getSettings() {
+                return List.of(UNSTABLE_NODES);
+            }
+        }
+    }
+
+    public void testDuplicateCreateRepository() throws Exception {
```
DaveCTurner, line 91: Why "duplicate"? How about this name?

```suggestion
    public void testCreateInvalidRepository() throws Exception {
```

(similarly in the class name and the repository name)

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
```diff
@@ -0,0 +1,146 @@
...
+            Arrays.stream(internalCluster().getNodeNames())
+                .filter(name -> name.equals(internalCluster().getMasterName()) == false)
+                .toArray()
+        );
+
+        // put repository for the first time: only let master node create repository successfully
+        createRepository(
+            repositoryName,
+            UnstableRepository.TYPE,
+            Settings.builder().put("location", randomRepoPath()).put(UnstableRepository.UNSTABLE_NODES.getKey(), unstableNode)
```
DaveCTurner, line 103: Suggest using `putList` instead of `put`, and drop the `Strings.arrayToCommaDelimitedString` above.

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
```diff
@@ -0,0 +1,146 @@
...
+            @Override
+            public List<Setting<?>> getSettings() {
+                return List.of(UNSTABLE_NODES);
+            }
+        }
+    }
+
+    public void testDuplicateCreateRepository() throws Exception {
+        final String repositoryName = "test-duplicate-create-repo";
+        String unstableNode = Strings.arrayToCommaDelimitedString(
```
DaveCTurner, line 93: We need at least one node that isn't the master, but we can do that by calling `internalCluster().ensureAtLeastNumDataNodes(2)` or `internalCluster().startDataOnlyNode()`.

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
```diff
@@ -0,0 +1,146 @@
...
+
+        // put repository for the first time: only let master node create repository successfully
+        createRepository(
+            repositoryName,
+            UnstableRepository.TYPE,
+            Settings.builder().put("location", randomRepoPath()).put(UnstableRepository.UNSTABLE_NODES.getKey(), unstableNode)
+        );
+        // verification should fail with some node has InvalidRepository
+        try {
+            client().admin().cluster().prepareVerifyRepository(repositoryName).get();
```
DaveCTurner, line 107: Suggest using `expectThrows` here:

```suggestion
        final var verificationException = expectThrows(RepositoryVerificationException.class,
            () -> client().admin().cluster().prepareVerifyRepository(repositoryName).get());
```

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
```diff
@@ -0,0 +1,146 @@
...
+            fail("verification should fail when some node failed to create repository");
+        } catch (Exception e) {
+            assertThat(e, isA(RepositoryVerificationException.class));
+            assertEquals(e.getSuppressed().length, 2);
```
DaveCTurner, line 111: Should be able to replace that `2` with `internalCluster().numDataAndMasterNodes() - 1`

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
```diff
@@ -0,0 +1,146 @@
...
+            assertThatException(
+                e.getSuppressed()[0].getCause(),
+                RepositoryException.class,
+                equalTo("[" + repositoryName + "] repository type [" + UnstableRepository.TYPE + "] failed to create on current node")
```
DaveCTurner, line 115: Could we also assert that the inner exception reports the cause properly?

DaveCTurner (changes requested):
Looks good, I left a handful of comments and suggested changes but structurally this is good.

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
mushao999: We do not want data only node. What we really want is node which fail to create repository and then selected as master later.

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
mushao999: We can not use `expectThrows` if we want to assert inner exception.

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
mushao999: I named it `duplicate` because NPE happend when we put repository for the second time.

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
mushao999: yes, so I remove the `if (repository != null)` and add a asserting here

test/framework/src/main/java/org/elasticsearch/test/hamcrest/ThrowableAssertions.java
mushao999: ah, you are right. I'll update it and another same logci in `RepositoriesDuplicatedCreationIT`

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
mushao999: I thought repository is only created in data node so I add the settings. Aware of repository is created on all nodes, I think our need no special shape of cluster. I've removed this line.

server/src/internalClusterTest/java/org/elasticsearch/repositories/RepositoriesDuplicatedCreationIT.java
mushao999: Thanks for your suggestion.

server/src/internalClusterTest/java/org/elasticsearch/repositories/InvalidRepositoryIT.java
```diff
@@ -0,0 +1,151 @@
...
+                    UnstableRepository.UNSTABLE_NODES.getKey(),
+                    Arrays.stream(internalCluster().getNodeNames())
+                        .filter(name -> name.equals(internalCluster().getMasterName()) == false)
+                        .toList()
+                )
+        );
+        // verification should fail with some node has InvalidRepository
+        boolean verifyPass = false;
+        try {
+            client().admin().cluster().prepareVerifyRepository(repositoryName).get();
```
DaveCTurner, line 109: Mentioned before, but this should be `final var repositoryVerificationException = expectThrows(RepositoryVerificationException.class, () -> client().admin().cluster().prepareVerifyRepository(repositoryName).get());`

Note that `expectThrows` returns the exception it caught so you can assert things about it such as its causes and suppressed exceptions.

DaveCTurner (commented):
I think you missed that `expectThrows()` returns the exception so you can assert more things about it. I think we can use `expectThrows` rather than the utilities in `ThrowableAssertions` here, and I'll open a follow-up to remove `ThrowableAssertions` entirely.

Edit: see https://github.com/elastic/elasticsearch/pull/85671

DaveCTurner:
See https://github.com/elastic/elasticsearch/pull/85671 for the PR to remove ThrowableAssertions in favour of using expectThrows.

DaveCTurner (approved):
LGTM, thanks for the extra iterations @mushao999

PR #85565, opened 2022-03-31
Make the ILM Move to Error Step Batched

### LINKED ISSUE #81880: Batch up failure-related ILM master tasks
In https://github.com/elastic/elasticsearch/pull/78547 we introduced batching for the ILM master tasks that occur on the happy path. However if a high-shard-count cluster encounters problems while doing ILM-related things—perhaps some nodes are temporarily unavailable for taking a snapshot—then we process the resulting `ilm-retry-failed-step` and `ilm-move-to-error-step` tasks one-by-one which can significantly delay the cluster's recovery from its problems.

We should batch these things together too.

It looks like we also enqueue duplicate `ilm-retry-failed-step` on each poll interval too, although we do appear to treat the duplicates as no-ops at execution time.

Relates https://github.com/elastic/elasticsearch/issues/77466

original-brownbear:
Simple solution by just moving this to the infrastructure we already have for the other
step tasks. There might be optimizations here still but this should be a big enough improvement
in the short term for all practical cases.

closes #81880


x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/MoveToErrorStepUpdateTask.java
```diff
@@ -73,25 +70,35 @@ public ClusterState execute(ClusterState currentState) {
...
+                index,
+                currentStepKey
+            ),
+            e
         );
-        if (ExceptionsHelper.unwrap(e, NotMasterException.class, FailedToCommitClusterStateException.class) != null) {
-            logger.debug(messageSupplier, e);
-        } else {
-            logger.error(messageSupplier, e);
-            assert false : new AssertionError("unexpected exception", e);
```
dakrone, line 94: I honestly don't remember or know why we have this assert, but are we okay dropping it?

dakrone (approved):
LGTM, thanks Armin!

original-brownbear:
Jenkins run elasticsearch-ci/bwc (Jenkins issue)

original-brownbear:
@elasticmachine update branch (there's some BwC serialization failure ...)

original-brownbear:
Npnp +Thanks Lee!

Issue #84481, opened 2022-03-01
Expand CCS tests coverage


javanna:
We currently have [integration tests](https://github.com/elastic/elasticsearch/tree/master/qa/multi-cluster-search) that cover cross-cluster search. What they don't cover is cross-cluster communication across multiple versions. We should make sure that these tests run against the previous minor version and double check their coverage (e.g. do they use the fields API? do they also cover field_caps? Which features of search and field_caps they cover?).

As a natural follow-up, it might make sense to consider reusing the existing field_caps and search tests yaml tests and run them in a CCS scenario, so that we don't need to maintain parallel CCS tests manually, as they tend to be forgotten over time.

Perhaps related to #78472 as it involves more CCS testing

Currently the steps we identified to reach better coverage are

- [x] run the existing qa:multi-cluster-search also against the previous minor version (#85281)
- [x] extend the ability to skip yaml for certain remote cluster versions (#85566) (follow up needed for #85281)
- [x] investigate if we should extend qa:multi-cluster-search to also test forward compatibility, i.e. all wire compatible local cluster versions agains a "current" remote ( #86521)
- [x] investigate if we can re-use existing search rest tests in a multi-cluster CCS setup (#86521)
- [x] investigate if the `proxy mode` in CCS is covered well in tests or if it also needs more coverage
* we decided that the [current coverage](https://github.com/elastic/elasticsearch/issues/84481#issuecomment-1128842367) is okay and added randomization of the connection mode for better coverage in (#86865)

javanna:
> From what I can see so far it seems to me we do have some good coverage of the basic scenarios here. We might consider moving randomization of the connection to AbstractMultiClustersTestCase (similar to the setup in SearchStatesIT) and see if that breaks any existing tests. Also we can probably randomize the connection mode for the CCS test we currently add in https://github.com/elastic/elasticsearch/pull/86521

Agreed @cbuescher thanks for digging.

javanna:
@cbuescher sounds like this can be closed? :)

PR #85566, opened 2022-03-31
[Tests] Enhance multi cluster search test skipping

cbuescher:
Currently the "skip" section in out rest yaml tests take into account the lowest
minor version of nodes in the connected local cluster. For some multi cluster
CCS test we will also need the ability to skip certain tests based on the
connected remote cluster version, e.g. if in bwc tests some functionality isn't
available yet on some bwc versions we test against. This change adds that
ability to yaml rest test in the :qa:multi-cluster-search module.

Relates to #84481


qa/multi-cluster-search/src/test/java/org/elasticsearch/search/MultiClusterSearchYamlTestSuiteIT.java
```diff
@@ -13,12 +13,40 @@
...
+
+    protected ClientYamlTestExecutionContext createRestTestExecutionContext(
+        ClientYamlTestCandidate clientYamlTestCandidate,
+        ClientYamlTestClient clientYamlTestClient,
+        boolean randomizeContentType
+    ) {
+        ClientYamlTestExecutionContext restTestExecutionContext = super.createRestTestExecutionContext(
+            clientYamlTestCandidate,
+            clientYamlTestClient,
+            randomizeContentType()
```
javanna, line 44: you are ignoring the last argument?

test/yaml-rest-runner/src/main/java/org/elasticsearch/test/rest/yaml/ClientYamlTestExecutionContext.java
```diff
@@ -213,6 +214,24 @@ public Version esVersion() {
         return clientYamlTestClient.getEsVersion();
     }
 
+    /**
+     * @return the version that skipping tests is based on, usually the same as {@link #esVersion()}
+     */
+    public Version skipVersion() {
```
javanna, line 220: I would not call this skipVersion, but rather getEsVersion like in the test client. For the multi_cluster tests, the es version will take into account the remote cluster.

qa/multi-cluster-search/src/test/java/org/elasticsearch/search/MultiClusterSearchYamlTestSuiteIT.java
```diff
@@ -13,12 +13,40 @@
...
 @TimeoutSuite(millis = 5 * TimeUnits.MINUTE) // to account for slow as hell VMs
 public class MultiClusterSearchYamlTestSuiteIT extends ESClientYamlSuiteTestCase {
 
+    private static Version remoteEsVersion = null;
+
+    @BeforeClass
+    public static void determineRemoteClusterMinimumVersion() {
+        String remoteClusterVersion = System.getProperty("tests.rest.remote-cluster-version");
+        if (remoteClusterVersion != null) {
+            remoteEsVersion = Version.fromString(remoteClusterVersion);
```
javanna, line 32: I wonder if we should take the remote cluster version blindly or rather the min between remote and local. That looks not needed today but maybe it is more future proof?

javanna (changes requested):
I left some comments, thanks @cbuescher

qa/multi-cluster-search/build.gradle
```diff
@@ -64,6 +64,7 @@ BuildParams.bwcVersions.withWireCompatible(ccsSupportedVersion) { bwcVersion, ba
     doFirst {
       nonInputProperties.systemProperty('tests.rest.suite', 'multi_cluster')
       nonInputProperties.systemProperty('tests.rest.cluster', localCluster.map(c -> c.allHttpSocketURI.join(",")))
+      nonInputProperties.systemProperty('tests.rest.remote-cluster-version', bwcVersion.toString())
```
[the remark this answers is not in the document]
cbuescher: Of course, that was my initial plan, but talking to Nhat about this we both thought that since we control the version of the remote cluster here completely this would just complicate determining the remote version unnecessarily. I think we would need to pass in the remote clusters transport URIs to the MultiClusterSearchYamlTestSuiteIT, create a client for the "_cat_nodes" call and then use that. But since we know the version here already I'm not convinced about what benefit this would give us.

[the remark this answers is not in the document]
javanna: I see what you mean. right the two can't get out of sync because this one determines which remote cluster we test against. I am fine with this. Nit: should we use remote_cluster_version (aligned with the multi_cluster above)?

qa/multi-cluster-search/src/test/java/org/elasticsearch/search/MultiClusterSearchYamlTestSuiteIT.java
javanna: I guess if the bwc version is provided through the system property, we should not complicate things here for no reason.

test/yaml-rest-runner/src/main/java/org/elasticsearch/test/rest/yaml/ESClientYamlSuiteTestCase.java
```diff
@@ -162,6 +162,14 @@ public void initAndResetContext() throws Exception {
         restTestExecutionContext.clear();
     }
 
+    protected ClientYamlTestExecutionContext createRestTestExecutionContext(
+        ClientYamlTestCandidate clientYamlTestCandidate,
+        ClientYamlTestClient clientYamlTestClient,
+        boolean randomizeContentType
+    ) {
+        return new ClientYamlTestExecutionContext(clientYamlTestCandidate, clientYamlTestClient, randomizeContentType());
```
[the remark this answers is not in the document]
cbuescher: I changed this as part of another refactoring

qa/multi-cluster-search/src/test/java/org/elasticsearch/search/MultiClusterSearchYamlTestSuiteIT.java
cbuescher: done as part of the other changes

cbuescher:
@javanna I pushed changes that I think simplify this PR quite a bit, I think this is ready for another look

qa/multi-cluster-search/src/test/java/org/elasticsearch/search/MultiClusterSearchYamlTestSuiteIT.java
```diff
@@ -13,12 +13,47 @@
...
+    }
+
+    protected ClientYamlTestExecutionContext createRestTestExecutionContext(
+        ClientYamlTestCandidate clientYamlTestCandidate,
+        ClientYamlTestClient clientYamlTestClient
+    ) {
+        return new ClientYamlTestExecutionContext(clientYamlTestCandidate, clientYamlTestClient, randomizeContentType()) {
+
+            /**
+             * for mixed cluster tests we also take the remote cluster version into account and return the minimum here
```
javanna, line 43: I assume that skipping tests is the only way we use this esVersion? Maybe worth adding a comment here to explain that we use this to skip tests, it may not be straight-forward to figure out from the code.

javanna (approved):
LGTM

qa/multi-cluster-search/src/test/java/org/elasticsearch/search/MultiClusterSearchYamlTestSuiteIT.java
cbuescher: Updated the docs

PR #60024, opened 2020-07-21
DOC Mapping updates are deprecated for ingestion privileges

albertzaharovits:
I've forgot to adjust the docs and the migration notice following https://github.com/elastic/elasticsearch/pull/58784 .

The text describes the deprecated behaviour in the next subsequent minor releases, which then turns into breaking behaviour in version 8, but the code (and the accompanying  notice) that it refers to, is still due. I will raise that PR promptly.

PR #85570, opened 2022-03-31
DOC Reinstate deprecation for ingest privileges with mapping update

albertzaharovits:
We missed actually breaking the deprecation from https://github.com/elastic/elasticsearch/pull/60024 in v8.
Consequently, the deprecated behavior still works in 8.* and it still generates deprecation logs and headers.
This PR reinstates the doc side of the deprecation notice, the plan being that we're now going to break the behavior in v9.


x-pack/docs/en/security/authorization/privileges.asciidoc
```diff
@@ -199,8 +199,12 @@ bulk request on an index or data stream that contains new fields that may
 be mapped rather than an explicit <<indices-put-mapping,update mapping>> request.
 
 `create`::
-Privilege to index documents. Also grants access to the update mapping
-action.
+Privilege to index documents.
++
+deprecated:[8.2] Also grants the permission to update the index mapping (but not
+the data stream mapping), using the update mapping action, or relying on
+{ref}/dynamic-mapping.html[dynamic mappings]. In the next major release,
```
bytebilly, line 206: Should we use a more generic statement like "in a next major release" instead? Explicitly mentioning the next major release already caused some problems in the past, and I'm not sure that it gives more value than a generic one. The same in all other changes.

What do you think?

lockewritesdocs:
@elasticmachine update branch

lockewritesdocs:
>Consequently, the deprecated behavior still works in 8.* and it still generates deprecation logs and headers.

@albertzaharovits, should we also apply these changes to `8.0`? The addition to the migration guide in #60024 explains the changes to these privileges, so I think that the privileges themselves should indicate the deprecation information that this PR adds, especially if messages are generated in the deprecation logs.

x-pack/docs/en/security/authorization/privileges.asciidoc
lockewritesdocs: I suggested "a future major release" in my comments.

lockewritesdocs (commented):
Overall, LGTM. I left some questions and suggested changes that you can consider incorporating.

x-pack/docs/en/security/authorization/privileges.asciidoc
albertzaharovits: I have accepted "a future major release".

x-pack/docs/en/security/authorization/privileges.asciidoc
```diff
@@ -199,8 +199,12 @@ bulk request on an index or data stream that contains new fields that may
...
+this privilege will not grant any mapping update permission.
```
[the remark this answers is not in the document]
albertzaharovits: I may be misunderstanding your suggestion, but I think the note is still valid, even in the presence of the deprecated behavior.
Both privileges exhibit the deprecated behavior but `create_doc` is still a restricted version of `create` and users requiring strict append should be aware of it.

albertzaharovits:
> @albertzaharovits, should we also apply these changes to 8.0? 

Yes, I think these doc deprecation notices should get into all 8.* release branches.

albertzaharovits:
@lockewritesdocs please take another look.

lockewritesdocs (approved):
LGTM 🦖  Thanks @albertzaharovits!

PR #85574, opened 2022-03-31
Fix BWC versions to 8.3.0

joegallo:
Follow up to #85504 (that PR touched 8.2, only, but it has BWC implications w.r.t. master, too, which I had failed to consider).

Updates the BWC versions on master for #83832, since that was actually removed from the 8.2 branch on #85504.

I reproduced the failing build locally and confirmed that this commit fixes that.


imotov (approved):
LGTM

PR #85349, opened 2022-03-24
Redirect testclusters ES process stdout/stderr to log file

rjernst:
Although most of the time the stdout and stderr of the Java process is
not needed since ES has its own log file, there are some cases where it
is very important to have the output before ES has initialized its
logging system. This commit changes the stdout/stderr of the ES process
created by testclusters to use the same log file that ES will write to.
There should not be contention with ProcessBuilder and ES trying to
write at the same time because ES swaps out the stream handles to write
to its logging system, so after that point ProcessBuilder will never see
anymore output.

relates #68333

PR #85576, opened 2022-03-31
Do not reuse ES log file in testclusters

### LINKED ISSUE #85553: Test logs include duplicate lines
### Elasticsearch Version

8.2.0+, 8.1.2+, 7.17+

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

n/a

### Problem Description

After https://github.com/elastic/elasticsearch/pull/85349 log lines are now duplicated in the output.

### Steps to Reproduce

Simply run a cluster with `./gradlew run` and notice the log duplication

### Logs (if relevant)

_No response_

rjernst:
In #85349 we restored capturing the stdout and stderr of Elasticsearch
processes started by testclusters. However, since testclusters launches
Elasticsearch in the forground for each process, the log output is
duplicated to stdout.

This commit changes the stdout redirect back to discard. This should be
ok for the purposes of capturing the process output, since we want to
capture jvm startup errors which should be sent to stderr.

closes #85553


mark-vieira (commented):
Yeah, this is kind of why my preference would be to _disable_ file logging, similarly to how we do for our Docker images and then just capture everything via stdout.

Won't we _still_ have duplication here though, just only for errors since we'll effectively be capturing anything written to stderr twice.

mark-vieira:
> We don't write to stderr in ES, so no, we should not have duplication.

So even errors are written to stdout? That doesn't seem right.

mark-vieira:
> The only case we log to stderr and to the log that I see is in the uncaught exception handler, but I don't think that matters here (and it is something we should be able to fix).

Well I think in cases where we have nodes explode (a common example is assertion errors) we'll get duplicate logs. I suppose that's still better.

Couldn't one option be to capture the log output from stdout/stderr and use that for test clusters reporting, and leave the file log created by default as it is? We'd be duplicating log output in terms of disk storage but it's probably not a ton.

rjernst:

mark-vieira:
> I don't think this would work currently because in Gradle we write a bunch of pre-start information to the log file before trying to start the process, and so rely on reading from the log file to capture that later on failure.

What I'm proposing is that the processbuilder captures stdout and stderr and writes it to a different location than the standard ES log file and we use _that_ for error reporting in the build.

rjernst:
Great idea Mark, I understand now. I pushed c575200

mark-vieira (approved):
LGTM

PR #85581, opened 2022-03-31
Bump commons-compress to 1.21

dakrone:
This updates the dependency version for the Apache commons compress project to 1.21


plugins/ingest-attachment/build.gradle
```diff
@@ -55,7 +55,7 @@ dependencies {
   // MS Office
   api "org.apache.poi:poi-scratchpad:${versions.poi}"
   // Apple iWork
-  api 'org.apache.commons:commons-compress:1.19'
+  api 'org.apache.commons:commons-compress:1.21'
```
mark-vieira, line 58: As far as I can see this is the only use of this dependency in actual production code. Everything else is either build logic or test code. Does that seem right?

mark-vieira (approved):

plugins/ingest-attachment/build.gradle
dakrone: Yep that's correct

rjernst (approved):
LGTM

dakrone:
@elasticmachine update branch

dakrone:
@elasticmachine update branch

PR #85585, opened 2022-03-31
Use ByteBuffer as slice holder in BytesReferenceStreamInput

rjernst:
This commit converts the handle for each slice in
BytesReferenceStreamInput to be a ByteBuffer. The ByteBuffer class
better interoperates with JDK based methods for decoding streams, and
itself uses intrinsics internally to provide faster decoding.


server/src/main/java/org/elasticsearch/common/bytes/BytesReferenceStreamInput.java
```diff
@@ -80,32 +84,38 @@ public long readLong() throws IOException {
...
     public int readVInt() throws IOException {
-        if (slice.length - sliceIndex >= 5) {
-            final byte[] buf = slice.bytes;
-            final int offset = slice.offset;
-            byte b = buf[offset + sliceIndex++];
+        if (slice.remaining() >= 5) {
+            final byte[] buf = slice.array();
+            final int offset = slice.arrayOffset();
+            int position = slice.position();
+            byte b = buf[offset + position++];
```
ChrisHegarty, line 91: why not just use ByteBuffer::get here? (rather than operating on the underlying byte array)

ChrisHegarty (commented):
This is a great change - would be good to see how if affects performance, but it should have a net positive affect.

original-brownbear:
++ I like this, looks much better than what we have now.

It's hard to tell though what this does for us in terms of performance in the real world.
It's just too close there for numbers in a real world test and it beats the current solution in JMH. The reason for that though, at least to me seems to be that we eventually  inline the slow path in the existing solution (at which point the benchmark visibly slows down) for the variable length encodings but don't inline it in this solution because the fast path compiles larger (this might change if we use `get` as suggested, didn't try it).
It seems this also comes with a bit of a slowdown in string reading because we don't inline the `vint` read there anymore. But that would probably also change by just extracting the exception throwing from `readArraySize`.

Would be nice if we had benchmarks on the deserialization of some of our very hot transport messages but unfortunately we don't.

ChrisHegarty:
>  It seems this also comes with a bit of a slowdown in string reading because we don't inline the `vint` read there anymore. But that would probably also change by just extracting the exception throwing from `readArraySize`.

Maybe we should roll up vint and vlong, like what was done in https://github.com/apache/lucene/pull/602

> Would be nice if we had benchmarks on the deserialization of some of our very hot transport messages but unfortunately we don't.

It would be great to have such.

rjernst:
@original-brownbear @ChrisHegarty I've brought this PR up to date, I think we should try to get it in, can you take another look?

original-brownbear (approved):
LGTM, if CI is happy I'm happy :)

ChrisHegarty (approved):
LGTM

PR #85590, opened 2022-04-01
[Test] Add a dedicate helper class for randomizing Authentication

ywangd:
The Authentication class has many internal logics about what values can
or cannot be combined together. It is not straightforward to always get
the logic right when trying to create such an object for tests. This
could lead to spurious failures or incomplete test coverage.

This PR adds a helper class for creating such an object with necessary
configurabililty. The relevant methods in AuthenticationTests now
delegate to the new class to avoid having to touch too many files in one
PR. The ultimate goal is to have it used in every place where an
Authentication object is needed for test to replace any calls to
constructors or mocking.


albertzaharovits (approved):
LGTM
It looks to me that all sorts possible of `Authentication`s can be built with this new `AuthenticatioinTestHelper` class.

PR #85595, opened 2022-04-01
Fix fields wildcard support to vector tile search API

### LINKED ISSUE #85592: Wildcards in fields parameter are not supported in vector tile search API
Currently if you add a field in the fields parameter with a wildcard, no fields will be added to the hits layer. The situation is even worst if the wildcard matches the field name of the geo_point or geo_shape field, in that case a ClassCastException is thrown.

iverase:
We are currently fetching fields added by the fields API by name which breaks support for wildcards. More over, in case wildcards are used and those wildcards matches the geo field, then a ClassCastException is thrown.

This commit changes the strategy on how those fields are fetched by just iterating over the returned map. In addition, the geo field is added the last one on the fields list so it does not get overridden.

fixes  https://github.com/elastic/elasticsearch/issues/85592


x-pack/plugin/vector-tile/src/main/java/org/elasticsearch/xpack/vectortile/rest/RestVectorTileAction.java
```diff
@@ -265,12 +266,10 @@ private static SearchRequestBuilder searchRequestBuilder(RestCancellableNodeClie
                 featureBuilder.mergeFrom((byte[]) feature);
                 VectorTileUtils.addPropertyToFeature(featureBuilder, layerProps, ID_TAG, searchHit.getId());
                 VectorTileUtils.addPropertyToFeature(featureBuilder, layerProps, INDEX_TAG, searchHit.getIndex());
-                if (fields != null) {
-                    for (FieldAndFormat field : fields) {
-                        final DocumentField documentField = searchHit.field(field.field);
-                        if (documentField != null) {
-                            VectorTileUtils.addPropertyToFeature(featureBuilder, layerProps, field.field, documentField.getValue());
-                        }
+                final Map<String, DocumentField> fields = searchHit.getFields();
```
craigtaverner, line 269: I presume `searchHits.getFields()` contains the subset of `request.getFieldAndFormats()` that was actually found? So this way we skip fields not found?

x-pack/plugin/vector-tile/src/main/java/org/elasticsearch/xpack/vectortile/rest/VectorTileUtils.java
```diff
@@ -60,8 +60,10 @@ public static void addToXContentToFeature(VectorTile.Tile.Feature.Builder featur
      * Adds the provided key / value pair into the feature as tags.
      */
     public static void addPropertyToFeature(VectorTile.Tile.Feature.Builder feature, MvtLayerProps layerProps, String key, Object value) {
-        feature.addTags(layerProps.addKey(key));
-        feature.addTags(layerProps.addValue(value));
+        if (key != null && value != null) { // just make sure we don't try to add a null values
```
craigtaverner, line 63: I presume this defensive coding it not really required because the fix in RestVectorTileAction should already remove the nulls? Do we have this code in case there are other code paths we've not thought of that might cause nulls? Would it make more sense to throw an error here so we don't hide bugs?

craigtaverner (approved):
It looks good, but I did have a few clarifying questions.

x-pack/plugin/vector-tile/src/main/java/org/elasticsearch/xpack/vectortile/rest/RestVectorTileAction.java
```diff
@@ -177,10 +178,11 @@ private static SearchRequestBuilder searchRequestBuilder(RestCancellableNodeClie
         searchRequestBuilder.setFetchSource(false);
         searchRequestBuilder.setTrackTotalHitsUpTo(request.getTrackTotalHitsUpTo());
         String args = request.getZ() + "/" + request.getX() + "/" + request.getY() + "@" + request.getExtent() + ":" + request.getBuffer();
```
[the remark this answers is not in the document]
iverase: done

x-pack/plugin/vector-tile/src/main/java/org/elasticsearch/xpack/vectortile/rest/RestVectorTileAction.java
```diff
@@ -177,10 +178,11 @@ private static SearchRequestBuilder searchRequestBuilder(RestCancellableNodeClie
...
-        searchRequestBuilder.addFetchField(new FieldAndFormat(request.getField(), "mvt(" + args + ")"));
         for (FieldAndFormat field : request.getFieldAndFormats()) {
             searchRequestBuilder.addFetchField(field);
         }
+        // added last so in case there is a wildcard, this one is picked
+        searchRequestBuilder.addFetchField(new FieldAndFormat(request.getField(), "mvt(" + args + ")"));
```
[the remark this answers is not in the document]
iverase: Yes, it populates a map so if it finds the same name it gets written over. If the behaviour changes, test should notice it.

x-pack/plugin/vector-tile/src/main/java/org/elasticsearch/xpack/vectortile/rest/RestVectorTileAction.java
iverase: I added a test for a non-existing field and for a field with null value. Everything seems to behave as expected

x-pack/plugin/vector-tile/src/main/java/org/elasticsearch/xpack/vectortile/rest/VectorTileUtils.java
iverase: I removed the defensive coding. I added more test and it seems we are never passing null values here.

iverase:

PR #85614, opened 2022-04-01
keep unassigned reasons

idegtiarenko:
Keep track of unassigned shards

### DIFF
--- a/server/src/internalClusterTest/java/org/elasticsearch/cluster/allocation/AwarenessAllocationIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/cluster/allocation/AwarenessAllocationIT.java
@@ -150,7 +150,6 @@ public class AwarenessAllocationIT extends ESIntegTestCase {
         assertThat(counts.get(B_0), anyOf(equalTo(2), equalTo(3)));
     }
 
-    // @AwaitsFix(bugUrl = "TODO")
     public void testAwarenessZonesIncrementalNodes() {
         Settings commonSettings = Settings.builder()
             .put("cluster.routing.allocation.awareness.force.zone.values", "a,b")
--- a/server/src/internalClusterTest/java/org/elasticsearch/cluster/allocation/AwarenessAllocationIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/cluster/allocation/AwarenessAllocationIT.java
@@ -150,7 +150,7 @@ public class AwarenessAllocationIT extends ESIntegTestCase {
         assertThat(counts.get(B_0), anyOf(equalTo(2), equalTo(3)));
     }
 
-    @AwaitsFix(bugUrl = "TODO")
+    // @AwaitsFix(bugUrl = "TODO")
     public void testAwarenessZonesIncrementalNodes() {
         Settings commonSettings = Settings.builder()
             .put("cluster.routing.allocation.awareness.force.zone.values", "a,b")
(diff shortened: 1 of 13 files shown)


server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceService.java
```diff
@@ -238,13 +238,19 @@ boolean updateDesiredBalanceAndReroute(DesiredBalanceInput desiredBalanceInput,
...
+        }
+
         logger.trace(
             hasChanges
                 ? "newer cluster state received, publishing incomplete desired balance and restarting computation"
                 : "desired balance computation converged"
         );
 
-        final DesiredBalance newDesiredBalance = new DesiredBalance(desiredAssignments);
+        final DesiredBalance newDesiredBalance = new DesiredBalance(desiredAssignments, unassigned);
```
idegtiarenko, line 253: We might consider only keeping only the count for now

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceService.java
```diff
@@ -238,13 +238,19 @@ boolean updateDesiredBalanceAndReroute(DesiredBalanceInput desiredBalanceInput,
             );
         }
 
+        final var unassigned = new HashMap<ShardId, UnassignedInfo>();
+        for (var ignored : routingAllocation.routingNodes().unassigned().ignored()) {
+            assert ignored.unassignedInfo() != null;
+            unassigned.put(ignored.shardId(), ignored.unassignedInfo());
```
DaveCTurner, line 244: I think we need something finer-grained than `ShardId` for the key - we might have just created a new copy for an existing shard (e.g. increased `number_of_replicas`). The primary could be assigned, the existing replica could be `DECIDERS_NO` and the new replica could be `NO_ATTEMPT`.

Also I think the only meaningful status here will be `DECIDERS_NO` - the other possibilities can only come from one of the existing-shards allocators I think.

server/src/internalClusterTest/java/org/elasticsearch/cluster/allocation/AwarenessAllocationIT.java
```diff
@@ -150,7 +150,7 @@ public void testAwarenessZones() {
         assertThat(counts.get(B_0), anyOf(equalTo(2), equalTo(3)));
     }
 
-    @AwaitsFix(bugUrl = "TODO")
+    // @AwaitsFix(bugUrl = "TODO")
```
DaveCTurner, line 153: Maybe just drop this?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceService.java
```diff
@@ -253,7 +262,7 @@ boolean updateDesiredBalanceAndReroute(DesiredBalanceInput desiredBalanceInput,
                     final var newNodes = desiredAssignment.getValue();
                     final var oldNodes = desiredBalance.desiredAssignments().get(shardId);
                     if (newNodes.equals(oldNodes)) {
-                        logger.trace("{} desired balance unchanged,   allocating to {}", shardId, newNodes);
+                        logger.trace("{} desired balance unchanged, allocating to {}", shardId, newNodes);
```
DaveCTurner, line 265: The extra spaces were to keep it lined up with the `now allocating` message in the `else` branch in the logs. I'm ok to remove them, but they seemed helpful sometimes.

DaveCTurner (approved):
LGTM, left a couple of nonblocking comments.

PR #85621, opened 2022-04-01
Speed up JsonXContentParser token conversion and text()

original-brownbear:
This reduces the compile size of convert token by about 20 bytes making it inline in a couple
of places that wouldn't inline before once `unknownTokenException` is deoptimized as dead code.
Same for the change to `text()` that saves some method size there so it inlines into the much used
textOrNull here and there.

With this change all calls in to `convertToken` in the common hot paths inline as do all the `isValue` checks from removing the strange overrides on the enum.


grcevski (approved):
LGTM!

original-brownbear:
Thanks Nikola!

PR #85623, opened 2022-04-01
Fix typos in comments

DaveCTurner:


server/src/main/java/org/elasticsearch/cluster/metadata/MetadataMappingService.java
```diff
@@ -225,9 +225,9 @@ public void putMapping(final PutMappingClusterStateUpdateRequest request, final
         for (Index index : request.indices()) {
             final IndexMetadata indexMetadata = metadata.index(index);
             if (indexMetadata == null) {
-                // local store recovery sends a mapping update request during application of a cluster state on t he data node which
-                // might we receive here before the CS update that created the index has been applied on all nodes and thus the index
-                // isn't found in the state yet but will be visible to the CS update below
+                // local store recovery sends a mapping update request during application of a cluster state on the data node which we might
+                // receive here before the CS update that created the index has been applied on all nodes and thus the index isn't found in
+                // the state yet but will be visible to the CS update below
```
dakrone, line 230:
```suggestion
                // local store recovery sends a mapping update request during application of a cluster state on the data node which we might
                // receive here before the CS update that created the index has been applied on all nodes and thus the index isn't found in
                // the state yet, but will be visible to the CS update below
```

test/framework/src/main/java/org/elasticsearch/index/engine/EngineTestCase.java
```diff
@@ -1137,22 +1137,18 @@ public static void assertOpsOnReplica(
...
             if (op instanceof Engine.Index) {
                 Engine.IndexResult result = replicaEngine.index((Engine.Index) op);
-                // replicas don't really care to about creation status of documents
-                // this allows to ignore the case where a document was found in the live version maps in
-                // a delete state and return false for the created flag in favor of code simplicity
-                // as deleted or not. This check is just signal regression so a decision can be made if it's
-                // intentional
+                // Replicas don't really care to about creation status of documents. This allows to ignore the case where a document was
+                // found in the live version maps in a delete state and return false for the created flag in favor of code simplicity as
+                // deleted or not. This check is just signal regression so a decision can be made if it's intentional.
```
dakrone, line 1142:
```suggestion
                // Replicas don't really care about the creation status of documents. This allows us to ignore the case where a document was
                // found in the live version maps in a delete state and return false for the created flag in favor of code simplicity as
                // deleted or not. This check is just a simple regression check so a decision can be made if it's intentional.
```

(I made some editorial changes based on what I *think* this is trying to say, but it was confusing so I could be incorrect)

test/framework/src/main/java/org/elasticsearch/index/engine/EngineTestCase.java
```diff
@@ -1137,22 +1137,18 @@ public static void assertOpsOnReplica(
...
             } else {
                 Engine.DeleteResult result = replicaEngine.delete((Engine.Delete) op);
-                // Replicas don't really care to about found status of documents
-                // this allows to ignore the case where a document was found in the live version maps in
-                // a delete state and return true for the found flag in favor of code simplicity
-                // his check is just signal regression so a decision can be made if it's
-                // intentional
+                // Replicas don't really care to about found status of documents. This allows to ignore the case where a document was found
+                // in the live version maps in a delete state and return true for the found flag in favor of code simplicity. This check is
+                // just signal regression so a decision can be made if it's intentional.
```
dakrone, line 1151:
```suggestion
                // Replicas don't really care about the "found" status of documents. This allows us to ignore the case where a document was found
                // in the live version maps in a delete state and return true for the found flag in favor of code simplicity. This check is
                // just a simple regression check so a decision can be made if it's intentional.
```

(same caveat as previous comment)

dakrone (approved):
LGTM, but please double check that my last two suggestions conform with the intended meaning of the comments (they are a little hard to understand as-is).

DaveCTurner:
++ thanks @dakrone, I think 093cca60f68119fe63165dede7debb6949dc07eb is closer to the intended meaning

PR #85625, opened 2022-04-01
[ML] Simplify shortcircuiting scores renormalizer

droberts195:
In #85555 a Phaser was used to track ongoing renormalizations
and wait for them to complete. However, this is a very complex
way to do things.

It turns out there's an easier way, as submitting a task to a
threadpool returns a Future that can be used to track when the
submitted task completes.

There was also a bug in the ShortCircuitingRenormalizer class
where the possibility of RejectedExecutionException was not being
considered.

### DIFF
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/job/process/normalizer/ShortCircuitingRenormalizer.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/job/process/normalizer/ShortCircuitingRenormalizer.java
@@ -152,18 +152,22 @@ public class ShortCircuitingRenormalizer implements Renormalizer {
         do {
             AugmentedQuantiles latestAugmentedQuantiles = getLatestAugmentedQuantilesAndClear();
             assert latestAugmentedQuantiles != null;
-            Quantiles latestQuantiles = latestAugmentedQuantiles.getQuantiles();
-            CountDownLatch latch = latestAugmentedQuantiles.getLatch();
-            try {
-                scoresUpdater.update(
-                    latestQuantiles.getQuantileState(),
-                    latestQuantiles.getTimestamp().getTime(),
-                    latestAugmentedQuantiles.getWindowExtensionMs()
-                );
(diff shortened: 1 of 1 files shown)


droberts195:
`>non-issue` as logically it's part of #85555.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/job/process/normalizer/ShortCircuitingRenormalizer.java
```diff
@@ -116,37 +116,53 @@ private synchronized AugmentedQuantiles getLatestAugmentedQuantilesAndClear() {
...
-        phaser.arriveAndDeregister();
+        semaphore.release();
+        latestTask = null;
         return true;
     }
 
     private void doRenormalizations() {
         do {
             AugmentedQuantiles latestAugmentedQuantiles = getLatestAugmentedQuantilesAndClear();
             assert latestAugmentedQuantiles != null;
```
benwtrent, line 154: Just because we do an assert, production runs with asserts off. We should still prevent an NPE.

droberts195: My thinking was that the logic is now simple enough that it cannot be `null`.  In the code before #85555 there was an expected way that it could be `null`.  In the code of #85555 it really should never have been `null`, but the `Phaser` was so complex that I wasn't convinced.  Whereas now I really cannot see any way that it could be `null`.

I guess it doesn't hurt to add an extra `if`.  But I'll also add a `TODO` to remove it if the `assert` doesn't trip in the next year or so.

benwtrent: I see that as well. THe logic now is much simpler and easier to understand. ++ refactor from me!

benwtrent (approved):
Much nicer!

PR #85628, opened 2022-04-01
Remove ILM docs redirect

joegallo:
https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-put-lifecycle.html looks like this:

Screen Shot 2022-04-01 at 11 09 32 AM

And the "ILM policy definition" link there goes to https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-policy-definition.html

But that page is just a redirect to https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-index-lifecycle.html, so lets link directly to that from https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-put-lifecycle.html and save the user a click.


docs/reference/ilm/set-up-lifecycle-policy.asciidoc
```diff
@@ -1,6 +1,6 @@
 [role="xpack"]
 [[set-up-lifecycle-policy]]
-== Configure a lifecycle policy [[ilm-policy-definition]]
+== Configure a lifecycle policy [[ilm-configure-policy]]
```
lcawl, line 3: > == Configure a lifecycle policy [[ilm-configure-policy]]

From the explanation in the PR, I'm not sure what the goal of this change is.  If your intention is to change https://www.elastic.co/guide/en/elasticsearch/reference/master/ilm-policy-definition.html, the source file is actually https://raw.githubusercontent.com/elastic/elasticsearch/master/docs/reference/redirects.asciidoc

lcawl: I'm actually surprised it didn't cause an error as-is, since the anchor IDs are supposed to be unique.  It makes me think it's not even recognized. My recommendation would be to remove that anchor rather than change it.

lcawl (approved):
LGTM, thanks!

PR #83055, opened 2022-01-25
[ML] add a frequent items aggregation

hendrikmuhs:
The PR adds an aggregation called `frequent_items`, a bucket aggregation which finds frequent item sets. It is a form of association rules mining that identifies items that often occur together. It also helps you to discover relationships between different data points (items).

For more information about usage have a look at #86037.

This implements frequent items using an algorithm called _eclat_.

PR #85630, opened 2022-04-01
Implement a serializable version of a BytesRefHash

hendrikmuhs:
Implement a serializable version of a BytesRefHash, useful to send hashtables over the wire

Relates #83055

**Background**

This PR is a spin off from a bigger task to implement a `frequent_items` aggregation. For this aggregation I need a data structure that collects data in form of largish frequency tables. These tables need to be centrally merged before the core algorithm mines for frequent items (I happily provide more details, just ping me). The closest re-usable structure I could find was `BytesRefHash` used in several places, however it wasn't able serialize and deserialize. It also turned out that hashing is not deterministic, the hash seed used internally is randomized. (Note: `BytesRefHash` has circuit breaker support)

This PR adds a serializable version of `BytesRefHash` by keeping changes to the current `BytesRefHash` minimal. Instead a new `SerializableBytesRefHash` got added which provides the necessary functionality.

Serialization/Deserialization: The implemented serialization encodes the internal data structures as is. However it uses variable length encodings to keep the size small. This works ok for the use case I am interested. It makes sense to think about smarter serialization schemes that avoids overhead due to data sparsity (e.g. RLE could work well), that's why the code encodes a serialization type. It might also be possible to just dump keys in order. I left this as future optimization (feel free to suggest better approaches).


server/src/main/java/org/elasticsearch/common/util/SerializableBytesRefHash.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        out.writeFloat(maxLoadFactor);
+        out.writeVLong(size);
+        out.writeVLong(maxSize);
+
+        long capacity = capacity();
+        out.writeLong(capacity);
```
droberts195, line 175: Could this be `writeVLong` or is there some situation when `capacity` can be negative?

server/src/main/java/org/elasticsearch/common/util/SerializableBytesRefHash.java
```diff
@@ -0,0 +1,214 @@
...
+        out.writeVLong(sizeOfBytes);
+        for (int i = 0; i < sizeOfBytes; ++i) {
+            out.writeByte(bytes.get(i));
+        }
+
+        // hashes
+        out.writeVLong(sizeOfHashes);
+
+        for (int i = 0; i < sizeOfHashes; ++i) {
+            out.writeZInt(hashes.get(i));
```
droberts195, line 203: I am not sure it's a good idea to use `writeZInt` here.  Looking at how it works, it takes 5 bytes to send a value where bits 28-31 are non-zero.  If the hash function is good then there's only a 1 in 16 chance of the 4 most significant bits all being zero.  So 15 out of 16 times this will send 5 bytes.  It seems like it would be more efficient to just use `writeInt` instead.

droberts195 (commented):
My biggest comment about this is how it's intended to be used when it gets to the receiving node.

In #80867 I also needed to serialize strings held in a `BytesRefHash` between nodes, but instead of serializing the `BytesRefHash` I serialized by converting back to strings and serializing those instead, then adding to a different `BytesRefHash` on the receiving node. For my use case a key requirement was assigning a unique numeric ID to particular strings.  This is needed for efficient storage and comparison of sequences of strings.  So when reducing the partial aggregation results I need all the partial results to have assigned unique IDs using the same `BytesRefHash`.  If all the `BytesRefHashes` from the different nodes were serialized to the coordinating node then they'd still need to be merged on the coordinating node.  Additionally, because of the way categorization discards the strings that aren't common between the string sequences then number of strings to be serialized is nowhere near as high as the number of strings seen in the raw data (unless we have the situation of just a handful of docs aggregated on each node, but then by definition the data volume is low anyway).

So my main comment is just to double check that this is needed at all.  Maybe the approach of serializing the actual strings would be too inefficient for your use case.  But if you eventually want all the strings from different nodes to be in the same `BytesRefHash` on the coordinating node and there's already been some degree of reduction on each data node then maybe it's OK.

hendrikmuhs:
First of all thanks for looking at it! Great feedback!

> So my main comment is just to double check that this is needed at all. Maybe the approach of serializing the actual strings would be too inefficient for your use case. But if you eventually want all the strings from different nodes to be in the same `BytesRefHash` on the coordinating node and there's already been some degree of reduction on each data node then maybe it's OK.

Yes, the hashes get merged, however I described it on slack:

> Right now, it transfers all data, including the internal hash values. But I wonder if I should simplify and only transfer keys and values. The hash values itself can be recalculated.
> It can be even further simplified: When merging the tables I actually don't need all of it. I only need the hashtable of the 1st table, for all other tables I only need keys and values.

I think the hashes are only necessary for appending and rehashing. I will check if I can drop the serialization of `hashes`. To solve my case, I can add a `Unmodifiable` version of the `SerializableBytesRefHash`, this is not only good for semantics, but I can skip the rehashing part for it.

PR #85507, opened 2022-03-30
Avoiding watcher validation errors when a data stream points to more than one index

masseyke:
We are seeing this error in 8.0.1 if a node is restarted after the `.watcher-history-16` data stream is pointing to more than one index.
```
[instance-0000000000] error validating to start watcher
java.lang.IllegalStateException: Alias [.watcher-history-16] points to more than one index
	at org.elasticsearch.xpack.watcher.watch.WatchStoreUtils.getConcreteIndex(WatchStoreUtils.java:32) ~[x-pack-watcher-8.1.0.jar:8.1.0]
	at org.elasticsearch.xpack.watcher.history.HistoryStore.validate(HistoryStore.java:79) ~[x-pack-watcher-8.1.0.jar:8.1.0]
	at org.elasticsearch.xpack.watcher.WatcherService.validate(WatcherService.java:158) [x-pack-watcher-8.1.0.jar:8.1.0]
	at org.elasticsearch.xpack.watcher.WatcherLifeCycleService.clusterChanged(WatcherLifeCycleService.java:162) [x-pack-watcher-8.1.0.jar:8.1.0]
	at org.elasticsearch.cluster.service.ClusterApplierService.callClusterStateListener(ClusterApplierService.java:564) [elasticsearch-8.1.0.jar:8.1.0]
	at org.elasticsearch.cluster.service.ClusterApplierService.callClusterStateListeners(ClusterApplierService.java:550) [elasticsearch-8.1.0.jar:8.1.0]
	at org.elasticsearch.cluster.service.ClusterApplierService.applyChanges(ClusterApplierService.java:510) [elasticsearch-8.1.0.jar:8.1.0]
	at org.elasticsearch.cluster.service.ClusterApplierService.runTask(ClusterApplierService.java:428) [elasticsearch-8.1.0.jar:8.1.0]
	at org.elasticsearch.cluster.service.ClusterApplierService$UpdateTask.run(ClusterApplierService.java:154) [elasticsearch-8.1.0.jar:8.1.0]
	at org.elasticsearch.common.util.concurrent.ThreadContext$ContextPreservingRunnable.run(ThreadContext.java:717) [elasticsearch-8.1.0.jar:8.1.0]
	at org.elasticsearch.common.util.concurrent.PrioritizedEsThreadPoolExecutor$TieBreakingPrioritizedRunnable.runAndClean(PrioritizedEsThreadPoolExecutor.java:260) [elasticsearch-8.1.0.jar:8.1.0]
[...]

PR #85632, opened 2022-04-01
Log at WARN level for Watcher cluster state validation errors

williamrandolph:
One of our users found that Watcher was not starting up correctly, but left no information in the log. We had to enable debug logging to see the log message that revealed the problem. This PR bumps that log line to WARN so that we can resolve similar future issues more quickly.

If I understand the code correctly, an error here will always prevent Watcher from starting, so I think it makes sense to log it at WARN level.

Relates https://github.com/elastic/elasticsearch/pull/85507

### DIFF
--- a/docs/changelog/85632.yaml
+++ b/docs/changelog/85632.yaml
@@ -1,0 +1,5 @@
+pr: 85632
+summary: Log at WARN level for Watcher cluster state validation errors
+area: Watcher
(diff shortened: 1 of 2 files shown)


masseyke (approved):
Looks good to me. I'm not sure why we were logging it at such a low level in the first place.

PR #84919, opened 2022-03-13
Block joins while applier is busy

DaveCTurner:
Certain cluster appliers perform I/O or other heavy computations and in
extreme circumstances may take a significant amount of time to apply a
cluster state. The master considers a node to be unhealthy if it takes
too long to apply a cluster state. Nodes that time out like this are
removed from the cluster, but they will carry on with the slow cluster
state application regardless. If the application takes long enough then
there may be time for the node to rejoin the cluster, time out applying
the state resulting from the rejoin, and be removed again. This can
happen repeatedly and is disruptive to the rest of the cluster.

There is no point in trying to rejoin the cluster while the applier is
busy. With this commit we send the join request from the applier thread,
ensuring that it is not occupied with other work and will therefore be
available to apply the joining state. The commit also adds periodic
logging of the status of ongoing joins to help clarify why it is taking
longer than expected to rejoin the cluster.

PR #85635, opened 2022-04-01
Report pending joins in ClusterFormationFailureHelper

DaveCTurner:
Extracted from #84919 for use by cluster health API


server/src/main/java/org/elasticsearch/cluster/coordination/JoinHelper.java
```diff
@@ -284,25 +284,29 @@ public void onResponse(Releasable connectionReference) {
...
 
                             @Override
                             public void handleException(TransportException exp) {
+                                pendingJoinInfo.message = PENDING_JOIN_FAILED;
                                 pendingOutgoingJoins.remove(dedupKey);
                                 FailedJoinAttempt attempt = new FailedJoinAttempt(destination, joinRequest, exp);
                                 attempt.logNow();
                                 lastFailedJoinAttempt.set(attempt);
                                 unregisterAndReleaseConnection(destination, connectionReference);
+                                pendingJoinInfo.message = PENDING_JOIN_FAILED;
```
andreidan, line 301: is this needed again?

andreidan (approved):
LGTM, thanks for adding this David

server/src/main/java/org/elasticsearch/cluster/coordination/ClusterFormationFailureHelper.java
```diff
@@ -249,5 +254,37 @@ private static String describeQuorum(VotingConfiguration votingConfiguration) {
...
             }
         }
+
+        private String getJoinStatusDescription() {
+            if (inFlightJoinStatuses.isEmpty()) {
+                return "";
+            }
+
+            final var stringBuilder = new StringBuilder();
+            inFlightJoinStatuses.stream()
```
arteam, line 264: Nit: I think it would be possible to collect results to the `stringBuilder` directly with the Streams API
```java
return inFlightJoinStatuses.stream()
                .sorted(Comparator.comparing(JoinStatus::age).reversed())
                .limit(10)
                .collect(
                    StringBuilder::new,
                    (sb, joinStatus) -> sb.append("; joining [")
                        .append(joinStatus.remoteNode().descriptionWithoutAttributes())
                        .append("] in term [")
                        .append(joinStatus.term())
                        .append("] has status [")
                        .append(joinStatus.message())
                        .append("] after [")
                        .append(timeValueWithMillis(joinStatus.age()))
                        .append("]"),
                    StringBuilder::append
                )
                .toString();
```

DaveCTurner: I think that's right, but I don't like this pattern because the execution order is poorly defined and doesn't exclude the possibility of doing extra unnecessary copying. It's cheap to append a few characters to the end of a `StringBuilder` but we never want to append a whole `StringBuilder`.

Not that it really matters here, there's only 10 items and this isn't exactly hot code, but still I don't like the pattern.

PR #77866, opened 2021-09-16
Speed up StreamInput#readString for Network Buffers

original-brownbear:
Mostly motivated by getting us some extra savings on the transport threads. 
In almost all cases we can just read the string straight off the buffer without
copying to the intermediary buffer used by `StreamInput` and make this read much
faster and more cache friendly.
Also, reduced the hot path size of some methods to make things compile nicer.

PR #85636, opened 2022-04-01
Extract some cold paths in StreamInput around String reading

original-brownbear:
Extracted the less controversial bits from #77866 here to make things inline
a little better. This is already quite helpful for some transport messages.


rjernst (approved):
LGTM

original-brownbear:
Thanks Ryan!

Issue #85542, opened 2022-03-31
GeoIP download will fail if airgapped environment only allows access to geoip.elastic.co


philippkahr:
### Elasticsearch Version

8.1.1

### Java Version

_bundled_

### OS Version

macOS 12.3

### Problem Description

When starting Elasticsearch it tries to download the GeoIP Package from geoip.elastic.co as documented [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/geoip-processor.html). However this is just a "redirect" to a google storage.

Thus in an airgapped environment one must allow access to geoip.elastic.co and the `storage.googleapis.com` which opens up a ton of possible security issues.

The geoip package should be served through the geoip.elastic.co URL.

```bash
curl -s 'https://geoip.elastic.co/v1/database?elastic_geoip_service_tos=agree' | jq
[
  {
    "age": 127526,
    "md5_hash": "e6bef7d0e11181b37f0288353f07179f",
    "name": "GeoLite2-ASN.tgz",
    "provider": "maxmind",
    "updated": 1648684830,
    "url": "https://storage.googleapis.com/elastic-paisano-production/maxmind/GeoLite2-ASN.tgz?X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=elastic-paisano-production%40elastic-apps-163815.iam.gserviceaccount.com%2F20220331%2Fhenk%2Fstorage%2Fgoog4_request&X-Goog-Date=20220331T000031Z&X-Goog-Expires=86400&X-Goog-SignedHeaders=host&X-Goog-Signature=3dcf41eae54a1be2cbe546e82489b91177e9f70b8338d98add6d5773ced2d8e10b11adcad03fca3acbc0842a60cc2dddc35955ef6d22388e92bcfa9cec10b002b2b597fff3ffd340748631dae4aedc81d379e59d1c27875fa65a4db0e432c9780b1b373253f31d2d6f7a8f586fa0c912ea3b397d6327ba4a50ba1c17a1b5a170ab5bfe79549ed88657780bfd0c810d57f616c21786f9f34382e94d560728b7d9f720b41fc98d960f83b64462c62d31eb5d4206be52c9c2524104f3ac8d12672efb88d1a3f1e93e4dcb892905fbf13820eb12c1a3bddaad29d992fed9d47630599e7d063dc3220494937fc10e85422a8295291ba83792e0b5effa072bf6e2ab0c"
  },
  {
    "age": 125938,
[...]

PR #85637, opened 2022-04-01
[DOC] Air gapped environments and GEOIP

lucabelluccini:
Closing https://github.com/elastic/elasticsearch/issues/85542

Please review/amend the text and formatting if necessary.

Also, backport to previous version (up to 7.14 I suppose).

### DIFF
--- a/docs/reference/ingest/processors/geoip.asciidoc
+++ b/docs/reference/ingest/processors/geoip.asciidoc
@@ -327,6 +327,12 @@ a secure proxy. You can then specify the proxy endpoint URL in the
 <<ingest-geoip-downloader-endpoint,`ingest.geoip.downloader.endpoint`>> setting
 of each node’s `elasticsearch.yml` file.
 
+[IMPORTANT]
+====
+In air gapped environments, the Elasticsearch nodes require access to `https://geoip.elastic.co`
+and `https://storage.googleapis.com/`.
+====
+
 [[use-custom-geoip-endpoint]]
 **Use a custom endpoint**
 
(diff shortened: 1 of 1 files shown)


dakrone (approved):
LGTM

dakrone (changes requested):
Actually, @lucabelluccini no PRs should be targeting the `main` branch directly yet. Could you change where this PR targets please?

lucabelluccini:
Sorry @dakrone should I target branch 8.2 and then we backport?

dakrone:
@lucabelluccini target `master` and then backport from there

lucabelluccini:
Should be targeting `master` now - I thought they were the same 🤦

wallrik:
Please do not put this in the documentation. I think there's a misunderstanding somewhere... You cannot access the internet across an air gap. **That's the whole point.** If you connect your network to the internet you no longer have an air gap.

If you want to put a similar note in the documentation it should probably say something like,

> If your cluster is behind a default denying firewall, or the software firewall on your nodes deny outbound traffic, the Elasticsearch nodes require access to `https://geoip.elastic.co` and `https://storage.googleapis.com`.

Please see the most well-established definition of an air gap in [RFC 4949](https://datatracker.ietf.org/doc/html/rfc4949#section-4). Also, see [Wikipedia](https://en.wikipedia.org/wiki/Air_gap_(networking)).

lucabelluccini:
I am not a technical writer and thank you @wallrik for the suggestion.

lockewritesdocs:
@lucabelluccini, should this PR also target previous versions back to 8.0? If so, I can add those version labels and also add the label to backport this PR automatically.

lucabelluccini:
I think we can backport in all versions back to 7.14

lockewritesdocs (approved):
LGTM 🦖 I added the appropriate version labels and the auto-backport label, so this is ready to merge.

PR #85639, opened 2022-04-01
Remove hppc from ClusterStatsNodes

rjernst:
The internals of ClusterStatsNodes keep a mapping from unique
jvm/architecture/etc to the count of nodes with a particular value.
These are small maps, and there is no need for them to use hppc classes.
This commit converts them to normal HashMaps.

relates #84735


grcevski (approved):
LGTM!

Issue #82595, opened 2022-01-14
`_recovery_source` sometimes remains after merge


jtibshirani:
If `_source` is disabled or filtered in the mappings, we add a `_recovery_source` field to support shard recoveries and CCR. Once it's no longer needed, then future merges will drop the `_recovery_source` field to reclaim space.

In certain cases, it appears that `_recovery_source` can stick around even after a merge. I noticed this issue through the dense vector rally track. This command indexes 100,000 documents with `_source` disabled, then force merges to 1 segment:
```
esrally race --track=dense_vector --challenge=index-and-search --track-params="ingest_percentage:10" --on-error abort
```

At the end, the shard was larger than expected:
```
195M	data/indices/gPefBjHjTCCxU_EnbSuGrQ/0/index
```

Using the disk usage API, we see this is due to recovery source:
```
   "_recovery_source" : {
        "total" : "149.9mb",
        "total_in_bytes" : 157209753,
        ....
```

There are no replicas, so the force merge should have removed recovery source. I can reproduce this with both 1 and 2 shards. I haven't found a small-scale reproduction yet.

ruslaniv:
I'm seeing the exact same behavior with:
```
"mappings": {
        "_source": {"excludes": ["title_vector"]},
        "properties": {
        ...}

``` 
Upon inspecting the index and based on this field size and amount of documents I indexed, it looks like the field stores vectors as plain floats without any compression.
FYI one 1024-dim vector stored as plain floats takes approximately 21-22 Kb whereas the same vector compressed and optimized by ES takes about 4 Kb. So that's quite a difference!

ruslaniv:
@elasticmachine is there any progress on this issue? 
Right now this problem is causing our index to grow to 250Gb instead of estimated 50Gb and the index no longer fits available RAM which severely degrades search performance.

DaveCTurner:
The `_recovery_source` field should be removed at merge time in all docs that are not retained for recovery by any retention lease. There is no real coordination between merges and retention lease movements so it's possible for documents including `_recovery_source` to end up in a large-ish segment that doesn't see another merge for a long time. But there shouldn't normally be very many documents like that, at least not unless there are some retention leases which lag a long way behind the max seqno for some reason.

You can get information about the retention leases and sequence numbers with `GET /<INDEX>/_stats?level=shards&filter_path=indices.*.shards.*.retention_leases,indices.*.shards.*.seq_no`. Can you share the output of that command here?

[...]

PR #85649, opened 2022-04-01
Synthetic source

nik9000:
image

This attempts to shrink the index by implementing a "synthetic _source" field.
You configure it by in the mapping:
```
{
  "mappings": {
    "_source": {
      "synthetic": true
    }
  }
}
```

And we just stop storing the `_source` field - kind of. When you go to access
the `_source` we regenerate it on the fly by loading doc values. Doc values
don't preserve the original structure of the source you sent so we have to
make some educated guesses. And we have a rule: the source we generate would
result in the same index if you sent it back to us. That way you can use it
for things like `_reindex`.

Fetching the `_source` from doc values does slow down loading somewhat. See
numbers further down.

## Supported fields
This only works for the following fields:
* `boolean`
* `byte`
* `date`
* `double`
* `float`
* `geo_point` (with precision loss)
* `half_float`
* `integer`
* `ip`
* `keyword`
* `long`
* `scaled_float`
* `short`
* `text` (when there is a `keyword` sub-field that is compatible with this feature)

## Educated guesses

The synthetic source generator makes `_source` fields that are:
* sorted alphabetically
* as "objecty" as possible
* pushes all arrays to the "leaf" fields
* sorts most array values
* removes duplicate text and keyword values

These are mostly artifacts of how doc values are stored.

### sorted alphabetically
```
{
  "b": 1,
  "c": 2,
  "a": 3
}
```
becomes
```
{
  "a": 3,
  "b": 1,
  "c": 2
}
```

### as "objecty" as possible
```
{
  "a.b": "foo"
}
```
becomes
```
{
  "a": {
    "b": "foo"
  }
}
```

### pushes all arrays to the "leaf" fields
```
{
  "a": [
    {
      "b": "foo",
      "c": "bar"
    },
    {
      "c": "bort"
    },
    {
      "b": "snort"
    }
}
```
becomes
```
{
  "a" {
    "b": ["foo", "snort"],
    "c": ["bar", "bort"]
  }
}
```

### sorts most array values
```
{
  "a": [2, 3, 1]
}
```
becomes
```
{
  "a": [1, 2, 3]
}
```

### removes duplicate text and keyword values
```
{
  "a": ["bar", "baz", "baz", "baz", "foo", "foo"]
}
```
becomes
```
{
  "a": ["bar", "baz", "foo"]
}
```
## `_recovery_source`

Elasticsearch's shard "recovery" process needs `_source` *sometimes*. So does
cross cluster replication. If you disable source or filter it somehow we store
a `_recovery_source` field for as long as the recovery process might need it.
When everything is running smoothly that's generally a few seconds or minutes.
Then the fields is removed on merge. This synthetic source feature continues
to produce `_recovery_source` and relies on it for recovery. It's *possible*
to synthesize `_source` during recovery but we don't do it.

That means that synethic source doesn't speed up writing the index. But in the
future we might be able to turn this on to trade writing less data at index
time for slower recovery and cross cluster replication. That's an area of
future improvement.

## perf numbers

I loaded the entire tsdb data set with this change and the size:

```
           standard -> synthetic
store size  31.0 GB ->  7.0 GB  (77.5% reduction)
_source  24695.7 MB -> 47.6 MB  (99.8% reduction - synthetic is in _recovery_source)
```

A second _forcemerge a few minutes after rally finishes should removes the
remaining 47.6MB of _recovery_source.

With this fetching source for 1,000 documents seems to take about 500ms. I
spot checked a lot of different areas and haven't seen any different hit. I
*expect* this performance impact is based on the number of doc values fields
in the index and how sparse they are.


nik9000:
I hacked together something to test the differences:
```
export from=$(curl -s -HContent-Type:application/json -uelastic:D2MVMBAUE0fUDu30A6yO -XPOST -k 'https://localhost:9201/tsdb/_search?size=0&pretty' -d'{
  "aggs": {
    "min": {
      "min": {
        "field": "@timestamp",
        "format": "epoch_millis"
      }
    }
  }
}' | jq -r .aggregations.min.value_as_string)
export to=$(curl -s -HContent-Type:application/json -uelastic:D2MVMBAUE0fUDu30A6yO -XPOST -k 'https://localhost:9201/tsdb/_search?size=0&pretty' -d'{
  "aggs": {
    "max": {
      "max": {
        "field": "@timestamp",
        "format": "epoch_millis"
      }
    }
  }
}' | jq -r .aggregations.max.value_as_string)
for date in $(seq $from 1000 $to); do
  for id in $(curl -s -HContent-Type:application/json -uelastic:D2MVMBAUE0fUDu30A6yO -XPOST -k 'https://localhost:9201/tsdb/_search?size=10000&pretty' -d'{
    "stored_fields": ["__none__"],
    "sort": {"@timestamp": "desc"},
    "query": {
      "range": {
        "@timestamp": {
          "gte": '$date',
          "format": "epoch_millis"
        }
      }
    }
  }' | jq -r .hits.hits[]._id); do
    echo $date $id
    diff \
      <(curl -s -HContent-Type:application/json -uelastic:D2MVMBAUE0fUDu30A6yO -XGET -k 'https://localhost:9201/tsdb/_doc/'$id | jq 'del(._seq_no)' -S) \
      <(curl -s -HContent-Type:application/json -uelastic:D2MVMBAUE0fUDu30A6yO -XGET -k 'https://localhost:9200/tsdb/_doc/'$id | jq 'del(._seq_no)' -S)
  done
done | tee diffs
```

Which spits out:
```
1619630303410 aJUt5LaG4q6Jz3sDAAABeR4M1KU
1619630303410 JFpDQXhLo-DNlumpAAABeR4M1KU
1619630303410 uwFiL6jzbCaBLd-LAAABeR4M1KU
1619630303410 DDQFZr8fpaviwfqcAAABeR4M1GI
36c36
<               "pct": 0.0076478614634146345
---
>               "pct": 0.008
40c40
<               "pct": 0.003919529
---
>               "pct": 0.004
72c72
<               "pct": 0.4777109375
---
>               "pct": 0.47800000000000004
75c75
<               "pct": 0.015884058091486793
---
>               "pct": 0.016
97c97
<         "start_time": "2021-04-29T08:18:44Z"
---
>         "start_time": "2021-04-29T08:18:44.000Z"
1619630303410 iHhCyXEez-elWsWXAAABeR4M1GI
36c36
<               "pct": 0.2534879995
---
>               "pct": 0.253
40c40
<               "pct": 0.2534879995
---
>               "pct": 0.253
72c72
<               "pct": 0.01896543080120626
---
>               "pct": 0.019
75c75
<               "pct": 0.01896543080120626
---
>               "pct": 0.019
97c97
<         "start_time": "2021-04-29T14:31:13Z"
---
>         "start_time": "2021-04-29T14:31:13.000Z"
1619630303410 wWEXn97ymH76klIbAAABeR4M1GI
36c36
<               "pct": 0.165416355
---
>               "pct": 0.165
40c40
<               "pct": 0.00827081775
---
>               "pct": 0.008
72c72
<               "pct": 0.619265625
---
>               "pct": 0.619
75c75
<               "pct": 0.010295400826531081
---
>               "pct": 0.01
97c97
<         "start_time": "2021-04-07T10:08:31Z"
---
>         "start_time": "2021-04-07T10:08:31.000Z"
```

The test data has values like `"pct": 0.010295400826531081` and has configured the mapping to use a scaled float with a scaling factor of `1000` so the what we actually store in doc values is `"pct": 0.01` - and that is what we put in synthetic source.

nik9000:
One thing I've noticed that we probably don't want but I don't know how to get rid of is `copy_to` - if you use `copy_to` to make an index of `foo.bar.message` at `message` then doc values for both will have doc values for it and the `_source` will contain it twice. This is ok for now, but would prevent us from using the source for recovery.

nik9000:
> One thing I've noticed that we probably don't want but I don't know how to get rid of is `copy_to` - if you use `copy_to` to make an index of `foo.bar.message` at `message` then doc values for both will have doc values for it and the `_source` will contain it twice. This is ok for now, but would prevent us from using the source for recovery.

I've forbidden copy_to for synthetic source indices in this PR. We can figure out how to allow it later.

nik9000:
I've modifies the description to change the enabling logic from

```
"_source": {
  "synthetic": true
}
```

to
```
"_source": {
  "enabled": "synthetic"
}
```

I know turning a boolean into a tri-valued thing is annoying, but it really is the most concise way to describe what's going on here - `_source` is enabled via synthetic reconstruction. And it's invalid to set `"enabled": false, "synthetic": true`.

I haven't yet implemented this change, but it's fairly small.

I also have more tests I need to add to this - round tripping tests. I haven't quite figured out how to do that yet though.

mushao999:
Hi, @nik9000 , it's really a great job you've done. I read the commits and have some questions:
1. Why not storing the non-doc-values fields still in the source and assembling them back together with doc-values fields while fetching? We can make our feature more perfect while not affect performance too much in this way.
2. Did your benchmark take the page-cache into consideration? In our case ,we clean the page-cache for every single query to get rid of the impact of cache.
3. You've used the new `SourceLoader` logic in `FetchPhase` and `ShardGetService`. I was wondering are there any other places that need to change to the new `SourceLoader` logic?
4. You changes separated the doc-values-fetching logic into every `FieldMapper`.It maybe a obstacle for concurrency if we want to improve the fetching performance.
Look forward to your reply. Thanks

nik9000:
> Hi, @nik9000 , it's really a great job you've done. I read the commits and have some questions:

Thanks! I'm kind of excited by it.
> 
>     1. Why not storing the non-doc-values fields still in the source and assembling them back together with doc-values fields while fetching? We can make our feature more perfect while not affect performance too much in this way.

Let me give an example here for those following along. Say you have a text field `text` and an integer `i`. And you send this `_source`:

```
{
  "text": "the quick brown fox",
  "i": 1
}
```

If the `text` field doesn't have any other place to save it's data then we'd *just* remove `i` from the `_source` and save the `_source` as:


```
{
  "text": "the quick brown fox",
}
```

The we'd apply the techniques from this PR on load to `i` and get the original `_source back.

It's a really good idea. One that @mushao999 and I have talked about in the past over video. But one I chose not to implement. I probably should have told him earlier that I wasn't planning on implementing it. That's my bad.

Anyway! I didn't implement that because I wanted to do the simplest thing that'd push performance as much as possible. And my ultimate goal is not to write `_source` at all when you are in this mode. Or to filter it. And I like that when you set `_source: {enabled: synthetic}` that behavior is forced. Sort of - see all my comments about `_recovery_source`.

But I don't think it's unreasonable to do the other thing too. I just don't really think I want to do that right now.

>     2. Did your benchmark take the page-cache into consideration? In our case ,we clean the page-cache for every single query to get rid of the impact of cache.

Good call. I tried thwart the page cache by scrolling the dates. That should effectively stop lots of caches, but the page cache, maybe not actually. I thought it had at the time, but I think I was wrong. If my page cache was hot before I started my benchmarks then it'd stay hot throughout them. And I had only loaded 7.0GB of data so it probably *was* hot. I'll try and rerun it.

> 
>     3. You've used the new `SourceLoader` logic in `FetchPhase` and `ShardGetService`. I was wondering are there any other places that need to change to the new `SourceLoader` logic?

There's really on more big spot: `LuceneChangesSnapshot`. That's part of replica recovery and cross cluster replication. But I'm able to ignore it for now by not touching the `_recovery_source` field. I'd *love* to remove that, but this PR is already plenty big. And, luckily, if I don't touch it everything "just works". And you get a lot of space savings! You just have to write `_recovery_source` to disk. And then remove it. Which feels like a waste. That's just so so so many bits to compress and write. But everything in time, I guess.

> 
>     4. You changes separated the doc-values-fetching logic into every `FieldMapper`.It maybe a obstacle for concurrency if we want to improve the fetching performance.

I don't think it will be. Or, at least, I'm hopeful it won't be. I was thinking that the next step would be to change the signature of:
```
public Leaf leaf(LeafReader reader) throws IOException
```
to:
```
public Leaf leaf(LeafReader reader, int ids) throws IOException
```

And then to have them load all the values "column-wise". I think that'd be a pretty fair performance improvement without any parallelization. After that I think we could maybe parallelize the calls to `leaf`, but I'm not sure precisely.

nik9000:
run elasticsearch-ci/part-2

mushao999:
@nik9000 Thanks for your detailed reply.
1. As to saving non-doc-values fields still in source, I think it has many advantages:
- can saving fields that does not save or support doc-values
- can avoid precision loss in some field such as geo_point
It's true we can force user to use it in our way, however it will limit the use of our feature in many cases, where user do not want to save all fields in doc-values and precision loss is unacceptable.
As to the implementation and performance, it will not introduce too many changes to make it work upon your current PR, it could be:
- indexing: filter source in `SourceFilterMapper`(10% tradeoff in our benchmark)
- fetching: load source into XContentBuilder beforce load doc-values in `FetchPhase`(time most cost in doc-values fetching, so this will not affect fetching performance too much)
I got the idea you may not want to do it right. But it's really necessary in our production. How about I open a PR to do it? 
2. Page cache thing:
in our case we clean page cache in this way:
- clean page cache using: `echo 3 > /proc/sys/vm/drop_caches`
- restart cluster

nik9000:
Oooh! I've realized that this doesn't plug into scripting - it won't synthesize the source for scripting. For now, at least, I think that's ok. It is a little janky though - because I believe it should work for fetch phase scripting but not, say, aggregation phase scripting. hmmmm.

nik9000:
Oooh! I've realized that this doesn't plug into scripting - it won't synthesize the source for scripting. For now, at least, I think that's ok. It is a little janky though - because I believe it should work for fetch phase scripting but not, say, aggregation phase scripting. hmmmm.

nik9000:
> 1. As to saving non-doc-values fields still in source, I think it has many advantages:
> 
> 
>     * can saving fields that does not save or support doc-values
> 
>     * can avoid precision loss in some field such as geo_point
> 
> 
> It's true we can force user to use it in our way, however it will limit the use of our feature in many cases, where user do not want to save all fields in doc-values and precision loss is unacceptable.
> 
> As to the implementation and performance, it will not introduce too many changes to make it work upon your current PR, it could be:
> 
>     * indexing: filter source in `SourceFilterMapper`(10% tradeoff in our benchmark)
> 
>     * fetching: load source into XContentBuilder beforce load doc-values in `FetchPhase`(time most cost in doc-values
>       fetching, so this will not affect fetching performance too much)

I think that's pretty reasonable though I wonder if it's right to do now. A lot of the folks I've talked to are really ok with the loss of precision. In fact, they'd kind of like a version of this that would prefer to drop fields rather than reject them. A "more sloppy" version of this where the `_source` is "best effort". I'm not super comfortable with it.

I wonder if it'd make sense in the end to have different ways of doing this. A "conservative" mode like your talking about, a "strict" mode like this, and a "best effort" mode like some other folks have been talking about. I like the characteristics of my strict mode a lot -namely that you *know* it'll never write any `_source`.



> 3\. Page cache thing:
>         in our case we clean page cache in this way:
> 
> 
>     * clean page cache using: `echo 3 > /proc/sys/vm/drop_caches`
> 
>     * restart cluster

I'll give this a try at some point.

test/framework/src/main/java/org/elasticsearch/index/mapper/MapperServiceTestCase.java
```diff
@@ -644,4 +646,36 @@ protected BiFunction<MappedFieldType, Supplier<SearchLookup>, IndexFieldData<?>>
...
+        try (Directory directory = newDirectory()) {
+            RandomIndexWriter iw = new RandomIndexWriter(random(), directory);
+            iw.addDocument(mapper.parse(source(build)).rootDoc());
+            iw.close();
+            try (DirectoryReader reader = DirectoryReader.open(directory)) {
+                SourceLoader loader = mapper.sourceMapper().newSourceLoader(mapper.mapping().getRoot());
+                String syntheticSource = loader.leaf(getOnlyLeafReader(reader)).source(null, 0).utf8ToString();
+                try (Directory roundTripDirectory = newDirectory()) {
+                    RandomIndexWriter roundTripIw = new RandomIndexWriter(random(), roundTripDirectory);
+                    roundTripIw.addDocument(mapper.parse(source(build)).rootDoc());
```
nik9000, line 660: This line is wrong. oops.

axw:
> I really like the above approach. @axw I could see this making also quite an impact for apm data.

Indeed. @dgieselaar spotted the PR and did some poking in our test clusters. In edge-oblt, he was seeing 35-50% savings (~35% for trace data, ~50% for metrics data). So, pretty significant.

> Source cleanup: I have to confess, I almost like the cleaned up source more than the source we have today. I like the alphabetical sorting, make everything an object etc. The only thing I stumbled over is the removal of duplicated array values but I could not find a use case where this would be relevant.

One place where this will be an issue for APM is how we're storing "span links" -- as an array of objects. As each object field is [stored by Elasticsearch as separate arrays](https://www.elastic.co/guide/en/elasticsearch/reference/current/nested.html#nested-arrays-flattening-objects), de-duplication and sorting would break our ability reconstruct the array of objects.

jsoriano:
> * The only thing I stumbled over is the removal of duplicated array values but I could not find a use case where this would be relevant. @jsoriano Anything comes to mind on your end?

In general we discourage the use of [arrays of objects](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/array.html), because they are not what people use to expect. We have some cases of the `nested` type, but I am not sure how these fields are consumed. I think these uses can be grouped in:
* Stack monitoring modules: I think nested type is used there to mimic the documents expected by the monitoring UI. @jasonrhodes may have more details on how needed this is in the future.
* A couple of security use cases, but all these cases are in logs data streams.

In other cases were we may have duplicated values in arrays I think that it is completely fine if they are removed.

ruflin:
The nice thing about the solution above is, that the indices which can't work with this features can opt-out. So I would not hold back on these edge cases.

romseygeek:
> as "objecty" as possible

I've just realised that this is going to clash with #86166, where if you bar subobjects on an object field you can only send documents with fields that look like `"a.b"`, and `"a":{"b":{...}}` will be rejected.  We can worry about it later, but it needs to be on our radar.

rockdaboot:
At continuous profiling we currently store stack traces as arrays of frames (basically two arrays that must stay in sync). So de-duplication and sorting forces us to store the array values as one object (as concatenated bytes, base64 encoded).
On the other hand side, the concatenation enables prefix compression that gives us another ~50% storage reduction with my test data. Means, we will do the concatenation anyway :-).

nik9000:
I was really hoping we'd have the time to build something special for
continuous profiling. A "stack trace" field or something. But there are
only so many hours. Sorry.

rockdaboot:
Just trying to test with ES built on top of `https://github.com/elastic/elasticsearch/pull/85649/commits/adfc9a390288485f908553f2ea3230bb57c7ce5c`.

Index creation fails with
```
    "_source": {
      "enabled": "synthetic"
    },
    ...
```
```
Failed to parse mapping: Failed to parse value [synthetic] as only [true] or [false] are allowed.
```
How do I enable this feature ?

nik9000:
> Index creation fails when putting this into the mapping

Sorry! I haven't yet switched to the tri-valued version of enabled - so it's `enabled: true, synthetic: true`. Sorry to mess that up for you!

Also! I just talked to @tvernum about field about field level security. It looks like it *should* just work but it sure does need a test.

jasonrhodes:
> We have some cases of the `nested` type, but I am not sure how these fields are consumed. I think these uses can be grouped in:
> 
> * Stack monitoring modules: I think nested type is used there to mimic the documents expected by the monitoring UI. @jasonrhodes may have more details on how needed this is in the future.

I think it was the logstash module only that uses nested field types and we've wanted to remove that for a long time to make that data available to Kibana more widely. I'll need to check to be sure.

nik9000:
> We have some cases of the `nested` type

@romseygeek and I were talking today about maybe supporting nested with this. It *should* be possible. But now is not the time to implement it. If you need it, we can get it in a follow up. Eventually. If no one needs it I'm happy to wait. Very happy to wait. Nested is complex internally.

ruflin:
>  If no one needs it I'm happy to wait.

Lets wait. It is also possible that we have it today but in reality we don't need it so this even forces us to get rid of it which is good.

romseygeek:
> I've just realised that this is going to clash with https://github.com/elastic/elasticsearch/pull/86166, where if you bar subobjects on an object field you can only send documents with fields that look like "a.b", and "a":{"b":{...}} will be rejected. We can worry about it later, but it needs to be on our radar.

Looking more closely at this, I don't think it's going to be a problem, because the way it is implemented is by delegating the source generation to object mappers, and an object mapper that disallows subobjects should just do the right thing here.

I'm pretty happy with the way this is implemented.  Let me know if I can help with more test writing or anything.

nik9000:
> I'm pretty happy with the way this is implemented. Let me know if I can help with more test writing or anything.

I'm presently fighting with round trip testing scaled floats. When you round trip a scaled float it produces values that *encode* to different numbers but when you decode them they are the same. I currently have no idea why. I'm checking the other round trip tests now to see if it's just scaled_float that's betraying me.

nik9000:
@romseygeek could you have another look at this? I've pushed some extra testing for round trips and it all passes. Well, sort of. I have to stub out a little of it because of mystery precision things. But I think we can get those in a follow up change.

x-pack/plugin/mapper-constant-keyword/src/internalClusterTest/java/org/elasticsearch/xpack/constantkeyword/mapper/ConstantKeywordFieldMapperTests.java
```diff
@@ -203,4 +206,24 @@ protected void randomFetchTestFieldConfig(XContentBuilder b) throws IOException
     protected boolean allowsNullValues() {
         return false;   // null is an error for constant keyword
     }
+
```
romseygeek, line 209: We have enough test cases that have to implement these four identical 'empty' methods, that maybe it's worth consolidating them into a `NoSyntheticSourceTest` interface with default methods and the test cases can just implement them?

nik9000:
For those following along at home this used to be activated with `enabled: synthetic` and but now it is activated with `synthetic: true`. I'm debating with a few folks about which is better. But, because this is behind a feature flag, I think it's safe to merge it either way. And, since the code currently supports `synthetic: true`, that's what I'd like to merge in the first cut.

nik9000:
I think we're getting closer to merging this. There are a *bunch* of follow ups:

### Before removing the feature flag
* [ ] Finish round trip tests
* [ ] Figure out how much performance we can get by turning off `_recovery_source`. Don't actually turn it off
* [ ] Figure out final API to turn this on

### Whenever
* [ ] Add support for more field types
* [ ] Speed up fetching by loading column-wise
* [ ] Synthesize instead of using `_recovery_source` if we find above that it can speed up indexing a lot.

romseygeek (approved):
LGTM.  Thanks for all the back and forth, let's get this merged and look at the follow-ups.

nik9000:
I have some perf numbers from a hack that turns off `_recovery_source`. This shows the potential indexing speed improvement we could get from using synthetic source on the recovery side:
```
|                 Min Throughput | 18024.6    | 19495.8     | 1471.23    | docs/s |  +8.16% |
|                Mean Throughput | 19426.9    | 21727.1     | 2300.2     | docs/s | +11.84% |
|              Median Throughput | 19169.7    | 21310.8     | 2141.08    | docs/s | +11.17% |
|                 Max Throughput | 22742.4    | 26960.5     | 4218.16    | docs/s | +18.55% |
|       Cumulative indexing time |   829.772  |   768.031   |  -61.741   |    min |  -7.44% |
|          Cumulative merge time |   235.641  |   230.36    |   -5.28152 |    min |  -2.24% |
| Cumulative merge throttle time |    39.8483 |    61.9718  |   22.1234  |    min | +55.52% |
|        Cumulative refresh time |    11.0899 |     7.02558 |   -4.06432 |    min | -36.65% |
|          Cumulative flush time |    41.3977 |    30.4483  |  -10.9495  |    min | -26.45% |
```

The short version is about 11% improvement in docs per second in TSDB, probably more in non-TSDB. Significantly faster merges, flushes, and refreshes - at least in TSDB, probably much faster in non-TSDB.

TSDB in it's current form has a somewhat inefficient indexing pipeline, mostly because it can never skip the `_id` lookup on write. We will fix that eventually, but for now TSDB is known slower to write. So the 11% speed boost on write here will likely jump once that slowness is resolved. I'm running a test against a non-TSDB index now to see.

The merge time is funny to read - it looks like a 2% speed up, but I believe a lot of that speed up is being throttled. See the 55% bump in merge throttling time. My guess is that we're looking at a reduction in load from merge in the 25% range, similar to flush and refresh.

Here's what the disk looks like with `_recovery_source` enabled:
```
Device   r/s     w/s  rMB/s     wMB/s  ... wareq-sz  svctm  %util
md0     0.00  176.67   0.00     30.34  ...   175.88   0.00   0.00
md0     0.00  157.33   0.00     31.66  ...   206.07   0.00   0.00
md0     0.00  196.67   0.00     35.70  ...   185.91   0.00   0.00
md0     0.00  288.00   0.00     62.00  ...   220.44   0.00   0.00
md0     0.00  185.00   0.00     41.12  ...   227.62   0.00   0.00
md0     0.00  126.33   0.00     27.97  ...   226.71   0.00   0.00
md0     0.00  192.00   0.00     25.95  ...   138.38   0.00   0.00
md0     0.00  208.64   0.00     47.55  ...   233.36   0.00   0.00
md0     0.00 1167.00   0.00    206.48  ...   181.18   0.00   0.00
md0     0.00  206.64   0.00     23.55  ...   116.69   0.00   0.00
md0     0.00  221.67   0.00     30.54  ...   141.06   0.00   0.00
md0     0.00  158.33   0.00     29.90  ...   193.36   0.00   0.00
md0     0.00  208.00   0.00     33.23  ...   163.59   0.00   0.00
md0     0.00  266.33   0.00     70.11  ...   269.54   0.00   0.00
md0     0.00  122.67   0.00     12.91  ...   107.77   0.00   0.00
md0     0.00  184.67   0.00     28.96  ...   160.57   0.00   0.00
md0     0.00  951.67   0.00    103.28  ...   111.13   0.00   0.00
md0     0.00  214.00   0.00     31.92  ...   152.72   0.00   0.00
md0     0.00  184.00   0.00     31.07  ...   172.93   0.00   0.00
```

Note the bursty writes. Here's what it looks like without `_recovery_source`:

```
Device   r/s     w/s  rMB/s     wMB/s   ... wareq-sz  svctm  %util
md0     0.00  252.00   0.00     43.25   ...   175.74   0.00   0.00
md0     0.00  703.00   0.00     51.49   ...    75.00   0.00   0.00
md0     0.00  250.00   0.00     41.11   ...   168.40   0.00   0.00
md0     0.00  194.67   0.00     44.45   ...   233.82   0.00   0.00
md0     0.00  192.00   0.00     44.57   ...   237.71   0.00   0.00
md0     0.00  176.00   0.00     41.98   ...   244.23   0.00   0.00
md0     0.00  157.67   0.00     26.27   ...   170.62   0.00   0.00
md0     0.00  854.67   0.00     77.48   ...    92.83   0.00   0.00
md0     0.00  174.67   0.00     38.78   ...   227.35   0.00   0.00
md0     0.00  186.67   0.00     40.56   ...   222.51   0.00   0.00
md0     0.00  174.67   0.00     36.90   ...   216.34   0.00   0.00
md0     0.00  219.67   0.00     44.61   ...   207.96   0.00   0.00
md0     0.00  187.67   0.00     41.46   ...   226.21   0.00   0.00
md0     0.00   79.33   0.00     13.53   ...   174.67   0.00   0.00
md0     0.00  670.33   0.00     71.00   ...   108.46   0.00   0.00
md0     0.00  307.67   0.00     33.58   ...   111.75   0.00   0.00
md0     0.00  182.00   0.00     44.03   ...   247.74   0.00   0.00
md0     0.00  219.00   0.00     46.32   ...   216.57   0.00   0.00
md0     0.00  210.67   0.00     41.19   ...   200.21   0.00   0.00
md0     0.00  204.00   0.00     48.01   ...   241.01   0.00   0.00
md0     0.00  101.33   0.00     18.07   ...   182.63   0.00   0.00
md0     0.00  773.33   0.00    105.11   ...   139.18   0.00   0.00
md0     0.00  287.67   0.00     28.89   ...   102.82   0.00   0.00
md0     0.00  176.00   0.00     40.39   ...   234.99   0.00   0.00
md0     0.00  209.67   0.00     40.71   ...   198.82   0.00   0.00
md0     0.00  209.67   0.00     41.95   ...   204.86   0.00   0.00
md0     0.00  145.67   0.00     25.39   ...   178.52   0.00   0.00
md0     0.00  203.33   0.00     38.60   ...   194.41   0.00   0.00
md0     0.00  737.00   0.00     66.14   ...    91.90   0.00   0.00
```

The writes are *less* bursty. Still bursty, but less so. I believe the infrastructure that I used to run this captured graphs of the this data over a longer period of time, but I don't know how to access it. I'm digging.

nik9000:
I got charts! Here's disk write for non-tsdb indices:
image

It's a stacked line graph writes on all physical disks on the machine. So md0 above is basically the top most line. The first run has `_recovery_source` and the second one doesn't. It writes faster and hits the disk less hard.

nik9000:
Here's the TSDB run:
image

This time the *second* run has `_recovery_source` and the first one doesn't. It's the same picture thing - turning off `_recovery_source` hits the disk less hard and increases write speed.

ruslaniv:
Is there any way to disable creation of `_recovery_source` because of this: 
https://github.com/elastic/elasticsearch/issues/82595#issuecomment-1331773350

nik9000:
We've talked a little about this - rebuilding the _source on the fly using synthetic _source. At the time we decided it wasn't worth it because folks were looking at doing other kinds of replication. I believe they are still working on that. In that replication mechanism we wouldn't need `_recovery_source` at all. That'd be lovely. No synthetic _source required. I still think that's a good plan.

ruslaniv:
Nik, thank you for your answer! 
Do you think the issue of "dangling" `_recovery_source` could be addressed in the near future? Right now this issue is causing our index to grow to 250Gb instead of 50Gb. Not only this is wasting 200Gb of disk space which is not critical, but  the index no longer fits available RAM which is very critical.

nik9000:
> Not only this is wasting 200Gb of disk space which is not critical, but the index no longer fits available RAM which is very critical.

Bleh. And `_source` is stored next to `_id` and friends so you'll end up paging it in even if you weren't intending to load it from disk. Lovely. It looks like @DaveCTurner is talking to you on the linked issue about the dangling `_recovery_source` which is a good sign. He should be able to figure out what's going on for you.

I do think `_recovery_source` is being used much more now - mostly because folks are removing dense vectors from the _source, but partly because of synthetic _source. I wouldn't be surprised if we found more "fun" things in it now - but it should work as he describes. I read a lot of that code when working on this. But computers are sneaky.

PR #85659, opened 2022-04-03
Cleanup XContentParserUtils throw on method signatures

original-brownbear:
No need to make all callers pass in the the location object here
as all call sites take the current location of the parser anyway.


rjernst:
Is there a technical reason for this change? To me the existing signature ensures better isolation of concern, since the throw method doesn’t need the entire parser.

original-brownbear:
> Is there a technical reason for this change?

It's a small thing but it does reduce the method size after JIT a little. Once dead code elimination does away with the throwing path (the tiny method should get inlined right away at the start, we save a method call. It's a similar optimization to how `ensureExpectedToken` works above.

rjernst (approved):
LGTM. We could consider a super interface in the future to narrow it just to the get location method.

PR #85660, opened 2022-04-03
Speed up Strings.validFileName

original-brownbear:
This method takes up ~10-15% of the CPU of updating an instance of `BlobStoreIndexShardSnapshots`
at the moment. This does matter for large repositories where there's a lot of shards to update, particularly
during deletes which all run on the same node.
Boxing and comparing chars via the set does not appear to get optimized by the compiler in any way
while the new method inlines fully and is essentially free compared to the rest of the logic in the caller methods.

### DIFF
--- a/server/src/main/java/org/elasticsearch/common/Strings.java
+++ b/server/src/main/java/org/elasticsearch/common/Strings.java
@@ -32,6 +32,8 @@ import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.TreeSet;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class Strings {
 
@@ -351,7 +353,9 @@ public class Strings {
         return sb.toString();
     }
 
-    public static final Set<Character> INVALID_FILENAME_CHARS = Set.of('\\', '/', '*', '?', '"', '<', '>', '|', ' ', ',');
+    public static final String INVALID_FILENAME_CHARS = "["
+        + Stream.of('\\', '/', '*', '?', '"', '<', '>', '|', ' ', ',').map(c -> "'" + c + "'").collect(Collectors.joining(","))
+        + "]";
 
     public static boolean validFileName(String fileName) {
         for (int i = 0; i < fileName.length(); i++) {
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/elasticsearch/common/Strings.java
```diff
@@ -355,8 +355,7 @@ private static String changeFirstCharacterCase(String str, boolean capitalize) {
 
     public static boolean validFileName(String fileName) {
         for (int i = 0; i < fileName.length(); i++) {
-            char c = fileName.charAt(i);
-            if (INVALID_FILENAME_CHARS.contains(c)) {
```
rjernst, line 359: We should remove the Set? That can be a followup, but keeping it around is confusing.

server/src/main/java/org/elasticsearch/common/Strings.java
```diff
@@ -366,13 +365,22 @@ public static boolean validFileName(String fileName) {
...
-            if (c != '*' && INVALID_FILENAME_CHARS.contains(c)) {
+            if (c != '*' && isInvalidFileNameCharacter(c)) {
                 return false;
             }
         }
         return true;
     }
 
+    private static boolean isInvalidFileNameCharacter(char c) {
+        switch (c) {
```
rjernst, line 376: I think this method can be written a little more succinctly with a switch expression:

```
return switch(c) {
            case '\\', '/', '*', '?', '"', '<', '>', '|', ' ', ',' -> true;
            default -> false;
        };
```

rjernst (approved):
LGTM, one suggestion and one followup ask.

server/src/main/java/org/elasticsearch/common/Strings.java
original-brownbear: ++ done

server/src/main/java/org/elasticsearch/common/Strings.java
original-brownbear: Did it here in https://github.com/elastic/elasticsearch/pull/85660/commits/421b5c60fbc6bd4ede3bbf4ee8bf08a46313a2cf the current set is only used to render an exception message in production and it rendered in a broken way (just inlined the char set in the one test using it, which seemed good enough to me).

original-brownbear:
Thanks Ryan!

PR #85661, opened 2022-04-03
Cleanup and slight speedup for field name parsing in snapshot metadata

original-brownbear:
No need to use parse fields here, there's' no deprecated names or similar.
Also, cleanup the complicated if clauses in the file info parser.


idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

PR #85672, opened 2022-04-04
[ML] Allow retrieving `boolean` fields from _source in DFA jobs

dimitris-athanasiou:
This commit fixes a bug where `boolean` mapped fields may only
be retrieved from doc values in Data Frame Analytics jobs.

While this problem may occur for `boolean` fields with disabled
doc values, a more common way of being affected by this bug is
running an analysis against data with a very high number of fields.
If there are more fields than the `index.max_docvalue_fields_search`
setting, we fall back to retrieving the fields from _source. Then
`boolean` mapped fields would fail with the error message "cannot
apply boolean mapping to field [<field>]".

### DIFF
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/extractor/ExtractedFields.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/extractor/ExtractedFields.java
@@ -8,6 +8,7 @@ package org.elasticsearch.xpack.ml.extractor;
 
 import org.elasticsearch.action.fieldcaps.FieldCapabilities;
 import org.elasticsearch.action.fieldcaps.FieldCapabilitiesResponse;
+import org.elasticsearch.core.Booleans;
 import org.elasticsearch.index.mapper.BooleanFieldMapper;
 import org.elasticsearch.search.SearchHit;
 import org.elasticsearch.xpack.core.ml.utils.MlStrings;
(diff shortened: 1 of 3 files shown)


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/extractor/ExtractedFields.java
```diff
@@ -235,39 +235,62 @@ private static boolean isNestedOrObject(Map<String, FieldCapabilities> fieldCaps
...
+
         @Override
         public Object[] value(SearchHit hit) {
-            DocumentField keyValue = hit.field(getName());
-            if (keyValue != null) {
-                return keyValue.getValues().stream().map(v -> Boolean.TRUE.equals(v) ? trueValue : falseValue).toArray();
+            Object[] value = field.value(hit);
+            if (value != null) {
+                return Arrays.stream(value).map(v -> Boolean.TRUE.equals(v) ? trueValue : falseValue).toArray();
             }
```
benwtrent, line 266: I am not 100% sure this works, in `_source` the boolean representation could be `"true"`.

dimitris-athanasiou: Well spotted! I'm now using `Booleans.parseBoolean` which is also what's used by the `BooleanFieldMapper` when the values are strings.

dimitris-athanasiou: Somehow I forgot to push that change. It should be available now.

benwtrent (approved):

PR #85677, opened 2022-04-04
Use the WEIGHT_MATCHES implementation of the unified highlighter

romseygeek:
Cut over to using WEIGHT_MATCHES by default; this should improve matching
for custom queries and complex booleans.

Note that with this change, phrases are highlighted as a whole rather than as
a set of individual terms.


server/src/test/java/org/elasticsearch/lucene/search/uhighlight/CustomUnifiedHighlighterTests.java
```diff
@@ -187,7 +187,7 @@ public void testMultiPhrasePrefixQuerySingleTerm() throws Exception {
 
     public void testMultiPhrasePrefixQuery() throws Exception {
         final String[] inputs = { "The quick brown fox." };
-        final String[] outputs = { "The <b>quick</b> <b>brown</b> <b>fox</b>." };
+        final String[] outputs = { "The <b>quick brown fox</b>." };
```
mayya-sharipova, line 190: Does this output change imply that we are dealing with breaking change?

server/src/main/java/org/elasticsearch/common/lucene/search/MultiPhrasePrefixQuery.java
```diff
@@ -299,6 +302,15 @@ private static boolean termArraysEquals(List<Term[]> termArrays1, List<Term[]> t
...
     public void visit(QueryVisitor visitor) {
         if (visitor.acceptField(field)) {
+            for (int i = 0; i < termArrays.size() - 1; i++) {
+                visitor.consumeTerms(this, termArrays.get(i));
+            }
+            for (Term term : termArrays.get(termArrays.size() - 1)) {
+                visitor.consumeTermsMatching(this, field, () -> {
+                    Automaton a = PrefixQuery.toAutomaton(term.bytes());
+                    return new CompiledAutomaton(a).runAutomaton;
+                });
```
mayya-sharipova, line 312: I wonder if this automaton will respect `maxExpansions` parameter?  Or since we expect mostly use highlighter based on matches this is not important?

server/src/main/java/org/elasticsearch/common/lucene/search/MultiPhrasePrefixQuery.java
```diff
@@ -299,6 +302,15 @@ private static boolean termArraysEquals(List<Term[]> termArrays1, List<Term[]> t
...
+            }
```
mayya-sharipova, line 313: Should we remove "// TODO implement term visiting" on the next line?

mayya-sharipova (approved):
This change overall LGTM after we fix all test failures; I've left a couple of questions to clarify.

romseygeek:
`WEIGHT_MATCHES` won't work for runtime fields either, so I think we will have to fall back to the default impl if the field to be highlighted is a runtime one?

jimczi:
Superseded by #96068

PR #85684, opened 2022-04-04
Make FieldMapper.parse inline better and slightly faster

original-brownbear:
`MultiFields` needs no map, we only ever iterate the values so we can
use an array to get no-overhead iteration (the overhead of setting
up the iterator for single entry multi-fields is considerable especially
with the unmodifiable map wrapping). This also speeds up mapper
serialization.
Also, extract cold exception throwing path which makes this method go from 100b+
to 39b and inline in many places and extract multi-field iteration which
isn't always used for the same reason (though it's a lesser concern).

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
@@ -12,6 +12,7 @@ import org.apache.lucene.index.LeafReaderContext;
 import org.elasticsearch.Version;
 import org.elasticsearch.common.Explicit;
 import org.elasticsearch.common.TriFunction;
+import org.elasticsearch.common.collect.Iterators;
 import org.elasticsearch.common.logging.DeprecationCategory;
 import org.elasticsearch.common.logging.DeprecationLogger;
 import org.elasticsearch.common.settings.Setting;
@@ -30,6 +31,7 @@ import org.elasticsearch.xcontent.support.AbstractXContentParser;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.EnumSet;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
```diff
@@ -495,44 +509,43 @@ public MultiFields build(Mapper.Builder mainFieldBuilder, MapperBuilderContext c
...
         }
 
-        private final Map<String, FieldMapper> mappers;
+        private final FieldMapper[] mappers;
 
         private MultiFields(Map<String, FieldMapper> mappers) {
-            this.mappers = mappers;
+            this.mappers = mappers.values().toArray(EMPTY_MAPPERS);
+            // sort for consistent iteration order + serialization
+            Arrays.sort(this.mappers, Comparator.comparing(FieldMapper::name));
```
idegtiarenko, line 522: It looks like there are only 2 usages of this constructor: empty case and a builder one.
Could we change constructor to accept a [sorted] array of field mappers already?
It looks like builder create the maps specifically for this constructor and it could be changed to array/list/stream instead

server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
```diff
@@ -222,39 +224,60 @@ public boolean parsesArrayValue() {
     public void parse(DocumentParserContext context) throws IOException {
         try {
             if (hasScript) {
-                throw new IllegalArgumentException("Cannot index data directly into a field with a [script] parameter");
+                throwIndexingWithScriptParam();
```
idegtiarenko, line 227: NIT: I would not extract this unless absolutely necessary. Not sure if this is saving a lot method size wise...

server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
```diff
@@ -222,39 +224,60 @@ public boolean parsesArrayValue() {
...
-                );
-            }
+            rethrowAsMapperParsingException(context, e);
+        }
+        // TODO: multi fields are really just copy fields, we just need to expose "sub fields" or something that can be part
+        // of the mappings
+        if (multiFields.mappers.length == 0) {
+            return;
+        }
+        doParseMultiFields(context);
```
idegtiarenko, line 238: Looks like an inlined version of parse?
```suggestion
        if (multiFields.mappers.length > 0) {
            doParseMultiFields(context);
        }
```

idegtiarenko (approved):

server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
original-brownbear: Right ... that makes perfect sense and is yet another neat speedup to constructing these (which is nice for many-shards benchmarks). We only really (sort of) need the map in the builder but not in the constructor.
I made it use an array now for the constructor argument but kept the sorting in the constructor to make things easier to read (otherwise it's yet another mental step to figuring out that the array is sorted).
Let me know what you think :)

original-brownbear:
Thanks Ievgen!

PR #85688, opened 2022-04-05
Speed up NumberFieldMapper

original-brownbear:
No need to create an intermediary list here. Creating it and adding it
to the document tended to take more time than the parsing of the number itself.

From our benchmarking during on-week:

before:

image

after:

image

As part of another relatively complicated mapping the number field mappers see a speedup of ~3x with this change.
This makes good sense given how the previous implementation would needlessly create array lists with a capacity of 10 (default) to index one to three fields.


server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
```diff
@@ -349,18 +349,24 @@ public Query rangeQuery(
             }
 
             @Override
-            public List<Field> createFields(String name, Number value, boolean indexed, boolean docValued, boolean stored) {
-                List<Field> fields = new ArrayList<>();
+            public void createFields(
```
idegtiarenko, line 352: NIT: `appendFields`

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
```diff
@@ -349,18 +349,24 @@ public Query rangeQuery(
...
+                Consumer<IndexableField> consumer,
```
idegtiarenko, line 353: I wonder if it is better to supply list at first place to avoid creating consumer from `List#add()`?

original-brownbear: Refactored to using `LuceneDocument` directly here without any indirection now :)

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
```diff
@@ -370,9 +374,13 @@ public IndexFieldData.Builder getFieldDataBuilder(String name) {
 
             private static void validateParsed(float value) {
                 if (Float.isFinite(HalfFloatPoint.sortableShortToHalfFloat(HalfFloatPoint.halfFloatToSortableShort(value))) == false) {
-                    throw new IllegalArgumentException("[half_float] supports only finite values, but got [" + value + "]");
+                    throwOnInfiniteHalfFloat(value);
```
henningandersen, line 377: Is this important to the optimization? I think I'd prefer to leave this as is unless there is a measurable effect?

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
```diff
@@ -496,9 +510,13 @@ public IndexFieldData.Builder getFieldDataBuilder(String name) {
 
             private static void validateParsed(float value) {
                 if (Float.isFinite(value) == false) {
-                    throw new IllegalArgumentException("[float] supports only finite values, but got [" + value + "]");
+                    throwOnInfiniteFloat(value);
```
henningandersen, line 513: Same as previous comment, is this bit important?

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
```diff
@@ -349,18 +347,24 @@ public Query rangeQuery(
             }
 
             @Override
-            public List<Field> createFields(String name, Number value, boolean indexed, boolean docValued, boolean stored) {
-                List<Field> fields = new ArrayList<>();
+            public void createFields(
+                LuceneDocument document,
```
henningandersen, line 351: I would prefer to pass in an explicit interface or alternatively `Consumer<Field>`.

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
```diff
@@ -1449,18 +1510,22 @@ protected void parseCreateField(DocumentParserContext context) throws IOExceptio
         try {
             value = value(context.parser(), type, nullValue, coerce());
         } catch (IllegalArgumentException e) {
-            if (ignoreMalformed.value() && context.parser().currentToken().isValue()) {
-                context.addIgnoredField(mappedFieldType.name());
-                return;
-            } else {
-                throw e;
-            }
+            handleIllegalArgumentException(context, e);
```
henningandersen, line 1513: Is this important to the optimization? The exception flow reads less naturally with this extracted.

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
```diff
@@ -1032,7 +1086,14 @@ public abstract Query rangeQuery(
 
         public abstract Number parsePoint(byte[] value);
 
-        public abstract List<Field> createFields(String name, Number value, boolean indexed, boolean docValued, boolean stored);
+        public abstract void createFields(
```
henningandersen, line 1089: I wonder if this should be renamed to `addFields` now that it no longer just instantiates the objects?

Also, would be nice to add javadoc on the expected behavior.

henningandersen (commented):
Removing the additional list makes sense to me. I wonder if you have a rally comparison showing before/after so that we can see how much it means for an indexing workload?

The other parts (extracting throw methods) seems more like micro-optimizations. I think we need more thorough input on why this is important before we make those in general. And if we find we need them, we also need a way to capture changes to the code that adds such throws in the future, otherwise we risk regressing performance. Doe they have visible effects on a rally run?

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
original-brownbear: Hard to say. It certainly gives the compiler a better shot at generating good code than not doing this.

This is what current master looks like when profiling:

image

Obviously, fixing the intermediary list is the biggest saving we can make but it's fairly visible that just the pure byte parsing side of things isn't free either CPU wise.
You can pretty easily make this a visible effect in a JMH benchmark. Whether or not this will translate into a real-world effect will very much depend on the use case (mapping) and environment.
It's rather hard to make this kind of thing show in a rally run because of how noisy they are.

That said, my thinking as far whether it's worth it optimising this kind of code path to this degree would be:

We spend roughly half the CPU time on the write thread during indexing on parsing documents in benchmarks like the Rally logging track. There's not that all that many obvious optimisations in this code (the list thing here being a notable exception) but it's slower in terms of bytes/s than it could theoretically be. In the past optimisations like this around x-content have been quite successful and Jackson itself goes to great lengths to do the same optimisations on their end.
Each individual change like this will not necessarily show up much outside of JMH but the aggregate effects of these changes will eventually by allowing the compiler to optimise with greater scope after inlining more.
-> while admittedly there's always a little voodoo to this kind of optimisation and a single digit % change in the performance of this code is hard to measure in Rally, I'd still go for it simply because this is among the hottest code paths we have and any bit of saving here will translate into real world cost savings for users now and might reveal optimisations to us (and/or the compiler :)) down the line.

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
original-brownbear: See https://github.com/elastic/elasticsearch/pull/85688#discussion_r857494685

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
original-brownbear: also https://github.com/elastic/elasticsearch/pull/85688#discussion_r857494685 :)

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
original-brownbear: ++ Renamed and added javadoc.

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
original-brownbear: I thought the same initially https://github.com/elastic/elasticsearch/pull/85688#pullrequestreview-931539042 :) But I think I agree with Ievgen and prefer the current solution now. What's the point of adding that kind of indirection here? It may look a little nicer in isolation, but makes the big picture of what this method does less obvious (which makes reasoning about performance harder and invites problems like the intermediary list we remove here). Also, while the compiler can mostly figure this kind of thing out and make it fully inline, I'd rather not push luck on that front either and rather make the performance obvious.

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
henningandersen: I think I'd like to leave out these minor optimizations from this PR and then have a separate discussion on those. I think we need a firmer stance on the necessity of this and if we end up needing this, we should have some verification that we are not inadvertently violating this in our tests or build. We should probably discuss that in a meeting in the coming weeks. But I think it is worthwhile getting the main optimization (the intermediary list optimization) in, so would like to decouple the two.

original-brownbear: Sounds good, reverted the JIT optimizations for now then :)

idegtiarenko (approved):

javanna:
I had another look and I like the progress made, thanks @henningandersen for reviewing this! I think the PR is now focused on a single improvement, which is good. Do I understand correctly that it's all about avoiding the indirection of creating a list of fields and rather adding those fields directly to the LuceneDocument? If so could you update the description of the PR to reflect that? 

Have you run the benchmarks again after updating the PR to only address this aspect?

original-brownbear:
Thanks for taking a look @javanna !

> Do I understand correctly that it's all about avoiding the indirection of creating a list of fields and rather adding those fields directly to the LuceneDocument? If so could you update the description of the PR to reflect that?

Jup, this is the only change now. Updated the description accordingly.

> Have you run the benchmarks again after updating the PR to only address this aspect?

Just ran it again, still exactly the same picture as in the initial description and still about a 3x speedup in the context of `org.elasticsearch.benchmark.index.mapper.KeywordFieldMapperBenchmark` for example (where we're just doing a standard long field).

henningandersen:
@original-brownbear can you also share the rally comparison so we can see what size of benefit the change provides from the outside?

henningandersen (approved):
LGTM. I'd like someone from the search team to also approve this before merging though.

javanna (approved):
LGTM

original-brownbear:
Thanks all!

PR #85704, opened 2022-04-05
reuse the delegate

idegtiarenko:
This should reuse the result of a `delegate()` call that is backed by a `ArrayDeque#peek()` operation.

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/mapper/DotExpandingXContentParser.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/DotExpandingXContentParser.java
@@ -74,14 +74,15 @@ class DotExpandingXContentParser extends FilterXContentParserWrapper {
             if (subpaths.length == 1 && field.endsWith(".") == false) {
                 return;
             }
-            XContentLocation location = delegate.getTokenLocation();
             Token token = delegate.nextToken();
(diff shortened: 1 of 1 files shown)


original-brownbear (approved):

server/src/main/java/org/elasticsearch/index/mapper/DotExpandingXContentParser.java
```diff
@@ -73,14 +74,15 @@ private void expandDots() throws IOException {
...
+            Token token = delegate.nextToken();
+            if (token == Token.END_OBJECT || token == Token.END_ARRAY) {
                 throw new IllegalStateException("Expecting START_OBJECT or START_ARRAY or VALUE but got [" + token + "]");
             } else {
-                parsers.push(new DotExpandingXContentParser(new SingletonValueXContentParser(delegate()), delegate(), subpaths, location));
+                var location = delegate.getTokenLocation();
+                var parser = (token == Token.START_OBJECT || token == Token.START_ARRAY)
+                    ? new XContentSubParser(delegate)
+                    : new SingletonValueXContentParser(delegate);
+                parsers.push(new DotExpandingXContentParser(parser, delegate, subpaths, location));
```
idegtiarenko, line 85: This caused the test failure at: `org.elasticsearch.index.mapper.DotExpandingXContentParserTests#testGetTokenLocation`

PR #85705, opened 2022-04-05
Adding support for filtering pit requests by index and shard

### LINKED ISSUE #85703: Support specifying index and shard in point in time request
### Description

While looking into replacing scrolls with PITs in https://github.com/elastic/elasticsearch-hadoop/issues/1925, I noticed that PITs are missing two features that scrolls have: (1) the ability to ask for results only in certain indices and (2) the ability to ask for results only in certain shards. I'm not sure if this is needed for many other use cases, but it is critical for es-hadoop/es-spark because we create one task per shard.
I think this is pretty straightforward to do, and I'll open a draft PR for it.

masseyke:
This makes it so that you can have a PIT query only search given indices and shards, bringing it to parity with what scrolls support.
I'm marking this as draft because I'm not familiar with this code, and not sure if the search team even wants this feature.
Closes #85703
Relates https://github.com/elastic/elasticsearch-hadoop/issues/1925

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/search/TransportSearchAction.java
+++ b/server/src/main/java/org/elasticsearch/action/search/TransportSearchAction.java
@@ -28,6 +28,7 @@ import org.elasticsearch.cluster.node.DiscoveryNode;
 import org.elasticsearch.cluster.node.DiscoveryNodes;
 import org.elasticsearch.cluster.routing.GroupShardsIterator;
 import org.elasticsearch.cluster.routing.OperationRouting;
+import org.elasticsearch.cluster.routing.Preference;
 import org.elasticsearch.cluster.routing.ShardIterator;
 import org.elasticsearch.cluster.routing.ShardRouting;
 import org.elasticsearch.cluster.service.ClusterService;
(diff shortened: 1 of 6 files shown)


masseyke:
This PR is currently failing to pass all tests because data streams don't work the way I've done it. Adding data stream support to the REST path seems straightforward enough, but EQL is using a different path than I've been testing with (TransportOpenPointInTimeAction), and that seems a lot harder to fix.

server/src/main/java/org/elasticsearch/action/search/TransportSearchAction.java
```diff
@@ -1433,11 +1445,59 @@ private static RemoteTransportException wrapRemoteClusterFailure(String clusterA
...
         }
         return iterators;
     }
+
+    private static boolean passesIndexCheck(String indexNameFromPit, Set<String> indexNamesFromRequest) {
+        if (indexNamesFromRequest.contains(indexNameFromPit)) {
+            return true;
+        }
+        for (String indexNameFromRequest : indexNamesFromRequest) {
+            if (indexNameFromPit.startsWith(".ds-" + indexNameFromRequest)) {
```
masseyke, line 1476: We don't want to merge this. This is just to get everything working with data streams.

masseyke:
This PR works well for testing es-hadoop use of PITs, but I'm going to close it because it's not really close to ready. Here's my known to do list (there might be more)

1. Properly support datastreams (as opposed to the hack I did in TransportSearchAction).
2. Add support for all `preference` values (currently only supports `_shards`).
3. Fix all the tests that are currently braking.

Issue #74660, opened 2021-06-28
Add better support for metric data types (TSDB)


imotov:
#### Phase 0 - Inception
- [x] Obtain schemas annotated with dimensions and metrics from the Metrics team (small) @nik9000 
- [x] Prototyping Lucene Data Pull Mechanism(medium) @imotov
- [x] Prototyping Data Pull Mechanism in elasticsearch @imotov

#### Phase 1 - Mappings
- [x] Add `time_series_dimension` mapping parameter to fields 
  - [X] #74450 @csoulios
  - [X] #74939 @csoulios
  - [X] #78012 @csoulios
- [x] #76766 @csoulios
- [X] #78790 @imotov
- [x] #79136 @weizijun

#### Phase 2 - Ingest
- [x] Dimension-based tsid generator
  - [x] #77154 @nik9000 
  - [x] Add TSDB-specific tests 
    - [X] #78208 @nik9000 
    - [X] #78042 @nik9000 
    - [X] #78038 @nik9000 
    - [X] #78034 @nik9000 
    - [X] #78028 @nik9000 
    - [X] #78022 @nik9000 
  - [x] #80276 @csoulios
  - [x] #81382 @csoulios ([prototype](https://github.com/elastic/elasticsearch/pull/75638/files#diff-4c1601543024812997c062c3b5e2dfc1aae96707d9aec8bba2c00e0b0a69c86d)) 
  - [x] #81998 @csoulios 

- [x] Routing
  - [x] #77211 @nik9000 
  - [x] #77731 @nik9000 
  - [x] #79384 @nik9000
  - [x] #79520 @nik9000
  - [x] #81125 @csoulios 
  - [x] #79826 (Speed up xcontent filtering) @weizijun 
  - [x] Have a good hard look at the switch statement in `BulkOperation`. Maybe we can make this simpler. 
    - [X] #79394 @nik9000
    - [X] #79472 @nik9000
    - [x] #80624
  - [x] #81436 @csoulios 
  - [x] Test and fix get-by-id #82633 (See linked issue for greater description of sub-points)
    - [x] Initial implementation of `_id` for tsid (#82633)
    - [x] Generate better error messages when `_id` is automatically generated (#84903, #84962)
    - [x] Improve error messages on version conflict to include `_tsid` and `@timestamp` (#84957)
    - [x] Size
[...]

PR #82944, opened 2022-01-24
[RollupV2]: make RollupAction available and improve some features

weizijun:
### Background

1. Rollup is very useful in TSDB case. TSDB can use rollup to do timestamp downsample.
2. The Rollup refactoring is under development at https://github.com/elastic/elasticsearch/issues/42720, and it seems to be blocked.

I know rollups is a bit of a complicated beast of a project. Yes, there may be a lot of design details to discuss.  But downsampling is very useful in our use case. So I'm continuing to develop rollup refactoring. 

### Detail

I think the RollupAction is the basement of the rollup refactoring. So I do some improvement to make the API available to use and add some feature in the RollupAction API.

Here's a list of what I've done:
1. Fix RollupAction API bugs:

- Fix when write sparse metric column, it will cause write exception.
- Fix when read sparse group column, it will cause NPE exception.
- Fix when date_histogram field spare, calculate range query exception.
- Fix when source index not found, it will case NPE exception.
- Fix when rollup index is exists, it will continue to rollup until finished rollup, cause resize exception.
- Fix when origin index's shard is more than 1,  RollupIndexerAction get shard list error.

2. Add some features

- Rollup group, metrics support wildcard fields.
- Support rollup index reuse the origin index's settings.
- Support get rollup tasks status, and support cancel rollup tasks.
- Support rollup group terms configured `_tsid` field.
- Support rollup time_series index in the new execution mode(https://github.com/elastic/elasticsearch/pull/82129). It greatly improves the performance of rollup tasks.
- Logging optimize: log the execution cost, optimize index failed log, and so on.
- Rollup ilm policy index optimize, remove the ilm police in tmp index, and add back in the resize settings
- Add many tests, to make the code coverage more than 90%.

Since this is the first PR of my rollup feature, I won't change the code that might affect other features.

Some features are ready in our internal repository. Here is the TODO list in the next PRs:

- Support rollup `aggregate_metric_double` type metrics, as rollup index use `aggregate_metric_double` to store metric type, so we can't rollup a rollupe index.
- Support search rollup index, now when rollup a data stream index, the rollup index will also add to data stream's concrete indices, It will cause error search result. And the search rollup function will select the best indices for a search DSL that match the rollup config.

There are a few more things to discuss:

- If the origin index must be a read-only index? if yes, it will add the check logic.
[...]

PR #85708, opened 2022-04-05
TSDB: Implement downsampling on time-series indices

### LINKED ISSUE #65769: Rollup Action must throw exception when rollup index already exists
Currently, the Rollup Action defined in #64900 does not have safeguards against the API running on anything other than an existing-non-rollup-index as the source index, and a rollup-index that does not exist yet.

If rollup index already exists, an exception should be thrown.

relates #42720.

### LINKED ISSUE #69799: Various RollupActionSingleNodeTests failing with IndexNotFoundException
**Build scan**: 

- https://gradle-enterprise.elastic.co/s/ib5r6mw23rtcs
- https://gradle-enterprise.elastic.co/s/ezyp34u27mc36
- https://gradle-enterprise.elastic.co/s/s6zuen2tr2dva

**Repro line**: `./gradlew ':x-pack:plugin:rollup:test' --tests "org.elasticsearch.xpack.rollup.v2.RollupActionSingleNodeTests.testMaxMetric" -Dtests.seed=AF76AE54E383FA3F -Dtests.security.manager=true -Dtests.locale=nb -Dtests.timezone=Asia/Pontianak -Druntime.java=11`

**Reproduces locally?**: No

**Applicable branches**: `master`, `7.x` and `7.12`

**Failure history**:

https://build-stats.elastic.co/goto/de41e6b94aa2ae417cbe342ddc4d0741 indicates a few failures a day.

**Failure excerpt**:

```
  1> [2021-03-02T15:01:32,165][INFO ][o.e.x.r.v.RollupActionSingleNodeTests] [testAvgMetric] before test
  1> [2021-03-02T15:01:32,167][DEPRECATION][o.e.d.c.m.MetadataCreateIndexService] [node_s_0] index [jnmin] matches multiple legacy templates [one_shard_index_template, random-soft-deletes-template], composable templates will only match a single template
  1> [2021-03-02T15:01:32,173][INFO ][o.e.c.m.MetadataCreateIndexService] [node_s_0] [jnmin] creating index, cause [api], templates [one_shard_index_template, random-soft-deletes-template], shards [1]/[0]
  1> [2021-03-02T15:01:32,232][DEPRECATION][o.e.d.c.m.MetadataCreateIndexService] [node_s_0] index [.rolluptmp-kglmlq] matches multiple legacy templates [one_shard_index_template, random-soft-deletes-template], composable templates will only match a single template
  1> [2021-03-02T15:01:32,233][DEPRECATION][o.e.d.c.m.MetadataCreateIndexService] [node_s_0] index name [.rolluptmp-kglmlq] starts with a dot '.', in the next major version, index names starting with a dot are reserved for hidden indices and system indices
  1> [2021-03-02T15:01:32,236][INFO ][o.e.c.m.MetadataCreateIndexService] [node_s_0] [.rolluptmp-kglmlq] creating index, cause [auto(bulk api)], templates [one_shard_index_template, random-soft-deletes-template], shards [1]/[0]
  1> [2021-03-02T15:01:32,362][INFO ][o.e.c.r.a.AllocationService] [node_s_0] Cluster health status changed from [YELLOW] to [GREEN] (reason: [shards started [[.rolluptmp-kglmlq][0]]]).
  1> [2021-03-02T15:01:32,762][INFO ][o.e.c.m.MetadataCreateIndexService] [node_s_0] [.rolluptmp-ecjbud] creating index, cause [api], templates [], shards [1]/[1]
  1> [2021-03-02T15:01:32,834][INFO ][o.e.c.m.MetadataMappingService] [node_s_0] [.rolluptmp-kglmlq/3YQHCIUPTW6YJhWTeug1Og] create_mapping [_doc]
  1> [2021-03-02T15:01:33,091][INFO ][o.e.x.r.v.RollupShardIndexer] [node_s_0] Successfully sent [262], indexed [262]
  1> [2021-03-02T15:01:33,149][INFO ][o.e.c.m.MetadataCreateIndexService] [node_s_0] applying create index request using existing index [.rolluptmp-kglmlq] metadata
  1> [2021-03-02T15:01:33,153][INFO ][o.e.c.m.MetadataCreateIndexService] [node_s_0] [kglmlq] creating index, cause [clone_index], templates [], shards [1]/[1]
  1> [2021-03-02T15:01:33,220][INFO ][o.e.x.r.v.RollupShardIndexer] [node_s_0] Successfully sent [10], indexed [10]
  1> [2021-03-02T15:01:33,328][INFO ][o.e.c.m.MetadataCreateIndexService] [node_s_0] applying create index request using existing index [.rolluptmp-ecjbud] metadata
  1> [2021-03-02T15:01:33,330][INFO ][o.e.c.m.MetadataCreateIndexService] [node_s_0] [ecjbud] creating index, cause [clone_index], templates [], shards [1]/[1]
  1> [2021-03-02T15:01:33,435][INFO ][o.e.c.m.MetadataMappingService] [node_s_0] [ecjbud/3VA-iINcTl6RCqNgz1jHxQ] create_mapping [_doc]
  1> [2021-03-02T15:01:33,509][INFO ][o.e.c.m.MetadataMappingService] [node_s_0] [kglmlq/g1PmmRXuQQ-6VUQ-duXhUA] create_mapping [_doc]
  1> [2021-03-02T15:01:33,633][INFO ][o.e.c.m.MetadataDeleteIndexService] [node_s_0] [.rolluptmp-ecjbud/EbxCDhmdS-SltrHrWilOOQ] deleting index
  1> [2021-03-02T15:01:33,768][INFO ][o.e.c.m.MetadataDeleteIndexService] [node_s_0] [.rolluptmp-kglmlq/3YQHCIUPTW6YJhWTeug1Og] deleting index
  1> [2021-03-02T15:01:33,780][INFO ][o.e.x.r.v.RollupActionSingleNodeTests] [testAvgMetric] after test
  2> REPRODUCE WITH: ./gradlew ':x-pack:plugin:rollup:test' --tests "org.elasticsearch.xpack.rollup.v2.RollupActionSingleNodeTests.testAvgMetric" -Dtests.seed=5EC5B0A6CCAA9904 -Dtests.security.manager=true -Dtests.locale=is -Dtests.timezone=Africa/Lubumbashi -Druntime.java=8
  2> [.rolluptmp-kglmlq/3YQHCIUPTW6YJhWTeug1Og] IndexNotFoundException[no such index [.rolluptmp-kglmlq]]
        at __randomizedtesting.SeedInfo.seed([5EC5B0A6CCAA9904:73AACC3E1E9D7A9D]:0)
        at org.elasticsearch.cluster.metadata.Metadata.getIndexSafe(Metadata.java:719)
        at org.elasticsearch.cluster.metadata.MetadataDeleteIndexService.deleteIndices(MetadataDeleteIndexService.java:80)
        at org.elasticsearch.cluster.metadata.MetadataDeleteIndexService$1.execute(MetadataDeleteIndexService.java:67)
        at org.elasticsearch.cluster.ClusterStateUpdateTask.execute(ClusterStateUpdateTask.java:48)
        at org.elasticsearch.cluster.service.MasterService.executeTasks(MasterService.java:691)
        at org.elasticsearch.cluster.service.MasterService.calculateTaskOutputs(MasterService.java:313)
        at org.elasticsearch.cluster.service.MasterService.runTasks(MasterService.java:208)
        at org.elasticsearch.cluster.service.MasterService.access$000(MasterService.java:62)
        at org.elasticsearch.cluster.service.MasterService$Batcher.run(MasterService.java:140)
        at org.elasticsearch.cluster.service.TaskBatcher.runIfNotProcessed(TaskBatcher.java:139)
        at org.elasticsearch.cluster.service.TaskBatcher$BatchedTask.run(TaskBatcher.java:177)
        at org.elasticsearch.common.util.concurrent.ThreadContext$ContextPreservingRunnable.run(ThreadContext.java:673)
        at org.elasticsearch.common.util.concurrent.PrioritizedEsThreadPoolExecutor$TieBreakingPrioritizedRunnable.runAndClean(PrioritizedEsThreadPoolExecutor.java:241)
        at org.elasticsearch.common.util.concurrent.PrioritizedEsThreadPoolExecutor$TieBreakingPrioritizedRunnable.run(PrioritizedEsThreadPoolExecutor.java:204)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

csoulios:
This PR implements downsampling operation on time series indices.

The PR creates a `_rollup` endpoint that allows users to downsample an index and can be
accessed by the following call:

```
POST /<source_index>/_rollup/<rollup_index>
{
    "fixed_interval": "1d"
}
```

### Requirements
An index can be downsampled if all of the following requirements are met:
 
 - Must be a time series index (have the `index.mode: time_series` index setting)
 - Must not be writeable (have the `index.blocks.write: true` index setting)
 - Must have dimension fields marked with mapping parameter `time_series_dimension: true`
 - Must have metric fields marked with mapping parameter `time_series_metric`

### Steps

The steps followed by the downsampling action are the following: 

1. Read the index mapping from the source index and extract the dimension and metric fields.
1. Create a temporary rollup index with the following settings/mapping:
    1. The following index settings will be copied from the original index
        - `index.mode`
        - `index.routing_path`
        - `index.time_series.start_time`
        - `index.time_series.end_time`
    1. Set the following index settings to the values of the original index:
        - `index.rollup.source.uuid`
        - `index.rollup.source.name`
    1. All fields mapped as `time_series_dimension` will be created in the rollup index with exactly the same mapping.
    1. All fields mapped as `time_series_metric` will be created and their field type will change to `aggregate_metric_double`.
1. Use the [`TimeSeriesIndexSearcher`](https://github.com/elastic/elasticsearch/blob/a89d4c34cbd8e5b3b3879df3aec632e7494e3c35/server/src/main/java/org/elasticsearch/search/aggregations/timeseries/TimeSeriesIndexSearcher.java) implementation to traverse over the (sorted by `_tsid`, `@timestamp`) source index and create the rollup index. 
1. Rename the temporary rollup index to its defined name.
1. If the source index is a member of a data stream, we must replace the original index with the rollup index. So the rollup index is added and the source index is deleted. The addition and deletion of the indices should be an atomic action.

### TODO 

- Implement logic for storing  fields that are neither dimensions nor metrics (aka labels)
- Make rollup task cancellable
- More testing

Fixes #65769
Fixes #69799
Finally, this PR is based on the code written for https://github.com/elastic/elasticsearch/pull/64900


server/src/main/java/org/elasticsearch/search/aggregations/timeseries/TimeSeriesIndexSearcher.java
```diff
@@ -78,7 +78,7 @@ public void search(Query query, BucketCollector bucketCollector) throws IOExcept
         PriorityQueue<LeafWalker> queue = new PriorityQueue<>(searcher.getIndexReader().leaves().size()) {
             @Override
             protected boolean lessThan(LeafWalker a, LeafWalker b) {
-                return a.timestamp < b.timestamp;
+                return a.timestamp > b.timestamp;
```
csoulios, line 81: I have changed the sorting order because I was getting results in a wrong timestamp order when receiving documents from different segments. By changing this, the doc order problem was fixed. 

I think this bug is fixed by this PR: https://github.com/elastic/elasticsearch/pull/85526

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/XPackClientPlugin.java
```diff
@@ -411,12 +413,6 @@
             )
         );
 
-        // rollupV2
-        if (RollupV2.isEnabled()) {
```
csoulios, line 415: I removed the feature flag we had used for Rollups v2. Now, rollups code works as an operation for time_series indices

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/RollupActionConfig.java
```diff
@@ -6,157 +6,206 @@
...
-            new ParseField(RollupActionGroupConfig.NAME)
+
+        PARSER.declareField(
+            constructorArg(),
+            p -> new DateHistogramInterval(p.text()),
+            new ParseField(FIXED_INTERVAL),
+            ObjectParser.ValueType.STRING
         );
-        PARSER.declareObjectArray(optionalConstructorArg(), (p, c) -> MetricConfig.fromXContent(p), new ParseField(MetricConfig.NAME));
+        PARSER.declareStringOrNull(optionalConstructorArg(), new ParseField(TIME_ZONE));
```
csoulios, line 78: Since the team has decided that we are going to support only the UTC timezone for now, maybe we should remove the `timezone` parameter from the `_rollup` operation and add it when we add support for other timezones.

csoulios: The team decided to remove the `time_zone` parameter from the `_rollup` API and add it later more timezones are supported.

server/src/main/java/org/elasticsearch/search/aggregations/timeseries/TimeSeriesIndexSearcher.java
csoulios: Issue fixed by PR https://github.com/elastic/elasticsearch/pull/85526

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/RollupActionConfig.java
csoulios: Removed `time_zone` rollup param at [d905038](https://github.com/elastic/elasticsearch/pull/85708/commits/d905038b8eb281c2933d209af57d4bd1e9a95d20)

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -250,112 +298,165 @@ public void onFailure(Exception e) {
...
+                    List<Index> backingIndices = new ArrayList<>(originalDataStream.getIndices().size());
+                    // Adding the rollup index to the beginning of the list will prevent it from ever being
                     // considered a write index
                     backingIndices.add(rollupIndex);
-                    backingIndices.addAll(originalDataStream.getIndices());
+                    // Add all indices except the source index
+                    backingIndices.addAll(originalDataStream.getIndices().stream().filter(i -> i.getName() != sourceIndexName).toList());
                     DataStream dataStream = new DataStream(
                         originalDataStream.getName(),
                         backingIndices,
```
csoulios, line 462: Adding the rollup index to the data stream and remove the source index, hopefully in an atomic action.

@martijnvg your feedback on this would be greatly appreciated. Thanks

martijnvg: The data stream in cluster state is updated in a atomic way here, since
the original index is removed and rolled up index is added in the same `ClusterStateUpdateTask` instance.

csoulios: Thank you. I pushed new code that uses `DataStream#replaceBackingIndex(...)`

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/action/RollupIndexerAction.java
```diff
@@ -85,26 +115,28 @@ public ActionRequestValidationException validate() {
...
+            builder.array("dimension_fields", dimensionFields);
+            builder.array("metric_fields", metricFields);
             builder.endObject();
             return builder;
         }
 
         @Override
         public int hashCode() {
-            return Objects.hash(rollupRequest);
+            return Objects.hash(rollupIndex, rollupRequest, dimensionFields, metricFields);
```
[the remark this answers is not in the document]
csoulios: Good catch. Changed it to a more conventional `hashCode()` implementation.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/RollupActionConfig.java
```diff
@@ -6,157 +6,205 @@
...
-    public void validateMappings(
-        final Map<String, Map<String, FieldCapabilities>> fieldCapsResponse,
-        final ActionRequestValidationException validationException
-    ) {
-        groupConfig.validateMappings(fieldCapsResponse, validationException);
-        for (MetricConfig m : metricsConfig) {
-            m.validateMappings(fieldCapsResponse, validationException);
-        }
+    @Override
+    public void writeTo(final StreamOutput out) throws IOException {
```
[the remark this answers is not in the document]
csoulios: Fixed

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -109,137 +109,184 @@ protected void masterOperation(
...
+            );
+            return;
+        }
+        // Assert source index is read-only
+        if (state.blocks().indexBlocked(ClusterBlockLevel.WRITE, sourceIndexName) == false) {
+            listener.onFailure(
+                new ElasticsearchException(
+                    "Rollup requires setting [" + IndexMetadata.SETTING_BLOCKS_WRITE + " = true] for index [" + sourceIndexName + "]"
+                )
+            );
```
martijnvg, line 141: missing return statement?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -109,137 +110,184 @@ protected void masterOperation(
...
+        final String rollupIndexName = request.getRollupIndex();
+        if (state.getMetadata().index(rollupIndexName) != null) {
+            listener.onFailure(new ResourceAlreadyExistsException("Rollup index [{}] already exists.", rollupIndexName));
+            return;
+        }
+        final String tmpIndexName = createTmpIndexName(rollupIndexName);
+        if (state.getMetadata().index(tmpIndexName) != null) {
+            listener.onFailure(new ResourceAlreadyExistsException("Temporary rollup index [{}] already exists.", tmpIndexName));
             return;
         }
```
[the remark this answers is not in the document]
martijnvg: Without knowing all the details here, if this action is doing many things that are not atomic then there is room for failure here that needs to be taken into account. But maybe this action should only do the actual rollup and an the rollup ilm action should be changed to do the action listed below here? In case of failure, ILM can retry from the step that has failed.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: Fixed in https://github.com/elastic/elasticsearch/pull/85708/commits/cd23fd713ba6adbaa3be8c3987f27900b6a0b476

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -108,137 +108,185 @@ protected void masterOperation(
...
+                        request,
+                        dimensionFieldCaps.keySet().toArray(new String[0]),
+                        metricFieldCaps.keySet().toArray(new String[0])
+                    );
+
                     client.execute(RollupIndexerAction.INSTANCE, rollupIndexerRequest, ActionListener.wrap(indexerResp -> {
                         if (indexerResp.isCreated()) {
-                            // 4.
+                            // 4. Make temp index read-only
+                            UpdateSettingsRequest updateSettingsReq = new UpdateSettingsRequest(WRITE_BLOCKED_SETTINGS, tmpIndexName);
```
imotov, line 242: I wonder if we should propagate parent task here as well.

imotov (commented):
Looks reasonable to me, but I would defer to @nik9000 for the final review.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/MetricFieldProducer.java
```diff
@@ -0,0 +1,237 @@
...
+
+        abstract void collect(double number);
+
+        abstract Number get();
+
+        abstract void reset();
+    }
+
+    static class Max extends Metric {
+        private Double max;
```
salvatore-campagna, line 72: I see we are using double here to keep track of metrics. In the aggregation framework using double as the underlying common data type is a source of issues/pain. Should we consider using BigDecimal? Precision would benefit...

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/MetricFieldProducer.java
```diff
@@ -0,0 +1,237 @@
...
+
+        @Override
+        public Object value() {
+            Map<String, Object> metricValues = new HashMap<>();
+            for (MetricFieldProducer.Metric metric : metrics()) {
+                if (metric.get() != null) {
+                    metricValues.put(metric.name, metric.get());
+                }
+            }
+            return metricValues;
```
salvatore-campagna, line 217: Maybe return as unmodifiable?

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/IndexLifecycle.java
```diff
@@ -320,15 +319,11 @@ private static List<NamedXContentRegistry.Entry> xContentEntries() {
                     new ParseField(SearchableSnapshotAction.NAME),
                     SearchableSnapshotAction::parse
                 ),
-                new NamedXContentRegistry.Entry(LifecycleAction.class, new ParseField(MigrateAction.NAME), MigrateAction::parse)
+                new NamedXContentRegistry.Entry(LifecycleAction.class, new ParseField(MigrateAction.NAME), MigrateAction::parse),
+                new NamedXContentRegistry.Entry(LifecycleAction.class, new ParseField(RollupILMAction.NAME), RollupILMAction::parse)
```
nik9000, line 323: Does this need to live behind the tsdb feature flag?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
```diff
@@ -248,319 +184,212 @@ public void afterBulk(long executionId, BulkRequest request, Throwable failure)
...
+                    final int docCount = docCountProvider.getDocCount(docId);
+                    rollupBucketBuilder.collectDocCount(docCount);
+                    for (Map.Entry<String, FormattedDocValues> e : metricsFieldLeaves.entrySet()) {
+                        String fieldName = e.getKey();
+                        FormattedDocValues leafField = e.getValue();
+
+                        if (leafField.advanceExact(docId)) {
+                            for (int i = 0; i < leafField.docValueCount(); i++) {
+                                Object obj = leafField.nextValue();
+                                if (obj instanceof Number number) {
```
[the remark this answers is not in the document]
nik9000: I talked with @csoulios and it's something that he does want to move out but would like to save for a follow up change.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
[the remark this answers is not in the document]
nik9000: Should we remove the temporary index?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/MetricFieldProducer.java
csoulios: You are right that double is potentially a source of drouble. However, all our aggs already use doubles everywhere. Also, `aggregate_metric_double` will store the result as a double

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/MetricFieldProducer.java
csoulios: Fixed [ab1a4d8](https://github.com/elastic/elasticsearch/pull/85708/commits/ab1a4d899e60009b41a5c41b9b96f0c069723941)

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/MetricFieldProducer.java
```diff
@@ -135,40 +131,25 @@ void reset() {
...
+            assert fieldType.getMetricType() != null;
+
+            final List<Metric> list = new ArrayList<>();
+            for (String metricName : fieldType.getMetricType().supportedAggs()) {
+                switch (metricName) {
+                    case "min" -> list.add(new Min());
+                    case "max" -> list.add(new Max());
+                    case "sum" -> list.add(new Sum());
+                    case "value_count" -> list.add(new ValueCount());
+                    default -> throw new IllegalArgumentException("Unsupported metric type [" + metricName + "]");
```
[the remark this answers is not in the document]
csoulios: I totally changed this part. Now there are two implementations of `MetricFieldProducer` (`GaugeMetricFieldProducer` and `CounterMetricFieldProducer`) that are created based on metric type of the field

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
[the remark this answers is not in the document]
csoulios: Removed temp index in [286cdbd](https://github.com/elastic/elasticsearch/pull/85708/commits/286cdbd3af0dcd6c504285d2f35be661a175ecd0)

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/RollupActionConfig.java
```diff
@@ -6,157 +6,205 @@
...
+        timeZone = in.readString();
     }
 
-    public RollupActionGroupConfig getGroupConfig() {
-        return groupConfig;
+    @Override
+    public void writeTo(final StreamOutput out) throws IOException {
+        out.writeString(FIXED_INTERVAL);
+        fixedInterval.writeTo(out);
+        out.writeString(timeZone);
```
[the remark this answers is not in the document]
csoulios: I chose to serialize timezone here because I thought we are going make it configurable in the future., so we better send a fixed string now. This will keep the protocol stable in the future. The same for interval type, I wanted to keep this information.

But you never know we may never find it useful to add support for multiple timezones and interval. We will definitely find something else to add there. So, for the sake of YAGNI principle, you think I should remove them?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/TimeseriesLifecycleType.java
```diff
@@ -78,7 +78,7 @@ public class TimeseriesLifecycleType implements LifecycleType {
         AllocateAction.NAME,
         MigrateAction.NAME,
         FreezeAction.NAME,
-        RollupV2.isEnabled() ? RollupILMAction.NAME : null
+        RollupILMAction.NAME
```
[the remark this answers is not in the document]
csoulios: The rollup action will be available but it will not operate on indices that are not time series indices. 

So, in a similar way as we did for `TimeSeriesIdFieldMapper` I did not put this behind a feature flag. I can add it though, if you think it makes sense.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/ilm/RollupILMActionTests.java
```diff
@@ -82,9 +81,11 @@ RollupILMAction notCopy(RollupILMAction rollupILMAction) {
         String newRollupPolicy = rollupILMAction.rollupPolicy();
         switch (randomIntBetween(0, 1)) {
             case 0 -> {
-                List<MetricConfig> metricConfigs = new ArrayList<>(rollupILMAction.config().getMetricsConfig());
-                metricConfigs.add(new MetricConfig(randomAlphaOfLength(4), Collections.singletonList("max")));
-                newConfig = new RollupActionConfig(rollupILMAction.config().getGroupConfig(), metricConfigs);
+                DateHistogramInterval fixedInterval = ConfigTestHelpers.randomInterval();
```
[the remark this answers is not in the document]
csoulios: I didn't know there was such a method. Thanks :)

Done in [c22b6ba](https://github.com/elastic/elasticsearch/pull/85708/commits/c22b6baadf30b95880e0218e195270be3ffba5be)

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/Rollup.java
```diff
@@ -160,15 +156,13 @@ public List<RestHandler> getRestHandlers(
                 new ActionHandler<>(GetRollupCapsAction.INSTANCE, TransportGetRollupCapsAction.class),
                 new ActionHandler<>(GetRollupIndexCapsAction.INSTANCE, TransportGetRollupIndexCapsAction.class),
                 new ActionHandler<>(XPackUsageFeatureAction.ROLLUP, RollupUsageTransportAction.class),
-                new ActionHandler<>(XPackInfoFeatureAction.ROLLUP, RollupInfoTransportAction.class)
+                new ActionHandler<>(XPackInfoFeatureAction.ROLLUP, RollupInfoTransportAction.class),
+                // Rollup / Downsampling
+                new ActionHandler<>(RollupIndexerAction.INSTANCE, TransportRollupIndexerAction.class),
```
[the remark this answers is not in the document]
csoulios: Fixed in [86c3347](https://github.com/elastic/elasticsearch/pull/85708/commits/86c3347410aaf35140a79d839ff0494d1edcae35)

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/IndexLifecycle.java
csoulios: Fixed in [86c3347](https://github.com/elastic/elasticsearch/pull/85708/commits/86c3347410aaf35140a79d839ff0494d1edcae35)

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/TimeseriesLifecycleType.java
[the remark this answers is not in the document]
csoulios: Fixed in [86c3347](https://github.com/elastic/elasticsearch/pull/85708/commits/86c3347410aaf35140a79d839ff0494d1edcae35)

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/XPackClientPlugin.java
csoulios: Added the TSDB feature flag

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/MetricFieldProducer.java
```diff
@@ -0,0 +1,237 @@
...
+import org.elasticsearch.index.mapper.MappedFieldType;
+import org.elasticsearch.index.query.SearchExecutionContext;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+abstract class MetricFieldProducer {
```
nik9000, line 19: It's probably worth sprinkling javadoc in a bunch of places, here included.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
```diff
@@ -229,14 +168,18 @@ public void afterBulk(long executionId, BulkRequest request, BulkResponse respon
...
                 }
             }
 
             @Override
             public void afterBulk(long executionId, BulkRequest request, Throwable failure) {
-                long items = request.numberOfActions();
-                numSent.addAndGet(-items);
+                if (failure != null) {
+                    long items = request.numberOfActions();
+                    numSent.addAndGet(-items);
```
nik9000, line 180: Even if it failed it was still "sent", right?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
```diff
@@ -229,14 +168,18 @@ public void afterBulk(long executionId, BulkRequest request, BulkResponse respon
                                 (msg1, msg2) -> Objects.equals(msg1, msg2) ? msg1 : msg1 + "," + msg2
                             )
                         );
-                    logger.error("failures: [{}]", failures);
+                    numFailed.addAndGet(failures.size());
+                    logger.error("Shard {} failed to populate rollup index: [{}]", indexShard.shardId(), failures);
```
nik9000, line 172: Should this abort the load process?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
```diff
@@ -229,14 +168,18 @@ public void afterBulk(long executionId, BulkRequest request, BulkResponse respon
...
 
             @Override
             public void afterBulk(long executionId, BulkRequest request, Throwable failure) {
-                long items = request.numberOfActions();
-                numSent.addAndGet(-items);
+                if (failure != null) {
+                    long items = request.numberOfActions();
+                    numSent.addAndGet(-items);
+                    numFailed.addAndGet(items);
+                }
```
nik9000, line 182: Should this abort the load process?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
```diff
@@ -248,319 +191,203 @@ public void afterBulk(long executionId, BulkRequest request, Throwable failure)
...
-
-    private void indexBucket(BucketKey key, List<FieldMetricsProducer> fieldsMetrics, int docCount) {
-        IndexRequestBuilder request = client.prepareIndex(tmpIndex);
-        Map<String, Object> doc = Maps.newMapWithExpectedSize(2 + key.groupFields.size() + fieldsMetrics.size());
-        doc.put(DocCountFieldMapper.NAME, docCount);
-        doc.put(timestampField.name(), timestampFormat.format(key.timestamp));
+    private class TimeSeriesBucketCollector extends BucketCollector {
+        private final BulkProcessor bulkProcessor;
+        private long docsProcessed;
+        private long bucketsCreated;
```
nik9000, line 197: I think it'd be useful to plumb these into the task so the get task API can show you the progress. It'd be fine to do in a follow up though.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -277,126 +398,126 @@ private static XContentBuilder getDynamicTemplates(XContentBuilder builder) thro
...
-            .field("type", DateFieldMapper.CONTENT_TYPE)
-            .startObject("meta")
-            .field(dateIntervalType, dateInterval)
-            .field(RollupActionDateHistogramGroupConfig.CalendarInterval.TIME_ZONE, tz)
-            .endObject()
-            .endObject();
+            @Override
+            public void clusterStateProcessed(ClusterState oldState, ClusterState newState) {
+                // 7. Delete the source index
+                deleteSourceIndex(sourceIndexName, rollupIndexName, listener);
```
nik9000, line 424: Will we always want to delete the source index? I get that we do if we're replacing the source in the data stream but if it isn't in a data stream is that required? Maybe I'm thinking about this funny - maybe we have to because of index patterns or something.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -277,126 +398,126 @@ private static XContentBuilder getDynamicTemplates(XContentBuilder builder) thro
...
+    }
+
+    private void deleteSourceIndex(final String sourceIndex, final String rollupIndex, ActionListener<AcknowledgedResponse> listener) {
+        client.admin().indices().delete(new DeleteIndexRequest(sourceIndex), new ActionListener<>() {
+            @Override
+            public void onResponse(AcknowledgedResponse acknowledgedResponse) {
+                if (acknowledgedResponse.isAcknowledged()) {
+                    // Source index was deleted successfully.
+                    // 8. Finalize the rollup index by setting its state to SUCCESS
+                    finalizeRollupIndex(rollupIndex, listener);
```
nik9000, line 489: Is it possible to submit the delete action and for the source index to be deleted and then something to go wrong and you don't get the response here so the rollup index isn't "finalized"? I think it'd be nice to make the replacement in the data stream, the delete, and the finalize atomic in one action.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -108,137 +104,161 @@ protected void masterOperation(
...
+                Settings.builder()
+                    .put(IndexMetadata.INDEX_HIDDEN_SETTING.getKey(), true)
+                    .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, sourceIndexMetadata.getNumberOfShards())
+                    .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 0)
+                    .put(IndexMetadata.INDEX_ROLLUP_STATUS.getKey(), IndexMetadata.RollupTaskStatus.STARTED)
+                    .build()
+            ).mappings(XContentHelper.convertToJson(BytesReference.bytes(mapping), false, XContentType.JSON));
+
+            // 2. Create rollup index
+            submitUnbatchedTask("create-rollup-index", new ClusterStateUpdateTask() {
```
nik9000, line 211: I think maybe we can't use `submitUnbatchedTask`, right? The docs say " This method exists for legacy reasons but is deprecated and forbidden in new production code because unbatched tasks are a source of performance and stability bugs."

x-pack/plugin/security/qa/operator-privileges-tests/build.gradle
```diff
@@ -39,7 +39,6 @@ testClusters.configureEach {
   setting 'path.repo', repoDir.absolutePath
 
   requiresFeature 'es.user_profile_feature_flag_enabled', Version.fromString("8.1.0")
-  requiresFeature 'es.rollup_v2_feature_flag_enabled', Version.fromString("8.0.0")
```
nik9000, line 42: Should this have the time series flag instead?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: Do not deleting the source index when an index is not a data stream backing index has certain benefits (mainly flexibility), but also adds complexity in certain areas such as ILM.

We have discussed this topic in the past and we have agreed that the souce index should be deleted in any case, even if it's not a data stream index.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: I have seen the deprecation comment and researched a bit more into this. Method `clusterService.submitUnbatchedStateUpdateTask()` is deprecated by currently used all over the place. 

My goal was to get this PR merged and then work in this issue in a separate PR. 

@DaveCTurner do you have an opinion if we should merge this PR while using the deprecated method?

DaveCTurner: I would very much prefer to avoid adding more unbatched tasks, they are truly awful for performance in large clusters. Here for instance if ILM were paused for a while then it might need to create thousands of rollup indices when restarted, and as currently written it would put all of those tasks in the queue at once and then work through them one-at-a-time.

That said, here it looks like we're just reworking an existing unbatched task rather than adding a new one right? If so this isn't really changing anything in terms of the cluster state tasks so it seems ok to proceed.

DaveCTurner: > That said, here it looks like we're just reworking an existing unbatched task rather than adding a new one right? 

Actually now that I look closer it seems a bit more than that, we're introducing tasks that weren't there before which makes me concerned. Is there a good reason to avoid batching these these things? It shouldn't be too hard. See `MetadataIndexTemplateService#TEMPLATE_TASK_EXECUTOR` for a reasonably simple example.

csoulios: @DaveCTurner thanks for the feedback.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
nik9000: We're holding off on early abort for a follow up PR.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
nik9000: We're holding off on early abort for a follow up PR.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: Fixed in [332bef6](https://github.com/elastic/elasticsearch/pull/85708/commits/332bef648bbbddae58756acc40fca1adc9d3b257)

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: I modified the code so that the following operations happen within a single state update request, so they are all performed atomically.

 - If source index belongs to a data stream, add the rollup index and remove the source index
 - Mark rollup index as completed successfully
 - Delete the source index


See [ee42ef5](https://github.com/elastic/elasticsearch/pull/85708/commits/ee42ef5aacc5328ceef64b4f052f2e11a0537f74)

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: Fixed. Now it propagates parent task for all client calls

x-pack/plugin/security/qa/operator-privileges-tests/build.gradle
csoulios: Fixed in [270ea69](https://github.com/elastic/elasticsearch/pull/85708/commits/270ea69604e30fc6f3e1fc6993f9ddf8d84c91ba)

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
csoulios: I think here there may be a problem with retries. If failures are retried, then same docs are counted multiple times numSent. Right?

In the end we require that numSent == numIndexed, so if there are retries we may get failures because of this

nik9000 (commented):
I looked at it for a while and left some comments. I think I want to have another look on Monday too.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -108,144 +129,165 @@ protected void masterOperation(
...
+
+        // Rollup will perform the following tasks:
+        // 1. Extract rollup config from source index field caps
+        // 2. Create the rollup index
+        // 3. Run rollup indexer
+        // 4. Make rollup index read-only and set replicas
+        // 5. Add rollup index to the data stream
+        // 6. Mark rollup index as "completed successfully"
+        // 7. Delete the source index
+        // 8. Refresh rollup index
```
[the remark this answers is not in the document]
csoulios: Happy to do it, but I am not sure I understand why data may vanish

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -108,144 +129,165 @@ protected void masterOperation(
...
+             */
+            for (Map.Entry<String, Map<String, FieldCapabilities>> e : fieldCapsResponse.get().entrySet()) {
+                String field = e.getKey();
+                FieldCapabilities fieldCaps = e.getValue().values().iterator().next();
+                if (fieldCaps.isDimension()) {
+                    dimensionFieldCaps.put(field, fieldCaps);
+                } else if (e.getValue().values().iterator().next().getMetricType() != null) {
+                    metricFieldCaps.put(field, fieldCaps);
+                } else {
+                    // TODO: Field is not a dimension or a metric. Treat it as a tag
```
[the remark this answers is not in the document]
csoulios: Yes, in a following PR we will add the tags. I left this there as a placeholder

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -108,144 +129,165 @@ protected void masterOperation(
...
+                            UpdateSettingsRequest updateSettingsReq = new UpdateSettingsRequest(settings, rollupIndexName);
+                            updateSettingsReq.setParentTask(clusterService.localNode().getId(), task.getId());
                             client.admin().indices().updateSettings(updateSettingsReq, ActionListener.wrap(updateSettingsResponse -> {
                                 if (updateSettingsResponse.isAcknowledged()) {
-                                    // 5.
-                                    client.admin().indices().resizeIndex(resizeRequest, ActionListener.wrap(resizeResponse -> {
-                                        if (resizeResponse.isAcknowledged()) {
-                                            // 6.
-                                            publishMetadata(originalIndexName, tmpIndexName, rollupIndexName, listener);
+                                    // 5. Add rollup index to data stream and publish rollup metadata
```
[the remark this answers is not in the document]
csoulios: Not sure why we should wait for the replicas to recover. I was thinking that this can happen in parallel. Any concerns we may lose the primary shard while deleting the source index?

I am afraid that waiting for replicas to recover may take lots of time.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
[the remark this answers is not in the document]
nik9000: It's just vanish for a few seconds. If you put the new index into the data stream and we start querying it *before* an explicit refresh then some of the rollup that your wrote may not yet be visible. So data will appear to "vanish" until the refresh takes.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -277,132 +421,157 @@ private static XContentBuilder getDynamicTemplates(XContentBuilder builder) thro
...
+             * When creating the rollup index, we copy the index.number_of_shards from source index,
+             * and we set the index.number_of_replicas to 0, to avoid replicating the temp index.
+             * We will set the correct number of replicas later.
+             */
+            Settings.builder()
+                .put(IndexMetadata.INDEX_HIDDEN_SETTING.getKey(), true)
+                .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, sourceIndexMetadata.getNumberOfShards())
+                .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 0)
+                .put(IndexMetadata.INDEX_ROLLUP_STATUS.getKey(), IndexMetadata.RollupTaskStatus.STARTED)
+                .build()
```
nik9000, line 446: Maybe we should set the refresh_interval to -1 here. The default interval tries to make data visible while writing at the cost of less efficient writes.

nik9000: Also - I wonder if we should force_merge after we're done writing. It doesn't have to be now, but it should make the target index more efficient to query and we're not going to write to it again.

weizijun:
I added a PR(https://github.com/elastic/elasticsearch/pull/82944) in January, and discuss with @csoulios about the downsampling design several times. Now, I'm waiting for this PR to be merged. Then I can contribute my code together to the downsampling requirement.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: Fixed in [0ff80dc](https://github.com/elastic/elasticsearch/pull/85708/commits/0ff80dc4792a1caa76bbb44566e9f4f6e5c4774d)

- Set refresh_interval to -1
- Force merge index in the end

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -277,132 +421,157 @@ private static XContentBuilder getDynamicTemplates(XContentBuilder builder) thro
...
+        ActionListener<AcknowledgedResponse> listener
+    ) {
+        CreateIndexClusterStateUpdateRequest createIndexClusterStateUpdateRequest = new CreateIndexClusterStateUpdateRequest(
+            "rollup",
+            rollupIndexName,
+            rollupIndexName
+        ).settings(
+            /*
+             * When creating the rollup index, we copy the index.number_of_shards from source index,
+             * and we set the index.number_of_replicas to 0, to avoid replicating the temp index.
```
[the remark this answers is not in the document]
csoulios: Fixed in [0ff80dc](https://github.com/elastic/elasticsearch/pull/85708/commits/0ff80dc4792a1caa76bbb44566e9f4f6e5c4774d)

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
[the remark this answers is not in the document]
csoulios: Fixed in [0ff80dc](https://github.com/elastic/elasticsearch/pull/85708/commits/0ff80dc4792a1caa76bbb44566e9f4f6e5c4774d)

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
csoulios: Fixed in [839a29d](https://github.com/elastic/elasticsearch/pull/85708/commits/839a29df872cfdae88af8c8885297e8811ced7b3)

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -108,144 +129,165 @@ protected void masterOperation(
...
+            final Map<String, FieldCapabilities> dimensionFieldCaps = new HashMap<>();
+            final Map<String, FieldCapabilities> metricFieldCaps = new HashMap<>();
+            /*
+             * Rollup runs on a single index, and we do not expect multiple mappings for the same
+             * field. So, it is safe to select the first and only value of the FieldCapsResponse
+             * by running: e.getValue().values().iterator().next()
+             */
+            for (Map.Entry<String, Map<String, FieldCapabilities>> e : fieldCapsResponse.get().entrySet()) {
+                String field = e.getKey();
+                FieldCapabilities fieldCaps = e.getValue().values().iterator().next();
```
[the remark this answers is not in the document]
csoulios: Fixed in [1dd28fd](https://github.com/elastic/elasticsearch/pull/85708/commits/1dd28fdba12c1b9865ceb57e414001cc013da59a)

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/MetricFieldProducer.java
csoulios: Fixed in [de6b766](https://github.com/elastic/elasticsearch/pull/85708/commits/de6b766575f2ab5cd64067d6dced334d14c07208)

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -108,144 +133,214 @@ protected void masterOperation(
...
+                                                        rollupIndexName,
+                                                        parentTask,
+                                                        listener,
+                                                        new ElasticsearchException(
+                                                            "Failed to publish new cluster state with rollup metadata"
+                                                        )
+                                                    );
+                                                }
+                                            },
+                                                e -> deleteRollupIndex(
```
nik9000, line 290: I wonder if there is a good way to reduce the duplicate calls here. Oh well, in a follow up.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -108,144 +133,214 @@ protected void masterOperation(
...
+                                                    forceMergeIndex(
+                                                        rollupIndexName,
+                                                        parentTask,
+                                                        ActionListener.wrap(
+                                                            mergeIndexResp -> listener.onResponse(AcknowledgedResponse.TRUE),
+                                                            e -> listener.onFailure(
+                                                                new ElasticsearchException(
+                                                                    "Failed to force-merge index [" + rollupIndexName + "]",
+                                                                    e
+                                                                )
```
nik9000, line 274: Will this fail the request but not cause the destination index to be deleted? What will ILM think of that?

nik9000 (approved):
Let's get this in. There's certainly more to do but I feel like this is a good start for rollups.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: In [393d7c2](https://github.com/elastic/elasticsearch/pull/85708/commits/393d7c2cd2c8c75f88970ca6caee55c2be9b1f08), I changed it so that it does not fail the rollup action.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -267,12 +273,19 @@ protected void masterOperation(
...
-                                                                    "Failed to force-merge index [" + rollupIndexName + "]",
+                                                            e -> {
+                                                                /*
+                                                                 * At this point rollup has been successful even if force-merge fails.
+                                                                 * Also, we have deleted the source index and there is no way we can
+                                                                 * roll back and restart the operation. So, we should not fail the rollup
+                                                                 * operation.
+                                                                 */
+                                                                logger.error(
+                                                                    "Failed to force-merge rollup index [" + rollupIndexName + "]",
```
nik9000, line 284: :+1:

It's probably worth a test that fails here and makes sure we don't fall over. I think this will do the job but it's probably worth tracing this path. Both as a "for now" thing and to make sure that changes in the future don't cause us to lose this property.

nik9000 (approved):

PR #85712, opened 2022-04-06
update analysis-common plugin description to ensure json validity

urgerestraint:
### Context
* ES version: 7.9.1
* Issue: GETs to endpoints which mix-in plugin descriptions, such as `/_nodes`, result in an invalid JSON in Kibana response body due to the quoting in this plugin description, making cluster level integrations a little harder to test. Example:

In Kibana devtools or Kibana proxied GETs:
```
GET /_nodes

# copy json and pipe to jq, python, jsonlint etc.

parse error: Expected separator between values at line 205, column 33
```

With line 205 being the first instance of the `analysis-common` module under `.nodes.modules`:
```
        "description" : """Adds "built in" analyzers to Elasticsearch.""",
```

This issue appears due to Kibana string parsing in devtools, but this is the only instance I've seen of that output method producing invalid JSON. 

I couldn't find any other plugins which included double quotes in their descriptions.

### Changes
* Drops the heavy quoting around `"built in"`; understandable emphasis given it's not truly built in, couldn't think of a better alternative without additional verbiage.

### Tests
* `gradle check` passed locally


javanna:
@elasticmachine update branch

javanna:

javanna:
Sorry that it took us a long time to get back to you on this proposed PR.

Good catch, you did guess right where the description came from, but when I call the `_nodes` endpoint I get `Adds \"built in\" analyzers to Elasticsearch.` with escaped double quotes which seems correct. It may be that it's because I am testing on a recent version and something has recently changed.

When I test from Kibana dev tools, I get `"description": """Adds "built in" analyzers to Elasticsearch.""",` though, which is slightly different and I am not sure what is changing the response, but that is not the same response I get from curl. Also I do not see an error in Dev Tools, hence I am leaning towards closing this PR as I don't see an issue within Elasticsearch.

Please correct me if I am missed some details trying to reproduce it.

PR #85714, opened 2022-04-06
Faster ipv4 parser

idegtiarenko:
According to the `BeatsMapperBenchmark`: single ip field parsing is visible (highlighted) in the async flame graph.

image

Updating implementation with one that would not need intermediate list and string allocations:
```
Benchmark                         (ip)         Mode  Cnt    Score   Error  Units
InetAddressesBenchmark. current   192.168.0.1  avgt    5  146.564 ± 6.164  ns/op
InetAddressesBenchmark. new       192.168.0.1  avgt    5   29.605 ± 1.845  ns/op
```


original-brownbear (approved):
LGTM :) nice one

PR #85719, opened 2022-04-06
Add a highlighter unit test base class

romseygeek:
The vast majority of our highlighter tests are integration or rest tests, which exercise
the full ES stack but take a long time to run and are difficult to debug.  We have a few
unit tests but they are testing very low-level behaviour, and don't interact with the fetch
phase or hit contexts.  This commit adds a new HighlighterTestCase base class with some
helper methods that should fill the gap between these two sets of tests.  It includes a
method that takes a MapperService, ParsedDocument and SearchSourceBuilder, and then
runs the appropriate highlighter fetch subphase over the resulting hit.

### DIFF
--- a/server/src/test/java/org/elasticsearch/search/fetch/subphase/highlight/CustomUnifiedHighlighterTests.java
+++ b/server/src/test/java/org/elasticsearch/search/fetch/subphase/highlight/CustomUnifiedHighlighterTests.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.search.fetch.subphase.highlight;
+
+import org.elasticsearch.index.mapper.MapperService;
+import org.elasticsearch.index.mapper.ParsedDocument;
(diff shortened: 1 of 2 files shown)


server/src/test/java/org/elasticsearch/search/fetch/subphase/highlight/CustomUnifiedHighlighterTests.java
```diff
@@ -0,0 +1,41 @@
...
+            """);
+
+        ParsedDocument doc = mapperService.documentMapper().parse(source("""
+            { "field" : "this is some text" }
+            """));
+
+        SearchSourceBuilder search = new SearchSourceBuilder().query(QueryBuilders.termQuery("field", "some"))
+            .highlighter(new HighlightBuilder().field("field"));
+
+        Map<String, HighlightField> highlights = highlight(mapperService, doc, search);
```
mayya-sharipova, line 37: Is it worth to have even higher level abstraction, and provide a mapping and a document as Strings to `highlight` method, and `highlight` method itself inside create `MapperService` and `ParsedDocument`?

mayya-sharipova (approved):
@romseygeek Thanks Alan, this looks very nice and simplify our highlight tests a lot.

server/src/test/java/org/elasticsearch/search/fetch/subphase/highlight/CustomUnifiedHighlighterTests.java
romseygeek: I did think about doing things that way, but I think this is slightly more flexible - it allows you to build mappings from XContent if you want to build something programmatically, for example.  If it turns out that we're almost always using this pattern we can always add a sugar method that just takes strings.

PR #85720, opened 2022-04-06
Allow bucket paths to specify _count within a bucket

### LINKED ISSUE #76320: `_count` buckets_path syntax doesn't work with filters aggregation
When attempting to get `_count` via `bucket_selector.buckets_path`, I get a path validation error:

`"Invalid path element [filter_thing['phase_create']._count] in path [filter_thing['phase_create']._count]"`

It seems to me that since `filters` is a multi-bucket aggregation, and we are selecting a single bucket, accessing `_count` should be valid.

Here is a minimal repeatable code set:
```
PUT test_index
{
  "mappings": {
    "properties": {
      "special_id": {
        "type": "long"
      },
      "phase": {
        "type": "keyword"
      }
    }
  }
}

POST test_index/_bulk?refresh=true
{"index":{}}
{"special_id":1,"phase":"create"}
{"index":{}}
{"special_id":1,"phase":"update"}
{"index":{}}
{"special_id":2,"phase":"update"}
{"index":{}}
{"special_id":3,"phase":"create"}
{"index":{}}
{"special_id":3,"phase":"update"}


POST test_index/_search
{
  "size": 0,
  "aggs": {
    "All_IDs": {
      "terms": {
        "field": "special_id"
      },
      "aggs": {
        "filter_thing": {
          "filters": {
            "filters": {
              "phase_create": {
                "match": {
                  "phase": "create"
                }
              }
            }
          }
        },
        "select": {
          "bucket_selector": {
            "buckets_path": {
              "c": "filter_thing['phase_create']._count"
            },
            "script": "params.c==0"
          }
        }
      }
    }
  }
}
```

benwtrent:
Users should be able to specify specific metrics/keys within a specific bucket key. 

An example is `agg["bucket_foo"]._count`. 

This change now allows that.

closes: https://github.com/elastic/elasticsearch/issues/76320


not-napoleon (approved):
That path parser is a house of cards, and we should rewrite the whole thing to be more sensible, but that's out of scope for this I think.

PR #85726, opened 2022-04-06
Add indices permissions to Enterprise Search service account

carlosdelest:
Enterprise Search can create a new Engine type that uses already existing Elasticsearch indices.

Elasticsearch indices starting with `search-*` (or with a `search-*` alias) can be used as the backend for these new Engine types. The suffix has been chosen so users need to create an alias or rename their indices to be available to Enterprise Search; this way, we avoid Enterprise Search to have full read ability over all indices.

For this to happen, permissions must be added to the Enterprise Search service account so it can read `search-*` indices.

Fixes: https://github.com/elastic/enterprise-search-team/issues/1773
Related: https://github.com/elastic/cloud-assets/pull/1004, https://github.com/elastic/cloud/pull/100471


ioanatia (approved):

ywangd (approved):
LGTM

PR #85729, opened 2022-04-06
Update bucket metric pipeline agg paths to allow intermediate single bucket and bucket qualified multi-bucket aggs

benwtrent:
Bucket metric pipeline aggregation paths currently always expect the sibling aggregation to be a multi-bucket agg. 

This honestly doesn't have to be the case for bucket metric pipeline aggs to work. 

Consider the following path:

```
filter_agg>filters_agg['bucket_foo']>histo>some_metric_agg
```

Since `filter_agg>filters_agg['bucket_foo']` are well defined and are not crossing bucket threasholds, metrics should still be able to be calculated against the bucket values for `histo`

This commit allows any combination of single bucket aggs (e.g. filter) and bucket specific multi-bucket aggs before reaching the desired multi-bucket used for the metric calculation.


not-napoleon (commented):
I need to think about this one a bit more, will take another pass at it tomorrow.  Here's some initial feedback.

not-napoleon (approved):
LGTM.  Thanks.

benwtrent:
@elasticmachine update branch

benwtrent:
@elasticmachine update branch

benwtrent:
@elasticmachine update branch

PR #85732, opened 2022-04-06
Use Faster API for Parsing Maps from XContent

original-brownbear:
We can be a little more efficient when parsing maps and exploit
the fact that we know the next token is a name in a couple of cases.
I fixed the most performance relevant one but there's a couple more
that could make use of this API in a follow up.

The `nextFieldName` call itself doesn't seem to do too much for us in benchmarking, but saving the `nextToken()` call is quite helpful in making things inline better because it saves the huge switch we have when converting from Jackson to our own tokens.


rjernst (approved):
LGTM

original-brownbear:
Thanks Ryan!

PR #85734, opened 2022-04-07
Simplify stream concatenation

idegtiarenko:
This change simplifies the Stream concatenation. It replaces:
* reduce/concat for 2 arguments usage with a simple concat
* reduce/concat for multiple streams with a flatMap.


server/src/main/java/org/elasticsearch/index/mapper/MappingLookup.java
```diff
@@ -291,15 +291,12 @@ private void checkObjectDepthLimit(long limit) {
...
-        Stream.of(objectMappers.values().stream(), fieldMappers.values().stream())
-            .reduce(Stream::concat)
-            .orElseGet(Stream::empty)
-            .forEach(mapper -> {
-                String name = mapper.simpleName();
-                if (name.length() > limit) {
-                    throw new IllegalArgumentException("Field name [" + name + "] is longer than the limit of [" + limit + "] characters");
-                }
-            });
+        Stream.concat(objectMappers.values().stream(), fieldMappers.values().stream()).forEach(mapper -> {
```
original-brownbear, line 294: Could we just solve this with a normal loop (or two)? The stream stuff is just needless overhead here I think.

server/src/main/java/org/elasticsearch/common/util/iterable/Iterables.java
```diff
@@ -35,11 +35,7 @@
 
         @Override
         public Iterator<T> iterator() {
-            return Stream.of(inputs)
-                .map(it -> StreamSupport.stream(it.spliterator(), false))
-                .reduce(Stream::concat)
-                .orElseGet(Stream::empty)
-                .iterator();
+            return Stream.of(inputs).flatMap(it -> StreamSupport.stream(it.spliterator(), false)).iterator();
```
original-brownbear, line 38: I looked into the uses of this method (and concatenating iterables in general). It appears there's literally only 3 spots and we already have concatenated iterators. Maybe we could just do away with the whole concept of concatenated iterable outright and replace the 3 use cases with either concatenating iterators or just doing some other simple thing? :)

original-brownbear (commented):
Looks better than what we have, but I left some comments wondering if we can go for even simpler here. For the mapping lookup checks in particular, simpler and faster would make the many-shards benchmarks happier :)

server/src/main/java/org/elasticsearch/index/mapper/MappingLookup.java
idegtiarenko: Yeah, good point

server/src/main/java/org/elasticsearch/common/util/iterable/Iterables.java
idegtiarenko: Correct. I would like to merge this as is (simplify concatenation in various places) and then open a followup pr to remove/replace operations from `org.elasticsearch.common.util.iterable.Iterables` that are no longer needed.

original-brownbear (approved):

PR #85740, opened 2022-04-07
Relax docs about using new repos across major versions

DaveCTurner:
Today the docs say that new major versions need new snapshot
repositories or else corruption may occur. This isn't true, we support
using the same repository across upgrades.


henningandersen (approved):
LGTM.

DaveCTurner:
Going to add this to 7.17 too to avoid any doubt that this only applies to 8.x clusters.

Issue #81626, opened 2021-12-10
Pending task batching can be a bottleneck


DaveCTurner:
**Elasticsearch version** (`bin/elasticsearch --version`): 7.15.1 (likely others, this mechanism has existed for quite a while)

**Plugins installed**: []

**JVM version** (`java -version`): Bundled

**OS version** (`uname -a` if on a Unix-like system): ESS

**Description of the problem including expected versus actual behavior**:

I saw a cluster with symptoms of stuck transport threads, in which most of the transport threads had the following stack trace:

```
       app//org.elasticsearch.cluster.service.TaskBatcher.submitTasks(TaskBatcher.java:59)
       app//org.elasticsearch.cluster.service.MasterService.submitStateUpdateTasks(MasterService.java:802)
       app//org.elasticsearch.cluster.service.ClusterService.submitStateUpdateTasks(ClusterService.java:271)
       app//org.elasticsearch.cluster.service.ClusterService.submitStateUpdateTask(ClusterService.java:252)
       app//org.elasticsearch.cluster.action.shard.ShardStateAction$ShardFailedTransportHandler.messageReceived(ShardStateAction.java:271)
       app//org.elasticsearch.cluster.action.shard.ShardStateAction$ShardFailedTransportHandler.messageReceived(ShardStateAction.java:255)
```

This cluster had been unhealthy for other reasons for an extended period of time and the nodes have apparently accumulated a rather large collection of unprocessed `shard-failed` tasks. When the new master is elected, these tasks are all retried individually which forms up into a thundering herd that block on the `tasksPerBatchingKey` mutex and keep the transport workers from doing more meaningful things. Consequently the nodes start to fail their follower checks and drop out of the cluster, ultimately causing the master to stand down. Another master is elected and the cycle begins again.

I see room for improvement in a couple of areas. Firstly, I think we should be batching the shard-failed requests - at least the retries - on the data nodes. Secondly, I think we should streamline the batching of tasks on the master. Rather than submitting every single task to a `PrioritizedEsThreadPoolExecutor` and having to maintain a separate map of the batches I think we should have a queue of the `ClusterStateTaskExecutor` things each of which maintains its own queue to which we can add tasks without blocking on a mutex.

Relates #77466

PR #85751, opened 2022-04-07
Radix-sorted, lookup-free, typesafe, nonblocking, priority-boosting pending tasks

DaveCTurner:
Today the master's pending task queue is just the
`PriorityBlockingQueue<Runnable>` belonging to the underlying
`ThreadPoolExecutor`. The reasons for this date back a long way but it
doesn't really reflect the structure of the queue as it exists today. In
particular, we must keep track of batches independently of the queue
itself, and must do various bits of unchecked casting to process
multiple items of the same type at once.

This commit introduces an new queueing mechanism, independent of the
executor's queue, which better represents the conceptual structure of
the master's pending tasks:

* Today we use a priority queue to allow important tasks to preempt
less-important ones. However there are only a small number of priority
levels, so it is simpler to maintain a queue for each priority,
effectively replacing the sorting within the priority queue with a
radix sort.

* Today when a task is submitted we perform a map lookup to see if it
can be added to an existing batch or not. With this change we allow
client code to create its own dedicated queue of tasks. The entries in
the per-priority-level queues are themselves queues, one for each
executor, representing the batches to be run.

* Today each task in the queue holds a reference to its executor, but
the executor used to run a task may belong to a different task in the
same batch. In practice we know they're the same executor (that's how
batches are defined) but we cannot express this knowledge in the type
system so we have to do a bunch of unchecked casting to work around it.
With this change we associate each per-executor queue directly with its
executor, avoiding the need to do all this unchecked casting.

* Today the master service must block its thread while waiting for each
task to complete, because otherwise the executor would start to process
the next task in the queue. This makes testing using a
`DeterministicTaskQueue` harder (see `FakeThreadPoolMasterService`).
With this change we avoid submitting a task to the `ThreadPoolExecutor`
until the previous task is complete, which means we can make the
implementation avoid blocking while a task is running and therefore run
the whole production implementation in deterministic tests[^1].

* Today it's possible for a steady drip of high-priority tasks to starve
low-priority tasks of access to the master for an extended period of
time. With this change we separate the queue of tasks from the queue
which determines the execution order. This allows us to implement more
intelligent execution policies. For instance, if we detect that the
queue has not processed any low-priority tasks for too long then we can
make the decision to boost their priorities[^2].

Relates #81626

[^1]: Not done yet but this is a step in the right direction.

[^2]: Not done yet but this is a step in the right direction.


DaveCTurner:

DaveCTurner:
Superseded by #92021

PR #85758, opened 2022-04-07
Replace most shell script logic with Java

rjernst:
Elasticsearch provides several command line tools, as well as the main script to start elasticsearch. While most of the logic is abstracted away for cli tools, the main elasticsearch script has hundreds of lines of platform specific shell code. That code is difficult to maintain because it uses many special shell features which then must also exist in other platforms (ie windows batch files). Additionally, the logic in these scripts are not easy to test, we must be on the actual platform and test with a full installation of Elasticsearch, which is relatively slow (compared to most in process tests).

This commit replaces most of the shell specific logic with Java code. It introduces a singular entrypoint, the Launcher, to start any Elasticsearch CLI. Each shell script must then only describe the tool to call and the lib directories it needs to load.

There is a small amount of shell logic that remains. Specifically, that is to identify the location of ES_HOME from the shell script path, and then to find which java installation should be used. After that, the cli can be launched, using a small heap (as we do already for CLIs). For the main Elasticsearch server, the cli figures out all the jvm options and such necessary, then launches the real server process. If run in the foreground, the launcher will stay alive for the lifetime of Elasticsearch; the streams are effectively inherited so all output from Elasticsearch still goes to the console. If daemonizing, the launcher waits around until Elasticsearch is "ready" (this means the Node startup completed), then detaches and exits.


rjernst:
Note: This PR is very much a draft. It shows the idea is possible, and there are a lot of advantages to it, like easier maintenance of the logic, as well as future potential (for example, making native calls from java to the Windows API to manage our own service, thus replacing Apache procrun). It's about +500 lines at the moment, but so much of that is normal java boilerplate. Additionally, many of the changes here can go into master directly, so I plan to open additional PRs so that this PR becomes more and more reviewable.

rjernst:
@elasticmachine run elasticsearch-ci/packaging-tests-unix

rjernst:
This PR is now ready for review.

A note to reviewers: the total diff looks large, +2k lines. However, there are basically 2k work of new tests that were not possible before, so the shell scripts vs the new code is net zero. The main classes to look at are ServerProcess, which handles starting and interacting with the jvm process, and the changes to Elasticsearch.java which handling sending and receiving signals on stderr/stdin. The cli specific logic is mostly moved to ServerCli from Elasticsearch, and the tests for the cli options were retained and moved to ServerCliTests.

server/src/main/java/org/elasticsearch/bootstrap/Elasticsearch.java
```diff
@@ -120,6 +145,29 @@ static void printLogsSuggestion() {
...
+     *
+     * @param stdin Standard input for this process
+     */
+    private static void startCliMonitorThread(InputStream stdin) {
+        new Thread(() -> {
+            int msg = -1;
+            try {
+                msg = stdin.read();
+            } catch (IOException e) {}
+            if (msg == BootstrapInfo.SERVER_SHUTDOWN_MARKER) {
```
grcevski, line 162: Maybe we should also put this in finally?

grcevski (commented):
This looks great, the tests we can write now! I left some questions/comments. We should probably add a test-windows label to the PR.

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ErrorPumpThread.java
```diff
@@ -0,0 +1,104 @@
...
+                } else if (line.isEmpty() == false && line.charAt(0) == SERVER_READY_MARKER) {
+                    ready = true;
+                    readyOrDead.countDown();
+                } else {
+                    writer.println(line);
+                }
+            }
+        } catch (IOException e) {
+            ioFailure = e;
+        }
```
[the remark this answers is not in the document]
mark-vieira: 👍 I agree. We should do both in a finally block in case we encounter some error reading from the input stream.

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ServerProcess.java
```diff
@@ -0,0 +1,259 @@
...
+ */
+public class ServerProcess {
+
+    // the actual java process of the server
+    private final Process jvmProcess;
+
+    // the thread pumping stderr watching for state change messages
+    private final ErrorPumpThread errorPump;
+
+    // a flag marking whether the java process has been detached from
```
mark-vieira, line 59: This comment reads like it's been truncated.

distribution/src/bin/elasticsearch
```diff
@@ -1,139 +1,5 @@
...
-  if ! ps -p $pid > /dev/null ; then
-    exit 1
-  fi
-  exit 0
-fi
-
-exit $?
+CLI_NAME=server
+CLI_LIBS=lib/tools/server-cli
+source "`dirname "$0"`"/elasticsearch-cli
```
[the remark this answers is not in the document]
grcevski: Yeah, I was thinking about it. If most of our cli tools are generally short running we should use it by default.

mark-vieira (approved):

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ServerProcess.java
rjernst: Just poor grammar on my part. I've reworded.

rjernst:
@elasticmachine run elasticsearch-ci/packaging-tests-windows

rjernst:

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ServerProcess.java
```diff
@@ -0,0 +1,266 @@
...
+        Path esHome = processInfo.workingDir();
+        Path javaHome = PathUtils.get(processInfo.sysprops().get("java.home"));
+        List<String> command = new ArrayList<>();
+        boolean isWindows = processInfo.sysprops().get("os.name").startsWith("Windows");
+        command.add(javaHome.resolve("bin").resolve("java" + (isWindows ? ".exe" : "")).toString());
+        command.addAll(jvmOptions);
+        command.add("-cp");
+        // The '*' isn't allowed by the windows filesystem, so we need to force it into the classpath after converting to a string.
+        // Thankfully this will all go away when switching to modules, which take the directory instead of a glob.
+        command.add(esHome.resolve("lib") + (isWindows ? "\\" : "/") + "*");
```
ChrisHegarty, line 220: Trivially, this could simplified to use java.io.File.separator

ChrisHegarty (approved):
LVGTM

distribution/src/bin/elasticsearch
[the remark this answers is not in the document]
grcevski: Good point, I'll take it upon myself to get some data on CPU/memory while running without the optimizing compiler and maybe if it's beneficial to a great extent, I'll follow up with a PR.

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ServerProcess.java
rjernst: I don't think we can! We specifically use the local isWindows here so that we can check the unix and linux behavior in tests, regardless of which platform we are actually on. See `ServerProcessTests.testCommandLine`

grcevski (approved):
LGTM!

PR #85761, opened 2022-04-08
Remove legacy versioned logic for DefaultSystemMemoryInfo

rjernst:
The machine dependent heap logic uses SystemMemoryInfo to determine how
much total memory is available on the system. The default implementation
delegates to Java's MX bean to get OS stats. In the past this was only
available in Java 14, but since the main branch is now on Java 17, we do
not need this logic.

This commit refactors the default implementation to no longer need the
version check, and to also remove unnecessary forbidden api suppressions
for the entire class, instead using a LongSupplier for the memory to
isolate the platform specific bean references.

relates #85758


distribution/tools/launchers/src/main/java/org/elasticsearch/tools/launchers/DefaultSystemMemoryInfo.java
```diff
@@ -10,32 +10,29 @@
...
-        }
+    @SuppressForbidden(reason = "Using com.sun internals is the only way to query total system memory")
+    private static LongSupplier getOSBeanMemoryGetter() {
+        OperatingSystemMXBean bean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
+        return bean::getTotalMemorySize;
+    }
 
-        return operatingSystemMXBean.getTotalPhysicalMemorySize();
+    @Override
+    public long availableSystemMemory() {
```
mark-vieira, line 35: Since we've removed `SystemMemoryInfoException` from the signature here we should probably remove it everywhere else as well (catch statements, `SystemMemoryInfo` interface, etc). As far as I can tell this was the only place we ever threw such an exception, unless we want to keep this in place for furture implementations?

mark-vieira (changes requested):

distribution/tools/launchers/src/main/java/org/elasticsearch/tools/launchers/DefaultSystemMemoryInfo.java
rjernst: I've removed `SystemMemoryInfoException`

rjernst:

rjernst:
@mark-vieira This is ready for another review.

mark-vieira (approved):
👍

PR #85762, opened 2022-04-08
Remove pre Java 17 system jvm options

rjernst:
This commit removes the one conditional jvm option that was dependent on
the Java version priority 17, which was whether to add
`-XX:+ShowCodeDetailsInExceptionMessages`. This option is available
since Java 14, so the conditional is no longer needed in the main
branch.

relates #85758


grcevski (approved):
LGTM!

Since JDK 15 this is also on by default, so perhaps removing the command line might even be better.

https://github.com/openjdk/jdk/blob/7edd18612155b8e66f214a17e65ef775b470955e/src/hotspot/share/runtime/globals.hpp#L642

rjernst:
> Since JDK 15 this is also on by default

Great! I did not realize this. I've removed the flag altogether.

PR #85763, opened 2022-04-08
Remove leftover meta plugin descriptor

rjernst:
This commit removes a legacy file that was necessary when xpack was made
up of multiple plugins. The meta plugin functionality was removed many
years ago.


grcevski (approved):
LGTM!

PR #85764, opened 2022-04-08
Remove xshare shell logic

rjernst:
When adding aarch64 support to Elasticsearch, a bug existed in the jdk
which caused an error message to be printed on every invocation of Java
due to page size misalignment (see
https://bugs.openjdk.java.net/browse/JDK-8236847). This bug was fixed in
Java 17. Since the main branch always uses Java 17, the workaround is no
longer necessary.

relates #85758


rjernst:
Note: I confirmed the CDS error message no longer appears when invoking `java -version` on a t4g graviton instance of Ubuntu using openjdk 17.

mark-vieira (approved):
We also set `-Xshare:off` by default [here](https://github.com/elastic/elasticsearch/blob/01c5bc04e83ef0ce2bda79d7e7f41f46fa9f4070/distribution/tools/launchers/src/main/java/org/elasticsearch/tools/launchers/JvmOption.java#L98), I suspect this is no longer necessary?

PR #76896, opened 2021-08-24
Remove no-jdk distributions

mark-vieira:
Starting with Elasticsearch 8.0 we will no longer build separate "no-jdk" distributions. All distributions now include by default a bundled JDK and users can configure an alternative JDK distribution via `ES_JAVA_HOME`.

I've left some references to no-jdk distribution handling code to make future backports easier but all the build scripts under `distribution/` have had conditional jdk logic stripped out. Additionally,  there was a lot of lingering code related to OSS distributions which no longer exist which has been removed as well and will be separately backported to `7.x`.

Closes https://github.com/elastic/elasticsearch/issues/65109

PR #85765, opened 2022-04-08
Remove no-jdk deprecations

rjernst:
The no-jdk distributions exist in 7.x and before. They were removed with
8.0. This commit removes the remaining deprecation messages for using
the no-jdk distribution. Note that when talking with an older node, we
drop the bundledJdk attribute. This is ok because it is only possible
for this to not be true when talking with a 7.17 node, during an upgrade,
and the usingBundledJdk is retained, which is the important thing if
debugging a problem.

relates #76896
relates #85758


docs/reference/cluster/stats.asciidoc
```diff
@@ -1440,7 +1436,6 @@ The API returns the following response:
                "vm_name": "OpenJDK 64-Bit Server VM",
                "vm_version": "12+33",
                "vm_vendor": "Oracle Corporation",
-               "bundled_jdk": true,
```
mark-vieira, line 1443: Is removing this item from this API considered a breaking change?

mark-vieira (approved):
:+1:

docs/reference/cluster/stats.asciidoc
rjernst: I had thought the cluster stats API was not considered covered under our backcompat, but it appears it is in a bit of a gray area. I've added it back here, just always returning `true`.

PR #85766, opened 2022-04-08
Only check java version with custom ES_JAVA_HOME

rjernst:
The java version checker is intended to ensure the version of the JDK
that Elasticsearch will use is at least the version Elasticsearch was
compiled against. It prevents cryptic class version errors upon loading.
The bundled jdk is always of the appropriate version.

This commit moves the version check to only occur when a custom
ES_JAVA_HOME is used. If a user were to swap out the bundled jdk with
their own, it could result in the cryptic errors being protected
against, but replacing the bundled jdk is not something we need nice
error messages for.

relates #85758


mark-vieira (commented):
Why bother with this change if the intention is fold the version check into the launcher anyway. In that scenario we'd _always_ be doing this check, right?

rjernst:
My original idea was to fold this in. However, starting in java 7 compatible means we can never start with modules, so it would complicate that later on. I instead realized that we really only need to do this check for custom JDKs, thus this alternative to what we discussed before.

mark-vieira (approved):
:+1:

PR #85767, opened 2022-04-08
Use multi-release-jar for java version checker

rjernst:
The java version checker tests whether the current jdk is a sufficient
version to run Elasticsearch. Since it is compatible with java 7, it
must parse Java's version sysprop.

This commit reworks the checker to no longer need to parse the version.
Instead, it uses a multi-release-jar, so that there are two versions of
the main class. The Java 7 compatible version simply exits with an error
message. The Java 17 version simple exits with no error.

relates #85758


distribution/tools/java-version-checker/build.gradle
```diff
@@ -1,12 +1,27 @@
-apply plugin: 'elasticsearch.build'
+apply plugin: 'elasticsearch.java'
 
-targetCompatibility = JavaVersion.VERSION_1_7
+sourceSets {
+  unsupportedJdkVersionEntrypoint
+}
+
+tasks.named('compileUnsupportedJdkVersionEntrypointJava').configure {
```
mark-vieira, line 7: FYI, we can use `sourceSets.unsupportedJdkVersionEntrypoint.compileJavaTaskName` here.

distribution/tools/java-version-checker/build.gradle
```diff
@@ -1,12 +1,27 @@
...
+  manifest {
+    attributes("Multi-Release": "true")
+  }
 
-// java_version_checker do not depend on core so only JDK signatures should be checked
-tasks.named('forbiddenApisMain').configure {
-  replaceSignatureFiles 'jdk-signatures'
+  from(sourceSets.unsupportedJdkVersionEntrypoint.output)
+  eachFile { details ->
+    if (details.path.equals("org/elasticsearch/tools/java_version_checker/JavaVersionChecker.class") &&
```
mark-vieira, line 18: I don't know why but I'm triggered by underscores in a package name :smile:

distribution/tools/java-version-checker/src/main/java/org/elasticsearch/tools/java_version_checker/JavaVersionChecker.java
```diff
@@ -9,55 +9,18 @@
 package org.elasticsearch.tools.java_version_checker;
 
 import java.util.Arrays;
-import java.util.Locale;
 
 /**
- * Simple program that checks if the runtime Java version is at least 17.
+ * Java 17 compatible main which exits with an error.
```
mark-vieira, line 14: I find this confusing. The whole point of this is to _not_ exit with an error, right?

distribution/tools/java-version-checker/src/unsupportedJdkVersionEntrypoint/java/org/elasticsearch/tools/java_version_checker/JavaVersionChecker.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Java 7 compatible main which exits with an error.
+ */
+final class JavaVersionChecker {
+
+    private JavaVersionChecker() {}
+
+    public static void main(final String[] args) {
+        // no leniency!
+        if (args.length != 0) {
```
mark-vieira, line 23: Should we even bother checking the arguments here?

mark-vieira (approved):
Couple of comments. Just so I recall, the whole point of this is that this is going to get folded into a singular "launcher" right that will compile against Java 17, yes?

distribution/tools/java-version-checker/build.gradle
rjernst: I agree, I'm happy to change this, it was just already existing.

mark-vieira: We can definitely leave it as-is.

distribution/tools/java-version-checker/src/main/java/org/elasticsearch/tools/java_version_checker/JavaVersionChecker.java
rjernst: Oops, this was a copy/paste problem.

distribution/tools/java-version-checker/src/unsupportedJdkVersionEntrypoint/java/org/elasticsearch/tools/java_version_checker/JavaVersionChecker.java
rjernst: I just left this like in the old implementation. No leniency!

PR #85768, opened 2022-04-09
[Spacetime] Macaroon access tokens

albertzaharovits:
### Macaroons

A form of bearer credentials (as opposed to proof-of-possesion credentials).
Macaroons are an instance of what has been termed claims-based, proof-carrying or simply credentials-based authorization.
The claims are stored client-side, in the token itself, as opposed to server-side in a database entry referred to by the token. The client-side storage requires that the macaroon token be signed with a secret key. The signing key must be shared among the systems that verify and generate the token.

Macaroons allow for attenuation, via caveats, which are conditions further restricting the granted access of the token. The restrictions can apply to the object as well as to the subject. The subject restrictions take the form of third-party attenuation, an interesting concept, but which has not been explored in this POC.

The trully interesting thing is that the attenuation does not require calling any API endpoint. Anyone can attenuate the token they're in possesion of, before sharing it further, without knowing the signing key which is only required when the token was initially generated. Of course no one is able to waive attenuations.
The level of granularity of the access restrictions from caveats is implementation dependent (the server-side implementation of the caveat verifier), but in any case should be much higher than the typical oauth2/OIDC scopes.

For further reading on the topic, refer to:

https://research.google/pubs/pub41892/
https://github.com/rescrv/libmacaroons/blob/master/README
https://evancordell.com/posts/macaroons-101-contextual-confinement/
https://neilmadden.blog/2020/09/09/macaroon-access-tokens-for-oauth-part-2-transactional-auth/

### POC ES implementation:

In this POC, all the access tokens that ES generates, automatically, without any other configuration, are macaroons containing access tokens. There is no need to configure a shared key between the cluster nodes because the POC reuses some already existing code in the ES token service that uses the cluster state to share a secret (auto-generated) key.

For this POC, three kinds of caveats (restrictions) are implemented:
 * data read-only
 * specific PIT
 * before time instant

#### Goal
A goal of this POC is to facilitate the sharing of point-in-time references.
Today, users can open PITs, and can thereafter share them with other users. The users the PIT is shared with must themselves have the necessary credentials granting them access to the underlying indices that the PIT refers to.

Macaroons improve this experience by allowing the user that opened the PIT to be able to create a short term credential 
that is scoped to only allow access to the PIT. The credential is the attenuated version of the user's macaroon access token. The user can then share the credential together with the PIT, such that the receiving user doesn't have to have any credentials of its own (like sharing a read-only link to a google doc with anyone, without requiring the user be part of the google workspace organization).

#### Example flow

1. enable the token service by appending the following in `config/elasticsearch.yml`
```
xpack.license.self_generated.type: "trial"
```
2. generate a macaroon access token with the usual token API
```
curl -k -u "elastic:$ELASTIC_PASSWORD" -X POST "https://localhost:9200/_security/oauth2/token?pretty" -H 'Content-Type: application/json' -d'
{
  "grant_type" : "client_credentials"
}
'
```
3. open (create) a PIT for any index
```
curl -k -u "elastic:$ELASTIC_PASSWORD" -X PUT "https://localhost:9200/twitter/_doc/1?" -H 'Content-Type: application/json' -d'
{
  "category": "click", "tag": "one"
}
'
curl -k -u "elastic:$ELASTIC_PASSWORD" -X POST "https://localhost:9200/twitter/_pit?keep_alive=999m&pretty"
```
4. locally, without calling an ES API, restrict the access token from `2` to the PIT from `3`
```
jshell --class-path "$(pwd)/lib/*:$(pwd)/modules/x-pack-security/*"

import org.elasticsearch.xpack.security.authz.accesscontrol.MacaroonUtils

MacaroonUtils.attenuateTokenForPITOnly("<the access token from the response in 2>", "<the PIT from the response in 3>")
```
5. the returned attenuated token can be further restricted in duration with `MacaroonUtils#attenuateTokenInDuration` and be shared, together with the PIT itself, so anyone (a user or a role are not required) can use the macaroon token to only query the PIT, nothing else.


albertzaharovits:
@elasticmachine test elasticsearch-ci/eql-correctness

PR #85769, opened 2022-04-09
Make KeywordFieldMapper slighty faster

original-brownbear:
Extracting the cold code paths from the keyword field mapper to make it inline
a little better in some cases. 
Also, avoid one costly `currentToken()` call by using the same pattern for xcontent parsing
we use in e.g. the text field mapper and not manually check the curent token (we do that
in the parser itself anyway).
This is slightly (but measurably) faster in the
benchmarking we did on this during spacetime.

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java
@@ -57,7 +57,6 @@ import org.elasticsearch.search.runtime.StringScriptFieldPrefixQuery;
 import org.elasticsearch.search.runtime.StringScriptFieldRegexpQuery;
 import org.elasticsearch.search.runtime.StringScriptFieldTermQuery;
 import org.elasticsearch.search.runtime.StringScriptFieldWildcardQuery;
-import org.elasticsearch.xcontent.XContentParser;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
@@ -66,7 +65,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.function.Supplier;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -897,14 +886,17 @@ private void indexValue(DocumentParserContext context, String value) {
...
 
+        final String name = fieldType().name();
         if (value.length() > ignoreAbove) {
-            context.addIgnoredField(name());
+            context.addIgnoredField(name);
             return;
         }
 
-        value = normalizeValue(fieldType().normalizer(), name(), value);
+        if (normalizer != null) {
```
henningandersen, line 895: It seems like we are optimizing either:

1. A constant vs null check
2. A method call that gets a final field.
3. The method call to normalizeValue when normalizer is keyword.

I am not sure either are important. Can you elaborate on the intention here? I fear such optimizations will easily fall off.

server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -915,54 +907,47 @@ private void indexValue(DocumentParserContext context, String value) {
...
-                + "\" (whose "
-                + "UTF8 encoding is longer than the max length "
-                + MAX_TERM_LENGTH
-                + "), all of which were "
-                + "skipped. Please correct the analyzer to not produce such terms. The prefix of the first immense "
-                + "term is: '"
-                + Arrays.toString(prefix)
-                + "...'";
-            throw new IllegalArgumentException(msg);
+            throwOnMaxTermSize(name, binaryValue);
```
henningandersen, line 910: Can we instead keep the throw here but calculate the message in a message? It is also sort of annoying that this is necessary to help java, are we sure this helps enough, i.e., do you have rally numbers for it?

henningandersen (commented):
> a little better in some cases.

Can you elaborate on this, perhaps produce a rally comparison?

It is quite detailed changes and I worry that such optimizations fall off as we continue development.

Issue #85721, opened 2022-04-06
[CI] ReadinessClusterIT testReadinessDuringRestarts failing


martijnvg:
**Build scan:**
https://gradle-enterprise.elastic.co/s/k26mmnltxc7yw/tests/:server:internalClusterTest/org.elasticsearch.readiness.ReadinessClusterIT/testReadinessDuringRestarts

**Reproduction line:**
`./gradlew ':server:internalClusterTest' --tests "org.elasticsearch.readiness.ReadinessClusterIT.testReadinessDuringRestarts" -Dtests.seed=66FCB40B017BCF62 -Dtests.locale=id-ID -Dtests.timezone=Asia/Barnaul -Druntime.java=17`

**Applicable branches:**
master, 8.2

**Reproduces locally?:**
No

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.readiness.ReadinessClusterIT&tests.test=testReadinessDuringRestarts

**Failure excerpt:**
```
com.carrotsearch.randomizedtesting.UncaughtExceptionError: Captured an uncaught exception in thread: Thread[id=1731, name=Thread-10, state=RUNNABLE, group=TGRP-ReadinessClusterIT]

  at __randomizedtesting.SeedInfo.seed([66FCB40B017BCF62:7374F15AB6D0E16D]:0)

  Caused by: java.lang.AssertionError: java.lang.IllegalArgumentException: remote node [{node_t2}{BAmYUvGFRxqIhcJ3aQ82aw}{4NMnoo4vSgGEBKSvT_RfnQ}{127.0.0.1}{127.0.0.1:45408}{cdfhilrstw}] is build [1ed69c17444b90d524fe6a85e758f75e0fa148a6] of version [8.2.0] but this node is build [unknown] of version [8.2.0] which has an incompatible wire format

    at __randomizedtesting.SeedInfo.seed([66FCB40B017BCF62]:0)
    at org.elasticsearch.transport.InboundHandler.handleResponse(InboundHandler.java:346)
    at org.elasticsearch.transport.InboundHandler.messageReceived(InboundHandler.java:142)
    at org.elasticsearch.transport.InboundHandler.inboundMessage(InboundHandler.java:94)
    at org.elasticsearch.transport.TcpTransport.inboundMessage(TcpTransport.java:795)
    at org.elasticsearch.transport.InboundPipeline.forwardFragments(InboundPipeline.java:149)
[...]

Issue #83034, opened 2022-01-25
[CI] NodeTests failing with incompatible wire format error


cbuescher:
Could not reproduce, haven't seen any recent failures so far either so keeping this unmuted for now.

**Build scan:**
https://gradle-enterprise.elastic.co/s/auabxhkpzbxn2/tests/:server:test/org.elasticsearch.node.NodeTests/testCloseOnInterruptibleTask

**Reproduction line:**
`./gradlew ':server:test' --tests "org.elasticsearch.node.NodeTests.testCloseOnInterruptibleTask" -Dtests.seed=3E36C1C728F289EA -Dtests.locale=pt-PT -Dtests.timezone=America/Ensenada -Druntime.java=17`

**Applicable branches:**
master

**Reproduces locally?:**
No

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.node.NodeTests&tests.test=testCloseOnInterruptibleTask

**Failure excerpt:**
```
com.carrotsearch.randomizedtesting.UncaughtExceptionError: Captured an uncaught exception in thread: Thread[id=851, name=Thread-64, state=RUNNABLE, group=TGRP-NodeTests]


  Caused by: java.lang.AssertionError: java.lang.IllegalArgumentException: remote node [{TS_B}{TS_B}{HlNEkIWGTwKCjy--a1FF3w}{127.0.0.1}{127.0.0.1:9302}] is build [unknown-modified] of version [8.1.0] but this node is build [unknown] of version [8.1.0] which has an incompatible wire format

    at __randomizedtesting.SeedInfo.seed([3E36C1C728F289EA]:0)
    at org.elasticsearch.transport.InboundHandler.handleResponse(InboundHandler.java:346)
    at org.elasticsearch.transport.InboundHandler.messageReceived(InboundHandler.java:142)
    at org.elasticsearch.transport.InboundHandler.inboundMessage(InboundHandler.java:94)
    at org.elasticsearch.transport.TcpTransport.inboundMessage(TcpTransport.java:796)
    at org.elasticsearch.transport.InboundPipeline.forwardFragments(InboundPipeline.java:149)
    at org.elasticsearch.transport.InboundPipeline.doHandleBytes(InboundPipeline.java:121)
    at org.elasticsearch.transport.InboundPipeline.handleBytes(InboundPipeline.java:86)
[...]

PR #85773, opened 2022-04-11
Log node identity at startup

DaveCTurner:
At startup we construct a `DiscoveryNode` which identifies the local
node, and we report various components of this identity in the startup
logs. You can often deduce other parts (e.g. the ephemeral and
persistent node IDs) from other logs, but this is hard to do reliably.
With this commit we report the node identity in full at startup to avoid
any ambiguity.

Relates #83034, #85721, and other test failures explainable by mistaken
identities.

### DIFF
--- a/server/src/main/java/org/elasticsearch/node/Node.java
+++ b/server/src/main/java/org/elasticsearch/node/Node.java
@@ -1267,7 +1267,7 @@ public class Node implements Closeable {
             }
         }
 
-        logger.info("started");
+        logger.info("started {}", transportService.getLocalNode());
 
         pluginsService.filterPlugins(ClusterPlugin.class).forEach(ClusterPlugin::onNodeStarted);
 
--- a/docs/changelog/85773.yaml
+++ b/docs/changelog/85773.yaml
@@ -1,0 +1,5 @@
+pr: 85773
+summary: Log node identity at startup
+area: Network
+type: enhancement
+issues: []


server/src/main/java/org/elasticsearch/node/Node.java
```diff
@@ -1267,7 +1267,7 @@ public void onTimeout(TimeValue timeout) {
             }
         }
 
-        logger.info("started");
+        logger.info("started {}", transportService.getLocalNode());
```
idegtiarenko, line 1270: This logs something like `started {runTask-0}{xxhYOWXfQdyqY2Au33UKmA}{XN9z8jtRT5OiRzxJxDqFFg}{127.0.0.1}{127.0.0.1:9300}{cdfhilmrstw}{ml.machine_memory=16557490176, ml.max_jvm_size=536870912, testattr=test, xpack.installed=true}` :+1: 

Is it also worth to update https://github.com/elastic/elasticsearch/blob/327163d07e431bdcd2ba26f2e65fba8df3b47481/server/src/main/java/org/elasticsearch/node/Node.java#L320-L321

to log the exact build hash if the version is a SNAPSHOT?

idegtiarenko (approved):

server/src/main/java/org/elasticsearch/node/Node.java
DaveCTurner: I think we log this already, the third field in the `build[{}/{}/{}/{}]` bit is `Build.CURRENT.hash()`.

Issue #85127, opened 2022-03-19
array_index_out_of_bounds_exception when terms agg ordered by top_metrics agg


andrewctate:
### Elasticsearch Version

8.2

### Java Version

_bundled_

### OS Version

OSX 11.6

### Problem Description

A `terms` aggregation on a keyword field ordered by a `top_metrics` aggregation on a date field for a particular time range results in an `array_index_out_of_bounds_exception` on the Elasticsearch side.

### Steps to Reproduce

Add the `kibana_sample_data_ecommerce` to the cluster (I was using the 8.2 snapshot). Then, run the following query that runs a `terms` agg ordered by a `top_metrics` agg.

```
{
  "aggs": {
    "0": {
      "terms": {
        "field": "customer_first_name.keyword",
        "order": {
          "1-bucket>1-metric[customer_birth_date]": "desc"
        },
        "size": 5
      },
      "aggs": {
        "1-bucket": {
          "filter": {
            "bool": {
              "must": [],
              "filter": [
                {
                  "bool": {
                    "should": [
                      {
                        "exists": {
                          "field": "customer_birth_date"
                        }
                      }
                    ],
                    "minimum_should_match": 1
                  }
                }
              ],
              "should": [],
              "must_not": []
            }
          },
          "aggs": {
            "1-metric": {
              "top_metrics": {
                "metrics": {
                  "field": "customer_birth_date"
                },
                "size": 1,
                "sort": {
                  "order_date": "desc"
                }
              }
            }
          }
        }
      }
    }
  },
  "size": 0,
  "fields": [
    {
      "field": "customer_birth_date",
      "format": "date_time"
    },
    {
      "field": "order_date",
      "format": "date_time"
    },
    {
      "field": "products.created_on",
      "format": "date_time"
    }
  ],
  "script_fields": {},
  "stored_fields": [
    "*"
  ],
  "runtime_mappings": {},
  "_source": {
    "excludes": []
  },
  "query": {
    "bool": {
      "must": [],
      "filter": [
        {
          "range": {
            "order_date": {
              "format": "strict_date_optional_time",
              "gte": "2021-12-03T23:36:40.426Z",
              "lte": "2022-03-18T22:36:40.426Z"
            }
          }
        }
      ],
      "should": [],
      "must_not": []
    }
  }
}
```

## Response Body
```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "array_index_out_of_bounds_exception",
[...]

PR #85774, opened 2022-04-11
fix: ordering terms aggregation on top metrics null values

salvatore-campagna:
If top_metrics buckets are never collected (i.e. a filter never
passing data to a top_metrics nested aggregator) sorting results
in an index out of bounds exception. The data array, filled at
collection time, is actually empty and we try to access it to do
the sorting using ordinals as indices. To be more precise, we do the
comparison when filling the priority queue used to build aggregation
results. The first insertion is successful since the priority queue is empty
and there is no comparison to perform. Anyway, as soon as we insert
the second element in the priority queue, the comparator is invoked
and access to the empty array takes place.

Here we use the special Double.NaN value to report missing data
for the top_metrics metric field. This happens for empty or null
fields. The comparator we use is a 'NaN-aware' comparator.
As a consequence, data is sorted using another comparator,
the default '_key' comparator added when creating the compound
comparator.

This is a fix for #85127.


x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/analytics/nested_top_metrics_sort.yml
```diff
@@ -278,3 +396,653 @@ setup:
...
+
+  - match: { aggregations.name.buckets.4.key: "daniel" }
+  - match: { aggregations.name.buckets.4.doc_count: 3 }
+  - match: { aggregations.name.buckets.4.field_exists.doc_count: 3 }
+  - length: { aggregations.name.buckets.4.field_exists.top_metrics.top: 1 }
+  - length: { aggregations.name.buckets.4.field_exists.top_metrics.top.0.sort: 1 }
+  - match: { aggregations.name.buckets.4.field_exists.top_metrics.top.0.sort.0: "2021-11-09T17:29:00.000Z" }
+  - match: { aggregations.name.buckets.4.field_exists.top_metrics.top.0.metrics.version: 1 }
+
+---
```
salvatore-campagna, line 958: The last two tests are here just to highlight that the behaviour of the top_metric aggregation changes with the value of `rewrite_to_filter_by_filter`. This is not correct, we would like to have consistent behaviour. I will submit an issue to track this and point the issue to these two tests (and their counterpart where the setting is true).

salvatore-campagna:
@elasticmachine update branch

salvatore-campagna:
@elasticmachine update branch

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/analytics/nested_top_metrics_sort.yml
salvatore-campagna: I will create the issue after merging this PR so that I can reference the two tests in the ticket.

not-napoleon (approved):
LGTM

PR #85239, opened 2022-03-22
Fix leaking listeners bug on frozen tier

tlrx:
We identified a bug in searchable snapshot code that could prevent some search queries to be completed. This bug results in search tasks lingering on a `data_frozen` node for hours or days, filling the `search` thread pool and blocking other search requests to be processed.

This bug can be reproduced on a `data_frozen` node with:
- multiple search queries requesting the same region from the shared cache 
- the region corresponds to cached data not yet available in cache
- the searchable snapshot's thread pool that fetches cached data has no active workers
- the thread that will write the cached data (ie, the thread that picks up the `Gap` to fill) has restrictive permissions that prevents new thread to be created

When such a situation occurs, the thread pool in charge of executing the `AbstractRunnable` that fills the cache throws an exception (in our case, an `AccessControlException`) that is not caught and not logged. In consequence the `Gap` that should have been written in cache is never filled with data and the listeners waiting for the cached data to be available won't be called back, blocking the execution of the searches waiting for the data.

This pull request introduces an integration test that reproduce this situation:
```
java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "modifyThreadGroup")
	at java.security.AccessControlContext.checkPermission(AccessControlContext.java:485) ~[?:?]
	at java.security.AccessController.checkPermission(AccessController.java:1068) ~[?:?]
	at java.lang.SecurityManager.checkPermission(SecurityManager.java:416) ~[?:?]
	at org.elasticsearch.secure_sm.SecureSM.checkThreadGroupAccess(SecureSM.java:186) ~[?:8.2.0-SNAPSHOT]
	at org.elasticsearch.secure_sm.SecureSM.checkAccess(SecureSM.java:135) ~[?:8.2.0-SNAPSHOT]
	at java.lang.ThreadGroup.checkAccess(ThreadGroup.java:335) ~[?:?]
	at java.lang.Thread.<init>(Thread.java:422) ~[?:?]
	at java.lang.Thread.<init>(Thread.java:708) ~[?:?]
	at org.elasticsearch.common.util.concurrent.EsExecutors$EsThreadFactory.newThread(EsExecutors.java:274) ~[?:8.2.0-SNAPSHOT]
	at java.util.concurrent.ThreadPoolExecutor$Worker.<init>(ThreadPoolExecutor.java:630) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.addWorker(ThreadPoolExecutor.java:920) ~[?:?]
[...]

PR #85217, opened 2022-03-22
Better logging when EsThreadPoolExecutor throws unexpected exceptions

tlrx:
We recently had issues with `EsThreadPoolExecutor` throwing unexpected exceptions (in our case an `AccessControlException`) before the `Runnable` or `AbstractRunnable` has been submitted for execution. In the case of `AbstractRunnable` the task is not rejected and it causes some resources to leak.

This pull request catches unexpected exceptions thrown when calling the `EsThreadPoolExecutor#execute()` method and logs a message. It also assert that this situation should never happen, and if that's the case an `AssertionError` is thrown in tests.

PR #85775, opened 2022-04-11
Remove special rejection handling in Cold/Frozen cache services

tlrx:
Now #85217 is merged the special handling introduced by #85239
to fail Gaps when a task is rejected can be removed. The
`onRejection` method of the  `AbstractRunnable` task will be
executed if the task is rejected or if an unexpected exception
is thrown before the task is added to the thread pool executor
queue.


DaveCTurner (approved):
Looks good, thanks @tlrx.

fcofdez (approved):
LGTM 👍

tlrx:
Thanks David and Francisco

PR #85776, opened 2022-04-11
https://github.com/glennvill/elasticsearch.git

glennvill:
fixed query builder so that top can be the same as bottom

As the same bounding box on a geo_shape query works. Lucene supports it so this fix is a must


craigtaverner (changes requested):
Missed the second check. Also, there really should be tests for this.

craigtaverner:
It appears this work was already done by @jchrys on the same date (11th April) in the PR at https://github.com/elastic/elasticsearch/pull/85788, which was reviewed by @iverase and merged into 8.3.0.

PR #82737, opened 2022-01-18
Add support for HTTP Proxies for the GCS repository

arteam:
The change adds 3 new client properties for the GCS repository:

* `gcs.client.default.proxy.type`
* `gcs.client.default.proxy.host`
* `gcs.client.default.proxy.port`

They allow to configure a [java.net.Proxy](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/Proxy.html) for the GCS SDK to use when communicating with the GCS API.

Resolves #82444

PR #85785, opened 2022-04-11
Expose proxy settings for GCS repositories

### LINKED ISSUE #84569: Repository GCS Plugin makes the elasticsearch crash on startup when client's proxy settings are specified in elasticsearch.yml.
### Elasticsearch Version

8.0

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

win10

### Problem Description

I am using GCP plugin with Elasticsearch 8.0 and it seems that the settings that are specified in the documentation [here](https://www.elastic.co/guide/en/elasticsearch/reference/8.0/repository-gcs.html) are not available to use in Elasticsearch. I have further gone into investigation to investigate GCS repository plugin and below are my observations:
The settings are declared in GoogleCloudStorageClientSettings.java class [here](https://github.com/elastic/elasticsearch/blob/690bac5dc375b5a1e8f6b3b091d38ee6cdeb0b05/modules/repository-gcs/src/main/java/org/elasticsearch/repositories/gcs/GoogleCloudStorageClientSettings.java#L101) but they are not loaded in the implementation getSettings() method [here](https://github.com/elastic/elasticsearch/blob/690bac5dc375b5a1e8f6b3b091d38ee6cdeb0b05/modules/repository-gcs/src/main/java/org/elasticsearch/repositories/gcs/GoogleCloudStoragePlugin.java#L66).
Due to this elasticsearch.bat is throwing error and exiting instead of starting Elasticsearch.

### Steps to Reproduce

1. Specify the below settings in elasticsearch.yml:
gcs.client.default:
  proxy.host: "abc.com"
  proxy.port: 80
  proxy.type: "http"
2. Start elasticsearch.bat and this will generate **java.lang.IllegalArgumentException: unknown setting [gcs.client.default.proxy.host] please check that any required plugins are installed, or check the breaking changes documentation for removed settings** followed by exiting the bat script.

### Logs (if relevant)

@KarenSumner @liladler Could you please help prioritize of fixing this bug?
elastic_bug

arteam:
They were added in #82737, but not exposed via the `GCSPlugin` to end users.

Resolves #84569

### DIFF
--- a/docs/changelog/85785.yaml
+++ b/docs/changelog/85785.yaml
@@ -1,0 +1,6 @@
+pr: 85785
+summary: Expose proxy settings for GCS repositories
+area: Snapshot/Restore
+type: bug
+issues:
+ - 84569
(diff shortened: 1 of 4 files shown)


henningandersen (approved):
LGTM, though the labels look wrong, please correct before merging.

PR #85791, opened 2022-04-11
Fixes CORS headers needed by Elastic clients

swallez:
This PR updates the default values for CORS headers so that Elasticsearch client libraries that can run in a web browser (e.g. JavaScript client, Go & Rust clients compiled to WebAssembly) work out of the box by just enabling `http.cors.enabled: true` and setting `http.cors.allow-origin`.

It adds the following headers to `http.cors.allow-headers`'s default value: `Authorization` for webapp-level authentication, `Accept`, and `User-Agent` and `X-Elastic-Client-Meta` that are useful for telemetry.

Additionally, a new `http.cors.expose-headers` setting allows configuring [headers that are exposed to the code in the browser](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers), defaulting to `X-elastic-product` which allows clients to successfully run product checks.

See also https://github.com/elastic/elasticsearch-rs/pull/200


docs/reference/modules/http.asciidoc
```diff
@@ -110,9 +110,16 @@ Which methods to allow. Defaults to `OPTIONS, HEAD, GET, POST, PUT, DELETE`.
 // tag::http-cors-allow-headers-tag[]
 `http.cors.allow-headers` {ess-icon}::
 (<<static-cluster-setting,Static>>)
-Which headers to allow. Defaults to `X-Requested-With, Content-Type, Content-Length`.
+Which headers to allow. Defaults to `X-Requested-With, Content-Type, Content-Length, Authorization, Accept, User-Agent, X-Elastic-Client-Meta`.
```
tvernum, line 113: I find it strange that we default `access-control-allow-headers` to include `Authorization` but `access-control-allow-credentials` to `false`?

Is that intentional?

swallez: Yes, it is intentional:
* [`access-control-allow-headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers), in a response to a preflight request, defines what headers (outside of the safelist) are allowed in CORS requests. Since clients set the `Authorization` header in their requests, it has to be allowed by the server in preflight responses.
* Defaulting [`access-control-allow-credentials`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials) to `false` forbids the use of browser-level credentials by clients, and thus requires clients to be provided with credential information that is then set in the `Authorization` header.

CORS is sometimes confusing 😅

tvernum (approved):
LGTM, but we should check whether @rjernst still wants to review.

rjernst:
I don't have any thoughts here.

swallez:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|7.17|[](https://github.com/elastic/elasticsearch/pull/93659)|

### Questions ?

PR #85792, opened 2022-04-11
Handle .geoip_databases being an alias or a concrete index

### LINKED ISSUE #85756: GeoIP Ingest plugin should be not break with feature migration
### Bug

It is possible for a system feature migration to break the geoip-ingest plugin on a cluster that has been upgraded from 7.17 to 8.x. The system features migration will reindex the `.geoip_databases` index into a new index called `.geoip_databases-reindexed-for-8`, then delete the original `.geoip_databases` index and replace it with an alias. The plugin is written to expect `.geoip_databases` to be a concrete index, so it fails to reload its database, and geoip ingest processors stop working.

We need to fix the code so that running a system feature migration doesn't create this problem.

### Reproducing

Steps to reproduce:
1. On a 7.17 cluster, [create a pipeline with a geoip processor](https://www.elastic.co/guide/en/elasticsearch/reference/current/geoip-processor.html).
2. Upgrade the cluster to 8.1.
3. Run a [system features migration](https://www.elastic.co/guide/en/elasticsearch/reference/current/feature-migration-api.html) and wait for it to complete.
4. Wait for the geoip-ingest plugin to reload its databases, or restart the cluster to make changes take effect
5. Ingest a document using the pipeline.

Here are some curl commands I used on my local, single-node cluster:

<details><summary>Expand for details</summary>

```
# ON 7.17 - create geoip processor

curl -s -XPUT -H'content-type:application/json' \
   'localhost:9200/_ingest/pipeline/geoip' \
   -d '{"description":"test","processors":[{"geoip":{"field":"ip"}}]}'

curl -s -XPUT -H'content-type:application/json' \
    'localhost:9200/my-index-00001/_doc/pre_migration?pipeline=geoip' \
    -d '{"ip":"8.8.8.8"}'

curl -s -XGET 'localhost:9200/my-index-00001/_doc/pre_migration'

# ON 7.17 - create snapshot

curl -s -XPUT -H'content-type:application/json' \
    'localhost:9200/_snapshot/fs_backup' \
    -d '{"type":"fs","settings":{"location":"/Users/wbrafford/work/es-builds/snapshots"}}'

curl -s -XPUT -H'content-type:application/json' \
    'localhost:9200/_snapshot/fs_backup/pre_migration' \
    -d '{"indices":"-*","feature_states":["geoip"],"include_global_state":false}'

# ON 8.1 - run feature migration

curl -s -XPOST 'localhost:9200/_migration/system_features'

# ON 8.1 - restart and test doc

curl -s -XPUT -H'content-type:application/json' \
    'localhost:9200/my-index-00001/_doc/post_migration?pipeline=geoip' \
    -d '{"ip":"8.8.8.8"}'

curl -s -XGET 'localhost:9200/my-index-00001/_doc/post_migration'

# restore snapshot

curl -s -XPOST -H'content-type:application/json' \
    'localhost:9200/_snapshot/fs_backup/pre_migration/_restore' \
    -d '{"indices":"-*","feature_states":["geoip"],"include_global_state":false}'

curl -s -XPUT -H'content-type:application/json' \
    'localhost:9200/my-index-00001/_doc/post_restore?pipeline=geoip' \
    -d '{"ip":"8.8.8.8"}'

curl -s -XGET 'localhost:9200/my-index-00001/_doc/post_restore'
```

</details>

### Workaround

The only fix I know of is to restore a geoip feature state from a snapshot taken before the system feature migration:

```
POST /_snapshot/<repo_name>/<snapshot_name>/_restore
{
  "indices": "-*",
  "feature_states": ["geoip"],
  "include_global_state": false
}
```

It doesn't really matter how old the snapshot is, because once the plugin is restored to a good state, it can update the geoip index.

### Open questions

What approach should we take to fix this?

1. The `.geoip_databases` index doesn't contain any user data. We could just "migrate" it by deleting it and recreating it. This kind of fix would be the responsibility of the core-infra team.
2. We could make the geoip ingest plugin robust for the case where `.geoip_databases` is an alias, not a concrete index. This doesn't seem like something the plugin needs intrinsically, but it might be easier to do than changing the migration code.

We should also look at giving users a more convenient way to reset the state of the geoip-ingest plugin. https://github.com/elastic/elasticsearch/issues/70426 would have been really useful to have for this bug.

cc @dakrone, @gwbrown, @joegallo

joegallo:
Closes #85756

There's two places we need to handle the maybe-it's-an-alias-or-maybe-it's-not logic. First, `DatabaseNodeService` wants to make sure all the primary shards are active before the service fully initializes. Second, `GeoIpDownloaderTaskExecutor` needs to delete the real index (not the alias!) when the downloader is disabled, so we need to potentially dereference the alias there, too.

Manually testing with a persistent data directory is pretty straightforward. I used a collection of requests like this:

```
PUT _cluster/settings
{
  "persistent": {
    "ingest.geoip.downloader.enabled": true
  }
}

PUT _cluster/settings
{
  "persistent": {
    "ingest.geoip.downloader.enabled": false
  }
}

POST _migration/system_features

PUT _ingest/pipeline/geoip
{
  "description": "Add geoip info",
  "processors": [
    {
      "geoip": {
        "field": "ip",
        "target_field": "geo",
        "database_file": "GeoLite2-Country.mmdb"
      }
    }
  ]
}

PUT my-index-00001/_doc/my_id?pipeline=geoip
{
  "ip": "8.8.8.8"
}

GET my-index-00001/_doc/my_id

GET _cat/indices?expand_wildcards=all
```

And then tested like this:

1. Start up a 7.17 node, enable the downloader, put the geoip pipeline.
2. Test some docs (they should have the ip turned into geo data).
3. Stop the node, upgrade to 8.3.0-SNAPSHOT, run the system features migration, restart the node.
4. Verify that the index exists, and it is actually behind an alias now because of the migration.
5. Test some docs (they should have the ip turned into geo data).
6. Disable the downloader and verify that the index was deleted.

### DIFF
--- a/docs/changelog/85792.yaml
+++ b/docs/changelog/85792.yaml
@@ -1,0 +1,6 @@
+pr: 85792
+summary: Handle `.geoip_databases` being an alias or a concrete index
+area: Ingest
+type: bug
+issues:
+ - 85756
--- a/modules/ingest-geoip/qa/full-cluster-restart/build.gradle
+++ b/modules/ingest-geoip/qa/full-cluster-restart/build.gradle
@@ -1,0 +1,70 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+
+import org.elasticsearch.gradle.Version
+import org.elasticsearch.gradle.internal.info.BuildParams
+import org.elasticsearch.gradle.testclusters.StandaloneRestIntegTestTask
+
+apply plugin: 'elasticsearch.internal-testclusters'
+apply plugin: 'elasticsearch.standalone-rest-test'
+apply plugin: 'elasticsearch.internal-test-artifact'
+apply plugin: 'elasticsearch.bwc-test'
+
+BuildParams.bwcVersions.withWireCompatible(v -> v.before("8.0.0")) { bwcVersion, baseName ->
+  def baseCluster = testClusters.register(baseName) {
+    testDistribution = "DEFAULT"
+    if (bwcVersion.before(BuildParams.bwcVersions.minimumWireCompatibleVersion)) {
+      // When testing older versions we have to first upgrade to 7.last
+      versions = [bwcVersion.toString(), BuildParams.bwcVersions.minimumWireCompatibleVersion.toString(), project.version]
+    } else {
+      versions = [bwcVersion.toString(), project.version]
+    }
+    numberOfNodes = 2
+    // some tests rely on the translog not being flushed
+    setting 'indices.memory.shard_inactive_time', '60m'
+    setting 'path.repo', "${buildDir}/cluster/shared/repo/${baseName}"
+    setting 'xpack.security.enabled', 'false'
+    requiresFeature 'es.index_mode_feature_flag_registered', Version.fromString("8.0.0")
+  }
+
+  tasks.register("${baseName}#oldClusterTest", StandaloneRestIntegTestTask) {
+    useCluster baseCluster
+    mustRunAfter("precommit")
+    doFirst {
+      delete("${buildDir}/cluster/shared/repo/${baseName}")
+    }
+
+    systemProperty 'tests.is_old_cluster', 'true'
+  }
+
+  tasks.register("${baseName}#upgradedClusterTest", StandaloneRestIntegTestTask) {
+    useCluster baseCluster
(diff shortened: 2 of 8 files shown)


joegallo:
I'm 65% of the way to a pretty nice integration test, but there's something interesting going on with that remaining 35% that makes me think there's still something off here. Added the WIP tag back.

joegallo:

joegallo:
@dakrone @williamrandolph I dropped the WIP label again (Jenkins gave me the ✅, but of course dropping the WIP label means I got a new commit added -- I fully expect another clean run from CI). This is ready for review now.

dakrone (approved):
LGTM, thanks for adding the test so it could catch the other part of the bug in `DatabaseNodeService` :)

williamrandolph (commented):
Apologies for not getting to this yesterday! I'm really impressed by the two-hop integration test, and of course the actual fix looks great. Thanks!

PR #71163, opened 2021-04-01
Remove legacy role settings

jasontedor:
This commit removes the previously deprecated legacy role settings. These settings have been replaced by node.roles.

Relates #54998
Relates #71143
Closes #66409

PR #85794, opened 2022-04-11
Remove remaining references to legacy role settings

rjernst:
The legacy node.ROLENAME settings were removed in 8.0 (#71163), but
machine dependent heap still handles them. This commit cleans up the
heap settings determination to no longer look for these legacy settings.

relates #85758


mark-vieira (approved):
LGTM

PR #85796, opened 2022-04-11
Remove flavor from build

rjernst:
The default distribution is the only remaining build flavor, and has been for
quite a while now. This commit removes flavor from the internal Build
class. It keeps rest api compat for nodes info for now by hardcoding
`default`.


distribution/tools/plugin-cli/src/main/java/org/elasticsearch/plugins/cli/InstallPluginAction.java
```diff
@@ -225,7 +225,7 @@ public void execute(List<PluginDescriptor> plugins) throws Exception {
             terminal.println(logPrefix + "Installing " + pluginId);
             try {
                 if ("x-pack".equals(pluginId)) {
-                    handleInstallXPack(buildFlavor());
+                    throw new UserException(ExitCodes.CONFIG, "this distribution of Elasticsearch contains X-Pack by default");
```
mark-vieira, line 228: Do we even need to handle this anymore at this point. The "x-pack" plugin hasn't been a thing for many versions. I assume this was only added to ease migration after x-pack was folded into ES.

server/src/main/java/org/elasticsearch/Build.java
```diff
@@ -22,50 +22,14 @@
...
-        }
-
-        public static Flavor fromDisplayName(final String displayName, final boolean strict) {
-            switch (displayName) {
-                case "default":
-                    return Flavor.DEFAULT;
-                case "oss":
-                    return Flavor.OSS;
-                case "unknown":
-                    return Flavor.UNKNOWN;
```
mark-vieira, line 56: What do we use for the integ-test distribution? Was this "unknown" before?

mark-vieira (approved):
One question, otherwise LGTM.

distribution/tools/plugin-cli/src/main/java/org/elasticsearch/plugins/cli/InstallPluginAction.java
rjernst: That is a fair point, but I would like to think about that separately. We certainly might get users who have used ES before trying to install x-pack, though at some point we should definitely remove this.

server/src/main/java/org/elasticsearch/Build.java
rjernst: Yes it would have been unknown, though I don't think it matters, since nothing will be checking it now.

PR #85822, opened 2022-04-12
Remove ILM classes from HLRC

joegallo:

There were two dangling references (see first and third commits), but then everything else was just a simple delete.


dakrone (approved):
LGTM

PR #85826, opened 2022-04-12
refactor array part into a BytesRefArray which can be serialized and …

hendrikmuhs:
refactor BytesRefHash and extract BytesRefArray. The BytesRefArray can be serialized for communication.

**Background**

This PR is a spin off from a bigger task to implement a frequent_items aggregation. For this aggregation I need a data structure that collects data in form of largish frequency tables. These tables need to be centrally merged before the core algorithm mines for frequent items (I happily provide more details, just ping me). The closest re-usable structure I could find was BytesRefHash used in several places, however it wasn't able serialize and deserialize. 

This PR extracts part that stores the keys out of BytesRefHash. 

Serialization/Deserialization: For communication, get the BytesRefsArray. On the receiver side, serialize the BytesRefsArray and re-create the BytesRefHash. The added constructor uses the given BytesRefsArray and re-creates the hashtable part.

Relates https://github.com/elastic/elasticsearch/pull/83055

### DIFF
--- a/server/src/main/java/org/elasticsearch/common/util/BytesRefHash.java
+++ b/server/src/main/java/org/elasticsearch/common/util/BytesRefHash.java
@@ -40,11 +40,31 @@ public final class BytesRefHash extends AbstractHash implements Accountable {
 
     // Constructor with configurable capacity and load factor.
     public BytesRefHash(long capacity, float maxLoadFactor, BigArrays bigArrays) {
+        this(null, capacity, maxLoadFactor, bigArrays);
+    }
+
+    public BytesRefHash(BytesRefArray bytesRefArray, BigArrays bigArrays) {
+        this(bytesRefArray, DEFAULT_MAX_LOAD_FACTOR, bigArrays);
+    }
+
+    public BytesRefHash(BytesRefArray bytesRefs, float maxLoadFactor, BigArrays bigArrays) {
+        this(bytesRefs, bytesRefs.size() + 1, maxLoadFactor, bigArrays);
+
+        // recreate hashes
+        for (int i = 0; i < bytesRefs.size(); ++i) {
+            bytesRefs.get(i, spare);
+            reset(rehash(spare.hashCode()), i);
+        }
+    }
+
+    private BytesRefHash(BytesRefArray byteRefs, long capacity, float maxLoadFactor, BigArrays bigArrays) {
         super(capacity, maxLoadFactor, bigArrays);
+
         boolean success = false;
         try {
+            this.bytesRefs = byteRefs != null ? byteRefs : new BytesRefArray(capacity, bigArrays);
+
             // `super` allocates a big array so we have to `close` if we fail here or we'll leak it.
-            bytesRefs = new BytesRefArray(capacity, bigArrays);
             hashes = bigArrays.newIntArray(capacity, false);
             success = true;
         } finally {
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/elasticsearch/common/util/BytesRefHash.java
```diff
@@ -51,6 +55,41 @@ public BytesRefHash(long capacity, float maxLoadFactor, BigArrays bigArrays) {
         spare = new BytesRef();
     }
 
+    public BytesRefHash(BytesRefArray bytesRefArray, BigArrays bigArrays) {
+        this(bytesRefArray, DEFAULT_MAX_LOAD_FACTOR, bigArrays);
+    }
+
+    public BytesRefHash(BytesRefArray byteRefs, float maxLoadFactor, BigArrays bigArrays) {
```
grcevski, line 62: It looks like this BytesRefHash constructor will re-allocate twice the hashes and the spare? We also seem to allocate bytesRefs by calling the super constructor, but then we immediately overwrite that reference with the incoming BytesRefArray. I might be confused at what's happening, can you please clarify?

server/src/main/java/org/elasticsearch/common/util/BytesRefHash.java
```diff
@@ -51,6 +55,41 @@ public BytesRefHash(long capacity, float maxLoadFactor, BigArrays bigArrays) {
...
+                close();
+            }
+        }
+        spare = new BytesRef();
+
+        // recreate hashes
+        for (int i = 0; i < byteRefs.size(); ++i) {
+            byteRefs.get(i, spare);
+            int code = rehash(spare.hashCode());
+            final long slot = slot(code, mask);
```
grcevski, line 82: There's a reset method on this class that we can use to eliminate this code duplication.

server/src/main/java/org/elasticsearch/common/util/BytesRefHash.java
hendrikmuhs: it calls `super(...)` -> the constructor of `AbstractHash`, double allocation would happen if calling `this(...)`

grcevski: Ah, I see what I did there :). That makes sense, would it be possible to refactor the code such that we reuse most of the code from `public BytesRefHash(long capacity, float maxLoadFactor, BigArrays bigArrays)`, and then just recompute the hashes?

hendrikmuhs: see [commit](https://github.com/elastic/elasticsearch/pull/85826/commits/e9b6b262b5de7da83105811c5e9df712a3afa392), it is a little complex, because we have to ensure we do not leak BytesRefArray.

server/src/main/java/org/elasticsearch/common/util/BytesRefHash.java
```diff
@@ -34,13 +40,31 @@ public BytesRefHash(long capacity, BigArrays bigArrays) {
...
+        this(bytesRefs, bytesRefs.size() + 1, maxLoadFactor, bigArrays);
+
+        // recreate hashes
+        for (int i = 0; i < bytesRefs.size(); ++i) {
+            bytesRefs.get(i, spare);
+            reset(rehash(spare.hashCode()), i);
+        }
+    }
+
+    private BytesRefHash(BytesRefArray byteRefs, long capacity, float maxLoadFactor, BigArrays bigArrays) {
```
grcevski, line 60: I thought of a hypothetical situation, but I'm not sure if it's an issue or not. In the case where we deserialize the ByteRefArray and pass it in to the constructor of BytesRefHash, we will sometimes release this reference and sometimes not, depending on if super fails to allocate and/or throws an exception. 

The super call isn't in the try catch, so if the super fails we will not call release on the passed in BytesRefArray, assuming the caller of the constructor would be responsible for that. However, if we fail to allocate the hashes, we've already set the reference `this.bytesRefs`, so the close() call would release BytesRefArray. 

My question is, do we ever want to release a passed in BytesRefArray if we haven't allocated it here? Perhaps there's no danger in releasing a reference twice, however releasing it sometimes might cause bugs in the caller code if they don't expect the BytesRefArray to be released by being passed in.

Swapping the lines such that hashes is allocated first would eliminate the problem, but I'm still not sure how do we document or inform users of the passed in byteRefs that releasing is delegated to the ByteRefsHash object once passed in.

grcevski: Perhaps writing tests for these exceptional situations would ensure we have the right contracts in place.

hendrikmuhs: I documented the correct usage and created 2 tests, see https://github.com/elastic/elasticsearch/pull/85826/commits/f670bd928a23e38304ef0008da53850fd0c544f2

grcevski (approved):
Thanks! LGTM!

Issue #84773, opened 2022-03-08
Implement impacts block for the shards availability indicator


andreidan:
Add "impact" block to indicators and populate it for the shards availability indicator. 
Impacts should include high level language, and wherever possible all the resources that are affected (in the shards availability indicator case the indices).

The impacts should also have a severity assigned.

ie. 
- if primary shards are unavailable (severity 2)
```
Cannot add data to 7 indices [index-1, index-2...] currently. Search results might return incomplete results.
```

- if replica shards are unavailable (severity 3)
```
Redundancy for 7 indices [index-1, index-2...] is currently disrupted. Fault tolerance and search scalability is reduced.
```

- if both primary and replica shards are unavailable (severity 1)
```
Indices [index-1, index-2, ...] are unavailable. You are at risk of losing the data in these indices.
```

masseyke:
I'm thinking a format like this (ignore the actual message and severity).
```
...
            "indicators": {
...
                "shards_availability": {
                    "details": {
                        "creating_primaries": 0,
                        "initializing_primaries": 0,
                        "initializing_replicas": 1,
                        "restarting_primaries": 0,
                        "restarting_replicas": 0,
                        "started_primaries": 3,
                        "started_replicas": 2,
                        "unassigned_primaries": 0,
                        "unassigned_replicas": 0
                    },
                    "impact": {
                        "impact_description": "Searches may be slower than usual",
                        "severity": 3
                    },
                    "status": "yellow",
                    "summary": "This cluster has 1 unavailable replica."
                }
            },
...
```

masseyke:
@andreidan any thoughts on severity, and how it compares to status? Isn't status the same thing as severity? Would I ever have status = red, but severity anything other than 1?

andreidan:
@masseyke I think the severity will provide a way for us to order the impacts reported by all the indicators. I believe even now for the shards availability indicator there's a bit more nuance to the impacts severity (eg. sev 1 and 2 both would report a RED status, however one's more serious than the other)

PR #85830, opened 2022-04-12
Add support for impact_areas to health impacts 

### LINKED ISSUE #85829: Add `impact_areas` field to impacts
eg
```
"impacts": [
            {
              "severity": 3,
              "description": "Searches might return slower than usual. Fewer redundant copies of the data exist on 2 indices [.ds-logslogslogs-2022.03.29-000001, my-index-000001].",
              "impact_areas": [ "search" ]
            }
          ]
```

andreidan:
This adds an `impact_areas` section to the impacts returned by the
health API.

eg. the `shards_availability` indicator might be returning:
```
"impacts": [
  {
    "severity": 1,
    "description":"Cannot add data to 1 index [red-index]. Searches might return incomplete results.",
    "impact_areas": [
      "ingest","search"
    ]
  },
  {
    "severity": 2,
    "description": "Searches might return slower than usual. Fewer redundant copies of the data exist on 2 indices [.ds-logslogslogs-2022.04.12-000001, my-index-000001].",
    "impact_areas": [
      "search"
    ]
  }
]
```

Resolves #85829

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
@@ -259,7 +259,7 @@ public class ShardsAvailabilityHealthIndicatorService implements HealthIndicator
                     indicesWithUnavailableReplicasOnly.size() == 1 ? "index" : "indices",
                     getTruncatedIndicesString(indicesWithUnavailableReplicasOnly, clusterMetadata)
                 );
-                impacts.add(new HealthIndicatorImpact(3, impactDescription, List.of(ImpactArea.SEARCH)));
+                impacts.add(new HealthIndicatorImpact(2, impactDescription, List.of(ImpactArea.SEARCH)));
             }
             return impacts;
         }
--- a/server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
@@ -23,6 +23,7 @@ import org.elasticsearch.health.HealthIndicatorImpact;
 import org.elasticsearch.health.HealthIndicatorResult;
 import org.elasticsearch.health.HealthIndicatorService;
 import org.elasticsearch.health.HealthStatus;
+import org.elasticsearch.health.ImpactArea;
 import org.elasticsearch.health.SimpleHealthIndicatorDetails;
 
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
```diff
@@ -258,7 +259,7 @@ public List<HealthIndicatorImpact> getImpacts() {
                     indicesWithUnavailableReplicasOnly.size() == 1 ? "index" : "indices",
                     getTruncatedIndicesString(indicesWithUnavailableReplicasOnly, clusterMetadata)
                 );
-                impacts.add(new HealthIndicatorImpact(3, impactDescription));
```
masseyke, line 261: Was the priority an intentional change? The description of https://github.com/elastic/elasticsearch/issues/84773 has unassigned replicas as a 3.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
```diff
@@ -241,7 +242,7 @@ public List<HealthIndicatorImpact> getImpacts() {
                     primaries.indicesWithUnavailableShards.size() == 1 ? "index" : "indices",
                     getTruncatedIndicesString(primaries.indicesWithUnavailableShards, clusterMetadata)
                 );
-                impacts.add(new HealthIndicatorImpact(1, impactDescription));
+                impacts.add(new HealthIndicatorImpact(1, impactDescription, List.of(ImpactArea.INGEST, ImpactArea.SEARCH)));
```
masseyke, line 245: I'm not sure what the use case for the impact areas is, but do "ingest" and "search" capture the impact areas of a missing primary? Should there be one like "cluster instability" or "data loss"? Calling missing data a search problem might not alert users enough.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
andreidan: Yes. The description was not updated to reflect that we ended up having only 2 impacts.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
andreidan: We'll iterate on this as we go along (and keep track with the UI development). For now I think this would suffice (and we need to keep the language high level as it's impacts related)

masseyke (approved):
LGTM

PR #85835, opened 2022-04-12
Remove Task classes from HLRC

joegallo:

Drops some methods that have no callers, and then these `org.elasticsearch.client.tasks` classes end up with no callers and can also be dropped.


dakrone (approved):
LGTM

Issue #84847, opened 2022-03-09
Node shutdown cluster state task improvements


DaveCTurner:
### Elasticsearch Version

master (and more)

### Installed Plugins

N/A

### Java Version

_bundled_

### OS Version

Cloud

### Problem Description

Today `TransportPutShutdownNodeAction` and `TransportDeleteShutdownNodeAction` submit their cluster state update tasks at `NORMAL` priority, which may mean that they take a long time to execute in an overloaded cluster. They also use the (now-forbidden) unbatched executor, and always generate a new cluster state update even if the change is a no-op. I think we should:

- [x] Implement a batching executor, since the orchestrator will sometimes retry on a timeout. (https://github.com/elastic/elasticsearch/pull/82296)
- [x] Increase the priority to `URGENT`. These tasks look to be pretty lightweight, and it's important for orchestration that they run ASAP. (https://github.com/elastic/elasticsearch/pull/85838)
- [x] Complete the listener without waiting for the follow-up reroute, so that we can ack that we processed the shutdown request quickly. I don't think real clients need to wait for the first reroute after the shutdown request was published; test clients might but we have other ways to wait for this in tests. (https://github.com/elastic/elasticsearch/pull/85846)
- [x] Avoid publishing a cluster state if the update is a no-op. Can we even avoid submitting the task if it looks like we already marked the node for shutdown? (https://github.com/elastic/elasticsearch/pull/85914)
- [ ] Bypass circuit breakers when manipulating shutdown metadata (in the REST and Transport layers). (https://github.com/elastic/elasticsearch/pull/86047)

DaveCTurner:
Henning and I noticed another problem with shutdown task resiliency: we respect circuit breakers at both REST and transport layers. I think we should skip circuit breakers on these things, the requests themselves don't take much extra heap and it's important for them to succeed even if the cluster is mostly overloaded.

rjernst:
Reopening as I didn't see the newly added ask.

PR #85838, opened 2022-04-12
Use urgent priority for node shutdown cluster state update

rjernst:
Node shutdown requests submit a cluster state update to initiate the
shutdown. Since the cluster may be overloaded, this update could take a
long time to be processed. This commit changes the priority to urgent,
so that shutdown requests can be processed as quickly as possible.

relates #84847


gwbrown:
Also tagging @pgomulka for awareness.

AthenaEryma (approved):
LGTM, thanks for taking care of this fix!

PR #85840, opened 2022-04-12
Refactor Cardinality Aggregation

not-napoleon:
This is a relatively small refactoring to lay the ground work for addressing some issues with the cardinality aggregation. It pulls some of the type based logic for selecting the collection strategy up to the factory level, using the values source registry.


not-napoleon:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/search/aggregations/AggregatorFactory.java
```diff
@@ -17,6 +17,12 @@
 
 import static org.elasticsearch.search.aggregations.support.AggregationUsageService.OTHER_SUBTYPE;
 
+/**
+ * Aggregator factories are responsible for creating the per-shard aggregator instances.  They should select the aggregator instance type
+ * based on the type of the data being aggregated (using the {@link org.elasticsearch.search.aggregations.support.ValuesSourceRegistry}
+ * when appropriate), and any optimizations that factory can make.  The factory layer is the correct place for heuristics to select
+ * different operating modes (such as using Global Ordinals or not).
+ */
```
nik9000, line 25: I typically wrap these manually around 80 or 100 characters to make them easier to read inline. Our layout tools don't force it so you don't have to do it, but it's a thing I like.

Also I'm super ok importing `ValuesSourceRegistry` even though it's only used by the comment. It doesn't hurt anything.

server/src/main/java/org/elasticsearch/search/aggregations/metrics/CardinalityAggregator.java
```diff
@@ -40,22 +40,23 @@
 /**
  * An aggregator that computes approximate counts of unique values.
  */
-public class CardinalityAggregator extends NumericMetricsAggregator.SingleValue {
+public abstract class CardinalityAggregator extends NumericMetricsAggregator.SingleValue {
 
-    private final int precision;
+    protected final int precision;
+    protected ValuesSourceConfig valuesSourceConfig;
```
nik9000, line 46: This can be final I think?

server/src/main/java/org/elasticsearch/search/aggregations/metrics/CardinalityAggregator.java
```diff
@@ -40,22 +40,23 @@
...
     private final ValuesSource valuesSource;
 
     // Expensive to initialize, so we only initialize it when we have an actual value source
     @Nullable
-    private HyperLogLogPlusPlus counts;
+    protected HyperLogLogPlusPlus counts;
```
nik9000, line 51: Is this still nullable?

server/src/main/java/org/elasticsearch/search/aggregations/metrics/DirectCollectorCardinalityAggregator.java
```diff
@@ -0,0 +1,43 @@
...
+import org.elasticsearch.search.aggregations.support.ValuesSourceConfig;
+
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * Cardinality aggregator that operates directly on the bytes values of a field.  This is necessary for fields that don't use ordinals at
+ * all, such as Range Fields.
+ */
+public class DirectCollectorCardinalityAggregator extends CardinalityAggregator {
```
nik9000, line 24: Maybe `BytesCollectingCardinalityAggregator`?

server/src/main/java/org/elasticsearch/search/aggregations/metrics/CardinalityAggregator.java
```diff
@@ -40,22 +40,23 @@
...
-    private int emptyCollectorsUsed;
-    private int numericCollectorsUsed;
-    private int ordinalsCollectorsUsed;
-    private int ordinalsCollectorsOverheadTooHigh;
-    private int stringHashingCollectorsUsed;
+    protected int emptyCollectorsUsed;
+    protected int numericCollectorsUsed;
+    protected int ordinalsCollectorsUsed;
+    protected int ordinalsCollectorsOverheadTooHigh;
+    protected int stringHashingCollectorsUsed;
```
nik9000, line 59: I think these might not make sense to have any more because you can use the name of the aggregator to figure out which collector it made - except for the one that does heuristics on ordinals. That one probably just needs the last three. I think.

server/src/main/java/org/elasticsearch/search/aggregations/metrics/CardinalityAggregatorFactory.java
```diff
@@ -28,6 +29,11 @@ class CardinalityAggregatorFactory extends ValuesSourceAggregatorFactory {
 
     private final Long precisionThreshold;
     private final CardinalityAggregatorSupplier aggregatorSupplier;
+    private int emptyCollectorsUsed;
+    private int numericCollectorsUsed;
+    private int ordinalsCollectorsUsed;
+    private int ordinalsCollectorsOverheadTooHigh;
+    private int stringHashingCollectorsUsed;
```
nik9000, line 36: I haven't pulled this one locally, but I think these might be unused.

server/src/main/java/org/elasticsearch/search/aggregations/metrics/CardinalityAggregatorFactory.java
```diff
@@ -65,10 +84,12 @@ public static void registerAggregators(ValuesSourceRegistry.Builder builder) {
                                 metadata
                             );
                         }
+                        // fallback in the default aggregator
+                        return new KeywordCardinalityAggregator(name, valuesSourceConfig, precision, context, parent, metadata);
```
nik9000, line 88: I wonder if `Ordinal` is an important word to have in the name of this thing. Its really like the bytes collecting one but it reasons about ordinals.

server/src/main/java/org/elasticsearch/search/aggregations/metrics/CardinalityAggregator.java
not-napoleon: Yes, because `createUnmapped` still makes an instance of `CardinalityAggregator` with a null values source.  I might get around to cleaning that up at some point (I'd like to) but I didn't do it in this pass.

server/src/main/java/org/elasticsearch/search/aggregations/metrics/CardinalityAggregatorFactory.java
not-napoleon: Yeah, I went back and forth on what to call this.  The next step of this fix is going to be to break this one up, but I didn't want to make that change along with the refactor.  I'll rename it for now, and we can take another pass at it in a few days.

server/src/main/java/org/elasticsearch/search/aggregations/metrics/CardinalityAggregator.java
nik9000: ah! But could it be final?

server/src/main/java/org/elasticsearch/search/aggregations/metrics/CardinalityAggregator.java
not-napoleon: Yeah, I think it's safe to get rid of the numeric count.  I'll do that.

not-napoleon:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/search/aggregations/metrics/NumericCardinalityAggregator.java
```diff
@@ -0,0 +1,53 @@
...
+    public NumericCardinalityAggregator(
+        String name,
+        ValuesSourceConfig valuesSourceConfig,
+        int precision,
+        AggregationContext context,
+        Aggregator parent,
+        Map<String, Object> metadata
+    ) throws IOException {
+        super(name, valuesSourceConfig, precision, context, parent, metadata);
+        if (valuesSourceConfig.getValuesSource()instanceof ValuesSource.Numeric numericVS) {
```
nik9000, line 38: Maybe just hard cast here? It's not going to fail unless something is seriously wrong with the values source registry. Then `source` could be final which feels nice.

nik9000 (approved):
LGTM.

PR #85842, opened 2022-04-12
Remove hppc from cat allocation api

rjernst:
The cat allocation endpoint uses an hppc int map for counting the number
of allocations for each node. There are not that many nodes, and this is
short lived anyways. This commit switches to a HashMap.

relates #84735


dakrone (approved):
LGTM

PR #85843, opened 2022-04-12
Remove hppc from tcp transport

rjernst:
The TcpTransport uses an hppc int set when falling back to
determine if there is a unique port bound. Since the expectation is this
is a single value, the map should be very small. This commit switches to
using a HashSet.

relates #84735


ChrisHegarty (approved):
LGTM

PR #85845, opened 2022-04-12
Remove SnapshotClient from HLRC

dakrone:
This removes the `SnapshotClient` from the high level rest client, rewriting the tests to use the
low-level client instead.

Relates to https://github.com/elastic/elasticsearch/issues/83423


joegallo (approved):
LGTM, your call whether my ✅ is good enough given my commits here. And also if you don't like any of my commits, we can totally undo them.

dakrone:
@elasticmachine update branch

PR #85846, opened 2022-04-12
Ack node shutdown requests after cluster state update is complete

rjernst:
The adding or removing a node to the node shutdown api stores submits a
cluster state update task to store the node metadata to be shutdown.
Sometimes a reroute is required, and in those cases we currently wait
for the reroute to complete before acknowledging the original request.
This commit changes the behavior to acknowledge success after the
cluster state update is complete. A warning is still logged if the
reroute fails.

relates #84847


x-pack/plugin/shutdown/src/main/java/org/elasticsearch/xpack/shutdown/TransportDeleteShutdownNodeAction.java
```diff
@@ -92,21 +92,15 @@ public void onFailure(Exception e) {
 
             @Override
             public void clusterStateProcessed(ClusterState oldState, ClusterState newState) {
+                listener.onResponse(AcknowledgedResponse.TRUE);
```
DaveCTurner, line 95: I suggest completing the listener _after_ submitting the task to the reroute service, so that if there's ever a need to wait for the followup reroute to complete you can safely do so with `GET _cluster/health?wait_for_events=languid`.

pgomulka (approved):
LGTM,

x-pack/plugin/shutdown/src/main/java/org/elasticsearch/xpack/shutdown/TransportDeleteShutdownNodeAction.java
DaveCTurner: I suggest completing the listener after submitting the task to the reroute service, so that if there's ever a need to wait for the followup reroute to complete you can safely do so with `GET _cluster/health?wait_for_events=languid`.

(sorry for the noise, repeating my previous comment in a new review)

DaveCTurner (changes requested):
Sorry, forgot to mark my previous review as `request changes`, resubmitting...

DaveCTurner (approved):
LGTM

PR #85847, opened 2022-04-12
Wildcard support in Logstash pipelines search API.

mashhurs:
## What is happening?
`_logstash/pipeline?id={}` API does not support searching pipelines by `wildcard IDs`.

**Why it is important?**
Currently, every Logstash instance fetches all pipelines (from `.logstash` index) every 5s (configurable) and filters locally based on Logstash instance pipeline settings. Pipeline config may include wildcard IDs. After this change, every Logstash instance should consider ROI pipelines config. That's said, LS instance sends its pipeline IDs to Elasticsearch and gets the pipeline configs only belong to itself. This dramatically decreases network load as well as gives an opportunity to control pipelines in dynamic way with wildcards.

### Issue
Closes #84772

## Does Kibana get affected by this change?
- **NO**: we are not changing API interface, extending its behavior.

## Documentation
- No change required in [Elasticsearch (Kibana CPM) ](https://www.elastic.co/guide/en/kibana/8.2/logstash-configuration-management-api.html)side since no feature affected in Kibana.
- Logstash team documents when client side behavior/code changes (after merging this PR).

## Acceptance Criteria
API should support wildcard IDs and return pipeline configs based on requested IDs (explicit and wildcard).

## Solution
We had two solution approaches:
- fetching all docs from `.logstash` index and filter based on requested ID.
- support wildcard ES query
Why we choose #option 1? - Based on team's decision ([RFC](https://docs.google.com/document/d/1NtTB0L9oBno4yEjpr8PkliFSMV8JXa5silyBizFZxJw)), customers are less affected.

#### Prerequisites
- Make sure you cloned the Elasticsearch repo
- Make sure you installed Kibana on your local machine
- Make sure you cloned the Logstash repo (required LS to ES communication)
- Make sure python is installed on your local machine
- Run Elasticsearch with `./gradlew :run`
- Go to Postman and send a request to generate Kibana token, add user name & password to Basic auth header.
`POST: localhost:9200/_security/service/elastic/kibana/credential/token/kibanaToken`
- Set received Kibana token to kibana.yml under the Kibana config folder
- Run the Kibana and go to the Stack Management -> Users page. Note that default authentication will be uid: elastic, pwd: password when accessing the Kibana.
- Create a user named logstash with logstash-admin & logstash-system roles. They will be required to directly access the .logstash index.
- Run the [python script](https://gist.github.com/mashhurs/79441a7380ae18c32931dc544ec549ab) to generate bunch of pipelines

#### Test
- Now you can go to Postman and do wildcard pipelines search, eg.:
`GET http://localhost:9200/_logstash/pipeline?id=ry*,rob*,ki*chy*`

### DIFF
--- a/x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineAction.java
+++ b/x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineAction.java
@@ -35,7 +35,10 @@ import org.elasticsearch.xpack.logstash.Logstash;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.Consumer;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static org.elasticsearch.xpack.core.ClientHelper.LOGSTASH_MANAGEMENT_ORIGIN;
@@ -43,6 +46,13 @@ import static org.elasticsearch.xpack.core.ClientHelper.LOGSTASH_MANAGEMENT_ORIG
 public class TransportGetPipelineAction extends HandledTransportAction<GetPipelineRequest, GetPipelineResponse> {
 
     private static final Logger logger = LogManager.getLogger(TransportGetPipelineAction.class);
+
+    private static final Integer SIZE = 10000;
+
+    private static final String WILDCARD = "*";
+
+    private static final Pattern WILDCARD_PATTERN = Pattern.compile("[^*]+|(\\*)");
+
     private final Client client;
 
     @Inject
@@ -53,7 +63,74 @@ public class TransportGetPipelineAction extends HandledTransportAction<GetPipeli
 
     @Override
     protected void doExecute(Task task, GetPipelineRequest request, ActionListener<GetPipelineResponse> listener) {
-        if (request.ids().isEmpty()) {
+        final Set<String> explicitPipelineIds = request.ids()
+            .stream()
+            .filter(pipeline -> pipeline.contains(WILDCARD) == false)
+            .collect(Collectors.toSet());
+
+        final Set<String> wildcardPipelineIdPatterns = request.ids()
+            .stream()
+            .filter(pipeline -> pipeline.contains(WILDCARD))
+            .map(this::toWildcardPipelineIdPattern)
+            .collect(Collectors.toSet());
+
+        if (explicitPipelineIds.size() > 0 && wildcardPipelineIdPatterns.size() == 0) {
+            getPipelinesByIds(explicitPipelineIds, listener);
+            return;
+        }
+
+        client.prepareSearch(Logstash.LOGSTASH_CONCRETE_INDEX_NAME)
+            .setSource(
+                SearchSourceBuilder.searchSource().fetchSource(true).query(QueryBuilders.matchAllQuery()).size(SIZE).trackTotalHits(true)
+            )
+            .setScroll(TimeValue.timeValueMinutes(1L))
+            .execute(ActionListener.wrap(searchResponse -> {
+                final int numHits = Math.toIntExact(searchResponse.getHits().getTotalHits().value);
+                final Map<String, BytesReference> pipelineSources = Maps.newMapWithExpectedSize(numHits);
+                final Consumer<SearchResponse> clearScroll = (response) -> {
+                    if (response != null && response.getScrollId() != null) {
+                        ClearScrollRequest clearScrollRequest = new ClearScrollRequest();
+                        clearScrollRequest.addScrollId(response.getScrollId());
+                        client.clearScroll(
+                            clearScrollRequest,
+                            ActionListener.wrap(
+                                (r) -> {},
+                                e -> logger.warn(
+                                    new ParameterizedMessage("clear scroll failed for scroll id [{}]", response.getScrollId()),
+                                    e
+                                )
+                            )
+                        );
+                    }
+                };
+                final int maxNumHitsToFetch = 0;
+                handleFilteringSearchResponse(
+                    searchResponse,
+                    pipelineSources,
+                    explicitPipelineIds,
+                    wildcardPipelineIdPatterns,
+                    maxNumHitsToFetch,
+                    clearScroll,
+                    listener
+                );
+            }, e -> handleFailure(e, listener)));
+    }
+
+    private void getPipelinesByIds(Set<String> ids, ActionListener<GetPipelineResponse> listener) {
+        client.prepareMultiGet()
+            .addIds(Logstash.LOGSTASH_CONCRETE_INDEX_NAME, ids)
+            .execute(ActionListener.wrap(mGetResponse -> {
+                logFailures(mGetResponse);
+                listener.onResponse(
+                    new GetPipelineResponse(
+                        Arrays.stream(mGetResponse.getResponses())
+                            .filter(itemResponse -> itemResponse.isFailed() == false)
+                            .filter(itemResponse -> itemResponse.getResponse().isExists())
+                            .map(MultiGetItemResponse::getResponse)
+                            .collect(Collectors.toMap(GetResponse::getId, GetResponse::getSourceAsBytesRef))
+                    ));
+            }, e -> handleFailure(e, listener)));
+        /*if (request.ids().isEmpty()) {
             client.prepareSearch(Logstash.LOGSTASH_CONCRETE_INDEX_NAME)
                 .setSource(
                     SearchSourceBuilder.searchSource()
@@ -106,7 +183,7 @@ public class TransportGetPipelineAction extends HandledTransportAction<GetPipeli
                         )
                     );
                 }, e -> handleFailure(e, listener)));
-        }
+        }*/
     }
 
     private void handleFailure(Exception e, ActionListener<GetPipelineResponse> listener) {
@@ -118,37 +195,66 @@ public class TransportGetPipelineAction extends HandledTransportAction<GetPipeli
         }
     }
 
-    private void handleSearchResponse(
+    private void handleFilteringSearchResponse(
         SearchResponse searchResponse,
         Map<String, BytesReference> pipelineSources,
+        Set<String> explicitPipelineIds,
+        Set<String> wildcardPipelineIdPatterns,
+        int maxNumHitsToFetch,
         Consumer<SearchResponse> clearScroll,
         ActionListener<GetPipelineResponse> listener
     ) {
+        maxNumHitsToFetch += searchResponse.getHits().getHits().length;
         for (SearchHit hit : searchResponse.getHits().getHits()) {
-            pipelineSources.put(hit.getId(), hit.getSourceRef());
+            if (explicitPipelineIds.isEmpty() && wildcardPipelineIdPatterns.isEmpty()) {
+                pipelineSources.put(hit.getId(), hit.getSourceRef());
+                continue;
+            }
+
+            // take if ID is in request IDs set
+            if (explicitPipelineIds.contains(hit.getId())) {
+                pipelineSources.put(hit.getId(), hit.getSourceRef());
+                continue;
+            }
+            // take if id matches request wildcard pattern
+            for (String wildcardPipelineIdPattern : wildcardPipelineIdPatterns) {
+                if (Pattern.matches(wildcardPipelineIdPattern, hit.getId())) {
+                    pipelineSources.put(hit.getId(), hit.getSourceRef());
+                    break;
+                }
+            }
         }
 
-        if (pipelineSources.size() > searchResponse.getHits().getTotalHits().value) {
+        if (maxNumHitsToFetch > searchResponse.getHits().getTotalHits().value) {
             clearScroll.accept(searchResponse);
             listener.onFailure(
                 new IllegalStateException(
                     "scrolling returned more hits ["
-                        + pipelineSources.size()
+                        + maxNumHitsToFetch
                         + "] than expected ["
                         + searchResponse.getHits().getTotalHits().value
                         + "] so bailing out to prevent unbounded "
                         + "memory consumption."
                 )
             );
-        } else if (pipelineSources.size() == searchResponse.getHits().getTotalHits().value) {
+        } else if (maxNumHitsToFetch == searchResponse.getHits().getTotalHits().value) {
             clearScroll.accept(searchResponse);
             listener.onResponse(new GetPipelineResponse(pipelineSources));
         } else {
+            int finalNumHitsToFetch = maxNumHitsToFetch;
             client.prepareSearchScroll(searchResponse.getScrollId())
                 .setScroll(TimeValue.timeValueMinutes(1L))
                 .execute(
                     ActionListener.wrap(
-                        searchResponse1 -> handleSearchResponse(searchResponse1, pipelineSources, clearScroll, listener),
+                        searchResponse1 -> handleFilteringSearchResponse(
+                            searchResponse1,
+                            pipelineSources,
+                            explicitPipelineIds,
+                            wildcardPipelineIdPatterns,
+                            finalNumHitsToFetch,
+                            clearScroll,
+                            listener
+                        ),
                         listener::onFailure
                     )
                 );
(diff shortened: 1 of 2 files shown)


x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineAction.java
```diff
@@ -32,20 +32,26 @@
...
 
 import static org.elasticsearch.xpack.core.ClientHelper.LOGSTASH_MANAGEMENT_ORIGIN;
 
 public class TransportGetPipelineAction extends HandledTransportAction<GetPipelineRequest, GetPipelineResponse> {
 
     private static final Logger logger = LogManager.getLogger(TransportGetPipelineAction.class);
+
     private final Client client;
 
+    private final Integer SIZE = 10000;
```
mashhurs, line 53: I will change this to static.

mashhurs:
I will change the `format` caused CI failure but no idea why `Task :modules:transport-netty4:internalClusterTest` is failed.

x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/util/LogstashPipelinesUtil.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public static String toRegex(String input) {
+        StringBuffer stringBuffer = new StringBuffer(input.length());
+        stringBuffer.append('^');
+        for (int i = 0; i < input.length(); i++) {
+            switch(input.charAt(i)) {
+                case '*':
+                    stringBuffer.append(".*");
+                    break;
+                case '?':
```
mashhurs, line 30: I need to remove special chars since Kibana also does not allow special chars for pipeline name.

x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineAction.java
```diff
@@ -169,4 +214,8 @@ private void logFailures(MultiGetResponse multiGetResponse) {
             logger.info("Could not retrieve logstash pipelines with ids: {}", ids);
         }
     }
+
+    private String toRegex(String input) {
+        return "^".concat(input.replace(WILDCARD, ".*$"));
```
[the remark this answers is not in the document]
mashhurs: Thanks @kaisecheng for good points.
I was also aware of your 2nd point (pipeline search worked when I sent regex patter through URL query string). Plz take a look at my latest (`Wildcard regex pattern improvement`) commit so that it covers both your concerns.

My opinion for your additional following point would be "we assume users are installed standard (not customized) Logstash and Elasticsearch and the API gets the validated pipelines by [Logstash bootstrap check](https://github.com/elastic/logstash/blob/611953bbe96b4178bfbec5029e127af880ee0fa7/x-pack/lib/config_management/bootstrap_check.rb#L19)." Let me plz know if you think we should check pattern anywhere in all pipelines related ES APIs. I think it is not necessary.
```
Currently Logstash accept alphabet, number, _, - as pipeline id and only accept * as wildcard to match any character in pipeline id.
```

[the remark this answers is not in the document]
kaisecheng: I agree with you that we don't need to give pattern checking to all APIs in this enhancement. The goal is to bring wildcard to search system index. Logstash do not have upsert feature to central management. So, aligning the search part is good enough for now. It is nice to have for future to align Kibana and Elasticsearch API.

kaisecheng (commented):
Good improvement for the wildcard handling. The search experience aligns with Logstash now. One suggestion is to have unit test for wildcard matching part.

For example
pipeline ids: super_generator, another_generator, hello
pattern: *generator
matching result: super_generator, another_generator

x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineAction.java
```diff
@@ -122,37 +141,66 @@ private void handleFailure(Exception e, ActionListener<GetPipelineResponse> list
...
+                continue;
+            }
+
+            // take if ID is in request IDs set
+            if (explicitPipelineIds.contains(hit.getId())) {
+                pipelineSources.put(hit.getId(), hit.getSourceRef());
+                continue;
+            }
+            // take if id matches request wildcard pattern
+            for (String wildcardPipelineIdPattern : wildcardPipelineIdPatterns) {
```
jbaiera, line 166: Readability nit: You could simplify this to avoid needing the break if you changed this to a stream call:


(Assuming that `wildcardPipeilneIdPatterns` is converted to a `List<Pattern>`)
```java
if (wildcardPipelineIdPatterns.stream().anyMatch(pattern -> pattern.matcher(hit.getId()).matches())) {
        // Add entry to result map
}
```

x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineAction.java
```diff
@@ -122,37 +141,66 @@ private void handleFailure(Exception e, ActionListener<GetPipelineResponse> list
...
+    private void handleFilteringSearchResponse(
         SearchResponse searchResponse,
         Map<String, BytesReference> pipelineSources,
+        Set<String> explicitPipelineIds,
+        Set<String> wildcardPipelineIdPatterns,
+        int maxNumHitsToFetch,
         Consumer<SearchResponse> clearScroll,
         ActionListener<GetPipelineResponse> listener
     ) {
+        maxNumHitsToFetch += searchResponse.getHits().getHits().length;
```
jbaiera, line 153: Since we assign a new variable later to hold this value (`finalNumHitsToFetch`) can we instead make a new variable here to hold this mutation instead of overwriting the parameter passed in?

x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineAction.java
```diff
@@ -122,37 +141,66 @@ private void handleFailure(Exception e, ActionListener<GetPipelineResponse> list
...
+            for (String wildcardPipelineIdPattern : wildcardPipelineIdPatterns) {
+                if (Pattern.matches(wildcardPipelineIdPattern, hit.getId())) {
+                    pipelineSources.put(hit.getId(), hit.getSourceRef());
+                    break;
+                }
+            }
         }
 
-        if (pipelineSources.size() > searchResponse.getHits().getTotalHits().value) {
+        if (maxNumHitsToFetch > searchResponse.getHits().getTotalHits().value) {
```
jbaiera, line 174: I think we can check for this failure scenario before iterating over the records from this scroll response. I think it'd be better to error out earlier in that case and save the work of checking against patterns.

x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineAction.java
```diff
@@ -122,37 +141,66 @@ private void handleFailure(Exception e, ActionListener<GetPipelineResponse> list
...
                         + "] so bailing out to prevent unbounded "
                         + "memory consumption."
                 )
             );
-        } else if (pipelineSources.size() == searchResponse.getHits().getTotalHits().value) {
+        } else if (maxNumHitsToFetch == searchResponse.getHits().getTotalHits().value) {
             clearScroll.accept(searchResponse);
             listener.onResponse(new GetPipelineResponse(pipelineSources));
         } else {
+            int finalNumHitsToFetch = maxNumHitsToFetch;
```
jbaiera, line 190: Variable name is a little confusing here, can we make this something like "numberOfHitsSeenSoFar"? Then the method parameter could be something like "numberOfHitsSeenPreviously"?

The name "maxNumHitsToFetch" makes me think this is a hard limit on the results, but in reality it's just a counter and the hard limit is the scroll total hits.

x-pack/plugin/logstash/src/test/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineActionTests.java
```diff
@@ -38,6 +49,7 @@
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+@ThreadLeakScope(ThreadLeakScope.Scope.NONE)
```
jbaiera, line 52: Is there something breaking these tests on thread leak detection now?

jbaiera (changes requested):
Looking good! Left a few comments

x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineAction.java
```diff
@@ -122,37 +141,66 @@ private void handleFailure(Exception e, ActionListener<GetPipelineResponse> list
...
+                pipelineSources.put(hit.getId(), hit.getSourceRef());
+                continue;
+            }
+            // take if id matches request wildcard pattern
+            for (String wildcardPipelineIdPattern : wildcardPipelineIdPatterns) {
+                if (Pattern.matches(wildcardPipelineIdPattern, hit.getId())) {
+                    pipelineSources.put(hit.getId(), hit.getSourceRef());
+                    break;
+                }
+            }
```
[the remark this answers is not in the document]
mashhurs: Thanks @kaisecheng for the details,
If you are worrying about performance, as my understanding it will not matter since JIT optimizer takes care.
From clean code perspective, it really varies from person to person. My approach is based on lambda where I use `filter` and `map` in plain loop as well. Where I am filtering by continuing and mapping at the end. If we have strict code formatting I am happy to change.

x-pack/plugin/logstash/src/test/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineActionTests.java
mashhurs: Couldn't figure out why my test case `testGetPipelinesByExplicitAndWildcardIds` is causing thread leak. Do you have any?

```
// log
com.carrotsearch.randomizedtesting.ThreadLeakError: 1 thread leaked from SUITE scope at org.elasticsearch.xpack.logstash.action.TransportGetPipelineActionTests: 
   1) Thread[id=24, name=elasticsearch[testGetPipelinesByExplicitAndWildcardIds][[timer]], state=TIMED_WAITING, group=TGRP-TransportGetPipelineActionTests]
        at java.base@17.0.2/java.lang.Thread.sleep(Native Method)
        at app//org.elasticsearch.threadpool.ThreadPool$CachedTimeThread.run(ThreadPool.java:671)

	at __randomizedtesting.SeedInfo.seed([DACFC603D5697F8A]:0)
```

x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineAction.java
```diff
@@ -122,37 +141,66 @@ private void handleFailure(Exception e, ActionListener<GetPipelineResponse> list
...
+
+            // take if ID is in request IDs set
+            if (explicitPipelineIds.contains(hit.getId())) {
+                pipelineSources.put(hit.getId(), hit.getSourceRef());
+                continue;
+            }
+            // take if id matches request wildcard pattern
+            for (String wildcardPipelineIdPattern : wildcardPipelineIdPatterns) {
+                if (Pattern.matches(wildcardPipelineIdPattern, hit.getId())) {
```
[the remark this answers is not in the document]
mashhurs: Oh thanks, hidden performance bug.

x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineAction.java
mashhurs: From paging point of view, this makes sense.

x-pack/plugin/logstash/src/test/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineActionTests.java
jbaiera: It looks like there is a ThreadPool instance that hasn't been shut down at the end of the test? Though I'm not seeing one set up at all - perhaps its something downstream of some classes in the test?

jbaiera: Ah, found it - it's the NoOpClient - it creates a TestThreadPool instance which is leaving a timer thread laying about at the end of the test. Solution is simple - close the NoOpClient in a try-with-resources block

x-pack/plugin/logstash/src/test/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineActionTests.java
```diff
@@ -103,6 +115,55 @@ public void onFailure(Exception e) {
...
+            public void onFailure(Exception e) {
+                failure.set(e);
+            }
+        };
+
+        new TransportGetPipelineAction(mock(TransportService.class), mock(ActionFilters.class), getMockClient(searchResponse)).doExecute(
+            null,
+            request,
+            testActionListener
+        );
```
jbaiera, line 162:
```suggestion
        try (Client client = getMockClient(searchResponse)) {
            new TransportGetPipelineAction(mock(TransportService.class), mock(ActionFilters.class), client).doExecute(
                null,
                request,
                testActionListener
            );
        }
```

x-pack/plugin/logstash/src/test/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineActionTests.java
jbaiera, line 52:
```suggestion
```

jbaiera (commented):
This should clear up the thread leak issues.

x-pack/plugin/logstash/src/test/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineActionTests.java
mashhurs: Can you plz explain what is happening behind the scene?

x-pack/plugin/logstash/src/test/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineActionTests.java
mashhurs: Removed this line while resolving conflicts with upstream.

x-pack/plugin/logstash/src/main/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineAction.java
```diff
@@ -118,37 +137,64 @@ private void handleFailure(Exception e, ActionListener<GetPipelineResponse> list
...
+
+            // take if ID is in request IDs set
+            if (explicitPipelineIds.contains(hit.getId())) {
+                pipelineSources.put(hit.getId(), hit.getSourceRef());
+                continue;
+            }
+            // take if id matches request wildcard pattern
+            if (wildcardPipelinePatterns.stream().anyMatch(pattern -> pattern.matcher(hit.getId()).matches())) {
+                pipelineSources.put(hit.getId(), hit.getSourceRef());
+            }
```
jbaiera, line 178: There's a lot of repetition here - Each of these conditionals has the same body - can we combine the conditions together into one boolean?

jbaiera (approved):
LGTM - Left a small comment about code repetition

x-pack/plugin/logstash/src/test/java/org/elasticsearch/xpack/logstash/action/TransportGetPipelineActionTests.java
jbaiera: The thread leak detection is taking a snapshot of all threads currently created before the test and when the test completes. If there are any threads that are left lying around, it throws an assertion error. Hence the failures. 

The `getMockClient()` method is returning a new instance of `NoOpClient`. Inside of its constructor, it is creating a new instance of `TestThreadPool` which is a subclass of `ThreadPool`. The `ThreadPool` instance creates a number of executors, schedulers, and threads (like the time caching thread). All of those threads need to be shutdown before the end of the test case. Putting client in the try-with-resources block will ensure that the client is closed after the block, which will close the thread pool underneath.

PR #85850, opened 2022-04-13
Aggregation Execution Context add timestamp provider

weizijun:
As LeafWalker already get the timestamp from doc values, It can add a timestampProvider in the AggregationExecutionContext. 
It's useful in the time serise downsampling sense. downsampling collect can get the timestamp from the AggregationExecutionContext.
And It will have other more use case.


csoulios:
@elasticmachine ok to test

csoulios (approved):
LGTM!

This is a good improvement, that will definitely be used by the downsampling operation, which loads timestamps. I expect this to also improve performance.

csoulios:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/search/aggregations/AggregationExecutionContext.java
```diff
@@ -23,11 +23,17 @@
...
 
     private final CheckedSupplier<BytesRef, IOException> tsidProvider;
+    private final CheckedSupplier<Long, IOException> timestampProvider;
     private final LeafReaderContext leafReaderContext;
 
-    public AggregationExecutionContext(LeafReaderContext leafReaderContext, CheckedSupplier<BytesRef, IOException> tsidProvider) {
+    public AggregationExecutionContext(
+        LeafReaderContext leafReaderContext,
+        CheckedSupplier<BytesRef, IOException> tsidProvider,
+        CheckedSupplier<Long, IOException> timestampProvider
```
imotov, line 32: Why is this a checked supplier? Under what condition can it ever throw IOException?

imotov (approved):
LGTM, except I don't think we need IOException there. I suspect we can convert CheckedSupplier to Supplier in case of tsid as well. I think it is CheckedSupplier for purely historical reasons.

server/src/main/java/org/elasticsearch/search/aggregations/AggregationExecutionContext.java
weizijun: > LGTM, except I don't think we need IOException there. I suspect we can convert CheckedSupplier to Supplier in case of tsid as well. I think it is CheckedSupplier for purely historical reasons.

Yeah, I also think Supplier is better, I have converted CheckedSupplier to Supplier for timestamp and tsid.

weizijun:
@elasticmachine update branch

csoulios:
Thank you for this contribution @weizijun

weizijun:
> Thank you for this contribution @weizijun

Thanks @imotov , @csoulios !

PR #85855, opened 2022-04-13
Reduce cli tests calling main directly

rjernst:
The main method of Command is not normally called by cli tests. Instead
they call the execute helper which calls mainWithoutErrorHandling.
Sometimes, though, it is desirable to test the top level behavior. This
commit cleans up the base class tests to not call main directly. This
will help with future refactorings to change the signature of main, so
that less uses need to be changed.


grcevski (approved):
LGTM! Very cool refactor!

pugnascotia (approved):

PR #85863, opened 2022-04-13
[ML] Fix serialisation of text embedding updates

davidkyle:
Fixes an issue where text_embeddding updates could not be serialised between nodes.


droberts195 (approved):
LGTM

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/MlInferenceNamedXContentProvider.java
```diff
@@ -603,7 +603,7 @@ public List<NamedWriteableRegistry.Entry> getNamedWriteables() {
             new NamedWriteableRegistry.Entry(
                 InferenceConfigUpdate.class,
                 TextEmbeddingConfigUpdate.NAME,
-                TextClassificationConfigUpdate::new
+                TextEmbeddingConfigUpdate::new
```
benwtrent, line 606: 🤦

benwtrent (approved):

PR #85867, opened 2022-04-13
[ML] add nlp config update serialization tests

benwtrent:
adds inference request serialization tests that include nlp config updates.

relates: https://github.com/elastic/elasticsearch/pull/85863

PR #85870, opened 2022-04-13
[ML] Tests of the named writable registry for inference updates

davidkyle:
Same motivation as #85867, different approach.

Related to #85863


benwtrent (approved):
Oooo! I like this. So clean. It will be good to include along side: https://github.com/elastic/elasticsearch/pull/85867

One for a focused serialization to capture all updates, another to exercise the inference request objects.

davidkyle:
@elasticmachine update branch

PR #85871, opened 2022-04-13
Remove lazy evaluation from plugin properties building

rjernst:
Plugins and modules use PluginBuildPlugin in gradle to build their
plugin properties file. Internally that uses an afterEvaluate since the
plugin properties extension will not have been filled in yet when the
plugin is first applied. This commit removes the afterEvaluate in favor
of configuring the plugin properties map for filling in the template
within the pluginProperties task configuration, which is applied when
the task is actually realized in the task graph.


breskeby (approved):
lgtm

PR #85872, opened 2022-04-13
[ML] Replace the implementation of the categorize_text aggregation

droberts195:
This replaces the implementation of the `categorize_text` aggregation
with the new algorithm that was added in #80867. The new algorithm
works in the same way as the ML C++ code used for categorization jobs.

The docs are updated to reflect the workings of the new implementation.


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/aggs/categorization/CategorizeTextAggregationBuilder.java
```diff
@@ -362,6 +315,6 @@ public String getType() {
 
     @Override
     public Version getMinimalSupportedVersion() {
-        return Version.V_7_16_0;
+        return Version.V_8_3_0;
```
benwtrent, line 318: How does this work when somebody does `categorize_text` from a 7.16 node and it attempts to serialize to a 8.3 node? 

I know this minimal support version prevents us from WRITING to older nodes, but what if an older node is writing to it? 

We may need to add a clause at the top of stream input parsing to make sure the input stream is at least 8.3?

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/aggs/categorization/InternalCategorizationAggregation.java
```diff
@@ -23,143 +20,58 @@
...
 import java.util.Comparator;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
-import static org.elasticsearch.xpack.ml.aggs.categorization.CategorizationBytesRefHash.WILD_CARD_REF;
-
 public class InternalCategorizationAggregation extends InternalMultiBucketAggregation<
```
benwtrent, line 34: My main concern is that, for some weird reason, older nodes did the aggregating and the coordinator is trying to read in old InternalCategorizationAggregation objects. 

I would hope the `getMinimalSupportedVersion` in the builder would prevent this as the builder implies the factory is created on that node. Which means that only nodes after 8.3.0 would receive the new builder and reply back.

droberts195: Good point.  I've made the stream input and output safe for older nodes.

In the case of a coordinating node on an older version it's possible for the newer nodes to use the new algorithm and then serialize their results in the old format.  So that scenario can work to some extent - probably not brilliantly as the merging might not work well for categories created in completely different ways, but better than nothing.

Like you say, the other way around should not be possible because of the minimal supported version, but I've made it safe if there's some unforeseen loophole.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/aggs/categorization/CategorizeTextAggregationBuilder.java
droberts195: This is a good point.

While I was making changes I realised that a newer node _can_ actually use the new algorithm and respond to an older coordinating node in the format that the older node wants.  So I have implemented that.  The merging of categories on the coordinating node might not work brilliantly if different nodes have created their local categories in different ways, but it should be better than nothing.

With the changes of b3fe740f3dac21d207e0f42e1afd8cceab5752e6 I think the communications between old and new nodes should at least be safe, even though the results might be a bit strange sometimes.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/aggs/categorization/CategorizeTextAggregationBuilder.java
```diff
@@ -116,6 +118,13 @@ public CategorizeTextAggregationBuilder(StreamInput in) throws IOException {
         super(in);
         this.bucketCountThresholds = new TermsAggregator.BucketCountThresholds(in);
         this.fieldName = in.readString();
+        // If the coordinating node is an older version then we might still receive messages from older
+        // nodes. In this case we can send back results for this node created using the new algorithm.
+        // They won't necessarily merge well with results from other nodes, but are better than nothing.
```
benwtrent, line 123: Since its experimental, I think we should throw. This is not a critical functionality, nothing systemic is using this aggregation. We should throw, especially since running in a mixed cluster environment is a temporary situation.

benwtrent: The new internal agg implementation serializing back to the old coordinator will throw anyways (due to things not being read of the wire well). It would be cheaper for the users (and more obvious the cause of the issue) if we just throw.

droberts195: > The new internal agg implementation serializing back to the old coordinator will throw anyways (due to things not being read of the wire well).

It won't throw on the old coordinator with the changes of b3fe740f3dac21d207e0f42e1afd8cceab5752e6 - I changed the new nodes to serialize in the format the old nodes will expect.

But still, you're probably right that a simple "not supported in mixed cluster" exception is better as it's less likely to cause people to waste time debugging strange results.  I'll change the 4 methods to throw exceptions instead.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/aggs/categorization/CategorizeTextAggregationBuilder.java
```diff
@@ -269,6 +280,16 @@ protected CategorizeTextAggregationBuilder(
 
     @Override
     protected void doWriteTo(StreamOutput out) throws IOException {
+        // Disallow this aggregation in mixed version clusters that cross the algorithm change boundary.
+        if (out.getVersion().before(ALGORITHM_CHANGED_VERSION)) {
```
benwtrent, line 284: we MAY get this for free with the versioned named writeable stuff. But, keeping this here is cool with me.

benwtrent (approved):

droberts195:
@elasticmachine update branch

droberts195:
@elasticmachine update branch

droberts195:
@elasticmachine update branch

droberts195:
@elasticmachine update branch

droberts195:
https://github.com/elastic/elasticsearch/pull/85872/commits/d5b01130a66a22343c8ebb3ad2a72f6e020dea3a and https://github.com/elastic/elasticsearch/pull/85872/commits/758938787b1414c19142241bd7364f6ebb20b6f6 implement the changes of elastic/ml-cpp#2277 on the Java side.

PR #85877, opened 2022-04-13
[DOCS] Add manage_enrich role to built-in roles

lockewritesdocs:
Adds the `manage_enrich` user to the list of built-in roles.


ywangd (approved):
LGTM

The title says "... to built-in _roles_" but the issue description says "... to the list of built-in _users_". I think we should correct the issue description.

Also ping @elastic/clients-team because I believe manage_enrich is also missing from elasticsearch-specification.

sethmlarson:
Thanks for the ping @ywangd, [could you open a `missing type` issue](https://github.com/elastic/elasticsearch-specification/issues) pointing to this PR?

lockewritesdocs:
>The title says "... to built-in roles" but the issue description says "... to the list of built-in users". I think we should correct the issue description.

@ywangd, I updated the description to match the title 👍

PR #85888, opened 2022-04-14
Remove hppc from multi*shard request and responses

rjernst:
The Multi API requests and responses use an hppc int array for
term location data. The use of hppc here dates back to the beginning of
the multi apis, yet all the other members have been converted to Java
collections. This commit converts these uses to ArrayLists.

relates #84735


grcevski (approved):
LGTM!

PR #85889, opened 2022-04-14
Remove hppc from task manager

rjernst:
The task manager uses hppc to keep track of the count of child tasks for
each network connection. This is bookkeeping code and not performance
critical, nor is it memory intensive (the map won't be that large). This
commit converts the code to use a HashMap.

relates #84735

### DIFF
--- a/server/src/main/java/org/elasticsearch/tasks/TaskManager.java
+++ b/server/src/main/java/org/elasticsearch/tasks/TaskManager.java
@@ -8,9 +8,6 @@
 
 package org.elasticsearch.tasks;
 
-import com.carrotsearch.hppc.ObjectIntHashMap;
-import com.carrotsearch.hppc.ObjectIntMap;
-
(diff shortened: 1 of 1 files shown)


rjernst:

grcevski (commented):
This looks good to me! The only issue I see is that TaskManagerTests has no coverage for the register and unregister methods, where we bump and decrease the task counts. Do we have other tests that would validate the code?

rjernst:
@grcevski  You are right TaskManagerTests does not explicitly call unregister (though it does call register). The task framework is used indirectly by a lot of tests. I agree It would be much better to have a unit test of this functionality, but these tests are a bit complex right now, and this change is just a trivial type change. I'm going to go ahead and merge, and will look to refactor these tests for better direct coverage in the future.

grcevski (approved):
LGTM! I agree, I'll take on the task to add some tests for this code.

PR #85890, opened 2022-04-14
User Profile - Add hint support to SuggestProfile API

ywangd:
This PR adds both uids and labels hints to SuggestProfile API. It also
add supports to the data parameter to request body (in addition to URL
query parameter) and fuzzy search for name.

PS: Changes to yaml tests and docs will be addressed in a separate PR to reduce scope.


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -254,6 +236,50 @@ public void setEnabled(String uid, boolean enabled, RefreshPolicy refreshPolicy,
...
+        }
+        final SuggestProfilesRequest.Hint hint = request.getHint();
+        if (hint != null) {
+            if (hint.getUids() != null) {
+                assert false == hint.getUids().isEmpty() : "uids hint cannot be empty";
+                query.should(QueryBuilders.termsQuery("user_profile.uid", hint.getUids()));
+            }
+            if (hint.getLabels() != null) {
+                assert hint.getLabels().size() == 1 : "labels hint support exactly one key";
+                final String labelKey = hint.getLabels().keySet().iterator().next();
```
n1v0lg, line 264: Sanity check: it's the intention here that if we somehow end up with no labels, we throw `NoSuchElelement` here, correct? 

Would consider encapsulating this into a `getSingleLabel()` method on the Hint class, together with the assertion, and possibly throw an `IllegalStateException` with a "labels hint support exactly one key" error message there

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
```diff
@@ -66,6 +82,95 @@ public ActionRequestValidationException validate() {
         if (size < 0) {
             validationException = addValidationError("[size] parameter cannot be negative but was [" + size + "]", validationException);
         }
+        if (hint != null) {
+            validationException = validateHint(validationException);
+        }
+        return validationException;
+    }
+
+    private ActionRequestValidationException validateHint(ActionRequestValidationException validationException) {
```
n1v0lg, line 91: Optional but feels like this belongs on the `Hint` class

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -254,6 +236,50 @@ public void setEnabled(String uid, boolean enabled, RefreshPolicy refreshPolicy,
...
+                ).type(MultiMatchQueryBuilder.Type.BOOL_PREFIX).fuzziness(Fuzziness.AUTO)
+            );
+        }
+        final SuggestProfilesRequest.Hint hint = request.getHint();
+        if (hint != null) {
+            if (hint.getUids() != null) {
+                assert false == hint.getUids().isEmpty() : "uids hint cannot be empty";
+                query.should(QueryBuilders.termsQuery("user_profile.uid", hint.getUids()));
+            }
+            if (hint.getLabels() != null) {
```
n1v0lg, line 262: Optional: we are using `hint.getLabels()` (and `getUids()`) multiple times, so would consider pulling those into vars

n1v0lg (approved):
LGTM. A few optional nits and clarifications (primarily for my own sake) but nothing blocking.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
```diff
@@ -66,6 +82,95 @@ public ActionRequestValidationException validate() {
...
+        private final Map<String, List<String>> labels;
+
+        @SuppressWarnings("rawtypes")
+        public Hint(List<String> uids, Map<String, Object> labelsInput) {
+            this.uids = uids == null ? null : List.copyOf(uids);
+            // Convert the data type. Business logic, e.g. single key, is enforced in request validation
+            if (labelsInput != null) {
+                final HashMap<String, List<String>> labels = new HashMap<>();
+                for (Map.Entry<String, Object> entry : labelsInput.entrySet()) {
+                    final Object value = entry.getValue();
```
n1v0lg, line 132: Very optional but I think following structure is a little easier to parse:

```java
if (value instanceof final String stringValue) {
    labels.put(entry.getKey(), List.of(stringValue));
} else if (value instanceof final List listValue) {
    final ArrayList<String> values = new ArrayList<>();
    for (Object v : listValue) {
        if (v instanceof final String stringValue) {
            values.add(stringValue);
        } else {
            throw new IllegalArgumentException("[labels] hint supports either string value or list of strings");
        }
    }
    labels.put(entry.getKey(), List.copyOf(values));
} else {
    throw new IllegalArgumentException("[labels] hint supports either string or list of strings as its value");
}
```

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
```diff
@@ -66,6 +82,95 @@ public ActionRequestValidationException validate() {
...
+                    if (false == value instanceof String && false == value instanceof List) {
+                        throw new IllegalArgumentException("[labels] hint supports either string or list of strings as its value");
+                    }
+
+                    if (value instanceof final List listValue) {
```
n1v0lg, line 137: If we use `List<?>` here, can we remove the `@SuppressWarnings("rawtypes")`?

docs/changelog/85890.yaml
```diff
@@ -0,0 +1,5 @@
+pr: 85890
+summary: User Profile - Add hint supprot `SuggestProfile` API
```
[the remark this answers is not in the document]
ywangd: Updated

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
```diff
@@ -66,6 +82,95 @@ public ActionRequestValidationException validate() {
...
+            }
+        }
         return validationException;
     }
+
+    public static class Hint implements Writeable {
+        @Nullable
+        private final List<String> uids;
+        @Nullable
+        private final Map<String, List<String>> labels;
```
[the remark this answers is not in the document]
ywangd: The single key for the `labels` field is an intentional restriction that we agreed to keep the scope as minimal as possible to cater for only known, concrete requirements. We might decide to allow more keys in the future. But it's hard to tell for now. There is also inherent limitation on the expressiveness of a map structure. For example, we often interpret the relationship between multiple keys to be "AND". So there is no easy way to express "OR". 

> If that's not the case, might go with a record Label with name and values fields, or possibly including those fields on the Hint class directly. That way the structure is more explicit and validation and accessing labels gets simpler.

For just a single key, this makes sense. I considered that but didn't want to let the "design level restriction" impact on the underlying data struture. I have no strong opinion. If Albert also thinks it's better to get rid of the Map, I'll be happy to follow the suggestion.

> If we want to support multiple labels in the future, wondering what that will look like in terms of bwc
In this particular case, bwc is not a practical issue because the request should never be sent across nodes (also transport client is no longer a thing in 8.x). But if we do encouter such situation where the remote node cannot support what the local node requires, the common strategy is to throw error.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/rest/action/profile/RestSuggestProfilesAction.java
```diff
@@ -55,16 +66,28 @@ public String getName() {
 
     @Override
     protected RestChannelConsumer innerPrepareRequest(RestRequest request, NodeClient client) throws IOException {
-        final Set<String> dataKeys = Strings.tokenizeByCommaToSet(request.param("data", null));
         final Payload payload = request.hasContentOrSourceParam()
             ? PARSER.parse(request.contentOrSourceParamParser(), null)
-            : new Payload(null, null);
+            : new Payload(null, null, null, null);
+
+        final String data = request.param("data", null);
```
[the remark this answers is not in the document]
ywangd: Having `data` as a query parameter is for consistency with the GetProfileApi which does not support request body. Having it in the request body is to support having one place to specify all necessary parameters. I think each has its own usages. But we can choose to keep just one if you feel strong about it.

Conceptually, `source` and request body are really the same thing. I don't think you can provide both `source` parameter and request body: Request body takes precedence, but any unconsumed query parameter also results into error. We accept `source` as an alternative to request body because not all HTTP client supports GET with request body. So it is more for compatibility and I don't consider it as an alternative way for specifying requests (conceptually).

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/rest/action/profile/RestSuggestProfilesAction.java
```diff
@@ -55,16 +66,28 @@ public String getName() {
...
+        if (data != null && payload.data() != null) {
+            throw new IllegalArgumentException(
+                "The [data] parameter must be specified in either request body or as query parameter, but not both"
```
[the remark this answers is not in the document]
ywangd: I think this should be covered already. The `payload` is parsed from `contentOrSourceParamParser` which covers both request body and `source` query parameter. As said above, if `source` is not consumed (because request body is provided), the whole request should error out.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/rest/action/profile/RestSuggestProfilesAction.java
```diff
@@ -31,12 +32,22 @@ public class RestSuggestProfilesAction extends SecurityBaseRestHandler {
...
     static final ConstructingObjectParser<Payload, Void> PARSER = new ConstructingObjectParser<>(
         "suggest_profile_request_payload",
-        a -> new Payload((String) a[0], (Integer) a[1])
+        a -> new Payload((String) a[0], (Integer) a[1], (PayloadHint) a[2], (String) a[3])
+    );
+
+    @SuppressWarnings("unchecked")
+    static final ConstructingObjectParser<PayloadHint, Void> HINT_PARSER = new ConstructingObjectParser<>(
+        "suggest_profile_request_payload_hint",
+        a -> new PayloadHint((List<String>) a[0], (Map<String, Object>) a[1])
```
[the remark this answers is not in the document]
ywangd: Yes using `Object` is due to having a mix of `List<String>` and `String`. Otherwise we can parse for more specific Maps, e.g. `Map<String, List<String>>`. But the error message will be different, parsing error vs validation error. Often times, validation error is more specific and easier to understand as oppose to parsing error which just tells you something like "expecting a start of object, but get xxx" etc.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -254,6 +236,50 @@ public void setEnabled(String uid, boolean enabled, RefreshPolicy refreshPolicy,
...
+                assert hint.getLabels().size() == 1 : "labels hint support exactly one key";
+                final String labelKey = hint.getLabels().keySet().iterator().next();
+                final List<String> labelValues = hint.getLabels().get(labelKey);
+                if (labelValues.size() == 1) {
+                    query.should(QueryBuilders.termQuery("user_profile.labels." + labelKey, labelValues.get(0)));
+                } else {
+                    query.should(QueryBuilders.termsQuery("user_profile.labels." + labelKey, labelValues));
```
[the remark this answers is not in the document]
ywangd: 👍

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
ywangd: `labels` itself can be `null` but if it is not `null`, it must have exactly one key/value pair. 

Having `getSingleLabel()` is a good idea. I added it and it also allowed me to remove the more general `getLabels` method.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
ywangd: Agreed!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
ywangd: 👍

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
ywangd: Indeed. Thanks!

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
ywangd: 👍

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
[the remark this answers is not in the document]
n1v0lg: Thanks for the context 👍 

>> If Albert also thinks it's better to get rid of the Map, I'll be happy to follow the suggestion.

Sounds good! I don't feel strongly either way, esp. if since based on what you said there _is_ a chance we will extend this to necessitate a map in the future

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/rest/action/profile/RestSuggestProfilesAction.java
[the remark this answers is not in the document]
n1v0lg: Coolio, clarification makes sense. Since `source` is more about compatibility, good with supporting `data` as a param.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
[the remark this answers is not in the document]
albertzaharovits: > If Albert also thinks it's better to get rid of the Map, I'll be happy to follow the suggestion.

I think it's OK like it is.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
```diff
@@ -66,6 +83,96 @@ public ActionRequestValidationException validate() {
...
+                    return addValidationError(
+                        "[labels] hint supports a single key, got [" + Strings.collectionToCommaDelimitedString(labels.keySet()) + "]",
+                        validationException
+                    );
+                }
+
+                final List<String> values = labels.values().iterator().next();
+                if (values.isEmpty()) {
+                    return addValidationError("[labels] hint cannot be empty", validationException);
+                }
```
albertzaharovits, line 173: Maybe we should also validate that `labels.keys().iterator().next()` (and values too) does not contain `*`?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
```diff
@@ -66,6 +83,96 @@ public ActionRequestValidationException validate() {
...
+                        for (Object v : listValue) {
+                            if (v instanceof final String stringValue) {
+                                values.add(stringValue);
+                            } else {
+                                throw new IllegalArgumentException("[labels] hint supports either string value or list of strings");
+                            }
+                        }
+                        labels.put(entry.getKey(), List.copyOf(values));
+                    } else {
+                        throw new IllegalArgumentException("[labels] hint supports either string or list of strings as its value");
```
albertzaharovits, line 118: I think throwing `ElasticsearchParseException` is used in general for these, but it's mostly equivalent with `IllegalArgumentException`.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -254,6 +237,50 @@ public void setEnabled(String uid, boolean enabled, RefreshPolicy refreshPolicy,
...
+                query.should(QueryBuilders.termsQuery("user_profile.uid", hintedUids));
+            }
+            final Tuple<String, List<String>> label = hint.getSingleLabel();
+            if (label != null) {
+                final List<String> labelValues = label.v2();
+                if (labelValues.size() == 1) {
+                    query.should(QueryBuilders.termQuery("user_profile.labels." + label.v1(), labelValues.get(0)));
+                } else {
+                    query.should(QueryBuilders.termsQuery("user_profile.labels." + label.v1(), labelValues));
+                }
```
albertzaharovits, line 271: I wonder why the term/terms distinction for "labels" but not for "uid".

albertzaharovits (approved):
LGTM

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
ywangd: We do want to support `*` in the value because Kibana needs it to indicate access to all spaces (existing and future). 

We can prevent `*` in the key since it is [not supported by flattened field type](https://www.elastic.co/guide/en/elasticsearch/reference/current/flattened.html#supported-operations) anyway. Thanks

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/profile/SuggestProfilesRequest.java
ywangd: Updated

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
ywangd: In short, I forgot. But on a second thought, the distincation does not seem necessary. I remember seeing this pattern in other places and I assumed it is for performance. But I cannot find it in other places now and there are examples where "terms" is used for even for a single value. So I updated to always use termsQuery.

docs/changelog/85890.yaml
```diff
@@ -0,0 +1,5 @@
+pr: 85890
+summary: User Profile - Add hint support tp SuggestProfiles API
```
ywangd, line 2:
```suggestion
summary: User Profile - Add hint support to SuggestProfiles API
```

PR #85895, opened 2022-04-14
Add benchmarks for FieldMappers

idegtiarenko:
Add a jmh benchmarks for FieldMappers that allow us to investigate
performance and hotspots of the field mappers related code.


benchmarks/src/main/java/org/elasticsearch/benchmark/index/mapper/BeatsMapperBenchmark.java
```diff
@@ -0,0 +1,205 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+
+@Fork(value = 1) // jvmArgs = { "-XX:+PrintCompilation", "-XX:+UnlockDiagnosticVMOptions", "-XX:+PrintInlining" })
```
original-brownbear, line 62: We can probably remove the comment here. It's not the job of the benchmark to document how to toy around with this kind of thing IMO :)

benchmarks/src/main/java/org/elasticsearch/benchmark/index/mapper/BeatsMapperBenchmark.java
```diff
@@ -0,0 +1,205 @@
...
+    @SuppressWarnings("varargs")
+    private <T> T randomFrom(T... items) {
+        return items[random.nextInt(items.length)];
+    }
+
+    private static Path pathToResource(String path) throws URISyntaxException {
+        return Paths.get(BeatsMapperBenchmark.class.getResource(path).toURI());
+    }
+
+    private static MapperService createMapperService(String mappings) {
```
original-brownbear, line 160: I think we can add a TODO here at least, asking for this to be dried up with similar logic in the benchmark we copied this from?

benchmarks/src/main/resources/org/elasticsearch/benchmark/index/mapper/sample_documents.json
```diff
@@ -0,0 +1,25 @@
+{"@timestamp": 893964617, "client.ip": "40.135.0.0", "http.request.method": "GET", "url.path": "/images/hm_bg.jpg HTTP/1.0", "http.response.status_code": 200, "http.response.bytes": 24736}
```
original-brownbear, line 1: Should we generate these in some form as well like we do in the other benchmark maybe? This seems a little noisy when we already have code to generate documents anyway?

original-brownbear (commented):
Looks really nice, just a few smaller comments.
I also wonder if we should store the huge JSON mapping as a compressed binary after all (sorry for getting back to this I know we had this discussion before).
In the end, the json will never be modified on a line-by-line basis, so having it as outright text really just blows up the size of the repo by a couple more KB than necessary right?

idegtiarenko:

benchmarks/src/main/java/org/elasticsearch/benchmark/index/mapper/BeatsMapperBenchmark.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private static String readSampleMapping() throws IOException {
+        // Uncompressed mapping is around 1mb and 29k lines.
+        // It is unlikely that it will be modified so keeping the compressed version instead to minimize the repo size.
+        return readCompressedMapping("filebeat-mapping-8.1.2.json.gz");
+    }
+
+    private static String readCompressedMapping(String resource) throws IOException {
+        try (var in = new InputStreamReader(new GZIPInputStream(BeatsMapperBenchmark.class.getResourceAsStream(resource)), UTF_8)) {
```
original-brownbear, line 67: NIT: no need to optimize anything here IMO :)

```java
    private static String readCompressedMapping(String resource) throws IOException {
        try (var in = new GZIPInputStream(BeatsMapperBenchmark.class.getResourceAsStream(resource))) {
            return new String(in.readAllBytes(), UTF_8);
        }
    }
```

should do just fine

original-brownbear (approved):
LGTM, thanks!

PR #85898, opened 2022-04-14
Has privileges API for profiles

albertzaharovits:
This introduces a new Security API (`_security/profile/_has_privileges`) that can be used to verify which Users, given their associated [User Profiles](https://www.elastic.co/guide/en/elasticsearch/reference/master/user-profile.html#user-profile) (multiple profile uids can be specified in a single request), have the requested privileges.

This is analogous to the existing [Has privileges API | Elasticsearch Guide [8.1] | Elastic](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-has-privileges.html). It also uses the same format for specifying the privileges to be checked, and should be used in the same situations (ie to run an authorization preflight check or to verify privileges over application resources). However, unlike the existing has privilege API, this can be used to check the privileges of **multiple** users (not of the currently authenticated one), but the users must have an existing profile, and the response is binary only (has or has not the requested privileges).
Calling this API requires the `manage_user_profile` cluster privilege.


albertzaharovits:
@elasticmachine update branch

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/RBACEngine.java
```diff
@@ -141,6 +139,11 @@ public void resolveAuthorizationInfo(RequestInfo requestInfo, ActionListener<Aut
         );
     }
 
+    @Override
+    public void resolveAuthorizationInfo(Subject subject, ActionListener<AuthorizationInfo> listener) {
+        rolesStore.getRole(subject, listener.map(role -> new RBACAuthorizationInfo(role, role)));
+    }
```
ywangd, line 145: Not related to this PR, but this usage shows why I didn't remove the `AnonymousUser` argument from `Subject.getRoleReferenceIntersection`. As in its current state, this works correctly because anonymous user roles are handled inside `rolesStore.getRole` and the caller does not have to think about it. If we remove it, this means the subject must be prepared with anonymous roles before calling to this method which could be trappy.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportProfileHasPrivilegesAction.java
```diff
@@ -0,0 +1,131 @@
...
+                                logger.debug(
+                                    new ParameterizedMessage("Failed to check privileges for profile [{}]", profileUid),
+                                    checkPrivilegesException
+                                );
+                                errorProfiles.add(profileUid);
+                                if (counter.decrementAndGet() == 0) {
+                                    allDone.run();
+                                }
+                            })
+                        );
```
ywangd, line 115: I am surprised that we don't have an existing ActionListener subtype that does this aggregation and handling. Maybe worthwhile to extract the pattern here?

ywangd (commented):
The changes look mostly reasonable to me. Other than following comments, I have one question about changes to the AuthorizationEngine interface: Should we try maintain existing methods and provide default implementation of new methods as much as possible (the default implementation can even just be throwing errors but at least won't cause compilation errors)? Not sure if it is worth the effort?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportProfileHasPrivilegesAction.java
albertzaharovits: Honestly I haven't even looked for one. I know about the `GroupedActionListener` but that one is more strict about the erros. I think the abstraction we would need here would have to be a base class for `GroupedActionListener`.

Although I like the idea, I feel like I've expended the inovation budged with this PR, and would propose to defer it for another time.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -101,13 +106,32 @@ public ProfileService(Settings settings, Clock clock, Client client, SecurityInd
...
 
-    public void getProfile(String uid, @Nullable Set<String> dataKeys, ActionListener<Profile> listener) {
+    public void getProfile(String uid, Set<String> dataKeys, ActionListener<Profile> listener) {
         getVersionedDocument(
             uid,
             listener.map(versionedDocument -> versionedDocument != null ? versionedDocument.toProfile(dataKeys) : null)
         );
     }
 
+    public void getProfileSubjects(Collection<String> uids, ActionListener<Tuple<Map<String, Subject>, List<String>>> listener) {
```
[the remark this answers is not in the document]
albertzaharovits: > If we return Profile instead of Subject, we can replace Map<String, Subject> with List<Subject>.

We discussed that this option would require retrieving the `Subject` of a `Profile`, which is a less favorable option (`Subject` is internal-ish and `Profile` is external facing).

> It may still be cumbersome to have Tuple<List<Subject>, List<String>>. We probably want to encapsulate them into its own class which can also provide an opportunity to return "not-found" UIDs in addition to errored ones (which, practicially, are more likely to happen).

I agree, pushed 55d6d442a441094b .

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/ProfileHasPrivilegesRequestBuilder.java
```diff
@@ -0,0 +1,74 @@
...
+public class ProfileHasPrivilegesRequestBuilder extends ActionRequestBuilder<ProfileHasPrivilegesRequest, ProfileHasPrivilegesResponse> {
+
+    public ProfileHasPrivilegesRequestBuilder(ElasticsearchClient client) {
+        super(client, ProfileHasPrivilegesAction.INSTANCE, new ProfileHasPrivilegesRequest());
+    }
+
+    public ProfileHasPrivilegesRequestBuilder parse(XContentParser parser) throws IOException {
+        XContentParser.Token token = parser.nextToken();
+        if (token != XContentParser.Token.START_OBJECT) {
+            throw new ElasticsearchParseException(
```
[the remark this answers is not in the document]
albertzaharovits: I agree, pushed [08b7488](https://github.com/elastic/elasticsearch/pull/85898/commits/08b7488a721aa9ffb7beecb6c89a3d00a59f839e) .

albertzaharovits:
Thank you Yang for the review on a draft!

albertzaharovits:
> I have one question about changes to the AuthorizationEngine interface: Should we try maintain existing methods and provide default implementation of new methods as much as possible (the default implementation can even just be throwing errors but at least won't cause compilation errors)? Not sure if it is worth the effort?

I don't think it is worth it, but we agreed that we should implicate the broader team.
My reasoning is that I'm not aware of any hard **requirement** that we must not introduce changes that result in compilation failures from plugin authors (plugins must be recompiled for every version of Elasticsearch).

albertzaharovits:
@ywangd I've marked this ready for review, though it's missing a couple of less imortant tests.
I'll work on those in parallel to your review round.

Note that I don't plan to add docs and rest specs in this PR, but in the follow-up.

ywangd:
> Note that I don't plan to add docs and rest specs in this PR, but in the follow-up.

That is my preference as well. 👍

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java
```diff
@@ -171,30 +168,42 @@ public AuthorizationService(
...
+            wrapPreservingContext(
+                listener.delegateFailure(
+                    (delegateListener, authorizationInfo) -> authorizationEngine.getUserPrivileges(
+                        authorizationInfo,
+                        wrapPreservingContext(listener, threadContext)
+                    )
+                ),
+                threadContext
+            )
+        );
```
ywangd, line 206: The existing code gets `authorizationInfo` from the `threadContext` instead of computing it again. I think we should maintain this behaviour. It might be necessary to compute it if we support retrieving privileges by profileId in future. But we can update the code here when it is needed.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -305,6 +329,52 @@ private void getVersionedDocument(String uid, ActionListener<VersionedDocument>
         });
     }
 
+    private void getVersionedDocuments(
+        Collection<String> uids,
+        ActionListener<Tuple<Map<String, VersionedDocument>, List<String>>> listener
```
ywangd, line 334: We can simplify this to
```suggestion
        ActionListener<Tuple<List<ProfileDocument>, List<String>>> listener
```

because:
1. `ProfieDocument` already has `profileUid` so no need for the map key
2. We don't need version control info in this use case so no need to wrap it within a VersionedDocument.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -679,4 +749,6 @@ Profile toProfile(Set<String> dataKeys) {
             );
         }
     }
+
+    public record MultiProfileSubjectResponse(Map<String, Subject> profileUidToSubject, List<String> failureProfileUids) {}
```
ywangd, line 753: Sorry to drag on this, but can we return `ProfileDocument` as part of this response? It can help remove the need for a Map and makes it a bit simpler to read.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/rest/action/user/RestProfileHasPrivilegesAction.java
```diff
@@ -0,0 +1,68 @@
...
+    protected RestChannelConsumer innerPrepareRequest(RestRequest restRequest, NodeClient client) throws IOException {
+        try (XContentParser parser = restRequest.contentOrSourceParamParser()) {
+            ProfileHasPrivilegesRequest request = ProfileHasPrivilegesRequest.PARSER.parse(parser, null);
+            return channel -> client.execute(ProfileHasPrivilegesAction.INSTANCE, request, new RestBuilderListener<>(channel) {
+                @Override
+                public RestResponse buildResponse(ProfileHasPrivilegesResponse response, XContentBuilder builder) throws Exception {
+                    response.toXContent(builder, ToXContent.EMPTY_PARAMS);
+                    return new BytesRestResponse(RestStatus.OK, builder);
+                }
+            });
```
ywangd, line 65: Nit
```suggestion
            return channel -> client.execute(ProfileHasPrivilegesAction.INSTANCE, request, new RestToXContentListener<>(channel));
```

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -305,6 +329,52 @@ private void getVersionedDocument(String uid, ActionListener<VersionedDocument>
...
+        ActionListener<Tuple<Map<String, VersionedDocument>, List<String>>> listener
+    ) {
+        if (uids.isEmpty()) {
+            listener.onResponse(new Tuple<>(Map.of(), List.of()));
+            return;
+        }
+        tryFreezeAndCheckIndex(listener).ifPresent(frozenProfileIndex -> {
+            new OriginSettingClient(client, SECURITY_PROFILE_ORIGIN).prepareMultiGet()
+                .addIds(frozenProfileIndex.aliasName(), uids.stream().map(ProfileService::uidToDocId).toArray(String[]::new))
+                .execute(ActionListener.wrap(multiGetResponse -> {
```
ywangd, line 343: I think we need wrap this inside a `frozenProfileIndex .checkIndexVersionThenExecute` call.

ywangd (commented):
I didn't really look into the tests. But have the following comments for the production code.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/RoleDescriptor.java
```diff
@@ -915,6 +928,10 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
             return builder.endObject();
         }
 
+        public static void write(StreamOutput out, IndicesPrivileges privileges) throws IOException {
+            privileges.writeTo(out);
+        }
```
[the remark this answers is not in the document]
albertzaharovits: It is used in `ProfileHasPrivilegesRequest#writeTo`.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/ProfileHasPrivilegesResponse.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public ProfileHasPrivilegesResponse(String[] hasPrivilegeUids, String[] errorUids) {
+        super();
+        this.hasPrivilegeUids = hasPrivilegeUids;
+        this.errorUids = errorUids;
+    }
+
+    public ProfileHasPrivilegesResponse() {
+        this(new String[0], new String[0]);
+    }
```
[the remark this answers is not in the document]
albertzaharovits: No, I have removed it.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/ProfileHasPrivilegesResponse.java
```diff
@@ -0,0 +1,84 @@
...
+    public ProfileHasPrivilegesResponse(StreamInput in) throws IOException {
+        super(in);
+        this.hasPrivilegeUids = in.readStringArray();
+        this.errorUids = in.readStringArray();
+    }
+
+    public ProfileHasPrivilegesResponse(String[] hasPrivilegeUids, String[] errorUids) {
+        super();
+        this.hasPrivilegeUids = hasPrivilegeUids;
+        this.errorUids = errorUids;
```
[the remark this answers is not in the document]
albertzaharovits: I made the change like suggested.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
albertzaharovits: Good tip, thank you!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
albertzaharovits: Good catch!

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/ProfileHasPrivilegesRequest.java
```diff
@@ -0,0 +1,155 @@
...
+            ObjectParser.ValueType.OBJECT
+        );
+    }
+
+    private List<String> uids;
+    private String[] clusterPrivileges;
+    private IndicesPrivileges[] indexPrivileges;
+    private RoleDescriptor.ApplicationResourcePrivileges[] applicationPrivileges;
+
+    public ProfileHasPrivilegesRequest(List<String> uids, RoleDescriptor privilegesToCheck) {
```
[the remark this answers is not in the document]
albertzaharovits: I agree that the parsed `RoleDescriptor`, which is incomplete, needs to be validated before it is split into the index, cluster and application privileges.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java
albertzaharovits: I have pushed d9b60fbf63e865a09a3d87fcdbaad1c72b which reverts to the before behavior that uses the authz info from the context rather than "resolving" it again from the subject.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/ProfileHasPrivilegesRequest.java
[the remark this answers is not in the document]
albertzaharovits: >Taking a step back, it is a concidence and conveinience that the privileges to check can re-use part of the RoleDescriptor code. But they are in fact two different things.

I think I haven't quite understood when I initially read your comment. 
I got that the constructor parameter should be `PrivilegesToCheck` instead of `RoleDescriptor`, but that the parsing from json should still return an incomplete `RoleDescriptor`.

I now get that we should avoid the `RoleDescriptor` class completely and instead use `PrivilegesToCheck`. I have pushed, which does that https://github.com/elastic/elasticsearch/pull/85898/commits/02c0b680ae07aba426cdfd6d53505305cb16aad4 .
Please let me know if this is what you were asking for, @ywangd .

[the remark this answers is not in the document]
albertzaharovits: Ooops, also https://github.com/elastic/elasticsearch/pull/85898/commits/347de51ac53fedd43c6f0ee1211628ed31bd69b4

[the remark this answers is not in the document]
ywangd: Thanks @albertzaharovits 
Yes that's what I meant. Thanks for digging it.

albertzaharovits:
This is ready for review @ywangd .

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/action/profile/ProfileHasPrivilegesRequestTests.java
```diff
@@ -0,0 +1,212 @@
...
+                randomBoolean() ? null : new RoleDescriptor.IndicesPrivileges[0],
+                randomApplicationResourcePrivileges()
+            ),
+            new PrivilegesToCheck(
+                randomClusterPrivileges(),
+                randomIndicesPrivileges(),
+                randomBoolean() ? null : new RoleDescriptor.ApplicationResourcePrivileges[0]
+            )
+        );
+    }
```
ywangd, line 168: If the randomisation picks the empty array, why is the overall `PrivilegesToCheck` invalid? That is, isn't 
`new PrivilegesToCheck(new String[0], randomIndicesPrivileges(), randomApplicationResourcePrivileges())` valid?

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/profile/ProfileIntegTests.java
```diff
@@ -530,6 +556,21 @@ public void testSetEnabled() {
...
+                List.of(profile1.uid()),
+                new PrivilegesToCheck(
+                    new String[] { "cluster:monitor/state" },
+                    new RoleDescriptor.IndicesPrivileges[0],
+                    new RoleDescriptor.ApplicationResourcePrivileges[0]
+                )
+            )
+        ).actionGet();
+        assertThat(profileHasPrivilegesResponse.hasPrivilegeUids(), contains(equalTo(profile1.uid())));
+        assertThat(profileHasPrivilegesResponse.errorUids(), emptyIterable());
```
ywangd, line 572: Honestly I haven't thought about this. The envisioned use case is that the UIDs from a pre-flight search request which means they are all **enabled** (because search filtered out disabled profiles). So I had always assumed the hasPrivileges would only work for enabled profiles. But even when all UIDs are from a search result, it is theoretically possible that they got disabled just before the hasPrivileges check. Should we care? 

Functional wise, whether profile is enabled or not feels orthogonal to hasPrivileges. So I am tempted to say hasPrivileges should *not* care about whether profiles are enabled. But in previous discussion to the Kibana team, we emphasized that the search result is a best effort in terms of privileges (spaces etc) and Kibana must filter them with the hasPrivileges API to be sure. That makes me think it is rather let-down if hasPrivileges API does not care about "enabled" status which can be argued to be the top level privilege. This means we cannot fully trust what hasPrivileges says because the Profile can be disabled (and thus the associated user should not recieved a notification) despite passing the privilege check.

In summary, I think we should **not** allow hasPrivileges to work for disabled profiles.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/profile/TransportProfileHasPrivilegesAction.java
```diff
@@ -0,0 +1,125 @@
...
+        this.profileService = profileService;
+        this.securityContext = securityContext;
+        this.threadPool = threadPool;
+    }
+
+    @Override
+    protected void doExecute(Task task, ProfileHasPrivilegesRequest request, ActionListener<ProfileHasPrivilegesResponse> listener) {
+        profileService.getProfileSubjects(request.profileUids(), ActionListener.wrap(profileSubjectsAndFailures -> {
+            if (profileSubjectsAndFailures.profileUidToSubject().isEmpty()) {
+                listener.onResponse(new ProfileHasPrivilegesResponse(Set.of(), Set.of()));
```
ywangd, line 72:
```suggestion
                listener.onResponse(new ProfileHasPrivilegesResponse(Set.of(), Set.copyOf(profileUidToSubject.failureProfileUids)));
```
?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/profile/TransportProfileHasPrivilegesAction.java
```diff
@@ -0,0 +1,125 @@
...
+                            }, checkPrivilegesException -> {
+                                logger.debug(
+                                    new ParameterizedMessage("Failed to check privileges for profile [{}]", profileUid),
+                                    checkPrivilegesException
+                                );
+                                errorProfiles.add(profileUid);
+                                if (counter.decrementAndGet() == 0) {
+                                    allDone.run();
+                                }
+                            })
```
ywangd, line 108: The `counter` check is a good use case for `ActionListener#runAfter` which likely helps inline `allDone` as well.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -311,6 +331,48 @@ private void getVersionedDocument(String uid, ActionListener<VersionedDocument>
...
 
+    private void getVersionedDocuments(Collection<String> uids, ActionListener<Tuple<List<ProfileDocument>, List<String>>> listener) {
+        if (uids.isEmpty()) {
+            listener.onResponse(new Tuple<>(List.of(), List.of()));
+            return;
+        }
+        tryFreezeAndCheckIndex(listener).ifPresent(frozenProfileIndex -> {
+            frozenProfileIndex.checkIndexVersionThenExecute(
+                listener::onFailure,
+                () -> new OriginSettingClient(client, SECURITY_PROFILE_ORIGIN).prepareMultiGet()
```
ywangd, line 342: Since #86345 is now merged, we'd want 
```suggestion
                () -> new OriginSettingClient(client, getActionOrigin()).prepareMultiGet()
```

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -311,6 +331,48 @@ private void getVersionedDocument(String uid, ActionListener<VersionedDocument>
...
+                    .addIds(frozenProfileIndex.aliasName(), uids.stream().map(ProfileService::uidToDocId).toArray(String[]::new))
+                    .execute(ActionListener.wrap(multiGetResponse -> {
+                        List<ProfileDocument> retrievedDocs = new ArrayList<>(multiGetResponse.getResponses().length);
+                        List<String> failures = new ArrayList<>(0);
+                        Exception loggedException = null;
+                        for (MultiGetItemResponse itemResponse : multiGetResponse.getResponses()) {
+                            if (itemResponse.isFailed() && itemResponse.getFailure() != null) {
```
ywangd, line 349: Internally `isFailed()` is just `return failure != null;` and `getFailure()` is `return this.failure;`. So I think just check `isFailed()` is enough here. Also other branches can be simplified as well because based on `MultiGetItemResponse`'s logic, one of `failure` and `response` must be non-null and we can directly call `MultiGetItemResponse#getId()` which internally delegates to either failure or response.

ywangd (commented):
I had some more comments. The most important one is about whether hasPrivileges should work for disabled profiles. My conclusion is No (see below for details). What do you think?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/RoleDescriptor.java
```diff
@@ -379,12 +381,70 @@ private static String[] readStringArray(String roleName, XContentParser parser,
...
+                );
+            }
+            if (Arrays.stream(indexPrivileges).anyMatch(IndicesPrivileges::isUsingDocumentLevelSecurity)) {
+                throw new ElasticsearchParseException("Field [{}] is not supported in a has_privileges request", Fields.QUERY);
+            }
+        }
+        return new PrivilegesToCheck(
+            clusterPrivileges != null ? clusterPrivileges : Strings.EMPTY_ARRAY,
+            indexPrivileges != null ? indexPrivileges : IndicesPrivileges.NONE,
+            applicationPrivileges != null ? applicationPrivileges : ApplicationResourcePrivileges.NONE
```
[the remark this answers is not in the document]
albertzaharovits: Doing this would change the request validation for the has privilege request (the existing one).

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/ProfileHasPrivilegesResponse.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public int hashCode() {
+        return Objects.hash(hasPrivilegeUids, errorUids);
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+        return builder.startObject()
+            .stringListField("has_privilege_uids", hasPrivilegeUids)
+            .stringListField("error_uids", errorUids)
```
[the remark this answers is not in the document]
albertzaharovits: I have pushed 64776cc8299db75c293960242f . `error_uids` is optionally returned, but `has_privilege_uids` is always returned.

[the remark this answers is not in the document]
ywangd: Yes `has_privilege_uids` should always present.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/RoleDescriptor.java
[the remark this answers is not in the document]
ywangd: My suggestion was based on the assumption `HasPrivilegesRequest` is only created by `HasPrivilegesRequestBuilder`, which uses `RoleDescriptor` with null-handling inside the constructor. But I now noticed it is also created in many other places. That's unfortunate. More reason for promoting the new `PrivilegesToCheck` class to bring consistency. But in a future PR. Thanks!

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/action/profile/ProfileHasPrivilegesRequestTests.java
```diff
@@ -0,0 +1,212 @@
...
+
+    @Override
+    protected Writeable.Reader<ProfileHasPrivilegesRequest> instanceReader() {
+        return ProfileHasPrivilegesRequest::new;
+    }
+
+    @Override
+    protected ProfileHasPrivilegesRequest createTestInstance() {
+        return new ProfileHasPrivilegesRequest(
+            randomList(5, () -> randomAlphaOfLengthBetween(0, 7)),
```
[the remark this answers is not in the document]
albertzaharovits: I have pushed 7b43d56 to prevent `null` fields.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/AuthorizationEngine.java
```diff
@@ -241,6 +255,93 @@ default AuthorizationInfo getAuthenticatedUserAuthorizationInfo() {
...
+        public ActionRequestValidationException validate(ActionRequestValidationException validationException) {
+            if (cluster == null) {
+                validationException = addValidationError("clusterPrivileges must not be null", validationException);
+            }
+            if (index == null) {
+                validationException = addValidationError("indexPrivileges must not be null", validationException);
+            }
+            if (application == null) {
+                validationException = addValidationError("applicationPrivileges must not be null", validationException);
+            } else {
```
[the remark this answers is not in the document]
albertzaharovits: The `null` validation is in place because theoretically (but not in practice) the request for the existing has privilege endpoint can be built without these fields being set.
Do you want me to change the request construction for the existing has privilege endpoint?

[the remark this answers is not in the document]
ywangd: No. Let's keep it this way. As metioned in another comment, I wasn't aware HasPrivilegesRequest is used in many other places. So I don't think this is the PR to change that.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/profile/TransportProfileHasPrivilegesAction.java
albertzaharovits: This is a nice catch, pushed [62cfa87](https://github.com/elastic/elasticsearch/pull/85898/commits/62cfa8712f618e7bde7b3f4b5683d7c6b5f974ba), which includes a test.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/profile/TransportProfileHasPrivilegesAction.java
albertzaharovits: This is a nice suggestion. I have pushed: [577c22c](https://github.com/elastic/elasticsearch/pull/85898/commits/577c22c564053dde08bc6ec6e52deb837f6c5d6e) .

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
albertzaharovits: I was well aware of the internal logic of the mget response that you're describing.
This is why I've put a catch-all else with
```
 logger.error("Inconsistent get item response [{}] [{}]", itemResponse.getIndex(), itemResponse.getId());
```

My approach was to be defensive wrt to inconsistent broken mget results.
But based on your feedback I have pushed 516784e8f2e175de374c6f16d1 . Hopefully you count it as an improvement.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/action/profile/ProfileHasPrivilegesRequestTests.java
albertzaharovits: It was a mistake, I pushed [f59f562](https://github.com/elastic/elasticsearch/pull/85898/commits/f59f562c31dd6f1bafa34089a97d0faf4af05719)

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/AuthorizationEngine.java
```diff
@@ -183,41 +205,33 @@ void validateIndexPermissionsAreSubset(
...
+     *                          from {@link #resolveAuthorizationInfo(Subject, ActionListener)}
+     * @param listener the listener to be notified of the get privileges response
      */
-    void getUserPrivileges(
-        Authentication authentication,
-        AuthorizationInfo authorizationInfo,
-        GetUserPrivilegesRequest request,
-        ActionListener<GetUserPrivilegesResponse> listener
-    );
+    void getUserPrivileges(AuthorizationInfo authorizationInfo, ActionListener<GetUserPrivilegesResponse> listener);
```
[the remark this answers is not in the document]
albertzaharovits: My understanding is that we agreed to return the privileges from the `AuthorizationInfo` from the threadContext, not the one that is resolved by the engine. We also discussed it here: https://github.com/elastic/elasticsearch/pull/85898#discussion_r861799741 .

>So I am OK if you decide to keep the change.

I think this is preferable in order to have a more reasonable interface to the AuthorizatinEngine.

albertzaharovits:
ready for another round @ywangd .

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -311,6 +333,49 @@ private void getVersionedDocument(String uid, ActionListener<VersionedDocument>
...
+                                    );
+                                }
+                            } else if (itemResponse.getResponse() != null) {
+                                if (itemResponse.getResponse().isExists()) {
+                                    retrievedDocs.add(buildProfileDocument(itemResponse.getResponse().getSourceAsBytesRef()));
+                                } else if (logger.isDebugEnabled()) {
+                                    logger.debug("Profile [{}] not found", docIdToUid(itemResponse.getId()));
+                                }
+                            } else {
+                                logger.error("Inconsistent get item response [{}] [{}]", itemResponse.getIndex(), itemResponse.getId());
```
ywangd, line 367: I think we should add an assertion here so it fails loudly in CI if ever happens.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/AuthorizationEngine.java
```diff
@@ -80,16 +87,31 @@
...
+     * This could include retrieval of permissions from an index or external system.
+     * See also {@link #resolveAuthorizationInfo(RequestInfo, ActionListener)}, for which this method is the more general
+     * sibling. This returns the {@code AuthorizationInfo} that is used for access checks outside the context of
+     * authorizing a specific request, i.e. {@link #checkPrivileges(AuthorizationInfo, PrivilegesToCheck, Collection, ActionListener)}
+     *
+     * @param subject object representing the effective user
+     * @param listener the listener to be notified of success using {@link ActionListener#onResponse(Object)}
+     *                 or failure using {@link ActionListener#onFailure(Exception)}
+     */
+    void resolveAuthorizationInfo(Subject subject, ActionListener<AuthorizationInfo> listener);
```
ywangd, line 113: [CustomAuthorizationEngine](https://github.com/elastic/elasticsearch/blob/master/plugins/examples/security-authorization-engine/src/main/java/org/elasticsearch/example/CustomAuthorizationEngine.java) needs to be updated as well for these interface changes. Otherwise CI will get compilation failures once this PR is merged (similar to https://github.com/elastic/elasticsearch/issues/86378). It does not fail here because it [does not run](https://github.com/elastic/elasticsearch/issues/86378#issuecomment-1116630079) on regular PR.

ywangd (approved):
LGTM thanks for the iterations!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
albertzaharovits: Pushed [98f9ff4](https://github.com/elastic/elasticsearch/pull/85898/commits/98f9ff437f5950bc69a636a246e16a091a350e38)

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/AuthorizationEngine.java
albertzaharovits: Nice catch, pushed [11d0c88](https://github.com/elastic/elasticsearch/pull/85898/commits/11d0c889953d17e376ca3a89e2b2109832f39b99) , but I didn't manage to verify that it actually works 🤞

PR #85911, opened 2022-04-14
Remove unnecessary uses of ObjectHashSet

rjernst:
A couple places in ES code use ObjectHashSet when a Java HashSet is
essentially equivalent. This commit converts these trivial uses to use
HashSet.

relates #84735


grcevski (approved):
LGTM!

PR #85913, opened 2022-04-14
Move docker env var settings handling out of bash

rjernst:
In docker ES allows settings to be set via environment variables. This
is currently handled in complex bash logic. This commit moves that logic
into EnvironmentAwareCommand where the rest of the initial settings are
found.

relates #85758


qa/os/src/test/java/org/elasticsearch/packaging/util/docker/Docker.java
```diff
@@ -78,7 +78,7 @@ public class Docker {
      * but that appeared to cause problems with repeatedly destroying and recreating containers with
      * the same name.
      */
-    static String containerId = null;
+    public static String containerId = null;
```
pugnascotia, line 81: I don't think we need this now?

server/src/test/java/org/elasticsearch/common/cli/EnvironmentAwareCommandTests.java
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            @Override
+            protected Build.Type getBuildType() {
+                return buildType;
+            }
+        };
+    }
+
+    public void testNonDockerEnvVarsIgnored() throws Exception {
```
pugnascotia, line 68:
```suggestion
    /**
     * Check that env vars are not respected when the build type is not Docker.
     */
    public void testNonDockerBuildIgnoresEnvVars() throws Exception {
```

server/src/test/java/org/elasticsearch/common/cli/EnvironmentAwareCommandTests.java
```diff
@@ -0,0 +1,111 @@
...
+        mockEnvVars.put("some.setting", "1");
+        callback = env -> {
+            Settings settings = env.settings();
+            assertThat(settings.hasValue("foo.bar"), is(false));
+            assertThat(settings.hasValue("some.settings"), is(false));
+        };
+        execute();
+    }
+
+    public void testDockerEnvSettings() throws Exception {
```
pugnascotia, line 79:
```suggestion
    /**
     * Check that when the build type is Docker, then the different supported styles of passing settings
     * via env vars are handled.
     */
    public void testDockerBuildRespectsEnvVars() throws Exception {
```

server/src/test/java/org/elasticsearch/common/cli/EnvironmentAwareCommandTests.java
```diff
@@ -0,0 +1,111 @@
...
+        mockEnvVars.put("ES_SETTING_SIMPLE_SETTING", "baz");
+        mockEnvVars.put("ES_SETTING_UNDERSCORE__HERE", "baz2");
+        mockEnvVars.put("ES_SETTING_UNDERSCORE__DOT_BAZ", "buzz");
+        mockEnvVars.put("ES_SETTING_DOUBLE____UNDERSCORE", "buzz");
+        mockEnvVars.put("ES_SETTING_TRIPLE___BAZ", "buzz");
+        mockEnvVars.put("lowercase.setting", "baz");
+        mockEnvVars.put("setting.Ignored", "baz");
+        callback = env -> {
+            Settings settings = env.settings();
+            assertThat(settings.hasValue("simple.setting"), is(true));
```
pugnascotia, line 90: What do you think about also checking that the settings have the correct values as well?

server/src/test/java/org/elasticsearch/common/cli/EnvironmentAwareCommandTests.java
```diff
@@ -0,0 +1,111 @@
...
+            assertThat(settings.hasValue("triple_.baz"), is(true));
+            assertThat(settings.hasValue("double__underscore"), is(true));
+            assertThat(settings.hasValue("lowercase.setting"), is(true));
+            assertThat(settings.hasValue("setting.Ignored"), is(false));
+        };
+        execute();
+    }
+
+    public void testDockerEnvSettingsOverride() throws Exception {
+        // docker env takes precedence over settings on the command line
```
pugnascotia, line 102:
```suggestion
    /**
     * Check that with the Docker build, env vars takes precedence over settings on the command line.
     */
    public void testDockerBuildEnvVarsOverrideCommandLine() throws Exception {
```

server/src/test/java/org/elasticsearch/common/cli/EnvironmentAwareCommandTests.java
```diff
@@ -0,0 +1,111 @@
...
+import org.junit.Before;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Consumer;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+public class EnvironmentAwareCommandTests extends CommandTestCase {
```
pugnascotia, line 28: Can we also please replicate `DockerTests.test087EnvironmentVariablesInIncorrectFormatAreIgnored()`? It checks that env vars, which look like maybe they ought to be respected, are ignored due to being malformed.

You'll also need to remove these tests from `DockerTests`, because they rely on being able to inspect the Elasticsearch command line for options:

   * `test086EnvironmentVariablesInSnakeCaseAreTranslated`
   * `test087EnvironmentVariablesInIncorrectFormatAreIgnored`
   * `test087EnvironmentVariablesInIncorrectFormatAreIgnored`

pugnascotia (changes requested):
There are a number of debug statements that need to be removed as well.

rjernst:
Thanks Rory! I think I have a addressed all your comments. I normally don't like javadoc comments on test methods *and* having super descriptive names (which are actually harder to read IMO becauseOfTheLongRunningCamelCaseThatIsHardToVisuallyParse ;) But I added some simple comments to each method to describe what they are checking and standardized on method naming.

rjernst:
Thanks for the docker test fixes Rory. I think this is ready for a final review now that all tests are passing.

pugnascotia (approved):

PR #85914, opened 2022-04-14
Add noop detection to node shutdown actions

rjernst:
When node shutdown apis are called to add or remove a node to be
shutdown, it is possible the given node is already shutting down. In
this case, there is no need to submit a cluster state update. This
commit detects when this no-op case occurs and simply returns.

relates #84847


pgomulka (approved):
LGTM

PR #85919, opened 2022-04-15
Convert license tools to use unified cli entrypoint

rjernst:
The license tools exists separate from the normal cli tools since they
are not part of the Elasticsearch distribution. However, they still
utilize the Elasticsearch cli infrastructure. This commit adds a common
cli base script that calls the cli launcher and converts each of the
license tools to use it.

relates #85758


grcevski (approved):
LGTM!

PR #85926, opened 2022-04-15
Execute self-reference checks once per pipeline

### LINKED ISSUE #85790: Reduce script processor overhead from safety check
Currently 99% of a profiled `script` processor's execution time comes from the invocation of `CollectionUtils.ensureNoSelfReferences()`, not the script itself (this could vary with script intensity, but it stands to reason it would take a lot to change this dramatically).
image

1. Is there another way we could approach this self-reference guard rail? It's not completely fool-proof but after the first document passes this check we can be reasonably sure the Script doesn't cause self-references, except in the case of some very specific conditional behavior one would almost have to intentionally code in. As this measure is to help users prevent shooting themselves in the foot, an ostensible 100x overhead is too high a cost, surely.
2. As a half-measure, should we not execute the check if the processor has `ignore_errors` set to `true` anyway? This is the case even for some published integrations, and could benefit greatly from the reduced overhead. It's not ideal as this particular problem really should probably not be ignored since it will cause issues in a later step, but this would preserve existing behavior and reduce cost.

DJRickyB:
Closes #85790 

## Purpose
Currently, in an ingest pipeline, each time a script is executed, we check the IngestDocument to ensure there are no circular references, thus making it unserializable. This change pulls that check out to the whole pipeline for performance reasons.

## Bonus
1. The `IllegalArgumentException` thrown by `ensureNoSelfReferences` is currently caught and ignorable by the script processor's `ignore_failures` flag. This better prevents unserializable documents from causing StackOverflowErrors at serialization time, as the exception explicitly fails the pipeline if it has not already failed for some other reason.
2. Script-based conditionals also were performing this check, so it is not just for script processors.
3. Conceivably a pipeline could introduce and clean up a circular reference, this allows for that.


DJRickyB:
Once-per-pipeline gives us about 3-4% back in mean indexing throughput in my logging solution benchmark, which is CPU bound, very consistently. example:

```diff
    ____        ____
   / __ \____ _/ / /_  __
  / /_/ / __ `/ / / / / /
 / _, _/ /_/ / / / /_/ /
/_/ |_|\__,_/_/_/\__, /
                /____/


Comparing baseline
  Race ID: f73bea7c-acfb-48bb-87e9-6e3c13c7dc29
  Race timestamp: 2022-04-15 18:03:07
  Challenge: logging-indexing
  Car: external
  User tags: env-id=956ebb87-41bf-4c61-9e9e-efbef38ce93e, setup=esbench

with contender
  Race ID: 8aa3cf7c-344a-4677-b672-56316544ef6a
  Race timestamp: 2022-04-15 19:26:50
  Challenge: logging-indexing
  Car: external
  User tags: env-id=956ebb87-41bf-4c61-9e9e-efbef38ce93e, setup=esbench

------------------------------------------------------
    _______             __   _____
   / ____(_)___  ____ _/ /  / ___/_________  ________
  / /_  / / __ \/ __ `/ /   \__ \/ ___/ __ \/ ___/ _ \
 / __/ / / / / / /_/ / /   ___/ / /__/ /_/ / /  /  __/
/_/   /_/_/ /_/\__,_/_/   /____/\___/\____/_/   \___/
------------------------------------------------------

|                                                        Metric |              Task |        Baseline |       Contender |           Diff |   Unit |   Diff % |
|--------------------------------------------------------------:|------------------:|----------------:|----------------:|---------------:|-------:|---------:|
|                    Cumulative indexing time of primary shards |                   |   136.276       |   137.906       |    1.63008     |    min |   +1.20% |
|             Min cumulative indexing time across primary shard |                   |     0           |     0           |    0           |    min |    0.00% |
|          Median cumulative indexing time across primary shard |                   |     2.96183     |     2.95748     |   -0.00435     |    min |   -0.15% |
|             Max cumulative indexing time across primary shard |                   |    76.8061      |    78.603       |    1.79687     |    min |   +2.34% |
|           Cumulative indexing throttle time of primary shards |                   |     0           |     0           |    0           |    min |    0.00% |
|    Min cumulative indexing throttle time across primary shard |                   |     0           |     0           |    0           |    min |    0.00% |
| Median cumulative indexing throttle time across primary shard |                   |     0           |     0           |    0           |    min |    0.00% |
|    Max cumulative indexing throttle time across primary shard |                   |     0           |     0           |    0           |    min |    0.00% |
|                       Cumulative merge time of primary shards |                   |    43.8519      |    42.7787      |   -1.07317     |    min |   -2.45% |
|                      Cumulative merge count of primary shards |                   |   338           |   342           |    4           |        |   +1.18% |
|                Min cumulative merge time across primary shard |                   |     0           |     0           |    0           |    min |    0.00% |
|             Median cumulative merge time across primary shard |                   |     0.00495     |     0.00418333  |   -0.00077     |    min |  -15.49% |
|                Max cumulative merge time across primary shard |                   |    40.0235      |    38.6845      |   -1.33903     |    min |   -3.35% |
|              Cumulative merge throttle time of primary shards |                   |    11.6338      |    12.3145      |    0.68062     |    min |   +5.85% |
|       Min cumulative merge throttle time across primary shard |                   |     0           |     0           |    0           |    min |    0.00% |
|    Median cumulative merge throttle time across primary shard |                   |     0           |     0           |    0           |    min |    0.00% |
|       Max cumulative merge throttle time across primary shard |                   |    10.6137      |    11.214       |    0.60032     |    min |   +5.66% |
|                     Cumulative refresh time of primary shards |                   |     2.1944      |     2.1603      |   -0.0341      |    min |   -1.55% |
|                    Cumulative refresh count of primary shards |                   |   465           |   459           |   -6           |        |   -1.29% |
|              Min cumulative refresh time across primary shard |                   |     0           |     0           |    0           |    min |    0.00% |
|           Median cumulative refresh time across primary shard |                   |     0.0206333   |     0.0210667   |    0.00043     |    min |   +2.10% |
|              Max cumulative refresh time across primary shard |                   |     1.80635     |     1.7837      |   -0.02265     |    min |   -1.25% |
|                       Cumulative flush time of primary shards |                   |     4.24188     |     4.25818     |    0.0163      |    min |   +0.38% |
|                      Cumulative flush count of primary shards |                   |   152           |   153           |    1           |        |   +0.66% |
|                Min cumulative flush time across primary shard |                   |     0           |     0           |    0           |    min |    0.00% |
|             Median cumulative flush time across primary shard |                   |     0.157567    |     0.16455     |    0.00698     |    min |   +4.43% |
|                Max cumulative flush time across primary shard |                   |     1.06583     |     1.07038     |    0.00455     |    min |   +0.43% |
|                                       Total Young Gen GC time |                   |     3.809       |     3.812       |    0.003       |      s |   +0.08% |
|                                      Total Young Gen GC count |                   |   205           |   204           |   -1           |        |   -0.49% |
|                                         Total Old Gen GC time |                   |     0           |     0           |    0           |      s |    0.00% |
|                                        Total Old Gen GC count |                   |     0           |     0           |    0           |        |    0.00% |
|                                                    Store size |                   |    13.6763      |    13.1561      |   -0.5202      |     GB |   -3.80% |
|                                                 Translog size |                   |     2.29121     |     1.85416     |   -0.43705     |     GB |  -19.08% |
|                                        Heap used for segments |                   |     0           |     0           |    0           |     MB |    0.00% |
|                                      Heap used for doc values |                   |     0           |     0           |    0           |     MB |    0.00% |
|                                           Heap used for terms |                   |     0           |     0           |    0           |     MB |    0.00% |
|                                           Heap used for norms |                   |     0           |     0           |    0           |     MB |    0.00% |
|                                          Heap used for points |                   |     0           |     0           |    0           |     MB |    0.00% |
|                                   Heap used for stored fields |                   |     0           |     0           |    0           |     MB |    0.00% |
|                                                 Segment count |                   |   170           |   167           |   -3           |        |   -1.76% |
|                                   Total Ingest Pipeline count |                   |     3.49102e+07 |     3.49102e+07 |    0           |        |    0.00% |
+|                                    Total Ingest Pipeline time |                   |     6.20946e+06 |     4.17524e+06 |   -2.03423e+06 |     ms |  -32.76% |
|                                  Total Ingest Pipeline failed |                   |     0           |     0           |    0           |        |    0.00% |
|                                                Min Throughput |  insert-pipelines |    13.9361      |    15.4517      |    1.51554     |  ops/s |  +10.87% |
|                                               Mean Throughput |  insert-pipelines |    13.9361      |    15.4517      |    1.51554     |  ops/s |  +10.87% |
|                                             Median Throughput |  insert-pipelines |    13.9361      |    15.4517      |    1.51554     |  ops/s |  +10.87% |
|                                                Max Throughput |  insert-pipelines |    13.9361      |    15.4517      |    1.51554     |  ops/s |  +10.87% |
|                                      100th percentile latency |  insert-pipelines |  1074.6         |   968.266       | -106.336       |     ms |   -9.90% |
|                                 100th percentile service time |  insert-pipelines |  1074.6         |   968.266       | -106.336       |     ms |   -9.90% |
|                                                    error rate |  insert-pipelines |     0           |     0           |    0           |      % |    0.00% |
|                                                Min Throughput |        insert-ilm |    31.9556      |    27.7967      |   -4.1589      |  ops/s |  -13.01% |
|                                               Mean Throughput |        insert-ilm |    31.9556      |    27.7967      |   -4.1589      |  ops/s |  -13.01% |
|                                             Median Throughput |        insert-ilm |    31.9556      |    27.7967      |   -4.1589      |  ops/s |  -13.01% |
|                                                Max Throughput |        insert-ilm |    31.9556      |    27.7967      |   -4.1589      |  ops/s |  -13.01% |
|                                      100th percentile latency |        insert-ilm |    29.5338      |    33.8879      |    4.35416     |     ms |  +14.74% |
|                                 100th percentile service time |        insert-ilm |    29.5338      |    33.8879      |    4.35416     |     ms |  +14.74% |
|                                                    error rate |        insert-ilm |     0           |     0           |    0           |      % |    0.00% |
|                                                Min Throughput |        bulk-index |   470.233       |   551.757       |   81.5237      | docs/s |  +17.34% |
+|                                               Mean Throughput |        bulk-index | 44389.7         | 46099.8         | 1710.05        | docs/s |   +3.85% |
+|                                             Median Throughput |        bulk-index | 45798.9         | 47388.6         | 1589.72        | docs/s |   +3.47% |
|                                                Max Throughput |        bulk-index | 46214.5         | 47914.3         | 1699.82        | docs/s |   +3.68% |
|                                       50th percentile latency |        bulk-index |   514.618       |   507.593       |   -7.02567     |     ms |   -1.37% |
|                                       90th percentile latency |        bulk-index |  1064.29        |  1011.06        |  -53.2232      |     ms |   -5.00% |
|                                       99th percentile latency |        bulk-index |  1488.1         |  1401.44        |  -86.6662      |     ms |   -5.82% |
|                                     99.9th percentile latency |        bulk-index |  2425.91        |  2198.29        | -227.622       |     ms |   -9.38% |
|                                    99.99th percentile latency |        bulk-index |  4464.13        |  4411.89        |  -52.2409      |     ms |   -1.17% |
|                                      100th percentile latency |        bulk-index |  5070.69        |  4585.85        | -484.838       |     ms |   -9.56% |
|                                  50th percentile service time |        bulk-index |   514.618       |   507.693       |   -6.92515     |     ms |   -1.35% |
|                                  90th percentile service time |        bulk-index |  1062.63        |  1010.68        |  -51.9506      |     ms |   -4.89% |
|                                  99th percentile service time |        bulk-index |  1488.4         |  1401.16        |  -87.2429      |     ms |   -5.86% |
|                                99.9th percentile service time |        bulk-index |  2425.91        |  2198.29        | -227.622       |     ms |   -9.38% |
|                               99.99th percentile service time |        bulk-index |  4464.13        |  4411.89        |  -52.2409      |     ms |   -1.17% |
|                                 100th percentile service time |        bulk-index |  5070.69        |  4585.85        | -484.838       |     ms |   -9.56% |
|                                                    error rate |        bulk-index |     0           |     0           |    0           |      % |    0.00% |
|                                       50th percentile latency | compression-stats | 11999           | 11986.2         |  -12.8506      |     ms |   -0.11% |
|                                       90th percentile latency | compression-stats | 30708.7         | 30032.5         | -676.116       |     ms |   -2.20% |
|                                      100th percentile latency | compression-stats | 31785           | 31334.1         | -450.936       |     ms |   -1.42% |
|                                  50th percentile service time | compression-stats | 11999           | 11986.2         |  -12.8506      |     ms |   -0.11% |
|                                  90th percentile service time | compression-stats | 30708.7         | 30032.5         | -676.116       |     ms |   -2.20% |
|                                 100th percentile service time | compression-stats | 31785           | 31334.1         | -450.936       |     ms |   -1.42% |
|                                                    error rate | compression-stats |    90.9091      |    81.8182      |   -9.09091     |      % |  -10.00% |


-------------------------------
[INFO] SUCCESS (took 0 seconds)
-------------------------------
```

From a contextual standpoint, once-per-doc may not be fine-grained enough, but since we're here I'll test that also.

DJRickyB:
Once-per-doc is not possible without re-factoring the ingest pipelines. Currently they flip the representation of the document between the `IngestDocument` and the `IndexRequest` between each pipeline invocation, the latter requiring a serializable structure. It might be worth investigating if this back-and-forth (the invocations of `sourceAsMap()` in IngestService.innerExecute()`) can also be lifted out of the ingest service execution itself, but that is out of scope for this investigation.

joegallo:
@elasticmachine update branch

joegallo:
The CI failure here looked like https://github.com/elastic/elasticsearch/issues/85949, which is closed, so I updated the branch to fix it (🤞).

server/src/main/java/org/elasticsearch/ingest/IngestService.java
```diff
@@ -851,6 +852,12 @@ private void innerExecute(
                 itemDroppedHandler.accept(slot);
                 handler.accept(null);
             } else {
+                try {
+                    CollectionUtils.ensureNoSelfReferences(result.getSourceAndMetadata(), "ingest pipeline [" + pipeline.getId() + "]");
```
martijnvg, line 856: Is this checked for each pipeline? I think we should only execute this for pipelines that have scripts in it?

martijnvg: As as far as I know, there is no easy of checking whether a pipeline uses scripts.
So I think adding this logic is currently not worth it. If at some point in time, 
there are use cases where with pipelines with no scripts that have a performance a reak regression,
then we can think about adding this logic, so that we self reference check can be skipped.

martijnvg: So it turns out scripts in conditionals can't change the document (see `ConditionalProcessor#evaluate(...)`).
We can use `ScriptService#getProcessorsInPipeline(...)` to find script processors, but the issue here is that `ScriptProcessor` is defined in a ingest-common module. But maybe we can get processors by type?

martijnvg: To clarify, scripts in conditions don't need to be checked for self referencing. Only script processors should be checked. I think we can add a method to `ScriptService` that fetches pipeline with processors of a specific type, just like `getProcessorsInPipeline(...)` does for fetching pipelines with processor of a certain `Class`.

martijnvg: We talked about this and this optimization can be added in a followup pr.

martijnvg (approved):
Thanks @DJRickyB! LGTM 👍

DJRickyB:
@martijnvg I added a32d60f8f1fd95883286a70001989d01a4d29260, and all tests have passed. still good to go?

DJRickyB:
Closes https://github.com/elastic/elasticsearch/issues/85790
Purpose

Currently, in an ingest pipeline, each time a script is executed, we check the IngestDocument to ensure there are no circular references, thus making it unserializable. This change pulls that check out to the whole pipeline for performance reasons.
Bonus:

* The IllegalArgumentException thrown by ensureNoSelfReferences is currently caught and ignorable by the script processor's ignore_failures flag. This better prevents unserializable documents from causing StackOverflowErrors at serialization time, as the exception explicitly fails the pipeline if it has not already failed for some other reason.
*   Script-based conditionals also were performing this check, so it is not just for script processors.
*  Conceivably a pipeline could introduce and clean up a circular reference, this allows for that.

jpountz:
I'm adding the `release highlight` label given the indexing speedup we observed on the Logging track.

server/src/main/java/org/elasticsearch/ingest/IngestService.java
dakrone: Open #87335 for this followup, so we don't forget.

PR #85929, opened 2022-04-15
Only use InitialMasterNodes during bootstrap

stefnestor:
👋🏼 [This doc](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/modules-discovery-bootstrap-cluster.html) currently says ... (referring to `cluster.initial_master_nodes`).

>  After the cluster has formed, this setting is no longer required.

This should instead push users that this setting ONLY makes sense during bootstrap and should be removed after.

### DIFF
--- a/docs/reference/modules/discovery/bootstrapping.asciidoc
+++ b/docs/reference/modules/discovery/bootstrapping.asciidoc
@@ -31,7 +31,7 @@ node:
 
 When you start a master-eligible node, you can provide this setting on the
 command line or in the `elasticsearch.yml` file. After the cluster has formed,
-this setting is no longer required. It should not be set for master-ineligible
+this setting should no longer be used. It should not be set for master-ineligible
 nodes, master-eligible nodes joining an existing cluster, or cluster restarts.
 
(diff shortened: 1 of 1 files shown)


docs/reference/modules/discovery/bootstrapping.asciidoc
```diff
@@ -31,7 +31,7 @@ node:
 
 When you start a master-eligible node, you can provide this setting on the
 command line or in the `elasticsearch.yml` file. After the cluster has formed,
-this setting is no longer required. It should not be set for master-ineligible
+this setting should no longer be used. It should not be set for master-ineligible
```
DaveCTurner, line 34: I think that's still weaker than it could be. How about making it a command to the user, like this?

```suggestion
remove this setting from the configuration. It should not be set for master-ineligible
```

DaveCTurner (changes requested):
I left a suggestion.

stefnestor:
Closing in favor of https://github.com/elastic/elasticsearch/pull/85948.

PR #76483, opened 2021-08-13
Lazily initialize template model - take II

felixbarny:
Fixes #76346

Similar to https://github.com/elastic/elasticsearch/pull/76356 but takes a different approach without API changes. This makes the compatibility tests pass and requires fewer code changes overall.

Before:
flamegraph

After:
flamegraph_lazy_model

PR #85930, opened 2022-04-15
Optimize FieldPath::<init> by avoiding split via regex

felixbarny:
A lot of methods in `IngestDocument` (such as `IngestDocument::hasField` or `IngestDocument::setFieldValue`) take a `String path` argument that represents a dot-separated path within the document.

Currently, the string is split not very efficiently - via a non-compiled regex (`path.split("\\.")`). Looking at the profiles I attached to https://github.com/elastic/elasticsearch/pull/76483, it becomes apparent that the overhead of splitting the string in this way is non-negligible.

Due to the pervasive use of `FieldPath` in almost every processor I think this can have a significant overall impact. It would be interesting to see the difference in observability rally tracks.

### DIFF
--- a/server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
+++ b/server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
@@ -898,6 +898,8 @@ public final class IngestDocument {
 
     private class FieldPath {
 
+        private static final String[] EMPTY_STRING_ARRAY = new String[0];
+
         private final String[] pathElements;
         private final Object initialContext;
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
```diff
@@ -898,6 +898,8 @@ public String getFieldName() {
 
     private class FieldPath {
 
+        private static final String[] EMPTY_STRING_ARRAY = new String[0];
```
dakrone, line 901: We already have `Strings.EMPTY_ARRAY` for this purpose

dakrone (commented):
I left a comment, but this seems like something we could pretty easily test with JMH. Could you put something together for this? It'll be easier to see than in something as noisy as the full nightly ingest benchmarks

felixbarny:
Thanks for the suggestion to create a JMH benchmark. The result was quite surprising: `path.split("\\.")` was actually faster than my custom implementation, although it was causing a few more allocations. It's also much faster compared to using a compiled regex. That's because there's a fast path in `String::split` if there's just one char or one escaped char.

I guess I jumped to conclusions too quickly when I saw a `Pattern.compile` frame in the "after" profile from https://github.com/elastic/elasticsearch/pull/76483. I'm really not sure why there's a `Pattern.compile` frame. Maybe I was testing with an odd JDK back then which didn't have the fast path.

PR #76511, opened 2021-08-13
Add reroute processor

felixbarny:
[Docs preview](https://elasticsearch_76511.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/reroute-processor.html)

The reroute processor allows to route a document to another target index or data stream. It has two main modes:

When setting the destination option, the target is explicitly specified and the dataset and namespace options can’t be set.

When the destination option is not set, this processor is in a data stream mode. Note that in this mode, the reroute processor can only be used on data streams that follow the [data stream naming scheme](https://elasticsearch_76511.docs-preview.app.elstc.co/guide/en/fleet/master/data-streams.html#data-streams-naming-scheme). Trying to use this processor on a data stream with a non-compliant name will raise an exception.

The name of a data stream consists of three parts: <type>-<dataset>-<namespace>. See the [data stream naming scheme](https://elasticsearch_76511.docs-preview.app.elstc.co/guide/en/fleet/master/data-streams.html#data-streams-naming-scheme) documentation for more details.

This processor can use both static values or reference fields from the document to determine the dataset and namespace components of the new target. See [Table 38, “Reroute options”](https://elasticsearch_76511.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/reroute-processor.html#reroute-options) for more details.

It’s not possible to change the type of the data stream with the reroute processor.
After a reroute processor has been executed, all the other processors of the current pipeline are skipped, including the final pipeline. If the current pipeline is executed in the context of a [Pipeline](https://elasticsearch_76511.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/pipeline-processor.html), the calling pipeline will be skipped, too. This means that at most one reroute processor is ever executed within a pipeline, allowing to define mutually exclusive routing conditions, similar to a if, else-if, else-if, … condition.

The reroute processor ensures that the data_stream.<type|dataset|namespace> fields are set according to the new target. If the document contains a event.dataset value, it will be updated to reflect the same value as data_stream.dataset.

Note that the client needs to have permissions to the final target. Otherwise, the document will be rejected with a security exception which looks like this:

```
{"type":"security_exception","reason":"action [indices:admin/auto_create] is unauthorized for API key id [8-dt9H8BqGblnY2uSI--] of user [elastic/fleet-server] on indices [logs-foo-default], this action is granted by the index privileges [auto_configure,create_index,manage,all]"}
```

Reroute options

| Name        | Required  | Default  | Description |
|-|-|-|-|
[...]

Issue #57968, opened 2020-06-11
Applying destination index pipelines after a processor changes the target index 


tvernum:
When evaluating ingest pipelines, we determine the full set of pipelines upfront (taking into account default_pipeline and final_pipeline) based on the index name that is included in the incoming request.

While executing those pipelines it is possible that a processor will change the destination index , for example (trivial example)
```
PUT /_ingest/pipeline/redirect
{
  "description": "Redirect to another index",
  "processors": [{ "set": { "field": "_index", "value": "new-destination" } }]
}
```

If that happens it is possible that the new destination index has its own set of default/final pipelines, but we do not execute them.

It would be preferable if the pipelines attached to the new destination were executed, although there are some questions/issues to work through 
- if the default_pipeline redirects to a new index, should we still execute the final_pipeline of the original index?
- should we execute the default_pipeline of the new destination, or just the final_pipeline?
- how do we prevent the document from being infinitely redirected between indices (could we set a rule that the pipeline of the destination index may not change the destination a second time?)

tvernum:
CC: @ruflin

ruflin:
@exekias @andresrc This is relevant for our discussion to potentially use a "routing" ingest pipeline.

martijnvg:
I think we should aim to keep logic of executing pipelines after the _index has been changed simple.

Prior to performing ingest, a list of pipelines is determined. These pipelines are executed. If
at the end the `_index` has been changed then only the final pipeline of this index (if exists) should
be executed.

> if the default_pipeline redirects to a new index, should we still execute the final_pipeline of the original index?

Given with the above logic, yes.

> should we execute the default_pipeline of the new destination, or just the final_pipeline?

I think only the final pipeline should be executed.

> how do we prevent the document from being infinitely redirected between indices (could we set a rule that the pipeline of the destination index may not change the destination a second time?)

I think it is fine if request or default pipelines change the index. However I think that final pipelines shouldn't be changing the index.

martijnvg:
We discussed this issue with a broader audience and think that the re-execution logic of pipeline should be limited:

* When a final pipeline is executed, it always should be the last pipeline being executed. Also the fact that a final pipeline can change the `_index` key during ingest is a bug, which needs to be addressed.
* If the request pipeline changes the `_index` then this changes the default pipeline and final pipeline to be executed based on whether the new index has a default or final pipeline. If the original index was configured with a default or final pipeline, then these pipelines aren't executed.
[...]

Issue #63798, opened 2020-10-16
[Feature] Document based ingest routing


ruflin:
## Problem

On the Ingest side many different dataset are available. Today our story of building integrations assumes that a single source only produces one data streams. One log file contains the access logs for nginx and other one contains the error logs for nginx and they are not mixed. This allows us to send each to different data streams with a specific ingest pipeline which knows how to process the events.

Unfortunately this is the ideal case and there are multiple scenarios where a single source can contain a mix of datasets. This happens in the logs and the metrics use case. A few examples:

* Docker logs: Docker only allows to log to stderr or stdout. If there are more than 2 log files, the data will be mixed together. 
* Prometheus metrics: A single prometheus collector might collect metrics from many prometheus endpoints
* Kinesis: Getting data out of kinesis mixes many different datasets
* Syslog: Many services send their data a single syslog input

The routing of an event to the right destination could be done on the edge or centrally. This feature request is around having centrally managed routing as it would work well in the context of the integrations installed by Fleet. Having it centrally has the advantage that when new integrations are installed or updated, no updates of all the Agents are needed to add new routing information and it also works well in the case of standalone agents.

## Story

In the following I’m going through a story on how I could imagine document based ingest routing to work. This is NOT a proposal for the API or how it should be designed, but it should help to better explain on how it would be used from Fleet.

The following story uses syslog as a data sink input and all data is sent to syslog.

### Add routing rules

A user gets started with Elastic Agent by enrolling into Fleet. The user starts collecting logs from syslog and all the logs go into `logs-generic-default`. The services that send logs to syslog are nginx and mysql. Based on some ML detection rules, Fleet recognises that the logs being shipped to `logs-generic-default` contain logs for nginx and mysql. It requests the user to install these two packages which the user happily does.

On the Fleet side, this installs all the assets like dashboards, ingest pipelines and templates for nginx and mysql. But on the config side nothing changes as it still just pulls in data from syslog. Each package contains some ingest routing rules. These routing rules are added to `logs-generic-*` to be applied to all data which lands there. Each package can contain a list of conditions. For nginx the following happens:

```
PUT _routing/logs-generic-*
{
  “id”: “nginx”
  “condition”: “ctx.input.type”==”syslog” && contains(“file.path”,”nginx”)
  // Data should be routed to the nginx data stream within the same namespace
  “target”: “logs-nginx-${data_stream.namespace}”
}
```

[...]

PR #85931, opened 2022-04-15
Invoke default pipeline of new index

felixbarny:
When a processor changes the `_index` field, only the final pipeline of the new index will be executed.
This change preserves that behavior but introduces another `_redirect` metadata field that behaves similarly to setting `_index`.
The difference is when setting this field, the default pipeline of the target index will be executed. This PR also detects and prevents loops in the index redirection.

Example: both index_1 and index_2 have a default and final pipeline. the default pipeline of index_1 sets the _redirect field to `index_2`. This executes the pipelines index_1.default, index_2.default, index_2.final.

Required for
* https://github.com/elastic/elasticsearch/pull/76511

Previous discussions:
* https://github.com/elastic/elasticsearch/issues/57968

### DIFF
--- a/server/src/main/java/org/elasticsearch/ingest/IngestService.java
+++ b/server/src/main/java/org/elasticsearch/ingest/IngestService.java
@@ -769,8 +769,6 @@ public class IngestService implements ClusterStateApplier, ReportingService<Inge
     }
 
     private Pipelines getPipelines(IndexRequest indexRequest) {
-        indexRequest.isPipelineResolved(false);
-        resolvePipelines(null, indexRequest, state.metadata());
         final String pipelineId = indexRequest.getPipeline();
         indexRequest.setPipeline(NOOP_PIPELINE_NAME);
         final String finalPipelineId = indexRequest.getFinalPipeline();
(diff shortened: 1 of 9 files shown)


rahuldimri:
@felixbarny is there anything where I can help ?

felixbarny:
@rahuldimri thanks for asking. We're still in internal discussion on how to solve https://github.com/elastic/elasticsearch/issues/63798. This PR is part of a prototype which may not make it in as-is.

Can I ask you why you're interested in this PR? Do you have a use case that could be solved by this? If so, what's your use case?

server/src/main/java/org/elasticsearch/script/Metadata.java
```diff
@@ -39,6 +39,7 @@
  */
 public class Metadata {
     protected static final String INDEX = "_index";
+    protected static final String REDIRECT = "_redirect";
```
felixbarny, line 42: I'm unsure if `_redirect` should also be added to `IndexRequest`. I guess that could potentially expose `_redirect` to the `_bulk` API where this doesn't make much sense. It's also more like an alias to `_index` but with slightly different semantics.

@dakrone do you have a suggestion?

server/src/main/java/org/elasticsearch/ingest/IngestService.java
```diff
@@ -763,21 +751,78 @@ public void onFailure(Exception e) {
                         });
 
                         IngestDocument ingestDocument = newIngestDocument(indexRequest);
-                        executePipelines(pipelines.iterator(), hasFinalPipeline, indexRequest, ingestDocument, documentListener);
-
+                        LinkedHashSet<String> indexRecursionDetection = new LinkedHashSet<>();
+                        indexRecursionDetection.add(indexRequest.index());
```
ruflin, line 755: More general question: I assume during the routing, `index()` will always be the data_stream name for data_streams and the data_stream is only resolved to a write index after the processing happened? Asking it here because the data stream name is the one that should be part of the index recursion detection, not the write index as this could even change during ingest time I assume.

felixbarny: Yes, _index isn't the internal write index of the data stream.

PR #85948, opened 2022-04-16
More docs re. removing cluster.initial_master_nodes

DaveCTurner:
Ensures that on every page of the docs that mentions
`cluster.initial_master_nodes` also mentions that this setting must be
removed after bootstrapping completes.


stefnestor (approved):
TY!

Issue #85695, opened 2022-04-05
Multiple Data Paths deprecation is marked as Critical rather than warning


rdnm:
### Elasticsearch Version

7.17

### Java Version

_bundled_

### OS Version

Any

### Problem Description

In the upgrade assistant in 7.17 a critical warning is shown saying "Multiple data paths are not supported". This should be downgraded to a warning, as Multiple Data Paths were not removed in 8.0, and thus this should not stop upgrade to 8.0. 

See https://github.com/elastic/elasticsearch/issues/71205#issuecomment-1088706578 for more details.

### Steps to Reproduce

Create an elasticsearch deployment with multiple data paths configured. Run the upgrade assistant in 7.17. You will see there is a Critical deprecation message for the multiple data paths feature.

rjernst:
I haven't been able to find this particular deprecation message in Elasticsearch, is it possible it is coming from Kibana directly?

I did find a couple critical deprecations related to this, which I've adjusted in https://github.com/elastic/elasticsearch/pull/85952.

Note that the deprecation log had MDP as critical earlier in 7.x but that was fixed in https://github.com/elastic/elasticsearch/pull/79463 for 7.16.

rjernst:
This was fixed in https://github.com/elastic/elasticsearch/pull/86102

PR #85952, opened 2022-04-18
Relax data path deprecations from critical to warn

rjernst:
The deprecations for index and shared data paths were defined at
critical level, but these are not removed in 8.0. This commit relaxes
these deprecation messages so they are just warnings.

relates #85695


grcevski (approved):
LGTM!

PR #85953, opened 2022-04-18
Move netutil to netty module

rjernst:
This utils class was previously shared across transport implementations,
but is only used by netty now.


williamrandolph (approved):
LGTM!

PR #85954, opened 2022-04-18
Move io utils to core package

rjernst:
Most classes under elasticsearch-core had been moved to the o.e.core
package. However, a couple io related classes remained in an "internal"
package. This commit moves Streams and IOUtils to the core package, as
they are no more "internal" than the rest of the classes in core.


ChrisHegarty (approved):
LGTM.

PR #85956, opened 2022-04-18
Remove hppc from term vector fields api

rjernst:
The term vectors code is some of the oldest in Elasticsearch. It
currently uses hppc for keeping track of terms to offsets. This commit
converts to using a HashMap for this mapping.

relates #84735


rjernst:
Note this PR is purely mechanical. The only tests I could find are high level, but they should be sufficient for this very basic change.

romseygeek (approved):
LGTM

PR #85957, opened 2022-04-18
Remove hppc from some "common" classes

rjernst:
This commit removes hppc from two trivial cases under o.e.common.

relates #84735


grcevski (approved):
LGTM!

PR #85958, opened 2022-04-18
[ML] adds new question_answering NLP task for extracting answers to questions from a document

benwtrent:
This commit adds a new `question_answering` task.

The `question_answering` task allows supplying a `question` in the inference config update.

When storing the model config for inference:
```
"inference_config": {
  "question_answering": {
    "tokenization": {...}, // tokenization settings, recommend doing 386 max sequence length with 128 span, and no truncating
    "max_answer_length": 15 // the max answer length to consider
  }
}
```

Then when calling `_infer` or running with in a pipeline, add the `question` you want answered given the context provided by the document text
```
{
  "docs":[{ "text_field": <some long text field to extract answer}],
  "inference_config": {
    "question_answering": {
      "question": <Question desiring answer>
    }
  }
}
```
The response then looks like:
```
{
    "predicted_value": <string subsection of the document that is the answer>
    "start_offset": <Char offset in document to start>,
    "end_offset": <char offset end of the answer,
    "prediction_probability": <prediction score>
}
```

Some models tested:

 - https://huggingface.co/distilbert-base-cased-distilled-squad
 - https://huggingface.co/bert-large-uncased-whole-word-masking-finetuned-squad
 - https://huggingface.co/deepset/electra-base-squad2
 - https://huggingface.co/deepset/tinyroberta-squad2


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/NlpTokenizer.java
```diff
@@ -207,14 +208,134 @@ public TokenizationResult.Tokens tokenize(
...
             tokenIdsSeq1.stream().map(DelimitedToken.Encoded::getEncoding).collect(Collectors.toList()),
             tokenPositionMapSeq1,
             tokenIdsSeq2.stream().map(DelimitedToken.Encoded::getEncoding).collect(Collectors.toList()),
             tokenPositionMapSeq2
-        ).build(seq1 + seq2, isTruncated, tokens, -1, sequenceId);
+        ).build(List.of(seq1, seq2), isTruncated, List.of(innerResultSeq1.tokens, innerResultSeq2.tokens), -1, sequenceId);
+    }
+
+    /**
+     * Tokenize the two sequences, allowing for spanning of one of the
```
davidkyle, line 220:
```suggestion
     * Tokenize the two sequences, allowing for spanning of the 2nd sequence
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/NlpTokenizer.java
```diff
@@ -207,14 +208,134 @@ public TokenizationResult.Tokens tokenize(
...
+     * @param seq1 The first sequence in the pair
+     * @param seq2 The second sequence
+     * @param truncate truncate settings
+     * @param span the spanning settings, how many tokens to overlap.
+     *             We split and span on seq2.
+     * @param sequenceId Unique sequence id for this tokenization
+     * @return tokenization result for the sequence pair
+     */
+    public List<TokenizationResult.Tokens> tokenize(String seq1, String seq2, Tokenization.Truncate truncate, int span, int sequenceId) {
```
davidkyle, line 229: This looks very similar to 

`tokenize(
        String seq1,
        InnerTokenization innerResultSeq1,
        String seq2,
        Tokenization.Truncate truncate,
        int sequenceId
    )`

but with the spanning logic added. Is there an opportunity to refactor here?

davidkyle (approved):
Nits, nothing more.

LGTM I added the `cloud-deploy` label so I can test it out

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/trainedmodel/QuestionAnsweringConfigUpdate.java
```diff
@@ -0,0 +1,258 @@
...
+            Optional.ofNullable(maxAnswerLength).orElse(questionAnsweringConfig.getMaxAnswerLength()),
+            questionAnsweringConfig.getVocabularyConfig(),
+            tokenizationUpdate == null
+                ? questionAnsweringConfig.getTokenization()
+                : tokenizationUpdate.apply(questionAnsweringConfig.getTokenization()),
+            Optional.ofNullable(resultsField).orElse(questionAnsweringConfig.getResultsField())
+        );
+    }
+
+    boolean isNoop(QuestionAnsweringConfig originalConfig) {
```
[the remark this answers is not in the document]
benwtrent: Noop is not really called anywhere for Q/A as we require the question to be provided. Having a "default question" stored in the context makes no sense. 

This is indeed unique among our current tasks.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/NlpTokenizer.java
```diff
@@ -98,7 +98,8 @@ public List<TokenizationResult.Tokens> tokenize(String seq, Tokenization.Truncat
             );
             // Make sure we do not end on a word
             if (splitEndPos != tokenIds.size()) {
-                while (Objects.equals(tokenPositionMap.get(splitEndPos), tokenPositionMap.get(splitEndPos - 1))) {
+                while (splitEndPos > splitStartPos + 1
```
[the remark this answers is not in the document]
benwtrent: We need to verify that the end is actually after the start. If we don't we could get into a death spiral where the splitting never moves forward (think adversarial case of very small sequences with overlaps that are almost the same as the sequence size).

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/NlpTokenizer.java
benwtrent: Maybe, but the previous one shouldn't do spanning at all. Mainly because its an optimization to allow a single tokenization to be used multiple times. We just have no way of representing a spanned batched encoding (4d tensor?). 

I think some refactoring could be done as a whole, but it would take more churn on the rest of the tokenization code.

benwtrent:
@elasticmachine update branch

benwtrent:
@elasticmachine update branch

benwtrent:
@elasticmachine update branch

PR #85959, opened 2022-04-18
Fix null_value for array-valued geo_point fields

nik9000:
If you configure a `geo_point` field with a `null_value` and then send
the value like:
```
"point_field": [
  {some valid geo_point},
  null
]
```
It would fail. This fixes that to be consistent with the way other
fields handle `null_value`.


server/src/main/java/org/elasticsearch/index/mapper/AbstractPointGeometryFieldMapper.java
```diff
@@ -120,8 +120,12 @@ public void parse(XContentParser parser, CheckedConsumer<T, IOException> consume
                     consumer.accept(validate(point));
                 } else {
                     while (token != XContentParser.Token.END_ARRAY) {
-                        parseAndConsumeFromObject(parser, point, consumer, onMalformed);
-                        point = pointSupplier.get();
+                        if (parser.currentToken() == XContentParser.Token.VALUE_NULL) {
+                            consumer.accept(nullValue);
```
imotov, line 124: I feel like this should be wrapped into `if (nullValue != null) { ... } ` otherwise it causes NPE inside consumer that it interprets as malformed data.

imotov (approved):
Added a small suggestion, otherwise LGTM

server/src/main/java/org/elasticsearch/index/mapper/AbstractPointGeometryFieldMapper.java
nik9000: :+1: I see how the other code had it. I added a test for this.

PR #85960, opened 2022-04-18
Move docker cgroup override to SystemJvmOptions

rjernst:
The system jvm options are those that Elasticsearch always adds to the
ES java command line. In docker, a hack exists for cgroups, which is
passed through the user's ES_JAVA_OPTS. This commit moves the processing
of that docker hack to the system jvm options.

relates #85758


mark-vieira (approved):
LGTM

PR #85961, opened 2022-04-18
Remove use of Terminal from uncaught exception handler

rjernst:
The uncaught excpetion handler watches for exceptions thrown from
Elasticsearch that are missed by all other exception handling. This
handler determines whether the exception is fatal and should cause
Elasticsearch to exit. Most of the time this occurs long after
Elasticsearch has started up, in which case we may have already closed
the underlying System.err. The exception handler currently writes
to System.err through the default Terminal.

This commit removes writing these exceptions to the Terminal. The
handler is already added after logging is initialized, so if an
exception occurs on startup, this will already be logged to the
console through the console logging appender.

relates #85758


server/src/main/java/org/elasticsearch/bootstrap/ElasticsearchUncaughtExceptionHandler.java
```diff
@@ -54,21 +53,11 @@ static boolean isFatalUncaught(Throwable e) {
     void onFatalUncaught(final String threadName, final Throwable t) {
         final String message = "fatal error in thread [" + threadName + "], exiting";
         logger.error(message, t);
-        Terminal.DEFAULT.errorPrintln(message);
-        t.printStackTrace(Terminal.DEFAULT.getErrorWriter());
-        // Without a final flush, the stacktrace may not be shown before ES exits
-        Terminal.DEFAULT.flush();
-
-        Elasticsearch.printLogsSuggestion();
```
grcevski, line 62: I have a question about removing the printLogsSuggestion(), do we not want to keep the friendly error message for the user to check the logs for more details? We show the last bit of logs on the screen, but there might be more insight of the error in the logs.

rjernst: printLogsSuggestion() writes to the Terminal. So it's effectively like the others, it might be broken because we've closed stderr.

However, your comment made me realize we are implicitly relying on the uncaught exception handler for startup as well, by allowing main have a checked Exception. I will change that to have a broad try/catch so that we can capture and print the exception ourselves instead of relying on the JVM to do so for us.

rjernst: I pushed 2d14865

server/src/main/java/org/elasticsearch/bootstrap/ElasticsearchUncaughtExceptionHandler.java
```diff
@@ -54,21 +53,11 @@ static boolean isFatalUncaught(Throwable e) {
     void onFatalUncaught(final String threadName, final Throwable t) {
         final String message = "fatal error in thread [" + threadName + "], exiting";
         logger.error(message, t);
-        Terminal.DEFAULT.errorPrintln(message);
-        t.printStackTrace(Terminal.DEFAULT.getErrorWriter());
-        // Without a final flush, the stacktrace may not be shown before ES exits
-        Terminal.DEFAULT.flush();
```
williamrandolph, line 60: If we are still using the terminal to print, do we need to do this flush before we exit?

williamrandolph (commented):
Looks good overall, but I'm puzzled by the test failures.

server/src/main/java/org/elasticsearch/bootstrap/ElasticsearchUncaughtExceptionHandler.java
rjernst: I thought no, but there is a small edge case where logging has not been setup in which we might not flush (otherwise we flush indirectly in printLogsSuggestion). I've added an explicit flush right before we exit.

rjernst:

rjernst:
@williamrandolph @grcevski CI is passing now, can you take another look?

grcevski (approved):
Looks good to me! LGTM! I tested with throwing exceptions in various places during bootime and it works well.

PR #85962, opened 2022-04-18
Remove Terminal.readSecret with max length

rjernst:
The Terminal class currently has two variants of readSecret, one that is
similar to Console.readPassword, and another that takes a maximum length
to read. This second variant was originally added as a protection when
reading the keystore password. However, there are no other uses of it,
and the passphrase has already been fully read in bin/elasticsearch by
the time it is read from Terminal, so it shouldn't be possible to
actually hit this edge case. This commit removes the maxLength variant.

relates #85758


rjernst:
Note that I did not yet remove readLineToCharArray, but that should be possible in a followup once https://github.com/elastic/elasticsearch/pull/85758 passes the passphrase through directly to Elasticsearch.

williamrandolph (approved):
Looks good to me! I think this cleans up the code nicely.

PR #85964, opened 2022-04-18
[DOCS] Remove references to bootstrap.system_call_filter

### LINKED ISSUE #84862: The docs still reference bootstrap.system_call_filter
### Description

The [bootstrap checks docs for 8.1](https://github.com/elastic/elasticsearch/blob/8.1/docs/reference/setup/bootstrap-checks.asciidoc#system-call-filter-check) still have a [section for disabling system call filters](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/_system_call_filter_check.html) even though that setting was removed in 8.0 (#72848) and its removal is [listed in the breaking changes](https://www.elastic.co/guide/en/elastic-stack/8.0/elasticsearch-breaking-changes.html)

grcevski:
This PR removes all references to the now removed option of `bootstrap.system_call_filter`.

Closes #84862

### DIFF
--- a/docs/reference/setup/bootstrap-checks.asciidoc
+++ b/docs/reference/setup/bootstrap-checks.asciidoc
@@ -194,10 +194,9 @@ filters are installed to prevent the ability to execute system calls
 related to forking as a defense mechanism against arbitrary code
 execution attacks on Elasticsearch. The system call filter check ensures
 that if system call filters are enabled, then they were successfully
-installed. To pass the system call filter check you must either fix any
+installed. To pass the system call filter check you must fix any
 configuration errors on your system that prevented system call filters
-from installing (check your logs), or *at your own risk* disable system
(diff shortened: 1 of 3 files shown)


rjernst (commented):
I don’t think we can remove this currently in master, since it would be a breaking change? That is, removing the actual setting. We could still make the docs change.

grcevski:
I think the setting was removed in 8.0 (https://github.com/elastic/elasticsearch/pull/72848) and its removal is [listed in the breaking changes](https://www.elastic.co/guide/en/elastic-stack/8.0/elasticsearch-breaking-changes.html). This just removes some outstanding docs + references to now removed setting. The actual setting static I removed was flagged as unused by IntelliJ, nothing depends on it.

rjernst (approved):
Ah, LGTM then!

PR #85967, opened 2022-04-18
Adding a deprecation info API check for too many shards

masseyke:
This commit makes sure that there is enough room in a cluster to add a small number of shards during an upgrade. The
information is exposed in the deprecation info API as a cluster configuration check.
Closes #85702


joegallo:
Do we actually need room for the replicas? That is, does something in the process truly require these indices to be **green** or is (merely) yellow sufficient?

As an edge case, I'm not sure your API is correctly shaped. Imagine a three data node cluster where two of the three nodes are already at `cluster.max_shards_per_node` and the last node is 20 shards under that. I think the math you're doing internally would indicate that "yes, I can totally allocate 5 new shards + 1 replica each", but that isn't actually so. I think `checkShardLimit` is closer to the right API already.

masseyke:
Yeah my initial thought was to just check that we're under the limit. But it sounds like the problem is actually that during an upgrade kibana creates more indices. See https://github.com/elastic/elasticsearch/issues/85702#issuecomment-1098549411 for a little more. So you need a little more room. My thinking (and maybe it's not great) was that on a tiny cluster like you describe, you wouldn't be on cloud and you wouldn't let something like the deprecation API stop you from upgrading. But I don't know how we handle both supporting upgrades on cloud and tiny self-managed clusters, unless we kick this check over to cloud to do.

x-pack/plugin/deprecation/src/main/java/org/elasticsearch/xpack/deprecation/ClusterDeprecationChecks.java
```diff
@@ -0,0 +1,30 @@
...
+    static DeprecationIssue checkShards(ClusterState clusterState) {
+        // Make sure we have room to add a small non-frozen index if needed
+        if (ShardLimitValidator.canAddShardsToCluster(5, 1, clusterState, false)) {
+            return null;
+        } else {
+            return new DeprecationIssue(
+                DeprecationIssue.Level.WARNING,
+                "The cluster has too many shards to be able to upgrade",
+                "https://ela.st/es-deprecation-7-shard-limit",
+                "Delete indices to clear up space",
```
dakrone, line 24: Is this message from another deprecation error? Otherwise maybe we can be more specific and say something like "there is not enough room for N more shards, increase the `cluster.max_shards_per_node` or `cluster.max_shards_per_node.frozen` settings or remove indices to clear up resources."

(using "space" is hard here, because it makes it sound like it's disk-related, when really it's just our arbitrary limits)

dakrone (commented):
> But I don't know how we handle both supporting upgrades on cloud and tiny self-managed clusters, unless we kick this check over to cloud to do.

Since this is a dynamic setting, this is not the end of the world for supporting. For example, a tiny self-managed cluster could probably bump the limit up prior to their upgrade.

x-pack/plugin/deprecation/src/main/java/org/elasticsearch/xpack/deprecation/ClusterDeprecationChecks.java
masseyke: It might be used in another error message, but in this case I think it was just me trying to walk the line between being too technical and not technical enough to be useful. I'll reword it.

masseyke: I intentionally didn't include `cluster.max_shards_per_node.frozen` in the reworded version because we're not checking that setting (the assumption is that the new small index would not be a frozen index).

masseyke:
@elasticmachine update branch

dakrone (approved):
LGTM, sorry I took so long for review on this.

PR #85977, opened 2022-04-19
Primary system indices should be write indices for system aliases

### LINKED ISSUE #85072: Re-indexing security system index results in broken index state
### Elasticsearch Version

local cluster running master commit 22824e47bc894e7d06e27a6fa7b221488516cb35, 8.2.0-SNAPSHOT

### Installed Plugins

_No response_

### Java Version

17

### OS Version

Darwin Kernel Version 21.3.0

### Problem Description

In ES 8.x, re-indexing `.security-tokens` to a new index in the right namespace, e.g., `.security-tokens-8`, results in the creation of a new "duplicate" index and a subsequent error state when attempting to use the `.security-tokens` index, effectively breaking it. The underlying issue is that the `.security-tokens` alias consequently points to two indeces. 

This likely extends to other system indeces as well, not just `.security-tokens`.

Instead of allowing the re-index to proceed, we should prevent the operation, similar to how we prevent creating a new index `.security-tokens-*` within the namespace if an index already exists.

### Steps to Reproduce

Trigger `.security-tokens` index creation:

```
POST /_security/oauth2/token
{
  "grant_type": "client_credentials"
}
```

Re-index:

```
POST _reindex
{
  "source": {
    "index": ".security-tokens"
  },
  "dest": {
    "index": ".security-tokens-42"
  }
}
```

Attempt to use index again:

```
POST /_security/oauth2/token
{
  "grant_type": "client_credentials"
}
```

Results in 400 and error log:

```
java.lang.IllegalStateException: Alias [.security-tokens] points to more than one index: [[.security-tokens-42/g6y3TELtQBCpNPo2ILrSww], [.security-tokens-7/vooe5E9lRvKz3AuqZbza7g]]
```

Running `GET /_cat/indices?expand_wildcards=all` shows that there are now two indeces for the `.security-tokens` alias (`.security-tokens-42` and `.security-tokens-7`)

williamrandolph:
Previously, when auto-creating a system index, we would assign the alias from the system index descriptor to any system index that matched the descriptor's pattern. This meant that, as a side effect, a reindex operation or unexpected autocreate operation can add an index to an alias and cause it to reject writes.

I could think of two fixes. The first was to adjust our code so that a system alias is only automatically added to the system index descriptor's primary index. ~That is what I have done in this pull request.~

The other option is to ensure that the primary index is always designated as the system alias's write index. **EDIT: That is what I have done in this pull request.**

- [x] Will the system index manager propagate this update to existing aliases?

Fixes https://github.com/elastic/elasticsearch/issues/85072


rjernst:
> The other option would be to find a way to ensure that the primary index is always designated as the system alias's write index

For non-externally managed system indices, this seems like the more correct solution? The intent of the system indices infrastructure is to completely manage system indices, so it would seem aligned to have the infrastructure completely control the alias.

williamrandolph:
All of the production code changes here are on code paths that only apply to managed system indices, so on second thought it is probably all right to be restrictive.

I've changed the code so that we always add the alias for any managed system index, and set the primary index as the write index for the alias.

I need to look at the system index manager code to see if we expect these changes to propagate on a version upgrade.

williamrandolph:
The `SystemIndexManager` class can handle mapping updates, but it doesn't do anything to perform fairly trivial fixes to system indices when the system index descriptors change.

We should probably file a different issue to cover this work. It could involve checking that the index is hidden, that its alias is hidden, and that its primary index is designated as the write index for its alias. One reason I'd want to spin this work off is that there could be some complexity around making sure that nodes with different versions aren't competing with each other to change index settings.

server/src/internalClusterTest/java/org/elasticsearch/action/admin/indices/create/AutoCreateSystemIndexIT.java
```diff
@@ -84,6 +99,24 @@ public void testAutoCreateNonPrimaryIndex() throws Exception {
 
         GetIndexResponse response = client().admin().indices().prepareGetIndex().addIndices(INDEX_NAME + "-2").get();
         assertThat(response.indices().length, is(1));
+        assertThat(response.aliases().size(), is(1));
+        assertThat(response.aliases().get(INDEX_NAME + "-2").size(), is(1));
+        assertThat(response.aliases().get(INDEX_NAME + "-2").get(0), equalTo(AliasMetadata.builder(INDEX_NAME).isHidden(true).build()));
+    }
+
+    public void testAutoCreateSystemIndicesAllowsWriteToAlias() throws Exception {
```
grcevski, line 107: Would it be possible to test the opposite scenario? Such that, the first API request makes the non primary index INDEX_NAME-2 and then we add the primary name - 1.

server/src/internalClusterTest/java/org/elasticsearch/action/admin/indices/create/CreateSystemIndicesIT.java
```diff
@@ -98,9 +100,29 @@ public void testSystemIndexIsAutoCreatedViaConcreteName() {
...
+        internalCluster().startNodes(1);
+
+        // Trigger the creation of the system index
+        indexDoc(nonPrimarySystemIndex, "1", "foo", "bar");
+        ensureGreen(nonPrimarySystemIndex);
+
+        assertFalse(indexExists(PRIMARY_INDEX_NAME));
+        assertTrue(indexExists(INDEX_NAME + "-2"));
+
+        // Check that a non-primary system index is not automatically assigned an alias
```
grcevski, line 114: I'm a little bit confused by the comment here, I guess there's an alias, but it's not an alias we can use for writing?

grcevski (approved):
LGTM! This looks good to me, I just left some minor comments and questions?

server/src/internalClusterTest/java/org/elasticsearch/action/admin/indices/create/CreateSystemIndicesIT.java
williamrandolph: Good catch. That comment is out of date. I'll change it!

williamrandolph:

williamrandolph:
@elasticmachine update branch

PR #85985, opened 2022-04-19
Remove index_mode property from index templates

martijnvg:
This PR removes the `index_mode` field from index templates (that was part of `data_stream` json object).
This field was used to determine whether a tsdb data stream should created or a regular data stream.
As part of this change, a tsdb data stream is now created when the template that matches with the data stream to be created has the following properties:

- A `index.routing_path` setting has been defined in the composable index template.
- The fields mentioned in the `index.routing_path` setting refers to fields in the mapping of this template that are of type `keyword` and the `time_series_dimension` attribute is set to `true`.

Prior to this change a template that creates tsdb data stream looks like:

```json
{
    "index_patterns": ["k8s*"],
    "data_stream": {
        "index_mode": "time_series"
    },
   "template": {
      "settings": {
         "index.routing_path": ["metricset.name"],
      }
      "mappings": {
  	  "properties": {
    	    "metrics": {
      	        "properties:": {
                     "name": {
                        "type": "keyword",
      	                "time_series_dimension": true
                     }
                }
    	    }
  	  }
     }
   }
}

```

And with this change a template that creates a tsdb data stream looks like:

```json
{
   "index_patterns": ["k8s*"],
  "data_stream": {},
  "template": {
     "settings": {
       "index.routing_path": ["metricset.name"]
     },
    "mappings": {
  	  "properties": {
    	    "metrics": {
      	        "properties:": {
                     "name": {
                        "type": "keyword",
      	                "time_series_dimension": true
                     }
                }
    	    }
  	  }
     }
  }
}

```

After this PR the following changes will be made:

- Generating  the `index.routing_path` setting if not defined based on the mapping. This will make specifying this setting no longer required.
- Allow tsdb data streams to be created in case when dynamic mappings are used. In this case the concrete routing fields may not be know until ingestions starts. In this case configuring the `index.routing_path` with a wildcard patten is required and in this case configuring these fields in the template mapping isn't required. Validation will happen upon indexing when dynamic fields are mapped.


imotov (commented):
Overall it looks reasonable to me, but I am a bit concern about switching from enum to boolean to indicate time series database. I think we should either rename this boolean to reflect what it actually does instead of which type of data streams it is used in or we should keep it an enum since I anticipate other types of specialized indices will follow if this one is successful.

weizijun:
> After this PR the following changes will be made:
> 
> * Generating  the `index.routing_path` setting if not defined based on the mapping. This will make specifying this setting no longer required.
> * Allow tsdb data streams to be created in case when dynamic mappings are used. In this case the concrete routing fields may not be know until ingestions starts. In this case configuring the `index.routing_path` with a wildcard patten is required and in this case configuring these fields in the template mapping isn't required. Validation will happen upon indexing when dynamic fields are mapped.

About Generating the `index.routing_path`,  I have deep discuss with @nik9000 , there is the detail : https://github.com/elastic/elasticsearch/pull/82633#issuecomment-1017504556
It seems that the Generating is hard to implement.

martijnvg:
> It seems that the Generating is hard to implement.

The plan is to not support the dynamic case, in this case the `routing_path` field must be defined.
The `routing_path` index setting will be only be generated if the template has mappings with fields of type keyword and the 
 `time_series_dimension` attribute is enabled. The reason we like to support this make getting started with tsdb easier. If the  first template someone defines has data stream enabled and time series dimension fields defined, then we create data streams in tsdb mode.

In case no time series dimension fields are defined as concrete field, but a dynamic mapping template is defined then we will require a `routing_path` (possible with field names containing wildcards). But this is for more advanced use cases and likely not for anyone getting started with tsdb.

dakrone (commented):
This generally looks good to me also, but I think I agree with @imotov about keeping it as an enum for future index types.

server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
```diff
@@ -896,6 +896,25 @@ public Map<String, ComposableIndexTemplate> templatesV2() {
...
     }
 
+    public boolean isTsdbTemplate(ComposableIndexTemplate indexTemplate) {
+        var template = indexTemplate.template();
+        if (indexTemplate.getDataStreamTemplate() == null || template == null) {
+            return false;
+        }
+
+        var settings = MetadataIndexTemplateService.resolveSettings(indexTemplate, componentTemplates());
+        if (IndexMetadata.INDEX_ROUTING_PATH.exists(settings)) {
```
nik9000, line 906: If we're going to auto-generate the routing path maybe we should use `index_mode` here?

martijnvg: This method is going to change once `index.routing_path` is going to be generated from the template's mapping. Maybe we can also use `index_mode` index setting here, but that isn't something we said a user would need to configure in order to create data streams in tsdb mode.

nik9000: I *think* configuring the mode is still a good thing. Right now we support configuring fields as `time_series_dimension`s and without modifying the index itself. It'd be pretty surprising if upgrading Elasticsearch caused new indices with those mappings to be time_series. I'm not sure that's a terrible idea, but I think it's worth double checking. Personally I'd only try and generate the `routing_path` if the index is configured in `time_series` mode. But I might be wrong here.

nik9000: Either way, that concern doesn't block this PR - it'd block the PR that generates the `routing_path`.

nik9000:
[weizijun](https://github.com/weizijun) commented [5 days ago](https://github.com/elastic/elasticsearch/pull/85985#issuecomment-1105946321)
> About Generating the `index.routing_path`, I have deep discuss with @nik9000 , there is the detail : [#82633 (comment)](https://github.com/elastic/elasticsearch/pull/82633#issuecomment-1017504556)
> It seems that the Generating is hard to implement.

[martijnvg](https://github.com/martijnvg) commented [yesterday](https://github.com/elastic/elasticsearch/pull/85985#issuecomment-1108562080)
> The plan is to not support the dynamic case, in this case the `routing_path` field must be defined. The `routing_path` index setting will be only be generated if the template has mappings with fields of type keyword and the `time_series_dimension` attribute is enabled. The reason we like to support this make getting started with tsdb easier. If the first template someone defines has data stream enabled and time series dimension fields defined, then we create data streams in tsdb mode.
> 
> In case no time series dimension fields are defined as concrete field, but a dynamic mapping template is defined then we will require a `routing_path` (possible with field names containing wildcards). But this is for more advanced use cases and likely not for anyone getting started with tsdb.

I think this is a pretty decent way to generate it, yeah. It's much better than what I was trying to do - building it dynamically from the keyword fields, updating it as things were added. Down that road lies madness.

I also spent a while thinking about how to generate the field's value when initializing the index itself for the first time. Like, build it from templates and stuff and then derive the routing_path statically from the keyword dimensions. This *should* have worked. But everything happens in the wrong order. So I never pushed on it.

I'm a little worried that this is abstracting away a complex and easy to misconfigure thing inside of templates that may not be owned by folks that know what's up. But it doesn't seem like a bad default. I've mostly listed all of the keyword dimensions in the `routing_path` for my tests. It's *usually* right, I think. And if it weren't for dynamic mappings the whole thing would be easier.....

server/src/main/java/org/elasticsearch/cluster/metadata/DataStream.java
```diff
@@ -81,6 +81,9 @@ public final class DataStream implements SimpleDiffable<DataStream>, ToXContentO
     private final boolean replicated;
     private final boolean system;
     private final boolean allowCustomRouting;
+    // This boolean field is for keeping track of whether a data stream is a tsdb data stream.
```
imotov, line 84: That looks like a leftover from the previous iteration.

server/src/main/java/org/elasticsearch/cluster/metadata/DataStream.java
```diff
@@ -265,24 +268,24 @@ public IndexMode getIndexMode() {
      *
      * @param writeIndex            new write index
      * @param generation            new generation
-     * @param indexModeFromTemplate the index mode as is defined in the template that created this data stream
+     * @param isTsdbTemplate the index mode as is defined in the template that created this data stream
      *
      * @return new {@code DataStream} instance with the rollover operation applied
      */
-    public DataStream rollover(Index writeIndex, long generation, IndexMode indexModeFromTemplate) {
+    public DataStream rollover(Index writeIndex, long generation, boolean isTsdbTemplate) {
```
imotov, line 275: This is internal, so it is probably ok to keep it a boolean here, but I would still prefer not to use the abbreviation Tsdb. I think we are calling it TimeSeries everywhere else in the code.

imotov (commented):
I think I would prefer to keep it enum everywhere, but I am ok with converting it to boolean internally. I am not sure I like the tsdb abbreviation though.

martijnvg:
> I'm a little worried that this is abstracting away a complex and easy to misconfigure thing inside of templates that may not be owned by folks that know what's up. But it doesn't seem like a bad default. I've mostly listed all of the keyword dimensions in the routing_path for my tests. It's usually right, I think. And if it weren't for dynamic mappings the whole thing would be easier.....

Right, dynamic field mapping make things for us more difficult. But I suspect, like you, that in most cases the keyword dimension fields are statically configured in the mapping of the template.

martijnvg:
> I am not sure I like the tsdb abbreviation though.

@imotov I've renamed the variables/parameters to use not use tsdb abbreviation and removed the stale comment.

server/src/main/java/org/elasticsearch/cluster/metadata/ComposableIndexTemplate.java
```diff
@@ -352,10 +336,13 @@ public DataStreamTemplate(boolean hidden, boolean allowCustomRouting, IndexMode
...
-            if (in.getVersion().onOrAfter(Version.V_8_1_0)) {
-                indexMode = in.readOptionalEnum(IndexMode.class);
-            } else {
-                indexMode = null;
+            if (in.getVersion().onOrAfter(Version.V_8_1_0) && in.getVersion().before(Version.V_8_3_0)) {
+                // Accidentally included index_mode to binary node to node protocol in previous releases.
+                // (index_mode is removed and was part of code based when tsdb was behind a feature flag)
+                // (index_mode was behind a feature in the xcontent parser, so it could never actually used)
+                // (this used to be an optional enum, so just need to (de-)serialize a false boolean value here)
+                boolean value = in.readBoolean();
```
imotov, line 344: I think this is the most important place to keep it enum.

imotov (commented):
LGTM, except for the enum serialization part

server/src/main/java/org/elasticsearch/cluster/metadata/ComposableIndexTemplate.java
martijnvg: The index_mode is removed from the the composable index templates, the only reason that a boolean is read here is because index mode wasn't behind a feature flag in the binary serialisation code (it was in the xcontent serialization). 

The index mode is kept in the `DataSteam` class. 

So I think this is good?

dakrone (approved):
LGTM also

PR #85988, opened 2022-04-19
Remove oldState from ILM's onClusterStateProcessed

DaveCTurner:
Relying on the original cluster state during publication completion
handling is incompatible with proper batching. ILM's
`IndexLifecycleClusterStateUpdateTask` passes the original state into
its custom `onClusterStateProcessed` method, but none of the
implementations use it. This commit removes this unused parameter.


dakrone (approved):
LGTM

DaveCTurner:
Thanks Lee

PR #85990, opened 2022-04-19
Allow to sort by script value using SemVer semantics

### LINKED ISSUE #82287: Scripting Fields API X-Pack Version should implement Comparable<Version>
The `Version` value class of the fields API should implement `Comparable<Version>` based on the implementation of [`VersionEncoder.EncodedVersion`](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/mapper-version/src/main/java/org/elasticsearch/xpack/versionfield/VersionEncoder.java#L178).

This will allow users to naturally sort that value class.

### LINKED ISSUE #85989: Sort by VERSION value behaves incorrectly when sort value returned by painless script
When a query `sort` is defined on a script that returns a VERSION value (https://www.elastic.co/guide/en/elasticsearch/reference/current/version.html), the result is sorted using lexicographic order instead of SemVer

## Reproducer

```
PUT /test

{
  "mappings": {
    "properties": {
      "version": {
        "type": "version" 
      }
    }
  }
}

```

```
POST /_bulk

{"index": {"_index":"test"}}
{"name" : "foo", "version":"1.11.3"}
{"index": {"_index":"test"}}
{"name" : "foo", "version":"1.1.4"}
{"index": {"_index":"test"}}
{"name" : "bar", "version":"1.2"}
```

```
POST /test/_search

{
    "size": 1000,
    "_source": false,
    "fields": [
        {
            "field": "name"
        },
        {
            "field": "version"
        }
    ],
    "sort": [
        {
            "_script": {
                "script": {
                    "lang": "painless",
                    "source": "doc['version'].value"
                },
                "type": "string"
            }
        }
    ]
}
```

Also the following does not seem to be an option:
```
"sort": [
        {
            "_script": {
                "script": {
                    "lang": "painless",
                    "source":"$('version', new Version('0'))"
                },
                "type": "string"
            }
        }
    ]
```

since it tries to cast Version to String (and fails of course)

---

Labeling as Core/Infra/Scripting, but not sure if it's for Search instead

luigidellaquila:
### Context: 

In SQL and EQL we are trying to add support for VERSION fields, see https://github.com/elastic/elasticsearch/pull/85502

But `ORDER BY <a calculated value>` does not work properly when the calculated value is of type [Version](https://www.elastic.co/guide/en/elasticsearch/reference/current/version.html), since the calculation is translated to a Painless script.

We tracked down the problem to the following issue: https://github.com/elastic/elasticsearch/issues/85989

We identified the following root problems:
- `sort _script` only accepts `string` and `number` as [type](https://www.elastic.co/guide/en/elasticsearch/painless/current/painless-sort-context.html); both are not a good fit for VERSION fields
- defining a new `ScriptSortType` to handle generic BytesRef values (that we can obtain from VERSION, allowing us to have the right sorting) does not help, since **Version BytesRef is not a valid UTF8 string**, so it cannot be properly formatted in the result
-  ScriptSortBuilder has a hardwired `DocValueFormat.RAW` formatter, that cannot be configured in the query
- `mapper-version` is an x-pack plugin, so we would probably not move the version parsing/formatting logic to Server

Fixes #85989
Fixes #82287


luigidellaquila:
Further reduced the impact of the changes, using `type` directly to infer the type

```
{
    "size": 1000,
    "_source": false,
    "fields": [
        {
            "field": "name"
        },
        {
            "field": "version"
        }
    ],
    "sort": [
        {
            "_script": {
                "script": {
                    "lang": "painless",
                    "source":"$('version', new Version('0'))"
                },
                "type": "version"
            }
        }
    ]
}

```

luigidellaquila:
@elasticmachine update branch

luigidellaquila:
Thank you very much for your quick feedback @cbuescher 

For now, the only blocker I have regarding x-pack is the following:
- `"source":"$('version', new Version('0'))"` returns a Version, and I have the BytesRef, so everything works fine
- `"source": "doc['version'].value"` returns a String, so it should be encoded to Version, but I didn't find a way to access VersionEncoder without breaking the dependencies

The solution with $ is enough for my immediate needs (ie. the SQL fix), but it seems a half-baked solution to me, and since many users are used to `doc['version'].value` syntax, they would probably expect it to work as well.

Please take your time to review the code and ping me if you have some time to discuss the solution

Thanks

Luigi

server/src/main/java/org/elasticsearch/search/sort/ScriptSortBuilder.java
```diff
@@ -246,10 +248,15 @@ public static ScriptSortBuilder fromXContent(XContentParser parser, String eleme
 
     @Override
     public SortFieldAndFormat build(SearchExecutionContext context) throws IOException {
-        return new SortFieldAndFormat(
-            new SortField("_script", fieldComparatorSource(context), order == SortOrder.DESC),
-            DocValueFormat.RAW
-        );
+        DocValueFormat docValueFormat = DocValueFormat.RAW;
+        if ("version".equals(this.type.toString())) {
+            MappedFieldType fieldType = context.getFieldType("version");
```
luigidellaquila, line 253: this is wrong... I thought it returned a field based on the type name, but it's based on the *field* name.

x-pack/plugin/mapper-version/src/main/java/org/elasticsearch/xpack/versionfield/Version.java
```diff
@@ -17,7 +19,7 @@
  * TODO(stu): implement {@code Comparable<Version>} based on {@code VersionEncoder#prefixDigitGroupsWithLength(String, BytesRefBuilder)}
  * See: https://github.com/elastic/elasticsearch/issues/82287
  */
-public class Version implements ToXContent {
+public class Version implements ToXContent, BytesRefProducer, Comparable<Version> {
     protected String version;
 
     public Version(String version) {
```
stu-elastic, line 25: I wonder if we should have another constructor that takes the encoded version, this String version is mostly for users.

luigidellaquila: Do you mean a `protected Version(String version, BytesRef ref)` for internal use only, so that we don't have to recalculate the BytesRef? I guess having a constructor with BytesRef would have the same problem, ie. we would have to decode it for the string value.

luigidellaquila:
@elasticmachine update branch

x-pack/plugin/mapper-version/src/main/java/org/elasticsearch/xpack/versionfield/Version.java
cbuescher: From a quick look I think we can have a ctor that only takes the BytesRef and internally decodes it to a String only once, then use one for the xContent/toString part and the other for comparison. `Version` only seems to be created from `VersionStringDocValuesField` where we can also directly get the bytes, in fact currently we do the decoding step there once in `getInternal`, that could happen in the ctor instead.

x-pack/plugin/mapper-version/src/main/java/org/elasticsearch/xpack/versionfield/VersionStringDocValuesField.java
```diff
@@ -143,8 +144,19 @@ public Version next() {
...
                     throw new NoSuchElementException();
                 }
-                return new Version(getInternal(index++));
+
+                return getVersionFromInternal(index++);
             }
         };
     }
+
+    private Version getVersionFromInternal(int index) {
```
cbuescher, line 153: Is this needed if we create the Version from a BytesRef directly?

cbuescher (commented):
I did a quick pass on this and have to admit I don't fully understand the details of setting up and wiring a new ScriptSortType but if making "Version" comparable gets the deal done, I think that approach works well even without having to move the VersionEncoder or other stuff to the server module. If that would make things easier on your side that's certainly something we can consider, I'd have to find out who to talk to for a move like that though because it would probably involve some slight license change on the code involved. 
I left a few minor comments. Maybe it would be possible to use some of the ordering tests from `VersionEncoderTests` to check that the `Version` classes compare method works as expected. Maybe just use the cases from `testEncodingOrderingSemver()` because I think they cover most of the interesting cases.

server/src/main/java/org/elasticsearch/script/BytesRefProducer.java
```diff
@@ -0,0 +1,16 @@
...
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.script;
+
+import org.apache.lucene.util.BytesRef;
+
+public interface BytesRefProducer {
```
cbuescher, line 13: Any plans to use this interface on anything other than `Version`? Otherwise maybe the instance check this is used here could simple test for `Version` instead. Or is this here so you don't need to reference `Version` from within the server module?

x-pack/plugin/mapper-version/src/main/java/org/elasticsearch/xpack/versionfield/VersionStringDocValuesField.java
luigidellaquila: Yes, it makes sense, we can do the decoding in the ctor directly. We won't need getVersionFromInternal.
Fixing.

server/src/main/java/org/elasticsearch/script/BytesRefProducer.java
luigidellaquila: The interface is there only to decouple this fix from x-pack, I'd be happy to remove it if we move Verson to server module.
Just for completeness, we could use this interface (and the same fix) for at least another case, that is [IP](https://www.elastic.co/guide/en/elasticsearch/reference/current/ip.html) field type, but the interface would still be overkill (an if/else would be more than enough probably)

x-pack/plugin/mapper-version/src/main/java/org/elasticsearch/xpack/versionfield/VersionStringDocValuesField.java
luigidellaquila: I started to do this change and I realised one thing that is worth pointing out: Version(String) constructor is public API (see last Painless test in [20_scripts.yml](https://github.com/elastic/elasticsearch/pull/85990/files#diff-dcc6c3b83691ea8470049bca3dcad374992a030cf25c65bd9e6be3ff0f67c5a6R82)) and it has to encode the string to get the BytesRef. So if we decide to move Version to Server module, we will have to move VersionEncoder as well.

server/src/main/java/org/elasticsearch/script/BytesRefProducer.java
cbuescher: I see, thats what I guessed. No worries, I think if moving complicates things right now the interface is fine. Maybe this could benefit from a small class comment why its there then, so the next person doesn't ask again.

x-pack/plugin/mapper-version/src/main/java/org/elasticsearch/xpack/versionfield/Version.java
stu-elastic: We still need the `Version(String)` constructor is [whitelisted](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/mapper-version/src/main/resources/org/elasticsearch/xpack/versionfield/org.elasticsearch.xpack.versionfield.txt#L15) for painless users.  Users have to have the ability to provide a `Version` as a default value when fetching version fields, as in this [test](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/mapper-version/src/yamlRestTest/resources/rest-api-spec/test/50_script_values.yml#L40).

luigidellaquila: Yes, I'm keeping both (see last commit)

luigidellaquila:
> We want to move users towards the fields API (the $ syntax). I would want to wait for strong user demand before adding a version that works with doc.

No problem for me, I can use the field API from SQL, so it's not a blocker.
If we agree on this, I can easily "disable" the `doc[]` syntax

x-pack/plugin/mapper-version/src/yamlRestTest/resources/rest-api-spec/test/50_script_values.yml
```diff
@@ -40,6 +42,70 @@ setup:
...
-  - match: { hits.hits.2.fields.field.0: "1.2.3-abc+def" }
-  - match: { hits.hits.3.fields.field.0: "1.0.0" }
+  - match: { hits.hits.1.fields.field.0: "1.12.3" }
+  - match: { hits.hits.2.fields.field.0: "1.2.3.4.5" }
+  - match: { hits.hits.3.fields.field.0: "1.2.3-abc+def" }
+  - match: { hits.hits.4.fields.field.0: "1.0.0" }
+
+
+---
+"Sort by Version script value (asc)":
```
stu-elastic, line 52: This will fail on other versions, you'll have to move this into a new yaml test with the setup
```
  - skip:
      version: " - 8.2.99"
      reason: "version script field sorting support was added in 8.3.0"
``

**Comment by luigidellaquila** [comment]
@stu-elastic I removed the support for `doc['version'].value` in this context. In case of user demand, the fix is easy

**Review by stu-elastic — APPROVED** [review]
Can you add the doc-values version back?  After reviewing the removal, it doesn't clean much up and I'm sure users would appreciate it.

The changes on the scripting side look good.

File: x-pack/plugin/mapper-version/src/main/java/org/elasticsearch/xpack/versionfield/Version.java [hunk 6]
```diff
@@ -7,21 +7,28 @@
...
 
 import java.io.IOException;
 
 /**
  * Script value class.
- * TODO(stu): implement {@code Comparable<Version>} based on {@code VersionEncoder#prefixDigitGroupsWithLength(String, BytesRefBuilder)}
- * See: https://github.com/elastic/elasticsearch/issues/82287
  */
-public class Version implements ToXContent {
+public class Version implements ToXContent, BytesRefProducer, Comparable<Version> {
```
cbuescher, line 20: Minor nit, unrelated to this change actually, but could you change this class to implement `ToXContentFragment` and remove the `isFragment` method? We use `ToXContentFragment`  as a marker interface for things that don't output full json object, and in this case I think the class itself only outputs a value to the xcontent builder.

cbuescher (approved):

luigidellaquila:
@stu-elastic @cbuescher thank you very much for you support.

@sethmlarson I see you labeled this PR for `Team:Clients` as well, is there anything specific to check?

sethmlarson (commented):
This looks good from a clients perspective, no comments from me. I have a script that adds `Team:Clients` to PRs that change API specs or YAML tests so we can be alerted of upcoming changes.

PR #85994, opened 2022-04-19
Ignore missing JDK 18 java.desktop finalize methods

### LINKED ISSUE #85992: Build fails with JDK 18 -  Method not found while parsing signature: java.awt.color.ICC_Profile#finalize()
```
$./gradlew precommit
...
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':client:client-benchmark-noop-api-plugin:forbiddenApisMain'.
> Parsing signatures failed: Method not found while parsing signature: java.awt.color.ICC_Profile#finalize()

...

Publishing build scan...
https://gradle-enterprise.elastic.co/s/rhm5f4imnnae2
```

JDK 18 removed a number of deprecated-for-removal empty finalize() in java.desktop module, see https://bugs.openjdk.java.net/browse/JDK-8273103

The jdk-deprecated forbidden APIs list should be updated to reflect this.

ChrisHegarty:
This change allows to build ES with JDK 18. Prior to this change the build fails because of missing java.desktop finalize methods, during the processing of the forbidden deprecated APIs list.

JDK 18 removed a number of deprecated-for-removal empty finalize() in java.desktop module, see https://bugs.openjdk.java.net/browse/JDK-8273103 

closes #85992

### DIFF
--- a/build-tools-internal/src/main/resources/forbidden/jdk-deprecated.txt
+++ b/build-tools-internal/src/main/resources/forbidden/jdk-deprecated.txt
@@ -9,7 +9,6 @@
 # This file contains all public, deprecated API signatures in Java version 17 (extracted from build 17).
 
 @ignoreMissingClasses
-@ignoreUnresolvable
 @defaultMessage Deprecated in Java 17
 
 java.applet.Applet
@@ -142,7 +141,6 @@ java.awt.Window#hide()
 java.awt.Window#postEvent(java.awt.Event)
 java.awt.Window#reshape(int,int,int,int)
 java.awt.Window#show()
-java.awt.color.ICC_Profile#finalize() # removed in Java 18
 java.awt.datatransfer.DataFlavor#equals(java.lang.String)
 java.awt.datatransfer.DataFlavor#normalizeMimeType(java.lang.String)
 java.awt.datatransfer.DataFlavor#normalizeMimeTypeParameter(java.lang.String,java.lang.String)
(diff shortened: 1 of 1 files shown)


ChrisHegarty:
The changes proposed, to add `@ignoreUnresolvable`, result in a warning during the build - but at least this is an incremental improvement over the build failure!   There is probably a better way to suppress these?

```
> Task :server:forbiddenApisMain
'@ignoreUnresolvable' inside signatures files is deprecated, prefer using '@ignoreMissingClasses' to ignore signatures where the class is missing.
Method not found while parsing signature: java.awt.color.ICC_Profile#finalize() # removed in Java 18 [signature ignored]
Method not found while parsing signature: java.awt.image.ColorModel#finalize() # removed in Java 18 [signature ignored]
Method not found while parsing signature: java.awt.image.IndexColorModel#finalize() # removed in Java 18 [signature ignored]
```

build-tools-internal/src/main/resources/forbidden/jdk-deprecated.txt
```diff
@@ -9,6 +9,7 @@
 # This file contains all public, deprecated API signatures in Java version 17 (extracted from build 17).
 
 @ignoreMissingClasses
+@ignoreUnresolvable
```
mark-vieira, line 12: How granular is this? Does this apply to _all_ unresolved methods or just the ones in this file? Can we be more granular here and just ignore these known removed methods?

Also, do we know which third party library is using these removed methods? Can we update to a version that's fixed the issue?

mark-vieira:
Looking at this again I think the changes here are fine. I wasn't aware what the issue was at first and that this is just about valiating the signature file itself. I think we can move ahead with the annotation, or even just remove those instances of `finalize()` altogether.

ChrisHegarty:
> I think we can move ahead with the annotation, or even just remove those instances of `finalize()` altogether.

I reverted the change to add the annotation, in favour of just removing the three AWT related members. This avoids the aforementioned warning. Additionally, it seems incredibly unlikely that any of these AWT methods would ever be used, and even if so then a JDK 18 build job would catch it quickly.

pugnascotia (approved):

PR #86004, opened 2022-04-19
[ML] Model assignment planner

dimitris-athanasiou:
Adds an assignment planner for trained model deployments.
The planner uses a linear programming solver (ojalgo) and randomized
rounding in order to compute an assignment plan that tries to maximize
the number of allocations whilst minimizing memory used and ensuring
deployments with assignments will not get fewer allocations than they
did before.


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlanner.java
```diff
@@ -0,0 +1,193 @@
...
+                }
+            }
+        }
+        return mergedPlanBuilder.build();
+    }
+
+    private enum PreserveAllocations {
+        ONE,
+        ALL;
+    }
```
benwtrent, line 159: I think all this code would be simpler, and probably less susceptible to bugs (as it will reduce predicates on this down to 1), if we had a 

`AbstractPreserveAllocation` class and a `PreserveOneAllocation` and `PreserveAllAllocation` or something. 

Their constructors take the `List<Node> nodes, List<Model> models` and can calculate the required values. 

Then, we would have only one switch statement (constructing those subclasses).

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlanner.java
```diff
@@ -0,0 +1,193 @@
...
+        this.models = models.stream().sorted(Comparator.comparing(Model::id)).toList();
+    }
+
+    public AssignmentPlan computePlan() {
+        AssignmentPlan planKeepingOneAllocationOnPreviousAssignments = solveKeepingOneAllocationOnPreviousAssignments();
+        AssignmentPlan bestPlan = planKeepingOneAllocationOnPreviousAssignments.satisfiesPreviousAssignments()
+            ? planKeepingOneAllocationOnPreviousAssignments
+            : solvePreservingAllPreviousAssignments();
+        logger.debug(() -> "Best plan =\n" + bestPlan.prettyPrint());
+        logger.debug(prettyPrintOverallStats(bestPlan));
```
benwtrent, line 56:
```suggestion
        logger.debug(() -> prettyPrintOverallStats(bestPlan));
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlan.java
```diff
@@ -0,0 +1,323 @@
...
+            return Objects.hashCode(id);
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            Node that = (Node) o;
+            return Objects.equals(id, that.id);
+        }
```
benwtrent, line 82: This is sort of sneaky. You do this because you want to have the `Node` object as a hashtable later in the linear solver. But these objects are not actually equivalent as the memory and cores could be drastically different.

Could we instead use the `id` in the hash tables directly?

The same concern goes for the `Model` class.

If we are stuck on this, then at least these classes should be package private.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
```diff
@@ -0,0 +1,341 @@
...
+            return AssignmentPlan.builder(nodes, models).build();
+        }
+
+        Tuple<Map<Tuple<Model, Node>, Double>, AssignmentPlan> weightsAndBinPackingPlan = calculateWeightsAndBinPackingPlan();
+
+        Map<Tuple<Model, Node>, Double> allocationValues = new HashMap<>();
+        Map<Tuple<Model, Node>, Double> assignmentValues = new HashMap<>();
+        if (solveLinearProgram(weightsAndBinPackingPlan.v1(), allocationValues, assignmentValues) == false) {
+            return weightsAndBinPackingPlan.v2();
+        }
```
benwtrent, line 92: Is it possible to know that you cannot get any better than the `weightsAndBinPackingPlan`? It seems like a big waste to run the linear program and THEN pick the bin packing solution.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
```diff
@@ -0,0 +1,341 @@
...
+                        assignmentPlan.assignModelToNode(m, n, allocations);
+                        weights.put(Tuple.tuple(m, n), w);
+                        w -= dw;
+                        break;
+                    }
+                }
+                if (lastW == w || assignmentPlan.getRemainingAllocations(m) == 0) {
+                    break;
+                }
+            }
```
benwtrent, line 147:
```suggestion
            double lastW = w;
            do {
                List<Node> orderedNodes = nodes.stream()
                    .sorted(Comparator.comparingDouble(n -> dsafNodeOrder(n, m, assignmentPlan)))
                    .toList();
                for (Node n : orderedNodes) {
                    int allocations = Math.min(
                        assignmentPlan.getRemainingCores(n) / m.threadsPerAllocation(),
                        assignmentPlan.getRemainingAllocations(m)
                    );
                    if (allocations > 0 && assignmentPlan.canAssign(m, n, allocations)) {
                        assignmentPlan.assignModelToNode(m, n, allocations);
                        weights.put(Tuple.tuple(m, n), w);
                        w -= dw;
                        break;
                    }
                }
            } while(lastW != w && assignmentPlan.getRemainingAllocations(m) != 0);
```

Seems safer than having a `while(true)` to me

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
```diff
@@ -0,0 +1,341 @@
...
+        }
+
+        for (Model m : models) {
+            for (Node n : nodes) {
+                if (weights.containsKey(Tuple.tuple(m, n)) == false) {
+                    weights.put(Tuple.tuple(m, n), random.nextDouble(minWeight(m, n, w), maxWeight(m, n, w)));
+                }
```
benwtrent, line 154:
```suggestion
                weights.computeIfAbsent(Tuple.tuple(m, n), () -> random.nextDouble(minWeight(m, n, w), maxWeight(m, n, w)));
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
```diff
@@ -0,0 +1,341 @@
...
+
+        return Tuple.tuple(weights, binPackingPlan);
+    }
+
+    private double dsafModelOrder(Model m) {
+        return (m.currentAllocationByNodeId().isEmpty() ? 1 : 2) * -normalizedMemoryPerModel.get(m) * m.threadsPerAllocation();
+    }
+
+    private double dsafNodeOrder(Node n, Model m, AssignmentPlan.Builder assignmentPlan) {
+        // as java.lang.Math#abs(long) is a forbidden API, we compute an abs manually
```
benwtrent, line 170: You should do something similar `BucketCountKSTestAggregator` except with `long`

```
 @SuppressForbidden(reason = "Math#abs(int) is safe here as we protect against MIN_VALUE")
    private static int nonNegative(int x) {
        if (x == Integer.MIN_VALUE) {
            throw new AggregationExecutionException("unexpected value while interpolating sampled values");
        }
        return Math.abs(x);
    }
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
```diff
@@ -0,0 +1,341 @@
...
+                .upper(coresPerNode.get(n))
+                .setLinearFactors(varsForNode(n, allocationVars), Access1D.wrap(threadsPerAllocationPerModel));
+        }
+
+        for (Node n : nodes) {
+            List<Variable> allocations = new ArrayList<>();
+            List<Double> modelMemories = new ArrayList<>();
+            models.stream().filter(m -> m.currentAllocationByNodeId().containsKey(n.id()) == false).forEach(m -> {
+                allocations.add(allocationVars.get(Tuple.tuple(m, n)));
+                modelMemories.add(normalizedMemoryPerModel.get(m) * m.threadsPerAllocation() / (double) coresPerNode.get(n));
```
benwtrent, line 272: Is this supposed to be
```
m_i * a_i_j * t_i / N_j
```
Or 
```
(a_i_j * t_i) / N_j
```

(If I am reading this correctly, I think its `m_i * a_i_j * t_i / N_j` as its t_i/N-j * memory and then we use `a_i_j` as the list of values...)

Overall comment to prevent confusion, it would be good in addition to the overall comment regarding the logic of the linear equations, that each line contains which constraint its building (according to the overall comment).

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,419 @@
...
+        if (softAssignmentQueue.isEmpty() == false) {
+            logger.debug(() -> new ParameterizedMessage("Random assignment rounding across [{}] rounds", rounds));
+            for (int i = 0; i < rounds; i++) {
+                AssignmentHolder randomizedAssignments = new AssignmentHolder(assignmentHolder);
+                randomizedAssignments.doRandomizedRounding(softAssignmentQueue);
+                AssignmentPlan randomizedPlan = randomizedAssignments.toPlan();
+                if (randomizedPlan.compareTo(bestPlan) > 0) {
+                    bestPlan = randomizedPlan;
+                }
+            }
```
benwtrent, line 74: arg, I wish there was some "epsilon" we were approaching here instead of literal "randomly do things X number of times and pick the best"

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,419 @@
...
+                remainingNodeMemory.compute(n, (k, v) -> v - m.memoryBytes());
+            }
+            remainingNodeCores.compute(n, (k, v) -> v - allocations * m.threadsPerAllocation());
+            remainingModelAllocations.compute(m, (k, v) -> v - allocations);
+        }
+    }
+
+    private static boolean isInteger(double value) {
+        // it is possible that the solver results in values that are really close to an int, we should treat those as ints
+        return Double.isFinite(value) && Math.abs(value - Math.rint(value)) < 1e-6;
```
benwtrent, line 417: please make `1e-6` a static variable or something, `EPS` is good enough.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,419 @@
...
+            Model m,
+            int remainingNodeCores,
+            long remainingNodeMemory,
+            int remainingModelAllocations
+        ) {
+            // as java.lang.Math#abs(long) is a forbidden API, we compute an abs manually
+            int distanceOfRemainingCoresToModelThreads = remainingNodeCores - remainingModelAllocations * m.threadsPerAllocation();
+            if (distanceOfRemainingCoresToModelThreads < 0) {
+                distanceOfRemainingCoresToModelThreads = -distanceOfRemainingCoresToModelThreads;
+            }
```
benwtrent, line 372: Again, you should create a protected and safe absolute function. See previous comment referring to `BucketCountKSTestAggregator`

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,419 @@
...
+            return (m.currentAllocationByNodeId().isEmpty() ? 1 : 2) * -m.memoryBytes();
+        }
+
+        private boolean hasSoftAssignments(Node n) {
+            return models.stream().anyMatch(m -> isSoftAssignment(m, n));
+        }
+
+        private boolean isSoftAssignment(Model m, Node n) {
+            Tuple<Model, Node> index = Tuple.tuple(m, n);
+            return (softAssignments.get(index) > 0 && softAssignments.get(index) < 1) || isInteger(softAllocations.get(index)) == false;
```
benwtrent, line 215: shouldn't this also do a `null` check?

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,419 @@
...
+            // random rounding finds.
+
+            Map<Tuple<Model, Node>, Integer> resultAllocations = new HashMap<>();
+
+            ResourceTracker resourceTracker = new ResourceTracker(nodes, models);
+
+            for (Model m : models) {
+                for (Node n : nodes) {
+                    Tuple<Model, Node> assignment = Tuple.tuple(m, n);
+                    int allocations = (int) Math.floor(softAllocations.getOrDefault(assignment, 0.0));
```
benwtrent, line 317: OK, I think this `softAllocations` and `softAllocations` all needs to be wrapped up. Instead of `put(0.0)` can we remove the key? 

And not access those maps directly except through specific methods? It seems like "knowledge of how to access values and update them, etc." is bleeding throughout with these maps and it may be easier to follow if they were contained somehow.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,419 @@
...
+            int distanceOfRemainingCoresToModelThreads = remainingNodeCores - remainingModelAllocations * m.threadsPerAllocation();
+            if (distanceOfRemainingCoresToModelThreads < 0) {
+                distanceOfRemainingCoresToModelThreads = -distanceOfRemainingCoresToModelThreads;
+            }
+            return (m.currentAllocationByNodeId().containsKey(n.id()) ? 0 : 1) + (remainingNodeCores <= remainingModelAllocations * m
+                .threadsPerAllocation() ? 0 : 0.5) + (0.01 * distanceOfRemainingCoresToModelThreads) + (0.01 * remainingNodeMemory);
+        }
+    }
+
+    private static class ResourceTracker {
```
benwtrent, line 378: why not a record? Because you want things `final`?

benwtrent (commented):
Just the first pass for now. 

Would be good for a second review to take a gander.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlan.java
```diff
@@ -0,0 +1,323 @@
...
+                + allocations
+                + ") (threads_per_allocation = "
+                + threadsPerAllocation
+                + ") (current_allocations = "
+                + currentAllocationByNodeId
+                + ")";
+        }
+    };
+
+    public record Node(String id, long availableMemoryBytes, int cores) {
```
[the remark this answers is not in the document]
dimitris-athanasiou: `NodeLoad` is much more complex than what we need here. I'd rather keep it as it is and see if it makes sense to replace when we wire all this up with execution if that is ok with you.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
dimitris-athanasiou: I don't think there's a way to know the bin-packing solution is better before running the solver. At least not without adding complexity and I'm not sure it's worth it given the runtime of this and the fact that it only happens on specific events (model starts, nodes joining, etc.). Also, `solveLinearProgram` may return `false` which means the solver couldn't find a feasible solution.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
dimitris-athanasiou: This doesn't work because `w` isn't final and can't be used in lambdas.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
dimitris-athanasiou: It is `m_i * a_i_j * t_i / N_j`. I'll add comments for each constraint.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
benwtrent: >  is better

I am thinking that we determine that it is "good enough" like we have some acceptable amount of waste.

But sure, I understand it's a premature optimization. I am just slightly concerned over the fact that this is running on a master node with a bunch of other things. 

I suppose we just need to be careful that plans are not spammed in a "rolling upgrade" scenario or something (which is a separate optimization

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
dimitris-athanasiou: We initialize `softAssignments` so that it contains an entry for every possible assignment (m, n).

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
dimitris-athanasiou: Given its constructor takes different arguments than its members, it wouldn't gain us much to make it a record.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
dimitris-athanasiou: Actually the `getOrDefault` is unnecessary here. There should always be a value for all possible `(m, n)`.

`AssignmentHolder` is all about interacting with those values. Not sure how I could wrap them further. I'll have a think.

benwtrent: @dimitris-athanasiou it just seems that every time the interaction takes place, the same logic is used: make tuple, get it, check if its zero or something. Additionally, it seems that almost in every case, both maps need to be updated.

I am thinking that this should be wrapped up in a method or something. Additionally, storing zero in the maps seems weird. Why not delete keys and treat `0` as the default value?

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlan.java
dimitris-athanasiou: It turns out I can get rid of those sneaky `equals` indeed. The only place where we need to match models/nodes that can be different is when we merge the assignment plan with its preserved allocations. It used to be different which is why I had done this hack. But later refactoring meant we actually don't need it anymore.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
dimitris-athanasiou: I have extracted a method for unassigning a soft assignment in 437ec1b80fd1122a100734b40a39ab2845e6d04b.

I'd rather we kept `0` values in there as otherwise we'd have to be doing `contains` checks in a bunch of places.

dimitris-athanasiou:
@elasticmachine update branch

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
benwtrent: > I'd rather we kept 0 values in there as otherwise we'd have to be doing contains checks in a bunch of places.

You shouldn't have to. `getOrDefault` is exactly what you want.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
benwtrent: ARG!

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
davidkyle: `softAssignments` could be a `m * n` array of doubles index by the model and node order avoiding the use of the Tuple to key the map entry.

davidkyle: `softAssignments` could be a `m * n` array of doubles index by the model and node order avoiding the use of the Tuple to key the map entry.

davidkyle (commented):
There are many uses of maps of maps and tuples where I think plain Java objects could be used. The maps hurt readability as the methods `get()`, `keys()`, `values()` etc are not as informative as say `numberOfAllocations()` or `models()`.

That's a stylistic preference and not a blocker to merging, otherwise the code looks good

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlan.java
```diff
@@ -0,0 +1,323 @@
...
+            return id + " (mem = " + ByteSizeValue.ofBytes(availableMemoryBytes) + ") (cores = " + cores + ")";
+        }
+    };
+
+    /**
+     * The model assignments to each node. Each assignment
+     * is a {@link Map<Node, Integer>} where the value is the number
+     * of allocations for each node.
+     */
+    private final Map<Model, Map<Node, Integer>> assignments;
```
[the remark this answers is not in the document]
dimitris-athanasiou: I got rid of the sneaky overrides. Not sure about the `Model` having an `assignments` member. The idea was that `Model` describes the requirements necessary for the model allocations, not the eventual assignments. In particular, the model wouldn't have its `assignments` populated before the planner does its work. Does this make sense?

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
```diff
@@ -0,0 +1,341 @@
...
+    private Tuple<Map<Tuple<Model, Node>, Double>, AssignmentPlan> calculateWeightsAndBinPackingPlan() {
+        logger.debug(() -> "Calculating weights and bin packing plan");
+
+        double w = INITIAL_W;
+        double dw = w / nodes.size() / models.size();
+
+        Map<Tuple<Model, Node>, Double> weights = new HashMap<>();
+        AssignmentPlan.Builder assignmentPlan = AssignmentPlan.builder(nodes, models);
+
+        for (Model m : models.stream().sorted(Comparator.comparingDouble(this::dsafModelOrder)).toList()) {
```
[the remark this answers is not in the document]
dimitris-athanasiou: Good point. It is `descending size any fit`. I'll rename the methods to prevent confusion.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
dimitris-athanasiou: But at places we need to access them having a `(m, n)` pair. How would we convert those to the right indices?

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlan.java
```diff
@@ -0,0 +1,295 @@
...
+
+        long getRemainingMemory(Node n) {
+            return remainingNodeMemory.get(n);
+        }
+
+        int getRemainingThreads(Model m) {
+            return remainingModelAllocations.get(m) * m.threadsPerAllocation();
+        }
+
+        int getRemainingAllocations(Model m) {
```
benwtrent, line 225: All these are unboxing and could easily through NPE if something provided a model that was not added to the hashtable (same goes for the node accessors). 

They are package private, which is good. This just all feels very brittle.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
benwtrent: store m and n in separate maps to an int the int is the index in the double array. I am not sure this is better or not.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,428 @@
...
+                    int allocations = (int) Math.floor(softAllocations.getOrDefault(assignment, 0.0));
+                    resultAllocations.put(assignment, allocations);
+                    if (allocations > 0) {
+                        resourceTracker.assign(m, n, allocations);
+                    }
+                }
+            }
+
+            for (Model m : models.stream().sorted(Comparator.comparingDouble(this::remainingModelOrder)).toList()) {
+                if (resourceTracker.remainingModelAllocations.get(m) > 0) {
```
benwtrent, line 331: Both of the sorted for loops (models and nodes) have a predicate check that just skips whole sections. Wouldn't it be better to filter before sorting?

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlan.java
```diff
@@ -0,0 +1,323 @@
...
+            remainingNodeMemory = new HashMap<>();
+            remainingNodeCores = new HashMap<>();
+            remainingModelAllocations = new HashMap<>();
+
+            for (Model m : models) {
+                Map<Node, Integer> nodeAssignments = new HashMap<>();
+                for (Node n : nodes) {
+                    nodeAssignments.put(n, 0);
+                    remainingNodeMemory.put(n, n.availableMemoryBytes());
+                    remainingNodeCores.put(n, n.cores());
```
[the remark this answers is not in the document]
dimitris-athanasiou: I will add sizing to those maps and populate the node related ones in a separate loop.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlan.java
```diff
@@ -0,0 +1,295 @@
...
+
+    /**
+     * The assignments of that model to each node. Each assignment
+     * is a {@link Map<Node, Integer>} where the value is the number
+     * of allocations for each node.
+     * @param model the model for which assignments are returned
+     * @return the model assignments per node
+     */
+    @Nullable
+    public Map<Node, Integer> assignments(Model model) {
```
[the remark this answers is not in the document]
dimitris-athanasiou: I'll change this to return optional

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlan.java
dimitris-athanasiou: These are methods on the `Builder` which should have been constructed with the list of nodes/models that are used. Passing a model/node that wasn't part of the collections used for constructing the builder would be a programmatic error.

tveasey (commented):
I've done a pass through. I need to read `RandomizedAssignmentRounding` more carefully in a second pass. Overall it all looks as intended and I like the design! I've made some minor suggestions.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AbstractPreserveAllocations.java
```diff
@@ -0,0 +1,97 @@
...
+    AssignmentPlan mergePreservedAllocations(AssignmentPlan assignmentPlan) {
+        final Map<Tuple<String, String>, Integer> assignmentsByModelNodeIdPair = new HashMap<>();
+        for (Model m : assignmentPlan.models()) {
+            Map<Node, Integer> assignments = assignmentPlan.assignments(m);
+            if (assignments != null) {
+                for (Map.Entry<Node, Integer> nodeAssignment : assignments.entrySet()) {
+                    assignmentsByModelNodeIdPair.put(Tuple.tuple(m.id(), nodeAssignment.getKey().id()), nodeAssignment.getValue());
+                }
+            }
+        }
```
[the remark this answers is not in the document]
dimitris-athanasiou: The problem is that the model/node objects of the `assignmentPlan` here could differ to the model/node objects stored in `models` and `nodes` members. The reason is that we modified them in order to account for previous assignments before computing the plan. That is why I create the lookup table based on `String` ids instead of the model/node objects themselves. I'll go back and add a comment to clarify that.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlan.java
```diff
@@ -0,0 +1,295 @@
...
+                        + allocations
+                        + "] allocations to model ["
+                        + model.id()
+                        + "]; required threads per allocation ["
+                        + model.threadsPerAllocation()
+                        + "]"
+                );
+            }
+
+            assignments.get(model).put(node, allocations);
```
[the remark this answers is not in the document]
dimitris-athanasiou: That is a good point. I'll do that.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,428 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Random;
+import java.util.Set;
+
+/**
+ * Solving allocation distribution using linear programming requires relaxing the allocation
+ * and assignment variables. This means that even though in reality they take discreet integer
```
tveasey, line 31: typo
```suggestion
 * and assignment variables. This means that even though in reality they take discrete integer
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,428 @@
...
+ * values, we allow the solver to provide solutions that are real numbers. This is a common
+ * technique used in LP. A common way to convert the relaxed solution back in integer values
+ * is to apply randomized rounding. This class performs randomized rounding expecting
+ * double values for the number of allocations each model gets on each node, as well as a double
+ * value in [0-1] for whether there is an assignment for each model to each node.
```
tveasey, line 36:
```suggestion
 * value in [0,1] for whether there is an assignment for each model to each node.
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
```diff
@@ -0,0 +1,346 @@
...
+        this.nodes = nodes;
+
+        // Filter out models that are not already assigned and do not fit on any node
+        long maxNodeMemory = nodes.stream().map(Node::availableMemoryBytes).max(Long::compareTo).orElse(0L);
+        this.models = models.stream()
+            .filter(m -> m.currentAllocationByNodeId().isEmpty() == false || m.memoryBytes() <= maxNodeMemory)
+            .toList();
+
+        maxNodeCores = this.nodes.stream().map(Node::cores).max(Integer::compareTo).orElse(0);
+        maxModelMemoryBytes = this.models.stream().map(Model::memoryBytes).max(Long::compareTo).orElse(0L);
```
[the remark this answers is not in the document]
dimitris-athanasiou: For this to end up being zero we'd have to have no models. But I guess it doesn't hurt to change it to `1` here. I'll do that.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,428 @@
...
+                        Tuple<Model, Node> index = Tuple.tuple(m, n);
+
+                        int allocationsToAssign = 0;
+                        if (isInteger(softAllocations.get(index))) {
+                            // We round this separately because if we used ceil and the value was just about the
+                            // integer value we'll use one additional allocation when we shouldn't.
+                            allocationsToAssign = (int) Math.rint(softAllocations.get(index));
+                        } else if (maxTotalThreads <= n.cores()) {
+                            allocationsToAssign = (int) Math.ceil(softAllocations.get(index));
+                        }
```
tveasey, line 148: I think we should use the same logic to compute `maxTotalThreads` as we use here, i.e. with rounding. Also `maxTotalThreads <= n.cores()` is redundant since you only enter if this is true. I would factor out into a method:
```
int allocationsToAssign(...) {
    if (isInteger(softAllocations.get(index))) {
        // We round this separately because if we used ceil and the value was just about the
        // integer value we'll use one additional allocation when we shouldn't.
        return (int) Math.rint(softAllocations.get(index));
    }
    return (int) Math.ceil(softAllocations.get(index));
}
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,428 @@
...
+
+                        if (softAssignments.get(index) > 0 && softAssignments.get(index) < 1) {
```
tveasey, line 150: Seems like this should be checked before computing `allocationsToAssign`.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
dimitris-athanasiou: OK, I'll do that.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,428 @@
...
+        private final Map<Tuple<Model, Node>, Double> softAllocations = new HashMap<>();
+        private final ResourceTracker resourceTracker;
+
+        private AssignmentHolder() {
+            resourceTracker = new ResourceTracker(nodes, models);
+        }
+
+        private AssignmentHolder(AssignmentHolder holder) {
+            softAssignments.putAll(holder.softAssignments);
+            softAllocations.putAll(holder.softAllocations);
```
tveasey, line 99: I find the naming of these a bit confusing since you convert them from soft into hard. One this basis I would prefer `assignments` and `allocations`.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,428 @@
...
+        private boolean isSoftAssignment(Model m, Node n) {
+            Tuple<Model, Node> index = Tuple.tuple(m, n);
+            return (softAssignments.get(index) > 0 && softAssignments.get(index) < 1) || isInteger(softAllocations.get(index)) == false;
+        }
+
+        private void zeroSoftAssignmentsOfSatisfiedModels() {
+            for (Model m : models) {
+                if (resourceTracker.remainingModelAllocations.get(m) <= 0) {
+                    for (Node n : nodes) {
+                        Tuple<Model, Node> index = Tuple.tuple(m, n);
```
tveasey, line 225: nit: inside if.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,428 @@
...
+            return Math.min(softAssignments.get(assignment), 1 - softAssignments.get(assignment));
+        }
+
+        private double assignmentMostRemainingThreadsOrder(Tuple<Model, Node> assignment) {
+            return -softAllocations.get(assignment) * assignment.v1().threadsPerAllocation();
+        }
+
+        private void doRandomizedRounding(List<Tuple<Model, Node>> softAssignmentQueue) {
+            for (Tuple<Model, Node> assignment : softAssignmentQueue) {
+                if (softAssignments.get(assignment) == 1 && isInteger(softAllocations.get(assignment))) {
```
tveasey, line 263: I think this warrants a comment: i.e. because other operations can snap nodes in the queue as a side effect. Also shouldn't this be
```suggestion
                if (isSoftAssignment(m, n) == false) {
                    ...
```
and just after unpacking the tuple.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
```diff
@@ -0,0 +1,428 @@
...
+        private AssignmentPlan toPlan() {
+            AssignmentPlan.Builder builder = AssignmentPlan.builder(nodes, models);
+            for (Map.Entry<Tuple<Model, Node>, Integer> assignment : tryAssigningRemainingCores().entrySet()) {
+                builder.assignModelToNode(assignment.getKey().v1(), assignment.getKey().v2(), assignment.getValue());
+            }
+            return builder.build();
+        }
+
+        private Map<Tuple<Model, Node>, Integer> tryAssigningRemainingCores() {
+            // Eagerly assign allocations to models with larger size first on first node
```
tveasey, line 307:
```suggestion
            // Eagerly assign allocations to models with larger size first on the first node
```

tveasey (commented):
I've read through `RandomizedAssignmentRounding` again and spotted a couple of extra things.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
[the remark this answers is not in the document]
tveasey: I realise it is an impossible case, but I guess it just feels wrong to actually set up a divide by zero.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/RandomizedAssignmentRounding.java
dimitris-athanasiou: Definitely!

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
```diff
@@ -0,0 +1,354 @@
...
+        // ojalgo's MILP functionality makes calls to forbidden APIs that we cannot grant permissions for.
+        // Thus, we relax our variables to be real numbers. After we obtain a solution we then apply randomized rounding in
+        // order to snap the variable values to an integer.
+        //
+        // Adding the y variables to the model results in memory complexity explosion. However, we can substitute y.
+        // In particular, a_i_j * t_i <= N_j * y_i_j is an equality because the objective always increase if we decrease y_i_j.
+        // Thus, we can express: y_i_j = (a_i_j * t_i) / N_j
+        //
+        // The memory constraint is then rewritten as: sum(m_i * a_i_j * t_i / N_j) <= M_j
+        // The objective function becomes: maximize sum(a_i_j * w_i_j) - L1 * sum(m_i * a_i_j * t_i / N_j).
```
tveasey, line 236:
```suggestion
        // The objective function becomes: maximize sum(a_i_j * w_i_j) - L1' * sum(m_i * a_i_j * t_i / N_j).
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/LinearProgrammingPlanSolver.java
```diff
@@ -0,0 +1,354 @@
...
+        // We want to choose L1 in such a way that we will always assign models to a node if there is capacity.
+        // Since w_i_j can be equal to 1 and m_i <= 1, for each node it has to be that L1 * t_i / N_j < 1.
```
tveasey, line 238:
```suggestion
        // Since w_i_j can be equal to 1 and m_i <= 1, for each node it has to be that L1' * t_i / N_j < 1.
```

tveasey (approved):
Just a couple more knock on changes to the explanation, but otherwise LGTM! 🚀

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlan.java
```diff
@@ -0,0 +1,301 @@
...
+                throw new IllegalArgumentException("there should be no duplicate nodes");
+            }
+            if (models.stream().collect(Collectors.toSet()).size() != models.size()) {
+                throw new IllegalArgumentException("there should be no duplicate models");
+            }
+
+            assignments = new HashMap<>(nodes.size() * models.size(), 1.0f);
+            remainingNodeMemory = new HashMap<>(nodes.size(), 1.0f);
+            remainingNodeCores = new HashMap<>(nodes.size(), 1.0f);
+            remainingModelAllocations = new HashMap<>(models.size(), 1.0f);
```
benwtrent, line 199: All of these should use `Maps.
```suggestion
            assignments = Maps.newMapWithExpectedSize(nodes.size() * models.size());
            remainingNodeMemory = Maps.newMapWithExpectedSize(nodes.size());
            remainingNodeCores = Maps.newMapWithExpectedSize(nodes.size());
            remainingModelAllocations = Maps.newMapWithExpectedSize(models.size());
```

benwtrent (approved):
One minor thing, but I think we are g2g. I am sure refactoring will occur as we plug in the other parts of the code.

PR #86015, opened 2022-04-19
Upgrading to tika 2.4

masseyke:
Tika 1.x is end of life as of later this year. This change updates the AttachmentProcessor to use tika 2.
The goal was to keep the functionality as close as possible, just with upgraded tika. The tests have been slightly modified because of a small change in tika functionality -- as of 2.4.0 it now adds an extra newline to the output for every embedded attachment in a document.
Also as part of this I have broken apart the tika-parsers into individual dependencies. The reason is that we are considering breaking this plugin apart, and want to know exactly which parsers we pull in.


masseyke:
@dadoonet I know you use this a lot. You don't happen to have any quality tests that are outside of the codebase do you? I assume that tika 2 is at least as good as tika 1, but we don't have a whole lot of tests to prove it.

masseyke:
In tika 1.x `org.apache.tika.language.LanguageIdentifier` had been deprecated. It was gone in 2.0, so I originally switched AttachmentProcessor to a LanguageDetector instead. I tried the native tika one, but it would throw an exception if the confidence was too low. So I used Optimaize instead. The quality of the results from Optimaize seemed better than the old LanguageIdentifier (on a very limited set of test data). But it looks like late in the 2.0 release they decided to just move LanguageIdentifier to a different package instead of removing it (see https://issues.apache.org/jira/browse/TIKA-3343 and  https://github.com/apache/tika/commit/40690250f9b20703b7a92da17d1f42e6108e9109). And since this PR is just about getting us to 2.0 and not about quality improvements, I'm removing the Optimaize change. We might want to revisit that in the near future.

masseyke:
Note: The tests are currently failing because tika 2.4.0 adds a `\n` for each embedded attachment, which tika 1.x did not do. So the content length for attachments has changed by one byte per embedded attachment. Once this PR is approved, I'll disable those checks in the older branches. #86481 is related to this.

masseyke:
@elasticmachine update branch

plugins/ingest-attachment/src/test/java/org/elasticsearch/ingest/attachment/AttachmentProcessorTests.java
```diff
@@ -284,7 +284,6 @@ public void testEpubDocument() throws Exception {
                 "content_type",
                 "content_length",
                 "date",
-                "keywords",
```
jbaiera, line 287: Is the fact that we don't generate this now a breaking change?

jbaiera (approved):
LGTM - one question is all

plugins/ingest-attachment/src/test/java/org/elasticsearch/ingest/attachment/AttachmentProcessorTests.java
masseyke: I think that's certainly debatable. We're still looking for it, but tika is no longer producing it for epub files (at least not for this one). Tika is still producing it for other file types. Their documentations mentions that they've removed the `Keywords` metadata field in favor of other ones (https://cwiki.apache.org/confluence/display/TIKA/Migrating+to+Tika+2.0.0). However I still see `Keywords` in some of the results we get from tika. I'll look into this a little more.

masseyke: OK, it looks like some parsers generate some of the old values (like "Keywords"), and some generate the new ones (like "dc:subject"). So I changed it to check both. I double-checked all of the metadata property names, and "dc:subject" is the only one of the new names that we weren't checking.

dadoonet: I believe you hit this one: https://issues.apache.org/jira/plugins/servlet/mobile#issue/TIKA-3629

PR #86016, opened 2022-04-19
Add optional module name to plugin properties

rjernst:
In preparation for modularizing Elasticsearch server, each plugin needs
a way to declare itself as modular. When modularized, the main class of
a plugin must be found in a named module. This commit adds a module name
property to the plugin properties template and gradle support for
setting it, as well as automatically inferring the module name when a
module-info.java exists in the source a plugin.


mark-vieira:
Can't we figure this out when we load the plugin by checking in it's jar for a `module-info.class`? Now it seems plugins have to duplicate this configuration.

rjernst:
We don't know which jar file could have it, so we could have to peek into every jar file. Furthermore, we can't just look for any module-info.class in any jar file, since only the jar file of the main plugin class determines whether the plugin is modularized (there could and will be dependent jars which have already been modularized, but that does not imply the plugin should be loaded as a module). So determining that means scanning every jar file for the classname in the plugin info. It seems especially silly to do this at runtime when we know this at build time.

rjernst:
@mark-vieira Given that explanation, do you still have concerns about doing this at build time?

rjernst:
@mark-vieira I reimplemented the properties file generation, using a dedicated task as we discussed. The new task now properly is up to date (even when java source has changed, since it only looks for module-info.java).

I tested this locally, with and without a module-info, but I guess we should probably have a real test for this task now? I could use some guidance on where to put that.

I also wasn't sure where to put the validation we currently do that certain properties on the extension are set, I just did them in a doFirst for now.

breskeby:
@rjernst I think having an integration test coverage for this added to `PluginBuildPluginFuncTest` in build-tools should be enough.

build-tools/src/main/java/org/elasticsearch/gradle/plugin/PluginBuildPlugin.java
```diff
@@ -114,64 +116,41 @@ private static void configureDependencies(final Project project) {
...
+
+                if (extension.getDescription() == null) {
+                    throw new InvalidUserDataException("description is a required setting for esplugin");
+                }
+
+                if (extension.getType().equals(PluginType.BOOTSTRAP) == false && extension.getClassname() == null) {
+                    throw new InvalidUserDataException("classname is a required setting for esplugin");
+                }
+            });*/
+            task.getPluginName().set(providerFactory.provider(extension::getName));
```
breskeby, line 136: We could make this cleaner by finally porting the plugin extension to use  lazy properties. But we should do this in a dedicated PR. I've created https://github.com/elastic/elasticsearch/issues/86057 to track this

breskeby, line 136: As said in a general comment, some dedicated test coverage for module handling in PluginBuildPluginFuncTest would be good.

build-tools/src/main/java/org/elasticsearch/gradle/plugin/PluginBuildPlugin.java
```diff
@@ -114,64 +116,41 @@ private static void configureDependencies(final Project project) {
...
-            map.put("hasNativeController", extension.isHasNativeController());
-            map.put("requiresKeystore", extension.isRequiresKeystore());
-            map.put("type", extension.getType().toString());
-            map.put("javaOpts", extension.getJavaOpts());
-            map.put("licensed", extension.isLicensed());
-
-            copy.expand(map);
-            copy.getInputs().properties(map);
+        final var buildProperties = project.getTasks().register("pluginProperties", GeneratePluginPropertiesTask.class, task -> {
+            /*task.doFirst(t -> {
```
breskeby, line 123: putting this in a doFirst block changes the behaviour here in a way that we fail later than before. If we keep the check in the configuration block we would fail when the task is materialized which is usually during task graph calculation. IMO we should just keep that behaviour and fail as early as we reliable can to shorten the feedback cycle for the user.

rjernst: If I can do those checks at that time, why do I need any of the lazy evaluation. Shouldn’t the entire extension be fully configured then?

mark-vieira: Are these explicit checks even  necessary? If these are `@Input`s on the task it should fail if they are `null`.

build-tools/src/main/java/org/elasticsearch/gradle/plugin/GeneratePluginPropertiesTask.java
```diff
@@ -0,0 +1,146 @@
...
+        props.put("licensed", getIsLicensed().get());
+        props.put("modulename", findModuleName());
+
+        SimpleTemplateEngine engine = new SimpleTemplateEngine();
+        Path outputFile = getOutputFile().get().getAsFile().toPath();
+        Files.createDirectories(outputFile.getParent());
+        try (
+            var inputStream = GeneratePluginPropertiesTask.class.getResourceAsStream("/" + PROPERTIES_FILENAME);
+            var reader = new BufferedReader(new InputStreamReader(inputStream));
+            var writer = Files.newBufferedWriter(outputFile, StandardCharsets.UTF_8)
```
mark-vieira, line 113: We should be creating this writer in a try-with-resources. As it is we are never closing this writer.

build-tools/src/main/java/org/elasticsearch/gradle/plugin/GeneratePluginPropertiesTask.java
```diff
@@ -0,0 +1,146 @@
...
+    @TaskAction
+    public void generatePropertiesFile() throws IOException {
+        Map<String, Object> props = new HashMap<>();
+        props.put("name", getPluginName().get());
+        props.put("description", getPluginDescription().get());
+        props.put("version", getPluginVersion().get());
+        props.put("elasticsearchVersion", getElasticsearchVersion().get());
+        props.put("javaVersion", getJavaVersion().get());
+        PluginType pluginType = getPluginType().get();
+        props.put("classname", pluginType.equals(PluginType.BOOTSTRAP) ? "" : getClassname().get());
```
mark-vieira, line 98: FWIW, this could be simplified using `getPluginType().map()`.

build-tools/src/main/java/org/elasticsearch/gradle/plugin/GeneratePluginPropertiesTask.java
mark-vieira: Same probably goes for the reader.

mark-vieira (approved):
Couple of comments, otherwise this is a big improvement. Thanks, Ryan.

build-tools/src/main/java/org/elasticsearch/gradle/plugin/GeneratePluginPropertiesTask.java
rjernst: These are in a try-with-resources already?

mark-vieira: I'm a moron that cannot differentiate a `(` from a `{`. Disregard.

build-tools/src/main/java/org/elasticsearch/gradle/plugin/PluginBuildPlugin.java
rjernst: I think it may be confusing getting the error message there, since a developer does not set this on the task, they set it on the extension.

rjernst:
@breskeby @mark-vieira I pushed a couple small tests. I also switched to trying to read the module-info.class instead of source. This is less error prone, since we then know the module info is well formed. However, the test I have for it then fails because compilation fails. It seems we don't have the dependencies the pluginbuildplugin adds, ie server.

```
* What went wrong:
Execution failed for task ':compileJava'.
> Could not resolve all files for configuration ':compileClasspath'.
   > Could not find org.elasticsearch:elasticsearch:8.3.0-SNAPSHOT.
     Required by:
         project :
   > Could not find org.locationtech.spatial4j:spatial4j:0.7.
     Required by:
         project :
   > Could not find org.locationtech.jts:jts-core:1.15.0.
     Required by:
         project :
   > Could not find org.apache.logging.log4j:log4j-api:2.17.1.
     Required by:
         project :
   > Could not find org.apache.logging.log4j:log4j-core:2.17.1.
     Required by:
         project :
   > Could not find net.java.dev.jna:jna:5.10.0.
     Required by:
         project :
```

I also realized from looking at this that jts and spatial4j can be removed, those were moved to a legacy-geo module a while ago. I'll open a separate PR for that next week.

rjernst:
Ok, I fixed the test to work by clearing the compile deps, as we do in other tests like it. I think this PR is ready.

Issue #77505, opened 2021-09-09
Provide a bulk API for creating ingest assets


joshdover:
When Fleet installs Elasticsearch ingest assets (index and component templates, ingest pipelines, ILM policies, etc.) for a package, we're currently bottlenecked by queueing behavior on cluster state updates as observed in this issue: https://github.com/elastic/kibana/issues/110500#issuecomment-911505775

This is causing some package installs to take upwards of 30s. This is a problem for Fleet, Kibana, and Elastic Agent for two primary reasons:
1. We need the ability to upgrade packages on Kibana upgrades to keep some ingest assets in sync with the rest of the Stack (eg. assets used by APM Server or Elastic Agents themselves for monitoring).
2. We also likely will want the ability to automatically downgrade packages and reinstall older version of assets when there was an issue with a Kibana upgrade that requires a rollback to the previous Kibana version. This would require that we re-write all ingest assets in Elasticsearch to be sure they're compatible with the older Kibana version.

For both of these use cases, if this process is slow, Kibana upgrades and rollbacks will be too slow and possibly time out depending on the configuration of the orchestration layer.

When executing Fleet's setup process which installs the `system` package, we're seeing cluster state updates take ~150ms each on a single node cluster running on the same machine as Kibana. See the node stats results taken here before and after the setup process: [node_stats.zip](https://github.com/elastic/elasticsearch/files/7137391/node_stats.zip), [es_logs.zip](https://github.com/elastic/elasticsearch/files/7137792/es_logs.zip)

@DaveCTurner mentioned that one way we could optimize this is by providing a bulk API to batch these cluster state updates in a single write.

DaveCTurner:
There's a question about why cluster state updates take ~150ms in these tests, and that falls under the `:Distributed/Cluster coordination` label, but the question about a bulk API for installing templates/pipelines/ILM policies etc is the domain of the data management team so I'm moving this over there.

dakrone:
@joshdover out of curiosity, do you have an idea of the number of items Fleet usually would want to do in a single request? 10s? 100s? 1000s?

dakrone:
Also, rather than batching cluster state updates, perhaps it would be better to teach Elasticsearch the concept of a package, with a set of templates, policies, pipelines, and metadata, and then make adding or removing a package an atomic operation from a cluster-state perspective.

jakelandis:
> perhaps it would be better to teach Elasticsearch the concept of a package

[...]

PR #86017, opened 2022-04-19
Batch execute template and pipeline cluster state operations

dakrone:
This commit changes the cluster state operations for templates (legacy, component, and composable) as well as ingest pipelines to be bulk executed. This means that they can be processed much faster when creating/updating many simultaneously.

Relates to #77505


server/src/main/java/org/elasticsearch/cluster/metadata/MetadataIndexTemplateService.java
```diff
@@ -1651,13 +1641,6 @@ private void validate(String name, @Nullable Settings settings, List<String> ind
         }
     }
 
-    public interface PutListener {
```
martijnvg, line 1654: 👍 to see these listeners being replaced by `ActionListener`

martijnvg (commented):
The PR looks good to me.

I do wonder about testing. 
Each type of operation is tested (e.g. putting a pipeline) on the unit level and integration level. I think batching implementation is tested as well in ClusterStateService is tested as well. So I think we're good here, right?

dakrone:
> So I think we're good here, right?

I think so too, I think adding a test for this would add a ton of complexity.

martijnvg:
> I think adding a test specifically for this would add a ton of complexity.

Yes, I had a similar feeling that these tests require a lot of work without adding much benefit.

martijnvg (approved):
LGTM 👍

DaveCTurner:
For future reference you can use `ClusterStateTaskExecutorUtils` to simplify the job of writing tests that execute batches of cluster state update tasks. I don't think it's that necessary here, the batching is pretty straightforward, but e.g. in #86446 I'm proposing a change to the `IngestServiceTests` which involves running the executor code too. The tests still only ever execute singleton batches, but hopefully it's obvious how to extend them to execute multiple tasks in a single batch too.

PR #86020, opened 2022-04-19
[Test] Remove calls to Authentication constructor

ywangd:
This PR is another step towards locking down how Authentication object
can be instantiated: It should be created using dedicated convenient
methods instead of constructors.

Production usage of constructors are mostly removed. But lots of test
code still uses them. This PR replaces one form of the usage with the
newly introduced test helper and removes the corresponding constructor.

Relates: #85590
Relates: #85905


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
```diff
@@ -72,10 +72,6 @@ public class Authentication implements ToXContentObject {
     private final AuthenticationType type;
     private final Map<String, Object> metadata; // authentication contains metadata, includes api_key details (including api_key metadata)
 
-    public Authentication(User user, RealmRef authenticatedBy, RealmRef lookedUpBy) {
-        this(user, authenticatedBy, lookedUpBy, Version.CURRENT, AuthenticationType.REALM, Collections.emptyMap());
-    }
```
ywangd, line 77: This constructor is now removed. The next PR will make the other constructor private.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/authc/AuthenticationTestHelper.java
```diff
@@ -338,7 +338,7 @@ public Authentication build(boolean runAsIfNotAlready) {
                         );
                     }
                     case TOKEN -> {
-                        if (isServiceAccount) {
+                        if (isServiceAccount != null && isServiceAccount) {
```
ywangd, line 341: A bug fix for the test helper itself.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/authc/AuthenticationTests.java
```diff
@@ -43,14 +43,23 @@ public class AuthenticationTests extends ESTestCase {
...
         final RealmRef authenticatedBy = new RealmRef("auth_by", "auth_by_type", "node");
         final RealmRef lookedUpBy = new RealmRef("lookup_by", "lookup_by_type", "node");
-        final Authentication authentication = new Authentication(new User("user"), authenticatedBy, lookedUpBy);
+        final Authentication authentication = AuthenticationTestHelper.builder()
+            .user(new User("not-user"))
+            .realmRef(authenticatedBy)
+            .runAs()
+            .user(new User("user"))
+            .realmRef(lookedUpBy)
+            .build();
```
ywangd, line 52: The authentication object was created incorrectly: The user does not have a run-as but the authentication has a non-null lookup realm. So the new code fixed it by adding an authenticating user as I believe the intention here was to have run-as. 

The similar issue existed in a few other places and they have been fixed similarly.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/authc/AuthenticationTests.java
```diff
@@ -209,38 +218,22 @@ public void testRunAsAccessResourceOf() {
     }
 
     public void testIsServiceAccount() {
```
ywangd, line 220: Service account does not support run-as nor run-by another user. So the old test is not really correct. It is fixed accordingly.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/async/AsyncTaskServiceTests.java
```diff
@@ -88,10 +89,16 @@ public String getFeatureDescription() {
     }
 
     public void testEnsuredAuthenticatedUserIsSame() throws IOException {
-        Authentication original = new Authentication(new User("test", "role"), new Authentication.RealmRef("realm", "file", "node"), null);
+        Authentication original = AuthenticationTestHelper.builder()
+            .user(new User("test", "role"))
+            .realmRef(new Authentication.RealmRef("realm", "file", "node"))
+            .build(false);
```
ywangd, line 95: The `false` argument here is to make sure the randomized Authentication is not run-as because that is how the Authentication is created in the current code (`null` for lookup realm). I suspect most of the tests would work with a run-as Authentication. But I chose to be conservative to reduce the complexity of this PR.

albertzaharovits (approved):
LGTM

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
ywangd: After a second thought, I restored this constructor for this PR because it is tagged as `>test` and this is technically production code. I will raise a separate PR _after_ all test usages are addressed to remove this constructor and change the other to private.

PR #86023, opened 2022-04-20
Remove hppc from percentiles agg

rjernst:
The percentiles bucket agg uses an hppc arraylist of doubles to store
the parsed percent values. This is a very small list and does not need
to be a native array. This commit changes to using a standard ArrayList.

relates #84735


iverase (approved):
LGTM

PR #86024, opened 2022-04-20
Remove hppc from http server transport

rjernst:
The http server transport uses an hppc integer set to find the first
open port. This commit changes it to use a standard HashSet.

relates #84735


grcevski (approved):
LGTM!

PR #86025, opened 2022-04-20
Remove hppc type leak from ImmutableOpenIntMap.Builder

rjernst:
The builder for ImmutableOpenMap was changed to not inherit from hppc
types, but the builder from ImmutableOpenIntMap was missed. This commit
removes the base hppc interface from the builder and removes unused
methods, and converts one place that was using the leaked types.

relates #84735


grcevski (approved):
LGTM!

PR #86029, opened 2022-04-20
[7.17] Apply precommit checks for non productive projects (#85533)

breskeby:
 - Apply precommit checks for non productive projects (#85533)


breskeby:

PR #85538, opened 2022-03-31
[Tests] Removing AwaitsFix for closed issue in CCSDuelIT

cbuescher:
The referenced issue #40005 is marked as closed, so we should either be able to remove the @AwaitsFix notation and be able to run the test or we should create another issue in case the test still fails.

PR #86042, opened 2022-04-20
Fix AdaptingAggregator toString method

cbuescher:
Most aggregations inheriting from AggregatorBase implement 'toString()' by
printing the name of the aggregation, but AdaptingAggregator and subclasses
don't, so they will print Object#toString(). In some tests where we e.g. compare
response json outputs for equality this can cause errors because two separate
instances will have different string representations. This changes fixes that by
also using the name of the aggregation here.

Relates to #85538


cbuescher:
See https://github.com/elastic/elasticsearch/pull/85538#issuecomment-1084434280 for an example how this can cause a problem in our tests.

nik9000 (approved):

cbuescher:
@nik thanks for the review

PR #86047, opened 2022-04-20
Avoid circuit breaker trips in shutdown node actions

### LINKED ISSUE #84847: Node shutdown cluster state task improvements

rjernst:
When adding or deleting node shutdown metadata, it is important that as
few things go wrong as possible. Since these actions do not do anything
with memory, there is no reason they should trip the circuit breaker,
but they might accidentally if there is pressure on the system. This
commit declares the rest and transport actions as not capable of
tripping the circuit breaker.

closes #84847

### DIFF
--- a/x-pack/plugin/shutdown/src/main/java/org/elasticsearch/xpack/shutdown/RestDeleteShutdownNodeAction.java
+++ b/x-pack/plugin/shutdown/src/main/java/org/elasticsearch/xpack/shutdown/RestDeleteShutdownNodeAction.java
@@ -27,6 +27,11 @@ public class RestDeleteShutdownNodeAction extends BaseRestHandler {
     }
 
     @Override
+    public boolean canTripCircuitBreaker() {
+        return false;
+    }
+
+    @Override
     protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) {
         String nodeId = request.param("nodeId");
         return channel -> client.execute(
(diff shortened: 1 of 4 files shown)


DaveCTurner (approved):
LGTM

pgomulka (approved):
LGTM

PR #86051, opened 2022-04-21
Upgrade jackson to 2.13.2

rjernst:
Most of the Jackson uses, eg in x-content and azure, have already been
upgraded. This commit upgrades the rest of the uses. Note that it does
not yet upgrade the aws sdk, this should also be done on its own.


nik9000 (approved):

mushao999:
hi @nik9000, I found that your PR : https://github.com/FasterXML/jackson-core/pull/729 is not released in 2.13.2 and may be releaed in 2.14. So we may have to upgrade jackson again after jackson 2.14 released to fix some of our awaitsfix for that PR.

PR #86052, opened 2022-04-21
Fix cli opts for windows cli script

rjernst:
The elasticsearch-cli script now uses CLI_JAVA_OPTS to control its jvm
options. However, the windows script missed this and was still using
ES_JAVA_OPTS. This commit changes it to use CLI_JAVA_OPTS.

relates #85758


williamrandolph (approved):
LGTM!

PR #86056, opened 2022-04-21
Run periodic example plugins build with included build

breskeby:
Fixes failing example builds ci jobs that cannot resolve build-tools


mark-vieira:
This is intentional. The purpose of this periodic job is to resolve these as _external_ dependencies, thus ensuring that we aren't accidentally introducing an Elasticsearch dependency that isn't published to a public repository. The side effect is that they fail for a few runs until the first successful unified snapshot build after a version bump.

breskeby:
@mark-vieira understood. makes sense

PR #86066, opened 2022-04-21
Added documentation on GeoJSON format for points and geo-points

craigtaverner:
Recent work added support for GeoJSON to the following field types:

* `geo-point` https://github.com/elastic/elasticsearch/pull/85120 (merged into 8.2.0)
* `point` https://github.com/elastic/elasticsearch/pull/85442 (targeting 8.3.0)

This PR provides documentation for these. Since this is based on the branch for cartesian point, it includes the commit from that branch. If that PR is merged first, then this PR will only contain the documentation commit. In either case the intention is to review only the documentation commit in this PR.

### DIFF
--- a/docs/painless/painless-guide/painless-execute-script.asciidoc
+++ b/docs/painless/painless-guide/painless-execute-script.asciidoc
@@ -77,8 +77,9 @@ sorted list of `double` values. See
 <<painless-runtime-double,double_field context>>.
 
 `geo_point_field`::
-The context for {ref}/geo-point.html[`geo-point` fields]. `emit` takes a
-`geo-point` value and the script returns coordinates for the geo point. See
+The context for {ref}/geo-point.html[`geo-point` fields]. `emit` takes two double
+parameters, the latitude and longitude values, and the script returns an object in
+GeoJSON format containing the coordinates for the geo point. See
 <<painless-runtime-geo,geo_point_field context>>.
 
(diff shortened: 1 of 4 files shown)


docs/painless/painless-guide/painless-execute-script.asciidoc
```diff
@@ -77,8 +77,9 @@ sorted list of `double` values. See
 <<painless-runtime-double,double_field context>>.
 
 `geo_point_field`::
-The context for {ref}/geo-point.html[`geo-point` fields]. `emit` takes a
-`geo-point` value and the script returns coordinates for the geo point. See
+The context for {ref}/geo-point.html[`geo-point` fields]. `emit` takes two double
```
craigtaverner, line 80: @iverase Could you verify this change. I think the old description seemed quite incorrect.

iverase: I had a look and yes the explanation of the emit is correct, thanks! What it is not clear to me what we refer here to what the script returns. This is currently a bug as it should behave as an indexed geo_point which is not the case.

iverase: Ok, this geo_point context is different to the out put of the fields API. lThe change looks good to me then.

iverase (approved):
LGTM

PR #86069, opened 2022-04-21
Some cleanup around the transport thread hot path

original-brownbear:
Make this code a little easier to read (both for humans and the JIT compiler)
by moving cold paths into their own methods and splitting up some huge methods.


original-brownbear:
closing this has become a little outdated

PR #86072, opened 2022-04-21
Remove hppc from deletion policy

rjernst:
The CombinedDeletionPolicy keeps ref counts for each index snapshot
using an hppc primitive map. This commit converts it to use a standard
HashMap.

relates https://github.com/elastic/elasticsearch/issues/84735


server/src/main/java/org/elasticsearch/index/engine/CombinedDeletionPolicy.java
```diff
@@ -164,7 +163,7 @@ synchronized boolean releaseCommit(final IndexCommit snapshotCommit) {
                 + "], releasing commit ["
                 + releasingCommit
                 + "]";
-        final int refCount = snapshottedCommits.addTo(releasingCommit, -1); // release refCount
+        final int refCount = snapshottedCommits.merge(releasingCommit, -1, Integer::sum); // release refCount
```
original-brownbear, line 166: NIT: I think we could even do 

```java
final Integer refCount = snapshottedCommits.compute(releaseCommit, (key, count) -> {
   if (count == 1) {
      return null;
   }
  return count - 1;
})
```

and adjust the logic below to a null check instead of `== 0`. `merge` seems kind of (slightly) wrong here when we assume that we have an entry in the map for sure?

original-brownbear (approved):
LGTM, one suggestion on making it potentially even faster and a little simpler than it is today but not important :)

PR #86073, opened 2022-04-21
Remove hppc from LocalCheckpointTracker

rjernst:
The LocalCheckpointTracker keeps mappings between sequence number and a
bitsets, using hppc primitive maps. This commit converts these to use
standard HashMaps.

relates #84735

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/seqno/LocalCheckpointTracker.java
+++ b/server/src/main/java/org/elasticsearch/index/seqno/LocalCheckpointTracker.java
@@ -8,8 +8,8 @@
 
 package org.elasticsearch.index.seqno;
 
-import com.carrotsearch.hppc.LongObjectHashMap;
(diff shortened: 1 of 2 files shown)


original-brownbear (approved):
LGTM after some digging :) 

I looked through some profiling of this code and think this is an ok change to make.
The code here is obviously not entirely trivial performance-wise, but boxing a long and some overhead for the larger map seems like a rather trivial cost increase in a method that costs ~<0.25% of an indexing operation to begin with. (somewhat stable number from a couple of profiles I examined).
If we actually find that we want to optimize this, we could e.g. use `io.netty.util.collection.LongObjectHashMap` or so easily but I don't believe it matters enough to pull in that kind of dependency here. But there's so so many things more expensive than this path in indexing to begin with that this really shouldn't be our focus IMO.

PR #86074, opened 2022-04-21
Remove hppc from translog classes

rjernst:
The translog writer uses hppc maps for mapping sequence numbers to
bitsets, and keeping track of which sequence numbers have been flushed
or not. This commit converts these uses to Java collections.

relates #84735


rjernst:

jtibshirani:
@rjernst it looks like this overlaps with #85657.

rjernst:
Thanks @jtibshirani! I had not seen that. I will close in favor of it.

PR #86076, opened 2022-04-21
Simplify DotExpandingXContentParser creation

javanna:
DotExpandingXContentParser currently takes as constructor arguments both the subparser used to make sure that we only read the original value or object and not go further, as well as the original root parser. The latter is not needed, hence can be removed in favour of always relying on the former.


romseygeek (approved):
I definitely had a reason for passing the two through separately, but I can't remember what it was and it may have become redundant after more recent changes.  If the tests pass then LGTM.

javanna:
> I definitely had a reason for passing the two through separately 

I vaguely remember you explaining the reason, but I don't recall it and I no longer see why we would need both. I trust that tests would fail if that was the case.

PR #86077, opened 2022-04-21
Persistent health task

gmarouli:
The health API, has the requirement to be as quick and as cheap as possible. To achieve this we want to have one node responsible to keep the health state of all the node of the cluster without querying all the nodes in the spot. We do not want to burden the master with this, so we are introducing a new role, the health role. One of the health nodes is going to be selected at each moment.

The selection of the health node happens via the persistent task framework. We introduce a new persistent task and whichever node is assigned to it, is the selected health node. If the selected health node is shut down we release the task so the master can assign it to a new node is asap.

This PR is introducing the health role and the persistent task. Currently the health node does nothing.


gmarouli:
We close this feature branch and we will create a new PR to master with the new feature behind a feature switch.

PR #86078, opened 2022-04-21
Remove hppc from nested aggs

rjernst:
The nested and reverse nested aggs use hppc maps and lists for mapping buckets to
ordinals. This commit converts these to use HashMaps and ArrayLists.

relates #84735

### DIFF
--- a/server/src/main/java/org/elasticsearch/search/aggregations/bucket/nested/NestedAggregator.java
+++ b/server/src/main/java/org/elasticsearch/search/aggregations/bucket/nested/NestedAggregator.java
@@ -7,8 +7,6 @@
  */
 package org.elasticsearch.search.aggregations.bucket.nested;
 
-import com.carrotsearch.hppc.LongArrayList;
-
 import org.apache.lucene.index.IndexReaderContext;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.ReaderUtil;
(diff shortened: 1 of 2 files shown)


martijnvg (approved):
I left one question. Otherwise LGTM 👍

server/src/main/java/org/elasticsearch/search/aggregations/bucket/nested/NestedAggregator.java
```diff
@@ -193,11 +194,14 @@ void processBufferedChildBuckets() throws IOException {
 
             for (; childDocId < currentParentDoc; childDocId = childDocs.nextDoc()) {
                 cachedScorer.doc = childDocId;
-                final long[] buffer = bucketBuffer.buffer;
-                final int size = bucketBuffer.size();
-                for (int i = 0; i < size; i++) {
-                    collectBucket(sub, childDocId, buffer[i]);
-                }
+                final int docId = childDocId;
+                bucketBuffer.stream().mapToLong(Long::longValue).forEach(bucket -> {
```
martijnvg, line 198: Maybe use a regular for loop here? I notice that we have been avoiding to use `steam()` in hot code paths. I think we should do this here too? But Maybe I'm wrong here.

rjernst: I had done the stream because I was trying to avoid materializing a primitive array. But I realized now I could just iterate over the boxed values and let java do the auto unboxing on the call to collectBucket.

PR #86079, opened 2022-04-21
[TEST] FIPS compliant hashing algo bwc test setup

### LINKED ISSUE #85640: [CI] Verify version constants failing on pre-8.0 clusters with FIPS enabled
This looks to have been failing for awhile. https://gradle-enterprise.elastic.co/s/2zxpfgsfn4s5k/failure

```
:qa:verify-version-constants:v7.17.3#integTest FAILED
Exec output and error:
| Output for ./bin/elasticsearch-users:
| ERROR: Only PBKDF2 is allowed for password hashing in a FIPS 140 JVM. Please set the appropriate value for [ xpack.security.authc.password_hashing.algorithm ] setting.
```

It's possible here this is an issue with _any_ BWC tests against 7.17 clusters as it seems we've changed some stuff related to FIPS between now and then?

n1v0lg:
Integration and bwc tests use `UsersTool` to setup test users, when
required. Under the hood, `UsersTool` relies on a password hash
function. When running in FIPS mode the function must be
FIPS-compliant, e.g. `pbkdf2_stretch`, otherwise `UsersTool` fails. For
v8.x, a FIPS compliant hash function is chosen by default but for older
versions this is not the case. This leads to intermittent bwc test
failures. This PR tweaks the build for the failing test to use a
FIPS-compliant hash function.

Closes #85640


qa/verify-version-constants/build.gradle
```diff
@@ -18,6 +18,12 @@ BuildParams.bwcVersions.withIndexCompatible { bwcVersion, baseName ->
   def baseCluster = testClusters.register(baseName) {
       version = bwcVersion.toString()
       setting 'xpack.security.enabled', 'true'
+      if (BuildParams.inFipsJvm && bwcVersion.before('8.0.0')) {
```
n1v0lg, line 21: This is the only test that has been failing with this particular issue. As such, I'm fixing only the build for this particular test (since it's been bugging people for some time now).

I will follow this up with another PR that addresses with more generally, probably here: https://github.com/elastic/elasticsearch/blob/master/build-tools-internal/src/main/groovy/elasticsearch.fips.gradle#L68

martijnvg (approved):
LGTM, pending successful build.

ywangd:
@n1v0lg Do you mind backport this to 8.2? It is also failing there https://gradle-enterprise.elastic.co/s/x2o3qarn6kg5w

n1v0lg:
@yang-wei I think we're actually better off forward-porting https://github.com/elastic/elasticsearch/pull/66841 since that fixes the issue more generally

PR #86081, opened 2022-04-21
Move embedded loader and provider locator to core

ChrisHegarty:
Move embedded loader and provider locator to core so that the mechanism can be used by areas other than x-content. The usage then becomes a small recipe from the call-site. Over on the modules branch we're using this loading mechanism to load log4j (rather than loading it in the system class loader).


ChrisHegarty:

libs/core/src/main/java/org/elasticsearch/core/internal/provider/ProviderLocator.java
```diff
@@ -0,0 +1,73 @@
...
+ * Otherwise, the provider implementation will be loaded as a non-module.
+ *
+ * @param <T> the provider type
+ */
+public final class ProviderLocator<T> implements Supplier<T> {
+
+    private final String providerName;
+    private final Class<T> providerType;
+    private final String providerModuleName;
+    private final Set<String> missingModules;
```
rjernst, line 37: I assume this is held over from the module branch, but not applicable to master right now?

libs/x-content/src/main/java/org/elasticsearch/xcontent/internal/XContentProviderLocator.java
```diff
@@ -0,0 +1,37 @@
...
+    static final String PROVIDER_MODULE_NAME = "org.elasticsearch.xcontent.impl";
+
+    /**
+     * Returns the provider instance.
+     */
+    public static final XContentProvider INSTANCE = provider();
+
+    private static XContentProvider provider() {
+        Module m = XContentProviderLocator.class.getModule();
+        if (m.isNamed() && m.getDescriptor().uses().stream().anyMatch(XContentProvider.class.getName()::equals) == false) {
```
rjernst, line 32: Can this be moved into the locator so we don't have to duplicate it? Shouldn't the module of the locator and the provider be the same, so we can just use `providerType.getModule()` there?

libs/x-content/src/main/java/org/elasticsearch/xcontent/internal/XContentProviderLocator.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.elasticsearch.core.internal.provider.ProviderLocator;
+import org.elasticsearch.xcontent.spi.XContentProvider;
+
+import java.util.ServiceConfigurationError;
+
+/**
+ * A provider locator for finding the {@link XContentProvider}.
+ */
+public final class XContentProviderLocator {
```
rjernst, line 19: Do we need a separate class, or could this simply be inside XContentProvider? It's odd that we have "INSTANCE" which is not this class itself, but XContentProvider, and there we *also* have instance which just pulls from this class.

rjernst (approved):
Looks good, just a couple questions/suggestions.

libs/core/src/main/java/org/elasticsearch/core/internal/provider/ProviderLocator.java
ChrisHegarty: Yes, I'll remove the modules stuff from this PR.

libs/x-content/src/main/java/org/elasticsearch/xcontent/internal/XContentProviderLocator.java
ChrisHegarty: Yeah, good suggestion. I'll drop it from this PR, but address it in the modules branch.

libs/x-content/src/main/java/org/elasticsearch/xcontent/internal/XContentProviderLocator.java
ChrisHegarty: I folder this into an inner class in the XContentProvider interface ( since we just want a single static instance ).

PR #86083, opened 2022-04-21
Use declared constant (rather than resource lookup)

ChrisHegarty:
Use the public static final constant `org.apache.lucene.analysis.icu.NORMALIZER`, rather than poking around inside lucene resources - the `Normalizer2` instance is equivalent. It would appear that this code, doing the resource lookup, predates the lucene public field.


ChrisHegarty:

rjernst (approved):
LGTM

ChrisHegarty:

PR #86089, opened 2022-04-21
Fix NPE when using object field as match field for enrich policy

### LINKED ISSUE #86058: Null pointer exception executing a range enrich policy with an "object" field
### Elasticsearch Version

8.1.1

### Installed Plugins

_No response_

### Java Version

_bundled_

###  OS Version

Darwin Kernel Version 21.4.0

### Problem Description

Under certain conditions, executing a range enrich policy on a field with wrong datatype gives a NPE instead of a more comprehensible error.

### Steps to Reproduce

```
PUT enrich_index
{
  "mappings": {
    "properties": {
      "object": {
        "properties": {
          "field": {
            "type": "keyword"
          }
        }
      }
    }
  }
}

PUT /_enrich/policy/enrich_policy
{
  "range": {
    "indices": "enrich_index",
    "match_field": "object",
    "enrich_fields": ["enrich_field"]
  }
}

POST /_enrich/policy/enrich_policy/_execute
```

Executing the policy returns:
```json
{
  "error" : {
    "root_cause" : [
      {
        "type" : "null_pointer_exception",
        "reason" : "Cannot invoke \"String.hashCode()\" because \"<local7>\" is null"
      }
    ],
    "type" : "null_pointer_exception",
    "reason" : "Cannot invoke \"String.hashCode()\" because \"<local7>\" is null"
  },
  "status" : 500
}
```

Interestingly enough, if the enrich index were created like this (without subfield in `object`):
```
PUT enrich_index
{
  "mappings": {
    "properties": {
      "object": {
        "properties": {
        }
      }
    }
  }
}
```
 

Executing the policy would return a much better error:
```json
{
  "error" : {
    "root_cause" : [
      {
        "type" : "exception",
        "reason" : "Field 'object' has type [object] which doesn't appear to be a range type"
      }
    ],
    "type" : "exception",
    "reason" : "Field 'object' has type [object] which doesn't appear to be a range type"
  },
  "status" : 500
}
```

### Logs (if relevant)

_No response_

martijnvg:
Fix NPE when specifying a field of type object as match field for an enrich policy of type range.

This NPE could only occur if type field wasn't specified in the mapping, in which case it is of type object.
Addressed this by returning a better to understand error, since a field of type object is unsupported as match field.

Closes #86058


jbaiera (approved):
LGTM

martijnvg:
@elasticmachine update branch

PR #86098, opened 2022-04-22
Remove old optional deps from plugin build

rjernst:
The main elasticsearch jar used to have optional deps. This was because
of the high level rest client, where users did not want to pull in these
additional heavyweight deps that are not actually exposed by the Java apis.
These optional deps no longer exist since we no longer ship the HLRC.
When building plugins, these deps were "upgraded" to normal compile
classpath deps, since they were in fact available at runtime for
plugins. Since they are no longer optional, they are included
transitively by the dependency on the elasticsearch jar. This commit
removes the addition of these deps in the plugin build.


Nick19880515 (commented):
request approval

pugnascotia (approved):

ChrisHegarty (approved):
LGTM

PR #86100, opened 2022-04-22
Remove xmlbeans jar hell exception

ChrisHegarty:
This is a minor cleanup for an exception to the jar hell check that is
no longer necessary. The xmlbeans issue has been resolved, and ES
depends upon an xmlbeans version that contains that fix.


rjernst (approved):
LGTM

PR #86103, opened 2022-04-22
Forward port MDP deprecation info api

rjernst:
Multiple data paths was deprecated in 7.13. At the time, the deprecation
was only added to the deprecation logs. Later, it was also added to the
deprecation info api, but by that time, MDP had already been removed
from master. When MDP was added back to master, the deprecation log
messages were added back, but the info api messages were not since they
never existed in master.

This commit forward ports the deprecation info messages for MDP to 8.x.

relates #85695


pgomulka (approved):
LGTM

PR #86107, opened 2022-04-23
Permit removal of archived index settings

DaveCTurner:
Today you cannot remove archived index settings by applying a setting
update `{"archived.*":null}` because `IndexSettings#same` incorrectly
treats such an update as a no-op. This commit fixes that.


grcevski (commented):
This looks great! Would it be possible to add unit tests in IndexSettingsTests for the same method? Just to make sure we return the correct result in a combination of archived and current settings. We do seem to have coverage for the same method through calling `updateIndexMetadata`, but not directly and not for archived settings.

grcevski (approved):
LGTM!

PR #86114, opened 2022-04-24
Faster string writes by saving stream flushes

original-brownbear:
We serialize the chars to the same thread-local byte buffer that we serialize
longer vint to as well. This is done to save the overhead of writing out
multiple bytes one by one to more expensive streams (like the translog checksum one).
This PR takes the concept one step further and combines serializing the vint at the beginning
of the string write and the character serialization into a single buffer.
In some ad-hoc benchmarking of writing shorter strings to network buffers this is a 25%+
speedup through effectively making the vint write free and inline fully. Prior to this change
the performance of the string write is a lot less predictable since writing the initial vint
prefix may or may not cause allocating new write buffers (as does writing the actual bytes)
so things inline in strange ways depending on usage and string sizes + offsets.

Also, use the same "trick" to significantly speed up writing optional strings which we do quite a bit.

### DIFF
--- a/server/src/main/java/org/elasticsearch/common/io/stream/StreamOutput.java
+++ b/server/src/main/java/org/elasticsearch/common/io/stream/StreamOutput.java
@@ -229,13 +229,26 @@ public abstract class StreamOutput extends OutputStream {
             return;
         }
         byte[] buffer = scratch.get();
-        int index = 0;
+        int index = putMultiByteVInt(buffer, i, 0);
+        writeBytes(buffer, 0, index);
+    }
+
+    private static int putVInt(byte[] buffer, int i, int off) {
+        if (Integer.numberOfLeadingZeros(i) >= 25) {
+            buffer[off] = (byte) i;
+            return 1;
+        }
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM!

original-brownbear:
Thanks Nikola!

server/src/main/java/org/elasticsearch/common/io/stream/StreamOutput.java
```diff
@@ -388,10 +403,21 @@ public void writeText(Text text) throws IOException {
...
+        writeString(str, scratch.get(), 0);
+    }
+
+    /**
+     * Write string as well as possibly the beginning of the given {@code buffer}. The given {@code buffer} will also be used when encoding
+     * the given string.
+     *
+     * @param str string to write
+     * @param buffer buffer that may hold some bytes to write
+     * @param off how many bytes in {code buffer} to write
```
dakrone, line 415: This is confusing to me; isn't this the `offset` at which writing happens, not how many bytes to write? This string is confusing.

PR #86120, opened 2022-04-25
Refactor bounded geogrid aggregations

iverase:
GeoGrid aggregations accepts an optional bounding box to limit the bins that the aggregation can generate to the bins intersecting the provided bounding box. The implementation for Geohash and GeoTile is actually duplicated, one for GeoGrid aggregations over points and anther one for the aggregation over shapes.

This PR refactors those implementations into a `GeoHashBoundedPredicate` and a `GeoTileBoundedPredicate`.


server/src/main/java/org/elasticsearch/search/aggregations/bucket/geogrid/GeoTileBoundedPredicate.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /** The top bound on geotile coordinates */
+    public int maxY() {
+        return maxY;
+    }
+
+    /** Check if the provided tile intersects with the provided bounds. It is assumed that the precision
+     * for the given x/y is the same as the provided in the constructor.*/
+    public boolean validTile(int x, int y) {
```
craigtaverner, line 82: This function is identical to the one that takes `z`, with `precision==z`, so would it not be simpler to have only one function? The cost of the additional `<<` and `+` is probably not high, I would guess.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/geogrid/GeoTileBoundedPredicate.java
```diff
@@ -0,0 +1,118 @@
...
+            } else {
+                return rightX > x && leftX <= x;
+            }
+        }
+        return false;
+    }
+
+    /** Check if the provided tile at the provided level intersects with the provided bounds. The provided precision must be
+     * lower or equal to the precision provided in the constructor.
+     * */
```
craigtaverner, line 95:
```suggestion
     */
```

server/src/main/java/org/elasticsearch/search/aggregations/bucket/geogrid/GeoHashCellIdSource.java
```diff
@@ -73,50 +72,25 @@ protected int advanceValue(org.elasticsearch.common.geo.GeoPoint target, int val
...
-                if (crossesDateline) {
-                    return bbox.left() < maxX || bbox.right() > minX;
-                } else {
-                    return bbox.left() < maxX && bbox.right() > minX;
-                }
-            }
-            return false;
-        }
-
         private boolean validPoint(double x, double y) {
```
craigtaverner, line 94: It feels like this could be moved to `GeoHashBoundedPredicate`, just as the similar method from `GeoTileCellIdSource` was moved.

craigtaverner (approved):
Looks good. Some minor suggestions.

.idea/checkstyle-idea.xml
```diff
@@ -3,17 +3,17 @@
   <component name="CheckStyle-IDEA">
     <option name="configuration">
       <map>
-        <entry key="active-configuration" value="PROJECT_RELATIVE:$PROJECT_DIR$/checkstyle_ide.xml:Elasticsearch" />
-        <entry key="checkstyle-version" value="9.2" />
+        <entry key="active-configuration" value="PROJECT_RELATIVE:/Users/ivera/forks/elasticsearch/checkstyle_ide.xml:Elasticsearch" />
```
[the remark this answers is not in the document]
craigtaverner: I see this is already fixed.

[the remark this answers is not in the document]
iverase: yes, I noticed before.

server/src/test/java/org/elasticsearch/search/aggregations/bucket/geogrid/GeoTileBoundedPredicateTests.java
```diff
@@ -0,0 +1,65 @@
...
+        );
+        long hash = GeoTileUtils.longEncodeTiles(precision, x, y);
+        GeoTileBoundedPredicate predicate = new GeoTileBoundedPredicate(precision, bbox);
+
+        int minX = GeoTileUtils.getXTile(bbox.left(), tiles);
+        int maxX = GeoTileUtils.getXTile(bbox.right(), tiles);
+        int minY = GeoTileUtils.getYTile(bbox.bottom(), tiles);
+        int maxY = GeoTileUtils.getYTile(bbox.top(), tiles);
+
+        assertPredicates(hash, predicate, minX, minY, precision);
```
craigtaverner, line 35: I ran this test and found all four corners tested here result in the predicate returning `false`? Somehow I was expecting at least one of the corners to return true based on the min/max comment in the predicate.

craigtaverner: In the other tests `GeoHashBoundedPredicateTests.testValidTile` the min/min test (left, bottom) returned true for the predicate. I expected the same here, but it returns false. This seems wrong, surely?

server/src/test/java/org/elasticsearch/search/aggregations/bucket/geogrid/GeoTileBoundedPredicateTests.java
```diff
@@ -0,0 +1,65 @@
...
+        assertPredicates(hash, predicate, maxX, minY, precision);
+        assertPredicates(hash, predicate, minX, maxY, precision);
+        assertPredicates(hash, predicate, maxX, maxY, precision);
+
+        for (int i = 0; i < 1000; i++) {
+            assertPredicates(hash, predicate, randomIntBetween(0, (1 << i) - 1), randomIntBetween(0, (1 << i) - 1), precision);
```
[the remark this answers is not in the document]
craigtaverner: When I did the same for the GeoHash version of this test, sometimes all 10000 were false, but once 37/1000 were true. I guess it all depends on the random numbers. I think these randomized tests are very useful for covering scope, but do feel like we should have a couple of specific non-random tests to assert on known cases. To anchor the behaviour.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/geogrid/GeoTileBoundedPredicate.java
iverase: Ok, I removed the other function.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/geogrid/GeoHashCellIdSource.java
iverase: I prefer not because `GeoHashBoundedPredicate` should only work on hash. This method is just an optimisation oof this class. If the function return true then we know it is a vlid hash, else we will still need to use the predicate.

server/src/test/java/org/elasticsearch/search/aggregations/bucket/geogrid/GeoTileBoundedPredicateTests.java
iverase: GeoTile transformation are not stable, in particular on the latitude because of the trigonometry that we need to use to compute the Y value (maths are base on https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#Java). So it might happen that `tileToLat` gives you a lat, and `getYTile` using that lat gives you a different tile. It is not important here at the moment but we should keep this in mind.

server/src/test/java/org/elasticsearch/search/aggregations/bucket/geogrid/GeoTileBoundedPredicateTests.java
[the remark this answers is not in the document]
iverase: I added non random test.

PR #86122, opened 2022-04-25
Small additions to the register-repo docs

DaveCTurner:
1. Adds a note that you can restore older snapshots (to recover from a
failed upgrade) even after newer snapshots were taken.

2. Copies the note about incompatible S3 repo implementations to the top
level to avoid misunderstandings.


docs/reference/snapshot-restore/register-repository.asciidoc
```diff
@@ -127,9 +130,27 @@ Other repository types are available through official plugins:
 
 * {plugins}/repository-hdfs.html[Hadoop Distributed File System (HDFS)]
 
-You can also use alternative implementations of these repository types, such as
-MinIO, as long as they're compatible. To verify a repository's compatibility,
-see <<snapshots-repository-verification>>.
+You can also use alternative storage implementations with these repository
+types, as long as the alternative implementation is fully compatible. For
```
henningandersen, line 134: I think "fully" is maybe too much. At least we do not claim that minio is "fully" compatible. I'd say compatible is fine, we do elaborate on it in the following section?

docs/reference/snapshot-restore/register-repository.asciidoc
```diff
@@ -127,9 +130,27 @@ Other repository types are available through official plugins:
...
+instance, https://minio.io[MinIO] provides an alternative implementation of the
```
henningandersen, line 135: "a compatible alternative implementation"? Otherwise the "so" later does not refer to compatibility, just some implementation.

arteam (approved):
LGTM! I appreciate the reference to the `repo-analysis-api` API for performing basic tests on third-party S3 API vendors.

henningandersen (approved):
LGTM.

docs/reference/snapshot-restore/register-repository.asciidoc
DaveCTurner: I'd rather err on the side of caution here. Everyone calls their implementation "compatible" even if it's only a subset of endpoints, but few are willing to claim _full_ compatibility. Experience shows it's best not to leave room for debate in this area.

docs/reference/snapshot-restore/register-repository.asciidoc
DaveCTurner: I see what you mean. Instead, I changed the `so` for an `and` to make the sentence more neutral about the level of compatibility offered by Minio.

PR #86132, opened 2022-04-25
Add debug information to ReactiveReason about assigned and unassigned shards

### LINKED ISSUE #85243: Provide more information about autoscaling policy decisions to improve debugging
In some contexts is unclear why an autoscaling policy decides to request
more resources, i.e. if there's an unassigned shard that remains unassigned
even after adding more resources. It would be help to have more information
around the decisions taken by the deciders so it is easier to understand and
track them.

arteam:
We would like to see not only the amount of assigned and unassigned shards, but also the ids of the respective shards.

Fixes #85243


arteam:
@elasticmachine update branch

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -101,9 +102,21 @@ public List<DiscoveryNodeRole> roles() {
 
     @Override
     public AutoscalingDeciderResult scale(Settings configuration, AutoscalingDeciderContext context) {
+        Set<ShardId> unassignedShardIds = StreamSupport.stream(context.state().getRoutingNodes().unassigned().spliterator(), false)
+            .map(ShardRouting::shardId)
+            .collect(Collectors.toSet());
+        Set<ShardId> assignedShardIds = context.state()
```
henningandersen, line 108: The meaning of "assigned" in this decider is somewhat different from this. It means the assigned shards that contributes to the autoscaling decision. It is shards that are assigned to a node, where `canRemain() == NO` and there is no other node where it can be assigned. Additionally, either the reason it cannot reamin or it cannot be allocated must be a single disk threshold reason (i.e., not for other reasons).

I think you will need to get this information out together with the amount of disk space needed, i.e., to change `storagePreventsRemainOrMove` to return both the number of the list of shards.

henningandersen (commented):
Did an early review, left a comment.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -101,14 +102,25 @@ public List<DiscoveryNodeRole> roles() {
...
         AutoscalingCapacity autoscalingCapacity = context.currentCapacity();
         if (autoscalingCapacity == null || autoscalingCapacity.total().storage() == null) {
-            return new AutoscalingDeciderResult(null, new ReactiveReason("current capacity not available", -1, -1));
+            return new AutoscalingDeciderResult(
+                null,
+                new ReactiveReason(
+                    "current capacity not available",
+                    -1,
+                    -1,
+                    unassignedBytesUnassignedShards.unassignedShards(),
```
fcofdez, line 116: I'm wondering if it would be better to return an empty list in this case as we're signalling that we cannot make an autoscaling decision in the current state.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -722,6 +798,8 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
             builder.field("reason", reason);
             builder.field("unassigned", unassigned);
             builder.field("assigned", assigned);
+            builder.field("unassignedShardIds", unassignedShardIds);
```
fcofdez, line 801: We use underscores in our json representation: `s/unassignedShardIds/unassigned_shard_ids/`

fcofdez: I'm not sure if `unassigned` is a right term for this API as it is possible that a shard that should move from hot to warm remains assigned to a hot node, but for the purposes of this decider we consider it as `unassigned` to trigger the tier creation.  But we use these terms already, so we can leave as it is.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
```diff
@@ -381,16 +381,40 @@ private static void verifyScale(ClusterState state, long expectedDifference, Str
...
                 equalTo(expectedDifference)
             );
-            assertThat(result.reason().summary(), equalTo(reason));
+            assertThat(resultReason.summary(), equalTo(reason));
         } else {
             assertThat(result.requiredCapacity(), is(nullValue()));
-            assertThat(result.reason().summary(), equalTo("current capacity not available"));
+            assertThat(resultReason.summary(), equalTo("current capacity not available"));
         }
+        assertThat(
```
fcofdez, line 396: Can we add a more explicit test case? i.e. when a shard moves from hot to warm and that triggers the autoscaling decision.

fcofdez (commented):
Looks in the good direction, I've left a few comments.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -722,6 +798,8 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
...
+            builder.field("assignedShardIds", assignedShardIds);
```
fcofdez, line 802: I wonder if we want to show all the assigned shards that might get large and I think it doesn't provide much information about the autoscaling decision.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
fcofdez: I was thinking that it might make sense to include some information around the allocation decision for these shards so it is easier to debug issues around that. The downside is that responses can get too large.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
arteam: Hm, I don't have a strong opinion on that. I think it makes sense to have this field for the consistency sake, but I can also see the side of the argument of avoid bloat in the autoscaling decision response.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
arteam: I guess it's a bit tricky, because it's a generic `verify` method which gets invoked from multiple tests. I will see if I can create a separate test.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
arteam: Done!

arteam: I also added a test for the JSON representation of `ReactiveReason`

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveReasonTests.java
```diff
@@ -0,0 +1,54 @@
...
+import org.elasticsearch.xcontent.json.JsonXContent;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class ReactiveReasonTests extends ESTestCase {
+
+    public void testXContent() throws IOException {
```
fcofdez, line 26: We have a base class to test XContent parsing, see `AbstractXContentTestCase` it takes care of all the boilerplate 👍

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
fcofdez: I think it's covered in `testStoragePreventsAllocation`. Can we verify there that we see the "unassigned" shards there?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
henningandersen: > I was thinking that it might make sense to include some information around the allocation decision 

I agree to finding at least one shard to get more info for. But I'd like to keep that separate to this PR, this has value on it's own.

henningandersen (commented):
Left more comments, but this direction looks good.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
henningandersen: It is not really all the assigned shards. It is only those that cannot move but needs to move AND where either the reason they cannot remain is only storage OR the reason they cannot move somewhere else is storage. So it is much like the unassigned list and should not grow large except if the cluster has issues.

We should still limit the list we show though to the first X shards.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -722,6 +793,8 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
             builder.field("reason", reason);
             builder.field("unassigned", unassigned);
             builder.field("assigned", assigned);
+            builder.field("unassigned_shard_ids", unassignedShardIds);
+            builder.field("assigned_shard_ids", assignedShardIds);
```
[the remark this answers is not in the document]
henningandersen: On second thought, perhaps we should have a count field and a truncated list of shard-ids instead, to use a json array for the list rather than a comma separated string.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveReasonTests.java
arteam: I don't think we need to use `AbstractXContentTestCase` here, because we only testing serialization

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveReasonTests.java
```diff
@@ -0,0 +1,77 @@
...
+            )
+        ) {
+            Map<String, Object> map = parser.map();
+            assertEquals(reason, map.get("reason"));
+            assertEquals(unassigned, map.get("unassigned"));
+            assertEquals(assigned, map.get("assigned"));
+            assertEquals(unassignedShardIds.stream().map(ShardId::toString).collect(Collectors.toList()), map.get("unassigned_shard_ids"));
+            List<String> xContentAssignedShardIds = (List<String>) map.get("assigned_shard_ids");
+            assertEquals(
+                assignedShardIds.stream().map(ShardId::toString).limit(512).collect(Collectors.toList()),
```
fcofdez, line 60: Why `.limit(512)`?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -714,6 +775,10 @@ public void writeTo(StreamOutput out) throws IOException {
             out.writeString(reason);
             out.writeLong(unassigned);
             out.writeLong(assigned);
+            if (out.getVersion().onOrAfter(Version.V_8_3_0)) {
```
fcofdez, line 778: nit: can you extract that version to a constant?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -678,17 +710,38 @@ public static class ReactiveReason implements AutoscalingDeciderResult.Reason {
...
             this.assigned = assigned;
+            this.unassignedShardIds = unassignedShardIds;
+            this.assignedShardIds = assignedShardIds;
         }
 
         public ReactiveReason(StreamInput in) throws IOException {
             this.reason = in.readString();
             this.unassigned = in.readLong();
             this.assigned = in.readLong();
+            if (in.getVersion().onOrAfter(Version.V_8_3_0)) {
```
fcofdez, line 738: nit: can you extract that version to a constant?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -166,6 +185,8 @@ static Optional<String> singleNoDecision(Decision decision, Predicate<Decision>
         }
     }
 
+    record BytesShardIds(long bytes, SortedSet<ShardId> shardIds) {}
```
fcofdez, line 188: I find this name a bit confusing, maybe we can use something else like `ShardsSize` or something more descriptive? otherwise we might want to use a `Tuple`?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -118,7 +124,20 @@ public AutoscalingDeciderResult scale(Settings configuration, AutoscalingDecider
             .total(autoscalingCapacity.total().storage().getBytes() + unassignedBytes + assignedBytes, null)
             .node(maxShardSize, null)
             .build();
-        return new AutoscalingDeciderResult(requiredCapacity, new ReactiveReason(message, unassignedBytes, assignedBytes));
+        return new AutoscalingDeciderResult(
+            requiredCapacity,
+            new ReactiveReason(
+                message,
+                unassignedBytes,
```
fcofdez, line 131: nit: I wonder if it makes sense to keep the total size and shard ids together as they are related?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -118,7 +124,20 @@ public AutoscalingDeciderResult scale(Settings configuration, AutoscalingDecider
...
+                assignedBytes,
+                unassignedBytesUnassignedShards.shardIds(),
```
fcofdez, line 133: I think we mentioned that it would be nice to provide some the allocation debug info for few of the unassigned shards, so it is easier to understand why a shard is unassigned? wdyt @henningandersen?

fcofdez (commented):
Direction looks good, I left a few comments. 👍

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
henningandersen: We discussed that [here](https://github.com/elastic/elasticsearch/pull/86132#discussion_r866652526) and I suggested to keep that to a separate PR. Do you think it need to be incorporated into this PR?

fcofdez: Oh sorry, I missed that. I think it's fine to tackle that in a separate PR.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
arteam: Added via 

```java
assertThat(resultReason.unassignedShardIds(), equalTo(decider.allocationState(context).storagePreventsAllocation().shardIds()))
```

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
arteam: Done!

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
arteam: Done!

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -714,14 +777,21 @@ public void writeTo(StreamOutput out) throws IOException {
...
 
         @Override
         public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
             builder.startObject();
             builder.field("reason", reason);
             builder.field("unassigned", unassigned);
+            builder.field("unassigned_shard_ids", unassignedShardIds);
             builder.field("assigned", assigned);
+            builder.field("assigned_shard_ids", assignedShardIds.stream().limit(MAX_ASSIGNED_SHARD_IDS).collect(Collectors.toList()));
+            builder.field("amount_of_assigned_shards", assignedShardIds.size());
```
henningandersen, line 794: Can we name this `assigned_shards` and `assigned_shards_count` instead?

```suggestion
            builder.field("assigned_shards", assignedShardIds.stream().limit(MAX_ASSIGNED_SHARD_IDS).collect(Collectors.toList()));
            builder.field("assigned_shards_count", assignedShardIds.size());
```

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -714,14 +777,21 @@ public void writeTo(StreamOutput out) throws IOException {
...
+                out.writeCollection(assignedShardIds);
+            }
         }
 
         @Override
         public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
             builder.startObject();
             builder.field("reason", reason);
             builder.field("unassigned", unassigned);
+            builder.field("unassigned_shard_ids", unassignedShardIds);
```
henningandersen, line 791: I think we should limit this too and also present a count.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ProactiveStorageDeciderServiceTests.java
```diff
@@ -277,7 +277,7 @@ public void testForecast() {
             List<Index> addedIndices = forecastDataStream.getIndices().subList(indices, forecastDataStream.getIndices().size());
             long actualTotal = totalSize(addedIndices, forecastRoutingTable, forecast.info());
 
-            // three round downs -> max 3 bytes lower and never above.
+            // three round downs -> max 3 sizeInBytes lower and never above.
```
henningandersen, line 280: This looks like a rename gone wrong, revert?

henningandersen (commented):
Did not complete this yet, but wanted to relay the comments I have so far.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -214,16 +233,19 @@ public static class AllocationState {
...
-                .sum();
+                .toList();
+            return new ShardsSize(
+                unassignedShards.stream().mapToLong(this::sizeOf).sum(),
+                unassignedShards.stream().map(ShardRouting::shardId).collect(Collectors.toCollection(TreeSet::new))
+            );
         }
 
-        public long storagePreventsRemainOrMove() {
+        ShardsSize storagePreventsRemainOrMove() {
```
henningandersen, line 248: Likewise for removing public here.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -255,7 +277,15 @@ && canAllocate(shard, allocation) == false) {
...
                 .sum();
 
-            return unallocatableBytes + unmovableBytes;
+            return new ShardsSize(
+                unallocatableBytes + unmovableBytes,
+                Stream.concat(
+                    unmovableShards.stream(),
+                    candidates.stream()
+                        .filter(Predicate.not(unmovableShards::contains))
+                        .filter(s -> cannotAllocateDueToStorage(s, allocation))
```
henningandersen, line 286: This bit is somewhat expensive and we just did it above. I'd prefer to collect the unallocatable shards in a list to avoid the repeated calculation.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -675,20 +705,45 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
     }
 
     public static class ReactiveReason implements AutoscalingDeciderResult.Reason {
+
+        static final int MAX_AMOUNT_OF_SHARDS = 512;
+        private static final Version SHARD_IDS_OUTPUT_VERSION = Version.V_8_3_0;
```
henningandersen, line 710: I think this should be updated to 8_4 now.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveReasonTests.java
```diff
@@ -0,0 +1,91 @@
...
+public class ReactiveReasonTests extends ESTestCase {
+
+    @SuppressWarnings("unchecked")
+    public void testXContent() throws IOException {
+        String reason = randomAlphaOfLength(10);
+        long unassigned = randomNonNegativeLong();
+        long assigned = randomNonNegativeLong();
+        String indexUUID = UUIDs.randomBase64UUID();
+        SortedSet<ShardId> unassignedShardIds = new TreeSet<>(randomUnique(() -> new ShardId("index", indexUUID, randomInt(1000)), 600));
+        SortedSet<ShardId> assignedShardIds = new TreeSet<>(randomUnique(() -> new ShardId("index", indexUUID, randomInt(1000)), 600));
```
henningandersen, line 38: Also randomize the index name?

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
```diff
@@ -405,6 +449,18 @@ private long numAllocatableSubjectShards() {
             .count();
     }
 
+    private SortedSet<ShardId> allocatableShards() {
```
henningandersen, line 452: I think this method and `numAllocatableSubjectShards` can be collapsed into one, returning the `ShardsSize` to use.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderReasonWireSerializationTests.java
```diff
@@ -47,6 +81,12 @@ protected ReactiveStorageDeciderService.ReactiveReason mutateInstance(ReactiveSt
 
     @Override
     protected ReactiveStorageDeciderService.ReactiveReason createTestInstance() {
-        return new ReactiveStorageDeciderService.ReactiveReason(randomAlphaOfLength(10), randomNonNegativeLong(), randomNonNegativeLong());
+        return new ReactiveStorageDeciderService.ReactiveReason(
+            randomAlphaOfLength(10),
+            randomNonNegativeLong(),
+            new TreeSet<>(randomUnique(() -> new ShardId("index", UUIDs.randomBase64UUID(), randomInt(5)), 8)),
```
henningandersen, line 87: Also randomize the index name?

henningandersen (commented):
Left a few final smaller comments, otherwise this looks good.

fcofdez (approved):
LGTM, thanks for the iterations @arteam. This change will be really useful to debug autoscaling decisions!

henningandersen (approved):
LGTM with 2 more changes. No need for another review round.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
```diff
@@ -381,28 +420,37 @@ private static void verifyScale(ClusterState state, long expectedDifference, Str
...
+            assertThat(resultReason.summary(), equalTo("current capacity not available"));
+            assertThat(resultReason.unassignedShardIds(), equalTo(Set.of()));
+            assertThat(resultReason.assignedShardIds(), equalTo(Set.of()));
         }
     }
 
-    private long numAllocatableSubjectShards() {
+    record AllocatableShards(long numOfShards, SortedSet<ShardId> shardIds) {}
+
+    private AllocatableShards allocatableShards() {
```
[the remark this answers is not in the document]
arteam: I deliberately collected the data to a list, because I believed have seen some test failing when the amount was different from the size of unique list of shards. Let me take a closer look and just use a `SortedSet` if it's not the case.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
```diff
@@ -381,28 +420,37 @@ private static void verifyScale(ClusterState state, long expectedDifference, Str
...
+        List<ShardId> allocatableShards = StreamSupport.stream(state.getRoutingNodes().unassigned().spliterator(), false)
             .filter(shard -> subjectShards.contains(shard.shardId()))
             .filter(
-                shard -> StreamSupport.stream(allocation.routingNodes().spliterator(), false)
-                    .anyMatch(node -> deciders.canAllocate(shard, node, allocation) != Decision.NO)
+                shard -> allocation.routingNodes().stream().anyMatch(node -> deciders.canAllocate(shard, node, allocation) != Decision.NO)
             )
-            .count();
+            .map(ShardRouting::shardId)
+            .toList();
```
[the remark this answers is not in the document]
fcofdez: @arteam can we collect directly to a `TreeSet` as Henning suggests?

fcofdez (approved):
LGTM 👍 

There's still a couple of Henning's suggestions that should be tackled, but otherwise it looks good.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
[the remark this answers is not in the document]
arteam: No, unfortunately, we can't. `unallocatableBytes` get calculated based on `List<ShardRouting>` which may contain duplicate shard ids. If we use `TreeSet` here, the test starts regularly failinf.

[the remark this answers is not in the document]
henningandersen: Ah, that makes sense now, thanks for checking that. Perhaps you can add a comment that the list may contain duplicates due to shard copies (replicas)?

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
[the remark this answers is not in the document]
henningandersen: Resolving this, since the size of the set is different from the number of shards when there are multiple shard copies.

arteam:
Thank you both!

PR #82085, opened 2021-12-26
Remove NIO Transport Plugin

original-brownbear:
Removes the `transport-nio` plugin completely.

PR #86133, opened 2022-04-25
Flatten Netty4 http pipeline handlers into a single handler

original-brownbear:
This sets up the implementation of chunked REST response encoding
that will be able to send large responses without serializing them
in full on heap upfront.
Having all the steps of sending a response in one place together with
the pipelining logic allows adding another response type that serializes
on the transport threads alongside `Netty4HttpResponse` in a follow-up
that will only require changes to `Netty4PipeliningHandler`.
Also, this change in isolation saves loads of indirection and now (that the NIO
transport is gone) redundant abstractions, making the path of sending a REST
response all around easier to follow.

relates #82085 

NOTE: I did copy a lot of the moved around code outright without too much renaming. While the resulting code could be made nicer by renaming and refactoring the way the methods are set up, I figured for this step it's far easier to review things if I copy as much as I can outright. The same reasoning went into the rather awkward way in which I made `Netty4HttpPipeliningHandlerTests` work for now.  A follow-up aligning it better with `Netty4HttpServerPipeliningTests` should be able to clean this up further.


modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -84,4 +173,49 @@ public void close(ChannelHandlerContext ctx, ChannelPromise promise) {
...
+
+        Netty4HttpChannel channel = ctx.channel().attr(Netty4HttpServerTransport.HTTP_CHANNEL_KEY).get();
+        if (cause instanceof Error) {
+            serverTransport.onException(channel, new Exception(cause));
+        } else {
+            serverTransport.onException(channel, (Exception) cause);
+        }
+    }
+
+    private List<Tuple<Netty4HttpResponse, ChannelPromise>> write(final Netty4HttpResponse response, ChannelPromise listener) {
```
henningandersen, line 190: Let us rename parameter to promise.

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpServerPipeliningTests.java
```diff
@@ -115,98 +112,44 @@ class CustomNettyHttpServerTransport extends Netty4HttpServerTransport {
...
-                );
-                response.addHeader("content-length", Integer.toString(buffer.readableBytes()));
-
-                final boolean slow = uri.matches("/slow/\\d+");
-                if (slow) {
-                    try {
-                        Thread.sleep(scaledRandomIntBetween(500, 1000));
-                    } catch (InterruptedException e) {
-                        throw new RuntimeException(e);
+        public void incomingRequest(HttpRequest httpRequest, HttpChannel httpChannel) {
```
henningandersen, line 122: This change puzzled me a bit in that you tried to keep most other things unchanged, but here went for a more radical change. I see the need for changes (using deleted types), but for my education perhaps you can explain the need for the radical change of the test.

henningandersen (commented):
This looks fine to me, left a couple of comments, I'd like to understand the test change more fully.

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpServerPipeliningTests.java
original-brownbear: Sure :) See this change: https://github.com/elastic/elasticsearch/pull/86133/files#diff-5fb205d075055657d087e3445ff349bcfed01425594e604f7bf7b3d3a6dfefa9L325

we had the `pipelining` handler and then the `handler` as separate channel handlers before, now they're a single handler. So previously the test could just replace the "handler" with `PossiblySlowUpstreamHandler` and still test the logic in the unchanged "pipelining" handler. Now that both handlers are one, this wasn't possible any more.
So since I couldn't replace part of the handler, I left the pipeline as is and overrode the http server transport to catch the requests a little further down the stack.
Or put differently, we previously replaced this code https://github.com/elastic/elasticsearch/pull/86133/files#diff-a3368fd4145592d305c99a7c2b0a6c21c9bd6d45c828515aed1f228b69af2a79L29 in the test, now we run it in this test as well and overrode the code it calls.

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
original-brownbear: Sure done :)

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpServerPipeliningTests.java
henningandersen: Thanks for the explanation!

henningandersen (approved):
LGTM.

Please do follow-up with some of the code cleanup PRs that you mentioned once this is merge.

original-brownbear:
Thanks Henning, will do!

PR #86134, opened 2022-04-25
Initialize active realms without logging a message (#81380).

### LINKED ISSUE #81380: Boot message possible confusion: reporting trial license
## Description

During Elasticsearch boot, users get the following message:

> license mode is [trial], currently licensed security realms are [reserved/reserved,file/default_file,native/default_native]

Even if it's technically true (and necessary), it may be confusing for end users since they may expect `[basic]` instead.

## Proposal

Avoid the message reporting the trial license, since the "correct" license level is written again later in the boot process.

slobodanadamovic:
This commit tries to remove confusing initial log message about the license and active realms.
When `Realms` is created, `activeRealms` property is initialized with default (`TRIAL`) license state and message is logged.
Few seconds later an event is triggered which updates license to the actual user's license.
This causes a second message to be logged. This introduces a confusion when the actual license (e.g. `BASIC`) is different than the initial license (`TRIAL`).

Closes #81380

### DIFF
--- a/docs/changelog/86134.yaml
+++ b/docs/changelog/86134.yaml
@@ -1,0 +1,6 @@
+pr: 86134
+summary: Initialize active realms without logging a message
+area: License
+type: enhancement
+issues:
+ - 81380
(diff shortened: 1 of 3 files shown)


albertzaharovits (approved):
LGTM
I've spent some good amount of time on this one, in order to make sure that the `Realms#recomputeActiveRealms` is indeed invoked in all the cases:

* new cluster with/without a license
* existing cluster
* a node that cannot form a cluster

Unfortunately this sort of thing is hard to write a test for. Also a missing log line on the enabled authn realms is a relatively mild issue. Therefore I'm content with this PR.

PR #83884, opened 2022-02-14
Upgrade Azure SDK

fcofdez:

PR #86135, opened 2022-04-25
Upgrade Azure SDK to 12.16.0

fcofdez:
In this version jackson was upgraded to the version `2.13.2.1` we were using
before. See https://github.com/Azure/azure-sdk-for-java/releases/tag/azure-core_1.27.0

Without the upgrade, the third party SAS tests were failing, once upgraded the tests pass again,
I think that we were hitting some incompatibility with the jackson version that we were 
providing. 

Relates #83884


arteam (approved):
LGTM!

fcofdez:
Thanks Artem!

PR #82842, opened 2022-01-20
Remove version_type "internal" from rest api specs

russcam:
This commit removes the version_type `"internal"` from the rest api specs.
Support for internal versioning for optimistic concurrency control was
removed in elastic/elasticsearch#38254.

Since this removes values from rest api specs, it will affect any artifact that is generated from specs, for example, clients.

Other observations
- `VersionType` enum still has an `INTERNAL` value, which might be needed for bwc?
- `SetProcessorTests` randomly selects a version type which includes `"internal"` - this should probably also be removed.
- `IngestServiceTests` randomly selects a version type which includes `"internal"` and `"external_gt"` - these should probably also be removed.

PR #86139, opened 2022-04-25
Remove version_type "internal" from rest api specs

arteam:
It's a rebased version of #82842 created in order to pass the new PR checks around the changelogs

> This commit removes the version_type "internal" from the rest api specs.
Support for internal versioning for optimistic concurrency control was
removed in elastic/elasticsearch#3825.


arteam:

PR #86142, opened 2022-04-25
Remove HLRC security package

joegallo:

There's just one class left in this package, and it's pretty easy to replace all the uses of it.


dakrone (approved):
LGTM

PR #86146, opened 2022-04-25
Default hashing algo setting refactor

n1v0lg:
Small refactor that removes code duplication in constructing the
default hashing algorithm setting between password and token setting. 

Also removes obsolete test cluster configuration tweak introduced in
#86079. This configuration fix is superseded by #86171.


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/XPackSettings.java
```diff
@@ -183,42 +184,24 @@ private XPackSettings() {
...
-                        e
-                    );
-                }
-            }
-        },
-        Property.NodeScope
+        }
     );
 
-    // TODO: This setting of hashing algorithm can share code with the one for password when pbkdf2_stretch is the default for both
```
n1v0lg, line 216: Spotted this TODO while working on https://github.com/elastic/elasticsearch/pull/86079 and figured I might as well!

ywangd (approved):
LGTM

I suggest we tag this as `>refactoring` instead of `>non-issue`.

PR #86148, opened 2022-04-25
FIPS compliant hashing algo for v7.x bwc tests

### LINKED ISSUE #86144: Security enabled BWC tests fail in FIPS mode for v7.x
### Problem Description

When xpack security and FIPS mode is enabled for a BWC test, the test requires that a FIPS-compliant password hashing algorithm is set via the `xpack.security.authc.password_hashing.algorithm` setting. Without a compliant algorithm, the cluster fails to start and the test fails. This happens automatically via a default setting value for v8.x clusters but does **not** happen for v7.x clusters. 

As such, v7.x BWC tests that have security enabled fail when run in FIPS mode.

The original test failure where this surfaced was `VerifyVersionConstantsIT` (see #85640) for which we put in a temporary fix to remove the test triage noise (#86079).

There are several BWC tests left that will still fail however and we should put a general fix in place. Based on an internal discussion with @es-security, we decided not to backport a default settings change to v7.x clusters. We therefore need to fix this issue within our test setup.

### Steps to Reproduce

`./gradlew -Dtests.fips.enabled=true :x-pack:plugin:shutdown:qa:rolling-upgrade:v7.17.4#oldClusterTest --tests org.elasticsearch.upgrades.NodeShutdownUpgradeIT.testShutdown`

will fail with

```
Exec output and error:
| Output for ./bin/elasticsearch-users:
| ERROR: Only PBKDF2 is allowed for password hashing in a FIPS 140 JVM. Please set the appropriate value for [ xpack.security.authc.password_hashing.algorithm ] setting.
```

n1v0lg:
Closes https://github.com/elastic/elasticsearch/issues/86144


n1v0lg:
Closing in favor of forward port: https://github.com/elastic/elasticsearch/pull/86171

n1v0lg:
Closing in favor of forward port: https://github.com/elastic/elasticsearch/pull/86171

PR #86149, opened 2022-04-25
Dedicated API for unbatched master tasks

DaveCTurner:
Today when submitting a cluster state update task the caller can
indicate that it is not to be included in a batch by using an executor
obtained from `ClusterStateTaskExecutor#unbatched()`. This is kind of
weird: the caller must not re-use the executor, and also the master
service has no practical way to handle this special case any
differently.

This commit introduces a dedicated API for unbatched tasks, opening the
door to some future simplifications.


DaveCTurner:
Note to reviewers: this is pretty much exactly as boring as it looks. Maybe look at the changes to `MasterService` and `ClusterService` but there's no other meaningful content here.

DaveCTurner:

idegtiarenko (approved):

PR #86150, opened 2022-04-25
Remove hppc from top hits agg

rjernst:
The top hits aggregator uses hppc for keeping track of leaf bucket
collectors. This commit converts it to use HashMap.

relates #84735


nik9000 (commented):
Could you use `LongObjectPagedHashMap` here instead? It's known not to allocate. I know `top_hits` isn't the fastest thing, but we may as well use the tools we build for aggs in aggs.

rjernst:
Ok, switched to `LongObjectPagedHashMap`

rjernst:
@nik9000 CI looks happy now, any other concerns?

nik9000 (approved):

PR #86151, opened 2022-04-25
Remove hppc from terms agg

rjernst:
The terms aggs take include/excludes for terms, and for numeric fields
use an hppc set of longs. This commit converts to using a HashSet.

relates #84735

### DIFF
--- a/server/src/main/java/org/elasticsearch/search/aggregations/bucket/terms/IncludeExclude.java
+++ b/server/src/main/java/org/elasticsearch/search/aggregations/bucket/terms/IncludeExclude.java
@@ -7,9 +7,6 @@
  */
 package org.elasticsearch.search.aggregations.bucket.terms;
 
-import com.carrotsearch.hppc.LongHashSet;
-import com.carrotsearch.hppc.LongSet;
-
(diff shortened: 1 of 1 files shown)


nik9000 (commented):
Do we know if the contains test allocates? This is deep in a high performance path where we really can't allocate. The size of the objects themselves doesn't matter - it's the `accept` method that calls `contains`.

Another option that we *know* doesn't allocate is `LongHash` - it's a bit silly here, but if we're allocating then it's probably worth it.

If we're not allocating we probably should have a comments in `accept` about it.

rjernst:
`contains` doesn't directly allocate, but the autoboxing that will occur could, if the value is outside the range of pre-cached boxed values.

I don't have any strong opinions, other than the overall goal of moving away from hppc. I'll switch this to LongHash.

rjernst:
`LongHash` would have been difficult to use because it requires BigArrays. Instead, I've switched the value type of the set to be our own private boxed type that allows modification. This way the accept method can use a spare instance as we do in other methods here (eg spare BytesRef).

nik9000 (approved):

rjernst:

PR #86152, opened 2022-04-25
Fix Uninformative Assertion in JdbcShardFailureIT

original-brownbear:
We have a hard to understand failure in this one when trying to make a change, lets
make it tell us the concrete numbers at least.


rjernst (approved):
LGTM

Issue #86044, opened 2022-04-20
Add label features to vector tile search API response


nreese:
### Description

There are [2 labeling issues in Kibana when using Vector tiles](https://github.com/elastic/kibana/issues/130675).
1) Multiple labels for MultiPolygon and MultiLineString geometries.
    <img width="500" alt="Screen Shot 2022-04-20 at 9 32 26 AM" src="https://user-images.githubusercontent.com/373691/164268037-0d6088f3-5bd6-4e0e-9541-d551091af843.png">
2) Multiple labels for features that span tiles. For example, hexagons that span tile boundaries.
    <img width="500" alt="Screen Shot 2022-04-20 at 9 35 52 AM" src="https://user-images.githubusercontent.com/373691/164268658-dfdd6b0e-10b7-46da-a1e9-15e99044fc38.png">

The recommended solution is to provide features for labels, where the feature geometry is the label location.

To resolve these issues, I propose adding a new parameter, `withLabels`,  to `_mvt` API that when `true` would add label features to the response. Label features could be added in 2 ways
1) another layer, "label",  in the response. The "label" layer will contain Point feature for each hit and agg feature. For agg features, the Point feature will be the center of the hexagon. For hit features, the Point feature will be the centroid of the largest polygon for MultiPolygon and the centroid for the longest line for MultiLineString.
2) Include label feature in existing "hits" and "aggs" layers. Give label feature a property like "__EsIsLabelFeature" so filter expressions can be used exclude features in style rules.

Kibana is creating these centroid features today for GeoJSON sources. Kibana includes label features in GeoJSON feature collection and uses filters to exclude features in style rules (option 2 above) 

cc @jsanz @thomasneirynck

PR #86154, opened 2022-04-25
Support geo label position as runtime field

craigtaverner:
There is a need to provide sensibly calculated label positions for polygons and lines in Kibana maps (as requested in https://github.com/elastic/elasticsearch/issues/86044). A very convenient way of satisfy this need is through a runtime field that the rest API can make use of when labels are requested. This has the advantage of providing painless access to the label position as well. This PR provides the runtime field.


craigtaverner:

server/src/main/java/org/elasticsearch/script/field/GeoPointDocValuesField.java
```diff
@@ -121,6 +121,11 @@ public GeoBoundingBox getInternalBoundingBox() {
         return boundingBox;
     }
 
+    @Override
+    public GeoPoint getInternalLabelPosition() {
+        return centroid;
```
iverase, line 126: I think we should return here one of the points in the array instead of the centroid.

craigtaverner: Good point. I notice that the centroid is calculated during setMultiValue using a loop over all points. Since we are looping there anyway, perhaps we could pick a point somewhat central as the label position already there?

craigtaverner: I think I came up with a reasonable algorithm for picking which point to use, and the index of the point in the values array is selected during the setSingleValue and setMultiValue methods, so hopefully has negligable impact. I'll push something soon.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/index/mapper/GeoShapeWithDocValuesFieldMapper.java
```diff
@@ -401,6 +401,16 @@ public GeoBoundingBox getInternalBoundingBox() {
...
         }
 
+        @Override
+        public GeoPoint getInternalLabelPosition() {
+            try {
+                return value.labelPosition();
+            } catch (IOException e) {
+                // Most of the time the centroid is a good compromise, but perhaps we should know about this error anyway?
+                return centroid;
+            }
```
iverase, line 411: Not sure if we should ignore the error, maybe throw an UncheckedIOException instead?

iverase (commented):
Just have a small comment about how we handle an error. Otherwise let's get CI green and it can be approved.

iverase (approved):
LGTM

Thank you Craig!

stu-elastic (approved):
Scripting looks good.

PR #86156, opened 2022-04-25
Optional sub claim in JwtAuthenticationToken

justincr-elastic:
JWT `sub` claim is required in JwtAuthenticationToken, but optional in JwtRealm via the `claims.principal` setting.

The goal of this PR is to make the `sub` claim optional in JwtAuthenticationToken. If `sub` is absent, use the other claims to compute a substitute for the `sub` claim. The computed substitute is all of the claims, ordered to make the cache key deterministic, and filtered to exclude 7 claims (i.e. `iss`, `aud`, `jti`, `iat`, `exp`, `nbf`, `auth_time`). The exclude list is hard-coded, because JwtAuthenticationToken handling is common to all JWT realms, not per-realm.

Note, the `aud` claim is ordered to make that part of the cache key deterministic too.

JwtRealmTestCase `createJwtRealmSettings` already randomizes the `claims.principal` setting to `sub` or `<realmname>_sub`. JwtRealmTestCase `randomJwt` has been updated to check if the `claims.principal` setting is `sub` or not. If not `sub`, a value for `sub` is randomized to be included or omitted.

Example values of the token principal can be seen in trace/debug logs of `JwtRealm`, like so:
1. If `claims.principal` setting is `sub`:
```
token=[iss1_4324/iss1_4324_aud1/principal0]
```
2. If `claims.principal` setting is `realm_iss1_4324_sub`:
```
token=[iss1_4324/iss1_4324_aud1/{gr0up=superuser, realm_iss1_4324_sub=principal0}]
```

### DIFF
--- a/docs/changelog/86156.yaml
+++ b/docs/changelog/86156.yaml
@@ -1,0 +1,5 @@
+pr: 86156
+summary: Optional sub claim in `JwtAuthenticationToken`
+area: Authentication
+type: enhancement
+issues: []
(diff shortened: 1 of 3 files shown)


justincr-elastic:
Closed this PR. A different approach will be used in #86533

PR #86166, opened 2022-04-25
Add support for dots in field names for metrics usecases

### LINKED ISSUE #63530: Dotted field names that conflict with objects
Elasticsearch assumes that dots in fields names are an object separator. This means that a document such as this one:

```
{
  "metric.value.max": 42
}
```

is actually indexed as if it was formatted like below:

```
{
  "metric": {
    "value": {
      "max": 42
    }
  }
}
```

And in the mappings, this translates into two `object` fields called `metric` and `metric.value` and a `long` field called `metric.value.max`.

This proves problematic when ingesting metrics that come from external systems such as Micrometer or OpenTelemetry, as it's not rare to have both `metric.value` and `metric.value.max` as metric names:

```
{
  "metric.value": 10,
  "metric.value.max": 42
}
```

Such a document will always fail indexing because `metric.value` would need to be an `object` field because of `metric.value.max` and a `long` field at the same time, which is illegal.

Some workarounds have been developed, such as replacing dots with underscores, or adding suffixes, but this creates a bad user experience in Kibana as users are not seeing the field names that they expect.

We should look into ways to make this supported in Elasticsearch.

javanna:
This PR adds support for a new mapping parameter to the configuration of the object mapper (root as well as individual fields), that makes it possible to store metrics data where it's common to have fields with dots in their names in the following format:

```
{
  "metrics.time" : 10,
  "metrics.time.min" : 1,
  "metrics.time.max" : 500
}
```

Instead of expanding dotted paths the their corresponding object structure, objects can be configured to preserve dots in field names, in which case they can only hold leaf sub-fields and no further objects.

The mapping parameter is called `subobjects` and controls whether an object can hold other objects (defaults to true) or not. The following example shows how it can be configured in the mappings:
```
{
  "mappings" : {
    "properties" : {
      "metrics" : {
        "type" : "object", 
        "subobjects" : false
      }
    }
  }
}
```

Closes #63530


docs/reference/mapping/params/collapsed.asciidoc
```diff
@@ -0,0 +1,102 @@
...
+The described default behaviour is reasonable for most scenarios, but causes problems in certain situations
+where for instance a field `metrics.time` holds a value too, which is common when indexing metrics data.
+A document holding a value for both `metrics.time.max` and `metrics.time` gets rejected given that `time`
+would need to be a leaf field to hold a value as well as an object to hold the `max` sub-field.
+
+The `collapsed` setting, which can be applied only to the top-level mapping definition and
+to <<object,`object`>> fields, makes it possible to store metrics where field names contain dots and share
+common prefixes. From the example above, if the object container `metrics` is collapsed, it can hold values
+for both `time` and `time.max` directly without the need for any intermediate object, as dots in field
+names are preserved.
```
romseygeek, line 17: Maybe worth saying here that you can't add a document with an object structure underneath a collapsed field?

server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
```diff
@@ -445,7 +445,13 @@ private static void parseObject(final DocumentParserContext context, ObjectMappe
...
         if (objectMapper != null) {
             context.path().add(currentFieldName);
+            if (objectMapper instanceof ObjectMapper objMapper) {
+                if (objMapper.isCollapsed()) {
+                    context.path().setWithinCollapsedPath(true);
+                }
+            }
             parseObjectOrField(context, objectMapper);
+            context.path().setWithinCollapsedPath(false);
             context.path().remove();
```
romseygeek, line 455: I think we can always set within collapsed path to false on `remove()` can't we, because by definition if we're moving back up the object hierarchy then the object we're moving to can't be collapsed.  So maybe we can have just one method on ContentPath, `setCollapsed()`?

server/src/main/java/org/elasticsearch/index/mapper/DotExpandingXContentParser.java
```diff
@@ -232,7 +269,7 @@ public Token currentToken() {
     @Override
     public String currentName() throws IOException {
         if (state == State.PARSING_ORIGINAL_CONTENT) {
-            assert expandedTokens == subPaths.length * 2 - 1;
+            // assert expandedTokens == subPaths.length * 2 - 1;
```
romseygeek, line 272: If you're replacing `subPaths` when we collapse, doesn't this assertion still hold?

server/src/test/java/org/elasticsearch/index/mapper/DocumentParserTests.java
```diff
@@ -1842,6 +1842,164 @@ public void testDotsOnlyFieldNames() throws Exception {
...
+        IllegalArgumentException err = expectThrows(IllegalArgumentException.class, () -> mapper.parse(source("""
+            {
+              "metrics": {
+                "service": {
+                  "time.max" : 10
+                }
+              }
+            }
+            """)));
+        assertEquals("Object [_doc] is collapsed and does not support inner object [metrics]", err.getMessage());
```
romseygeek, line 1892: I wonder if we can detect this case and replace `Object [_doc]` with `root`, or something else?  It's a bit confusing otherwise, as there isn't a `_doc` anywhere in the mappings that the user sees.

romseygeek: It would be nice to get the full object path here as well, but from a quick look at the parsing code it doesn't look like it will be simple to do, so happy to leave both of these as a TODO.

romseygeek (commented):
This is great! I left a couple of nits and suggestions. I think it's probably worth adding the tests for objects as arrays as well, just because that's an area where we are pretty under-tested and I can see it would be very easy for bugs to slip in there.

server/src/main/java/org/elasticsearch/index/mapper/DotExpandingXContentParser.java
javanna: yes leftover, I commented it out to try a different approach that did not require replacing the array, but it turned out not to be easy at all, so I reverted that. Will reinstate the assertion.

server/src/test/java/org/elasticsearch/index/mapper/DocumentParserTests.java
javanna: yea I agree about the error message, I believe that's what we already do for enabled though. I would look at changing this as a follow-up.

Which full object path do you mean?

server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
```diff
@@ -474,7 +480,13 @@ private static void parseObjectDynamic(DocumentParserContext context, ObjectMapp
                 throwOnCreateDynamicNestedViaCopyTo(dynamicObjectMapper);
             }
             context.path().add(currentFieldName);
+            if (dynamicObjectMapper instanceof ObjectMapper objectMapper) {
+                if (objectMapper.isCollapsed()) {
+                    context.path().setCollapsed(true);
+                }
+            }
             parseObjectOrField(context, dynamicObjectMapper);
+            context.path().setCollapsed(false);
```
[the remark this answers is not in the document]
javanna: we use add also for multi-fields, so I think resetting collapsed at remove may actually cause problems.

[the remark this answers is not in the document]
javanna: That though reminds me: can you please triple check that I am calling setCollapsed in all the right places? I am afraid I am missing some...

docs/reference/mapping/params/collapsed.asciidoc
javanna: yep i thought I had that but not here I see, good point

romseygeek (commented):
Nearly there, I had one question about nested objects.

docs/changelog/86166.yaml
```diff
@@ -0,0 +1,21 @@
+pr: 86166
+summary: Add support for dots in field names for metrics usecases
+area: Mapping
+type: feature
+issues:
+ - 63530
+highlight:
+  title: Add support for dots in field names for metrics usecases
+  body: "Add support for a new mapping parameter to the configuration of\n\
```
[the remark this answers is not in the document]
javanna: not sure, I am confused too.

server/src/main/java/org/elasticsearch/index/mapper/NestedObjectMapper.java
```diff
@@ -92,7 +92,7 @@ protected static void parseNested(String name, Map<String, Object> node, NestedO
     private final Query nestedTypeFilter;
 
     NestedObjectMapper(String name, String fullPath, Map<String, Mapper> mappers, Builder builder) {
-        super(name, fullPath, builder.enabled, builder.dynamic, mappers);
+        super(name, fullPath, builder.enabled, Explicit.IMPLICIT_FALSE, builder.dynamic, mappers);
         if (builder.indexCreatedVersion.before(Version.V_8_0_0)) {
```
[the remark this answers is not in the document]
javanna: good catch thanks. I am now wondering if there are other properties that are valid for object for not for nested which we end up parsing and possibly ignoring.

romseygeek:
>  Anything specific I need to test regarding the synthetic source support that has meanwhile landed?

Can you add a test to `SourceLoaderTests` that ensures that if we have a root or intermediate object with `subobjects:false` set on it, then we get the expected source back out? There's already a test for dots in field names, you should be able to add a couple of extra cases there.

romseygeek (approved):
LGTM.  Thanks for bringing this one over the line @javanna!

ruflin:
I filed https://github.com/elastic/package-spec/issues/349 to get support for this in package-spec.

PR #86167, opened 2022-04-25
Fleet: Add start_time and minimum_execution_duration attributes to actions

michel-laterman:
### What does this PR Do?

Add new start_time and minimum_execution_duration attributes to
.fleet-actions index mapping.

### Related Issues

- https://github.com/elastic/fleet-server/issues/1380

### DIFF
--- a/x-pack/plugin/core/src/main/resources/fleet-actions.json
+++ b/x-pack/plugin/core/src/main/resources/fleet-actions.json
@@ -22,6 +22,13 @@
         "expiration": {
           "type": "date"
         },
+        "start_time": {
+          "type": "date"
+        },
+        "minimum_execution_duration": {
+          "type": "keyword",
+          "enabled": false
+        },
         "input_type": {
           "type": "keyword"
         },
--- a/x-pack/plugin/core/src/main/resources/fleet-actions.json
+++ b/x-pack/plugin/core/src/main/resources/fleet-actions.json
@@ -26,8 +26,7 @@
           "type": "date"
         },
         "minimum_execution_duration": {
-          "type": "keyword",
-          "enabled": false
+          "type": "keyword"
         },
         "input_type": {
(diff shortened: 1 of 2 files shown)


x-pack/plugin/core/src/main/resources/fleet-actions.json
```diff
@@ -22,6 +22,12 @@
         "expiration": {
           "type": "date"
         },
+        "start_time": {
+          "type": "date"
+        },
+        "minimum_execution_duration": {
+          "type": "keyword"
```
ph, line 29: @michel-laterman What is the type of this field? duration looks more like a integer to me?

ph:
requires: https://github.com/elastic/elasticsearch/pull/86167 right?

williamrandolph (approved):
This looks good from an Elasticsearch perspective. I share the concern about `minimum_excecution_duration` being a keyword and not an integer, but that is an implementation detail for the Fleet application, so it is outside the scope of what I can competently review.

x-pack/plugin/core/src/main/resources/fleet-actions.json
michel-laterman: I was intending it to be a duration string i.e. `10m`
i was looking at a more recent draft of the json-schema we use on the fleet-server: https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.7.3.1

Let me know if I should change it to an integer

ph: So this would be a `Time.Duration`, I would just use an integer value here so we are aligned with the value of timeout which I believe is in seconds?

criamico (approved):
LGTM

aleksmaus (approved):
👍

ph (approved):

PR #86173, opened 2022-04-26
[doc] Explicitly mention about node shutdown remove for cluster shrink

pgomulka:

### DIFF
--- a/docs/reference/shutdown/apis/shutdown-delete.asciidoc
+++ b/docs/reference/shutdown/apis/shutdown-delete.asciidoc
@@ -26,7 +26,8 @@ you must be an operator to use this API.
 Enables a node to resume normal operations following a
 <<put-shutdown, put shutdown>> request.
 You must explicitly clear the shutdown request when a node rejoins the cluster,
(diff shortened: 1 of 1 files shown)


docs/reference/shutdown/apis/shutdown-delete.asciidoc
```diff
@@ -26,7 +26,8 @@ you must be an operator to use this API.
 Enables a node to resume normal operations following a
 <<put-shutdown, put shutdown>> request.
 You must explicitly clear the shutdown request when a node rejoins the cluster,
-regardless of the shutdown type.
+regardless of the shutdown type. When node is permanently removed (cluster shrink)
+the node shutdown request also has to be cleared.
```
gwbrown, line 30: I agree this needs to be mentioned explicitly. What do you think of this wording?

```suggestion
or when a node has permanently left the cluster. Shutdown requests are never
removed automatically by {es}.
```

AthenaEryma (approved):
LGTM, thanks for the iteration!

PR #86177, opened 2022-04-26
Dry up waiting for global Netty threads in internal cluster tests

original-brownbear:
We had this rule duplicated around the codebase. I don't think it's necessary these days
but it provides a neat speedup in tests by not wasting the 100ms hardcoded wait interval
in the thread leak checker if the global event executer doesn't shut down quickly enough.
Also the rule does avoid a bunch of unnecessary noisy WARN logging about waiting for leaked
threads.

-> Since we now use Netty as the exclusive transport implementation we might as well use
this rule everywhere to get its benefits and dry up the code a little.


x-pack/qa/smoke-test-plugins-ssl/src/test/java/org/elasticsearch/smoketest/SmokeTestMonitoringWithSecurityIT.java
```diff
@@ -68,35 +61,6 @@ public class TestRestHighLevelClient extends RestHighLevelClient {
...
     }
 
-    /**
-     * A JUnit class level rule that runs after the AfterClass method in {@link ESIntegTestCase},
-     * which stops the cluster. After the cluster is stopped, there are a few netty threads that
-     * can linger, so we wait for them to finish otherwise these lingering threads can intermittently
-     * trigger the thread leak detector
-     */
-    @ClassRule
-    public static final ExternalResource STOP_NETTY_RESOURCE = new ExternalResource() {
```
DaveCTurner, line 78: This is an `ESRestTestCase` which doesn't derive from `ESIntegTestCase` so I think we are missing another call to `awaitGlobalNettyThreadsFinish` here.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/test/SecuritySingleNodeTestCase.java
```diff
@@ -109,35 +104,6 @@ protected void before() {
...
     };
 
-    /**
-     * A JUnit class level rule that runs after the AfterClass method in {@link ESIntegTestCase},
-     * which stops the cluster. After the cluster is stopped, there are a few netty threads that
-     * can linger, so we wait for them to finish otherwise these lingering threads can intermittently
-     * trigger the thread leak detector
-     */
-    @ClassRule
-    public static final ExternalResource STOP_NETTY_RESOURCE = new ExternalResource() {
```
DaveCTurner, line 119: This is an `ESSingleNodeTestCase` which doesn't derive from `ESIntegTestCase` so I think we are missing another call to `awaitGlobalNettyThreadsFinish` here.

DaveCTurner (changes requested):
A couple of these places aren't `ESIntegTestCase` derivatives so I think they need more work. Also should we be waiting for these threads to finish in `AbstractMultiClustersTestCase`?

x-pack/qa/smoke-test-plugins-ssl/src/test/java/org/elasticsearch/smoketest/SmokeTestMonitoringWithSecurityIT.java
original-brownbear: Ah well spotted, didn't even notice that to be honest. Though I think this check is needless here, we don't spin up any Netty things in the rest tests so this wasn't doing anything and it's fine that it goes away here I think.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/test/SecuritySingleNodeTestCase.java
original-brownbear: Fair point, I'll just add it to `ESSingleNodeTestCase` also, so we cover all of those as well since they may spin up Netty tasks here and there when using REST.

x-pack/qa/smoke-test-plugins-ssl/src/test/java/org/elasticsearch/smoketest/SmokeTestMonitoringWithSecurityIT.java
DaveCTurner: Ah yes that makes sense we don't start any nodes ourselves in these tests.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/test/SecuritySingleNodeTestCase.java
DaveCTurner: ++ even if not using REST we still bind to a transport port in these tests.

original-brownbear:
Thanks @DaveCTurner and ++ to also doing this in the multi cluster test case, added it there as well.

DaveCTurner (approved):
LGTM

original-brownbear:
Thanks David!

Issue #86181, opened 2022-04-26
[Transform] Add validation for fields whose names start with underscore ('_')


przemekwitek:
### Description

Currently, the source fields whose names start with underscore character ('_') are dropped by the `latest` transform and don't show up in the destination index. 
This was the desired behavior as we considered such fields internal. Unfortunately, there are users specifying their own fields with '_' prefix and they are surprised when they don't see them in the destination index.
We should add a validation that will:
1. check for reserved fieldnames (available in `IndicesModule::initBuiltInMetadataMappers`)
2. warn about all other fields that have a leading underscore

benwtrent:
> warn about all other fields that have a leading underscore

We should probably just write a SINGLE warning for the lifetime of the transform on the node. Otherwise we spam way too much.

PR #86183, opened 2022-04-26
[Transform] Verify in yml tests that fields with leading underscore character are filtered out by the latest transform

przemekwitek:
This PR adds assertions verifying that:
1. fields with leading underscore character are filtered out by the `latest` transform
2. search runtime fields specified in the `latest` transform config are not included in the output

Relates https://github.com/elastic/elasticsearch/issues/86181


benwtrent (approved):
It would be really cool if our documentation mentioned this as well.

PR #86184, opened 2022-04-26
Fix 'autoGeneratedTimestamp should not be set externally' error when retrying IndexRequest

### LINKED ISSUE #83927: Our internal usage of BulkProcessor can fail indexing during retry
When BulkProcessor retries failed item requests then the same `IndexRequest` instance are used in a new `BulkRequest` for a retry. This is a problem when `BulkProcessor` is used from within Elasticsearch (this is the case for deprecation index logging, ilm/slm history and more), because the bulk request transport action may have set the `autoGeneratedTimestamp` field. 

If that is the case then error such a these may appear in the logs and the `BulkProcessor` retry fails:

```
Bulk write of 1 deprecation logs failed: autoGeneratedTimestamp should not be set externally
java.lang.IllegalArgumentException: autoGeneratedTimestamp should not be set externally
	at org.elasticsearch.action.bulk.TransportBulkAction.doInternalExecute(TransportBulkAction.java:223) ~[elasticsearch-7.16.1.jar:7.16.1]
	at org.elasticsearch.action.bulk.TransportBulkAction$1.doRun(TransportBulkAction.java:192) ~[elasticsearch-7.16.1.jar:7.16.1]
	at org.elasticsearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:26) [elasticsearch-7.16.1.jar:7.16.1]
	at org.elasticsearch.action.bulk.TransportBulkAction.doExecute(TransportBulkAction.java:200) [elasticsearch-7.16.1.jar:7.16.1]
	at org.elasticsearch.action.bulk.TransportBulkAction.doExecute(TransportBulkAction.java:91) [elasticsearch-7.16.1.jar:7.16.1]
```

In the bulk transport action, for each bulk item request of the `BulkRequest` its `process()` method is invoked. Incase of `IndexRequest` when no _id has been specified an id is generated and in that case the `IndexRequest#autoGeneratedTimestamp ` field is set. If the bulk request fails at a later stage, for example because there was no capacity on the node that hosts a primary shard to index documents then the bulk response will contain a `EsRejectedExecutionException` exception in the `BulkResponse` for the bulk items this failure occurred for. This is normal behaviour. However the `BulkProcessor` has retry logic and `EsRejectedExecutionException` is deemed retry-able. The `BulkProcessor` [reuses the same bulk request item instances ](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/action/bulk/Retry.java#L139)(`IndexRequest` in this case) upon retry. This causes the failure mentioned in the stack trace.

martijnvg:
We use `BulkProcessor` to index various features internally (ilm history, watcher history and more).
When an index request is retried then resending that same request can cause the following error:

```
java.lang.IllegalArgumentException: autoGeneratedTimestamp should not be set externally
	at org.elasticsearch.action.bulk.TransportBulkAction.doInternalExecute(TransportBulkAction.java:223) ~[elasticsearch-7.16.1.jar:7.16.1]
	at org.elasticsearch.action.bulk.TransportBulkAction$1.doRun(TransportBulkAction.java:192) ~[elasticsearch-7.16.1.jar:7.16.1]
	at org.elasticsearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:26) [elasticsearch-7.16.1.jar:7.16.1]
	at org.elasticsearch.action.bulk.TransportBulkAction.doExecute(TransportBulkAction.java:200) [elasticsearch-7.16.1.jar:7.16.1]
	at org.elasticsearch.action.bulk.TransportBulkAction.doExecute(TransportBulkAction.java:91) [elasticsearch-7.16.1.jar:7.16.1]
```

This happens when index requests fail because of a retryable error (e.g. EsRejectedExecutionException).
Prior to this change, the `BulkProcessor` (via `Retry`) then reuses the same instance.
With this change, the `autoGeneratedTimestamp` field is unset via a reset method added to `IndexRequest` class,
which should avoid the IAE.

Closes #83927


dakrone:
@martijnvg what's your thought on this solution versus doing something like adding a `.reset()` method on the `IndexRequest` that allows resetting that internal field so it can be retried? Do you have a preference?

martijnvg:
@dakrone I think I'm in favour of that instead of creating a new instance based on the original. The copying is error prone, especially if new fields are added in the future to IndexRequest. I will add the reset method and remove the IndexRequest best effort cloning.

dakrone (approved):
LGTM assuming CI is happy. Do we have a way to add a test for this that doesn't bend over backwards in complexity?

martijnvg:
> Do we have a way to add a test for this that doesn't bend over backwards in complexity?

I don't think so? This doesn't change behaviour or formats. The logic here is tested via `BulkProcessorRetryIT`, `AsyncBulkByScrollActionTests` and `RetryTests` (these test failed when making a mistake with the previous approach).
So I think we're good here.

PR #86189, opened 2022-04-26
Optimize StreamOutput#writeGenericString to fluss less

original-brownbear:
Using the same optimization used in #86114 to make this method a little more efficient.
This one gets quite hot for large cluster states that serialize lots of settings and greatly
benefits from fewer flushes to the deflater output stream.


rjernst (approved):
LGTM

PR #86194, opened 2022-04-26
Remove unused base methods in the HLRC

dakrone:
This removes the unused "base" methods (that interact with ES) from the `HighLevelRestClient`. The
others methods will be removed in subsequent work.

Relates to #83423


dakrone:
(Don't believe the diff lies, this is entirely removing code, git's diffs are just not-so-smart)

jbaiera (approved):
LGTM

PR #86196, opened 2022-04-26
Remove version skips for product header issue

### LINKED ISSUE #84243: Fix up version skips added in #84210 and friends
### Description

In #84210 we added logic to bypass an assertion when running BWC tests against some released versions since the assertion is known to fail on them. At the time there were no released versions on which the assertion holds, although it does hold on all the the relevant branches. This means we cannot be sure which versions will eventually carry the fix. When fixed versions are released, we should update the logic to reflect reality. This'll need doing in every active 8.x branch.

williamrandolph:
We added some logic to skep checking the product header on versions with known bugs in the product header response functionality. Fortunately, we managed to get all of our fixes out in the same 8.x versions where we added the assertion, so we should be able to remove the 8.x skips altogether.

Closes #84243


rjernst:
LGTM. The 8.0.2 label can be removed, that branch is no longer maintained.

rjernst:
LGTM. The 8.0.2 label can be removed, that branch is no longer maintained.

rjernst (approved):

grcevski (approved):
LGTM!

PR #84899, opened 2022-03-10
Adding impacts block to the health info API response

masseyke:
This change introduces the notion of an impact to a health API indicator, and adds impacts to all currently-existing
indicators.
Closes #84773

PR #86197, opened 2022-04-26
Adding potential impacts to remaining health indicators

masseyke:
The ability to add impacts to indicators was added in #84899, but impacts for all indicators other than shards availability were left empty. This commit adds potential impacts for the other indicators.


masseyke:

server/src/main/java/org/elasticsearch/snapshots/RepositoryIntegrityHealthIndicatorService.java
```diff
@@ -89,7 +91,13 @@ public HealthIndicatorResult calculate(boolean includeDetails) {
                 Collections.emptyList()
             );
         }
-
+        List<HealthIndicatorImpact> impacts = Collections.singletonList(
+            new HealthIndicatorImpact(
+                2,
+                "Snapshots in corrupted repositories cannot be restored. Data loss is possible.",
+                List.of(ImpactArea.SEARCH)
```
andreidan, line 98: Should we add a `BACKUP` impact area?  Maybe the corrupted repository should be severity 1?

I'm also wondering if we should include the repository names in the impact (similarly to how we do it in the `shards_availability` indicator) https://github.com/elastic/elasticsearch/blob/1cdb03ddb1c97b4869de33515305b7daee6cebb7/server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java#L762

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/IlmHealthIndicatorService.java
```diff
@@ -62,11 +65,18 @@ public HealthIndicatorResult calculate(boolean includeDetails) {
                 Collections.emptyList()
             );
         } else if (ilmMetadata.getOperationMode() != OperationMode.RUNNING) {
+            List<HealthIndicatorImpact> impacts = Collections.singletonList(
+                new HealthIndicatorImpact(
+                    3,
+                    "Indices are not being rolled over, which could lead to future instability.",
+                    List.of(ImpactArea.SEARCH)
```
andreidan, line 72: Should we add a `DEPLOYMENT MANAGEMENT` impact area?

I wonder about the impact message. Rollover is an implementation detail. Maybe we shouldn't mention it in this case. 

How about?
```
Automatic index lifecycle and data retention management is disabled. The performance and stability of your system could be impacted. 
```

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/slm/SlmHealthIndicatorService.java
```diff
@@ -62,11 +65,18 @@ public HealthIndicatorResult calculate(boolean includeDetails) {
                 Collections.emptyList()
             );
         } else if (slmMetadata.getOperationMode() != OperationMode.RUNNING) {
+            List<HealthIndicatorImpact> impacts = Collections.singletonList(
+                new HealthIndicatorImpact(
+                    3,
+                    "Scheduled snapshots are not running, which could lead to future data loss.",
+                    List.of(ImpactArea.SEARCH)
```
andreidan, line 72: `BACKUP` impact area?

andreidan (commented):
Thanks for adding these Keith.

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/IlmHealthIndicatorService.java
```diff
@@ -62,11 +65,18 @@ public HealthIndicatorResult calculate(boolean includeDetails) {
                 Collections.emptyList()
             );
         } else if (ilmMetadata.getOperationMode() != OperationMode.RUNNING) {
+            List<HealthIndicatorImpact> impacts = Collections.singletonList(
+                new HealthIndicatorImpact(
+                    3,
+                    "Indices are not being rolled over, which could lead to future instability.",
```
elastichelix, line 71: Is there anything more specific that could be said about "future instability"? As in, could you get more specific around potential for running out of resources or data loss or something that might be coming?

masseyke: Just thinking through what happens when shards get too big:

- Slow query response times
- OutOfMemoryErrors
- Disks filling up (leading to maybe slow query response times at first, server crashes eventually)
- Slow recovery time on node failure (which means data might be under-replicated for longer, so slower query response times).

I'm actually not sure if all of those are still the case in 8.0. I know I had some 1 TB shards in 1.x or 2.x, and the whole cluster was just an unstable mess at that point. Maybe slow query times is the thing to highlight here? @andreidan does that list above look right? Does slow query times seem like the thing to highlight here?

server/src/main/java/org/elasticsearch/snapshots/RepositoryIntegrityHealthIndicatorService.java
masseyke: Do we have a working definition of impact severity levels? My thinking was that it wasn't severity 1 because the system is still responsive and usable. But it would certainly be terrifying to not have a functioning snapshot repo.

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/slm/SlmHealthIndicatorService.java
```diff
@@ -62,11 +65,18 @@ public HealthIndicatorResult calculate(boolean includeDetails) {
                 Collections.emptyList()
             );
         } else if (slmMetadata.getOperationMode() != OperationMode.RUNNING) {
+            List<HealthIndicatorImpact> impacts = Collections.singletonList(
+                new HealthIndicatorImpact(
+                    3,
+                    "Scheduled snapshots are not running, which could lead to future data loss.",
```
Leaf-Lin, line 71:
```suggestion
                    "Scheduled snapshots are not running.  Enable schedule snapshot management to prevent loss of data in the future.",
```
I don't think disabling SLM **leads** to future data loss 😛 , more of insurance to prevent data loss... Not sure how to best word this?

x-pack/plugin/ilm/src/test/java/org/elasticsearch/xpack/slm/SlmHealthIndicatorServiceTests.java
```diff
@@ -68,7 +71,13 @@ public void testIsYellowWhenNotRunningAndPoliciesConfigured() {
                     YELLOW,
                     "SLM is not running",
                     new SimpleHealthIndicatorDetails(Map.of("slm_status", status, "policies", 1)),
-                    Collections.emptyList(),
+                    Collections.singletonList(
+                        new HealthIndicatorImpact(
+                            3,
+                            "Scheduled snapshots are not running, which could lead to future data loss.",
```
Leaf-Lin, line 77: Similar to my comment above
```suggestion
                            "Scheduled snapshots are not running.  Enable schedule snapshot management to prevent loss of data in the future."
```

Leaf-Lin (commented):
Just made a few suggestions around impact wording...

server/src/main/java/org/elasticsearch/snapshots/RepositoryIntegrityHealthIndicatorService.java
```diff
@@ -89,7 +92,17 @@ public HealthIndicatorResult calculate(boolean includeDetails) {
                 Collections.emptyList()
             );
         }
-
+        List<HealthIndicatorImpact> impacts = Collections.singletonList(
+            new HealthIndicatorImpact(
+                1,
+                String.format(
+                    Locale.ROOT,
+                    "Snapshots in corrupted repositories %s cannot be restored. Data loss is possible.",
```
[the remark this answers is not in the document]
masseyke: I could make the plurality dependent on how many are actually corrupted. We can have more than one repository though, so it's possible that more than one is corrupted.
My intention was to make it sound really scary. I can tone that down a little like you suggest though.

x-pack/plugin/ilm/src/test/java/org/elasticsearch/xpack/slm/SlmHealthIndicatorServiceTests.java
masseyke: Since this is the impacts block rather than user actions, how about `Scheduled snapshots are not running. There might not be backups of the data.`? Or `Scheduled snapshots are not running. There might not be backups of the data that could be used to restore if data is lost in the future.`?

server/src/main/java/org/elasticsearch/cluster/coordination/InstanceHasMasterHealthIndicatorService.java
```diff
@@ -27,6 +31,7 @@ public class InstanceHasMasterHealthIndicatorService implements HealthIndicatorS
 
     private static final String INSTANCE_HAS_MASTER_GREEN_SUMMARY = "Health coordinating instance has a master node.";
     private static final String INSTANCE_HAS_MASTER_RED_SUMMARY = "Health coordinating instance does not have a master node.";
+    private static final String NO_MASTER_IMPACT = "The cluster cannot create, delete, or rebalance indices, and is likely to be unstable";
```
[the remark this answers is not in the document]
masseyke: > In addition to "create", I think the cluster won't be able to accept any new writes to existing indices (append/update) either?

Ohh, you're right. Writes fail.

> All the _cat (and some other APIs) will not work.

I didn't realize this, but you're right about that, too.

> For "connected service"

Do we need to call this out separately? They only don't work because the various APIs we're already talking about don't work, right?

> Maybe somehow counterintuitively, search will continue to work when the master is missing. Is it worth pointing this out?

If you don't have a master, you probably ought to drop everything and fix that problem. There's no point in making it sound not quite so bad is there?

masseyke:
@elasticmachine update branch

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/IlmHealthIndicatorService.java
masseyke: I've updated this list based on feedback from @Leaf-Lin at https://github.com/elastic/elasticsearch/pull/86197#pullrequestreview-962768375.

andreidan (approved):
LGTM, thanks for working on this Keith

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/IlmHealthIndicatorService.java
andreidan: ++ Keith - I think for a user the "data retention" aspect would be very important (data staying for too long in the system could breach some user data laws and such) and financially, data not moving through the tiers will have a very unpleasant impact on the bills

server/src/main/java/org/elasticsearch/snapshots/RepositoryIntegrityHealthIndicatorService.java
andreidan: Ah not really for now. The severity will be very important when we'll have a global impacts section and we'll possibly only report `severity: 1` (requirements TBD)

Leaf-Lin (approved):
LGTM

x-pack/plugin/ilm/src/test/java/org/elasticsearch/xpack/slm/SlmHealthIndicatorServiceTests.java
Leaf-Lin: I know it's a bit wordy, but this feels more correct than what we had previous 👍 .

server/src/main/java/org/elasticsearch/snapshots/RepositoryIntegrityHealthIndicatorService.java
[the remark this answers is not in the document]
Leaf-Lin: Could you please clarify if the check only looks at the single repository `found-snapshot` that is defined by ESS? or does it check all possible repositories from `GET _snapshot/_all`?

server/src/main/java/org/elasticsearch/cluster/coordination/InstanceHasMasterHealthIndicatorService.java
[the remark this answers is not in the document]
Leaf-Lin: >> For "connected service"

> Do we need to call this out separately? They only don't work because the various APIs we're already talking about don't work, right?

I personally would prefer to make this a bit more explicitly but I will defer this to @shubhaat .  
The question is when master is not available, all connected services like "Kibana/EnterpriseSearch/Fleet" will stop working.  Should these connected services be listed as part of the `impact area`? and should the description of impact includes a statement on the "connected services"?

server/src/main/java/org/elasticsearch/snapshots/RepositoryIntegrityHealthIndicatorService.java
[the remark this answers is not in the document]
masseyke: It checks all repositories in the cluster state, which is equivalent to `GET _snapshot/_all`. I've changed it to use "repository" if there is only one corrupted, and "repositories" if there are more than one.

server/src/main/java/org/elasticsearch/cluster/coordination/InstanceHasMasterHealthIndicatorService.java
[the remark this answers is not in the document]
masseyke: It makes sense to me that we'd want to tell users that. But we don't usually reference Enterprise Search in Elasticsearch code. That seems like something we'd do at the Cloud level -- it sees that Elasticsearch is broken, so it knows to tell the user that all of the applications it is hosting that are built on Elasticsearch are not working. I'll go ahead and merge this for now, and can open a separate PR about this if the consensus is that we want to start referencing applications built on top of Elasticsearch more inside of Elasticsearch.

PR #86198, opened 2022-04-26
Cleanup Terminal to make it easier to subclass

rjernst:
Terminal is the abstraction Elasticsearch uses for all input and output,
both character based and binary. In an interactive shell, this is backed
by Java's Console, and in non-interactive it is backed by
stdin/stdout/stderr. Over time, the Terminal class has been amended to
support several different use cases, which has made constructing
subclasses for testing or filter based implementations complex. This
commit reworks Terminal so that the readers/writers/streams are
constructor arguments, instead of overrides. This allows subclasses to
simply call super with what is neeeded, rather than overloading several
methods and adding the same boilerplate implementation as others.

Note that the majority of the modifications here are to tests because
MockTerminal now has a factory method instead of direct constructor.

relates #85758

### DIFF
--- a/libs/cli/src/main/java/org/elasticsearch/cli/Terminal.java
+++ b/libs/cli/src/main/java/org/elasticsearch/cli/Terminal.java
@@ -10,9 +10,10 @@ package org.elasticsearch.cli;
 
 import org.elasticsearch.core.Nullable;
 
-import java.io.BufferedReader;
 import java.io.Console;
+import java.io.EOFException;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.PrintWriter;
@@ -51,14 +52,31 @@ public abstract class Terminal {
         VERBOSE /* printed only when cli is passed verbose option */
     }
 
+    private final Reader reader;
+
+    private final PrintWriter outWriter;
+
+    private final PrintWriter errWriter;
+
+    private final InputStream inStream;
+
+    private final OutputStream outStream;
+
     /** The current verbosity for the terminal, defaulting to {@link Verbosity#NORMAL}. */
     private Verbosity currentVerbosity = Verbosity.NORMAL;
 
-    /** The newline used when calling println. */
-    private final String lineSeparator;
-
-    protected Terminal(String lineSeparator) {
-        this.lineSeparator = lineSeparator;
+    protected Terminal(
+        Reader reader,
+        PrintWriter outWriter,
+        PrintWriter errWriter,
+        @Nullable InputStream inStream,
+        @Nullable OutputStream outStream
+    ) {
+        this.reader = reader;
+        this.outWriter = outWriter;
+        this.errWriter = errWriter;
+        this.inStream = inStream;
+        this.outStream = outStream;
     }
 
     /** Sets the verbosity of the terminal. */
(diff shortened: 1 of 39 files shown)


test/framework/src/main/java/org/elasticsearch/cli/MockTerminal.java
```diff
@@ -34,60 +64,43 @@ public class MockTerminal extends Terminal {
...
-    @Override
-    public PrintWriter getErrorWriter() {
-        return errorWriter;
+    private static PrintWriter newPrintWriter(OutputStream out) {
+        return new PrintWriter(new OutputStreamWriter(out, StandardCharsets.UTF_8));
     }
 
-    public void setHasOutputStream(boolean hasOutputStream) {
-        this.hasOutputStream = hasOutputStream;
+    public static MockTerminal create(boolean supportsBinary) {
```
williamrandolph, line 91: Am I correct in reading that we don't yet call this with an argument of `true`? That won't block approval but I'm curious whether there are plans to use a MockTerminal that supports binary in upcoming work.

libs/cli/src/main/java/org/elasticsearch/cli/Terminal.java
```diff
@@ -51,40 +52,87 @@ public enum Verbosity {
...
+
     /** The current verbosity for the terminal, defaulting to {@link Verbosity#NORMAL}. */
     private Verbosity currentVerbosity = Verbosity.NORMAL;
 
-    /** The newline used when calling println. */
-    private final String lineSeparator;
-
-    protected Terminal(String lineSeparator) {
-        this.lineSeparator = lineSeparator;
+    protected Terminal(
```
williamrandolph, line 68: Either here or on the field definitions, I think we should document that the `inStream` if non-null is likely not independent of the `reader`, and the `outStream` if non-null is likely not independent of the `outWriter`, or at least there is no expectation of independence between the items in those pairs.

libs/cli/src/main/java/org/elasticsearch/cli/Terminal.java
```diff
@@ -94,26 +142,29 @@ public final void println(CharSequence msg) {
...
 
     /** Prints message to the terminal's standard output at {@code verbosity} level, without a newline. */
     public final void print(Verbosity verbosity, String msg) {
-        print(verbosity, msg, false);
+        print(verbosity, outWriter, msg, false);
     }
 
     /** Prints message to the terminal at {@code verbosity} level, without a newline. */
-    protected void print(Verbosity verbosity, String msg, boolean isError) {
+    protected void print(Verbosity verbosity, PrintWriter writer, CharSequence msg, boolean newline) {
```
williamrandolph, line 154: I like the refactoring here.

server/src/main/java/org/elasticsearch/bootstrap/plugins/LoggerTerminal.java
```diff
@@ -48,32 +47,13 @@ public char[] readSecret(String prompt) {
...
-
-    @Override
-    public PrintWriter getErrorWriter() {
-        throw new UnsupportedOperationException();
-    }
-
-    @Override
-    protected void print(Verbosity verbosity, String msg, boolean isError) {
+    protected void print(Verbosity verbosity, PrintWriter writer, CharSequence msg, boolean newline) {
+        boolean isError = writer == getErrorWriter();
```
williamrandolph, line 51: If this class is initialized with `null` for the outWriter and for the errWriter, won't `isError` always evaluate to `true` here?

test/framework/src/main/java/org/elasticsearch/cli/MockTerminal.java
```diff
@@ -8,24 +8,54 @@
...
+            count += bytes.length;
+        }
+
+        void clear() {
+            pos = 0;
+            count = 0;
+        }
+    }
+
+    private static final byte[] NEWLINE = new byte[] { '\n' };
```
williamrandolph, line 55: Can we comment here that we are deliberately using Unix line endings in tests?

williamrandolph (commented):
Looking really good so far. I just have a few questions about things in the code.

test/framework/src/main/java/org/elasticsearch/cli/MockTerminal.java
rjernst: Yes, that is correct. I plan to work on that in a future change, so that we can fix the input reader to not consume past each line of input (leaving binary data that we could read on the stream).

server/src/main/java/org/elasticsearch/bootstrap/plugins/LoggerTerminal.java
rjernst: It would, but passing null for the readers is just a hack LoggingTerminal uses. The expectation is passing null for those means overriding this method.

libs/cli/src/main/java/org/elasticsearch/cli/Terminal.java
rjernst: I"ll take a pass at better documenting the entire class.

grcevski (approved):
LGTM! Out of curiosity, was there a particular reason to split text input and secret input in the past? Some of the test failures seem related? I really like the refactor.

rjernst:
> was there a particular reason to split text input and secret input in the past?

There wasn't, other than the way it was implemented was through overriding the separate readSecret and readText methods. I've had to fix several tests, though, that were relying on these being two separate streams of data, where now they are one stream so the order they are added matters. Long term, I also think it would be good to have some additional assertions, like making sure adding a secret is only read with readSecret. I'm going to add back the addSecretInput, but keep it as a single stream, so that we can at least add those assertions in the future.

PR #86207, opened 2022-04-27
Better failure for source-only snapshots of partially/fully mounted indices

tlrx:
Taking a snapshot of a partially or fully mounted index in a source-only repository is not supported and stops the data node by throwing an AssertionError.

Instead of throwing such an error, this change simply fails the snapshot of searchable snapshot shards in source-only repository and report a better snapshot failure.


x-pack/plugin/core/src/main/java/org/elasticsearch/snapshots/sourceonly/SourceOnlySnapshotRepository.java
```diff
@@ -148,7 +148,15 @@ public void snapshotShard(SnapshotShardContext context) {
         final Store store = context.store();
         Directory unwrap = FilterDirectory.unwrap(store.directory());
         if (unwrap instanceof FSDirectory == false) {
-            throw new AssertionError("expected FSDirectory but got " + unwrap.toString());
+            context.onFailure(
+                new IllegalStateException(
+                    context.indexCommit()
+                        + " is not a regular index and cannot be snapshotted into a source-only repository ["
+                        + unwrap.toString()
+                        + ']'
```
idegtiarenko, line 156: I believe `unwrap` directory is related to the index type, not a repository, is it? If so may be it is better to rephrase this to:
```
context.indexCommit() + " is not a regular index, but ["+ unwrap.toString()+"] and cannot be snapshotted into a source-only repository"
```

idegtiarenko (approved):

x-pack/plugin/core/src/main/java/org/elasticsearch/snapshots/sourceonly/SourceOnlySnapshotRepository.java
tlrx: Thanks for the suggestion, I'll borrow it.

tlrx:
Thanks Ievgen

PR #86210, opened 2022-04-27
Compute master task batch summary lazily

DaveCTurner:
Today we construct a string description of every batch of master tasks
in case we need to log it, grouping the tasks by `source` and calling
`toString()` on each one. Yet, we almost never emit the description in
the logs so this effort is usually wasted. This commit makes the summary
computation lazy, deferring the effort until we discover it's actually
needed and avoiding it entirely in most cases.

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/service/BatchSummary.java
+++ b/server/src/main/java/org/elasticsearch/cluster/service/BatchSummary.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.cluster.service;
+
+import org.elasticsearch.common.Strings;
+import org.elasticsearch.common.util.LazyInitializable;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class BatchSummary {
+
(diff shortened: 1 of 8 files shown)


idegtiarenko (approved):

server/src/main/java/org/elasticsearch/cluster/service/BatchSummary.java
```diff
@@ -0,0 +1,52 @@
...
+        lazyDescription = new LazyInitializable<>(() -> {
+            final Map<String, List<TaskBatcher.BatchedTask>> processTasksBySource = new HashMap<>();
+            for (final var task : allTasks) {
+                processTasksBySource.computeIfAbsent(task.source, s -> new ArrayList<>()).add(task);
+            }
+            final StringBuilder output = new StringBuilder();
+            Strings.collectionToDelimitedStringWithLimit((Iterable<String>) () -> processTasksBySource.entrySet().stream().map(entry -> {
+                String tasks = firstTask.describeTasks(entry.getValue());
+                return tasks.isEmpty() ? entry.getKey() : entry.getKey() + "[" + tasks + "]";
+            }).filter(s -> s.isEmpty() == false).iterator(), ", ", "", "", MAX_TASK_DESCRIPTION_CHARS, output);
```
idegtiarenko, line 35: I wonder if this could be simplified.

idegtiarenko: I was thinking about something like:
```
allTasks.stream().collect(groupingBy(task -> task.source, collectingAndThen(toList(), updateTask::describeTasks)));
```
that would return `source -> task descriptions` but this would still require second stream to add the keys...

Probably not worth it.

DaveCTurner: Yeah this is strangely ugly but I don't see any obviously-better implementations either. At least not without some substantial API changes that I don't want to work on right now.

Issue #77466, opened 2021-09-09
Fix Large Shard Count Scalability Issues


original-brownbear:
This meta issue tracks known issues with scaling clusters to large numbers of shards.

- Security
  - [ ] https://github.com/elastic/elasticsearch/issues/67987
     - [x] https://github.com/elastic/elasticsearch/issues/79632

- General
  - [ ] https://github.com/elastic/elasticsearch/issues/51992
  - [ ] https://github.com/elastic/elasticsearch/issues/87555
  - [x] https://github.com/elastic/elasticsearch/pull/77546
  - [ ] https://github.com/elastic/elasticsearch/issues/79563
  - [x] https://github.com/elastic/elasticsearch/pull/79793
  - [x] https://github.com/elastic/elasticsearch/issues/79906
  - [x] https://github.com/elastic/elasticsearch/pull/80064
  - [ ] https://github.com/elastic/elasticsearch/issues/80694
  - [x] https://github.com/elastic/elasticsearch/issues/81626
  - [x] https://github.com/elastic/elasticsearch/issues/81627
  - [x] https://github.com/elastic/elasticsearch/issues/81628
  - [ ] https://github.com/elastic/elasticsearch/issues/81846
  - [x] https://github.com/elastic/elasticsearch/issues/81880
  - [x] https://github.com/elastic/elasticsearch/issues/82337
  - [x] https://github.com/elastic/elasticsearch/issues/82342
  - [x] https://github.com/elastic/elasticsearch/pull/82608
  - [x] https://github.com/elastic/elasticsearch/pull/82227
  - [x] https://github.com/elastic/elasticsearch/pull/82830
  - [ ] https://github.com/elastic/elasticsearch/issues/83049
  - [ ] https://github.com/elastic/elasticsearch/issues/83203
  - [x] https://github.com/elastic/elasticsearch/issues/83204 -> #85380
  - [x] https://github.com/elastic/elasticsearch/issues/85839
  - [x] https://github.com/elastic/elasticsearch/issues/86639
  - [x] https://github.com/elastic/elasticsearch/issues/87681
  - [ ] https://github.com/elastic/elasticsearch/issues/90631
  - [x] https://github.com/elastic/elasticsearch/pull/92902
  - [x] https://github.com/elastic/elasticsearch/issues/97222
  - [x] https://github.com/elastic/elasticsearch/issues/97914
  - [ ] https://github.com/elastic/elasticsearch/issues/101805

- Snapshots + SLM
[...]

PR #86223, opened 2022-04-27
Remove shards per gb of heap guidance

original-brownbear:
This guidance does not apply any longer.
The overhead per shard has been significantly reduced in recent versions
and this kind of rule of thumb will be too pessimistic in many if not
most cases and might be too optimistic in other specific ones.
-> remove it

relates #77466


docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -68,7 +68,17 @@ This decreases the number of segments, which means less metadata is kept in
 heap memory.
 
 Every mapped field also carries some overhead in terms of memory usage and disk
-space. By default {es} will automatically create a mapping for every field in
+space. While the exact amount resource usage of each additional field depends
+on the type of field, an additional 512 bytes in memory use  for each additional
```
DaveCTurner, line 72: 512b seems tight, is that a worst-case guarantee? Even in 8.2? I think we can afford to be a bit conservative with the recommendation here.

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -68,7 +68,17 @@ This decreases the number of segments, which means less metadata is kept in
...
+mapped field on data nodes is a good approximation in most cases.
+As a result, the more fields the mappings in your indices contain, the fewer
+indices and shards will each individual data node be able to hold,
+all other things being equal. In addition to the per field overhead you should
+at least have an additional 512Mb of heap available per data node.
```
DaveCTurner, line 77: nit: bytes not bits

```suggestion
at least have an additional 512MB of heap available per data node.
```

Also is that 512MiB enough for everything else the node might need to do, or is that just enough for it to survive? It sounds pretty tight, maybe we need some more caveats here.

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -175,35 +185,15 @@ index                                 prirep shard store
...
-Some system indices for {enterprise-search-ref}/index.html[Enterprise Search]
-are nearly empty and rarely used. Due to their low overhead, you shouldn't
-count shards for these indices toward a node's shard limit.
-
-To check the current size of each node's heap, use the <<cat-nodes,cat nodes
-API>>.
-
-[source,console]
-----
-GET _cat/nodes?v=true&h=heap.current
```
DaveCTurner, line 195: I think it'd be helpful to keep these commands here.

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -175,35 +185,15 @@ index                                 prirep shard store
...
-----
-// TEST[setup:my_index]
+==== Aim for 3000 indices or fewer per GB of heap memory on master nodes
+
+The number of indices a master node can manage is proportional to the node's
+heap memory. The exact amount of heap memory each additional index requires
+depends on various factors. The include but are not limited to the size of its mapping,
+the number of shards per index or whether its mapping is shared with other indices.
+A good rule of thumb is to aim for 3000 indices or fewer per GB of heap on master nodes.
+For example, if your cluster contains 12,000 indices and each master node should have
```
DaveCTurner, line 195:
```suggestion
For example, if your cluster contains 12,000 indices then each dedicated master node should have
```

for non-dedicated masters should we say something about allowing for this much overhead too?

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -68,7 +68,17 @@ This decreases the number of segments, which means less metadata is kept in
 heap memory.
 
 Every mapped field also carries some overhead in terms of memory usage and disk
-space. By default {es} will automatically create a mapping for every field in
+space. While the exact amount resource usage of each additional field depends
+on the type of field, an additional 512 bytes in memory use  for each additional
+mapped field on data nodes is a good approximation in most cases.
+As a result, the more fields the mappings in your indices contain, the fewer
+indices and shards will each individual data node be able to hold,
```
henningandersen, line 75:
```suggestion
As a result, the more mapped fields per index, the fewer
indices and shards can be held on each individual data node,
```

docs/reference/how-to/size-your-shards.asciidoc
henningandersen: On the flip side of that is that running with a few indices with few fields on a 512MB heap with small workload is perfectly possible too. So either the limit need to be lower or a less tight wording than "at least".

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -175,35 +185,15 @@ index                                 prirep shard store
...
-GET _cat/shards?v=true
-----
-// TEST[setup:my_index]
+==== Aim for 3000 indices or fewer per GB of heap memory on master nodes
+
+The number of indices a master node can manage is proportional to the node's
+heap memory. The exact amount of heap memory each additional index requires
+depends on various factors. The include but are not limited to the size of its mapping,
+the number of shards per index or whether its mapping is shared with other indices.
+A good rule of thumb is to aim for 3000 indices or fewer per GB of heap on master nodes.
```
henningandersen, line 194: Can we put a few preconditions on this rule of thumb, like less than 1000 fields and with most indices created identically from the same template?

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -175,35 +185,15 @@ index                                 prirep shard store
...
-----
-GET _cat/shards?v=true
-----
-// TEST[setup:my_index]
+==== Aim for 3000 indices or fewer per GB of heap memory on master nodes
+
+The number of indices a master node can manage is proportional to the node's
+heap memory. The exact amount of heap memory each additional index requires
+depends on various factors. The include but are not limited to the size of its mapping,
+the number of shards per index or whether its mapping is shared with other indices.
```
henningandersen, line 193: > whether its mapping is shared with other indices.
I wonder if we could formulate this in less implementation detailed way as `whether indices are created identically from the same template with no dynamic mapping involved`?

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: It's quite conservative for 8.3 I'd say. it's ~4x what a heap dump for Beats indices would suggest in 8.3. 8.2 is probably about twice as expensive (this is only sorta true and depends on the field type).
=> how about 1kb per field then to have a nice round number?

henningandersen (commented):
Thanks Armin, left a few comments.

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: The idea was that the 512B above was chosen conservative enough that any additional overheads are covered. In the end 512M is plenty for just data node operations IMO if we have 0.5k/field and especially so if we go with 1k/field.
Maybe add caveats for when the data node is used for ingest and/or coordination? (it's hard to put numbers on either but we could just warn about them in general?)

DaveCTurner: The circuit breakers that apply to data node operations don't really take the shards-and-fields overhead into account tho, e.g. the fielddata and request breakers just act as if it's ok to use 40% and 60% of the total heap respectively. In the example here, 80% of the heap is already consumed with shards-and-fields overhead, so really (with the real-memory CB) there's only 15% left for doing any useful work.

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: My assumption was that once these numbers start to matter you'd be using dedicated master nodes anyway?
The overhead is definitely quite a bit less if you're on a non-dedicated master node because you share network buffers and such with other operations I'd say. Maybe better to add a line recommending dedicated masters here above say 5k indices or so instead?

DaveCTurner: Not sure, that seems a pretty low threshold to be going for dedicated masters and quite possibly the wrong variable to choose too. A 3x31GiB cluster with enough CPU to spare should be able to deal with more than 5k indices, and probably better to stick with 3 big nodes rather than add 3 tiny masters on top?

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: ++ added a verb to this though

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: These numbers are taken from Beats indices which do indeed come with 4.x k fields on average. The field count really doesn't matter much here, what matters is mostly that mappings are shared. Added a comment to that effect that also mentions the dynamic mapping updates now.

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: Sure brought them back. The shard per node count thing seemed a little off now but you're right these are still helpful :)

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: Reworded

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: >  A 3x31GiB cluster with enough CPU to spare should be able to deal with more than 5k indices, and probably better to stick with 3 big nodes rather than add 3 tiny masters on top?

Yea probably in most cases (disks would be my biggest worry here but with much more CS batching nowadays it's less of an issue probably). I think I'd keep it simple then and really just state that the numbers must be added for non-dedicated master nodes even if it's a little pessimistic heap wise if that's ok?

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: > 80% of the heap is already consumed with shards-and-fields overhead

Not really. Now that I made it 1024B in particular, we vastly overestimate the direct per-field overhead. So for larger nodes, we will have about a 50%+ margin here now. If we recommend 4G heap for the fields in the example, then really the steady state memory usage from them is probably 20% of the heap tops leaving enough room for the other operations.
The 512M was mostly meant to add a little more margin for error and to get reasonable numbers for smaller nodes.

original-brownbear:
Jenkins run elasticsearch-ci/part-2 (unrelated)

docs/reference/how-to/size-your-shards.asciidoc
DaveCTurner: sounds good to me

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -175,17 +185,18 @@ index                                 prirep shard store
...
-Some system indices for {enterprise-search-ref}/index.html[Enterprise Search]
-are nearly empty and rarely used. Due to their low overhead, you shouldn't
-count shards for these indices toward a node's shard limit.
+==== Aim for 3000 indices or fewer per GB of heap memory on master nodes
+
+The number of indices a master node can manage is proportional to the node's
+heap memory. The exact amount of heap memory each additional index requires
+depends on various factors. Theses include but are not limited to the size of its mapping,
+the number of shards per index or whether indices are created identically from the same
+template with no dynamic mapping involved.
```
DaveCTurner, line 194: slight rework suggestion here:

```suggestion
depends on various factors such as the size of the mapping and the number of
shards per index. Where possible, the master node will deduplicate the mapping
metadata across indices with identical mappings. Mapping metadata can be
deduplicated across indices that are created from the same index template as
long as they do not use dynamic mapping updates.
```

DaveCTurner (commented):
a few more nits but this looks pretty much good to go IMO, I'll let Henning take a final look too

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -5,6 +5,7 @@
     <inspection_tool class="PointlessBooleanExpression" enabled="false" level="WARNING" enabled_by_default="false">
       <option name="m_ignoreExpressionsContainingConstants" value="true" />
     </inspection_tool>
+    <inspection_tool class="SizeToLength" enabled="false" level="WARNING" enabled_by_default="false" />
```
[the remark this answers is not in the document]
original-brownbear: nah new Idea version is messing with this somehow, reverted :)

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: ++ nice thanks, applied!

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -72,12 +72,12 @@ space. While the exact amount resource usage of each additional field depends
 on the type of field, an additional 1024B in memory use  for each additional
```
Leaf-Lin, line 72:
```suggestion
on the type of field, an additional 1024B in memory used for each additional
```

Should this be `used` instead of `use`?  Also I removed an additional space.

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -72,12 +72,12 @@ space. While the exact amount resource usage of each additional field depends
...
 mapped field on data nodes is a good approximation in most cases.
 As a result, the more mapped fields you have per index, the fewer
-indices and shards will each individual data node be able to hold,
+indices and shards each individual data node will be able to hold,
 all other things being equal. In addition to the per field overhead you should
 at least have an additional 512MB of heap available per data node.
 For example, holding 1000 indices that each contain 4000 fields will require
```
Leaf-Lin, line 78:
```suggestion
For example, holding 1000 indices that each contains 4000 fields will require
```
Should this be contains following each being a singular?

Leaf-Lin, line 78: I think it's better to talk about a node holding 1000 shards not indices here for calculating heap usage?

```suggestion
For example, holding 1000 shards that each contain 4000 fields will require
```

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -72,12 +72,12 @@ space. While the exact amount resource usage of each additional field depends
...
-1000 x 4000 x 1024b = 4Gb for the fields and another 512Mb for a total of
-at least 4.5Gb on the data node.
+1000 × 4000 × 1024B = 4GB for the fields and another 512MB for a total of
+at least 4.5GB on the data node.
```
Leaf-Lin, line 80:
```suggestion
at least 4.5GB heap or 9GB RAM on the data node.
```

I would prefer we call out the heap vs RAM carefully here.

Leaf-Lin (commented):
In the example, we talked about having 1000 shards that each contains 4000 fields means a node of 4.5GB heap can hold 1000 shards. 

Later on we talked about for non-master node, each node should preserve `total_indices_in_cluster/3000` per GB heap.
> a good rule of thumb is to aim for 3000 indices or fewer per GB of heap on master nodes
> For non-dedicated master nodes, the same rule holds and should be added to the heap requirements of other node roles.

Does that mean we should also mention that one should adjust `cluster.max_shards_per_node` setting based on the heap available on a node? 
```
max_idling_shards_per_node = (per_node_memory_in_GB/2 - 0.5GB - total_indices_in_cluster/3000) / number_of_fields_per_shard x GB_to_B           ...(1)

total_indices_in_cluster   = number_of_data_nodes x max_idling_shards_per_node                                                                  ...(2)
```

From (1) and (2), you can derive:
```
max_idling_shards_per_node = (per_node_memory_in_GB/2 - 0.5GB) x 3000 / (3 x number_of_fields_per_shard/1000 + number_of_data_node)             ...(3)
```

The largest node (heap size) we recommend for Elasticsearch is 32GB heap or 64GB RAM would be able to hold 7,269 **idling** shards in 1-data-node-cluster or 4,295 **idling** shards in a 10-data-node-cluster given 4,000 fields per shard based on this formula. For a 60-data-node cluster, `max_idling shards_per_node` will be dropped further to 1,313.  The more data nodes in the cluster, the few shards per node each node can hold.   

The effect of fields count appears to be linear in a 1-node cluster changed to almost negligible as the cluster size grows large (>30 nodes).

[Later on on the doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/size-your-shards.html#_this_action_would_add_x_total_shards_but_this_cluster_currently_has_yz_maximum_shards_open), we refer to adjusting the default `cluster.max_shards_per_node` value from 1,000 to 1,200 temporarily.  Do we want to expand this section to include memory-based max _idling_ shards per node value?

⚠️ , we might want to emphasise _idling_ a bit more in the `Each index and shard has overhead` section?

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -68,7 +68,17 @@ This decreases the number of segments, which means less metadata is kept in
...
+space. While the exact amount resource usage of each additional field depends
+on the type of field, an additional 1024B in memory use  for each additional
+mapped field on data nodes is a good approximation in most cases.
+As a result, the more mapped fields you have per index, the fewer
+indices and shards each individual data node will be able to hold,
+all other things being equal. In addition to the per field overhead you should
+at least have an additional 512MB of heap available per data node.
+For example, holding 1000 indices that each contain 4000 fields will require
+1000 × 4000 × 1024B = 4GB for the fields and another 512MB for a total of
+at least 4.5GB on the data node.
```
Leaf-Lin, line 80:
```suggestion
at least 4.5GB on the data node. As a reminder, 4.5GB heap here refers to holding the idling data 
in heap on the data node only. Heap usage on data nodes is also heavily dependent on use cases  
as well as how often these shards are indexed, updated or queried, and if there is heavy 
aggregation performed on the data nodes. The max number of shards a node can hold should be 
 adjusted by taking your usage pattern into consideration.
```

How can we emphasise this is for idling shards only? Maybe add a sentence like this?

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -175,17 +185,20 @@ index                                 prirep shard store
...
+The number of indices a master node can manage is proportional to the node's
+heap memory. The exact amount of heap memory each additional index requires
+depends on various factors such as the size of the mapping and the number of
+shards per index. Where possible, the master node will deduplicate the mapping
+metadata across indices with identical mappings. Mapping metadata can be
+deduplicated across indices that are created from the same index template as
+long as they do not use dynamic mapping updates.
+Under the assumption that indices are mostly created identically from a small number of templates
+and no dynamic mapping updates are involved, a good rule of thumb is to aim for 3000 indices or
+fewer per GB of heap on master nodes. For example, if your cluster contains 12,000 indices then
```
Leaf-Lin, line 199: Can we explicitly call out that the rule of thumb here refers to an index having 1 primary and 1 replica shards?

docs/reference/how-to/size-your-shards.asciidoc
DaveCTurner: Either is acceptable IMO ("memory use" works here as a nounal phrase)

docs/reference/how-to/size-your-shards.asciidoc
DaveCTurner: Don't know how to explain this one, but `contain` is right to my ear. I might say "holding 1000 indices each containing 4000 fields" instead.

docs/reference/how-to/size-your-shards.asciidoc
DaveCTurner: +1 to mentioning heap, but I would rather not imply that RAM is always twice the heap. It's the heap size that matters here I think.

docs/reference/how-to/size-your-shards.asciidoc
DaveCTurner: I looked it up: in this construction `each` is an adverb modifying the verb `contain` rather than being its subject. The indices are the subject, so the verb is plural.

DaveCTurner: How about something like `holding shards from 1000 different indices`? Mappers are shared across shards from the same index so adding multiple shards from each index doesn't add to this memory usage.

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -68,7 +68,17 @@ This decreases the number of segments, which means less metadata is kept in
 heap memory.
 
 Every mapped field also carries some overhead in terms of memory usage and disk
-space. By default {es} will automatically create a mapping for every field in
+space. While the exact amount resource usage of each additional field depends
+on the type of field, an additional 1024B in memory use  for each additional
+mapped field on data nodes is a good approximation in most cases.
+As a result, the more mapped fields you have per index, the fewer
+indices and shards each individual data node will be able to hold,
```
DaveCTurner, line 75: Similarly here, could we expose relevant metrics in nodes stats?

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -68,7 +68,17 @@ This decreases the number of segments, which means less metadata is kept in
...
+all other things being equal. In addition to the per field overhead you should
+at least have an additional 512MB of heap available per data node.
```
henningandersen, line 77: I think this moves a proportion of our customers from inside to outside our recommendations, which I think is not the purpose here. I suggest to rephrase to something like:

In addition to the per field overhead you should have heap available for the base Elasticsearch heap usage as well as the heap usage of the workload. For a small workload, 512MB of heap in addition to the per field overhead suffice in many cases. For tight workloads, it is possible to go even lower.

docs/reference/how-to/size-your-shards.asciidoc
henningandersen: Is there not also a per shard overhead per field (in lucene)?

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -175,17 +185,20 @@ index                                 prirep shard store
...
-are nearly empty and rarely used. Due to their low overhead, you shouldn't
-count shards for these indices toward a node's shard limit.
+==== Aim for 3000 indices or fewer per GB of heap memory on master nodes
+
+The number of indices a master node can manage is proportional to the node's
+heap memory. The exact amount of heap memory each additional index requires
+depends on various factors such as the size of the mapping and the number of
+shards per index. Where possible, the master node will deduplicate the mapping
+metadata across indices with identical mappings. Mapping metadata can be
+deduplicated across indices that are created from the same index template as
```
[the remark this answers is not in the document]
original-brownbear: We could pretty easily since mapping stats computation uses deduplication internally these days.

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: Yea, as with mapping stats this should be quite straight forward actually

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: > Is there not also a per shard overhead per field (in lucene)?

it's relatively small compared to the ES overhead for beats indices. Given that we now use a value of 1KB here that gives us a lot of breathing space I think we more than cover the Lucene overheads in most cases. I'd rather keep the per index documentation here as it applies to the vast majority of cases (and as far as I can tell at least for the Beats mappings, would even apply well for significant outliers with say 5 shards per the same index on a data node).

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -68,7 +68,17 @@ This decreases the number of segments, which means less metadata is kept in
 heap memory.
 
 Every mapped field also carries some overhead in terms of memory usage and disk
-space. By default {es} will automatically create a mapping for every field in
+space. While the exact amount resource usage of each additional field depends
+on the type of field, an additional 1024B in memory use  for each additional
```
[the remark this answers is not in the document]
original-brownbear: ++ adjusted

docs/reference/how-to/size-your-shards.asciidoc
original-brownbear: ++ reworded to your suggestion

docs/reference/how-to/size-your-shards.asciidoc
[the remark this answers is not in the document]
DaveCTurner: :+1: should we do that before making these changes? It'd be a lot simpler to say "check this API" rather than just trying to explain the concepts in words.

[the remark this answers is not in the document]
original-brownbear: I'd say we can adjust the docs here a little more when making those changes? IMO these changes still get us to a better spot than where we currently are and I'm not sure we should delay this that much more?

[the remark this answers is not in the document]
DaveCTurner: Ok, I opened #86639 to track this.

docs/reference/how-to/size-your-shards.asciidoc
DaveCTurner: Similarly, see #86639

docs/reference/how-to/size-your-shards.asciidoc
Leaf-Lin: Let's remove that extra space then.

Leaf-Lin (approved):
I think we just need to remove that extra space in "memory use  for each", otherwise LGTM.

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -68,7 +68,20 @@ This decreases the number of segments, which means less metadata is kept in
 heap memory.
 
 Every mapped field also carries some overhead in terms of memory usage and disk
-space. By default {es} will automatically create a mapping for every field in
+space. While the exact amount resource usage of each additional field depends
+on the type of field, an additional 1KB in memory use for each additional
```
DaveCTurner, line 72: nit

```suggestion
on the type of field, an additional 1kB in memory use for each additional
```

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -68,7 +68,20 @@ This decreases the number of segments, which means less metadata is kept in
...
+mapped field on data nodes is a good approximation in most cases.
+As a result, the more mapped fields you have per index, the fewer
+indices and shards each individual data node will be able to hold,
+all other things being equal. In addition to the per field overhead you should
+have heap available for the base Elasticsearch heap usage as well as the heap
+usage of the workload. For a small workload, 512MB of heap in addition to the
+per field overhead suffice in many cases. For tight workloads, it is possible to
```
DaveCTurner, line 79:
```suggestion
per field overhead suffice in many cases. For light workloads, it may be possible to
```

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -68,7 +68,20 @@ This decreases the number of segments, which means less metadata is kept in
...
+go even lower.
+For example, holding 1000 indices that each contain 4000 fields will require
```
DaveCTurner, line 81: I sort of wonder whether the concrete figures given here should be another `Aim for` section below? Maybe later, just recording this thought for posterity here.

> Aim for 1000000 fields or fewer per GB of heap memory on data nodes, and account for overhead

DaveCTurner (approved):
I'm ok with this too although personally I'd wait for the stats APIs we discussed above before proceeding. I think we're going to end up fielding a bunch of questions about this guidance until those stats are available.

Your call ultimately.

original-brownbear:
@DaveCTurner thanks for b4f0e0f LGTM :)

@henningandersen could you take another look here too please? :) Thanks!

Leaf-Lin:
Hi, as this is just a doc change, are we able to merge this back into the 8.3.0?

henningandersen (commented):
Sorry for the delay here. I added a few more comments.

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -205,6 +211,22 @@ GET _cat/shards?v=true
 ----
 // TEST[setup:my_index]
 
+[discrete]
+[[field-count-recommendation]]
+==== Allow 1kB of heap per field per index on data nodes, plus overheads
+
+The exact resource usage of each mapped field depends on its type, but in many
+cases you should allow approximately 1kB of heap overhead per mapped field per
```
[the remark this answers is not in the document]
original-brownbear: Sure reworded :)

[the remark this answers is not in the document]
original-brownbear: Between the start of this PR and now I believe I almost halved the overhead per mapper through various fixes :D IMO we have a massive margin here already. 8.5G for 1k indices is complete overkill when the measured idle state heap usage of 10k Beats indices is under 4G now => I wouldn't be more pessimistic here :)

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -175,17 +175,23 @@ index                                 prirep shard store
...
 
-Some system indices for {enterprise-search-ref}/index.html[Enterprise Search]
-are nearly empty and rarely used. Due to their low overhead, you shouldn't
-count shards for these indices toward a node's shard limit.
+Where possible, the master node will deduplicate the mapping metadata across
+indices with identical mappings. Mapping metadata will be deduplicated across
+indices which are created from the same index template and
+<<explicit-mapping,do not use dynamic mapping updates>>.
+
+If your indices are are mostly created from a small number of templates and do
```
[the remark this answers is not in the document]
henningandersen: I think of something pretty simple and conservative. Perhaps just 1-2kb per field?

For dynamic mappings, it can make the impact of those very visible. Like you need a 1GB heap if you do not use them but a 30GB heap (or whatever the calculation says) if you use them. That can be a good motivation for a redesign?

Another case could be applications or search use cases using 100s of specific indices with specific mappings. They fall outside the above and we then leave them with no guidance. I think those cases will definitely also result in pretty small numbers anyway, but providing the guidance seems helpful.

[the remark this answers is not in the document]
DaveCTurner: I think it might make sense to defer this until https://github.com/elastic/elasticsearch/issues/86639 is addressed - IMO it's too complicated to explain here how dynamic mappings might work with the deduplication mechanism.

docs/reference/how-to/size-your-shards.asciidoc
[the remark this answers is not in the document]
henningandersen: It sounds good to me with those numbers though I am still worried about the massiveness of the change. It is probably mostly around the workload then, i.e., the old guidance included the ability to search and aggregate (to some extent), perhaps we can elaborate a bit on the 0.5GB added for workload to say that heavy workloads may require more (it is implicit today in the wording). 

Are all those improvements you mention in 8.3?

docs/reference/how-to/size-your-shards.asciidoc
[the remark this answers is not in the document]
henningandersen: I would set it up as a worst-case with no deduplication. I agree that anything in between is hard, but providing a no-sharing guidance should be relatively easy? It puts an upper boundary on the no-sharing required heap-size.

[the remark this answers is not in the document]
original-brownbear: > but providing a no-sharing guidance should be relatively easy? It puts an upper boundary on the no-sharing required heap-size.

But this totally depends on the size of the mappings used. Also, the effect of not sharing isn't as large as it is made out to be here I think. A compressed beats mapping for example is only *~6k*. So for 3k indices we're only talking about an 18M outright increase in the direct heap requirements for the master node. There's additional effects here from making stats APIs heavier without dedup, sending larger cluster states etc. but I find it very hard to capture these in a heap size number.

docs/reference/how-to/size-your-shards.asciidoc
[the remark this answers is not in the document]
original-brownbear: Sure, added a few words for the heavy workloads case now :)

> Are all those improvements you mention in 8.3?

Yes.

docs/reference/how-to/size-your-shards.asciidoc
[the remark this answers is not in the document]
henningandersen: We talked offline about this and it seems we could remove the deduplication assumption here, since it is in effect very little extra data on the masters and is more of a cpu concern than memory.

henningandersen:

original-brownbear:
@henningandersen @DaveCTurner how about [588bc87](https://github.com/elastic/elasticsearch/pull/86223/commits/588bc870eabf16c98afbed46d02f34e943a4e410) ?

henningandersen (approved):
LGTM, sorry for the back and forth on the deduplication piece.

DaveCTurner (approved):
LGTM2

original-brownbear:
Thanks all!

PR #86226, opened 2022-04-27
Deduplicate + intern field names in mappers and field caps responses

original-brownbear:
We already intern the type in field caps responses and interning the name
can save some more memory. Also, we can do the same for names in mappers
if which can save a little heap for mapper instances on data nodes
in case of nested field names (top level field names are already interned
by Jackson as far as I can tell).

I think this one will be quite diminished in importance by the other field caps fixes that landed lately but
seems like a safe enough change to maybe backport to 8.1 and 7.17 to address 
known issues in those versions to which we can't backport the larger behavior changes of field caps (see linked issue).
(the interning for mapper instance name fields could be left out of that backport, I just added it here since it seemed like
a freebie in this context)

### DIFF
--- a/docs/changelog/86226.yaml
+++ b/docs/changelog/86226.yaml
@@ -1,0 +1,5 @@
+pr: 86226
+summary: Deduplicate + intern field names in mappers and field caps responses
+area: Search
+type: bug
+issues: []
--- a/server/src/main/java/org/elasticsearch/action/fieldcaps/IndexFieldCapabilities.java
+++ b/server/src/main/java/org/elasticsearch/action/fieldcaps/IndexFieldCapabilities.java
@@ -13,6 +13,7 @@ import org.elasticsearch.common.io.stream.StreamInput;
 import org.elasticsearch.common.io.stream.StreamOutput;
 import org.elasticsearch.common.io.stream.Writeable;
 import org.elasticsearch.common.util.StringLiteralDeduplicator;
+import org.elasticsearch.index.mapper.Mapper;
 import org.elasticsearch.index.mapper.TimeSeriesParams;
 
 import java.io.IOException;
@@ -65,7 +66,6 @@ public class IndexFieldCapabilities implements Writeable {
         TimeSeriesParams.MetricType metricType,
         Map<String, String> meta
     ) {
-
         this.name = name;
         this.type = type;
         this.isMetadatafield = isMetadatafield;
@@ -77,7 +77,7 @@ public class IndexFieldCapabilities implements Writeable {
     }
 
     IndexFieldCapabilities(StreamInput in) throws IOException {
-        this.name = in.readString();
+        this.name = Mapper.internFieldName(in.readString());
         this.type = typeStringDeduplicator.deduplicate(in.readString());
         this.isMetadatafield = in.readBoolean();
         this.isSearchable = in.readBoolean();
--- a/server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
@@ -1446,5 +1446,4 @@ public abstract class FieldMapper extends Mapper implements Cloneable {
             return indexCreatedVersion.onOrAfter(minimumCompatibilityVersion);
         }
     }
-
 }
--- a/server/src/main/java/org/elasticsearch/index/mapper/MappedFieldType.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/MappedFieldType.java
@@ -74,7 +74,7 @@ public abstract class MappedFieldType {
         TextSearchInfo textSearchInfo,
         Map<String, String> meta
     ) {
-        this.name = Objects.requireNonNull(name);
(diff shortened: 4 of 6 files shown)


dnhatn:
Thanks, Armin. I am not sure if we should intern field names of indices from remote clusters. Perhaps just have a separate CHM (up to 1000 entries) to duplicate the field names of field-caps responses without interning? WDYT?

javanna:
I am not in favour of making this change without concrete measurements of what it improves, as well as tests that make sure that it won't get lost over time.

original-brownbear:
@javanna as far as mappers go the improvement is pretty straight forward. Currently we duplicate field name strings when parsing. Practically for 13k Beats indices on a data node (sorry for the odd number, that's what I have a heap dump for right now :)) this comes out to a whopping 1.2G of heap wasted for duplicate field names so this one is simple (IMO). Not sure how we'd test for this though. I guess we could have assertions on the key strings in the map in `MappingLookup` but not sure it's worth the noise?

image

For field caps it's more controversial admittedly. For the linked user issue, this change would likely fix them running OOM. That is for non 8.2 though. For 8.2+ we duplicate the field name string a lot less. I'm not so sure it's necessary here unless we want to backport to have a quick fix for older versions. It saves some memory and makes `org.elasticsearch.action.fieldcaps.TransportFieldCapabilitiesAction#innerMerge` a little faster via faster map lookups (not a huge effect). I guess overall, given how well the string deduplicator performs in benchmarking I think I'm a fan of this because it makes the memory usage more predictable but not too invested in this part :)

original-brownbear:
@dnhatn 

> I am not sure if we should intern field names of indices from remote clusters. Perhaps just have a separate CHM (up to 1000 entries) to duplicate the field names of field-caps responses without interning? WDYT?

Fair point. My thinking was this:

It's hard to guess the useful size of the map without interning. A beats mapping for example can have 4k fields. If we have a CHM with a capacity of 1k only and do not intern either, I could see cases where we evict at such a high rate that we save little memory while and pay for the CHM. You could make the map larger than 1k but it's just hard to tell how large you need it I guess?
Even for remote clusters though, we'll likely see the same strings over and over here (think beats mappings or any other solution) and I'd expect a lot of overlap across clusters in the average case. The deduplicator with interning performs well enough in benchmarking/profiling to justify the overhead and modern Java will eventually GC from the interned strings map.
I guess theoretically there could be cases where the interned string pool grows massively and becomes slow but in that scenario the field caps response also potentially blows up in size enormously anyway and we run into trouble that way anyway?

javanna:
@original-brownbear I see why the change itself is simple to make, and it looks like a great improvement too. Thanks for chasing this! 

I still think that we should brainstorm and figure out what the way forward is with performance improvements in general. These will be lost over time unless we have a plan to catch regressions or a way to test that such changes won't be accidentally reverted. It's because it's good changes that I care that they stick around over time and not just one-off. I also see how what I am asking requires a lot more thinking and work than simply making the change. Let's discuss this.

dnhatn:
> I guess theoretically there could be cases where the interned string pool grows massively and becomes slow but in that scenario the field caps response also potentially blows up in size enormously anyway and we run into trouble that way anyway?

@original-brownbear That was my concern. I am fine with this change, but I think we prefer a more isolated and predictable change in bug-fix versions. How about deduplicating type names within a single CCS request: https://github.com/elastic/elasticsearch/compare/8.1...dnhatn:81_deduplicate?expand=1. But let's merge this change if you prefer.

original-brownbear:
I opened https://github.com/elastic/elasticsearch/pull/86301 to pull the less controversial part of interning field names in mappers out of this one. Let's put a pass on the field caps part of this for a bit maybe, I'm very close to setting up the ability to manually benchmark this which would give us the ability to make more informed choices here, I'll report back soon if that's ok :)

dnhatn:
@original-brownbear @javanna As we have received some OOM reports on field-caps requests hitting many indices. I am in favor of merging this PR on 7.17. WDYT?

javanna:
@dnhatn you mean that you would backport #86301 which was extracted from this PR? Or is there more than we need to get in to decrease memory usage?

original-brownbear:
@dnhatn I'm not sure, you probably have a better handle on the severity of this issue than I do. This change definitely will save a lot of memory I think, but also runs the risk of slowing down reading these messages on the transport threads somewhat.
@javanna I think the issue here is that this PR is mostly helpful for older versions that don't have any of Nhat's more fundamental fixes that make it so that we don't have this many duplicate fieldname strings in the first place.

dnhatn:
@original-brownbear @javanna I've opened https://github.com/elastic/elasticsearch/pull/88238 as an alternative approach for this.

javanna:
@nhat now that #88238 is merged against 7.x and a part of this PR was merged to master, can this be closed or what is left to do here?

original-brownbear:
I think this can be closed, feel free to close, just leaving it open since you asked Nhat :)

javanna:
Closing then, thanks for the quick turn-around, we can always reopen if Nhat thinks otherwise ;)

PR #86228, opened 2022-04-27
Speed up map diffing computation for inserts

original-brownbear:
Found this in the many shards benchmarks. When adding a single (or a few elements) to a large map
but not removing any entries we needlessly iterate the `before` map which can be quite expensive
for very large maps (routing table with many indices, or in our case adding an additional datastream
to an already large map of DS).
We can detect whether or not we even need to look for deletes up-front by counting inserts trivially
which effectively makes it twice as fast to compute the map diff if there's only inserts and updates.

relates #77466


original-brownbear:
Jenkins run elasticsearch-ci/docs-check

joegallo:
I doubt it matters for performance, but for the sake of consistency, can you extend the same logic into the `ImmutableOpenIntMapDiff` code?

original-brownbear:
@joegallo 

> WDYT?

I went for this now, it was trivial enough and took 5 minutes, if it only helps one fringe use case it was worth it I guess :) Who knows when we'll finish up on the full removal of these maps.

@idegtiarenko 

> Is it possible to convert this to a static factory?

I just tried this, it's a surprisingly involved changed with how this stuff work (though worthwhile IMO). Maybe look into this in a follow-up? It seems like it would generally be nice to change the way the maps get created a little since it's not only less efficient than it could be but also quite confusing to read with the mutable fields being created in the parent constructor and then mutated in the implementations ... not a big fan but can't easily change it here I'm afraid :)

idegtiarenko (approved):

joegallo (approved):

original-brownbear:
Thanks Ievgen+Joe!

PR #86240, opened 2022-04-27
Add UNKNOWN status to health api if cluster state is unknown

### LINKED ISSUE #84792: Report UNKNOWN health status if the cluster state on the coordinating node is not fully recovered
We should report the health status as UNKNOWN if the cluster state on the coordinating node is not fully recovered (possibly with a reason message stating the node is not ready yet)

If the `instance-has-master` indicator reports RED the other indicators that rely on local cluster state should report UNKNOWN.

The order of statuses should be RED, YELLOW, UNKNOWN, GREEN
If any UNKNOWN is reported by an indicator, the indicator should advise retrying the health API call.

jbaiera:
This PR adds an unknown status to indicators when the cluster state is not yet recovered or if there is no master node.

The "has master node" indicator logic is left unchanged to signal poor health if a node is not part of a cluster, but all other indicators that rely on cluster state will report a generic UNKNOWN health response.

The UNKNOWN health status is the second lowest priority when merging health statuses. If any other indicators have YELLOW, or RED results, those will be chosen for the overall health in the result. If all other indicators are GREEN, the resulting status will be UNKNOWN.

The generic UNKNOWN health status respects the calculateDetails flag, and will add metadata to the result indicating the reason it could not determine the state of the indicator. Right now, these reasons are simply "has_master" and "cluster_state_recovered".

An example result from a data node that has no elected master:
```json
{
    "status": "red",
    "cluster_name": "runTask",
    "components": {
        "cluster_coordination": {
            "status": "red",
            "indicators": {
                "instance_has_master": {
                    "status": "red",
                    "summary": "Health coordinating instance does not have a master node."
                }
            }
        },
        "data": {
            "status": "unknown",
            "indicators": {
                "shards_availability": {
                    "status": "unknown",
                    "summary": "Health could not be determined. Try executing health api again."
                },
                "ilm": {
                    "status": "unknown",
                    "summary": "Health could not be determined. Try executing health api again."
                }
            }
        },
        "snapshot": {
            "status": "unknown",
            "indicators": {
                "repository_integrity": {
                    "status": "unknown",
                    "summary": "Health could not be determined. Try executing health api again."
                },
                "slm": {
                    "status": "unknown",
                    "summary": "Health could not be determined. Try executing health api again."
                }
            }
        }
    }
}
```

And the drilldown into data/shards_availability:
```json
{
    "cluster_name": "runTask",
    "components": {
        "data": {
            "indicators": {
                "shards_availability": {
                    "status": "unknown",
                    "summary": "Health could not be determined. Try executing health api again.",
                    "details": {
                        "reason": {
                            "has_master": false,
                            "cluster_state_recovered": false
                        }
                    }
                }
            }
        }
    }
}
```

Resolves #84792

### DIFF
--- a/server/src/main/java/org/elasticsearch/health/HealthIndicatorServiceBase.java
+++ b/server/src/main/java/org/elasticsearch/health/HealthIndicatorServiceBase.java
@@ -19,7 +19,7 @@ import java.util.Collections;
 
 public abstract class HealthIndicatorServiceBase implements HealthIndicatorService {
 
-    static final String COULD_NOT_DETERMINE_HEALTH = "Health could not be determined";
+    static final String COULD_NOT_DETERMINE_HEALTH = "Health could not be determined. Try executing health api again.";
 
     private final ClusterService clusterService;
 
--- a/server/src/main/java/org/elasticsearch/health/HealthIndicatorServiceBase.java
+++ b/server/src/main/java/org/elasticsearch/health/HealthIndicatorServiceBase.java
@@ -61,7 +61,7 @@ public abstract class HealthIndicatorServiceBase implements HealthIndicatorServi
         };
     }
 
-    private static boolean stateIsKnown(ClusterState clusterState) {
+    static boolean stateIsKnown(ClusterState clusterState) {
         return clusterState.getBlocks().hasGlobalBlock(GatewayService.STATE_NOT_RECOVERED_BLOCK) == false
             && clusterState.getBlocks().hasGlobalBlockWithId(NoMasterBlockService.NO_MASTER_BLOCK_ID) == false;
     }
--- a/server/src/main/java/org/elasticsearch/health/HealthIndicatorServiceBase.java
+++ b/server/src/main/java/org/elasticsearch/health/HealthIndicatorServiceBase.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.health;
+
+import org.elasticsearch.cluster.ClusterState;
+import org.elasticsearch.cluster.coordination.NoMasterBlockService;
+import org.elasticsearch.cluster.service.ClusterService;
+import org.elasticsearch.gateway.GatewayService;
+import org.elasticsearch.xcontent.ToXContent;
+import org.elasticsearch.xcontent.XContentBuilder;
+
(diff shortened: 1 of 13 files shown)


server/src/main/java/org/elasticsearch/health/HealthIndicatorServiceBase.java
```diff
@@ -0,0 +1,68 @@
...
+        if (stateIsKnown(clusterState)) {
+            return doCalculate(clusterState, calculateDetails);
+        } else {
+            return doCalculateUnknown(clusterState, calculateDetails);
+        }
+    }
+
+    protected abstract HealthIndicatorResult doCalculate(ClusterState clusterState, boolean calculateDetails);
+
+    protected HealthIndicatorResult doCalculateUnknown(ClusterState clusterState, boolean calculateDetails) {
```
andreidan, line 42: I believe we need another approach here. 

We have 2 use cases I believe:

1. This coordinating node has a block on the cluster state (which is what is currently tackled in this PR)

2. The indicator responsible with checking the master node (currently `instance-has-master` but soon it'll be the `master-is-stable` indicator) returned RED so some of the other indicators should return `UNKNOWN`. I believe we should have this ability to say "indicator 1 is RED so indicator 2,3,4 must be UNKNOWN now" (currently _all_ the other indicators should return `UNKNOWN` if `instance-has-master` returned RED as they rely on the cluster state, however in the future maybe the `network` indicator will be able to report some health information about this node's connections or such)

andreidan (commented):
Thanks Jimmy. I've left a suggestion on another use-case here which would influence the design a bit

server/src/main/java/org/elasticsearch/health/HealthIndicatorServiceBase.java
jbaiera: Are we sure we want to return an UNKNOWN in the case that master is not stable? If the master is unstable but we _are_ connected to a master node at this moment then it stands to reason that the other indicators should be able to produce their results right?

andreidan: That's a good question. The `master-is-stable` name is maybe not the best name for what it does. But it's a superset of `instance-has-master` and I believe in the case it returns `RED` ie. we don't have a master or we can't join a master, or even `YELLOW` (as that indicates "well, the master node is flapping and the cluster isn't reliably working") I don't think we can rely on the local cluster state to surely say there are no unassigned shards for e.g.

I believe we need a model that check both the cluster state blocks (like we have in this PR) but also a model that says "oh indicator X is RED, these other ones need to be UNKNOWN now"

I believe with the current implementation we can't express that. If we have _some_ cluster state without any blocks I can have `instance-has-master` returning RED (ie. no master in the cluster state) and at the same time `shards_availability` GREEN (which would be incorrect to indicate - this should be UNKNOWN)

jbaiera:
Closing this out and will open a new PR with new implementation.

PR #86242, opened 2022-04-27
Remove mention of "mapping type"

mrklaney:
"mapping types" were removed in v6.0

### DIFF
--- a/docs/reference/mapping/params/dynamic.asciidoc
+++ b/docs/reference/mapping/params/dynamic.asciidoc
@@ -42,7 +42,7 @@ GET my-index-000001/_mapping
 [[dynamic-inner-objects]]
 ==== Setting `dynamic` on inner objects
 <<object,Inner objects>> inherit the `dynamic` setting from their parent
-object or from the mapping type. In the following example, dynamic mapping is
+object. In the following example, dynamic mapping is
 disabled at the type level, so no new top-level fields will be added
 dynamically.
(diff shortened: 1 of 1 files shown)


docs/reference/mapping/params/dynamic.asciidoc
```diff
@@ -42,7 +42,7 @@ GET my-index-000001/_mapping
 [[dynamic-inner-objects]]
 ==== Setting `dynamic` on inner objects
 <<object,Inner objects>> inherit the `dynamic` setting from their parent
-object or from the mapping type. In the following example, dynamic mapping is
+object. In the following example, dynamic mapping is
 disabled at the type level, so no new top-level fields will be added
```
cbuescher, line 46: Maybe we should also not mention "type" here then. Maybe "root" level or "root mapping" level is more accurate now? Same goes for the mentions of "type" in the callouts after the example.

cbuescher (changes requested):
Hi @mrklaney, thanks for the suggested edits, clarification here makes sense to me. We should maybe go a step further and remove mentions of "type" in the following line and example callouts as well.

cbuescher:
Hi @mrklaney, just curious if you are still interested in merging this PR?

cbuescher (approved):
I'm going to merge this change in as is since its a step into removing the mentioning of "type" here.

Issue #86239, opened 2022-04-27
Remove uses of ImmutableOpen*Map


rjernst:
After https://github.com/elastic/elasticsearch/issues/84735, hppc usage in Elasticsearch is now isolated to `ImmutableOpenMap` and `ImmutableOpenIntMap`. This is a meta issue delineating the remaining uses of those classes that need to be removed.

- [ ] ImmutableOpenIntMap
  - [ ] server
    - [ ] o.e.Version
    - [ ] o.e.action.admin.cluster.reroute
    - [ ] o.e.action.admin.indices.shards
    - [ ] o.e.cluster.DiffableUtils
    - [ ] o.e.cluster.metadata.IndexMetadata
    - [ ] o.e.cluster.routing.IndexRoutingTable
  - [ ] server tests
    - [ ] o.e.action.admin.indices.shards
    - [ ] o.e.cluster.health
    - [ ] o.e.cluster.serialization
  - [ ] server integ tests
    - [ ] o.e.action.admin.indices.shards
    - [ ] o.e.cluster.routing
- [ ] ImmutableOpenMap
  - [ ] server
  - [ ] server tests
  - [ ] server integ tests
  - [ ] test framework
  - [ ] xpack core
  - [ ] data streams
  - [ ] autoscaling
  - [ ] ccr
  - [ ] deprecation
  - [ ] enrich
  - [ ] ilm tests
  - [ ] ML
  - [ ] QL
  - [ ] rollup
  - [ ] searchable snapshots
  - [ ] security
  - [ ] node shutdown
  - [ ] transforms
  - [ ] watcher

dakrone:
@rjernst do we still plan to keep some (most?) of these as immutable maps, or are we converting to mutable `HashMap`-like maps for these?

rjernst:
Yes I think they can and should remain immutable, just immutable Map instances with Map.of or Map.copyOf.

thecoop:
`ImmutableOpenMap` was *added* to the cluster/index metadata classes in #86673 for performance reasons. So we need to find a different collection type (maybe guava immutablemap?) if we want to remove `ImmutableOpenMap`

thecoop:
https://github.com/elastic/elasticsearch/pull/90006 caused a performance drop, needs to be thought about more

PR #86245, opened 2022-04-28
Remove ImmutableOpen*Map from Version

rjernst:
The Version class has two internal static maps for keeping track of the
known Version objects and quickly decoding from a version id into the
known Version object. This commit converts these maps to HashMaps.

relates #86239

### DIFF
--- a/server/src/main/java/org/elasticsearch/Version.java
+++ b/server/src/main/java/org/elasticsearch/Version.java
@@ -9,8 +9,6 @@
 package org.elasticsearch;
 
 import org.elasticsearch.common.Strings;
-import org.elasticsearch.common.collect.ImmutableOpenIntMap;
-import org.elasticsearch.common.collect.ImmutableOpenMap;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/elasticsearch/Version.java
```diff
@@ -158,8 +158,8 @@ public class Version implements Comparable<Version>, ToXContentFragment {
                 + "]";
         builder.put(V_EMPTY_ID, V_EMPTY);
         builderByString.put(V_EMPTY.toString(), V_EMPTY);
-        idToVersion = builder.build();
-        stringToVersion = builderByString.build();
+        idToVersion = Collections.unmodifiableMap(builder);
```
original-brownbear, line 161: Lets just do `Map.copyOf` for a smaller + faster map?

original-brownbear (approved):
LGTM lets use the faster immutable map though?

Issue #80117, opened 2021-11-01
Revisit and rationalise the Authentication class


ywangd:
We would like to revisit the `Authentication` class to make it provide better and easier to use interface to the consumers. 

#79809 enables run-as for all authentication schemes in addition to `realm` which is already supported. This adds more complexity to an `Authentication` object and how it should be used. For an example, building the role associated to an user now has to consider whether the user is the authenticated user or run-as user and the authentication scheme of the authenticated user. The existing `Authentication` class can answer all these question today. But its interface and internals are not aligned for these questions and the usage is error prone. For example, the `Authentication` object itself does not know whether the user has run-as. This information is kept by the `User` object. If the `User` object has an `authenticatedUser`, it's the run-as user. Otherwise it is not. But the information about realms for each `User` is directly kept in `Authentication`. Therefore, for full picture, the caller must check both `User` and `Authentication` realms and understands the nuance of their differenent combinations. 

- [x] Refactor CompositeRolesStore to separate roleDescriptors resolving (#80926)
- [ ] Move effective role descriptor checking from Resolver to CompositeRolesStore
- [x] https://github.com/elastic/elasticsearch/issues/81192
- [ ] Replace Authentication with AuthenticationContext in places other than CompositeRolesStore

tvernum:
My first draft of a suggested model is something more like this:
```
class AuthenticationContext {

    private Version version;

    private ResolvedUser effectiveUser;
    private ResolvedUser authenticatedUser;

    private AuthenticationInfo authInfo;

    boolean hasLookupUser() {
        return effectiveUser != authenticatedUser;
    }

}

class ResolvedUser {
    User user;
    RealmRef realm;

    boolean isServiceAccount() { .. }
    boolean isApiKey() { .. }

    boolean hasNamedRoles() { return isServiceAccount() == false && isApiKey() == false; }

    RoleReference roles() {
        if(hasNamedRoles()) {
            return NamedRoles(user.roles());
        } else {
            // ...
        }
    }

}
    
class AuthenticationInfo {
    AuthenticationType type;
    Map<String,Object> metadata;
}
```

I haven't put a lot of thought into the class names, but I think the overall structure would probably work.

The only question is whether to also use this as an opportunity to do something with `authorization_realms` and represent them in the model as well (I haven't done that here, but I would like to).

ywangd:
[...]

PR #86246, opened 2022-04-28
Internalise run-as user and simplify the regular User class

ywangd:
Today the run-as information is spread between User (authenticatedUser) and Authentication (lookup realm). They have to be configured consistently to work correctly. Previously there was also no inherent logic to ensure the consistency. Recent refacotring of Authentication class has made the situation better by favour Authentication creation with dedicated convenient methods over free-for-all constructors. #86206 is the ongoing PR that will finally remove public access of Authentication constructors. Now that the Authentication class is being tightly controlled, it makes possible to clean-up the User class. Specifically, the run-as information is already provided by the Authentication class, there is no need for the User class to also keep track of it. In fact, the way User class tracking the authenticating user with an inner user is less straightforward and not friendly to serialisation. Also, conceptually run-as is an information at Authentication level instead of User level.

This PR refactors the User class so that it no longer keeps track of run-as information so that there is a single consistent way to check whether an Authentication object is run-as. The essential changes are:

* Removes `User.isRunAs()` method
* Removes `User.authenticatedUser()` method and its backing instance variable
* Removes all User constructors that take `authenticatedUser` as an argument
* Adds a new private `RunAsUser` class inside `Authentication`

Note that this `RunAsUser` class is not really necessary in long term, the plan is to remove it in the later refactoring of Authentication class (where `Subject` variable and methods will become primary over the current `User`). It is added mostly to make the refactoring easier and reduce the change to how things work conceptually. This class is not exposed and there should be no need to use this class outside of `Authentication` itself. This tight scope should make it relatively easy to remove it later.

Relates: #86206
Relates: #80117


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/user/User.java
```diff
@@ -192,41 +143,6 @@ public final XContentBuilder toXContent(XContentBuilder builder, Params params)
...
-            return outerUser;
-        }
-    }
-
-    public static User readFrom(StreamInput input) throws IOException {
-        final boolean isInternalUser = input.readBoolean();
-        assert isInternalUser == false : "should always return false. Internal users should use the InternalUserSerializationHelper";
-        final String username = input.readString();
-        return partialReadFrom(username, input);
-    }
```
ywangd, line 216: This method is replaced by the new `AuthenticationSerializationHelper#readUserFrom`. The replacement is basically the old `InternalUserSerializationHelper.readFrom`.

Note that this method assert that the user cannot be internal. I don't think it is necessary at this method level. I have replaced it with assertions at the caller sites (there are only 2 of them).

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/GetUsersResponse.java
```diff
@@ -29,7 +30,9 @@ public GetUsersResponse(StreamInput in) throws IOException {
         } else {
             users = new User[size];
             for (int i = 0; i < size; i++) {
-                users[i] = User.readFrom(in);
+                final User user = Authentication.AuthenticationSerializationHelper.readUserFrom(in);
+                assert false == User.isInternal(user) : "should not get internal users";
```
ywangd, line 34: This is the new call site assertion metioned above.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
```diff
@@ -874,4 +886,127 @@ public enum AuthenticationType {
...
+            String[] roles = input.readStringArray();
+            Map<String, Object> metadata = input.readMap();
+            String fullName = input.readOptionalString();
+            String email = input.readOptionalString();
+            boolean enabled = input.readBoolean();
+            User outerUser = new User(username, roles, fullName, email, metadata, enabled);
+            boolean hasInnerUser = input.readBoolean();
+            if (hasInnerUser) {
+                User innerUser = readUserFrom(input);
+                assert false == User.isInternal(innerUser) : "authenticating user cannot be internal";
```
ywangd, line 993: This is the other call site assertion.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/user/UserTests.java
```diff
@@ -19,13 +19,5 @@ public void testUserToString() {
...
         user = new User("u1", new String[] { "r1", "r2" }, "user1", "user1@domain.com", Map.of("key", "val"), true);
         assertThat(user.toString(), is("User[username=u1,roles=[r1,r2],fullName=user1,email=user1@domain.com,metadata={key=val}]"));
-        user = new User("u1", new String[] { "r1" }, new User("u2", "r2", "r3"));
-        assertThat(
-            user.toString(),
-            is(
-                "User[username=u1,roles=[r1],fullName=null,email=null,metadata={},"
-                    + "authenticatedUser=[User[username=u2,roles=[r2,r3],fullName=null,email=null,metadata={}]]]"
-            )
-        );
```
ywangd, line 29: No need to test anymore because User no longer has inner user.

ywangd:
A large number of changes to test classes, e.g. `LoggingAuditTrailFilterTests`, are due to:
1. They randomise whether a user has inner user and ensure the effective user is picked up
2. They randomise authentication based on whether the randomised user has an inner user

For 1, the randomisation is simply removed.
For 2, I refactored the method to take two User arguments instead of one so that the inner user is provided as its own argument, e.g. previously `createAuthentication(User user, String realmName)` where `user` can have an inner user to indicate the Authentication should be created with run-as. This is changed to `createAuthentication(User user, @Nullable User authenticatingUser, String realmName)`. The call sites are also updated accordingly.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/audit/logfile/LoggingAuditTrailFilterTests.java
```diff
@@ -127,13 +128,7 @@ public void testPolicyDoesNotMatchNullValuesInEvent() throws Exception {
         final Settings.Builder settingsBuilder = Settings.builder().put(settings);
         // filter by username
         final List<String> filteredUsernames = randomNonEmptyListOfFilteredNames();
-        final List<User> filteredUsers = filteredUsernames.stream().map(u -> {
-            if (randomBoolean()) {
-                return new User(u);
-            } else {
-                return new User(new User(u), new User(UNFILTER_MARKER + randomAlphaOfLengthBetween(1, 4)));
-            }
-        }).collect(Collectors.toList());
```
ywangd, line 136: This is an example where it is no longer necessary to randomise whether User has an inner user.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/audit/logfile/LoggingAuditTrailFilterTests.java
```diff
@@ -1734,12 +1684,9 @@ public void testRolesFilter() throws Exception {
         final String[] unfilteredRoles = _unfilteredRoles.toArray(new String[0]);
         Authentication authentication;
         if (randomBoolean()) {
-            authentication = createAuthentication(
-                new User("user1", new String[] { "r1" }, new User("authUsername", new String[] { "r2" })),
-                "effectiveRealmName"
-            );
+            authentication = createAuthentication(new User("user1", "r1"), new User("authUsername", "r2"), "effectiveRealmName");
```
ywangd, line 1687: This is an example where a single User (with inner user) argument is split into 2 User arguments.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/AuthenticatorChainTests.java
```diff
@@ -95,8 +96,7 @@ public void init() {
         when(realms.getActiveRealms()).thenReturn(List.of(mock(Realm.class)));
         when(realms.getUnlicensedRealms()).thenReturn(List.of());
         final User user = new User(randomAlphaOfLength(8));
-        authentication = mock(Authentication.class);
-        when(authentication.getUser()).thenReturn(user);
+        authentication = AuthenticationTestHelper.builder().user(user).build(false);
```
ywangd, line 99: Mocking Authentication is no longer recommended. In future PPs, I plan to remove all mocking usage from tests and make Authentication final.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
```diff
@@ -241,11 +246,11 @@ public Authentication maybeRewriteForOlderVersion(Version olderVersion) {
      */
     public Authentication runAs(User runAs, @Nullable RealmRef lookupRealmRef) {
         Objects.requireNonNull(runAs);
-        assert false == runAs.isRunAs();
-        assert false == getUser().isRunAs();
+        assert false == runAs instanceof RunAsUser;
+        assert false == getUser() instanceof RunAsUser;
         assert AuthenticationType.REALM == getAuthenticationType() || AuthenticationType.API_KEY == getAuthenticationType();
```
tvernum, line 251: Should we also check that `runAs.getAuthenticationType() == AuthenticationType.REALM` ?

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/RBACEngineTests.java
```diff
@@ -143,33 +142,35 @@ public void testSameUserPermission() {
...
             : new AuthenticateRequestBuilder(mock(Client.class)).username(username).request();
         final String action = changePasswordRequest ? ChangePasswordAction.NAME : AuthenticateAction.NAME;
-        final Authentication authentication = mock(Authentication.class);
-        final Authentication.RealmRef authenticatedBy = mock(Authentication.RealmRef.class);
-        when(authentication.getUser()).thenReturn(user);
-        when(authentication.getAuthenticatedBy()).thenReturn(authenticatedBy);
+
         final String authenticationType = changePasswordRequest
             ? randomFrom(ReservedRealm.TYPE, NativeRealmSettings.TYPE)
             : randomAlphaOfLengthBetween(4, 12);
```
tvernum, line 156: I don't think we need this conditional do we?
This is used for the `authUser` - we don't care what realm type it has, we only care which realm `user` was `lookedUpBy`.

```suggestion
        final String authenticationType = randomAlphaOfLengthBetween(4, 12);
```

Though `authenticationRealmType` is probably a better name.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
```diff
@@ -874,4 +886,127 @@ public enum AuthenticationType {
...
+            if (SystemUser.is(user)) {
+                output.writeBoolean(true);
+                output.writeString(SystemUser.NAME);
+            } else if (XPackUser.is(user)) {
+                output.writeBoolean(true);
+                output.writeString(XPackUser.NAME);
+            } else if (XPackSecurityUser.is(user)) {
+                output.writeBoolean(true);
+                output.writeString(XPackSecurityUser.NAME);
+            } else if (SecurityProfileUser.is(user)) {
```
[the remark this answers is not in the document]
ywangd: Do we need BWC for this given the feature is still beta? I had assumed that we don't need it. For example, the profile index mappings changed in both 8.2 and 8.3 in a non-compatible way. It is unlikely things would work in a mixed cluster even without this new internal user. That said, maybe we should have phased the changes better so that they are BWC all the way. Unfortunately it was not the case. What do you think?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
ywangd: Not sure if I miss something, but `runAs` is a `User`, not an `Authentication`. So it does not have an `AuthenticationType`. Do you mean we should check `lookUpRealmRef` is not one of the sythetic realms? Unfortunately we cannot because of custom realms and we don't retrict their names or types.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/RBACEngineTests.java
ywangd: You are right. It does not matter. I inlined it with simply `randomAlphaOfLengthBetween(4, 12)`.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
tvernum: Sorry, yeah - too many classes with things in the wrong place.

Logically we ought to check that `runAs` is a realm based used, and not an Internal user or an API Key, or ServiceAccount.

Since the proposal is an `assert`, I don't think we need to worry about custom realms. ES doesn't run with assertions in production, and if it interferes with someone's development environment then it just serves as a reminder that they picked a terrible name & type for their realm.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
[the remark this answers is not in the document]
tvernum: But what happens when someone does a rolling upgrade from 7.17 to 8.7 ? Will the new internal user get passed around in cross-node requests and cause failures on the 7.17 nodes?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
ywangd: I added assertion for lookupRealmRef not having synthetic realm name or type. 

PS: I also plan to consolidate all the assertions from different Authentication methods into its own private method and call it inside the constructors. This method will also serve a good reference for the internal logic of authentication object.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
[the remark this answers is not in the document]
ywangd: If they call User Profile APIs on the 8.7 node, then yes, it is possible that they can see the failure on a 7.17 node. A BWC for this internal user will not fix the failure entirely. But one may argue that the error message could be slightly more friendy?

Instead of a mysterious serialisation failure (`user [_security_profile] is not an internal user`), it could be a 404 index_not_found (`no such index [_security]`) or 400 (`request body is required`) or some other errors. Now I actually don't know which one is less friendly. The serialisation failure is probably more mysterious but it fails consistently and mentions `profile` in the failure message, while the other option could be more unpredictable.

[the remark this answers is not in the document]
ywangd: Sorry I have to correct myself: the 400 or 404 error is what user can get on the REST layer. Since this talks about transport layer, I think the error should always be something along the line of action handler not found, which is arguably a more friend error message. I think this is enough to convince me to handle BWC for this user. I will raise a separate PR. Thanks for the discussion.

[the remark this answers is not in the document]
ywangd: I raised https://github.com/elastic/elasticsearch/pull/86345
Btw, the BWC with the above PR is better than I expected. As long as the request hits nodes with newer version first, it should just work because requests that actually got sent across nodes are basic index and search. Feature specific actions are all handled locally.

tvernum (approved):
LGTM

ywangd:
@elasticmachine update branch

PR #86248, opened 2022-04-28
Remove direct snakeyaml dependency from launcher tools

rjernst:
The launcher tools currently depend on snakeyaml for doing a rough
parsing of elasticsearch.yml to determine node roles for heap settings.
However, the launchers (and in the future, the server-cli) already have
the rest of server available on the classpath. This change removes
snakeyaml and replaces it with x-content.

relates #85758


grcevski (approved):
LGTM!

PR #86252, opened 2022-04-28
Upgrade to Netty 4.1.76

original-brownbear:
Staying up to date. This upgrade comes with two non-trivial
change. The thread local cache buffer cache has been disabled
for non-transport threads and the default pooled buffer chunk
size has been reduced to 4M from 16M.
The later change I think is without any effect for us since we
allocate buffers in 16k chunks on our end when doing outbound
writes and if anything should save us memory.
The disabling of thread-local buffer cache could have some effect
but is hardly measurable in practice. That combined with the fact
that we can switch back to the old behavior via the Netty system
property made me go with the new behavior here which does save us
some heap memory at hardly measurable CPU cost.


Tim-Brooks (approved):
> The thread local cache buffer cache has been disabled
for non-transport threads and the default pooled buffer chunk
size has been reduced to 4M from 16M.

We manually set all of the allocators we use for a max of 1MB I believe anyway.

NettyAllocator.java
```java
                    if (g1gcEnabled == false || g1gcRegionSizeIsKnown == false || g1gcRegionSizeInBytes >= (4 * 1024 * 1024)) {
                        // This combined with a 8192 page size = 1 MB chunk sizes
                        maxOrder = 7;
                    } else if (g1gcRegionSizeInBytes >= (2 * 1024 * 1024)) {
                        // This combined with a 8192 page size = 512 KB chunk sizes
                        maxOrder = 6;
                    } else {
                        // This combined with a 8192 page size = 256 KB chunk sizes
                        maxOrder = 5;
                    }
```

original-brownbear:
Thanks Tim!

PR #86253, opened 2022-04-28
speed up ip v4 parser

idegtiarenko:
Current implementation is using String#split that internally allocate
ArrayList that is later copied into array. It is possible to speed up
parser implementation by avoiding this extra allocations as we know
result array should have exactly 4 items in it.

```
Benchmark                         (ip)         Mode  Cnt    Score   Error  Units
InetAddressesBenchmark. current   192.168.0.1  avgt    5  146.564 ± 6.164  ns/op
InetAddressesBenchmark. new       192.168.0.1  avgt    5   29.605 ± 1.845  ns/op
```

This change is ported from a space time project focused on optimizing document parsing: https://github.com/elastic/elasticsearch/pull/85714

### DIFF
--- a/docs/changelog/86253.yaml
+++ b/docs/changelog/86253.yaml
@@ -1,0 +1,5 @@
+pr: 86253
+summary: Speed up ip v4 parser
+area: Infra/Core
+type: enhancement
+issues: []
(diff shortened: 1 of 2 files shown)


rjernst (changes requested):
The change looks good, but I'm not sure there is complete test coverage for the edge cases in InetAddressesTests. Can you make sure all of these exist like empty octet, wrong number of octets, too large number, etc?

idegtiarenko:
Hello @rjernst , yes this was already very well covered by tests: https://github.com/elastic/elasticsearch/blob/92e8ba2e744568389d128136d4f55929fbd8cdbe/server/src/test/java/org/elasticsearch/common/network/InetAddressesTests.java#L33-L94

rjernst (approved):
Great, thanks! LGTM

PR #86254, opened 2022-04-28
Reuse delegate() call result

idegtiarenko:
This change reuses the `delegate()` call result that is backed by 
`java.util.Deque#peek()` operation. This should simplify 
`expandDots()` and make it slightly cheaper to call.

Related to: https://github.com/elastic/elasticsearch/pull/85704


rjernst (approved):
LGTM

PR #81255, opened 2021-12-02
[s3-repository] Support IAM roles for Kubernetes service accounts

arteam:
There have been many requests to support `repository-s3` authentication via [IAM roles in Kubernetes service accounts](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/).

The AWS SDK is supposed to support them out of the box with the `aws-java-sdk-sts` library. Unfortunately, we can't use `WebIdentityTokenCredentialsProvider` from the SDK. It reads the token from `AWS_WEB_IDENTITY_TOKEN_FILE` environment variable which is usually mounted to `/var/run/secrets/eks.amazonaws.com/serviceaccount/token` and the S3 repository doesn't have the permissions to read it.  We don't want to hard-code a file permission for the S3 repository, because the location of `AWS_WEB_IDENTITY_TOKEN_FILE` can change at any time in the future and we would also generally prefer to restrict the ability of plugins to access things outside of their config directory.

To overcome this limitation, this change adds a custom `WebIdentityCredentials` provider that reads the service account from a symlink  to `AWS_WEB_IDENTITY_TOKEN_FILE` created in the repository's config directory. We expect the end user to create the symlink to indicate that they want to use service accounts for authentication like this: 

```bash
mkdir -p "${ES_PATH_CONF}/repository-s3"
ln -s $AWS_WEB_IDENTITY_TOKEN_FILE "${ES_PATH_CONF}/repository-s3/aws-web-identity-token-file"
```

Service accounts are checked and exchanged for session tokens by the AWS STS. To test the authentification flow, this change adds a test fixture which mocks  the `assume-role-with-web-identity` call to the service and returns a response with test credentials.

Fixes #52625

Issue #52625, opened 2020-02-21
Support for AWS IAM Kubernetes service account permissions


ErikvanDuren:
Amazon EKS recently supports IAM permissions for Kubernetes service accounts.

I would be nice to have support for service account permissions implemented in (at least) the S3 repository plugin so it is possible to create snapshots to an S3 repository without having to update  access keys and tokens on a regular basis.

DaveCTurner:
I had a quick check and this looks to be implemented by the `WebIdentityTokenCredentialsProvider` which is in the SDK version we're already using. The tricky bit will be to construct a test fixture that sets up an environment similar enough to EKS to show that it does integrate properly.

chonton:
What are the security concerns with using system properties, environment variables, or profiles to store credentials?  A web search reveals many opinions both for and against each method.  Should Elastic be enforcing a particular security policy or should Elastic make deployers aware of the pros/cons of each method?

In Kubernetes, mapped files holding the credentials seems to be one of the recommended practices at this time.  (Profile method)

chonton:
Elasticsearch team please chime in on which methods will be acceptable.

jim-barber-he:
For this to work this file needs changing: https://github.com/elastic/elasticsearch/blob/master/distribution/src/bin/elasticsearch-env-from-file

When I try to use IAM Roles for Service accounts it currently errors with:
```
ERROR: File ..data/token (target of symlink /var/run/secrets/eks.amazonaws.com/serviceaccount/token from AWS_WEB_IDENTITY_TOKEN_FILE) must have file permissions 400 or 600, but actually has: 640
```
The token gets created with the following permissions:
```
$ ls -l /var/run/secrets/eks.amazonaws.com/serviceaccount/..data/token
-rw-r----- 1 root 1001 1001 May 21 23:49 /var/run/secrets/eks.amazonaws.com/serviceaccount/..data/token
```
By default this file has permission 600 and owned by root:root, but when this is set:
```
podSecurityContext:
  fsGroup: 1001
  runAsUser: 1001
```
Then the group ownership of the token is changed to match `fsGroup` and the group read bit set so that the token can be read.
Without that change, the token wouldn't be readable, so the hard check for permission 400 or 600 enforced in the `elasticsearch-env-from-file` script isn't suitable for this case.

jim-barber-he:
[...]

PR #86255, opened 2022-04-28
[s3-repository] Support generating AWS role session name in case it's not provided

arteam:
The AWS SDK actually doesn't require the session name to be set and generates one if
it's not provided via the `AWS_ROLE_SESSION_NAME` environment variable.

See 
* https://github.com/elastic/elasticsearch/issues/52625#issuecomment-1105908937
* https://github.com/elastic/elasticsearch/pull/81255


tlrx (approved):
LGTM

PR #86256, opened 2022-04-28
Report size of bytesrefhash correctly after deserialization

hendrikmuhs:
After deserialization a bytesrefhash reported a size of `0` as the size wasn't set correctly.

relates #85826

Note:

 - the issue was missed in #85826, because of a test bug, see the fix for `assertAllIn`, it asserted equality of the _same_ value
 - technically this is a bug, but the PR is flagged as non-issue, because the code hasn't been part of a release


grcevski (approved):
LGTM!

PR #83576, opened 2022-02-07
Add new Threadpool for Cluster Coordination Activities

original-brownbear:
Using the generic pool for these all the activities surrounding cluster coordination
seems broken. They all get blocked on the same mutex in the `Coordinator` eventually
and run effectively sequentially.
Particularly in clusters with larger node counts this could lead to situations where
lots of generic threads are needlessly spun up only for the purpose of waiting on the mutex.
Since we at times also lock on the mutex in the coordinator on transport threads, it is
particularly unfortunate when there's lots of generic threads waiting on it.

=> fix this by using a single threaded pool for coordination work.
This also allows inspecting the queue size for these tasks for debugging purposes instead of having them hidden in the generic pool's queue.

PR #86259, opened 2022-04-28
Complete listener on rejection in FollowersChecker

DaveCTurner:
In #83576 we made it so that forking off the transport thread when
handling a follower check might be rejected if we're shutting down. In
that case we don't pass the rejection exception back out and instead
rely on the transport connection's close to notify the caller. We also
manipulate the transport channel directly which misses some exception
cases. This commit adjusts the implementation to fix these things.


DaveCTurner:
`>non-issue` as not really a big deal in production, we're shutting down so the caller gets an exception either way.

henningandersen (approved):
LGTM (when CI passes).

PR #67737, opened 2021-01-19
SQL: Fix elasticsearch-sql-cli under docker

palesz:
Calling `./bin/elasticsearch-sql-cli http://localhost:9200`under
docker failed with `ERROR: E is not a recognized option` error
message.

The root cause is that under docker the environmental variables are
automatically appended to the arguments of the sql cli. `"$@"` will
contain `-Evar=value` strings added by the `elasticsearch-env`
(line 122). This change uses the original argument list (saves it
before it is modified) when the sql cli is called.

Fixes #57744

PR #86264, opened 2022-04-28
Remove args capture in sql cli

rjernst:
The sql cli in Docker has a hack (#67737) to first capture the script arguments
before sourcing elasticsearch-env. This is because in Docker the env
script modified the arguments, adding in -E params based on the users
env settings. However, that Docker specific env logic was moved into
Java startup, so it is no longer necessary to capture the script args.

relates #85758


bpintea (approved):
LGTM

PR #86265, opened 2022-04-28
Remove errant use of jopt simple internals

rjernst:
The AutodetectProcessManager joins filter ids into a single string for
an internal request. Currently it uses an internal Strings helper class
from the joptsimple library. This commit changes it to use the JDK
provided String.join method.

relates #85758


droberts195 (approved):
LGTM - thanks

PR #86266, opened 2022-04-28
Relax jvm test throws

rjernst:
The jvm ergonomics tests have specific allowances of exception types
that might be thrown, based on the checked exceptions of the tested
methods. However, any exception is exceptional in tests, so the strict
checked exception list only makes modifying the checked exception list
of the production methods more cumbersom, needing to modify ever test.
This commit relaxes the throws clauses to allow for any Exception in
tests.

relates #85758


grcevski (approved):
LGTM!

PR #86268, opened 2022-04-28
[Cloud Security] refactor cloud security package policy name

CohenIdo:
Rename Clouse Security package name


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/store/ReservedRolesStore.java
```diff
@@ -806,7 +806,7 @@ public static RoleDescriptor kibanaSystemRoleDescriptor(String name) {
                     .build(),
                 // For src/dest indices of the Cloud Security Posture packages that ships a transform
                 RoleDescriptor.IndicesPrivileges.builder()
-                    .indices("logs-cis_kubernetes_benchmark.findings-*")
+                    .indices("logs-cloud_security_posture.findings*")
                     .privileges("read", "view_index_metadata")
                     .build(),
```
jportner, line 811: Just to clarify --

If any people have enabled the Cloud Security Posture app in 8.2 and they upgrade to 8.3, the kibana_system user will not be able to read any existing `logs-cis_kubernetes_benchmark.findings-*` indices. That's technically a breaking change, but it sounds like it's OK since the Cloud Security Posture is experimental and hidden behind a feature flag.

What's the upgrade path for users? Do they need to reinstall the Integration package for this, and then any old indices would just stick around until they're manually deleted? Would they lose any data / do they need to rename these indices before they upgrade?

It would be good to capture this in documentation somewhere. Maybe in the [Fleet release notes](https://www.elastic.co/guide/en/fleet/current/index.html)?

jportner (approved):

kfirpeled (changes requested):

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/store/ReservedRolesStore.java
kfirpeled: So the integration was not published yet. And the integration would only be published with a condition of kibana.version 8.3.0

So I think it covers that part. Unless I'm missing something

jportner: Makes sense, thanks!

kfirpeled (approved):

PR #86269, opened 2022-04-28
Add parameter to exclude indices in a snapshot from response

### LINKED ISSUE #82937: Add Parameter to not Return Index Name Lists in GET Snapshots API
The GET snapshots API currently returns the full list of all indices in a snapshot for each snapshot, regardless of the verbose setting.
This causes unmanageable response sizes in larger clusters and forces iterating through snapshots in small chunks if one wants to inspect a larger number of snapshots. E.g. a 50k indices cluster will see a million strings returned for just the index names with a common repository size of 200 snapshots.

I'll add a parameter to hide the indices list here to make the API more useful and less resource intensive to use for larger clusters.

relates #77466

original-brownbear:
Adds a parameter `index_names` to the get snapshots API so that users may exclude the potentially very long index name lists when listing out snapshots.

Not super optimized but good enough to fix the REST response in currently broken cases and generally make the API a little lighter to use for debugging common issues.

closes #82937


qa/smoke-test-http/src/test/java/org/elasticsearch/http/snapshots/RestGetSnapshotsIT.java
```diff
@@ -90,40 +90,45 @@ public void testSortOrder() throws Exception {
     }
 
     private void doTestSortOrder(String repoName, Collection<String> allSnapshotNames, SortOrder order) throws IOException {
-        final List<SnapshotInfo> defaultSorting = clusterAdmin().prepareGetSnapshots(repoName).setOrder(order).get().getSnapshots();
+        final boolean includeIndices = randomBoolean();
+        final List<SnapshotInfo> defaultSorting = clusterAdmin().prepareGetSnapshots(repoName)
+            .setOrder(order)
+            .setIndices(includeIndices)
```
idegtiarenko, line 96: In my opinion both `indices` and `includeIndices` sounds quiet generic and might be confusing.
Should they be called `includeIndexNames`?

idegtiarenko (approved):

docs/reference/snapshot-restore/apis/get-snapshot-api.asciidoc
```diff
@@ -99,6 +99,11 @@ version of Elasticsearch which took the snapshot, the start and end times of
 the snapshot, and the number of shards snapshotted. Defaults to `true`. If
 `false`, omits the additional information.
 
+`index_names`::
+(Optional, Boolean)
+If `true`, returns the list of snapshot names included in each snapshot in the response.
```
henningandersen, line 104:
```suggestion
If `true`, returns the list of index names included in each snapshot in the response.
```

henningandersen (commented):
Just a drive-by comment. Can we also update the PR description to mention what this PR changes?

original-brownbear:
Description added.

henningandersen (approved):
LGTM.

original-brownbear:
Thanks Ievgen + Henning!

PR #85940, opened 2022-04-15
Update license override list

AthenaEryma:
This commit updates the list of license expiration overrides as
requested by our compliance team.

PR #86272, opened 2022-04-28
Consistent response for starting basic license

### LINKED ISSUE #86244: Start basic license response may be incorrect
For `POST _license/start_basic` the API, the system decides whether a new basic license should be generated using the following logic:

https://github.com/elastic/elasticsearch/blob/3c15e4ebb1ae045d5cc690a0fd1e2f3c73173d39/x-pack/plugin/core/src/main/java/org/elasticsearch/license/StartBasicClusterTask.java#L101-L106

However, when building the response, the logic only takes account into old license type without checking maxNodes and expiryDate:

https://github.com/elastic/elasticsearch/blob/3c15e4ebb1ae045d5cc690a0fd1e2f3c73173d39/x-pack/plugin/core/src/main/java/org/elasticsearch/license/StartBasicClusterTask.java#L64-L65

This means it is possible that client gets an error response of 
```json
{
  "acknowledged": true,
  "basic_was_started": false,
  "error_message": "Operation failed: Current license is basic."
}
```
while the license is actually generated successfully at the backend. 

See also: https://github.com/elastic/elasticsearch/pull/85993#discussion_r859356499

n1v0lg:
The start basic license action returns an inconsistent response when a
basic license exists but a new basic license with different `maxNodes`
or `expiryDate` fields is started. In this case, the new license is
created but the response indicates that it was not. This PR makes the
response consistent.

Closes #86244


n1v0lg:
There doesn't seem to be a way that a client would ever run into this issue because both `expiryDate` and `maxNodes` are static constants, so this is very minor indeed. Still good to be consistent though. I'm happy to merge this or to close the issue, either way.   

Can also add a test; haven't yet since our current restrictions make it quite difficult to start a second basic license with `expiryDate` or `maxNodes` values that differ from the presets.

x-pack/plugin/core/src/main/java/org/elasticsearch/license/StartBasicClusterTask.java
```diff
@@ -33,7 +33,7 @@ public class StartBasicClusterTask extends ClusterStateUpdateTask {
     private final String description;
     private final ActionListener<PostStartBasicResponse> listener;
     private final Clock clock;
-    private AtomicReference<Map<String, String[]>> ackMessages = new AtomicReference<>(Collections.emptyMap());
+    private final AtomicReference<Map<String, String[]>> ackMessages = new AtomicReference<>(Collections.emptyMap());
```
n1v0lg, line 36: IntelliJ suggested and it looks sound

ywangd (commented):
> There doesn't seem to be a way that a client would ever run into this issue because both `expiryDate` and `maxNodes` are static constants, so this is very minor indeed. 

Indeed, it is very minor. But it is theoretically possible to run into this issue between ES versions since these variables can possibly change in a new version. Also the [license overrides](https://github.com/elastic/elasticsearch/pull/85940) is an example that existing license can have new attributes in a newer version. Though I don't think it applies to Basic licenses, it is another theorectical possibility. So I'd fix for consistency.

> Can also add a test; haven't yet since our current restrictions make it quite difficult to start a second basic license with `expiryDate` or `maxNodes` values that differ from the presets.

For unit test, can we just instanatiate or mock a `License` object for this purpose?

n1v0lg:
> For unit test, can we just instanatiate or mock a License object for this purpose?

Absolutely, will add one since we are going ahead with this change 👍

x-pack/plugin/core/src/test/java/org/elasticsearch/license/LicenseServiceTests.java
```diff
@@ -173,12 +262,7 @@ private void tryRegisterLicense(Settings baseSettings, License license, Consumer
             .put(DISCOVERY_TYPE_SETTING.getKey(), SINGLE_NODE_DISCOVERY_TYPE) // So we skip TLS checks
             .build();
 
-        final ClusterState clusterState = mock(ClusterState.class);
-        Mockito.when(clusterState.metadata()).thenReturn(Metadata.EMPTY_METADATA);
```
n1v0lg, line 177: We now also set the cluster name but that seems like a fair default and we get to re-use a bit of code

ywangd (approved):
LGTM

I don't quite understand why can't we just test `StartBasicClusterTask#execute` which should be much simpler and direct for the main code change. Your way of testing works, but it seems unnecessarily heavy for what we are really after.

n1v0lg:
@ywangd I went with including the test in `LicenceServiceTests` since it's consistent with the other tests already included in that test suite, and a "non-suprising" place for it to be. Testing `StartBasicClusterTask` would have made for a slightly easier setup but not by much (we'd still need the business of capturing the listener, generating or mocking an old license, etc).

n1v0lg:
@elasticmachine update branch

n1v0lg:
@elasticmachine update branch

n1v0lg:
@elasticmachine run elasticsearch-ci/part-2-fips again.....

PR #86274, opened 2022-04-28
Remove cli variant of SuppressForbidden

rjernst:
The cli lib has the SuppressForbidden annotation, but so does core,
which cli depends on. This commit removes the SuppressForbidden from
cli, in favor of the one from core.

relates #85758


grcevski (approved):
LGTM!

PR #86279, opened 2022-04-28
Move cli sysprops and envVars to execute parameter

rjernst:
The sysprops and envVars members of Command provide cli implementations
with information about the jvm process that is running. This is
convenient for runtime, but difficult for tests to mock because they
must subclass the cli class.

This commit adds a ProcessInfo record, and plumbs it through the
main and execute methods. The new record includes system properties,
environment variables and the working directory. By having this be a
single new parameter, additional information can be added in the future
without again needing to modify the method signatures.

relates #85758


ChrisHegarty (approved):
LGTM.

PR #86280, opened 2022-04-29
Better message in guessRootCauses

yan4105:
Just hoping to inspire better solutions, with respect to https://github.com/elastic/elasticsearch/issues/85127 and follow up discussion https://github.com/elastic/elasticsearch/pull/85774 where it seems a better error message for the users is needed, I suggest returning the method names that appear in the stack trace joined by “->”.

### DIFF
--- a/server/src/main/java/org/elasticsearch/ElasticsearchException.java
+++ b/server/src/main/java/org/elasticsearch/ElasticsearchException.java
@@ -635,7 +635,7 @@ public class ElasticsearchException extends RuntimeException implements ToXConte
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


jtibshirani:
I tagged the analytics/ geo team because it relates to your previous discussions. Feel free to remove or relabel.

not-napoleon:
I'm going to close this.  Stack traces don't make good error messages.  They tend to be long, sometimes very long, and are essentially useless to an end user.  For developers, we have the `error_trace=true` query parameter to get a full trace as needed.

PR #86217, opened 2022-04-27
Computation listener

idegtiarenko:
This pr introduces the listener that is called once all queued computations are complete.

PR #86283, opened 2022-04-29
Async allocator interface

idegtiarenko:
Wire listeners from create index operation into desired balance computation.
This pr ignores the other flows for now and uses a noop listener. 
They could be implemented in a followups.


server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
```diff
@@ -67,12 +73,12 @@ public void allocate(RoutingAllocation allocation) {
         desiredBalanceComputation.onNewInput(
             new DesiredBalanceInput(allocation.immutableClone(), new ArrayList<>(allocation.routingNodes().unassigned().ignored()))
         );
+        listener.onResponse(AcknowledgedResponse.TRUE);// TODO listener need to be passed to the above call
```
idegtiarenko, line 76: Could be done after https://github.com/elastic/elasticsearch/pull/86217 is merged

idegtiarenko: It looks like it is not enough to complete this listener once `desiredBalanceComputation` is complete as `processInput` might result in reroute that is async itself.

server/src/main/java/org/elasticsearch/cluster/metadata/MetadataCreateIndexService.java
```diff
@@ -285,13 +287,20 @@ public void createIndex(final CreateIndexClusterStateUpdateRequest request, fina
...
+
+        var future = new ListenableFuture<Void>();
+
         submitUnbatchedTask(
             "create-index [" + request.index() + "], cause [" + request.cause() + "]",
-            new AckedClusterStateUpdateTask(Priority.URGENT, request, listener) {
+            new AckedClusterStateUpdateTask(
+                Priority.URGENT,
+                request,
+                future.delegateFailure((delegate, response) -> future.addListener(listener.map(ignored -> response)))
```
DaveCTurner, line 298: Should this be `listener.delegateFailure`? We want to complete `listener` on a failure, we don't care about `future`.

test/framework/src/main/java/org/elasticsearch/test/ESIntegTestCase.java
```diff
@@ -2442,10 +2440,4 @@ public Settings onNodeStopped(String nodeName) throws Exception {
             }
         });
     }
-
-    protected static void awaitDesiredBalanceShardsAllocator() throws Exception {
-        if (internalCluster().getCurrentMasterNodeInstance(ShardsAllocator.class)instanceof DesiredBalanceShardsAllocator dbsa) {
-            assertBusy(() -> assertTrue(dbsa.isIdle()));
```
DaveCTurner, line 2448: I think this is the only caller of `isIdle` so we can remove it and the flag on which it depends.

server/src/main/java/org/elasticsearch/cluster/metadata/MetadataCreateIndexService.java
```diff
@@ -508,7 +522,8 @@ private ClusterState applyCreateIndexRequestWithV1Templates(
         final CreateIndexClusterStateUpdateRequest request,
         final boolean silent,
         final List<IndexTemplateMetadata> templates,
-        final BiConsumer<Metadata.Builder, IndexMetadata> metadataTransformer
+        final BiConsumer<Metadata.Builder, IndexMetadata> metadataTransformer,
+        final ActionListener<Void> listener
```
DaveCTurner, line 526: Could we call this `rerouteListener` (here and elsewhere in this PR) just to clarify what it's listening for?

DaveCTurner (commented):
Looks good, one change needed and a couple of other suggestions.

test/framework/src/main/java/org/elasticsearch/test/ESIntegTestCase.java
idegtiarenko: Unfortunately it is still used in `org.elasticsearch.cluster.coordination.Coordinator#hasIdleJoinValidationService` and `org.elasticsearch.cluster.coordination.JoinValidationServiceTests#testConcurrentBehaviour`

DaveCTurner: ah ok

DaveCTurner (approved):
LGTM

PR #86288, opened 2022-04-29
Update asm version in build tools for java19 support

breskeby:
This also fixes issues with using asm in functional tests where
spock dependencies brought in an older version of asm not compatible
with java 18


pugnascotia (approved):

ChrisHegarty (approved):
LGTM.

PR #86295, opened 2022-04-29
Add test utility methods for creating simple jar files

ChrisHegarty:
This PR adds test utility methods for creating simple jar files.

With this we can more easily write tests requiring jar archives for
the embedded class loader, and assert the loader's capabilities.

This PR simply adds the test utility methods for creating simple jar
files. Subsequent PR's will make use of these in test specific
scenarios.


rjernst (approved):
LGTM, one suggestion

PR #86297, opened 2022-04-29
Speed up parsing IndexMetadata from XContent

original-brownbear:
Leveraging #85732 : Cleaning up this code and more importantly helping out the master
node with the performance of parsing these a little as we need
to parse these during snapshot deletes which gets very expensive
in case of large index per snapshot counts.

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetadata.java
+++ b/server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetadata.java
@@ -8,7 +8,6 @@
 
 package org.elasticsearch.cluster.metadata;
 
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM! Pretty cool refactoring. Just for my curiosity, why did we check for Assertions.ENABLED in the past?

original-brownbear:
Thanks Nikola!

> Just for my curiosity, why did we check for Assertions.ENABLED in the past?

I have no idea I'm afraid. It seems completely redundant to what the JVM does anyway. I looked at the git history and these were added right when the assertions were first added so nothing helpful there either.

PR #86301, opened 2022-04-29
Intern field names in Mappers

original-brownbear:
Intern field names in mapper instances to save memory on nodes holding large
numbers of shards with similar mappings. In the example case of Beats mappings
this saves about 1G per 10k indices on a node which is a non-trivial win on
large frozen or warm nodes in particular.

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/mapper/Mapper.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/Mapper.java
@@ -10,6 +10,7 @@ package org.elasticsearch.index.mapper;
 
 import org.elasticsearch.Version;
 import org.elasticsearch.common.Strings;
+import org.elasticsearch.common.util.StringLiteralDeduplicator;
 import org.elasticsearch.xcontent.ToXContentFragment;
 
 import java.util.Map;
@@ -22,7 +23,7 @@ public abstract class Mapper implements ToXContentFragment, Iterable<Mapper> {
         protected final String name;
 
         protected Builder(String name) {
-            this.name = name;
+            this.name = internFieldName(name);
         }
 
         public String name() {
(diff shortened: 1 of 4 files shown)


dnhatn (approved):
LGTM.

original-brownbear:
Thanks Nhat!

server/src/main/java/org/elasticsearch/index/mapper/Mapper.java
```diff
@@ -79,4 +80,15 @@ public final String simpleName() {
...
     }
+
+    private static final StringLiteralDeduplicator fieldNameStringDeduplicator = new StringLiteralDeduplicator();
+
+    /**
+     * Interns the given field name string through a {@link StringLiteralDeduplicator}.
+     * @param fieldName field name to intern
+     * @return interned field name string
+     */
+    public static String internFieldName(String fieldName) {
```
javanna, line 91: does this need to be public?

javanna (commented):
Sorry for my super late review, I am catching up from a couple of weeks off. Is there a simple way to test that we don't duplicate strings here? Even if this is a small change, could we have a test that fails if the deduplication is removed?

server/src/main/java/org/elasticsearch/index/mapper/Mapper.java
original-brownbear: Right damn I extracted this from another change but in isolation here, this could've been `protected` indeed.

original-brownbear:
npnp + sorry for the late response as well :)

> could we have a test that fails if the deduplication is removed?

I could add one where we parse a mapping and then iterate all the mappers and check that `name() == name().intern()`, wdyt?

javanna:
that would work for me thanks!

PR #86303, opened 2022-04-29
Reroute completion listeners

idegtiarenko:
Introduce the listener that is called once queued computation and reroute are both complete.

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
@@ -18,6 +18,9 @@ import org.elasticsearch.common.Priority;
 import org.elasticsearch.threadpool.ThreadPool;
 
 import java.util.ArrayList;
+import java.util.Deque;
+import java.util.LinkedList;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
 
 /**
@@ -31,6 +34,11 @@ public class DesiredBalanceShardsAllocator implements ShardsAllocator {
     private final DesiredBalanceService desiredBalanceService;
     private volatile boolean pendingReroute;
 
+    private record DesiredBalancesListener(long index, Runnable listener) {}
+
+    private final AtomicLong indexGenerator = new AtomicLong(0);
+    private final Deque<DesiredBalancesListener> pendingListeners = new LinkedList<>();
+
     public DesiredBalanceShardsAllocator(
         ShardsAllocator delegateAllocator,
         ThreadPool threadPool,
@@ -43,8 +51,25 @@ public class DesiredBalanceShardsAllocator implements ShardsAllocator {
             protected void processInput(DesiredBalanceInput desiredBalanceInput) {
                 if (desiredBalanceService.updateDesiredBalanceAndReroute(desiredBalanceInput, this::isFresh)) {
                     pendingReroute = true;
-                    rerouteServiceSupplier.get()
-                        .reroute("desired balance changed", Priority.NORMAL, ActionListener.wrap(() -> pendingReroute = false));
+                    boolean isFreshInput = isFresh(desiredBalanceInput);
+                    rerouteServiceSupplier.get().reroute("desired balance changed", Priority.NORMAL, ActionListener.wrap(() -> {
+                        if (isFreshInput) {
+                            pendingReroute = false;
+                            triggerProcessedListeners(desiredBalanceInput.index());
+                        }
+                    }));
+                }
+            }
+
+            private void triggerProcessedListeners(long index) {
+                synchronized (pendingListeners) {
+                    while (true) {
+                        var trigger = pendingListeners.peekFirst();
+                        if (trigger == null || trigger.index > index) {
+                            break;
+                        }
+                        pendingListeners.pollFirst().listener.run();
+                    }
                 }
             }
 
@@ -64,8 +89,12 @@ public class DesiredBalanceShardsAllocator implements ShardsAllocator {
 
         // TODO must also capture any shards that the existing-shards allocators have allocated this pass, not just the ignored ones
 
+        var index = indexGenerator.incrementAndGet();
+        synchronized (pendingListeners) {
+            pendingListeners.addLast(new DesiredBalancesListener(index, () -> {/* TODO add async listener here */}));
+        }
         desiredBalanceComputation.onNewInput(
-            new DesiredBalanceInput(allocation.immutableClone(), new ArrayList<>(allocation.routingNodes().unassigned().ignored()))
+            new DesiredBalanceInput(index, allocation.immutableClone(), new ArrayList<>(allocation.routingNodes().unassigned().ignored()))
         );
 
         // TODO possibly add a bounded wait for the computation to complete?
--- a/server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
@@ -9,6 +9,7 @@
 package org.elasticsearch.cluster.routing.allocation.allocator;
 
 import org.elasticsearch.action.ActionListener;
+import org.elasticsearch.cluster.ClusterState;
 import org.elasticsearch.cluster.routing.RerouteService;
 import org.elasticsearch.cluster.routing.ShardRouting;
 import org.elasticsearch.cluster.routing.allocation.RoutingAllocation;
@@ -34,7 +35,7 @@ public class DesiredBalanceShardsAllocator implements ShardsAllocator {
     private final DesiredBalanceService desiredBalanceService;
     private volatile boolean pendingReroute;
 
-    private record DesiredBalancesListener(long index, Runnable listener) {}
+    private record DesiredBalancesListener(long index, ActionListener<Void> listener) {}
 
     private final AtomicLong indexGenerator = new AtomicLong(0);
     private final Deque<DesiredBalancesListener> pendingListeners = new LinkedList<>();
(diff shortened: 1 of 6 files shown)


DaveCTurner (commented):
Looks reasonable but I realised something I missed in our earlier conversation - see comment.

Also this deserves some changes to the `DesiredBalanceShardsAllocatorTests` I think.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
```diff
@@ -43,8 +51,25 @@ public DesiredBalanceShardsAllocator(
             protected void processInput(DesiredBalanceInput desiredBalanceInput) {
                 if (desiredBalanceService.updateDesiredBalanceAndReroute(desiredBalanceInput, this::isFresh)) {
                     pendingReroute = true;
-                    rerouteServiceSupplier.get()
-                        .reroute("desired balance changed", Priority.NORMAL, ActionListener.wrap(() -> pendingReroute = false));
+                    boolean isFreshInput = isFresh(desiredBalanceInput);
+                    rerouteServiceSupplier.get().reroute("desired balance changed", Priority.NORMAL, ActionListener.wrap(() -> {
```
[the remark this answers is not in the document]
idegtiarenko: Good point!

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
```diff
@@ -85,6 +86,21 @@ public void assertNoPendingReroute() {
...
+        private volatile boolean wasCalled = false;
+
+        @Override
+        public void onResponse(Void unused) {
+            wasCalled = true;
+        }
+
+        @Override
+        public void onFailure(Exception e) {
+
```
DaveCTurner, line 100: Nit: can assert no failures here too:

```suggestion
            fail("should not be called");
```

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
```diff
@@ -43,8 +52,44 @@ public DesiredBalanceShardsAllocator(
...
                 if (desiredBalanceService.updateDesiredBalanceAndReroute(desiredBalanceInput, this::isFresh)) {
                     pendingReroute = true;
-                    rerouteServiceSupplier.get()
-                        .reroute("desired balance changed", Priority.NORMAL, ActionListener.wrap(() -> pendingReroute = false));
+                    boolean isFreshInput = isFresh(desiredBalanceInput);
+                    var listener = new ActionListener<ClusterState>() {
+                        @Override
+                        public void onResponse(ClusterState clusterState) {
+                            if (isFreshInput) {
+                                pendingReroute = false;
```
DaveCTurner, line 60: We should clear this flag whether we're triggering listeners or not I think?

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceServiceTests.java
```diff
@@ -414,6 +414,7 @@ private static boolean executeAndReroute(
     ) {
         return desiredBalanceService.updateDesiredBalanceAndReroute(
             new DesiredBalanceInput(
+                0,
```
DaveCTurner, line 417: Hmm some of these tests execute this more than once, and in prod we would see a newer index each time. I think that means we're missing assertions somewhere to say that the index should be strictly increasing. I don't see an easy place to put them tho. Maybe `DesiredBalance` itself should record the index from which it was computed? I have a feeling this will be useful later on, at least in debugging. Up to you, we can leave it for now too.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
```diff
@@ -43,8 +52,44 @@ public DesiredBalanceShardsAllocator(
...
+                        }
+                        pendingListeners.pollFirst().listener.onResponse(null);
+                    }
+                }
+            }
+
+            private void triggerAllListenersFailure(Exception e) {
+                synchronized (pendingListeners) {
+                    DesiredBalancesListener listener;
+                    while ((listener = pendingListeners.pollLast()) != null) {
```
DaveCTurner, line 90: Why complete them in reverse order here? I think we can use a regular `Queue<>` rather than a `Deque<>` if we don't need this.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
```diff
@@ -43,8 +52,44 @@ public DesiredBalanceShardsAllocator(
...
+            }
+
+            private void triggerProcessedListeners(long index) {
+                synchronized (pendingListeners) {
+                    while (true) {
+                        var trigger = pendingListeners.peekFirst();
+                        if (trigger == null || trigger.index > index) {
+                            break;
+                        }
+                        pendingListeners.pollFirst().listener.onResponse(null);
```
DaveCTurner, line 82: I would rather complete the listener outside of the `synchronized` block. I think it's ok to acquire and release the lock on each iteration instead.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
```diff
@@ -43,8 +52,44 @@ public DesiredBalanceShardsAllocator(
...
+                    }
+                }
+            }
+
+            private void triggerAllListenersFailure(Exception e) {
+                synchronized (pendingListeners) {
+                    DesiredBalancesListener listener;
+                    while ((listener = pendingListeners.pollLast()) != null) {
+                        listener.listener.onFailure(e);
```
DaveCTurner, line 91: Similarly here - I would rather complete the listener outside of the `synchronized` block.

DaveCTurner (commented):
Looks good, I left just small comments.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
DaveCTurner: Oh wait I see, the whole point is that we don't want to be considered idle in this case. The `pendingReroute` flag will go away soon I guess, it's only used in a few tests to implement an artificial wait that shouldn't be necessary.

DaveCTurner:
Sorry, one more thing, we don't appear to have any tests that test that listeners are _not_ called if the computation is interrupted.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
idegtiarenko: Correct. I am also going to remove it in a following pull requests where I will be fixing remaining ITs that are still depending on it.

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceServiceTests.java
idegtiarenko: I guess this could be asserted inside of `org.elasticsearch.cluster.routing.allocation.allocator.ContinuousComputation#onNewInput`

DaveCTurner: Yeah I wondered about that but `ContinuousComputation` doesn't know that the computation object has a notion of "index".

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
```diff
@@ -57,15 +102,24 @@ public String toString() {
...
         assert MasterService.isMasterUpdateThread() || Thread.currentThread().getName().startsWith("TEST-")
             : Thread.currentThread().getName();
         // assert allocation.debugDecision() == false; set to true when called via the reroute API
         assert allocation.ignoreDisable() == false;
 
         // TODO must also capture any shards that the existing-shards allocators have allocated this pass, not just the ignored ones
 
+        var index = indexGenerator.incrementAndGet();
+        synchronized (pendingListeners) {
+            pendingListeners.addLast(new DesiredBalancesListener(index, listener));
```
DaveCTurner, line 119: Do we need to preserve the thread context of this listener? That's generally a problem with collecting listeners and then completing them all later. But I guess we're in system context here and we also complete them all in system context so maybe not. We should be able to assert that we're in system context tho (with some test adjustments because the tests won't be running this in system context I think)

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
```diff
@@ -226,4 +230,151 @@ public void afterPrimariesBeforeReplicas(RoutingAllocation allocation) {}
...
+            public ShardAllocationDecision decideShardAllocation(ShardRouting shard, RoutingAllocation allocation) {
+                throw new AssertionError("only used for allocation explain");
+            }
+        }, threadPool, () -> rerouteService);
+
+        var discoveryNode = createDiscoveryNode();
+        var index1 = createIndex(UUIDs.randomBase64UUID());
+        var index2 = createIndex(UUIDs.randomBase64UUID());
+
+        var listener = ActionListener.<Void>wrap(() -> {
```
DaveCTurner, line 277: Slight preference for failing the test if this listener gets an exception.

DaveCTurner (approved):
Looks good, one small comment but no need for another review.

PR #86304, opened 2022-04-29
Fix using FilterOutputStream without overriding bulk write

original-brownbear:
Fix using the filter output stream without overriding the bulk write.
The usage in `directFieldAsBase64` seems like a serious performance bug 
(marking this as >bug because it can cause serious slowness in async search I believe)
since the stream is used to write a potentially larger response here.
Had to move the no-close method to the core streams class to be able to use the 
existing solution in the x-content lib.

I also removed the `BlobOutputStream` that used to contain the same
fix now added to the no-close stream after realizing the class is pointless
to begin with to cut down on our usage of `FilterOutputStream` where the bulk
write fix is needed.

### DIFF
--- a/docs/changelog/86304.yaml
+++ b/docs/changelog/86304.yaml
@@ -1,0 +1,5 @@
+pr: 86304
+summary: Fix using `FilterOutputStream` without overriding bulk write
+area: Infra/Core
+type: bug
+issues: []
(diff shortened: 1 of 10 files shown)


grcevski (approved):
LGTM!

original-brownbear:
Thanks Nikola!

Issue #82255, opened 2022-01-05
[ML] Improve performance of job stats searches for data counts, model size stats and timing stats


droberts195:
When anomaly detection job stats are retrieved using the get job stats endpoint there seems to be a chance that a huge amount of CPU can be needed to determine authorization for the indices involved.

For example, from hot threads:

```
   100.0% [cpu=67.5%, other=32.5%] (500ms out of 500ms) cpu usage by thread 'elasticsearch[instance-0000000117][search][T#5]'
     2/10 snapshots sharing following 79 elements
       org.elasticsearch.xpack.core.security.authz.permission.IndicesPermission$Group.lambda$buildIndexMatcherPredicateForAction$0(IndicesPermission.java:565)
       org.elasticsearch.xpack.core.security.authz.permission.IndicesPermission$Group$$Lambda$7208/0x0000000801d679c0.test(Unknown Source)
       org.elasticsearch.xpack.security.authz.RBACEngine.resolveAuthorizedIndicesFromRole(RBACEngine.java:687)
       org.elasticsearch.xpack.security.authz.RBACEngine.loadAuthorizedIndices(RBACEngine.java:476)
       org.elasticsearch.xpack.security.authz.AuthorizationService.lambda$authorizeAction$7(AuthorizationService.java:411)
       org.elasticsearch.xpack.security.authz.AuthorizationService$$Lambda$7132/0x0000000801d30bd8.getAsync(Unknown Source)
       org.elasticsearch.xpack.security.authz.AuthorizationService$CachingAsyncSupplier.getAsync(AuthorizationService.java:1012)
       org.elasticsearch.xpack.security.authz.AuthorizationService.lambda$authorizeAction$10(AuthorizationService.java:426)
       org.elasticsearch.xpack.security.authz.AuthorizationService$$Lambda$7133/0x0000000801d30e00.getAsync(Unknown Source)
       org.elasticsearch.xpack.security.authz.AuthorizationService$CachingAsyncSupplier.getAsync(AuthorizationService.java:1012)
       org.elasticsearch.xpack.security.authz.RBACEngine.authorizeIndexAction(RBACEngine.java:343)
       org.elasticsearch.xpack.security.authz.AuthorizationService.authorizeAction(AuthorizationService.java:443)
       org.elasticsearch.xpack.security.authz.AuthorizationService.maybeAuthorizeRunAs(AuthorizationService.java:371)
[...]

PR #86305, opened 2022-04-29
[ML] improve reliability of job stats in larger clusters

benwtrent:
When gather job stats for closed jobs, we may be inadvertently executing on a transport thread. Typically, this is acceptable. But, when there are many jobs and many indices, this has a cascading effect and may cause the cluster to enter a troubling state.

This is main due to how slow security checks can be for search requests when the cluster has many indices. 

To alleviate, gathering information about closed jobs is forked to the ML utility thread pool

related: https://github.com/elastic/elasticsearch/issues/82255

### DIFF
--- a/docs/changelog/86305.yaml
+++ b/docs/changelog/86305.yaml
@@ -1,0 +1,5 @@
+pr: 86305
+summary: Improve reliability of job stats in larger clusters
+area: Machine Learning
+type: bug
+issues: []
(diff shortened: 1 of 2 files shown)


benwtrent:
@droberts195 I did some tests locally printing out the thread names on every call. 

Indeed we would be on a transport thread before this change. Meaning we would spin up N requests from that thread pool. Those executions would then occur on the search pool (eventually). 

In the call back, we are usually in the search pool, so I decided to fork that as well when gathering data stats (especially since in 7.17 its still multiple searches correct?).

Project Loom cannot come fast enough...

droberts195 (approved):
LGTM - thanks for doing this

benwtrent:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|7.17|[](https://github.com/elastic/elasticsearch/pull/86310)|

### Questions ?

PR #86306, opened 2022-04-29
Remove ImmutableOpenIntMap from ClusterStateHealthTests

rjernst:
These tests construct a map of shard to allocation ids. But they only do
so get the pairs to put into the index metadata builder. This commit
converts it to use HashMap.

relates #86239


williamrandolph (approved):
LGTM!

PR #86308, opened 2022-04-29
Add UNKNOWN health api status

jbaiera:
Adds an `UNKNOWN` health api status that will be used when the health of an indicator cannot be reliably determined.

UNKNOWN status has a priority that is higher than GREEN but lower than both YELLOW and RED.

### DIFF
--- a/server/src/main/java/org/elasticsearch/health/HealthStatus.java
+++ b/server/src/main/java/org/elasticsearch/health/HealthStatus.java
@@ -18,8 +18,10 @@ import java.util.stream.Stream;
 
 public enum HealthStatus implements Writeable {
     GREEN((byte) 0),
-    YELLOW((byte) 1),
-    RED((byte) 2);
+
+    UNKNOWN((byte) 1),
+    YELLOW((byte) 2),
+    RED((byte) 3);
 
     private final byte value;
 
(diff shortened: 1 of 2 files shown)


jbaiera:
relates #84792

server/src/main/java/org/elasticsearch/health/HealthStatus.java
```diff
@@ -18,8 +18,9 @@
 
 public enum HealthStatus implements Writeable {
     GREEN((byte) 0),
-    YELLOW((byte) 1),
-    RED((byte) 2);
+    UNKNOWN((byte) 1),
```
masseyke, line 21: Should the `merge` method use UNKNOWN as the default instead of GREEN? It seems more accurate.

andreidan: This is a good point. IMO the `orElse` that sort of implies "default" is a bit artificial there (as we shouldn't be merging a list of empty statuses - it'd be good to have an assert that the list is not empty IMO)

andreidan (approved):
LGTM, thanks Jimmy

server/src/main/java/org/elasticsearch/health/HealthStatus.java
jbaiera: Decided to have this throw instead of return a default since I think GREEN and UNKNOWN are reasonably competitive for a default value. If there aren't any problems, then things are GREEN right? Or I guess we don't know of any problems, so UNKNOWN? Nah, throw an error. This shouldn't be happening.

masseyke (approved):
LGTM

PR #86314, opened 2022-04-30
Simplify DeterministicTaskQueue's threadpool

DaveCTurner:
The `DeterministicTaskQueue` allows to create threadpools which don't
actually create any threads, instead executing the submitted tasks
deterministically on the main test thread. Today these fake threadpools
still unnecessarily create all the executor services and other machinery
that a real threadpool needs. This commit removes all that unnecessary
instantiation.


rjernst (approved):
LGTM

PR #86316, opened 2022-04-30
Make embedded class loader MRJAR aware

ChrisHegarty:
This PR updates the embedded class loader to be MRJAR aware.

A number of the implementation dependencies that are loaded by the 
embedded class loader are MRJARs. The loader should search for classes
and resources in the versioned section before the root.

### DIFF
--- a/libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoader.java
+++ b/libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoader.java
@@ -14,8 +14,6 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.UncheckedIOException;
 import java.net.MalformedURLException;
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.security.AccessController;
--- a/libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoader.java
+++ b/libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoader.java
@@ -13,6 +13,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.UncheckedIOException;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.security.AccessController;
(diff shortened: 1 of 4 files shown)


libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoader.java
```diff
@@ -136,25 +142,148 @@ public Class<?> findClass(String name) throws ClassNotFoundException {
...
+                final String jarPrefix = providerPrefix + "/" + jar;
+                JarMeta jam;
+                if (isMultiRelease(parent, jarPrefix)) {
+                    jam = new JarMeta(jarPrefix, true);
+                } else {
+                    jam = new JarMeta(jarPrefix, false);
+                }
+                map.put(jam, codeSource(listingURL, jar));
+            }
+            return Collections.unmodifiableMap(map);
```
rjernst, line 250: nit: Is Map.copyOf better to avoid an extra indirection on every access?

libs/core/src/test/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoaderTests.java
```diff
@@ -0,0 +1,459 @@
...
+        testResourcesVersioned(false, 0, 17);
+        testResourcesVersioned(false, 0, 11, 10, 9, 8);
+    }
+
+    /*
+     * Tests that the specific, 9, version of a resource is located, when the multi-release attribute
+     * is present and the versioned entry is less than or equal to the runtime version.
+     */
+    public void testResourcesVersioned9() throws Exception {
+        assumeTrue("JDK version not greater than or equal to 9", Runtime.version().feature() >= 9);
```
rjernst, line 238: Do we need these assumes since the build is on jdk 17+?

rjernst (approved):
LGTM, thanks for the extensive tests! I just left a couple very pedantic nits.

test/framework/src/main/java/org/elasticsearch/test/jar/JarUtils.java
```diff
@@ -69,6 +72,19 @@ public static void createJarWithEntries(Path jarfile, Map<String, byte[]> entrie
...
+    /**
+     * Creates a jar file with the given entries.
+     *
+     * @param jarfile the jar file path
+     * @param entries map of entries to add; jar entry name to String contents
+     * @param charset the charset used to convert the entry values to bytes
+     * @throws IOException if an I/O error occurs
+     */
+    public static void createJarWithEntries(Path jarfile, Map<String, String> entries, Charset charset) throws IOException {
+        var map = entries.entrySet().stream().collect(toUnmodifiableMap(Map.Entry::getKey, v -> v.getValue().getBytes(charset)));
```
rjernst, line 84: When would we ever want to have non-utf8? Could we just have this method assume that and remove the Charset param?

libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoader.java
```diff
@@ -136,25 +142,148 @@ public Class<?> findClass(String name) throws ClassNotFoundException {
...
+        if (jarMeta.isMultiRelease) {
+            url = findVersionedResourceForPrefixOrNull(name, jarMeta.prefix());
+            if (url != null) {
+                return url;
+            }
+        }
+        return parent.getResource(jarMeta.prefix() + "/" + name);
+    }
+
+    URL findVersionedResourceForPrefixOrNull(String name, String prefix) {
```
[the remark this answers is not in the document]
ChrisHegarty: Done

libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoader.java
ChrisHegarty: Done.

libs/core/src/test/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoaderTests.java
ChrisHegarty: We probably done need them, but they serve mostly as a comment to the reader.

test/framework/src/main/java/org/elasticsearch/test/jar/JarUtils.java
ChrisHegarty: Done.

ChrisHegarty:
This change updates the embedded class loader to be MRJAR aware.

PR #86319, opened 2022-04-30
Use faster and cleaner map parsing loop in ObjectParser.parse

original-brownbear:
Much easier to use the new API and not have to worry about all the manual
checks on the field name that were dead code anyway because they'd only be
hit on malformed input bytes which Jackson would throw on before entering
our code.
Also, removes some `currentToken()` calls by reusing the token as those aren't free and
speeding this up in any way possible is nice for the stability of snapshot inspection APIs
in large clusters.

### DIFF
--- a/libs/x-content/src/main/java/org/elasticsearch/xcontent/ObjectParser.java
+++ b/libs/x-content/src/main/java/org/elasticsearch/xcontent/ObjectParser.java
@@ -275,9 +275,6 @@ public final class ObjectParser<Value, Context> extends AbstractObjectParser<Val
             }
         }
 
-        FieldParser fieldParser = null;
-        String currentFieldName = null;
-        XContentLocation currentPosition = null;
         final List<String[]> requiredFields = this.requiredFieldSets.isEmpty() ? null : new ArrayList<>(this.requiredFieldSets);
         final List<List<String>> exclusiveFields;
         if (exclusiveFieldSets.isEmpty()) {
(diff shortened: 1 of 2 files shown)


original-brownbear:

libs/x-content/src/main/java/org/elasticsearch/xcontent/ObjectParser.java
```diff
@@ -289,36 +286,32 @@ public Value parse(XContentParser parser, Value value, Context context) throws I
...
-                fieldParser = parsers.get(currentFieldName);
+        while ((currentFieldName = parser.nextFieldName()) != null) {
+            currentPosition = parser.getTokenLocation();
+            fieldParser = parsers.get(currentFieldName);
+            token = parser.nextToken();
+            if (fieldParser == null) {
+                unknownFieldParser.acceptUnknownField(this, currentFieldName, currentPosition, parser, value, context);
             } else {
-                if (currentFieldName == null) {
-                    throwNoFieldFound(parser);
```
grcevski, line 300: I'm wondering about parsing '{}' START_OBJECT END_OBJECT, would we throw no field found before and now successfully parse?

original-brownbear: I don't think so because because we'd never enter the loop because the `nextToken` call would return `END_OBJECT` right away right?

grcevski: Yeah, you are right. I wrote a little test to convince myself this morning, no behaviour change:

```
    public void testEmpty() throws IOException {
        class TestStruct {
        }

        ObjectParser<TestStruct, Void> objectParser = new ObjectParser<>("the_parser");
        TestStruct s = new TestStruct();

        {
            XContentParser parser = createParser(JsonXContent.jsonXContent, "{}");
            objectParser.parse(parser, s,null);
        }
    }
```

grcevski (approved):
LGTM!

original-brownbear:
Thanks Nikola!

PR #86324, opened 2022-05-01
Add module descriptor test matchers

ChrisHegarty:
This PR adds test-only matchers for various aspect of module descriptors.

These matcher are used in the modules branch for improved readability
of tests that assert the exports, opens, requires, and services of
module descriptors.


rjernst (approved):
LGTM

PR #86330, opened 2022-05-02
[Docs] Fix url for feature migration APIs

ywangd:
Both GET and POST version of the API should have a leading underscore.


lockewritesdocs (approved):
Good catch 🦅 Thanks for fixing this issue @ywangd!

williamrandolph (approved):
Thanks for fixing this!

PR #86331, opened 2022-05-02
Fix over-allocation of mounted indices on a cold/frozen node

arteam:
`ClusterServiceInfo` can have stale information about free space in disk because gets updated periodically. If we allocate a lot of searchable snapshots on a cold node, we can break disk thresholds and over allocate on a node where there's no disk space left.

Preserve the expected shard size in `ShardRouting` when starting a new shard in case it was recovered from a searchable snapshot. Cache them in `unaccountedSearchableSnapshotSizes` for `RoutingAllocation`, so `DiskThresholdDecider` can account for the searchable snapshot shards that haven't been synced with cluster service info yet.

Fixes https://github.com/elastic/elasticsearch/pull/85018


arteam:
@elasticmachine update branch

arteam:
@elasticmachine update branch

arteam:

server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDecider.java
```diff
@@ -144,6 +143,24 @@ public static long sizeOfRelocatingShards(
...
+            if (indexMetadata.isSearchableSnapshot() == false) {
+                continue;
+            }
+            if (reservedSpace.containsShardId(shard.shardId())) {
+                continue;
+            }
+            if (clusterInfo.getShardSize(shard) != null) {
+                continue;
+            }
+            totalSize += Math.max(shard.getExpectedShardSize(), 0L);
```
henningandersen, line 161: I think we need this in the INITIALIZING case too for searchable snapshots? (or rather use the `shard.getExpectedShardSize` value in place of the default `0L` there).

I wonder if we can make a test that fail in that case. I.e., we have X initializing shards, allowing us to go X-1 shards too high.

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
```diff
@@ -505,7 +502,7 @@ public ShardRouting moveToStarted() {
             null,
             null,
             allocationId,
-            UNAVAILABLE_EXPECTED_SHARD_SIZE
+            expectedShardSize
```
henningandersen, line 505: We talked about only keeping this for searchable snapshots, where the size is immutable.

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
```diff
@@ -748,7 +745,7 @@ public String shortSummary() {
         if (this.unassignedInfo != null) {
             sb.append(", ").append(unassignedInfo.toString());
         }
-        if (expectedShardSize != UNAVAILABLE_EXPECTED_SHARD_SIZE) {
+        if (expectedShardSize != UNAVAILABLE_EXPECTED_SHARD_SIZE && state != ShardRoutingState.STARTED) {
```
henningandersen, line 748: I find this counter intuitive. If there is a shard size, I think we'd want the short summary and toString to contain it?

henningandersen (commented):
This direction looks good. I left a couple of comments to address.

I'd like to also see a test in `DiskThresholdDeciderUnitTests`, validating that the extra information is taken into account.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDecider.java
fcofdez: Yes, since we only account for relocating shards for the `INITIALIZING` shards. I agree that it would be nice to add a test for that case.

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
arteam: I believe there was an integration state checking the cluster state that relied on the `toString` output and it was failing because of the changed format. I believe it's should be ok to remove the check after we made sure `expectedShardSize` is set only for searchable snapshots.

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
arteam: Done!

arteam: I pass `Metatada` to `createStarted` in order to figure whether a new shard is searchable snapshot

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
arteam: Fixed it!

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -105,6 +106,7 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
                             .put(INDEX_SOFT_DELETES_SETTING.getKey(), true)
                             .put(INDEX_STORE_STATS_REFRESH_INTERVAL_SETTING.getKey(), "0ms")
                             .put(DataTier.TIER_PREFERENCE_SETTING.getKey(), DataTier.DATA_HOT)
+                            .put(MergePolicyConfig.INDEX_MERGE_ENABLED, false)
```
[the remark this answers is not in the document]
arteam: Yes, I've added a comment explaining it :+1:

server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDecider.java
arteam: I've changed `0L` to `Math.max(routing.getExpectedShardSize(), 0L)` for the `INITIALIZING` case, too, but I couldn't come up with a test case showing that it makes any difference :(

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,6 +275,73 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
...
+            client().admin()
+                .cluster()
+                .preparePutRepository("repository")
+                .setType("mock")
+                .setSettings(Settings.builder().put("location", randomRepoPath()).build())
+        );
+        var mockRepository = (MockRepository) internalCluster().getCurrentMasterNodeInstance(RepositoriesService.class)
+            .repository("repository");
+        // Prevent searchable snapshot shards from quickly jumping from INITIALIZED to STARTED
+        mockRepository.setBlockOnceOnReadSnapshotInfoIfAlreadyBlocked();
```
fcofdez, line 303: Let's move this before we mount the snapshot, to avoid possible interactions while we take the snapshot.

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,6 +275,73 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
...
+
+        Map<String, Long> indicesStoresSizes = sizeOfShardsStores("index-*");
+        assertAcked(client().admin().indices().prepareDelete("index-*"));
+
+        String otherDataNode = internalCluster().startNode(
+            Settings.builder().put(NodeRoleSettings.NODE_ROLES_SETTING.getKey(), DiscoveryNodeRole.DATA_COLD_NODE_ROLE.roleName()).build()
+        );
+        ensureStableCluster(3);
+
+        String otherDataNodeId = internalCluster().getInstance(NodeEnvironment.class, otherDataNode).nodeId();
```
fcofdez, line 315: nit: I think it's better to rely on the cluster state for this:
```
var clusterState = client().admin().cluster().prepareState().get().getState();
var nodeId = clusterState.nodes().resolveNode(otherNadaNode).getId();
```

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,6 +275,73 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
...
+        internalCluster().startNode(
+            Settings.builder()
+                .putList(NodeRoleSettings.NODE_ROLES_SETTING.getKey(), DiscoveryNodeRole.DATA_HOT_NODE_ROLE.roleName())
+                .build()
+        );
+
+        InternalClusterInfoService masterInfoService = (InternalClusterInfoService) internalCluster().getCurrentMasterNodeInstance(
+            ClusterInfoService.class
+        );
+        ClusterInfoServiceUtils.refresh(masterInfoService);
```
fcofdez, line 289: do we need to refresh the cluster info at this point?

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,6 +275,73 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
         });
     }
 
+    public void testOvercommitInitializingSearchableSnapshotShards() throws Exception {
+        internalCluster().startMasterOnlyNode();
+        internalCluster().startNode(
```
fcofdez, line 280: nit: shorter version `internalCluster().startNode(onlyRole(DATA_HOT_NODE_ROLE));`

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -155,11 +147,67 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
...
+                    false,
+                    storage
+                ),
+                ActionListener.wrap(response -> mountLatch.countDown(), e -> mountLatch.countDown())
+            );
+        }
+        mountLatch.await();
+    }
+
+    public void testHighWatermarkCanBeExceededOnColdNode() throws Exception {
```
fcofdez, line 181: s/testHighWatermarkCanBeExceededOnColdNode/testHighWatermarkCanNotBeExceededOnColdNode/ :)

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,6 +275,73 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
...
+        getTestFileStore(otherDataNode).setTotalSpace(totalSpace);
+
+        logger.info("--> refreshing cluster info");
+        ClusterInfoServiceUtils.refresh(masterInfoService);
+        assertThat(
+            masterInfoService.getClusterInfo().getNodeMostAvailableDiskUsages().get(otherDataNodeId).getTotalBytes(),
+            equalTo(totalSpace)
+        );
+
+        mountIndices(indicesStoresSizes, "mounted-", "repository", "snapshot", FULL_COPY);
```
fcofdez, line 336: We should validate that we do not allow allocating more shards that go over the disk threshold, right?

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
```diff
@@ -484,18 +483,27 @@ public ShardRouting reinitializeReplicaShard() {
...
+        return moveToStarted(null);
+    }
+
     /**
      * Set the shards state to <code>STARTED</code>. The shards state must be
      * <code>INITIALIZING</code> or <code>RELOCATING</code>. Any relocation will be
      * canceled.
      */
-    public ShardRouting moveToStarted() {
+    public ShardRouting moveToStarted(@Nullable Metadata metadata) {
```
fcofdez, line 495: I think we should have an extra field within `ShardRouting` to tell whether or not this shard is a searchable snapshot, that should simplify things and allows us to have clearer assertions. wdyt?

fcofdez (commented):
Thanks for adding the extra test, the direction looks good. I left a few extra comments.

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,6 +275,73 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
...
+        assertBusy(
+            () -> assertEquals(
+                ClusterHealthStatus.RED,
+                client().admin().cluster().health(new ClusterHealthRequest()).actionGet().getStatus()
+            )
+        );
```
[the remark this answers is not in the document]
arteam: Added `mockRepository.unblock()` before checking the cluster health status. Did you men that?

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
arteam: I believe we already do. The `createSnapshot("repository", "snapshot", nbIndices)` line is after the `setBlockOnceOnReadSnapshotInfoIfAlreadyBlocked` call

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
arteam: Done!

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
arteam: Exactly! We check that we can't allocate N shards when there's a space only for N-1 shards

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
fcofdez: But we already call just before the assertion, right?
```
        ClusterInfoServiceUtils.refresh(masterInfoService);
        assertThat(
            masterInfoService.getClusterInfo().getNodeMostAvailableDiskUsages().get(otherDataNodeId).getTotalBytes(),
            equalTo(totalSpace)
        );
```

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
fcofdez: I meant to move it before we call `mountIndices(indicesStoresSizes, "mounted-", "repository", "snapshot", FULL_COPY);`

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
fcofdez: Can we change the test to show that we indeed allocate `N-1` shards and the extra shards is not allocated?

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
fcofdez: I think it would add clarity and simplify this where we now have more API surface, i.e. why do we have 2 `moveToStarted` methods? additionally we could add more assertions to ensure that we only populate the expected shard size when it's needed.

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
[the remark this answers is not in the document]
fcofdez: My idea was:
1. Unblock the repo
2. Check that expected shards move to started

I feel like asserting that the cluster is RED is a bit coarse.

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
arteam: Right, we need to do `refresh` only once :+1:

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
arteam: WDYT about addressing this in a separate PR? `ShardRouting`'s constructor gets called from a lot of places, so adding a new field would add a significant amount of changes in the classes unrelated to the PR.

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
arteam: Moved

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
henningandersen: I would also like to get rid of the extra `moveToStarted` variant. Being lenient and assuming null in this central place seems dangerous.

But I do see the problems in doing so, at least in the test code, where we often do not have `Metadata` nor `IndexMetadata`. I wonder if we could fix this by instead letting `moveToStarted` take `expectedShardSize` as parameter. We can then let `AllocationService derive the `expectedShardSize` from the metadata it already has available.

I think this would help separate responsibilities too.

All tests that do not care can then pass in `UNAVAILABLE_EXPECTED_SHARD_SIZE`.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDecider.java
```diff
@@ -126,7 +127,11 @@ public static long sizeOfRelocatingShards(
 
         // Where reserved space is unavailable (e.g. stats are out-of-sync) compute a conservative estimate for initialising shards
         for (ShardRouting routing : node.shardsWithState(ShardRoutingState.INITIALIZING)) {
-            if (routing.relocatingNodeId() == null) {
+            IndexMetadata indexMetadata = metadata != null ? metadata.index(routing.index()) : null;
```
henningandersen, line 130: I think we can use `getIndexSafe` as in `getExpectedShardSize` and avoid the null checking? I think we should always have `metadata` here.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDecider.java
```diff
@@ -126,7 +127,11 @@ public static long sizeOfRelocatingShards(
...
+            if (indexMetadata == null) {
+                continue;
+            }
+            if (routing.relocatingNodeId() == null && indexMetadata.isSearchableSnapshot() == false) {
                 // in practice the only initializing-but-not-relocating shards with a nonzero expected shard size will be ones created
```
henningandersen, line 135:
```suggestion
                // in practice the only initializing-but-not-relocating non-searchable-snapshot shards with a nonzero expected shard size will be ones created
```

server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDecider.java
```diff
@@ -135,13 +140,37 @@ public static long sizeOfRelocatingShards(
...
+                    null,
+                    metadata,
+                    routingTable
+                );
+            }
+        }
+
+        // Count the STARTED searchable snapshot shards which are unaccounted in the cluster
+        for (ShardRouting shard : node.shardsWithState(ShardRoutingState.STARTED)) {
+            IndexMetadata indexMetadata = metadata.index(shard.index());
```
henningandersen, line 160: Use `getIndexSafe` and avoid the null check here?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDecider.java
```diff
@@ -135,13 +140,37 @@ public static long sizeOfRelocatingShards(
...
+            }
+            if (indexMetadata.isSearchableSnapshot() == false) {
+                continue;
+            }
+            if (reservedSpace.containsShardId(shard.shardId())) {
+                continue;
+            }
+            if (clusterInfo.getShardSize(shard) != null) {
+                continue;
             }
```
henningandersen, line 172: Can we make this a single if with `&&` conditions instead? It reads slightly verbose and backwards now. (I do know the one above is similar, but anyway I would prefer that for this new new block of code).

server/src/test/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDeciderUnitTests.java
```diff
@@ -471,16 +473,95 @@ public void testShardSizeAndRelocatingSize() {
...
+                            .put(INDEX_STORE_TYPE_SETTING.getKey(), SEARCHABLE_SNAPSHOT_STORE_TYPE)
+                    )
+                    .numberOfShards(3)
+                    .numberOfReplicas(1)
+            )
+            .build();
+
+        Index index = new Index("test", "1234");
+        String nodeId = "node1";
+        ShardRouting shard1 = ShardRoutingHelper.moveToStarted(
```
henningandersen, line 499: Can we instead create the 3 shards in a loop? Seems overly verbose spelled out here.

server/src/test/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDeciderUnitTests.java
```diff
@@ -471,16 +473,95 @@ public void testShardSizeAndRelocatingSize() {
...
+        );
+        ShardRouting shard3 = ShardRoutingHelper.moveToStarted(
+            ShardRoutingHelper.initialize(
+                ShardRouting.newUnassigned(
+                    new ShardId(index, 3),
+                    false,
+                    PeerRecoverySource.INSTANCE,
+                    new UnassignedInfo(UnassignedInfo.Reason.INDEX_CREATED, "foo")
+                ),
+                nodeId,
```
henningandersen, line 533: Can we add one additional shard that is added to another node? Just for checking that it does not impact it.

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,6 +271,97 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
...
+
+        var mockRepository = (MockRepository) internalCluster().getCurrentMasterNodeInstance(RepositoriesService.class)
+            .repository(repository);
+        // Prevent searchable snapshot shards from quickly jumping from INITIALIZED to STARTED
+        mockRepository.setBlockOnceOnReadSnapshotInfoIfAlreadyBlocked();
+
+        String prefix = "mounted-";
+        mountIndices(indicesToBeMounted.keySet(), prefix, repository, snapshotName, FULL_COPY);
+        mockRepository.unblock();
```
henningandersen, line 331: I am not sure I understand why this unblocks here. Should it not wait until after we mounted the `indexToSkip`? Otherwise the test might only be checking against started indices.

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,6 +271,97 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
         });
     }
 
+    public void testOvercommitInitializingSearchableSnapshotShards() throws Exception {
```
henningandersen, line 274: Can we use a similar name to the previous test, since they sort of test the same but in different flavors?

henningandersen: Also, I wonder why the tests are structured differently. The previous one mounts the same indices twice under different names. This one picks one not to mount, mounts the rest and then mounts the one extra index. I do not think there is a reason for that difference and maybe we can share most of the code between the two tests if we align the tests?

henningandersen (commented):
Did another round on this direction and left more comments.

server/src/test/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDeciderUnitTests.java
arteam: Not sure I'm following. I believe shards don't get filtered by the node id. `RoutingNode` just index all the shards that are supplied in its constructor.

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
arteam: It's a good point. I actually not sure whether we need to unblock the repository at all.

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
arteam: I've extracted some common code of the tests into methods:

* createIndices
* createRepository
* createSnapshot
* mountIndices

fcofdez:
@arteam I think something went wrong while the branch was updated, there're some unrelated changes, can you check please? Additionally, it's preferred to avoid force pushing in-review branches, as github gets confused sometimes. Thanks!

arteam:
Hey @fcofdez, sorry for the force push, it was a trivial change of the visibility modifier for the `createRepository` method. 
I don't see any unrelated changes, I had to fix some merge conflicts in `DiskThresholdDeciderUnitTests` when pulling the latest changes from master, though.

fcofdez:
I can see this and other unrelated changes https://github.com/elastic/elasticsearch/pull/86331/files#diff-db4b8f7b56ce1418ffd2304e9a50f41140917dd02b6c7bfd36cc780cc6de9dbc

arteam:
Thank you! You're right, they are result of auto-formatting for the merge commit. I'll fix the changes.

arteam:
Sorry, I had do redo the merge and force push the new change, because the previous was impossible to repair with a revert commit :(

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,6 +271,90 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
...
+                    .count(),
+                equalTo((long) indicesToBeMounted.size())
+            );
+        });
+
+        mountIndices(List.of(indexToSkip), prefix, repositoryName, snapshotName, FULL_COPY);
+        assertBusy(() -> {
+            var state = client().admin().cluster().prepareState().setRoutingTable(true).get().getState();
+            assertThat(
+                state.routingTable()
```
fcofdez, line 344: I think it's clearer to get the `IndexRoutingTable` for that particular index using `RoutingTable#index` and then assert that the index hasn't started, wdyt?

fcofdez, line 344: I think it's clearer to get the `IndexRoutingTable` for that particular index using `RoutingTable#index` and then assert that the index hasn't started, wdyt?

fcofdez (commented):
Looks mostly good, I left some comments in the tests.

server/src/test/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDeciderUnitTests.java
```diff
@@ -467,16 +471,73 @@ public void testShardSizeAndRelocatingSize() {
...
+            )
+            .build();
+
+        Index index = new Index("test", "1234");
+        String nodeId = "node1";
+        List<ShardRouting> shards = new ArrayList<>();
+        for (int i = 1; i <= 3; i++) {
+            int expectedSize = 10 * i;
+            shards.add(
+                ShardRoutingHelper.moveToStarted(
```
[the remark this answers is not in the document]
arteam: Split it.

server/src/test/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDeciderUnitTests.java
```diff
@@ -467,16 +471,73 @@ public void testShardSizeAndRelocatingSize() {
...
+                            .put(INDEX_STORE_TYPE_SETTING.getKey(), SEARCHABLE_SNAPSHOT_STORE_TYPE)
+                    )
+                    .numberOfShards(3)
+                    .numberOfReplicas(1)
+            )
+            .build();
+
+        Index index = new Index("test", "1234");
+        String nodeId = "node1";
+        List<ShardRouting> shards = new ArrayList<>();
```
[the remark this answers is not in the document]
arteam: Added

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
arteam: Good idea! Change it to `state.routingTable().index(prefix + indexToSkip)`

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,6 +271,90 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
...
+                    .allShards()
+                    .stream()
+                    .filter(s -> indicesToBeMounted.containsKey(s.shardId().getIndexName().replace(prefix, "")))
+                    .filter(s -> state.metadata().index(s.shardId().getIndex()).isSearchableSnapshot())
+                    .filter(s -> otherDataNodeId.equals(s.currentNodeId()))
+                    .filter(s -> s.state() == ShardRoutingState.STARTED)
```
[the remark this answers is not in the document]
arteam: Shards never reliably go a to a state where all of them are `INITIALIZED`. It's always some mixture of `INITIALIZED` and `STARTED` until all the shards go to `STARTED`. I'd love to add a check that can verify that a certain amount of shards are being initialized, but I have no idea how to control it (`setBlockOnceOnReadSnapshotInfoIfAlreadyBlocked` increases the amount of INITIALIZED shards, but doesn't specifically set how many).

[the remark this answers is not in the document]
fcofdez: I see what's going on here, `setBlockOnceOnReadSnapshotInfoIfAlreadyBlocked` only blocks once, we should have a new blob store implementation that blocks all the snapshot info reads until we deem it necessary. That way all the searchable snapshot shards should be in `ShardRoutingState.INITIALIZING` until we unblock the execution. We can pair to solve this if you want 👍.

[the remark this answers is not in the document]
arteam: I've created a custom `CustomMockRepository` that locks on `restoreShard` which prevents shards to go from `INITIALIZED` to `STARTED`. Now we can check that we can go overboard and trip the watermark with initializing shards taking space. It uses a static `CountDownLatch` which I don't like and am open to suggestions how to implement it better!

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,10 +283,159 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
...
+        String repositoryName = "repository";
+        createRepository(repositoryName, CustomMockRepositoryPlugin.TYPE);
+
+        String snapshotName = "snapshot";
+        createSnapshot(repositoryName, snapshotName, nbIndices);
+
+        Map<String, Long> indicesStoresSizes = sizeOfShardsStores("index-*");
+        assertAcked(client().admin().indices().prepareDelete("index-*"));
+
+        String otherDataNode = internalCluster().startNode(
```
fcofdez, line 301: nit: it might be clearer to name it `coldNodeName`

fcofdez, line 301: nit: it might be clearer to name it `coldNodeName`

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,10 +283,159 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
...
+        });
+
+        logger.info("--> Unlocking the initialized shards");
+        var mockRepository = (CustomMockRepository) internalCluster().getCurrentMasterNodeInstance(RepositoriesService.class)
+            .repository(repositoryName);
+        mockRepository.unlock();
+
+        assertBusy(() -> {
+            var state = client().admin().cluster().prepareState().setRoutingTable(true).get().getState();
+            assertThat(
```
fcofdez, line 363: nit: Since there's just one index that should be unassigned, I think it might be clearer to assert that the shard is `UNASSIGNED`?

fcofdez, line 363: nit: Since there's just one index that should be unassigned, I think it might be clearer to assert that the shard is `UNASSIGNED`?

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,10 +283,159 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
...
+            Environment environment,
+            NamedXContentRegistry namedXContentRegistry,
+            ClusterService clusterService,
+            BigArrays bigArrays,
+            RecoverySettings recoverySettings
+        ) {
+            super(metadata, environment, namedXContentRegistry, clusterService, bigArrays, recoverySettings);
+        }
+
+        private void unlock() {
```
fcofdez, line 418: nit: `unlockRestore()`?

fcofdez, line 418: nit: `unlockRestore()`?

fcofdez (commented):
I left a couple of small comments but this looks good to me, thanks for the iterations @arteam !

arteam:
@elasticmachine run elasticsearch-ci/docs-check

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/allocation/SearchableSnapshotDiskThresholdIntegTests.java
```diff
@@ -268,10 +283,156 @@ public void testHighWatermarkCanBeExceededOnColdOrFrozenNode() throws Exception
         });
     }
 
+    public void testHighWatermarkCanBeExceededOnColdNode() throws Exception {
```
fcofdez, line 286: nit: `testHighWatermarkCanNotBeExceededWithInitializingSearchableSnapshots`?

fcofdez (approved):
LGTM 👍. Let's wait for @henningandersen review before merging the changes.

arteam:
@elasticmachine update branch

build-tools-internal/version.properties
```diff
@@ -19,7 +19,7 @@ ecsLogging        = 1.2.0
 
 jna               = 5.10.0
 
-netty             = 4.1.77.Final
+netty=4.1.77.Final
```
henningandersen, line 22: Can we revert this, looks unrelated?

distribution/src/config/log4j2.properties
```diff
@@ -63,8 +63,8 @@ appender.deprecation_rolling.name = deprecation_rolling
 appender.deprecation_rolling.fileName = ${sys:es.logs.base_path}${sys:file.separator}${sys:es.logs.cluster_name}_deprecation.json
 appender.deprecation_rolling.layout.type = ECSJsonLayout
 # Intentionally follows a different pattern to above
-appender.deprecation_rolling.layout.dataset = deprecation.elasticsearch
-appender.deprecation_rolling.filter.rate_limit.type = RateLimitingFilter
+appender.deprecation_rolling.layout.dataset=deprecation.elasticsearch
+appender.deprecation_rolling.filter.rate_limit.type=RateLimitingFilter
```
henningandersen, line 67: Can we revert this, looks unrelated?

henningandersen (commented):
I started on this review, but it looks like a lot of unrelated changes have happened, making it a bit hard to review. Can you please revert all the unrelated changes?

arteam:
Ah, I already fixed it yesterday, but I didn't push the fixed merge commit. And sorry for the force push, unfortunately it's not straightforward to revert a mangled merge commit.

arteam:

henningandersen (commented):
This looks mostly good, left 2 comments to look into.

arteam:
@elasticmachine update branch

arteam:

server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDecider.java
```diff
@@ -125,21 +126,37 @@ public static long sizeOfRelocatingShards(
...
+                    clusterInfo,
+                    null,
+                    metadata,
+                    routingTable
+                );
+            }
+        }
+
+        // Count the STARTED searchable snapshot shards which are unaccounted in the cluster
+        for (ShardRouting shard : node.started()) {
```
henningandersen, line 155: I am a bit afraid of the time complexity here. With 1000 started shards here, this turns a bit expensive and runs many times on every reroute.

I'd propose to cache this in `RoutingAllocation` such that we cache the amount of total size to add by `RoutingNode` but also keep a "modification counter" from the `RoutingNode` in the cache to detect that it changed. We could also more explicitly clear the cache upon changes but that seems like a more advanced solution.

henningandersen (commented):
This looks mostly good, though I have a couple more asks here.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDecider.java
arteam: Hey @henningandersen! I created a static cache in `RoutingAllocation` based on `RoutingNode`. If `RoutingNode` changes, the cache will calculate a new value because `equals&hasCode` will be changed too. I also added a TTL for the cache, so we update the cache based on fresh `ClusterServiceInfo`

arteam:
@elasticmachine update branch

arteam:
@elasticmachine
run elasticsearch-ci/bwc
run elasticsearch-ci/docs-check

arteam:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -333,6 +346,33 @@ public void setHasPendingAsyncFetch() {
...
 
+    public long unaccountableSearchableSnapshotSize(RoutingNode node) {
+        try {
+            return unaccountableSearchableSnapshotSizes.computeIfAbsent(node, k -> {
+                long totalSize = 0;
+                // Count the STARTED searchable snapshot shards which are unaccounted in the cluster
+                for (ShardRouting shard : node.started()) {
+                    if (clusterInfo == null) {
+                        continue;
+                    }
```
henningandersen, line 357: Should this not go outside the loop - or even outside the `computeIfAbsent`?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -333,6 +346,33 @@ public void setHasPendingAsyncFetch() {
...
+                    DiskUsage usage = clusterInfo.getNodeMostAvailableDiskUsages().get(node.nodeId());
+                    ClusterInfo.ReservedSpace reservedSpace = clusterInfo.getReservedSpace(
+                        node.nodeId(),
+                        usage != null ? usage.getPath() : ""
+                    );
+                    if (clusterState.metadata().getIndexSafe(shard.index()).isSearchableSnapshot()
```
henningandersen, line 363: I wonder if we can check `shard.getExpectedShardSize() > 0` first, then check for `isSearchableSnapshot`, just to optimize a bit. Also, I'd like to defer looking up the `usage` and `reservedSpace` until we have determined that it is a searchable snapshot.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -333,6 +346,33 @@ public void setHasPendingAsyncFetch() {
...
+                        && reservedSpace.containsShardId(shard.shardId()) == false
+                        && clusterInfo.getShardSize(shard) == null) {
+                        totalSize += Math.max(shard.getExpectedShardSize(), 0L);
+                    }
+                }
+                return totalSize;
+            });
+        } catch (ExecutionException ignore) {
```
henningandersen, line 371: Once we make `unaccountableSearchableSnapshotSizes` non-static, we can use a `Map` instead and I think this exception is then no longer needed. I'd prefer not to swallow exceptions.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -77,6 +83,13 @@ public class RoutingAllocation {
     @Nullable
     private final DesiredNodes desiredNodes;
 
+    private static final TimeValue CACHE_TTL = TimeValue.timeValueMinutes(1);
+    private static final Cache<RoutingNode, Long> unaccountableSearchableSnapshotSizes = CacheBuilder.<RoutingNode, Long>builder()
```
henningandersen, line 87: I'd like to call the variable `unaccountedSearchableSnapshotSizes` instead. Also, please add javadoc describing what it means (searchable snapshots not yet registered in `ClusterInfo`).

henningandersen (commented):
A few more comments I'd like to see addressed.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -73,6 +79,13 @@ public class RoutingAllocation {
 
     private final Map<String, SingleNodeShutdownMetadata> nodeReplacementTargets;
 
+    private static final TimeValue CACHE_TTL = TimeValue.timeValueMinutes(1);
+    private static final Cache<RoutingNode, Long> unaccountableSearchableSnapshotSizes = CacheBuilder.<RoutingNode, Long>builder()
```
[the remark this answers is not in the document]
arteam: Yes, this was one the earlier designs, but then I went with the `static` approach

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
arteam: Good catch. I've moved it

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
arteam: Awesome! I've added the `shard.getExpectedShardSize() > 0` check and removed `Math.max(shard.getExpecredShardSize, 0) call

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
arteam: True that!

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
arteam: That indeed looks better! I've renamed it the field and the getter to `unaccountedSearchableSnapshotSizes` and added a comment to it.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -119,6 +124,28 @@ public RoutingAllocation(
         }
         this.nodeReplacementTargets = Map.copyOf(targetNameToShutdown);
         this.desiredNodes = DesiredNodes.latestFromClusterState(clusterState);
+        unaccountedSearchableSnapshotSizes = new HashMap<>();
```
fcofdez, line 127: nit: can we assign the map at the end and make it unmodifiable?

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
```diff
@@ -67,7 +70,11 @@ private RoutingNode(RoutingNode original) {
         this.shards = new LinkedHashMap<>(original.shards);
         this.relocatingShards = new LinkedHashSet<>(original.relocatingShards);
         this.initializingShards = new LinkedHashSet<>(original.initializingShards);
+        this.startedShards = new LinkedHashSet<>(original.startedShards);
         this.shardsByIndex = Maps.copyOf(original.shardsByIndex, HashSet::new);
+        for (Map.Entry<Index, Set<ShardRouting>> entry : original.shardsByIndex.entrySet()) {
```
fcofdez, line 75: is this necessary? isn't this map copied in the previous line?

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
```diff
@@ -43,6 +43,8 @@ public class RoutingNode implements Iterable<ShardRouting> {
 
     private final LinkedHashSet<ShardRouting> relocatingShards;
 
+    private final LinkedHashSet<ShardRouting> startedShards;
```
fcofdez, line 46: can we add some tests to check that the started shards are handled correctly in this class?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdMonitor.java
```diff
@@ -423,13 +423,14 @@ public void onNewInfo(ClusterInfo info) {
 
     // exposed for tests to override
     long sizeOfRelocatingShards(RoutingNode routingNode, DiskUsage diskUsage, ClusterInfo info, ClusterState reroutedClusterState) {
-        return DiskThresholdDecider.sizeOfRelocatingShards(
+        return DiskThresholdDecider.sizeOfUnaccountedShards(
```
fcofdez, line 426: I'm not sure if it is correct to use 0L here? this might mess up the disk thresholds monitoring?

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
```diff
@@ -485,13 +482,14 @@ public ShardRouting reinitializeReplicaShard() {
      * <code>INITIALIZING</code> or <code>RELOCATING</code>. Any relocation will be
```
fcofdez, line 482: In `public ShardRouting(ShardId shardId, StreamInput in)` we're serializing/de-serializing the shard size only when the state is `state == ShardRoutingState.UNASSIGNED || state == ShardRoutingState.INITIALIZING` I think that's not consistent with the current behaviour?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/AllocationService.java
```diff
@@ -571,8 +571,10 @@ private static void applyStartedShards(RoutingAllocation routingAllocation, List
                     + startedShard
                     + " but was: "
                     + routingNodes.getByAllocationId(startedShard.shardId(), startedShard.allocationId().getId());
-
-            routingNodes.startShard(logger, startedShard, routingAllocation.changes());
+            long expectedShardSize = routingAllocation.metadata().getIndexSafe(startedShard.index()).isSearchableSnapshot()
```
fcofdez, line 574: I think this is always `ShardRouting.UNAVAILABLE_EXPECTED_SHARD_SIZE`? I cannot find where we populate this with the value from `org.elasticsearch.snapshots.SnapshotShardSizeInfo#getShardSize(org.elasticsearch.cluster.routing.ShardRouting, long)`. Am I missing something?

fcofdez (commented):
I left a few comments, the most important one is a question about when do we populate the shard expected size with the searchable snapshot size, I couldn't find where we do that, but maybe I'm just missing it. Thanks for iterating into this.

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
arteam: Thank you for spotting this! That's a merge artifact from one the refactoring that happened during the development of the PR. I've removed it.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/AllocationService.java
arteam: It gets set in `ShardRouting#moveToStarted`

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
arteam: Good point! I've changed the serialization logic to write `expectedShardSize` for `STARTED` shards starting from `8.5.0`

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdMonitor.java
arteam: Don't have access to `RoutingAllocation` here, so I'm not sure what's a better approach here.

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
arteam: Good idea, I've added a test for checking that the `started` method accounts for added and removed nodes.

arteam:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/cluster/routing/allocation/AllocationService.java
fcofdez: I see it know, thanks for clarifying!

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -119,6 +125,29 @@ public RoutingAllocation(
         }
         this.nodeReplacementTargets = Map.copyOf(targetNameToShutdown);
         this.desiredNodes = DesiredNodes.latestFromClusterState(clusterState);
+        Map<String, Long> unaccountedSearchableSnapshotSizes = new HashMap<>();
```
fcofdez, line 128: nit: Can we move this into a static method?

server/src/test/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDeciderUnitTests.java
```diff
@@ -472,22 +477,113 @@ public void testShardSizeAndRelocatingSize() {
...
+        String nodeId = "node1";
+        String anotherNodeId = "another_node";
+
+        List<ShardRouting> shards = new ArrayList<>();
+        int anotherNodeShardCounter = 0;
+        for (int i = 1; i <= 3; i++) {
+            int expectedSize = 10 * i;
+            shards.add(createShard(index, nodeId, i - 1, expectedSize));
+            // randomly add shard for non-searchable snapshot index
+            if (randomBoolean()) {
```
fcofdez, line 519: isn't it possible to add more shards than expected for `another_index` here?

server/src/test/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDeciderUnitTests.java
```diff
@@ -472,22 +477,113 @@ public void testShardSizeAndRelocatingSize() {
...
+            .nodes(DiscoveryNodes.builder().add(node).add(anotherNode).build())
+            .build();
+        RoutingAllocation allocation = new RoutingAllocation(
+            null,
+            clusterState,
+            new DevNullClusterInfo(Map.of(), Map.of(), Map.of()),
+            null,
+            0
+        );
+        long sizeOfRelocatingShards = sizeOfUnaccountedShards(
```
fcofdez, line 557: can we add some initializing searchable snapshot shards randomly here to ensure that those are accounted as well?

fcofdez (approved):
This LGTM, I left a couple of minor nits. Maybe we want to get @henningandersen +1 too here? Thanks for all the iterations!

henningandersen (approved):
LGTM.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -119,6 +125,33 @@ public RoutingAllocation(
...
+        if (clusterInfo != null) {
+            for (RoutingNode node : clusterState.getRoutingNodes()) {
+                long totalSize = 0;
+                for (ShardRouting shard : node.started()) {
+                    DiskUsage usage = clusterInfo.getNodeMostAvailableDiskUsages().get(node.nodeId());
+                    ClusterInfo.ReservedSpace reservedSpace = clusterInfo.getReservedSpace(
+                        node.nodeId(),
+                        usage != null ? usage.getPath() : ""
+                    );
```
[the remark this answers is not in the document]
arteam: Good catch! It ended up in the inner loop as an artifact of one the review comments.

arteam:
Thank you both!

PR #86336, opened 2022-05-02
QL: Cleanup BWC tests for unsupported version

Luegg:
Since the jump to 8.x we no longer need some of the version switches in the bwc tests because we no longer test against version before 7.17.


Luegg:

astefan (approved):
LGTM

PR #86344, opened 2022-05-02
Introduce constants for trivial QueryCachingPolicies

original-brownbear:
Just a random find:
Cleaning this up, mainly to slightly simplify the already far too complex `IndexShard` constructor.

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/cache/query/TrivialQueryCachingPolicy.java
+++ b/server/src/main/java/org/elasticsearch/index/cache/query/TrivialQueryCachingPolicy.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/elasticsearch/index/cache/query/TrivialQueryCachingPolicy.java
```diff
@@ -0,0 +1,30 @@
...
+    NEVER(false);
+
+    private final boolean shouldCache;
+
+    TrivialQueryCachingPolicy(boolean shouldCache) {
+        this.shouldCache = shouldCache;
+    }
+
+    @Override
+    public void onUse(Query query) {}
```
idegtiarenko, line 24: Looks like every implementation is empty. Is it possible to remove this method?

idegtiarenko: Disregard that, just noticed that the interface is coming form lucene.

idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

PR #86345, opened 2022-05-02
User Profile - Use security origin for BWC cases

ywangd:
The security profile action origin and internal user is not available
before version 8.3.0. This PR makes all profile actions to use the
existing security origin if the cluster has any node that is older than
8.3.0.

The change makes it possible to use User Profile features in a mixed
cluster (7.17 and 8.3+) as long as the request always hits a node of
newer version first.

### DIFF
--- a/x-pack/qa/rolling-upgrade/src/test/resources/rest-api-spec/test/mixed_cluster/140_user_profile.yml
+++ b/x-pack/qa/rolling-upgrade/src/test/resources/rest-api-spec/test/mixed_cluster/140_user_profile.yml
@@ -1,0 +1,29 @@
+---
+"Test User Profile feature will work in a mixed cluster":
+
+  - skip:
+      features: [headers, node_selector]
+
+  - do:
+      node_selector:
+        version: " 8.3.0 - "
+      security.activate_user_profile:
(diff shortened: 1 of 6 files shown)


tvernum (approved):
LGTM

x-pack/qa/rolling-upgrade/src/test/resources/rest-api-spec/test/mixed_cluster/140_user_profile.yml
```diff
@@ -0,0 +1,33 @@
+---
+"Test User Profile feature will work in a mixed cluster":
+
+  - skip:
+      features: node_selector
+      version: " - 7.99.99"
+      reason: "https://github.com/elastic/elasticsearch/issues/86373"
```
ywangd, line 7: The BWC test fails with 7.17 node because we assert that mappings must have a key of single-mapping-name (`_doc`) in 7.17 which is not always true in a mixed cluster. Since it is an AssertionError, it does not seem to be a problem in production. I raised #86373 to discuss whether we can just remove the assertions. In the meantime, I am muting the tests for 7.x.

PR #86346, opened 2022-05-02
Deduplicate default FieldType in KeywordFieldMapper

original-brownbear:
The default type is incredibly common and instances are not trivial
in size with 16 fields. Heap dumps from larger data nodes holding many
keyword fields with the default field type can contain hundreds of MB
of heap used for these.
Same reasoning applies to the `TextSearchInfo` deduplication.
`TextSearchInfo` was turned into a record to give us an `equals` implementation.

On a frozen node holding 16k Beats indices this saves ~600M of heap in isolation.


romseygeek (approved):
LGTM, big +1 to making TextSearchInfo a record.

Issue #82342, opened 2022-01-07
`RestClusterGetSettingsAction` Requests the Full Metadata from Master


original-brownbear:
Similar issue as solved in https://github.com/elastic/elasticsearch/pull/78829.

This API is surprisingly expensive because it requests the full metadata via get cluster state request from master when called only to return just the settings from it to the user. This can mean requesting tens or hundreds of MB from master in a large cluster unexpectedly.

We should adjust this to not require fetching the full cluster state. One option would be to add a dedicate master node action for returning just the settings.

Relates #77466

PR #86347, opened 2022-05-02
Clean up request/response for get-cluster-settings

DaveCTurner:
Today there are request and response objects for the
get-cluster-settings action but the request is unused and the response
is only used in the REST layer. This commit removes the unused request
and renames the response to reflect that it's not a transport-layer
response. It also tidies a few things up in this area, removing the
unused `ActionResponse` superclass, making its fields final, and
replacing the overly-general `RestBuilderListener` with a regular
`RestToXContentListener` in the REST action.

Relates #82342 because to resolve that issue we will want to introduce
transport-layer request/response classes, and the classes involved in
this commit are in the way of that change.


grcevski (approved):
LGTM! Thanks David!

PR #86349, opened 2022-05-02
[8.2] Fix NPE when using object field as match field for enrich policy (#86089)

martijnvg:
 - Fix NPE when using object field as match field for enrich policy (#86089)


jbaiera:
Not sure if you used the backport tool but the labels that got set on this PR were mangled.

PR #86353, opened 2022-05-02
Add readImmutableMap for Java Map

rjernst:
The readImmutableMap method of StreamInput reads an ImmutableOpenMap. As
we work towards #86239, we want to continue using immutable maps, but
instead use those from the JDK. This commit renames the existing
readImmutableMap to readImmutableOpenMap, and reimplements the existing
method to return an immutable Map using Map.of and Map.ofEntries.

relates #86239


server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
```diff
@@ -680,13 +680,28 @@ public Map<String, Object> readMap() throws IOException {
         return (Map<String, Object>) readGenericValue();
     }
 
+    public <K, V> Map<K, V> readImmutableMap(Writeable.Reader<K> keyReader, Writeable.Reader<V> valueReader) throws IOException {
```
original-brownbear, line 683: Can we add a comment here, making it clear that `valueReader` must never return `null`? I see we're using the immutable open map with optional strings for the values, that will throw NPEs with `Map.of(` and can't be ported directly. Might be nice to make that clear in the docs?

original-brownbear (approved):

PR #86355, opened 2022-05-02
Make embedded class loader module aware

ChrisHegarty:
This PR makes embedded class loader module aware.

This change adds support to embedded class loader to load the provider
and implmentation dependencies as modules - within their own module 
layer - when the caller itself is a named module. Currently, this code
is not yet triggered during deployment, since the caller is always an
unnamed module, but the caller will be moularized in a subsequent
change.


libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedModulePath.java
```diff
@@ -0,0 +1,363 @@
...
+        return true;
+    }
+
+    private static final String MANIFEST_PATH = "META-INF/MANIFEST.MF";
+
+    static String moduleNameFromManifestOrNull(Path path) {
+        Path mp = path.resolve(MANIFEST_PATH);
+        if (Files.exists((mp))) {
+            try (InputStream is = Files.newInputStream(mp)) {
+                if (is != null) {
```
rjernst, line 353: When can newInputStream return null? I don't see it in the javadocs as a valid return

libs/core/src/main/java/org/elasticsearch/core/internal/provider/ProviderLocator.java
```diff
@@ -10,34 +10,82 @@
...
+            providerType,
+            ProviderLocator.class.getClassLoader(),
+            providerModuleName,
+            missingModules,
+            ProviderLocator.class.getModule().isNamed()
+        );
+    }
+
+    public ProviderLocator(
+        Module caller,
```
rjernst, line 67: Can we not find this ourselves? Or when would it differ from the module of the providerType?

libs/core/src/main/java/org/elasticsearch/core/internal/provider/ProviderLocator.java
```diff
@@ -10,34 +10,82 @@
...
 
-    public ProviderLocator(String providerName, Class<T> providerType) {
+    private final ClassLoader parentLoader;
+
+    private final Set<String> missingModules;
+
+    // whether to load the provider implementation as a module or not
+    private final boolean loadAsProviderModule;
+
+    public ProviderLocator(
```
rjernst, line 48: Why do we need two slightly different ctors, can we just always call the more explicit version below with the additional choice of whether to load as a module?

rjernst (commented):
This is great!I left some questions and comments.

libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoader.java
```diff
@@ -221,8 +228,85 @@ public URL nextElement() {
...
+
+    /**
+     * Returns a module finder capable of finding the modules that are loadable by this embedded
+     * impl class loader.
+     *
+     * <p> The module finder returned by this method can be used during resolution in order to
+     * create a configuration. This configuration can subsequently be materialized as a module layer
+     * in which classes and resources are loaded by this embedded impl class loader.
+     */
+    InMemoryModuleFinder moduleFinder(Set<String> missingModules) throws IOException {
```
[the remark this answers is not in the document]
rjernst: I see now from later cod this is how we filter out the modules we don't intend to provide. I still think it should be documented here or in InMemoryModuleFinder.

libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedModulePath.java
```diff
@@ -0,0 +1,363 @@
...
+            .flatMap(Optional::stream)
+            .collect(Collectors.toSet());
+
+        try {
+            Map<String, List<String>> map = new HashMap<>();
+            // parse each service configuration file
+            for (String sn : serviceNames) {
+                Path se = path.resolve(SERVICES_PREFIX + sn);
+                List<String> providerClasses = new ArrayList<>();
+                try (InputStream in = Files.newInputStream(se)) {
```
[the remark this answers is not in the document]
ChrisHegarty: Good idea. Done.

libs/core/src/main/java/org/elasticsearch/core/internal/provider/ProviderLocator.java
ChrisHegarty: we could, but it would require a stack walk. I could go either way on this, since it's only used for a module declaration check anyway.

libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoader.java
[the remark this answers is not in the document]
ChrisHegarty: Added

libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedModulePath.java
```diff
@@ -0,0 +1,363 @@
...
+        if (path.startsWith(MRJAR_PREFIX_PATH)) {
+            assert path.getNameCount() >= 3;
+            Path p = path.subpath(3, path.getNameCount());
+            return p;
+        }
+        return path;
+    }
+
+    static Set<String> explodedPackages(Path dir) {
+        String separator = dir.getFileSystem().getSeparator();
```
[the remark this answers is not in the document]
ChrisHegarty: Well, yes and no ;-) The JDK had a similar issue, which I fixed. This then allows explicit modules to be found in our embedded loader, but sadly not auto modules.

ChrisHegarty:
@elasticmachine run elasticsearch-ci/part-1-windows

ChrisHegarty:
@elasticmachine run elasticsearch-ci/part-1-windows

libs/core/src/main/java/org/elasticsearch/core/internal/provider/ProviderLocator.java
ChrisHegarty: the check has been simplified and reworked

libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoader.java
```diff
@@ -217,8 +223,88 @@ public URL nextElement() {
...
+        try {
+            URI uri = url.toURI();
+            if (uri.getScheme().equals("jar")) {
+                String s = uri.toString();
+                return URI.create(s.substring(0, s.lastIndexOf("!/")));
+            } else {
+                return URI.create(getParent(getParent(getParent(uri.toString()))));
+            }
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
```
rjernst, line 304: Since this is coming from a file, this should not be possible right? It could be an AssertionError?

libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedModulePath.java
```diff
@@ -0,0 +1,339 @@
...
+                .toList();
+            if (providerClasses.isEmpty() == false) {
+                map.put(sn, providerClasses);
+            }
+        }
+        return map;
+    }
+
+    /** Drops comments and trims the given string. */
+    static String dropCommentAndTrim(String line) {
```
rjernst, line 208: I understand this is package private so it can be unit tested, but having so many methods available makes it difficult to tell what is the real set of methods that are meant to be used here. Could this and maybe some others be made private and testing just done on the methods intended to be used elsewhere?

rjernst (approved):
LGTM, I noticed a couple other minor things, feel free to ignore them.

libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedImplClassLoader.java
ChrisHegarty: Done.

libs/core/src/main/java/org/elasticsearch/core/internal/provider/EmbeddedModulePath.java
ChrisHegarty: I moved the code around a bit, and remove the javadoc style comments from "internal" methods. This should help with readability, etc.

PR #81998, opened 2021-12-21
TSDB: Add support for composite aggregation on `_tsid` field

csoulios:
Currently, composite aggregation supports only string or numeric values for the fields in the `sources` part of the composite aggregation. Although `_tsid` is encoded and stored as a byte array, it is formatted as map `{"dim1": "value1", "dim2": value2", ...}` for input/output.

This PR adds support for composite aggregation on `_tsid` field.

Relates to #74660

PR #86359, opened 2022-05-02
Less complexity in after-key parsing for unmapped fields

### LINKED ISSUE #85928: Partial shard failures for non-string composite after_key values
### Elasticsearch Version

latest from master

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

windows

### Problem Description

Using composite's `after_key` with a non-String value to get the next "page" of results in a `composite` aggregation query results in a parsing exception with the message `Cannot set after key in the composite aggregation [keys] - incompatible value in the position 0: invalid value, expected string, got Integer`.

At least one of the indices used in the search request does not have a mapping for that specific field.

### Steps to Reproduce

`PUT /index1`

```
{
	"mappings": {
		"properties": {
			"number": {
				"type":"long"
			},
			"text": {
				"type":"keyword"
			}
		}
	}
}
```

`PUT /index2`

```
{
	"mappings": {
		"properties": {
			"date": {
				"type":"date"
			},
			"text": {
				"type":"keyword"
			}
		}
	}
}
```

`POST /index1/_bulk?refresh`

```
{"index":{"_id":1}}
{"number": 123,"text":"bla"}
{"index":{"_id":2}}
{"number": 100,"text":"test"}
{"index":{"_id":3}}
{"number": 200,"text":"foo"}
```

`POST /index2/_bulk?refresh`

```
{"index":{"_id":1}}
{"text":"bla"}
{"index":{"_id":2}}
{"text":"test"}
{"index":{"_id":3}}
{"text":"foo"}
```

`POST /index*/_search`

```
{
  "size": 0,
  "query": {
	"match_all": {}
  },
  "_source": true,
  "track_total_hits": -1,
  "aggregations": {
    "keys": {
      "composite": {
        "size": 1,
        "sources": [
          {
            "number": {
              "terms": {
                "field": "number",
                "missing_bucket": false,
                "order": "asc"
              }
            }
          },
          {
            "text": {
              "terms": {
                "field": "text",
                "missing_bucket": false,
                "order": "asc"
              }
            }
          }
        ]
      }
    }
  }
}
```

results in

```
{
    "took": 1,
    "timed_out": false,
    "_shards": {
        "total": 2,
        "successful": 2,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "max_score": null,
        "hits": []
    },
    "aggregations": {
        "keys": {
            "after_key": {
                "number": 100,
                "text": "test"
            },
            "buckets": [
                {
                    "key": {
                        "number": 100,
                        "text": "test"
                    },
                    "doc_count": 1
                }
            ]
        }
    }
}
```

The next query uses the after_key above:

```
{
  "size": 0,
  "query": {
	"match_all": {}
  },
  "_source": true,
  "track_total_hits": -1,
  "aggregations": {
    "keys": {
      "composite": {
        "size": 1,
        "sources": [
          {
            "number": {
              "terms": {
                "field": "number",
                "missing_bucket": false,
                "order": "asc"
              }
            }
          },
          {
            "text": {
              "terms": {
                "field": "text",
                "missing_bucket": false,
                "order": "asc"
              }
            }
          }
        ],
        "after": {
          "number": 100,
          "text": "test"
        }
      }
    }
  }
}
```

And the result is 

```
{
    "took": 1,
    "timed_out": false,
    "_shards": {
        "total": 2,
        "successful": 1,
        "skipped": 0,
        "failed": 1,
        "failures": [
            {
                "shard": 0,
                "index": "index2",
                "node": "EqTUabPYTgKzITFjW5SMiA",
                "reason": {
                    "type": "parse_exception",
                    "reason": "Cannot set after key in the composite aggregation [keys] - incompatible value in the position 0: invalid value, expected string, got Integer",
                    "caused_by": {
                        "type": "illegal_argument_exception",
                        "reason": "incompatible value in the position 0: invalid value, expected string, got Integer",
                        "caused_by": {
                            "type": "illegal_argument_exception",
                            "reason": "invalid value, expected string, got Integer"
                        }
                    }
                }
            }
        ]
    },
    "hits": {
        "max_score": null,
        "hits": []
    },
    "aggregations": {
        "keys": {
            "after_key": {
                "number": 123,
                "text": "bla"
            },
            "buckets": [
                {
                    "key": {
                        "number": 123,
                        "text": "bla"
                    },
                    "doc_count": 1
                }
            ]
        }
    }
}
```

I would have expected Elasticsearch to ignore the fact that index2 doesn't have that specific field used as after_key.
As is, the behavior is inconsistent: for a `string` value there is no exception, for a numeric value there are partial shard failures.

not-napoleon:
Resolves #85928

The after-key parsing is pretty weird, and there are probably more bugs there.  I did not take the opportunity to refactor the whole thing, but we should.  This fixes the immediate problem by treating after keys as bytes refs when we don't have a field but think we want a keyword.  We were already doing that if the user asked for a missing bucket, this just extends the behavior in the case that we don't.

Long term, the terms Composite source (and probably other Composite sources) should have specializations for unmapped fields.  That's the direction we want to take aggs in general.

### DIFF
--- a/docs/changelog/86359.yaml
+++ b/docs/changelog/86359.yaml
@@ -1,0 +1,6 @@
+pr: 86359
+summary: Less jank in after-key parsing for unmapped fields
+area: CompositeAggs
+type: bug
+issues:
+ - 85928
--- a/docs/changelog/86359.yaml
+++ b/docs/changelog/86359.yaml
@@ -1,6 +1,6 @@
 pr: 86359
 summary: Less jank in after-key parsing for unmapped fields
-area: CompositeAggs
+area: Aggregations
 type: bug
 issues:
  - 85928
--- a/server/src/main/java/org/elasticsearch/search/aggregations/bucket/composite/GlobalOrdinalValuesSource.java
+++ b/server/src/main/java/org/elasticsearch/search/aggregations/bucket/composite/GlobalOrdinalValuesSource.java
@@ -105,7 +105,7 @@ class GlobalOrdinalValuesSource extends SingleDimensionValuesSource<BytesRef> {
         if (missingBucket && value == null) {
             afterValue = null;
             afterValueGlobalOrd = MISSING_VALUE_FLAG;
-        } else if (value.getClass() == String.class || (missingBucket && fieldType == null)) {
+        } else if (value.getClass() == String.class || (fieldType == null)) {
             // the value might be not string if this field is missing in this shard but present in other shards
             // and doesn't have a string type
             afterValue = format.parseBytesRef(value);
--- a/server/src/test/java/org/elasticsearch/search/aggregations/bucket/composite/CompositeAggregatorTests.java
+++ b/server/src/test/java/org/elasticsearch/search/aggregations/bucket/composite/CompositeAggregatorTests.java
@@ -172,6 +172,8 @@ public class CompositeAggregatorTests extends AggregatorTestCase {
                 createDocument("keyword", "c")
             )
         );
+
+        // Only aggregate on unmapped field, no missing bucket => no results
         testSearchCase(
             Arrays.asList(new MatchAllDocsQuery(), new DocValuesFieldExistsQuery("keyword")),
             dataset,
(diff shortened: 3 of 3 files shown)


csoulios (approved):
LGTM! 

For future reference when we decide to refactor this part, let's have in mind that a BytesRef field can also be an encoded `_tsid` field. (See #81998)

PR #86361, opened 2022-05-02
[ML] add new trained_models/{model_id}/_infer endpoint for all supervised models and deprecate deployment infer api

### LINKED ISSUE #86032: [ML] Extend the _infer endpoint to support all model types
### Description

NLP models have a very useful [_infer](https://www.elastic.co/guide/en/elasticsearch/reference/current/infer-trained-model-deployment.html) API for quickly evaluation. However, Boosted tree models created by Data Frame Analytics jobs and the build in `lang_ident` model cannot be evaluated in the `_infer` endpoint as it only applies to NLP models which are deployed. This is evident in the full URL which contains the `deployment` keyword

`_ml/trained_models/<model_id>/deployment/_infer`

For models that do not need to have deployments the `_infer` URL should be:

`_ml/trained_models/<model_id>/_infer`

It is less confusing to have a single `_infer` endpoint, the second form should be preferred over the first and the first deprecated. 

For models that need to be deployed before they can be evaluated the current sensible error messages should be retained.

benwtrent:
This commit adds a new `_ml/trained_models/{model_id}/_infer` API. This api works for both native NLP models and supervised models trained via Data Frame analytics. 

The format of the API is the same as the old `_ml/trained_models/{model_id}/deployment/_infer`. Taking a `docs` and an `inference_config` parameter.

This PR also deprecates the old experimental `_ml/trained_models/{model_id}/deployment/_infer` API.

The biggest difference is that the response now nests all results under an "inference_results" object.

closes: https://github.com/elastic/elasticsearch/issues/86032


benwtrent:
Should wait until all the work with: https://github.com/elastic/ml-cpp/pull/2258 is completed and pytorch tests unmuted. That way we can fully test functionality in upgrade tests and integration tests.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/action/InferModelAction.java
```diff
@@ -23,44 +29,87 @@
...
 import java.util.Map;
 import java.util.Objects;
+import java.util.stream.Collectors;
 
-public class InternalInferModelAction extends ActionType<InternalInferModelAction.Response> {
-
-    public static final InternalInferModelAction INSTANCE = new InternalInferModelAction();
+public class InferModelAction extends ActionType<InferModelAction.Response> {
     public static final String NAME = "cluster:internal/xpack/ml/inference/infer";
+    public static final String EXTERNAL_NAME = "cluster:monitor/xpack/ml/inference/infer";
```
benwtrent, line 36: We needed an external name for interaction at the REST layer and since inference needs to work during rolling upgrade, we cannot get rid of the old action name.

So, both point to effectively the same transport class.

sethmlarson (approved):
LGTM from an API perspective

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/action/InferModelAction.java
```diff
@@ -179,6 +283,14 @@ public static Builder builder() {
             return new Builder();
         }
 
+        @Override
+        public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+            builder.startObject();
+            builder.field("inference_results", inferenceResults.stream().map(InferenceResults::asMap).collect(Collectors.toList()));
```
davidkyle, line 289: Alternatively `inferenceResults` which implement `ToXContentFragment` could be mapped to `ToXContentObject`

```
        static ToXContentObject toXContentObject(ToXContentFragment fragment) {
            return (builder, params) -> {
                builder.startObject();
                fragment.toXContent(builder, params);
                builder.endObject();
                return builder;
            };
        }
```

and 

```
builder.field("inference_results", inferenceResults.stream().map(Response::toXContentObject).collect(Collectors.toList()));
```

davidkyle (approved):
LGTM

lcawl:
If we're keeping both the old and new doc pages (with the old API marked as deprecated), shouldn't the same be done for the rest-api-spec/src/main/resources/rest-api-spec/api/ml.infer_trained_model.json? (i.e. leave the existing .json file and just add the "deprecated" property). That way when we get to the point of generating docs from those specs, it'll match.

PR #86365, opened 2022-05-02
[Refactor] Use records instead of tuples in PluginService

williamrandolph:
PluginsService used tuples to pair plugin instances with other objects. I've tried replacing the ones that are used in lists and passed between methods with records, for clarity. I left the transient usages within stream operations alone.

This is my first PR with Java Records, so please let me know if I've messed up any standard conventions.

### DIFF
--- a/server/src/main/java/org/elasticsearch/plugins/PluginsService.java
+++ b/server/src/main/java/org/elasticsearch/plugins/PluginsService.java
@@ -57,6 +57,20 @@ import static org.elasticsearch.common.io.FileSystemUtils.isAccessibleDirectory;
 
 public class PluginsService implements ReportingService<PluginsAndModules> {
 
+    /**
+     * A loaded plugin is one for which Elasticsearch has successfully constructed an instance of the plugin's class
+     * @param info Metadata about the plugin, usually loaded from plugin properties
+     * @param instance The constructed instance of the plugin's main class
+     */
+    record LoadedPlugin(PluginInfo info, Plugin instance) {}
+
+    /**
+     * A constructed plugin has an associated classloader
+     * @param plugin A constructed instance of a plugin's main class
+     * @param loader The classloader for the plugin
+     */
+    private record PluginClassloaderContext(Plugin plugin, ClassLoader loader) {}
+
     private static final Logger logger = LogManager.getLogger(PluginsService.class);
 
     private final Settings settings;
@@ -65,7 +79,7 @@ public class PluginsService implements ReportingService<PluginsAndModules> {
     /**
      * We keep around a list of plugins and modules
      */
-    private final List<Tuple<PluginInfo, Plugin>> plugins;
+    private final List<LoadedPlugin> plugins;
     private final PluginsAndModules info;
 
     public static final Setting<List<String>> MANDATORY_SETTING = Setting.listSetting(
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -57,6 +57,20 @@
...
+     * @param instance The constructed instance of the plugin's main class
+     */
+    record LoadedPlugin(PluginInfo info, Plugin instance) {}
+
+    /**
+     * A constructed plugin has an associated classloader
+     * @param plugin A constructed instance of a plugin's main class
+     * @param loader The classloader for the plugin
+     */
+    private record PluginClassloaderContext(Plugin plugin, ClassLoader loader) {}
```
rjernst, line 72: I think this can be combined into LoadedPlugin, adding the ClassLoader there.

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -703,7 +718,7 @@ private Plugin loadBundle(Bundle bundle, Map<String, Tuple<Plugin, ClassLoader>>
                 );
             }
             Plugin plugin = loadPlugin(pluginClass, settings, configPath);
-            loaded.put(name, Tuple.tuple(plugin, spiLoader));
+            loaded.put(name, new PluginClassloaderContext(plugin, spiLoader));
```
rjernst, line 721: We have the pluginInfo here as well, in bundle, so we can create a LoadedPlugin

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -476,17 +490,17 @@ private static void addSortedBundle(
...
+        Map<String, PluginClassloaderContext> loaded = new HashMap<>();
         Map<String, Set<URL>> transitiveUrls = new HashMap<>();
         List<Bundle> sortedBundles = sortBundles(bundles);
         for (Bundle bundle : sortedBundles) {
             if (bundle.plugin.getType() != PluginType.BOOTSTRAP) {
                 checkBundleJarHell(JarHell.parseClassPath(), bundle, transitiveUrls);
 
                 final Plugin plugin = loadBundle(bundle, loaded);
-                plugins.add(new Tuple<>(bundle.plugin, plugin));
+                plugins.add(new LoadedPlugin(bundle.plugin, plugin));
```
rjernst, line 503: Instead of adding these plugins as we go, we can get the `values()` from `loaded` at the end of this method.

rjernst (commented):
Thanks for this cleanup! I have one suggestion that I think can make this a bit simpler.

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
williamrandolph: I think I managed to correctly add the ClassLoader to the LoadedPlugin object. The main thing I am not sure about is that I needed to pick something arbitrary for the classpath plugins, so I just used the current class's classloader.

williamrandolph: Changed my mind on this -- I used `null` for classpath plugin classloaders and added an `assert` statement to verify we don't hit the bad case when building classloaders for extended plugins.

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -57,6 +58,14 @@
 
 public class PluginsService implements ReportingService<PluginsAndModules> {
 
+    /**
+     * A loaded plugin is one for which Elasticsearch has successfully constructed an instance of the plugin's class
+     * @param info Metadata about the plugin, usually loaded from plugin properties
+     * @param instance The constructed instance of the plugin's main class
+     * @param loader   The classloader for the plugin
+     */
+    record LoadedPlugin(PluginInfo info, Plugin instance, @Nullable ClassLoader loader) {}
```
rjernst, line 67: I don't think we need to use Nullable. In production this should not be null, we just have tests that abuse this contract, knowing how it is used.

rjernst (approved):
LGTM

PR #86368, opened 2022-05-03
Remove use of ImmutableOpenIntMap in IndexRoutingTable

rjernst:
The routing table builder uses an hppc based immutable map to keep track
of shard id to shard table. However, when building the final routing
table, it is just converted to an array. This commit switches to a
HashMap.

relates #86239


original-brownbear (commented):
Phew this is the one place where the performance of the int map matters I think. Ideally, we'd just ditch the use of a map instead of moving to boxed `Integer` keys. It should be possible to just use an array for `shards` like we do in other places in the routing table already?

original-brownbear (approved):
I was gonna benchmark this ... but looking at it again, just LGTM this code is slow elsewhere :)

PR #86375, opened 2022-05-03
Speed up map diffing (2)

idegtiarenko:
This pr is a followup to the https://github.com/elastic/elasticsearch/pull/86228

I was reviewing changes and noticed that we accumulate upserts and diffs into HashMaps however we never access them by key. This change replaces HashMaps with an ArrayLists so that we avoid calculating the hashes/resolving collisions/resize underlying arrays.

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java
+++ b/server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java
@@ -14,11 +14,10 @@ import org.elasticsearch.common.collect.ImmutableOpenMap;
 import org.elasticsearch.common.io.stream.StreamInput;
 import org.elasticsearch.common.io.stream.StreamOutput;
 import org.elasticsearch.common.io.stream.Writeable.Reader;
-import org.elasticsearch.common.util.Maps;
 
 import java.io.IOException;
+import java.util.AbstractMap;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -27,7 +26,7 @@ import java.util.Set;
 public final class DiffableUtils {
     private DiffableUtils() {}
 
-    private static final MapDiff<?, ?, ?> EMPTY = new MapDiff<>(null, null, List.of(), Map.of(), Map.of()) {
+    private static final MapDiff<?, ?, ?> EMPTY = new MapDiff<>(null, null, List.of(), List.of(), List.of()) {
         @Override
         public Object apply(Object part) {
             return part;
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java
```diff
@@ -224,31 +223,47 @@ private static <K, T, H> MapDiff<K, T, H> diffOrEmpty(MapDiff<K, T, H> diff) {
...
             assert after != null && before != null;
 
             int inserts = 0;
-            for (Map.Entry<K, T> partIter : after.entrySet()) {
-                final K key = partIter.getKey();
-                T beforePart = before.get(key);
-                if (beforePart == null) {
-                    upserts.put(key, partIter.getValue());
+            var upserts = new ArrayList<Map.Entry<K, T>>();
+            var diffs = new ArrayList<Map.Entry<K, Diff<T>>>();
```
idegtiarenko, line 250: I wonder if we could come up with reasonable predictions for initial sizes? @original-brownbear maybe you have any suggestions?

server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java
```diff
@@ -224,31 +223,47 @@ private static <K, T, H> MapDiff<K, T, H> diffOrEmpty(MapDiff<K, T, H> diff) {
...
+        ) {
             assert after != null && before != null;
 
             int inserts = 0;
-            for (Map.Entry<K, T> partIter : after.entrySet()) {
-                final K key = partIter.getKey();
-                T beforePart = before.get(key);
-                if (beforePart == null) {
-                    upserts.put(key, partIter.getValue());
+            var upserts = new ArrayList<Map.Entry<K, T>>();
```
original-brownbear, line 249: We could use `Tuple` instead of `Map.Entry` here couldn't we?

original-brownbear (approved):
LGTM just one suggestion :)

server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java
original-brownbear: I don't think so. Best you could maybe do is some upper bound but that's only really helpful (and to a very minor degree only as well) in case you have less than the default size of 10 in both maps.
Probably not worth it :)

server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java
idegtiarenko: Using `Map.Entry` allows to reuse entry returned by the `entrySet()` iterator while calculating the diff.
It is also less confusing to know where is the key and value.

PR #86380, opened 2022-05-03
Use faster and smaller maps in MappingLookup

original-brownbear:
It's in the title. These data structures are a huge (if not the largest) contributor to the idle-state
heap consumption of large data nodes at the moment, lets make them a little more compact and as
a side effect also faster.

relates #77466


server/src/main/java/org/elasticsearch/index/mapper/FieldTypeLookup.java
```diff
@@ -81,6 +87,10 @@ final class FieldTypeLookup {
             // this will override concrete fields with runtime fields that have the same name
             fullNameToFieldType.put(fieldType.name(), fieldType);
         }
+        this.fullNameToFieldType = Map.copyOf(fullNameToFieldType);
```
romseygeek, line 90: Can you add a comment here explaining that using `Map.copyOf` results in smaller memory usage and faster access?

server/src/main/java/org/elasticsearch/index/mapper/MappingLookup.java
```diff
@@ -169,8 +172,11 @@ private MappingLookup(
 
         this.fieldTypeLookup = new FieldTypeLookup(mappers, aliasMappers, mapping.getRoot().runtimeFields());
         this.indexTimeLookup = new FieldTypeLookup(mappers, aliasMappers, Collections.emptyList());
-        this.fieldMappers = Collections.unmodifiableMap(fieldMappers);
-        this.objectMappers = Collections.unmodifiableMap(objects);
+        this.fieldMappers = Map.copyOf(fieldMappers);
```
romseygeek, line 175: Same here

romseygeek (approved):
LGTM.  I think it's worth adding comments in to explain why we're using these constructs (particularly the Map-of-Set one which isn't obvious at a glance).

original-brownbear:
Thanks Alan, comments added!

PR #86388, opened 2022-05-03
Do not use recovery from snapshots in searchable snapshots

fcofdez:
This commit disables recovering from snapshots for searchable snapshots
as the snapshot for these type of indices consist in a pointer to the
original snapshot and it produced confusing error messages.


server/src/main/java/org/elasticsearch/indices/recovery/RecoverySourceHandler.java
```diff
@@ -598,6 +599,12 @@ void phase1(IndexCommit snapshot, long startingSeqNo, IntSupplier translogOps, A
         }
     }
 
+    private boolean canUseSnapshots() {
+        return useSnapshots && request.canDownloadSnapshotFiles()
+        // Avoid using snapshots for searchable snapshots as these are implicitly recovered from a snapshot
+            && "snapshot".equals(INDEX_STORE_TYPE_SETTING.get(shard.indexSettings().getSettings())) == false;
```
tlrx, line 605:
```suggestion
            && shard.indexSettings().getIndexMetadata().isSearchableSnapshot() == false;
```

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/SearchableSnapshotsRecoverFromSnapshotIntegTests.java
```diff
@@ -0,0 +1,89 @@
...
+        Loggers.addAppender(logger, mockAppender);
+
+        // Relocate the searchable snapshot shard to the new node
+        client().admin()
+            .indices()
+            .prepareUpdateSettings(restoredIndexName)
+            .setSettings(Settings.builder().put("index.routing.allocation.require._name", newNode).build())
+            .get();
+
+        ensureGreen(restoredIndexName);
```
tlrx, line 83: We should also run some search queries.

server/src/main/java/org/elasticsearch/indices/recovery/RecoverySourceHandler.java
```diff
@@ -558,7 +559,7 @@ void phase1(IndexCommit snapshot, long startingSeqNo, IntSupplier translogOps, A
             }
             if (canSkipPhase1(recoverySourceMetadata, request.metadataSnapshot()) == false) {
                 cancellableThreads.checkForCancel();
-                final boolean canUseSnapshots = useSnapshots && request.canDownloadSnapshotFiles();
+                final boolean canUseSnapshots = canUseSnapshots();
```
tlrx, line 562: I'm confused we ever get to this place with searchable snapshot shards. I'd expect the target shard to act as if all files already exist before executing the start recovery request. This is the role of the `indexShard.preRecovery()` hook in `PeerRecoveryTargetService` which triggers the load of the files from the snapshot.

I quickly looked at the `canSkipPhase1()` method which is based on the now deprecated synced flush and I wonder if this is:
```java
if (source.getSyncId() == null || source.getSyncId().equals(target.getSyncId()) == false) {
            return false;
        }
```
is still relevant today. I suspect replacing it with `if (Objects.equal(source.getSyncId(), target.getSyncId()) == false)` should be enough (and assert that we don't run into phase1 with searchable snapshot shards).

fcofdez: > I'm confused we ever get to this place with searchable snapshot shards. I'd expect the target shard to act as if all files already exist before executing the start recovery request. This is the role of the indexShard.preRecovery() hook in PeerRecoveryTargetService which triggers the load of the files from the snapshot.

We compute what needs to be recovered in the target, either via snapshot or sent by the source, in `recoveryPlannerService#computeRecoveryPlan` that's why we end up here even in searchable snapshots.

About synced flush, I guess that we could remove it in a separate PR as I think that at this point we shouldn't encounter shards using synced flush?

tlrx: We discussed this together and we agreed to move forward with setting the "can use snapshot" flag to false either on the source or in the start recovery request.

I'm still confused by the sync id that prevent phase 1 to be skipped at all; in case of searchable snapshots the number of docs and the local checkpoint and max sequence number should be identical on the source and target, making the phase 1 skippable. That's something we should dig into.

fcofdez: I did some renaming and added some comments to hopefully clarify the code in that area. As next steps we could work to know if it's safe to remove that behaviour and simplify the code a bit.

fcofdez:
@elasticmachine update branch

tlrx (approved):
LGTM - And really sorry for the time it took me to review this.

fcofdez:
@elasticmachine update branch

fcofdez:
@elasticmachine update branch

fcofdez:
@elasticmachine update branch

fcofdez:
Thanks Tanguy!

Issue #84400, opened 2022-02-28
Open point-in-time can return partial results with skip_unavailable


dnhatn:
### Elasticsearch Version

8.0

### Java Version

_bundled_

### OS Version

linux

### Problem Description

[Opening](https://www.elastic.co/guide/en/elasticsearch/reference/current/point-in-time-api.html#point-in-time-api-example) a point-in-time can return partial results if some remote clusters are disconnected and `skip_unavailable` is set to true.

### Steps to Reproduce

Relates #82407

PR #86389, opened 2022-05-03
[ML][Transform] retry transform search if remote clusters are skipped

### LINKED ISSUE #84090: [Transform] retry searches despite skip_unavailable is set to true for a CCS remote
Transform by design retries if search returns partial results to not create results for incomplete data. CCS offers a similar setting called `skip_unavailable`. However this is not part of the search DSL but a cluster setting. A discussion about this choice can be found in #82407.

We've discovered this issue recently for data feeds: #83838. Apart from transform rollup(v1) is probably affected, too.

**Mitigation**

Do not use a remote connection that uses `skip_unavailable`. In case you need to set `skip_unavailable` to `true`, use separate remote connections.

**Fix**

If the current design stays, we need a similar fix as done in #84052. If the fix is applied in the super classes it is not necessary to have a separate fix for rollup(v1).

benwtrent:
This commit changes transforms and rollups handling `skipped` of remote clusters to be similar to how each handles shard failures. 

For transforms, we simply retry again later (up to a limit). 

For rollups, the failure is logged and rollups is tried again.

When `skip_unavailable: true` is set for the cluster, the user probably doesn't mean for rollups && transforms. It very well could be that they unexpectedly notice that data is "lost" since skipping the remote cluster is a silent operation. Transforms and rollups are not expected to miss data silently.

This commit addresses this bug and throws if any remote clusters were skipped in the search. 

closes: https://github.com/elastic/elasticsearch/issues/84090


benwtrent:
This is blocked by: https://github.com/elastic/elasticsearch/issues/84400

We need to be able to see if a cluster was skipped in a PIT creation or not.

PR #86398, opened 2022-05-03
User APIs allow use of internal usernames

### LINKED ISSUE #86326: Internal usernames are not special
Today, User APIs, e.g. [PutUser](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-user.html), throw errors when the given username clashes with any of the internal usernames, such as `_xpack`. But this restriction is not necessary because:

1. [Internal users](https://www.elastic.co/guide/en/elasticsearch/reference/current/internal-users.html#:~:text=The%20Elastic%20Stack%20security%20features,originate%20from%20within%20the%20cluster.) are used for issuing request from within the cluster. They are defined in code, immutable and not accessible at the REST layer. More specifically, there should be no REST API that can operate with or against them.
2. User APIs target users from either Native realm or Reserved realm, while internal users are from synthetic realms such as `__fallback` or `__attach`. A user is fully qualified by its username and the realm it is from. As such, an internal user is a completely different user from a Native realm user even when their usernames are identical.
3. There is no such restriction on any other realms, e.g. File realm, or LDAP realm. It is possible to have users with internal usernames from these realms and Elasticsearch is capable of handling them correctly because underlyingly internal users have their own distinct code path.

Based on the above, the proposal is to remove the API level restriction for internal usernames because they are not special. Dropping the user impacting restriction also makes it easier to evolve internal users in future (see also https://github.com/elastic/elasticsearch/pull/86026#discussion_r854303848).

NOTE: The similar distinction between Native realm and Reserved realm usernames must still be maintained. We believe the initial intention is to distinguish between usernames from these two realms (which are still valid) and the extra restriction on internal usernames was an accidental change (which can be removed as proposed above).

n1v0lg:
This PR removes API restrictions that prevent user-related actions
(`PutUser`, `DeleteUser`, etc.) to be performed on users with names
clashing with internal users (e.g., `_system`). Internal users are used
for issuing requests within the cluster and not accessible at the REST
layer; they are conceptually and practically separate from users
created through the API (they do not belong to the same realm and
follow different authentication flows). As such it's safe to remove the
current restriction. By removing, we are more consistent with how
usernames are treated across other realms (name re-use is allowed), and
ease evolving internal users without interfering with end-user
experience.

Closes #86326


x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/esnative/NativeRealmIntegTests.java
```diff
@@ -269,7 +270,15 @@ public void testAddAndGetRole() throws Exception {
...
+    public void testAddUserAndRoleThenAuth() {
+        testAddUserAndRoleThenAuth("joe");
+    }
+
+    public void testAddUserWithInternalUsernameAndRoleThenAuth() {
+        testAddUserAndRoleThenAuth(randomInternalUsername());
+    }
+
+    public void testAddUserAndRoleThenAuth(String username) {
         logger.error("--> creating role");
```
n1v0lg, line 282: Not sure why these are `error`

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/user/TransportDeleteUserActionTests.java
```diff
@@ -269,4 +230,14 @@ public void onFailure(Exception e) {
         assertThat(throwableRef.get(), sameInstance(e));
         verify(usersStore, times(1)).deleteUser(eq(request), anyActionListener());
     }
+
+    private String randomInternalUsername() {
```
n1v0lg, line 234: This gets pretty repetitive but it feels like too specific a thing to include in ESTestCase/create a util for

n1v0lg:
I'll check if we need to update anything around this in docs e.g. in https://www.elastic.co/guide/en/elasticsearch/reference/current/internal-users.html

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/esnative/NativeRealmIntegTests.java
```diff
@@ -269,7 +270,24 @@ public void testAddAndGetRole() throws Exception {
...
-    public void testAddUserAndRoleThenAuth() throws Exception {
+    public void testAddUserAndRoleThenAuth() {
+        testAddUserAndRoleThenAuth("joe");
+    }
+
+    public void testAddUserWithInternalUsernameAndRoleThenAuth() {
+        testAddUserAndRoleThenAuth(randomInternalUsername());
+    }
+
+    public void testAuthWithInternalUsernameFailsWithoutCorrespondingUser() {
```
n1v0lg, line 281: Maybe overly paranoid. LMK if I should drop this since it's quite heavy

n1v0lg:
@elasticmachine run elasticsearch-ci/part-1-fips plz

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/user/SystemUser.java
```diff
@@ -44,10 +46,15 @@ public static boolean is(User user) {
         return INSTANCE.equals(user);
     }
 
-    public static boolean is(String principal) {
+    public static boolean deserializesFrom(String principal) {
         return NAME.equals(principal);
     }
 
+    public static void write(StreamOutput output) throws IOException {
```
n1v0lg, line 53: Example change for `SystemUser`. Will apply to all other internal users if this looks good.

ywangd: I'd have this method with signature `public void writeTo(StreamOutput)` which feels more consistent with other places. Static `write` method often takes an extra argument which is the thing to be written.

ywangd:
> With internal usernames **not** being reserved anymore, I think it's quite error prone and misleading to have a `SystemUser.is(String username)` method; it implies that the system user _is_ defined by a username which is not correct. I see someone misunderstanding this method, and incorrectly using it resolve a non internal user to an internal one. Likewise, ideally we should encapsulate `NAME` as much as possible to prevent relying on it elsewhere.
> 
> The only place `is` and `SystemUser.NAME` are used is inside serialization. I think it makes sense to tweak reading and writing internal users so that `NAME` is not used anymore and `is` becomes `deserializesFrom` (open to better name suggestions!) to make it's scope more obvious.

I agree `SystemUser.is(String)` is now an awkward method to have. So I am all for changing it. But I don't think it's really necessary to hide the `NAME`. It is after all accessible via `SystemUser.INSTANCE.principal()`. So I think it is *not* that important. Instead of renaming `is` to `deserializesFrom`, I'd say we just inline this logic inside `AuthenticationSerializationHelper#readUserFrom()` and remove the method.

n1v0lg:
@ywangd thanks for the quick response! I thought about `SystemUser.INSTANCE.principal()` leaking the name as well. I'm good with keeping `NAME` exposed. I'll tweak this and apply the change to all other internal users 👍

n1v0lg:
@ywangd actually one more consideration for encapsulating NAME: to me it's analogous to `username` on the User object, i.e., an internal detail whereas, principal is a public method that currently happens to return username but doesn't have to. I don't feel strongly about this just wanted to raise this

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/user/SystemUser.java
n1v0lg: Making it not static would require using `INSTANCE`, which is of type `User` and so won't have the custom `writeTo` method exposed. 

I'll go back to inlining this in Authentication, along with inlining `is`. Will then revisit in a small follow up PR, if we decide that encapsulating `NAME` is worth it after all.

n1v0lg: We'd have to use it here:

https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java#L978-L979

i.e., 

```
if (SystemUser.is(user)) {
    SystemUser.INSTANCE.writeTo(output);
}
```

But `INSTANCE` is a `User` and does not expose `writeTo`

ywangd: > But `INSTANCE` is a `User` and does not expose `writeTo`

Right. That can be changed too now that User class is simplified (#86246). Previously, user can have inner user which causes recursion problem for a regular `writeTo` method. This is no longer the case. Also the `SystemUser.is(User)` method isn't necessary either. If we mark `SystemUser` as `final`, the `is` method can be replaced by `instanceOf` or maybe even object identity comparison. 

All is to say there are quite a few thing we can do to improve internal users. But again, I think we can revisit later.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/user/TransportDeleteUserActionTests.java
ywangd: You can add method to `AuthenticationTestHelper` to avoid repitition and it might also be a general useful utility. Alternatively, you can use `AuthenticationTestHelper.builder().internal().build().getUser().principal()`. But it is likely way too verbose.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/esnative/NativeRealmIntegTests.java
```diff
@@ -269,7 +270,24 @@ public void testAddAndGetRole() throws Exception {
...
+        String token = basicAuthHeaderValue(randomInternalUsername(), new SecureString("s3krit-password"));
+        ElasticsearchSecurityException e = expectThrows(
+            ElasticsearchSecurityException.class,
+            () -> client().filterWithHeader(Collections.singletonMap("Authorization", token)).prepareSearch("idx").get()
+        );
+        assertThat(e.status(), is(RestStatus.UNAUTHORIZED));
+    }
+
+    public void testAddUserAndRoleThenAuth(String username) {
```
ywangd, line 290:
```suggestion
    private void testAddUserAndRoleThenAuth(String username) {
```
? 
Otherwise, it shows up to be testable in IDE.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/esnative/NativeRealmIntegTests.java
ywangd: I like it.

docs/changelog/86398.yaml
```diff
@@ -0,0 +1,6 @@
+pr: 86398
+summary: User API actions allow use of internal usernames
```
ywangd, line 2: This will be in the release notes. So I'd avoid using "internal usernames" because (1) End-users likely won't understand what it means; and (2) We don't want to (unintentionally) encourage the usage of internal usernames. So can we have something more generic like: "Slightly relax username restrictions for User APIs"?

ywangd (approved):
LGTM

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportSetEnabledAction.java
```diff
@@ -51,9 +50,6 @@ protected void doExecute(Task task, SetEnabledRequest request, ActionListener<Ac
         if (securityContext.getUser().principal().equals(request.username())) {
             listener.onFailure(new IllegalArgumentException("users may not update the enabled status of their own account"));
             return;
```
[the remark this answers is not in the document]
n1v0lg: uff nice catch! Will address in a follow up PR 👍

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/esnative/NativeRealmIntegTests.java
n1v0lg: Cheers, coulda sworn I made this private!

docs/changelog/86398.yaml
n1v0lg: Sounds good! We document internal users here: https://www.elastic.co/guide/en/elasticsearch/reference/current/internal-users.html but it's true that for most end-users this will be a foreign concept. +1 on not encouraging `_xpack` users for now reason 😅 

I will go with "Relax username restrictions for User APIs"

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/user/TransportDeleteUserActionTests.java
n1v0lg: Perfect place for it, thanks for the pointer.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportSetEnabledAction.java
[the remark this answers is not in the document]
n1v0lg: > Also, since we prevent user from disabling itself, why should we allow it delete itself (TransportDeleteUserAction)?

Agreed that this is inconsistent. Two options: remove the restriction in set enabled or add it to `TransportDeleteUserAction`. Former isn't ideal since it seems like a sensible safeguard (plus prevents users from re-enabling themselves). Latter makes sense but would be a breaking change. Not sure if it's worth the hassle tbh.

n1v0lg:

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportSetEnabledAction.java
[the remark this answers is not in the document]
ywangd: Let's be cautions about changing the behaviour of deleting users. I was merely raising the question, not a proposal. But the set enabled user behaviour is a bug and should be fixed (by checking realm alongside username). Btw, users cannot re-enable themseleves even without this check. Disabled users fail at authentication time

n1v0lg:
@elasticmachine run elasticsearch-ci/part-2-fips

n1v0lg:

PR #86399, opened 2022-05-03
[ML] Adjust automatic JVM heap sizing for dedicated ML nodes

droberts195:
Previously the JVM heap size on dedicated ML nodes was as follows
when automatic JVM heap sizing was used:

- 40% of memory if total memory < 2GB
- 25% of memory for total memory between 2GB and 8GB
- 2GB if total memory > 8GB

This formula had two major drawbacks:

1. There was no way to increase the ML node JVM size beyond 2GB.
   If for some reason 2GB wasn't enough (e.g. a huge cluster state)
   then there was no workaround.
2. The formula was not reversible - there are JVM heap sizes that
   could arise from 2 different total memory sizes. This complicated
   autoscaling calculations.

This PR switches to a simpler formula that does not suffer these
drawbacks. The new formula is:

JVM heap size = 40% of total memory up to 16GB
              + 10% of total memory over 16GB

This is always rounded down to the next lower multiple of 4 megabytes.
This is because Java itself will round to a multiple of 4 megabytes if the
provided argument isn't already rounded, and we want to minimise
discrepancies.

It's now possible to increase the JVM heap size on dedicated ML nodes
indefinitely by increasing the node size. And there are no nasty
discontinuities in the formula that make the inverse hard to calculate.


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/NativeMemoryCapacity.java
```diff
@@ -7,39 +7,60 @@
...
+            capacity.tierMlNativeMemoryRequirementExcludingOverhead,
+            capacity.nodeMlNativeMemoryRequirementExcludingOverhead,
+            capacity.jvmSize
+        );
     }
 
-    private final long tierMlNativeMemoryRequirement;
-    private final long nodeMlNativeMemoryRequirement;
+    private final long tierMlNativeMemoryRequirementExcludingOverhead;
+    private final long nodeMlNativeMemoryRequirementExcludingOverhead;
```
droberts195, line 39: This change to doing capacity calculations in terms of memory _excluding_ the per-node overhead considerably simplifies the calculations.  It allows other code to avoid the need to think about how many nodes will be added on a scale-up.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java
```diff
@@ -813,11 +831,13 @@ Optional<AutoscalingDeciderResult> checkForScaleUp(
             NativeMemoryCapacity updatedCapacity = NativeMemoryCapacity.from(currentScale)
                 .merge(analyticsCapacityAndNewLoad.v1())
                 .merge(anomalyCapacityAndNewLoad.v1())
-                .merge(modelCapacityAndNewLoad.v1())
-                // Since we require new capacity, it COULD be we require a brand new node
-                // We should account for overhead in the tier capacity just in case.
-                .merge(new NativeMemoryCapacity(MachineLearning.NATIVE_EXECUTABLE_CODE_OVERHEAD.getBytes(), 0));
```
droberts195, line 819: While it _COULD_ be the case that we require a brand new node, it also _COULD_ be the case that we don't.  Including the overhead here unconditionally was the case of an inconsistency that could lead to scale up/scale down loops.  Fixing this is a nice side effect of changing native capacity to be memory _excluding_ the per-node overhead.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/action/StartTrainedModelDeploymentAction.java
```diff
@@ -47,10 +47,14 @@ public class StartTrainedModelDeploymentAction extends ActionType<CreateTrainedM
...
-     * We also subtract 30MB that we always add as overhead (see MachineLearning.NATIVE_EXECUTABLE_CODE_OVERHEAD).
+     * 30MB of this is accounted for via the space set aside to load the code into memory
+     * that we always add as overhead (see MachineLearning.NATIVE_EXECUTABLE_CODE_OVERHEAD).
+     * That would result in 270MB here, although the problem with this is that it would
+     * mean few models would fit on a 2GB ML node in Cloud with logging and metrics enabled.
+     * Therefore we push the boundary a bit and require a 240MB per-model overhead.
      * TODO Check if it is substantially different in other platforms.
      */
-    private static final ByteSizeValue MEMORY_OVERHEAD = ByteSizeValue.ofMb(270);
+    private static final ByteSizeValue MEMORY_OVERHEAD = ByteSizeValue.ofMb(240);
```
droberts195, line 57: This relates to the discussion about what models will fit on a 2GB Cloud node.  I can pull this out into a separate PR to merge first so that the history is clearer in the future.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -537,7 +537,7 @@ Optional<String> handleNodeLoad(NodeLoad load, String nodeId, StartTrainedModelD
                 )
             );
         }
-        if (load.getFreeMemory() < params.estimateMemoryUsageBytes()) {
```
droberts195, line 540: This is a bug fix which I can pull out into a separate PR and merge to 8.2 as well.  The fix would look slightly different in 8.2, but never considering the per-node native code overhead here is a bug in released product.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/NativeMemoryCapacity.java
```diff
@@ -48,61 +69,151 @@ public NativeMemoryCapacity(long tierMlNativeMemoryRequirement, long nodeMlNativ
...
-        // necessarily what "auto" will imply after scaling. Because the JVM occupies a
-        // smaller proportion of memory the bigger the node, the memory percent might be
-        // higher than we calculate here, potentially resulting in a bigger ML tier than
-        // required. The effect is most pronounced when the minimum node size is small and
-        // the minimum tier size is a lot bigger. For example, if the minimum node size is
-        // 1GB and the total native ML memory requirement for the tier is 32GB then the memory
-        // percent will be 41%, implying a minimum ML tier size of 78GB. But in reality a
-        // single 64GB ML node would have an auto memory percent of 90%, and 90% of 64GB is
-        // plenty big enough for the 32GB of ML native memory required.
-        // TODO: improve this in the next refactoring
```
droberts195, line 88: This problem is fixed in this PR.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/NativeMemoryCapacity.java
```diff
@@ -48,61 +69,151 @@ public NativeMemoryCapacity(long tierMlNativeMemoryRequirement, long nodeMlNativ
...
-        double memoryPercentForMl = NativeMemoryCalculator.modelMemoryPercent(requiredNodeSize, jvmSize, maxMemoryPercent, useAuto);
-        double inverseScale = memoryPercentForMl <= 0 ? 0 : 100.0 / memoryPercentForMl;
-        long requiredTierSize = Math.round(Math.ceil(tierMlNativeMemoryRequirement * inverseScale));
```
droberts195, line 91: The new code in this method doesn't use floating point arithmetic. I found it caused unpredictable rounding problems because multi-gigabyte byte values we're working with are too close to the maximum 15 digit precision of `double`.  It's safer to stick to integer arithmetic.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/NativeMemoryCapacity.java
```diff
@@ -48,61 +69,151 @@ public NativeMemoryCapacity(long tierMlNativeMemoryRequirement, long nodeMlNativ
      * @return a new instance with the merged capacity values
      */
     NativeMemoryCapacity merge(final NativeMemoryCapacity nativeMemoryCapacity) {
-        if (this == nativeMemoryCapacity) return this;
-        long tier = this.tierMlNativeMemoryRequirement + nativeMemoryCapacity.tierMlNativeMemoryRequirement;
-        long node = Math.max(nativeMemoryCapacity.nodeMlNativeMemoryRequirement, this.nodeMlNativeMemoryRequirement);
-        // If the new node size is bigger, we have no way of knowing if the JVM size would stay the same
-        // So null out
```
droberts195, line 55: I think this was another bug in the old code.  If either node or tier size is different (higher or lower) then the JVM size could change after scaling.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/NodeAvailabilityZoneMapper.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Maintains a list of the nodes in each availability zone, where an availability zone
+ * means a unique combination of values of the attributes listed in the cluster setting
+ * cluster.routing.allocation.awareness.attributes.
+ */
+public class NodeAvailabilityZoneMapper implements ClusterStateListener {
```
droberts195, line 33: I can factor out adding this into a separate PR for future clarity.  But I wanted to develop it initially in conjunction with this change so it got a proper workout.

droberts195 (commented):
Some notes for reviewers.

Like I say, I can pull out bits of this into separate PRs, so feel free to leave review of this PR for after that.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java
```diff
@@ -863,19 +888,16 @@ Optional<AutoscalingDeciderResult> checkForScaleUp(
                 // we have no info, allow the caller to make the appropriate action, probably returning a no_scale
                 return Optional.empty();
             }
-            long newTierNeeded = 0L;
+            long newTierNeeded = -futureFreedCapacity.getTierMlNativeMemoryRequirementExcludingOverhead();
```
benwtrent, line 891: 👍

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java
```diff
@@ -896,17 +915,18 @@ Optional<AutoscalingDeciderResult> checkForScaleUp(
...
-                // Is there "future capacity" on a node that could run this job? If not, we need that much more in the tier.
-                if (futureFreedCapacity.getNodeMlNativeMemoryRequirement() < requiredMemory) {
-                    newTierNeeded = Math.max(requiredMemory, newTierNeeded);
-                }
+                newTierNeeded += requiredMemory;
                 newNodeMax = Math.max(newNodeMax, requiredMemory);
             }
-            if (newNodeMax > currentScale.getNodeMlNativeMemoryRequirement() || newTierNeeded > 0L) {
+            if (newNodeMax > currentScale.getNodeMlNativeMemoryRequirementExcludingOverhead() || newTierNeeded > 0L) {
                 NativeMemoryCapacity newCapacity = new NativeMemoryCapacity(newTierNeeded, newNodeMax);
```
benwtrent, line 922: May want to snap `newTierNeeded` to `0` if its below here?

benwtrent (commented):
Removing the overhead from being tracked in all the calculations is an excellent simplification. 

I had to read through the capacity calculations multiple times to track between the "tier level node size" and the "required node size" and how they are used separately. That calculation is indeed complicated now, but tier requirement sizes will be much more accurate.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/utils/NativeMemoryCalculator.java
```diff
@@ -175,34 +148,37 @@ public static long allowedBytesForMl(long machineMemory, int maxMemoryPercent, b
...
-            return Math.round((nativeAndOverhead / 0.6) * 0.4);
-        }
-        if (nativeAndOverhead < (ByteSizeValue.ofGb(8).getBytes() * 0.75)) {
-            return Math.round((nativeAndOverhead / 0.75) * 0.25);
+        if (nativeAndOverhead <= (JVM_SIZE_KNOT_POINT * 0.6)) {
+            return (nativeAndOverhead * 2 / 3 / BYTES_IN_MB) * BYTES_IN_MB;
         }
-        return STATIC_JVM_UPPER_THRESHOLD;
+        return Math.min(
+            ((long) (JVM_SIZE_KNOT_POINT * 0.4 + (nativeAndOverhead - JVM_SIZE_KNOT_POINT * 0.6) / 0.9 * 0.1) / BYTES_IN_MB) * BYTES_IN_MB,
```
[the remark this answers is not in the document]
droberts195: I think what's there is correct, but it's obviously not clear so I'll refactor it to make it clearer.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/utils/NativeMemoryCalculator.java
```diff
@@ -175,8 +176,9 @@ public static long dynamicallyCalculateJvmSizeFromNativeMemorySize(long nativeMa
         if (nativeAndOverhead <= (JVM_SIZE_KNOT_POINT * 0.6)) {
             return (nativeAndOverhead * 2 / 3 / BYTES_IN_MB) * BYTES_IN_MB;
         }
+        double nativeAndOverheadAbove16GB = nativeAndOverhead - JVM_SIZE_KNOT_POINT * 0.6;
```
benwtrent, line 179: Ah, I see, 60% of the 16gb is taken away to account for the rest of the bytes.

x-pack/plugin/ml/src/test/java/org/elasticsearch/xpack/ml/utils/NativeMemoryCalculatorTests.java
```diff
@@ -131,26 +132,14 @@ public void testAllowedBytesForMlWhenAutoIsTrue() {
...
-    public void testAllowedBytesForMlWhenBothJVMAndNodeSizeAreUnknown() {
-        int percent = randomIntBetween(5, 200);
-        DiscoveryNode node = newNode(null, null);
-        Settings settings = newSettings(percent, randomBoolean(), ByteSizeValue.ofMb(randomIntBetween(0, 5000)));
-        ClusterSettings clusterSettings = newClusterSettings(percent, randomBoolean(), ByteSizeValue.ofMb(randomIntBetween(0, 5000)));
-
-        assertThat(NativeMemoryCalculator.allowedBytesForMl(node, settings), equalTo(OptionalLong.empty()));
-        assertThat(NativeMemoryCalculator.allowedBytesForMl(node, clusterSettings), equalTo(OptionalLong.empty()));
-        assertThat(NativeMemoryCalculator.allowedBytesForMl(node, percent, randomBoolean()), equalTo(OptionalLong.empty()));
-    }
```
droberts195, line 152: This test is deleted because in 8.x it should be impossible that we don't know the JVM size for an ML node.  We add the attribute ourselves and have done since before 7.last, so an 8.x node will never be in a cluster with a node that didn't set it.

This also allows some assertions to be added and a `Long` changed to `long` which reduces the risk of subtle bugs being accidentally introduced.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java
```diff
@@ -922,52 +942,54 @@ Optional<AutoscalingDeciderResult> checkForScaleUp(
...
-    // This calculates the the following the potential future free capacity
-    // Since jobs with lookback only datafeeds, and data frame analytics jobs all have some potential future end date
-    // we can assume (without user intervention) that these will eventually stop and free their currently occupied resources.
-    //
-    // The capacity is as follows:
-    // tier: The sum total of the resources that will be eventually be available
-    // node: The largest block of memory that will be free on a given node.
-    // - If > 1 "batch" ml tasks are running on the same node, we sum their resources.
+    /**
+     * This calculates the the following the potential future free capacity
```
dimitris-athanasiou, line 946: Simplify this sentence? Perhaps just `This calculates the potential future free capacity.`

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java
```diff
@@ -922,52 +942,54 @@ Optional<AutoscalingDeciderResult> checkForScaleUp(
...
+     * Since jobs with lookback only datafeeds, and data frame analytics jobs all have some potential future end date
+     * we can assume (without user intervention) that these will eventually stop and free their currently occupied resources.
+     *
+     * The capacity is as follows:
+     * - tier: The sum total of the resources that will be eventually be available
```
dimitris-athanasiou, line 951:
```suggestion
     * - tier: The sum total of the resources that will eventually be available
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/NodeAvailabilityZoneMapper.java
```diff
@@ -0,0 +1,162 @@
...
+        Map<List<String>, Collection<DiscoveryNode>> updatedNodesByAvailabilityZone = new HashMap<>();
+        Map<List<String>, Collection<DiscoveryNode>> updatedMlNodesByAvailabilityZone = new HashMap<>();
+        for (DiscoveryNode node : nodes) {
+            List<String> orderedNodeAttributeValues = awarenessAttributes.stream()
+                .map(a -> node.getAttributes().get(a))
+                .filter(Objects::nonNull)
+                .toList();
+            if (orderedNodeAttributeValues.size() != awarenessAttributes.size()) {
+                // This indicates bad configuration - there shouldn't be nodes that don't have all the awareness attributes
+                logger.trace(
```
dimitris-athanasiou, line 141: Should this be a `warn` instead? It sounds like a serious enough misconfiguration that the user should find out.

dimitris-athanasiou (commented):
This is a good change! Left some minor comments.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/NodeAvailabilityZoneMapper.java
droberts195: It won't ever happen in Elastic Cloud.

It could easily happen in a self-managed cluster. If it's a warning it could spam the log continually.

The way it gets reported for shard allocation is that it's silently treated as a "NO" unless debug is enabled, in which case a debug message gets generated: https://github.com/elastic/elasticsearch/blob/788df35e339d5f1c90592d5e12c929b3f959bc78/server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/AwarenessAllocationDecider.java#L174

So maybe the equivalent here is to make this a `debug` log?

dimitris-athanasiou: Fair enough. Yeah, `debug` should be good enough.

droberts195: Done in c57a8b249e5bfd70b0cb1a14941449a3ded55215

dimitris-athanasiou (approved):
LGTM

benwtrent (approved):

droberts195:
I have just found that there's still a discrepancy in JVM sizing due to the fact that Java rounds `-Xmx` arguments to the next higher multiple of 4MB.  (#86431 is also related to this.)

I'm going to change the calculation of the JVM size on ML nodes to round _down_ to a whole multiple of 4MB so we avoid the discrepancy without further reducing the memory available for ML processes.

Unfortunately this means all those tests where I painstakingly checked the outputs by hand are going to need redoing. 😢

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/MachineDependentHeap.java
```diff
@@ -140,11 +140,13 @@ enum MachineNodeRole {
          *     <li>The absolute maximum heap size is 31 gigabytes.</li>
          * </ul>
          *
+         * In all cases the result is rounded down to the next whole multiple of 4 megabytes.
```
dimitris-athanasiou, line 143: Should we add here an explanation for why we're rounding down to the next multiple of 4 MB?

dimitris-athanasiou (commented):
Still LGTM, just a suggestion for adding to a comment.

droberts195:

PR #85491, opened 2022-03-30
Remove elasticsearch.rest-test gradle plugin

breskeby:
Removes and remaining usages of `elasticsearch.rest-test` and the plugin itself from the codebase

Relates to https://github.com/elastic/elasticsearch/issues/63696

PR #86400, opened 2022-05-03
Do not apply StandaloneRestTestPlugin in StandaloneTestPlugin

breskeby:
As we are moving away from StandaloneRestTestPlugin (see https://github.com/elastic/elasticsearch/pull/85491) we want to reduce the surface of that plugin.
If rest tests are used a rest test related plugin should be applied

### DIFF
--- a/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/test/StandaloneTestPlugin.java
+++ b/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/test/StandaloneTestPlugin.java
@@ -8,9 +8,16 @@
 
 package org.elasticsearch.gradle.internal.test;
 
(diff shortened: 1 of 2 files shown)


mark-vieira (commented):
We want to move away from _this_ plugin as well, right? I mean, I'm pretty sure we could just swap this out for the `elasticsearch.java` plugin everywhere and it would just work. The only side-effect is we would be creating a `main` sourceset that doens't have any source in it and so those tasks would be noops. Do you intend to do this in a follow up PR?

breskeby:
Long term yes but at the moment, `elasticsearch.java` is still doing a bit too much for a test only project imo as it also applies plugins, adds a lot unneeded task configurations. After we removed `elasticsearch.rest-test` and `standalone-rest-test` we can look into this. I think we might want something like `production` vs. `qa` plugins to differ between production code intended to be shipped and qa projects.

PR #86401, opened 2022-05-03
Allow serial_diff under min_doc_count aggs

nik9000:
Before 6.6 we allowed the `serial_diff` agg in lots of places, including
under `date_histogram` aggregations with `min_doc_count: 1`. This
allowed you to take the difference of two adjacent buckets, skipping any
that don't have any value. So if you have a value at 10am, no value at
11am, and another at noon the `serial_diff` will diff the 10am and noon
values. In 6.6 we disabled support for this. We'd like it back.


nik9000:
I took the opportunity to remove an annoying TODO with a bunch of `instanceof`s and add some yaml tests for serial_diff.

server/src/main/java/org/elasticsearch/search/aggregations/PipelineAggregationBuilder.java
```diff
@@ -216,6 +209,11 @@ public void addBucketPathValidationError(String error) {
          */
         public abstract void validateParentAggSequentiallyOrdered(String type, String name);
 
+        /**
+         * Validates that the parent is sequentially ordered.
```
not-napoleon, line 213: There's a method right above that has the same javadoc.  I know these two methods don't do the same thing, so they should explain how they're different.

server/src/main/java/org/elasticsearch/search/aggregations/PipelineAggregationBuilder.java
```diff
@@ -134,6 +131,11 @@ public void validateParentAggSequentiallyOrdered(String type, String name) {
                         + "] must have a histogram, date_histogram or auto_date_histogram as parent but doesn't have a parent"
                 );
             }
+
+            @Override
+            public void validateParentAggSequentiallyOrderedWithoutSkips(String type, String name) {
```
not-napoleon, line 136: This needs javadoc.  It's subtly different from the version in `ForInsideTree`, and it's not clear why they do two different things without loading the whole context.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java
```diff
@@ -501,4 +502,14 @@ public boolean equals(Object obj) {
...
         return Version.V_EMPTY;
     }
+
+    @Override
+    protected void validateSequentiallyOrdered(String type, String name, Consumer<String> addValidationError) {}
+
+    @Override
+    protected void validateSequentiallyOrderedWithoutGaps(String type, String name, Consumer<String> addValidationError) {
+        if (minDocCount != 0) {
+            addValidationError.accept("parent histogram of " + type + " aggregation [" + name + "] must have min_doc_count of 0");
```
not-napoleon, line 512: Do we need to think about extended/hard bounds here?  I'm pretty sure the answer is "no", but I want to say it out loud.

not-napoleon (approved):
LGTM.

sethmlarson (approved):
Looks good from an API perspective. Thanks for adding test cases!

server/src/main/java/org/elasticsearch/search/aggregations/PipelineAggregationBuilder.java
nik9000: :+1: I meant to do this but skipped it. thanks.

server/src/main/java/org/elasticsearch/search/aggregations/PipelineAggregationBuilder.java
nik9000: It always throws an exception. I'll jiggle it so it's more clear.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java
nik9000: We never have in the past. I think it's ok because those'll just make the result wider without adding any gaps.

nik9000:
run elasticsearch-ci/part-2

PR #86403, opened 2022-05-03
Use HashMap for IndexMetadata.inSyncAllocationIds

rjernst:
The in sync allocation ids is a mapping from shard to the set of ids
currently being processed. When being built, the metadata uses a sparse
map, only filling a value for a shard as they are put into the builder.
When the final metadata is built, the map is made dense.

This commit converts to using a HashMap instead of ImmutableOpenIntMap.
The boxed keys should not cause allocations, as long as number of shards
is lower than 128. Long term the map itself should become an array, as
we know the number of shards (much like the dense primaryTerms array
here). However, that change will be a little trickier to make, since we
will need to be backward compatible with how diffs are built, currently
using Map differences.

relates #86239


original-brownbear (approved):
LGTM. I'll try not to forget the follow up to move this to an array :)

PR #86404, opened 2022-05-03
Add preflight checks to Health API to ensure health is obtainable

jbaiera:
This PR introduces an idea of preflight health indicator services to the new health service. Preflight indicators are structurally identical to regular indicators, but they are executed first when calculating health and conditionally block downstream indicators from running on an unstable or unknown cluster state.

The health service is configured with an optional list of preflight indicators. If the preflight indicators are present, the health service will execute them first when calculating a health response. If any of these preflight indicators returns a non-green result status, then the remaining health indicators will not be executed. Instead, an UNKNOWN status will be created for them and returned in the response. If `computeDetails` flag is set, then UNKNOWN responses will contain details about which preflight indicator blocked its execution.


andreidan (commented):
Thanks for working on this Jimmy.

This is looking great.

Can we add one more check that the cluster state is RECOVERED ? If it is not everything should be UNKNOWN and we should recommend the user to retry calling the health api

server/src/test/java/org/elasticsearch/health/HealthServiceTests.java
```diff
@@ -114,11 +121,177 @@ public void testMissingComponentOrIndicator() {
...
+        assertThat(result.status(), is(equalTo(UNKNOWN)));
+        assertThat(result.summary(), is(equalTo(HealthService.UNKNOWN_RESULT_SUMMARY)));
+    }
+
+    public void testPreflightIndicatorFailureTriggersUnknownResults() {
+        var preflight1 = new HealthIndicatorResult("preflight1", "component1", RED, null, null, null, null);
+        var preflight2 = new HealthIndicatorResult("preflight2", "component2", GREEN, null, null, null, null);
+        var indicator1 = new HealthIndicatorResult("indicator1", "component1", GREEN, null, null, null, null);
+        var indicator2 = new HealthIndicatorResult("indicator2", "component1", YELLOW, null, null, null, null);
+        var indicator3 = new HealthIndicatorResult("indicator3", "component2", GREEN, null, null, null, null);
```
andreidan, line 194: Could we generally use real values in tests? 

Seeing a failure that has component2 and indicator3 in the error message is rather hard to "parse"

server/src/main/java/org/elasticsearch/health/HealthService.java
```diff
@@ -11,25 +11,42 @@
...
 
+    // Visible for testing
+    static final String UNKNOWN_RESULT_SUMMARY = "Could not determine indicator state";
+
+    private final List<HealthIndicatorService> preflightHealthIndicatorServices;
     private final List<HealthIndicatorService> healthIndicatorServices;
 
-    public HealthService(List<HealthIndicatorService> healthIndicatorServices) {
+    /**
+     * @param preflightHealthIndicatorServices indicators that are run first and block health calculation until GREEN.
```
andreidan, line 42: Can we be more verbose on the documentation here - I don't think these indicators "block" the health calculation per se but rather indicate a severe problem in the system (?)

server/src/test/java/org/elasticsearch/health/HealthServiceTests.java
jbaiera: How real should these be? I want to make sure that preflight indicators that are in the same component get correctly mixed together in the results, but the master indicator doesn't share a component with anything right now. For now I'll give them some hypothetical names, but if you'd like them to be more real than that we can discuss further.

jbaiera:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/HealthService.java
```diff
@@ -9,28 +9,70 @@
...
 
 /**
  * This service collects health indicators from all modules and plugins of elasticsearch
  */
 public class HealthService {
 
+    // Visible for testing
+    static final String UNKNOWN_RESULT_SUMMARY_PREFLIGHT_FAILED = "Could not determine indicator state. Cluster state is not stable. Check "
+        + "details for critical issues keeping this indicator from running.";
+    static final String UNKNOWN_RESULT_SUMMARY_NOT_RECOVERED = "Could not determine indicator state. Cluster state is not recovered. Try "
```
andreidan, line 40:
```suggestion
    static final String UNKNOWN_RESULT_SUMMARY_NOT_RECOVERED = "Could not determine indicator state. The current node handling the health request is not ready to assess the health of the cluster. Try "
```

server/src/main/java/org/elasticsearch/health/HealthService.java
```diff
@@ -9,28 +9,70 @@
...
+        + "again once cluster has been joined and state has been recovered.";
```
andreidan, line 41:
```suggestion
        + "again later or execute the health API against a different node.";
```

andreidan (approved):
LGTM, thanks for working on this Jimmy

andreidan:
Relates to #84792

jbaiera:
@elasticmachine update branch

PR #86405, opened 2022-05-03
Faster GET _cluster/settings API

### LINKED ISSUE #82342: `RestClusterGetSettingsAction` Requests the Full Metadata from Master

grcevski:
The existing API to retrieve the cluster settings relies on
pulling in the full cluster state, which can be very expensive.

This change adds a dedicated cluster settings API, avoiding
serializing the full cluster state.

Closes https://github.com/elastic/elasticsearch/issues/82342


server/src/main/java/org/elasticsearch/action/admin/cluster/settings/ClusterGetSettingsAction.java
```diff
@@ -0,0 +1,104 @@
...
+        public Settings persistentSettings() {
+            return persistentSettings;
+        }
+
+        public Settings transientSettings() {
+            return transientSettings;
+        }
+
+        public Settings settings() {
+            return Settings.builder().put(persistentSettings).put(transientSettings).build();
```
grcevski, line 101: I'm not sure this is the right way to get the combined settings, it didn't make sense to double serialize across the transport protocol.

DaveCTurner: It looks right:

https://github.com/elastic/elasticsearch/blob/579156717a284091dfe3810769a2a1194f6b76e8/server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java#L1875

However I think it would be fine (preferable even) to send the combined settings separately from the transient and persistent ones, rather than duplicating this logic and risking getting it wrong.

server/src/main/java/org/elasticsearch/action/admin/cluster/settings/ClusterGetSettingsAction.java
```diff
@@ -0,0 +1,116 @@
...
+import java.io.IOException;
+import java.util.Objects;
+
+import static org.elasticsearch.action.admin.cluster.settings.RestClusterGetSettingsResponse.PERSISTENT_FIELD;
+import static org.elasticsearch.action.admin.cluster.settings.RestClusterGetSettingsResponse.TRANSIENT_FIELD;
+
+public class ClusterGetSettingsAction extends ActionType<ClusterGetSettingsAction.Response> {
+
+    public static final ClusterGetSettingsAction INSTANCE = new ClusterGetSettingsAction();
+    public static final String NAME = "cluster:admin/settings/get";
```
DaveCTurner, line 32: I think this needs to be under `cluster:monitor/*` (maybe `cluster:monitor/settings`?) so that it's still permitted by clients that only have the monitor privilege.

server/src/main/java/org/elasticsearch/action/admin/cluster/settings/ClusterGetSettingsAction.java
```diff
@@ -0,0 +1,116 @@
...
+
+        public RequestBuilder(ElasticsearchClient client, ClusterGetSettingsAction action) {
+            super(client, action, new Request());
+        }
+    }
+
+    /**
+     * Response for cluster settings
+     */
+    public static class Response extends ActionResponse implements ToXContentObject {
```
DaveCTurner, line 67: Does this need to be `implements ToXContentObject`? I think we don't expose it as XContent directly, it's always converted to a `RestClusterGetSettingsResponse` first.

server/src/main/java/org/elasticsearch/action/admin/cluster/settings/ClusterGetSettingsAction.java
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Request to retrieve the cluster settings
+     */
+    public static class Request extends MasterNodeReadRequest<Request> {
+        public Request() {}
+
+        public Request(StreamInput in) throws IOException {
+            super(in);
+        }
```
DaveCTurner, line 46: Could we `assert in.getVersion().onOrAfter(Version.V_8_3_0)` to ensure the BwC logic is doing the right thing? Also in the `writeTo()` method, and similarly in the `Response` class.

server/src/main/java/org/elasticsearch/action/admin/cluster/settings/ClusterGetSettingsAction.java
```diff
@@ -0,0 +1,116 @@
...
+        @Override
+        public ActionRequestValidationException validate() {
+            return null;
+        }
+    }
+
+    /**
+     * Cluster get settings request builder
+     */
+    public static class RequestBuilder extends MasterNodeReadOperationRequestBuilder<Request, Response, RequestBuilder> {
```
DaveCTurner, line 57: I wonder, do we really need all this client-facing machinery? AFAIK this pattern was good for the transport client, but we have no transport client any more so I think we can just call the action directly in the one place it's currently used.

server/src/main/java/org/elasticsearch/action/admin/cluster/settings/TransportClusterGetSettingsAction.java
```diff
@@ -0,0 +1,64 @@
...
+            ClusterGetSettingsAction.NAME,
+            transportService,
+            clusterService,
+            threadPool,
+            actionFilters,
+            ClusterGetSettingsAction.Request::new,
+            indexNameExpressionResolver,
+            ClusterGetSettingsAction.Response::new,
+            ThreadPool.Names.SAME
+        );
```
DaveCTurner, line 46: This means that circuit breakers apply to this action, but they don't apply to the cluster state action or the REST action. I think we should continue to ignore circuit breakers.

```suggestion
        super(
            ClusterGetSettingsAction.NAME,
            false,
            transportService,
            clusterService,
            threadPool,
            actionFilters,
            ClusterGetSettingsAction.Request::new,
            indexNameExpressionResolver,
            ClusterGetSettingsAction.Response::new,
            ThreadPool.Names.SAME
        );
```

server/src/main/java/org/elasticsearch/client/internal/ClusterAdminClient.java
```diff
@@ -756,4 +757,26 @@ public interface ClusterAdminClient extends ElasticsearchClient {
...
      */
     ActionFuture<AcknowledgedResponse> deleteDanglingIndex(DeleteDanglingIndexRequest request);
+
+    /**
+     * The cluster settings.
+     *
+     * @param request The cluster settings request
+     * @return The result future
+     */
+    ActionFuture<ClusterGetSettingsAction.Response> clusterSettings(ClusterGetSettingsAction.Request request);
```
DaveCTurner, line 767: Similarly here, I think this stuff is all unused in practice and best dropped.

server/src/main/java/org/elasticsearch/rest/action/admin/cluster/RestClusterGetSettingsAction.java
```diff
@@ -50,20 +61,55 @@ public String getName() {
 
     @Override
     public RestChannelConsumer prepareRequest(final RestRequest request, final NodeClient client) throws IOException {
-        ClusterStateRequest clusterStateRequest = Requests.clusterStateRequest().routingTable(false).nodes(false);
         final boolean renderDefaults = request.paramAsBoolean("include_defaults", false);
-        clusterStateRequest.local(request.paramAsBoolean("local", clusterStateRequest.local()));
-        clusterStateRequest.masterNodeTimeout(request.paramAsTime("master_timeout", clusterStateRequest.masterNodeTimeout()));
+
+        if (nodesInCluster.get().getMinNodeVersion().onOrAfter(Version.V_8_3_0) == false) {
```
DaveCTurner, line 66:
```suggestion
        if (nodesInCluster.get().getMinNodeVersion().before(Version.V_8_3_0)) {
```

server/src/main/java/org/elasticsearch/rest/action/admin/cluster/RestClusterGetSettingsAction.java
```diff
@@ -50,20 +61,55 @@ public String getName() {
...
             );
     }
 
+    private RestChannelConsumer prepareLegacyRequest(final RestRequest request, final NodeClient client, final boolean renderDefaults) {
+        ClusterStateRequest clusterStateRequest = Requests.clusterStateRequest().routingTable(false).nodes(false);
+        clusterStateRequest.local(request.paramAsBoolean("local", clusterStateRequest.local()));
+        clusterStateRequest.masterNodeTimeout(request.paramAsTime("master_timeout", clusterStateRequest.masterNodeTimeout()));
+        return channel -> client.admin()
+            .cluster()
+            .state(clusterStateRequest, new RestToXContentListener<RestClusterGetSettingsResponse>(channel).map(response -> {
```
DaveCTurner, line 99: I think it would be slightly nicer to use the same `ActionListener<ClusterGetSettingsAction.Response>` for both the legacy and the regular case, with a `.map(clusterState -> new ClusterGetSettingsAction.Response(...))` to do the extra conversion step in the legacy case.

DaveCTurner (changes requested):
Looks good. I left some small comments.

In terms of tests, yes this seems mostly pretty well covered. I think we should have some tests that verify that we haven't mixed up the persistent and transient settings anywhere along the way, and that the new transport action applies settings filters correctly. Also I think we should have an `AbstractWireSerializingTestCase<ClusterGetSettingsAction.Response>` (including a `mutateInstance` implementation) to ensure it round-trips properly.

server/src/main/java/org/elasticsearch/action/admin/cluster/settings/ClusterGetSettingsAction.java
grcevski: Good idea, this way I can undo some of the changes I needed to do in xpack.

server/src/test/java/org/elasticsearch/action/admin/cluster/settings/ClusterGetSettingsTests.java
```diff
@@ -0,0 +1,88 @@
...
+                assertFalse(response.transientSettings().hasValue("transient.foo.filtered"));
+                assertTrue(response.transientSettings().hasValue("transient.foo.non_filtered"));
+            }
+
+            @Override
+            public void onFailure(Exception e) {
+                fail("Shouldn't reach here");
+            }
+        });
+    }
```
DaveCTurner, line 87: The trouble with this pattern is that it would pass even if the listener was not completed. There's various ways to avoid that, for instance using a future:

```suggestion
        final PlainActionFuture<ClusterGetSettingsAction.Response> future = new PlainActionFuture<>();
        action.masterOperation(null, null, clusterState, future);
        assertTrue(future.isDone());

        final ClusterGetSettingsAction.Response response = future.get();

        assertFalse(response.persistentSettings().hasValue("persistent.foo.filtered"));
        assertTrue(response.persistentSettings().hasValue("persistent.foo.non_filtered"));

        assertFalse(response.transientSettings().hasValue("transient.foo.filtered"));
        assertTrue(response.transientSettings().hasValue("transient.foo.non_filtered"));
```

server/src/main/java/org/elasticsearch/client/internal/ClusterAdminClient.java
```diff
@@ -756,4 +756,5 @@ public interface ClusterAdminClient extends ElasticsearchClient {
      * Delete specified dangling indices.
      */
     ActionFuture<AcknowledgedResponse> deleteDanglingIndex(DeleteDanglingIndexRequest request);
+
```
DaveCTurner, line 759: Nit: this file only contains whitespace changes now

server/src/main/java/org/elasticsearch/rest/action/admin/cluster/RestClusterGetSettingsAction.java
```diff
@@ -50,16 +60,46 @@ public String getName() {
...
+        ClusterGetSettingsAction.Request clusterSettingsRequest = new ClusterGetSettingsAction.Request();
+
+        clusterSettingsRequest.local(request.paramAsBoolean("local", clusterSettingsRequest.local()));
+        clusterSettingsRequest.masterNodeTimeout(request.paramAsTime("master_timeout", clusterSettingsRequest.masterNodeTimeout()));
+
+        return channel -> client.execute(
+            ClusterGetSettingsAction.INSTANCE,
+            clusterSettingsRequest,
+            new RestToXContentListener<RestClusterGetSettingsResponse>(channel).map(
+                r -> response(r, renderDefaults, settingsFilter, clusterSettings, settings)
```
DaveCTurner, line 78: I'd still sort of like to reduce the duplication here between the two branches but I see it's not so simple because we have to consume the request params before we get hold of the channel. Not a blocking suggestion.

DaveCTurner (approved):
LGTM

joegallo:

PR #86312, opened 2022-04-29
Fix test for Windows line endings

williamrandolph:
Adjustments to the MockTerminal class in #86198 and elsewhere have broken tests that depended on specific line endings. In many cases, this was not what the test was actually trying to check for. This fix makes the strings as system-agnostic as possible, in several cases using the `String#lines` method to handle line endings in a system-agnostic way.

Closes #86294. Closes #86341

PR #86408, opened 2022-05-03
Use Windows newlines when listing plugin information on Windows

### LINKED ISSUE #86352: Support Windows line separators in PluginInfo.toString
We have hardcoded Unix-style newlines in PluginInfo.toString. We should use `System.lineSeparator()` instead.

williamrandolph:
We hard-coded *nix line endings in the PluginInfo toString method, which means that the output would look garbled on Windows. We caught this when fixing up some Windows unit tests.

This PR makes things work for Windows. I also refactored for what I think is readability.

Fixes #86352 
Related #86312

### DIFF
--- a/server/src/main/java/org/elasticsearch/plugins/PluginInfo.java
+++ b/server/src/main/java/org/elasticsearch/plugins/PluginInfo.java
@@ -445,52 +445,34 @@ public class PluginInfo implements Writeable, ToXContentObject {
     }
 
     public String toString(String prefix) {
-        final StringBuilder information = new StringBuilder().append(prefix)
-            .append("- Plugin information:\n")
-            .append(prefix)
-            .append("Name: ")
-            .append(name)
-            .append("\n")
-            .append(prefix)
-            .append("Description: ")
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/elasticsearch/plugins/PluginInfo.java
```diff
@@ -445,52 +445,34 @@ public String toString() {
...
 
-        information.append(prefix)
-            .append("Extended Plugins: ")
-            .append(extendedPlugins)
-            .append("\n")
-            .append(prefix)
-            .append(" * Classname: ")
-            .append(classname);
+        addln(information, prefix, "Extended Plugins: ", extendedPlugins);
+        add(information, prefix, " * Classname: ", classname);
```
rjernst, line 465: You could avoid this one off method by using a `List<String> lines` instead of a StringBuilder, then returning `String.join(System.lineSeparator(), lines);`

rjernst (approved):
Looks good, one suggestion.

PR #86410, opened 2022-05-03
Health API explain query param

### LINKED ISSUE #86215: Add an include_details query param to the API
As the API is designed for user consumption (but can also be polled by various monitoring tools)
we have to balance the amount of work the API does in terms of RCA vs returning as much information 
as possible (ie. the details)

This proposes the introduction of an `include_details` query parameter that defaults to `true`.

This will be changed to `false` when the API is polled, but the users can simply call `_internal/_health` to
retrieve all the details.

On drill down (when the details are included by virtue of drilling down ie. `_internal/_health/data`) if
`include_details` contradicts the default drill down behaviour ( ie. `_internal/_health/data?include_details=false`) the 
details will be ommited (ie. `include_details` overrides the drill down behaviour)

masseyke:
The health API has a notion of details within each health indicator that is returned. These details can sometimes be
expensive to compute or transfer. This change allows a user to specify whether the details are generated and returned. By
default now all details are generated and returned (previously this was only the case if a component was specified in the
request). This behavior can be changed with the `explain` query param.
Closes #86215


andreidan (commented):

This LGTM, with the only mention that Leaf proposed we name the parameter `explain`.
(as the user actions are not under details but are controlled by this parameter `include_details` is a bit misleading).

Can you please rename the query param?
Can you please add this to the documentation too? https://github.com/elastic/elasticsearch/blob/master/docs/reference/health/health.asciidoc

masseyke:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/GetHealthAction.java
```diff
@@ -124,22 +124,22 @@ public String toString() {
     }
 
     public static class Request extends ActionRequest {
-        private final boolean computeDetails;
         private final String componentName;
         private final String indicatorName;
+        private final boolean includeDetails;
```
andreidan, line 129: I believe this should be named `explain` too? 

Together with the entire concept of `computeDetails` - https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/health/HealthService.java#L90 (and onwards)

I think dropping the `include...` and `compute...` concepts would simplify things.

What do you think?

masseyke: I had intentionally isolated the change to `explain` to the REST layer because it just doesn't make as much sense to me as include/computeDetails.

masseyke: Alright, I've change includeDetails and compuateDetails to explain everywhere.

andreidan (approved):
LGTM thanks for working on this Keith 🚀

PR #86414, opened 2022-05-04
DONE should mean fully processed in snapshot status

DaveCTurner:
Today in an active snapshot the snapshot status API sometimes shows
shards in state `DONE` but with `processed` stats that do not match the
`incremental` stats that show the work to be done. The discrepancy is
due to files that are stored in the shard snapshot metadata rather than
in their own blobs. This commit fixes the discrepancy.

It also addresses a test bug in `testConcurrentCreateAndStatusAPICalls`
which wasn't actually running the status API concurrently with the
snapshot creations.


server/src/internalClusterTest/java/org/elasticsearch/snapshots/SnapshotStatusApisIT.java
```diff
@@ -630,33 +633,61 @@ public void testGetSnapshotsRequest() throws Exception {
...
+        for (final var create : creates) {
+            create.get(10, TimeUnit.SECONDS);
+        }
+
+        var masterClusterService = internalCluster().getCurrentMasterNodeInstance(ClusterService.class);
+        assertBusy(() -> {
+            final var stillRunning = masterClusterService.state()
+                .custom(SnapshotsInProgress.TYPE, SnapshotsInProgress.EMPTY)
+                .isEmpty() == false;
             statuses.add(dataNodeClient.admin().cluster().prepareSnapshotStatus(repoName).setSnapshots(snapshotNames).execute());
```
original-brownbear, line 669: Could we retain the old loop as well? I'm not sure this will trigger the level of concurrency we needed to reproduce the original bug this was supposed to reproduce (the point of this test really is running enough gets and status requests in parallel so that the snapshot meta pool is fully occupied).
Running both loops should be fine IMO, this will only run 16 times at the most over 60s so running another 10-20 extra requests won't hurt IMO.

original-brownbear (approved):
LGTM thanks! One ask on the test that doesn't need another round of reviewing.

server/src/internalClusterTest/java/org/elasticsearch/snapshots/SnapshotStatusApisIT.java
DaveCTurner: Sure 7ec8e1e1ec471c68a1d8568de627db9b9889cbbd reinstates this (and fixes the number of iterations so it actually does what you said it should :grin:)

DaveCTurner: It strikes me that this bug might also need us to max out the threadpool after the snapshots all completed too - see 99cbdbb.

server/src/internalClusterTest/java/org/elasticsearch/snapshots/SnapshotStatusApisIT.java
```diff
@@ -630,33 +636,77 @@ public void testGetSnapshotsRequest() throws Exception {
...
+        final var waitForCompletion = randomBoolean();
+        final var createsListener = new PlainActionFuture<Void>();
+        final var createsGroupedListener = new GroupedActionListener<CreateSnapshotResponse>(
+            createsListener.map(ignored -> null),
+            snapshotNames.length
+        );
+        for (final var snapshotName : snapshotNames) {
+            clusterAdmin().prepareCreateSnapshot(repoName, snapshotName)
+                .setWaitForCompletion(waitForCompletion)
+                .execute(ActionListener.notifyOnce(createsGroupedListener));
```
original-brownbear, line 669: Sorry commented in the wrong spot:

why notifyOnce?

server/src/internalClusterTest/java/org/elasticsearch/snapshots/SnapshotStatusApisIT.java
original-brownbear: That's a nice idea, I think in practice it's good enough as is but this makes for a lot more action on the meta pool for sure :)

server/src/internalClusterTest/java/org/elasticsearch/snapshots/SnapshotStatusApisIT.java
DaveCTurner: Nvm, paranoia about double-notifying actions since `GroupedActionListener` only works by counting. But then I see it'd run out of result slots anyway.

PR #86415, opened 2022-05-04
[ML] Adding a node availability zone calculator

droberts195:
This is a utility class to help with node assignment and autoscaling
decisions in environments that have a concept of availability zones,
i.e. zones within the cluster where it's more likely to simultaneously
lose all nodes together due to hardware failure.

It maintains a list of the nodes in each availability zone, where an
availability zone means a unique combination of values of the attributes
listed in the cluster setting cluster.routing.allocation.awareness.attributes.

It's also possible to quickly find out how many availability zones there
are in total, and how many of them have ML nodes.


dimitris-athanasiou (approved):
LGTM

PR #86423, opened 2022-05-04
[ML] fix distribution change check for change_point aggregation

benwtrent:
There are certain scenarios where using apache math's built in ksTest check fails.

When `strict` is `true`, it requires the probability to be expressed as an inequality and typically causes `0` to be the pvalue calculated. This is most likely due to numerical resolution problems.

switching to `strict: false` fixes this. We only want to return that a change point is a distribution change if no only change type could be found or other change type pValues were large.

### DIFF
--- a/docs/changelog/86423.yaml
+++ b/docs/changelog/86423.yaml
@@ -1,0 +1,5 @@
+pr: 86423
+summary: Fix distribution change check for `change_point` aggregation
+area: Machine Learning
+type: bug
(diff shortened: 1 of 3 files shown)


tveasey (approved):
10000 is very safe regarding switching over to approximate KS; I would think 1000 is also indistinguishable. That said all looks fine to me and happy to stick to this since you've tested.

PR #86427, opened 2022-05-04
Remove remaining ImmutableOpen maps from shard stores response

rjernst:
The IndicesShardStoresResponse had ImmutableOpenInt map removed, but its
use accidentally remained in serialization. Thankfully the format for
immutable open maps are interchangable with read java Map. This commit
switches to using the new readImmutableMap to read the store statuses.

relates #86239

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/admin/indices/shards/IndicesShardStoresResponse.java
+++ b/server/src/main/java/org/elasticsearch/action/admin/indices/shards/IndicesShardStoresResponse.java
@@ -12,8 +12,6 @@ import org.elasticsearch.ElasticsearchException;
 import org.elasticsearch.action.ActionResponse;
 import org.elasticsearch.action.support.DefaultShardOperationFailedException;
 import org.elasticsearch.cluster.node.DiscoveryNode;
-import org.elasticsearch.common.collect.ImmutableOpenIntMap;
-import org.elasticsearch.common.collect.ImmutableOpenMap;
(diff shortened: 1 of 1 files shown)


idegtiarenko (approved):

server/src/main/java/org/elasticsearch/action/admin/indices/shards/IndicesShardStoresResponse.java
```diff
@@ -241,19 +239,15 @@ public IndicesShardStoresResponse(Map<String, Map<Integer, List<StoreStatus>>> s
...
-            int indexEntries = i.readVInt();
-            ImmutableOpenIntMap.Builder<List<StoreStatus>> shardEntries = ImmutableOpenIntMap.builder();
-            for (int shardCount = 0; shardCount < indexEntries; shardCount++) {
-                shardEntries.put(i.readInt(), i.readList(StoreStatus::new));
-            }
-            return shardEntries.build();
-        });
+        storeStatuses = in.readImmutableMap(
+            StreamInput::readString,
+            i -> i.readImmutableMap(StreamInput::readInt, j -> j.readList(StoreStatus::new))
```
idegtiarenko, line 249: It would be great to have a method to read immutable list as well if possible

rjernst: I opened https://github.com/elastic/elasticsearch/pull/86433

original-brownbear (approved):

PR #86433, opened 2022-05-04
Add immutable list reader to StreamInput

rjernst:
This commit adds a utility method for reading an immutable list to
StreamInput.

relates #86427

### DIFF
--- a/server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
+++ b/server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
@@ -1150,6 +1150,27 @@ public abstract class StreamInput extends InputStream {
     }
 
     /**
+     * Reads an list of objects. The list is expected to have been written using {@link StreamOutput#writeList(List)}.
+     * The returned list is immutable.
+     *
+     * @return the list of objects
+     * @throws IOException if an I/O exception occurs reading the list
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
```diff
@@ -1149,6 +1149,27 @@ public <T> List<T> readList(final Writeable.Reader<T> reader) throws IOException
...
+    /**
+     * Reads an list of objects. The list is expected to have been written using {@link StreamOutput#writeList(List)}.
+     * The returned list is immutable.
+     *
+     * @return the list of objects
+     * @throws IOException if an I/O exception occurs reading the list
+     */
+    public <T> List<T> readImmutableList(final Writeable.Reader<T> reader) throws IOException {
+        int count = readArraySize();
+        if (count == 0) {
```
original-brownbear, line 1161: I think it makes sense to special case `count == 1` and `== 2` as well here. These are quite common and the JDK does the same in `java.util.List#of(E...)` so I'd assume that's the right level of granularity?

original-brownbear (approved):
LGTM (after fixing check style ;)). Also one optional suggestion that is somewhat debatable, but this is fine with me as well if you disagree.

PR #86434, opened 2022-05-04
Add support for CPU ranges in desired nodes

fcofdez:
This commit adds support for CPU ranges in the desired nodes API. 

This aligns better with environments where administrators/orchestrators
can define lower and upper bounds for the amount of CPUs that the
desired node would get once deployed. 

This allows to provide information about the expected CPU and possible
allowed overcommit that the desired node will run on.

This was the previous expected body for the desired nodes API (we still support it):
```
PUT /_internal/desired_nodes/history/1
{
    "nodes" : [
        {
            "settings" : {
                 "node.name" : "instance-000187",
                 "node.external_id": "instance-000187",
                 "node.roles" : ["data_hot", "master"],
                 "node.attr.data" : "hot",
                 "node.attr.logical_availability_zone" : "zone-0"
            },
            "processors" : 8, 
            "memory" : "58gb",
            "storage" : "1700gb",
            "node_version" : "8.3.0"
        }
    ]
}
```

Now it's possible to define `processors` or `processors_range` as in:
```
PUT /_internal/desired_nodes/history/1
{
    "nodes" : [
        {
            "settings" : {
                 "node.name" : "instance-000187",
                 "node.external_id": "instance-000187",
                 "node.roles" : ["data_hot", "master"],
                 "node.attr.data" : "hot",
                 "node.attr.logical_availability_zone" : "zone-0"
            },
            "processors_range" : {"min": 8.0, "max": 16.0},
            "memory" : "58gb",
            "storage" : "1700gb",
            "node_version" : "8.3.0"
        }
    ]
}
```
Note that `max` in `processors_range` is optional.

This commit also moves from representing CPUs as integers to
accept floating point numbers.

Note: I disabled the bwc yamlRestTests for versions < 8.3 since we introduced
a few "breaking changes" but since this is an internal API it should be fine.


sethmlarson (approved):
LGTM, two questions:

- Are the `memory` and `storage` fields going to also have this change eventually?
- Can the `processors.min` be set without `processors.max` or is the value always required?

fcofdez:
Thanks for the review @sethmlarson

> Are the memory and storage fields going to also have this change eventually?

It might change as more internal users adopt this API.

> Can the processors.min be set without processors.max or is the value always required?

When `processors` is represented as an object, `min` and `max` are required. 
The API still supports the old version, where you can specify the number of processors as a number. In that case `min` and `max` are the supplied value.

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/cluster.desired_nodes/10_basic.yml
```diff
@@ -774,3 +767,135 @@ teardown:
             - { settings: { "node.external_id": "instance-000187"}, processors: 64, memory: "1b", storage: "1b", node_version: null }
   - match: { status: 400 }
   - match: { error.type: x_content_parse_exception }
+---
+"Test update desired nodes using legacy processors field":
```
henningandersen, line 771: I would not call this legacy, the min/max is not really the desired way to use this.

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/cluster.desired_nodes/10_basic.yml
```diff
@@ -774,3 +767,135 @@ teardown:
...
+  - skip:
+      reason: "contains is a newly added assertion"
+      features: contains
```
henningandersen, line 774: It looks like we are not using `contains` here (as well as the rest of the new tests)?

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNode.java
```diff
@@ -226,4 +251,88 @@ public String toString() {
...
+            if (out.getVersion().onOrAfter(MIN_MAX_PROCESSORS_VERSION)) {
+                out.writeFloat(min);
+                out.writeFloat(max);
+            } else {
+                out.writeInt(roundedMin());
+            }
+        }
+
+        public int roundedMin() {
+            return Math.round(min);
```
henningandersen, line 322: I wonder if we should round this down and round max up? That seems more intuitive, since then the actual quota of processors is still within the range.

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNode.java
```diff
@@ -86,21 +91,23 @@ private static Version parseVersion(String version) {
     }
 
     private final Settings settings;
-    private final int processors;
+    private final Processors processors;
```
henningandersen, line 94: I am not sure this will work in a mixed cluster, since we are changing the type of a field stored in cluster state (regardless of whether the new API has been used) and an old version cannot read it. I think we need a rolling upgrade test. Also, I wonder if we should simply make it two fields instead like `processors` and `processors_range` - to simplify parsing both for ourselves and clients? We could still make the two fields mutually exclusive.

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNode.java
```diff
@@ -226,4 +251,88 @@ public String toString() {
...
+
+        public static final ConstructingObjectParser<Processors, String> PROCESSORS_PARSER = new ConstructingObjectParser<>(
+            "processors",
+            false,
+            (args, name) -> new Processors((float) args[0], (float) args[1])
+        );
+
+        static {
+            PROCESSORS_PARSER.declareFloat(ConstructingObjectParser.constructorArg(), MIN_FIELD);
+            PROCESSORS_PARSER.declareFloat(ConstructingObjectParser.constructorArg(), MAX_FIELD);
```
henningandersen, line 269: Should this be optional? I think k8s might only have requests set, leading to no max cpu. In that case we would still like to know the minimum value.

henningandersen (commented):
Did an initial read and left a few comments.

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNode.java
fcofdez: I made the assumption that we would be able to break this since we don't have known users of this API, but that might be a risky assumption. I'm wondering if we should prevent upgrades during mixed-version clusters (at least when there are 8.2 and 8.3 master nodes) as we should be able to accept a processor range but the older nodes won't have that notion and we would have to send them the lower bound, maybe it's better to reject the updates in that phase to simplify things? wdty @henningandersen

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/cluster.desired_nodes/10_basic.yml
fcofdez: Copy-paste... 🤦

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNode.java
henningandersen: > but that might be a risky assumption

Yes, it can be a bit hard to control that it is never called on an 8.2 cluster. And if so, it will essentially break the cluster during upgrade, which sounds bad.

> we would have to send them the lower bound, maybe it's better to reject the updates in that phase to simplify things?

It sounds reasonable to me to reject a range specification during an upgrade to the release allowing the range specification, since otherwise we would be throwing away data.

I think we need to still accept a single processors specification in a mixed cluster though, i.e., as long as the client conforms to the old version API, everything should still work.

barkbay (commented):
I made a few tests using a dev version of the ECK operator (from this [PR](https://github.com/elastic/cloud-on-k8s/pull/5650)) and it works great 👍 

Some comments/questions:

* Setting a value of `0` for the `max` CPU capacity does not seem to clear a previous max value. It also does not raise an error as I _think_ it is the case in the current implementation. I'm wondering if setting a value of `0` should raise an error as it is the case now? (and also maybe explain it in the documentation?)

* I also want to double check the units: if a memory capacity of `2333Mi` is set in the K8S resource, it is translated by the ECK operator as `2446327808b`:

In K8S:

```yaml
  resources:
    requests:
      memory: 2333Mi
      cpu: 2222m
    limits:
      memory: 2333Mi
      cpu: 3141m
```

This the content of the API request by the ECK operator:

`PUT /_internal/desired_nodes/history/version`
```json
"processors_range": {
	"min": 2.222,
	"max": 3.141
},
"memory": "2446327808b"
```

It is however then stored in Elasticsearch as `2.2gb` in the desired node state:

`GET _internal/desired_nodes/_latest` 
```json
      "processors_range" : {
        "min" : 2.222,
        "max" : 3.141
      },
      "memory" : "2.2gb"
```

 I just want to be sure it is the expected result given that it should be either `2.2783203125 GiB` or `2.446327808 GB`. I think this is because `"2.2gb"` actually means `"2.2GiB"`?

Thanks!

server/src/test/java/org/elasticsearch/cluster/metadata/DesiredNodeTests.java
```diff
@@ -108,4 +151,79 @@ public void testGetRoles() {
...
+
+        {
+            final var desiredNode = new DesiredNode(settings, 2.0f, ByteSizeValue.ofGb(1), ByteSizeValue.ofGb(1), Version.CURRENT);
+            assertThat(desiredNode.isCompatibleWithVersion(Version.V_8_2_0), is(equalTo(true)));
+            assertThat(desiredNode.isCompatibleWithVersion(Version.V_8_3_0), is(equalTo(true)));
+        }
+    }
+
+    private Float randomInvalidFloatProcessor() {
+        return randomFrom(-1.0f, Float.NaN, Float.POSITIVE_INFINITY, Float.NEGATIVE_INFINITY);
```
henningandersen, line 227: Should we also include 0.0f?

server/src/test/java/org/elasticsearch/cluster/metadata/DesiredNodesTestCase.java
```diff
@@ -55,18 +55,23 @@ public static DesiredNode randomDesiredNodeWithRandomSettings() {
...
-        return randomDesiredNode(version, randomIntBetween(1, 256), settingsProvider);
+        return switch (randomInt(2)) {
+            case 0 -> randomDesiredNode(version, randomProcessor(), settingsProvider);
+            case 1 -> randomDesiredNode(version, randomIntBetween(1, 256), settingsProvider);
+            case 2 -> randomDesiredNode(version, randomIntBetween(1, 256) + randomFloat(), settingsProvider);
+            default -> throw new IllegalStateException();
+        };
     }
 
     public static DesiredNode randomDesiredNode(Version version, int processors, Consumer<Settings.Builder> settingsProvider) {
```
henningandersen, line 70: Can we simply remove this overload?

henningandersen (commented):
This looks mostly good. My only real concerns is with how `ProcessorsRange.max()` returns the `min` value if not specified, that looks wrong to me.

fcofdez:
Thanks for testing the PR @barkbay!

> Setting a value of 0 for the max CPU capacity does not seem to clear a previous max value. It also does not raise an error as I think it is the case in the current implementation. I'm wondering if setting a value of 0 should raise an error as it is the case now? (and also maybe explain it in the documentation?)

This is strange, it should fail (I even added a couple of tests that cover that case). When you mention that it does not clear the previous max value, do you mean that the Desired nodes remain in the previous version? 

> I also want to double check the units: if a memory capacity of 2333Mi is set in the K8S resource, it is translated by the ECK operator as 2446327808b:

Yes, it's `GiB` but it states that the unit is `gb`. Is that problematic?

server/src/test/java/org/elasticsearch/cluster/metadata/DesiredNodeTests.java
```diff
@@ -53,12 +53,55 @@ public void testExternalIdFallbacksToNodeName() {
...
+                ByteSizeValue.ofGb(1),
+                Version.CURRENT
+            )
+        );
+
+        expectThrows(
+            IllegalArgumentException.class,
+            () -> new DesiredNode(
+                settings,
+                new DesiredNode.ProcessorsRange(3.0f, 2.0f),
```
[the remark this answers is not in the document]
fcofdez: TIL about `ulp`. Thanks for the tip!

barkbay:
> This is strange, it should fail (I even added a couple of tests that cover that case). When you mention that it does not clear the previous max value, do you mean that the Desired nodes remain in the previous version?

I'm not able to reproduce, it might be a problem on my side.

> Yes, it's GiB but it states that the unit is gb. Is that problematic?

No, just want to be sure I was understanding the units correctly.

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNode.java
```diff
@@ -86,20 +96,61 @@ private static Version parseVersion(String version) {
...
+        if (processors == null && processorsRange == null) {
+            throw new IllegalArgumentException(
+                PROCESSORS_FIELD.getPreferredName()
+                    + " or "
+                    + PROCESSORS_RANGE_FIELD.getPreferredName()
+                    + " should be specified and none was specified"
+            );
+        }
+
+        if (processors != null & processorsRange != null) {
```
henningandersen, line 141:
```suggestion
        if (processors != null && processorsRange != null) {
```

henningandersen (approved):
LGTM, thanks Francisco.

fcofdez:
@elasticmachine upgrade branch

fcofdez:
Thanks all for the reviews!

PR #86437, opened 2022-05-04
Remove confusing alias to 'internalClusterTest' task

mark-vieira:
We initially added the "alias" to the `internalClusterTest` task mostly for legacy reasons as this task used to be called `icTest` and we didn't want to be overly disruptive. It's been awhile now and folks are used to the new task names, additionally, since it's not a true alias, but just a lifecycle task which depends on `internalClusterTest`, things like `icTest --tests ...` doesn't work the way you'd think which causes more confusion. For those who prefer a more terse task name, you can leverage Gradle's built-in support for truncating camel-case task names and run the `iCTest` task which will resolve to `internalClusterTest`.


DaveCTurner:
Change itself LGTM but this deserves a PSA email first since I imagine it will break some personal workflows.

mark-vieira:
@elasticmachine update branch

Issue #86429, opened 2022-05-04
Fix failing tests on `feature/desired-balance-allocator` branch


idegtiarenko:
Following tests are failing and need to be fixed before `DesiredBalanceShardsAllocator` could be merged to master.

- [x] DesiredBalanceServiceTests.* (https://github.com/elastic/elasticsearch/pull/86435)
- [x] DesiredBalanceReconcilerTests.* (https://github.com/elastic/elasticsearch/pull/86432)
- [x] ClusterAllocationExplainIT. testAllocationFilteringOnIndexCreation
- [x] ClusterHealthIT. testHealthOnIndexCreation
- [x] CorruptedFileIT. testCorruptionOnNetworkLayer (Replica shard is not started and remains in error state)
- [x] CorruptedFileIT. testReplicaCorruption
- [x] FilteringAllocationIT. testDecommissionNodeNoReplicas
- [x] IndexFoldersDeletionListenerIT. testListenersInvokedWhenIndexHasLeftOverShard (small probability to stuck after `logger.debug("--> creating a new index [{}]", indexName);`)
- [x] IndexRecoveryIT. testCancelNewShardRecoveryAndUsesExistingShardCopy
- [x] IndexRecoveryIT. testDoNotInfinitelyWaitForMapping (timed out waiting for green state: `ALLOCATION_FAILED, failed shard on node [y0Jt0-QrSu29efTbYU8AdQ]: failed to create index, failure org.elasticsearch.index.mapper.MapperParsingException: simulate mapping parsing error`)
- [x] IndexRecoveryIT. testCancelRecoveryWithAutoExpandReplicas (stuck after creating index [0-all] index in a cluster with a single master and no data nodes)
- [x] RareClusterStateIT. testDeleteCreateInOneBulk (consistently timing out on creating index with 0s timeout)
- [x] RecoveryFromGatewayIT. testSingleNodeNoFlush
- [x] ReplicaShardAllocatorIT. testDoNotCancelRecoveryForBrokenNode (timed out waiting for green state: `ALLOCATION_FAILED, failed recovery, failure org.elasticsearch.indices.recovery.RecoveryFailedException`)
- [x] ReplicaShardAllocatorIT. testPreferCopyCanPerformNoopRecovery
- [x] ReplicaShardAllocatorIT. testPreferCopyWithHighestMatchingOperations
- [x] ReplicaShardAllocatorIT. testPeerRecoveryForClosedIndices (<5% probability)
[...]

PR #86439, opened 2022-05-04
Delay response to update cluster settings pending reroute

DaveCTurner:
With this change we withhold the response to the update cluster settings
API until the corresponding reroute completes, fixing tests that do
things such as updating allocation filters and then waiting for all
relocations to complete, such as `AwarenessAllocationIT`.

Relates #86429


DaveCTurner:
Superseded by #86451, closing.

PR #86441, opened 2022-05-04
Require index.mode=time_series for creating time series data streams

martijnvg:
Prior to this change the `index.routing_path` was required in order for
data streams to be created in time series mode.

This commit changes that by also requiring `index.mode` index setting to
be set to `time_series`. The `index.routing_path` remains a required setting,
but it is no longer used to determine whether a data stream should be created
in time series mode.

The `index.mode` index setting will no longer automatically be added backing indices.
This was done by the `DataStreamIndexSettingsProvider` class and this class will not only
automatically add the `index.time_series.start_time` and `index.time_series.end_time`
index settings.

Relates #74660


modules/data-streams/src/internalClusterTest/java/org/elasticsearch/datastreams/TSDBIndexingIT.java
```diff
@@ -167,39 +165,6 @@ public void testTimeRanges() throws Exception {
         }
     }
 
-    public void testTsdbTemplatesWithComponentTemplate() throws Exception {
```
martijnvg, line 170: This test doesn't make a lot of sense now that `index.mode` setting is no longer automatically added to backing indices.

martijnvg: In exchange of removing this, I added some extra test at the end of this file that test validation that should kick in when creating template that creates time series data streams but index.mode or index.routing_path is missing.

nik9000 (approved):

martijnvg:
@elasticmachine update branch

PR #86447, opened 2022-05-04
Remove ImmutableOpenIntMap

rjernst:
All uses of the class have been removed, so this commit now removes the
tests and the class itself.

relates #86239


grcevski (approved):
LGTM!

PR #86448, opened 2022-05-04
Clean up and better document JvmOptionsParser

rjernst:
This change preparse the JvmOptionsParser for #85758 by moving the shell
specific behavior, like printing output, to the main method, and making
the main function take in all its parameters. Additionally the main
method to determine the jvm options is documented a bit.

relates #85758


distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/JvmOptionsParser.java
```diff
@@ -72,50 +75,73 @@ public static void main(final String[] args) throws InterruptedException, IOExce
...
+            Launchers.exit(e.exitCode);
+        }
+    }
+
+    /**
+     * Determines the jvm options that should be passed to the Elasticsearch Java process.
+     *
+     * <p> This method works by joining the options found in {@link SystemJvmOptions}, the {@code jvm.options} file,
+     * files in the {@code jvm.options.d} directory, and the options given by the {@code ES_JAVA_OPTS} environment
+     * variable. After the options are read, they are passed to a java subprocess which prints out the final options.
```
williamrandolph, line 97: The last sentence throws me off, since we pass to the subprocess in the `main()` method, not here.

williamrandolph (approved):
Disentangling the argument parsing and environment-variable wrangling from the string munging makes for much cleaner code! One little nit about the phrasing of the javadoc, but otherwise LGTM.

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/JvmOptionsParser.java
rjernst: The subprocess doesn't happen in main, it happens down inside `JvmErgonomics`, which internally calls `JvmOption.findFinalOptions`. I'll just remove this line though as I see how it is confusing.

PR #86449, opened 2022-05-05
Make verbosity in Terminal immutable

rjernst:
The Terminal class is almost immutable, except for the verbosity. In
production this is never modified after startup, but tests tweak the
verbosity, and startup does need to get at a terminal with a different
verbosity depending on cli args.

This commit adds a withVerbosity method to Terminal, which returns a new
Terminal, of the same underlying implementation, but with a modified
verbsity level.


rjernst:
I'm going to drop this for now because unfortunately it is quite tricky to get right due to how we manipulate terminal during startup with multi commands: when setting the verbosity on a subcommand, it leaks out to be the verbosity for any exceptions that occur, which are caught at the outer command.

PR #86453, opened 2022-05-05
[ML] Muting trained model upgrade tests

droberts195:
These tests also fail before the Java side of
https://github.com/elastic/ml-cpp/pull/2258 is merged.


davidkyle (approved):
LGTM

PR #86457, opened 2022-05-05
Cleanup SnapshotInProgressAllocationDecider a little

original-brownbear:
This should really be made faster via a better `SnapshotsInProgress` that has an index of what's going on per shard, but for now at least make it a little better.

We already have a stream here, lets not make this needlessly slow via the iterator
when we can have the stream iterate much faster when used directly.
Also, looks nicer this way.

relates #77466

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/SnapshotInProgressAllocationDecider.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/SnapshotInProgressAllocationDecider.java
@@ -48,7 +48,7 @@ public class SnapshotInProgressAllocationDecider extends AllocationDecider {
         return canAllocate(shardRouting, node, allocation);
     }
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/SnapshotInProgressAllocationDecider.java
```diff
@@ -51,45 +48,41 @@ public Decision canForceAllocateDuringReplace(ShardRouting shardRouting, Routing
         return canAllocate(shardRouting, node, allocation);
     }
 
+    private static final Decision YES_NO_SNAPSHOTS_RUNNING = Decision.single(Decision.Type.YES, NAME, "no snapshots are currently running");
```
idegtiarenko, line 51: Consider renaming this to `YES_NOT_RUNNING`. It was originally confused with `YES_NO` part

idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

PR #86458, opened 2022-05-05
Support geo label position through REST vector tiles API

### LINKED ISSUE #86044: Add label features to vector tile search API response

craigtaverner:
There is a need to provide sensibly calculated label positions for polygons and lines in Kibana maps (as requested in https://github.com/elastic/elasticsearch/issues/86044). A very convenient way to satisfy this need is through a runtime field that the rest API can make use of when labels are requested. This has the advantage of providing painless access to the label position as well.

Building onto https://github.com/elastic/elasticsearch/pull/86154, this PR adds support for the REST API to provide label positions to MVT queries, both for the HITS layer and the AGGS layer. To enable this feature, set `with_labels` to `true` as a query parameter to the vector tile search query.

Closes #86044

### DIFF
--- a/docs/reference/search/search-vector-tile-api.asciidoc
+++ b/docs/reference/search/search-vector-tile-api.asciidoc
@@ -371,7 +371,9 @@ polygon selected from the <<geoshape-indexing-approach,sorted triangle-tree>>.
 from the <<geoshape-indexing-approach,triangle-tree>>.
 * The aggregation results will provide one central point for each aggregation bucket.
 
-All label features added to the results will be distinguishable using the tag `_mvt_label_position`.
++
+All attributes from the original features will also be copied to the new label features.
+In addition, the new features will be distinguishable using the tag `_mvt_label_position`.
 // end::with_labels[]
 
 [role="child_attributes"]
--- a/docs/reference/search/search-vector-tile-api.asciidoc
+++ b/docs/reference/search/search-vector-tile-api.asciidoc
@@ -101,6 +101,11 @@ aggregation uses the `<zoom>/<x>/<y>` tile as a bounding box.
 on the `<field>`. The search only includes this aggregation if the
 `exact_bounds` parameter is `true`.
 
+* If the optional parameter `with_labels` is true, the internal search will include
+a dynamic runtime field that calls the `getLabelPosition` function of the geometry doc value.
+This enables the generation of new point features containing suggested geometry labels,
+so that, for example, multi-polygons will have only one label.
+
 For example, {es} may translate a vector tile search API request with a
 `grid_agg` argument of `geotile` and an `exact_bounds` argument of `true`
 into the following search:
@@ -353,6 +358,22 @@ If `true`, the exact number of hits is returned at the cost of some performance.
 If `false`, the response does not include the total number of hits matching the query.
 // end::track_total_hits[]
 
+// tag::with_labels[]
+`with_labels`::
+(Optional, Boolean)
+If true, the hits and aggs layers will contain additional point features representing
+suggested label positions for the original features.
+* `Point` and `MultiPoint` features will have one of the points selected.
+* `Polygon` and `MultiPolygon` features will have a single point generated,
+either the centroid, if it is within the polygon, or another point within the
+polygon selected from the <<geoshape-indexing-approach,sorted triangle-tree>>.
+* `LineString` features will likewise provide a roughly central point selected
+from the <<geoshape-indexing-approach,triangle-tree>>.
+* The aggregation results will provide one central point for each aggregation bucket.
+
+All label features added to the results will be distinguishable using the tag `_mvt_label_position`.
+// end::with_labels[]
+
 [role="child_attributes"]
 [[search-vector-tile-api-request-body]]
 ==== {api-request-body-title}
@@ -421,6 +442,8 @@ By default, the API calculates a bounding box for each feature. It sorts
 features based on this box's diagonal length, from longest to shortest.
 
 include::search-vector-tile-api.asciidoc[tag=track_total_hits]
+
+include::search-vector-tile-api.asciidoc[tag=with_labels]
 
 [role="child_attributes"]
 [[search-vector-tile-api-response]]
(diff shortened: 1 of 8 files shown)


iverase (changes requested):
I think we need to add a yaml test here: https://github.com/elastic/elasticsearch/blob/9a60aa8b64209740616266c2607b4600d90f3b6b/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/vector-tile/10_basic.yml

apart from that I tried locally and it is looking great!

iverase:
In addition, can we update the description of the PR because it is used to generate the release notes?

x-pack/plugin/vector-tile/src/main/java/org/elasticsearch/xpack/vectortile/rest/RestVectorTileAction.java
```diff
@@ -274,6 +297,21 @@ private static VectorTile.Tile.Layer.Builder buildHitsLayer(SearchHit[] hits, Ve
...
+                if (labelField != null) {
+                    Object labelPosValue = labelField.getValue();
+                    GeoPoint labelPos = GeoUtils.parseGeoPoint(labelPosValue, true);
+                    byte[] labelPosFeature = featureFactory.point(labelPos.lon(), labelPos.lat());
+                    if (labelPosFeature != null && labelPosFeature.length != 0) {
+                        featureBuilder.clear();
+                        featureBuilder.mergeFrom(labelPosFeature);
+                        VectorTileUtils.addPropertyToFeature(featureBuilder, layerProps, ID_TAG, searchHit.getId());
+                        VectorTileUtils.addPropertyToFeature(featureBuilder, layerProps, LABEL_POSITION_TAG, true);
+                        hitsLayerBuilder.addFeatures(featureBuilder);
```
iverase, line 311: I think we agreed that we should add all the tags that we added to the feature that is not a tag. That brings the question to what happens if the user has defined an attribute with the same name as the label?

x-pack/plugin/vector-tile/src/main/java/org/elasticsearch/xpack/vectortile/rest/RestVectorTileAction.java
```diff
@@ -309,6 +346,18 @@ private static VectorTile.Tile.Layer.Builder buildAggsLayer(
...
+                // Add label position as point
+                featureBuilder.clear();
+                GeoPoint labelPos = (GeoPoint) bucket.getKey();
+                byte[] labelPosFeature = featureFactory.point(labelPos.lon(), labelPos.lat());
+                if (labelPosFeature != null && labelPosFeature.length != 0) {
+                    featureBuilder.mergeFrom(labelPosFeature);
+                    VectorTileUtils.addPropertyToFeature(featureBuilder, layerProps, KEY_TAG, bucketKey);
+                    VectorTileUtils.addPropertyToFeature(featureBuilder, layerProps, LABEL_POSITION_TAG, true);
+                    aggLayerBuilder.addFeatures(featureBuilder);
+                }
```
iverase, line 359: I think we agreed that we should add all the tags that we added to the feature that is not a tag. That brings the question to what happens if the user has defined an attribute with the same name as the label?

iverase (changes requested):

x-pack/plugin/vector-tile/src/main/java/org/elasticsearch/xpack/vectortile/rest/RestVectorTileAction.java
craigtaverner: I think both concerns are considered in https://github.com/elastic/elasticsearch/pull/86458/commits/9fa8c0028b756e904beeeddc8a95c0cc4707fbe4

iverase (approved):
LGTM!

docs/reference/search/search-vector-tile-api.asciidoc
```diff
@@ -353,6 +358,22 @@ If `true`, the exact number of hits is returned at the cost of some performance.
...
+If true, the hits and aggs layers will contain additional point features representing
+suggested label positions for the original features.
+* `Point` and `MultiPoint` features will have one of the points selected.
+* `Polygon` and `MultiPolygon` features will have a single point generated,
+either the centroid, if it is within the polygon, or another point within the
+polygon selected from the <<geoshape-indexing-approach,sorted triangle-tree>>.
+* `LineString` features will likewise provide a roughly central point selected
+from the <<geoshape-indexing-approach,triangle-tree>>.
+* The aggregation results will provide one central point for each aggregation bucket.
+
```
nreese, line 373: Use `+` instead of empty line so statement below is indented under with_labels option

iverase: I will add that the label features inherit all the attributes from its corresponding original feature.

craigtaverner: I had tried that before, but it indented too much (made it part of the last bullet point). However, now I tried again, this time with an extra newline before the `+` and voila! It indented the right amount!

craigtaverner: I fixed both your suggestions in https://github.com/elastic/elasticsearch/pull/86458/commits/1c69a4aec99dea1891b087c1ed711c730729a0ee

nreese (approved):
doc changes LGTM

Issue #83784, opened 2022-02-10
Remove usages of LegacyClusterTaskResultActionListener


DaveCTurner:
### Description

In https://github.com/elastic/elasticsearch/pull/83562 we introduced a new API for better batching of master tasks, and also the `LegacyClusterTaskResultActionListener` to adapt old implementations to the new API as a temporary measure. This issue tracks the work to remove this adapter.

- [x] `o.e.cluster.LocalMasterServiceTask` and `o.e.cluster.ClusterStateTaskExecutor#unbatched()` (#83796)
- [x] `o.e.cluster.coordination.NodeRemovalClusterStateTaskExecutor` (#83797)
- [x] `o.e.cluster.coordination.JoinTaskExecutor` (#83803)
- [x] `o.e.cluster.metadata.MetadataIndexStateService` (#83807)
- [x] `o.e.cluster.metadata.MetadataMappingService` (#83811)
- [x] `o.e.cluster.service.MasterServiceTests` (#83840)
- [x] `o.e.cluster.action.shard.ShardStateAction` (#83842)
- [x] `o.e.action.admin.indices.create.AutoCreateAction` (#84170)
- [x] `o.e.action.admin.indices.rollover.TransportRolloverAction` (#84166)
- [x] `o.e.action.admin.cluster.desirednodes.DesiredNodesClusterStateTaskExecutor` (#85984)
- [x] `o.e.snapshots.SnapshotsService` (#86178)
- [x] `o.e.ingest.IngestService` (#86446)
- [x] `o.e.cluster.metadata.MetadataIndexTemplateService` (#86459)
- [x] `o.e.cluster.metadata.MetadataUpdateSettingsService` (#88643)
- [ ] `o.e.xpack.ilm.IndexLifecycleRunner` (#89459)
- [ ] Finally, push `o.e.cluster.ClusterStateTaskListener#clusterStateProcessed()` down to `o.e.cluster.ClusterStateUpdateTask` since it only makes sense for unbatched tasks (#89459)

PDTCCLF:
Hi, @DaveCTurner. It looks like no one is assigned for the last item "to push o.e.cluster.ClusterStateTaskListener#clusterStateProcessed() down to o.e.cluster.ClusterStateUpdateTask". As a first time contributor, I believe I could do it.

DaveCTurner:
[...]

PR #86446, opened 2022-05-04
Remove LegacyCTRAL from IngestService

DaveCTurner:
Relates #83784, #86017

PR #86459, opened 2022-05-05
Remove LegacyCTRAL from MetadataIndexTemplateService

DaveCTurner:
Relates #83784, #86017


DaveCTurner:
Not quite as elegant as #86446 - these tasks involve various components of a `Metadata` and in several places we need to rebuild a `Metadata` for validation. It's probably possible to push this down to a lower level but it'd be a much bigger change so I'm just sticking with iterating on the `ClusterState` here.

dakrone (approved):
LGTM

DaveCTurner:
Thanks Lee

PR #86460, opened 2022-05-05
Fix bounded hexagonal grids when they contain the bin on one of the poles

iverase:
Bounded hexagonal grids filter out all hexagons which their cartesian bounded box do not intersect the provided bounded box. Unfortunately the way we are computing the  cartesian bounded box for hexagons that contain one of the poles is incorrect. 

Because there is one hexagon containing the north pole and one contains the south pole, this PR computes the validity of those hexagons explicitly on the constructor of `GeoHexCellIdSource#BoundedCellValues` and it checks them explicitly in the method `validHex`.


x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/bucket/geogrid/GeoHexCellIdSource.java
```diff
@@ -76,10 +76,35 @@ private static class BoundedCellValues extends CellValues {
...
 
+        private final long northPoleHex, southPoleHex;
+        private final boolean northPoleHexValid, southPoleHexValid;
+
         protected BoundedCellValues(MultiGeoPointValues geoValues, int precision, GeoBoundingBox bbox) {
             super(geoValues, precision);
             this.crossesDateline = bbox.right() < bbox.left();
             this.bbox = bbox;
+            northPoleHex = H3.geoToH3(90, 0, precision);
+            northPoleHexValid = northPoleValidHex(northPoleHex);
```
craigtaverner, line 87: I think it might be a little more consistent to call this in the validHex function. I'm not sure there is value in caching these results in the constructor.

craigtaverner (approved):
It looks good, but I think it could be simplified slightly by removing the caching of northPoleHexValid and southPoleHexValid.

PR #86462, opened 2022-05-05
Lazy instantiate Cache in BitsetFilterCache

original-brownbear:
An empty cache instance costs a little over 50kb of heap. Lazy initializing the cache when needed
in this class comes with trivial overhead relative to the cost of using the cache
but can save hundreds of MB of heap on large data nodes.

relates #77466


idegtiarenko (approved):

ywelsch (approved):
This cache is only relevant in the context of nested fields. Not paying the extra price of instantiating the cache for every index that does not have nested fields is definitely great.

Another option would have been to make this cache a node-level instead of index-level service (similar to IndicesRequestCache). That would be a major refactoring, so I'm fine with this as well.

original-brownbear:
@elasticmachine update branch

original-brownbear:
Thanks Ievgen + Yannick!

PR #86465, opened 2022-05-05
Faster RoutingTable Builder

original-brownbear:
Copying the map one by one gets very expensive for large tables
=> leveraging the faster copy constructor that clones the map
internally gives a non-trivial speedup to many shards benchmark
bootstrapping.

relates #77466


idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

PR #86470, opened 2022-05-05
Cleanup+Speed-up DataStream Metadata handling

original-brownbear:
No need to copy maps so often, this was quite slow in many-shards
benchmarking which uses lots of datastreams.

relates #77466


server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
```diff
@@ -1515,14 +1510,7 @@ public Builder removeIndexTemplate(String name) {
...
-            DataStreamMetadata dataStreamMetadata = (DataStreamMetadata) customs.get(DataStreamMetadata.TYPE);
-            if (dataStreamMetadata != null) {
-                return dataStreamMetadata.dataStreams().get(dataStreamName);
-            } else {
-                return null;
-            }
+            return dataStreams().dataStreams().get(dataStreamName);
         }
 
         public Builder dataStreams(Map<String, DataStream> dataStreams, Map<String, DataStreamAlias> dataStreamAliases) {
```
joegallo, line 1516: In light of my previous comment, I'm curious whether this is better with the current signature versus perhaps being named `dataStreamMetadata` and receiving a `DataStreamMetadata`.

server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
```diff
@@ -1547,29 +1535,21 @@ public Builder put(DataStream dataStream) {
...
-            Map<String, DataStreamAlias> existingDataStreamAliases = Optional.ofNullable(
-                (DataStreamMetadata) this.customs.get(DataStreamMetadata.TYPE)
-            ).map(dsmd -> new HashMap<>(dsmd.getDataStreamAliases())).orElse(new HashMap<>());
-
-            this.customs.put(DataStreamMetadata.TYPE, new DataStreamMetadata(existingDataStreams, existingDataStreamAliases));
+            this.customs.put(DataStreamMetadata.TYPE, dataStreams().withAddedDatastream(dataStream));
             return this;
         }
 
+        private DataStreamMetadata dataStreams() {
```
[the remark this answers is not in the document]
original-brownbear: Sure makes sense :)

server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
original-brownbear: Yea lets do a follow-up for this IMO. Looking over this all these methods are needlessly complex and we can save tons of code here doing similar things like I did here already (as well as make the code a little more obviously correct in the process :)).

martijnvg (approved):
Thanks for making this better @original-brownbear @joegallo! LGTM 👍

PR #86471, opened 2022-05-05
Port javadoc configuration logic into a binary plugin

breskeby:
Also adding test coverage and fixing certain issues we stumbled into

1. when resolving project dependencies of compileClasspath we need to use `allDependencies` instead of `dependencies` as usually no dependencies are added directly to `compileClasspath` but via `implementation`, `compileonly` and `api` and friends
2. Fixed javadoc setup for projects using shadowed dependencies via shadow plugin
3. Fixed dealing with skipped javadoc tasks in referenced dependent projects.

In general this PR only fixes expected javadoc generation as it was intended before this PR. 

This also contains some tweaks to our gradle integration test fixtures for setting up and debugging test projects


build-tools/src/testFixtures/groovy/org/elasticsearch/gradle/fixtures/AbstractGradleFuncTest.groovy
```diff
@@ -51,14 +62,14 @@ abstract class AbstractGradleFuncTest extends Specification {
...
         return new NormalizeOutputGradleRunner(
-            new InternalAwareGradleRunner(
-                GradleRunner.create()
-                    .withDebug(ManagementFactory.getRuntimeMXBean().getInputArguments().toString().indexOf("-agentlib:jdwp") > 0)
-                    .withProjectDir(projectDir)
-                    .withPluginClasspath()
-                    .forwardOutput()
-            ),
-            projectDir
+                new InternalAwareGradleRunner(
```
pugnascotia, line 65: Did you mean to make all these whitespace changes?

build-tools/src/testFixtures/groovy/org/elasticsearch/gradle/fixtures/AbstractGradleFuncTest.groovy
```diff
@@ -151,4 +162,29 @@ abstract class AbstractGradleFuncTest extends Specification {
             System.err.println("Syserr: " + proc.errorStream.text)
         }
     }
+
+    def cleanup() {
+        if (Boolean.getBoolean('test.keep.samplebuild')) {
```
pugnascotia, line 167: TIL this existed.

pugnascotia (approved):

build-tools/src/testFixtures/groovy/org/elasticsearch/gradle/fixtures/AbstractGradleFuncTest.groovy
breskeby: nope. fixed

build-tools/src/testFixtures/groovy/org/elasticsearch/gradle/fixtures/AbstractGradleFuncTest.groovy
pugnascotia: Sorry, I wasn't clear - I didn't know that `Boolean.getBoolean(String)` existed 😅

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavadocPlugin.java
```diff
@@ -0,0 +1,134 @@
...
+            .sorted(Comparator.comparing(Dependency::getGroup))
+            .filter(d -> d instanceof ProjectDependency)
+            .map(d -> (ProjectDependency) d)
+            .filter(p -> p.getDependencyProject() != null)
+            .forEach(projectDependency -> configureDependency(shadow, projectDependency));
+    }
+
+    private void configureDependency(boolean shadowed, ProjectDependency dep) {
+        var upstreamProject = dep.getDependencyProject();
+        if (shadowed) {
```
mark-vieira, line 82: What projects are we actually publishing javadocs for that _also_ use the shadow plugin? I thought it used to only be the HLRC but that is no-more. Otherwise I think we also shadow the sql cli, but that's a CLI tool so we shouldn't be publishing Javadocs. Anything else?

mark-vieira (approved):

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavadocPlugin.java
breskeby: there's:

- `client/rest-high-level`
- `sql/jdbc`
- `sql/sql-cli`
- `repository-hdfs/hadoop-client-api`

from reading https://github.com/elastic/infra/blob/master/cd/release/release-manager/project-configs/master/elasticsearch.gradle I think from those only `sql/jdbc` is published to maven which includes javadoc I guess. Am I correct here?

PR #86473, opened 2022-05-05
Disabling users includes realm in same-user validation 

n1v0lg:
A validation check prevents users from disabling themselves. The check
is incorrect since it does not factor in realms. Users can have
overlapping usernames across realms, and should be able to disable
same-named users in other realms, authorization provided. This PR
tweaks the validation check to account for the source realm.


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportSetEnabledAction.java
```diff
@@ -62,4 +64,16 @@ protected void doExecute(Task task, SetEnabledRequest request, ActionListener<Ac
             listener.delegateFailure((l, v) -> l.onResponse(ActionResponse.Empty.INSTANCE))
         );
     }
+
+    private boolean isSameUserRequest(SetEnabledRequest request) {
+        final var effectiveSubject = securityContext.getAuthentication().getEffectiveSubject();
+        final var realmType = effectiveSubject.getRealm().getType();
+        if (ReservedRealm.TYPE.equals(realmType) == false && NativeRealmSettings.TYPE.equals(realmType) == false) {
```
n1v0lg, line 71: Usernames are mutually exclusive across these two realms so we don't have to worry about native user trying to disable same-named reserved user.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportSetEnabledAction.java
```diff
@@ -62,4 +64,16 @@ protected void doExecute(Task task, SetEnabledRequest request, ActionListener<Ac
             listener.delegateFailure((l, v) -> l.onResponse(ActionResponse.Empty.INSTANCE))
         );
     }
+
+    private boolean isSameUserRequest(SetEnabledRequest request) {
+        final var effectiveSubject = securityContext.getAuthentication().getEffectiveSubject();
```
n1v0lg, line 69: I'm matching the existing logic (aside from the realm improvement), however there is another possible case to prevent: currently, nothing keeps alice from running as bob to disable alice. I can extend the logic here to account for that but not sure if worth it.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/user/TransportSetEnabledActionTests.java
```diff
@@ -252,14 +259,87 @@ public void onFailure(Exception e) {
         );
     }
 
+    public void testUserCanModifySameNameUserFromDifferentRealm() throws Exception {
```
ywangd, line 262: This method is very similar to the existing `testValidUser(User)`. Can we somehow consolidate them?

ywangd (approved):
LGTM

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportSetEnabledAction.java
n1v0lg: Makes sense, thanks for the context!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportSetEnabledAction.java
```diff
@@ -62,4 +64,13 @@ protected void doExecute(Task task, SetEnabledRequest request, ActionListener<Ac
...
+        final var realmType = effectiveSubject.getRealm().getType();
+        // Only native or reserved realm users can be disabled via the API. If the realm of the effective subject is neither,
+        // the target must be a different user
+        return (ReservedRealm.TYPE.equals(realmType) || NativeRealmSettings.TYPE.equals(realmType))
```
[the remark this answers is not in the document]
n1v0lg: Hmm are you sure? My understanding is that we want reserved realm, not file.

I can add a util `isNativeOrReserved` though!

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/user/TransportSetEnabledActionTests.java
n1v0lg: Yup, should be able to pass in `authentication` - rest is generic

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportSetEnabledAction.java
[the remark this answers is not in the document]
ywangd: Sorry you are right. No need to do anything then.

PR #86482, opened 2022-05-05
Fix min node version before state recovery

DaveCTurner:
We use the minimum node version in the cluster state to make decisions
about backwards compatibility (e.g. to choose newer actions in the REST
layer only if all nodes will support it). Once the cluster is fully
formed we reject attempts by older nodes to join the cluster so that the
minimum node version only ever increases, which makes
backwards-compatibility decisions safe.

However, it's possible that the REST layer will make decisions about
backwards compatibility before the cluster is fully formed. In this
state, older nodes may still join the cluster and may therefore see
actions that they do not understand.

With this commit we report no nodes to the REST layer until the cluster
is fully-formed, and change the minimum node version in an empty cluster
to be the minimum compatible version. This means the REST layer will
operate in a maximally-compatible mode until the cluster is formed.

Relates #86405

### DIFF
--- a/docs/changelog/86482.yaml
+++ b/docs/changelog/86482.yaml
@@ -1,0 +1,5 @@
+pr: 86482
+summary: Fix min node version before state recovery
+area: Infra/REST API
+type: bug
+issues: []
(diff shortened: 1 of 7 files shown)


rjernst (commented):
I like the idea here! However, how would the nodes passed to the initRestHandlers ever get updated once the cluster state has stabilized on the new ES version? This happens only once when starting up the node.

rjernst:
Oh I see, we pass a supplier around everywhere, so once the cluster state gets updated locally, the next call to any rest handler using it will see the new nodes.

rjernst (approved):
LGTM. Might be nice to have a test checking a rest handler sees the newly reflected nodes, but it's fine as is too.

DaveCTurner:
++ yes it's a `Supplier<DiscoveryNodes>`, we get it afresh each time. Bit tricky to test but I think the one added in 8c4e1ed should do the job.

grcevski (approved):
LGTM! Thank you!

DaveCTurner:
Thanks both :)

PR #86484, opened 2022-05-05
Fix log message in JwtRealmGenerateTests for printing test ES-Client-Authentication header values

justincr-elastic:
This PR adds the missing value in a log message. It is not a functional change.

The tests in JwtRealmGenerateTests generate an example JWT issuer and use it to issue a JWT. The teams also create a JWT realm, and use the JWT to run an authentication test against the realm. At the end of the test, the issuer settings, realm settings, JWT contents, and request headers are printed.

The value for the request header ES-Client-Authentication should look like this.
```
ES-Client-Authentication: SharedSecret client-shared-secret-string
```

The test was printing this, with a missing value. Manually copying the generated examples to Elasticsearch and running it fails.
```
ES-Client-Authentication: client-shared-secret-string
```


slobodanadamovic (approved):
LGTM

PR #86493, opened 2022-05-05
Make the ILM and SLM history_index_enabled settings dynamic

joegallo:

Makes `indices.lifecycle.history_index_enabled` a dynamic setting, so that it's more useful as an escape hatch in situations where the ILM history writing ends up being a problem (as a dynamic setting, we'd be able to change the value without restarting the cluster).

For symmetry reasons, the same change is also made to `slm.history_index_enabled` although the practical need for a runtime escape hatch there is essentially irrelevant.

### DIFF
--- a/x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/history/ILMHistoryStore.java
+++ b/x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/history/ILMHistoryStore.java
@@ -58,7 +58,7 @@ public class ILMHistoryStore implements Closeable {
         ).getBytes()
     );
 
-    private boolean ilmHistoryEnabled = true;
+    private volatile boolean ilmHistoryEnabled = true;
     private final BulkProcessor processor;
     private final ThreadPool threadPool;
 
--- a/x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/history/ILMHistoryStore.java
+++ b/x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/history/ILMHistoryStore.java
@@ -59,12 +59,14 @@ public class ILMHistoryStore implements Closeable {
         ).getBytes()
     );
 
-    private final boolean ilmHistoryEnabled;
+    private boolean ilmHistoryEnabled;
     private final BulkProcessor processor;
     private final ThreadPool threadPool;
 
     public ILMHistoryStore(Settings nodeSettings, Client client, ClusterService clusterService, ThreadPool threadPool) {
-        this.ilmHistoryEnabled = LIFECYCLE_HISTORY_INDEX_ENABLED_SETTING.get(nodeSettings);
+        this.setIlmHistoryEnabled(LIFECYCLE_HISTORY_INDEX_ENABLED_SETTING.get(nodeSettings));
+        clusterService.getClusterSettings().addSettingsUpdateConsumer(LIFECYCLE_HISTORY_INDEX_ENABLED_SETTING, this::setIlmHistoryEnabled);
+
         this.threadPool = threadPool;
 
         this.processor = BulkProcessor.builder(new OriginSettingClient(client, INDEX_LIFECYCLE_ORIGIN)::bulk, new BulkProcessor.Listener() {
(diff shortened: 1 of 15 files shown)


x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/history/ILMHistoryStore.java
```diff
@@ -59,12 +58,17 @@ public class ILMHistoryStore implements Closeable {
...
 
-    private final boolean ilmHistoryEnabled;
+    private boolean ilmHistoryEnabled = true;
     private final BulkProcessor processor;
     private final ThreadPool threadPool;
 
-    public ILMHistoryStore(Settings nodeSettings, Client client, ClusterService clusterService, ThreadPool threadPool) {
-        this.ilmHistoryEnabled = LIFECYCLE_HISTORY_INDEX_ENABLED_SETTING.get(nodeSettings);
+    public ILMHistoryStore(Client client, ClusterService clusterService, ThreadPool threadPool) {
+        if (clusterService != null) {
```
dakrone, line 66: I don't think I like having a nullable `ClusterService` for the constructor, is there a way we can remove this check?

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/slm/history/SnapshotHistoryStore.java
```diff
@@ -38,12 +37,15 @@ public class SnapshotHistoryStore {
...
     private final ClusterService clusterService;
-    private final boolean slmHistoryEnabled;
+    private boolean slmHistoryEnabled = true;
 
-    public SnapshotHistoryStore(Settings nodeSettings, Client client, ClusterService clusterService) {
+    public SnapshotHistoryStore(Client client, ClusterService clusterService) {
         this.client = client;
         this.clusterService = clusterService;
-        slmHistoryEnabled = SLM_HISTORY_INDEX_ENABLED_SETTING.get(nodeSettings);
+        if (clusterService != null) {
```
dakrone, line 45: Same here about cluster service being nullable

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/history/ILMHistoryStore.java
```diff
@@ -59,12 +58,17 @@ public class ILMHistoryStore implements Closeable {
         ).getBytes()
     );
 
-    private final boolean ilmHistoryEnabled;
+    private boolean ilmHistoryEnabled = true;
```
dakrone, line 61: Most of the time we make these `volatile` so they are visible immediately across threads

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/slm/history/SnapshotHistoryStore.java
```diff
@@ -38,12 +37,15 @@ public class SnapshotHistoryStore {
 
     private final Client client;
     private final ClusterService clusterService;
-    private final boolean slmHistoryEnabled;
+    private boolean slmHistoryEnabled = true;
```
dakrone, line 40: This one should be `volatile` also.

dakrone (changes requested):
I left some comments, I think we shouldn't have a nullable `ClusterService` for these

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/history/ILMHistoryStore.java
joegallo: Ah, yep, good catch. 71778259957

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/slm/history/SnapshotHistoryStore.java
joegallo: Also 71778259957

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/history/ILMHistoryStore.java
joegallo: How about 33d9a309871b2e352882efff524dc82becb06d69?

dakrone (approved):
LGTM

PR #86498, opened 2022-05-06
TokenService decode JWTs, change warn to debug

justincr-elastic:
The goal of this PR is to change a warning in TokenService to debug. That warning is a false positive for multiple situations.

Three out of 12 authenticators look for  `Authorization` header with `bearer` scheme.
```
Authorization: bearer <servicetoken>
```
```
Authorization: bearer <token>
```
```
Authorization: bearer <jwt>
```

Also, OIDC and SAML realms issue a token after authentication success, so those also arrive in subsequent requests with `bearer` scheme too.

TokenService runs early and prints a debug if it cannot parse the Authorization header. The warning log does not make sense, so changing it to only log at debug, which is in line with how other authenticators work.


justincr-elastic:
@elasticmachine update branch

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/TokenService.java
```diff
@@ -616,11 +616,7 @@ void decodeToken(String token, ActionListener<UserToken> listener) {
             }
         } catch (Exception e) {
             // could happen with a token that is not ours
-            if (logger.isDebugEnabled()) {
-                logger.debug("built in token service unable to decode token", e);
-            } else {
-                logger.warn("built in token service unable to decode token");
```
n1v0lg, line 622: Surprised that debug cancelled out the higher-severity warning here. Was the intention originally to avoid the exception making it into the warning? (just out of curiosity!)

n1v0lg (approved):
LGTM

justincr-elastic:
@elasticmachine update branch

justincr-elastic:

justincr-elastic:
@elasticmachine update branch

Issue #86367, opened 2022-05-02
[Transform] Support range aggregations


simianhacker:
### Description

I'm currently working on transforming APM data into summary documents for implementing SLO's. The current latency for an APM transaction is stored using a `histogram` field. We would like to be able to create ratio for `good transactions / total transactions`. In the example below, a good transaction is anything under `200ms` which can be extracted from the histogram field using a `range` aggregation like `{ "range": { "field": "transaction.duration.histogram", "ranges": [{ "to": 200000 }] }` along with the total transactions like `{ "range": { "field": "transaction.duration.histogram", "ranges": [{ "from": 0 }] }`. These values are then stored as a `numerator` and a `denominator` using a `bucket_script`. Here is an example of the transform definition we would like to create.

```JSON
PUT _transform/slo_latency
{
  "source": {
    "index": "remote_cluster:metrics-apm*",
    "runtime_mappings": {
      "sloId": {
        "type": "keyword",
        "script": {
          "source": "emit('uuid-of-slo-definition')"
        }
      }
    }
  },
  "frequency": "1m",
  "dest": {
    "index": "slo_latency"
  },
  "sync": {
    "time": {
      "field": "@timestamp",
      "delay": "60s"
    }
  },
  "retention_policy": {
    "time": {
      "field": "@timestamp",
      "max_age": "30d"
    }
  },
  "pivot": {
    "group_by": {
      "transaction.name": {
        "terms": {
          "field": "transaction.name"
        }
      },
      "service.name": {
        "terms": {
          "field": "service.name"
        }
      },
      "sloId": {
        "terms": {
          "field": "sloId"
        }
      },
      "@timestamp": {
        "date_histogram": {
          "field": "@timestamp",
          "calendar_interval": "1m"
        }
      }
    },
    "aggregations": {
      "_numerator": {
        "range": {
          "field": "transaction.duration.histogram",
          "ranges": [
            {
              "to": 200000
            }
          ]
        }
      },
      "_demoninator": {
        "range": {
          "field": "transaction.duration.histogram",
          "ranges": [
            {
              "from": 0
            }
          ]
        }
      },
      "numerator": {
        "bucket_script": {
          "buckets_path": {
            "numerator": "_numerator['*-200000.0']>_count"
          },
          "script": "params.numerator"
        }
      },
      "denominator": {
        "bucket_script": {
          "buckets_path": {
            "denominator": "_demoninator['0.0-*']>_count"
          },
          "script": "params.denominator"
[...]

PR #86501, opened 2022-05-06
[Transform] Support `range` aggregation in transform

przemekwitek:
This PR adds support for `range` aggregation in transform.

Relates https://github.com/elastic/elasticsearch/issues/86367

### DIFF
--- a/x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/pivot/AggregationResultUtils.java
+++ b/x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/pivot/AggregationResultUtils.java
@@ -47,6 +47,7 @@ import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.util.stream.Collectors.toMap;
 import static org.elasticsearch.xpack.transform.transforms.pivot.SchemaUtil.dropFloatingPointComponentIfTypeRequiresIt;
 import static org.elasticsearch.xpack.transform.transforms.pivot.SchemaUtil.isNumericType;
 
@@ -344,12 +345,17 @@ public final class AggregationResultUtils {
         @Override
         public Object value(Aggregation agg, Map<String, String> fieldTypeMap, String lookupFieldPrefix) {
             Range aggregation = (Range) agg;
-            HashMap<String, Long> ranges = new HashMap<>();
-            for (Range.Bucket bucket : aggregation.getBuckets()) {
-                double from = bucket.getFrom() == null ? Double.NEGATIVE_INFINITY : ((Double) bucket.getFrom()).doubleValue();
-                double to = bucket.getTo() == null ? Double.POSITIVE_INFINITY : ((Double) bucket.getTo()).doubleValue();
-                ranges.put(generateKeyForRange(from, to), bucket.getDocCount());
-            }
+            Map<String, Long> ranges = aggregation.getBuckets()
+                .stream()
+                .collect(
+                    toMap(
+                        bucket -> bucket.getKeyAsString()
+                            .replace(".0-", "-")  // from: convert double to integer
+                            .replaceAll("\\.0$", "")  // to: convert double to integer
+                            .replace('.', '_'),  // convert remaining dots with underscores so that the key prefix is not treated as object
+                        bucket -> bucket.getDocCount()
+                    )
+                );
             return ranges;
         }
     }
--- a/x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/pivot/AggregationResultUtils.java
+++ b/x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/pivot/AggregationResultUtils.java
@@ -44,10 +44,11 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import static java.util.stream.Collectors.toMap;
 import static org.elasticsearch.xpack.transform.transforms.pivot.SchemaUtil.dropFloatingPointComponentIfTypeRequiresIt;
 import static org.elasticsearch.xpack.transform.transforms.pivot.SchemaUtil.isNumericType;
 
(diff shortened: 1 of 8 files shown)


x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/pivot/AggregationResultUtils.java
```diff
@@ -334,6 +340,27 @@ public Object value(Aggregation agg, Map<String, String> fieldTypeMap, String lo
...
+            return ranges;
+        }
+    }
+
+    static String generateKeyForRange(double from, double to) {
+        return new StringBuilder().append(Double.isInfinite(from) ? "*" : OutputFieldNameConverter.fromDouble(from))
+            .append("-")
+            .append(Double.isInfinite(to) ? "*" : OutputFieldNameConverter.fromDouble(to))
+            .toString();
+    }
```
benwtrent, line 362: Something that just came to mind is: How do we support multi-bucket  aggs when they are passed `keyed: false`?

It would be cool if our handling of range was easily extensible to `ip_range`, `date_range`, `geo_distance`, but I suppose this is an OK first pass.

benwtrent (commented):
Overall looks good. I have one small question that may not be addressed in this PR. 

My main concern is: Does having `*` in a field name cause issues with subsequent aggs or queries against those fields? I would assume not because we are allowed to create fields like that. But, just being a little paranoid :D

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/pivot/AggregationResultUtils.java
```diff
@@ -344,12 +345,17 @@ static class RangeAggExtractor implements AggValueExtractor {
...
-                ranges.put(generateKeyForRange(from, to), bucket.getDocCount());
-            }
+            Map<String, Long> ranges = aggregation.getBuckets()
+                .stream()
+                .collect(
+                    toMap(
+                        bucket -> bucket.getKeyAsString()
+                            .replace(".0-", "-")  // from: convert double to integer
+                            .replaceAll("\\.0$", "")  // to: convert double to integer
+                            .replace('.', '_'),  // convert remaining dots with underscores so that the key prefix is not treated as object
```
benwtrent, line 355: ++

The option I thought about was

```
str.endsWith(".0") ? str.subString(str.length() - 2, str.length()) : str
```

But regex works fine here too :)

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/pivot/AggregationResultUtils.java
```diff
@@ -334,6 +341,32 @@ public Object value(Aggregation agg, Map<String, String> fieldTypeMap, String lo
         }
     }
 
+    static class RangeAggExtractor implements AggValueExtractor {
+        @Override
+        public Object value(Aggregation agg, Map<String, String> fieldTypeMap, String lookupFieldPrefix) {
```
benwtrent, line 346: I am not sure this is extracting the sub-aggs. 

For multi-bucket, we should extract sub aggs as well.

Here is a `terms` with sub-aggs:

```
POST _transform/_preview
{
  "source": {
    "index": "kibana_sample_data_flights"
  },
  "pivot": {
    "group_by": {
      "dest": {"terms": {
        "field": "DestAirportID"
      }}
    }, 
    "aggregations": {
      "t": {
        "terms": {
          "field": "OriginCountry",
          "size": 1
        },
        "aggs": {
          "avg": {
            "avg": {
              "field": "DistanceMiles"
            }
          }
        }
      }
    }
  }
}
```
what a bucket looks like:
```
{
      "t" : {
        "AE" : {
          "avg" : 8160.76904296875
        }
      },
      "dest" : "ABQ"
    }
```

But when I try with range:

```
{
  "source": {
    "index": "kibana_sample_data_flights"
  },
  "pivot": {
    "group_by": {
      "dest": {
        "terms": {
          "field": "DestAirportID"
        }
      }
    },
    "aggregations": {
      "r": {
        "range": {
          "field": "AvgTicketPrice",
          "keyed": true,
          "ranges": [
            {
              "to": 500
            },
            {
              "from": 500,
              "to": 700
            },
            {
              "from": 700
            }
          ]
        },
        "aggs": {
          "avg": {
            "avg": {
              "field": "DistanceMiles"
            }
          }
        }
      }
    }
  }
}
```
I still get just counts:
```
{
      "r" : {
        "500-700" : 0,
        "*-500" : 0,
        "700-*" : 2
      },
      "dest" : "ABQ"
    }
```

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/pivot/AggregationResultUtils.java
przemekwitek: > My main concern is: Does having * in a field name cause issues with subsequent aggs or queries against those fields? I would assume not because we are allowed to create fields like that. But, just being a little paranoid :D

Yes, `*` in the field names seems strange but nothing failed in the integration tests and manual tests I made so far.

> How do we support multi-bucket aggs when they are passed keyed: false?

IIUC regardless of the value of `keyed` we index the documents into destination index in the same format i.e. a key-value map rather than a list.
I've added a few assertions to `TransformPivotRestIT.testPivotWithRanges` that verify it is actually the case for `range`.

> It would be cool if our handling of range was easily extensible to ip_range, date_range, geo_distance, but I suppose this is an OK first pass.

+1
ip_range, date_range, geo_distance are not supported in this PR but hopefully the logic will be reusable.

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/pivot/AggregationResultUtils.java
przemekwitek: Good spot!
I've just fixed it by reusing code for `MultiBucketsAggExtractor`.

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/pivot/TransformAggregations.java
```diff
@@ -207,14 +208,28 @@ public static Tuple<Map<String, String>, Map<String, String>> getAggregationInpu
...
+                    .stream()
                     .collect(
-                        Collectors.toMap(p -> percentilesAgg.getName() + "." + p, p -> { return percentilesAgg.getType(); }, (p1, p2) -> p1)
+                        Collectors.toMap(
+                            r -> rangeAgg.getName() + "." + AggregationResultUtils.generateKeyForRange(r.getFrom(), r.getTo()),
+                            r -> "range",
+                            (p1, p2) -> p1
+                        )
                     )
             );
```
benwtrent, line 234: This does not really map the values correctly. Especially if the agg has sub-aggs.

Running locally I continually get `Failed to deduce mapping for...` when there are sub-agg present or not. 

I think how this should be is:

 - `range` by itself should totally map to `long` (not source, as its just providing a count)
 - `range.sub_agg` if it exists should map to the sub-agg type (see ` if (agg.getSubAggregations().size() > 0) ` predicate).

And then we should have some tests that verify how it works with sub aggs or not.

hendrikmuhs:
Instead of creating string keys, I suggest to try out the native [range type](https://www.elastic.co/guide/en/elasticsearch/reference/current/range.html), this has the benefit of being better retrievable and further aggregateable.

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/pivot/TransformAggregations.java
przemekwitek: > I think how this should be is:
> 
> * `range` by itself should totally map to `long` (not source, as its just providing a count)
> * `range.sub_agg` if it exists should map to the sub-agg type (see `if (agg.getSubAggregations().size() > 0)` predicate).

Done.

> And then we should have some tests that verify how it works with sub aggs or not.

Done both in the unit test and integration test. Please LMK if you see other scenario that would benefit from more testing.

benwtrent (approved):

przemekwitek:
run elasticsearch-ci/bwc

przemekwitek:
run elasticsearch-ci/bwc

PR #86503, opened 2022-05-06
Make AllocationDecider canRemain loop tighter

original-brownbear:
No need to collect a multi decision here (which is surprisingly prominent in the profiling, accounting for almost 10% of the cost of reroute on average)
when we're not using it in the end anyway.

relates #77466


DaveCTurner (approved):
LGTM

PR #65362, opened 2020-11-23
Allow passing versioned media types to 7.x server

pgomulka:
a follow up after #63071 where it missed the XContentType.fromMediaType
method.
That method also have to remove the vendor specific substrings
(vnd.elasticsearch+ and compatible-with parameter) from mediaType value

relates #51816

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
- Have you followed the [contributor guidelines](https://github.com/elastic/elasticsearch/blob/master/CONTRIBUTING.md)?
- If submitting code, have you built your formula locally prior to submission with `gradle check`?
- If submitting code, is your pull request against master? Unless there is a good reason otherwise, we prefer pull requests against master and will backport as needed.
- If submitting code, have you checked that your submission is for an [OS and architecture that we support](https://www.elastic.co/support/matrix#show_os)?
- If you are submitting this code for a class then read our [policy](https://github.com/elastic/elasticsearch/blob/master/CONTRIBUTING.md#contributing-as-part-of-a-class) for that.

PR #63071, opened 2020-09-30
Allow passing versioned media types to 7.x server

pgomulka:
7.x client can pass media type with a version which will return a 7.x
version of the api in ES 8.
In ES server 7 this media type shoulld be accepted but it serve the same
version of the API (7x)
relates #51816

PR #86517, opened 2022-05-06
Enable HLRC compatibility mode by default

swallez:
HLRC compatibility mode allows it to communicate with Elasticsearch 8.x. Users are generally unaware of this feature, and this leads to a lot of questions and issues about the migration from 7.x to 8.x and the transition from HLRC to the new Java API client (we recently [added docs there](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/current/migrate-hlrc.html) as well).

This PR enables compatibility mode by default so that HLRC will work out of the box with versions 7.x and 8.x without any additional configuration.

It also adds docs about it, with a prominent note on the index page linking to compatibility mode explanations.

Reviewers:
* @dakrone who initially added compatibility mode
* @szabosteve for the doc part

### DIFF
--- a/client/rest-high-level/src/main/java/org/elasticsearch/client/RestHighLevelClient.java
+++ b/client/rest-high-level/src/main/java/org/elasticsearch/client/RestHighLevelClient.java
@@ -347,7 +347,12 @@ public class RestHighLevelClient implements Closeable {
                 .flatMap(Function.identity())
                 .collect(toList())
         );
-        if (useAPICompatibility == null && "true".equals(System.getenv(API_VERSIONING_ENV_VARIABLE))) {
+
+        // Compatibility mode is on by default, then env variable has precedence over builder setting
+        String apiVersioningEnv = System.getenv(API_VERSIONING_ENV_VARIABLE);
+        if (useAPICompatibility == null && apiVersioningEnv == null) {
+            this.useAPICompatibility = true;
+        } else if (useAPICompatibility == null && "true".equals(System.getenv(API_VERSIONING_ENV_VARIABLE))) {
             this.useAPICompatibility = true;
         } else {
             this.useAPICompatibility = Boolean.TRUE.equals(useAPICompatibility);
--- a/client/rest-high-level/src/test/java/org/elasticsearch/client/CustomRestHighLevelClientTests.java
+++ b/client/rest-high-level/src/test/java/org/elasticsearch/client/CustomRestHighLevelClientTests.java
@@ -147,7 +147,8 @@ public class CustomRestHighLevelClientTests extends ESTestCase {
      * Mocks the synchronous request execution like if it was executed by Elasticsearch.
      */
     private Response mockPerformRequest(Request request) throws IOException {
-        assertThat(request.getOptions().getHeaders(), hasSize(1));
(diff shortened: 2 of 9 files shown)


sethmlarson:
Enabling compatibility headers by default will cause errors when HLRC connects to older server versions (pre-7.14 iirc?)

dakrone:
This feels like a breaking change to me? This means someone going from a 7.17.3 to 7.17.4 version of the client might no longer be able to communicate with older versions of Elasticsearch. I'm not sure that I am comfortable making this change in a patch release.

szabosteve (approved):
Thanks for these changes, docs LGTM!

jakelandis:
> This means someone going from a 7.17.3 to 7.17.4 version of the client might no longer be able to communicate with older versions of Elasticsearch

This is a pretty big concern, and the reason we did not default in any of the 7.x clients. Is there specific reason why should change course, but only for the Java HRLC ? I will have to look up the exact version, but early 7.x clusters will fail all requests when presented with the compatibility header(s).

swallez:
@dakrone @jakelandis I understand your points, but don't consider them as a concern. This change is consistent with the [HLRC compatibility guidelines](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/java-rest-high-compatibility.html): "The High Level Client is guaranteed to be able to communicate with any Elasticsearch node running on the same major version and greater or equal minor version". Digging in the history, support for `vnd.elasticsearch` media types was added in 7.11 (see #63071 and #65362)

The goal is to remove a number of frustration points when migrating from 7 to 8. We've seen a number of reports where users try HLRC 7.x against ES 8.x, see failures as they did not enable compatibility mode, and end up considering that migration to the Java API client is a requirement. They not only feel we forced that migration on them, but feel blocked when encountering issues or missing features in the Java API client.

We've put a lot of engineering effort in making 8.x compatible with 7.x, but this feature is [hidden so well](https://www.elastic.co/guide/en/elasticsearch/reference/current/rest-api-compatibility.html) that users don't know about it. Better docs is one side of the equation (see also the [Java API client docs](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/current/migrate-hlrc.html)), provided that users actually read them. Having something that works out of the box provides the best experience, as it means users do not even have to know about it!

Regarding the change in a patch version, we have to consider that there will be no 7.18 so it's the only way for us to remove this roadblock.

The number of users running HLRC 7.17 against ES 7.10 or lower who may be impacted by this change is likely to be low and will fade over time, while migration from 7 to 8 is increasing and is the way forward. I'd rather explain users running HLRC 7.17.4 against ES 7.10 to disable compatibility mode or upgrade their cluster rather than deal with the frustration of users migrating from 7 to 8.

dakrone:
@swallez okay, since we document it that way, I think that it is reasonable to do this then. As a compromise, what do you think about marking this PR as `>breaking` so that it becomes more prevalent in the release documentation?

swallez:
Thanks @dakrone. It indeed totally makes sense to document this as a breaking change. I've updated the release notes and fixed the minimal compatible version (7.11) in the docs in f9b546b

swallez:
@dakrone I've updated the release notes and fixed a BWC test. Can you do a quick review?

dakrone (approved):
LGTM

PR #86519, opened 2022-05-06
Add getResolution method to H3

iverase:
There can be cases where it is useful to get the resolution of a H3 cell when we only have the h3 or heAddress value.  This PR just exposes the internal implementation.


craigtaverner (approved):
LGTM

PR #86522, opened 2022-05-06
Remove isMeta check during rewrite field-caps responses

dnhatn:
This check is no longer needed because 8.3+ won't handle index field-caps responses from 7.13 or before.

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/fieldcaps/TransportFieldCapabilitiesAction.java
+++ b/server/src/main/java/org/elasticsearch/action/fieldcaps/TransportFieldCapabilitiesAction.java
@@ -43,7 +43,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
-import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import static org.elasticsearch.action.search.TransportSearchHelper.checkCCSVersionCompatibility;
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/elasticsearch/action/fieldcaps/TransportFieldCapabilitiesAction.java
```diff
@@ -75,8 +73,6 @@ public TransportFieldCapabilitiesAction(
         this.clusterService = clusterService;
         this.indexNameExpressionResolver = indexNameExpressionResolver;
         this.indicesService = indicesService;
-        final Set<String> metadataFields = indicesService.getAllMetadataFields();
```
romseygeek, line 78: I think we can also remove IndicesService.getAllMetadataFields() now as well?

romseygeek: And indeed MapperRegistry.getAllMetadataMapperParsers()

romseygeek (approved):
LGTM! We can also remove a couple of other methods that are only used by this code.

server/src/main/java/org/elasticsearch/action/fieldcaps/TransportFieldCapabilitiesAction.java
dnhatn: Yes, I removed them in https://github.com/elastic/elasticsearch/pull/86522/commits/202f5a88f8d6cdbe07d165b6d0fecb11fc1bc7ee.

dnhatn:
@romseygeek Thanks for reviews.

PR #85941, opened 2022-04-15
Adding a view of master history

masseyke:
This commit adds the notion of an in-memory MasterHistory of which nodes have been master for the last 30 minutes that is maintained in memory on each node. It is exposed via the MasterHistoryService. This commit also has a transport action so that you can fetch the master history from any node in the cluster, represented as a List of NodeClients. The list is an ordered list of nodes that have been seen as master for the last 30 minutes, with the oldest first. This action is used by the MasterHistoryService exposed for use via the MasterHistoryService. The local and remote master history representations will be used  to determine if the master has been stable as part of the health API.

PR #86524, opened 2022-05-06
Master stability health indicator part 1 (when a master has been seen recently)

masseyke:
The health indicator for master stability is very large. This is the first PR for the master stability check. It handles the case when we have seen a master node recently (the first two steps in the bulleted list below). The more complicated case when we have not seen a master node recently will be in a subsequent PR.
This indicator reports the health of master stability.
 * If we have had a master within the last 30 seconds, and that master has not changed more than 3 times in the last 30 minutes, then this will report GREEN.
 * If we have had a master within the last 30 seconds, but that master has changed more than 3 times in the last 30 minutes (and that is confirmed by checking with the last-known master), then this will report YELLOW.
 * If we have not had a master within the last 30 seconds, then this will will report RED with one exception. That exception is when all of the following apply:
   * No node is elected master
   * This node is not master eligible
   * Some node is master eligible
   * We ask a master-eligible node to run this indicator
   * That master-eligible node comes back with a result that is not RED.

Since this indicator needs to be able to run when there is no master at all, it does not depend on the dedicated health node (which requires the existence of a master). The indicator in this PR replaces InstanceHasMasterHealthIndicatorService as the "pre-flight check" that is run before running any other health indicators.

Here is an example of a response when the master has changed more than 3 times in the last 30 minutes (so the `stable_master` indicator is yellow). Note that all other indicators respond with status "unknown" since we do not evaluate them if there is no stable master:
```
$ curl localhost:9200/_internal/_health | python3 -mjson.tool
{
    "status": "yellow",
    "cluster_name": "master-test",
    "components": {
        "cluster_coordination": {
            "status": "yellow",
            "indicators": {
                "stable_master": {
                    "status": "yellow",
                    "summary": "The master has changed 4 times in the last 30m",
                    "details": {
                        "recent_masters": [
                            {
                                "Bi7HMyQ8QmSk12bJk6RNiA": {
                                    "name": "master-node",
                                    "ephemeral_id": "PLml3YF1S5CpxR97Ttn2Tw",
                                    "transport_address": "127.0.0.1:9300",
                                    "external_id": "master-node",
                                    "attributes": {
                                        "xpack.installed": "true"
                                    },
                                    "roles": [
                                        "data",
                                        "master"
                                    ]
                                }
                            },
                            {
                                "Bi7HMyQ8QmSk12bJk6RNiA": {
                                    "name": "master-node",
                                    "ephemeral_id": "e56BVOZ_T1OFK7bmG0PTHQ",
                                    "transport_address": "127.0.0.1:9300",
                                    "external_id": "master-node",
                                    "attributes": {
                                        "xpack.installed": "true"
                                    },
                                    "roles": [
                                        "data",
                                        "master"
                                    ]
                                }
                            },
                            {
                                "Bi7HMyQ8QmSk12bJk6RNiA": {
                                    "name": "master-node",
                                    "ephemeral_id": "xQOwd4blR4GN3Kkx2jJB0Q",
                                    "transport_address": "127.0.0.1:9300",
                                    "external_id": "master-node",
                                    "attributes": {
                                        "xpack.installed": "true"
                                    },
                                    "roles": [
                                        "data",
                                        "master"
                                    ]
                                }
                            },
                            {
                                "Bi7HMyQ8QmSk12bJk6RNiA": {
                                    "name": "master-node",
                                    "ephemeral_id": "Frlk9HMVRL2OOPNQhg2E_w",
                                    "transport_address": "127.0.0.1:9300",
                                    "external_id": "master-node",
                                    "attributes": {
                                        "xpack.installed": "true"
                                    },
                                    "roles": [
                                        "data",
                                        "master"
                                    ]
                                }
                            },
                            {
                                "Bi7HMyQ8QmSk12bJk6RNiA": {
                                    "name": "master-node",
                                    "ephemeral_id": "n4ww22m9SHe-QC-xxenhGQ",
                                    "transport_address": "127.0.0.1:9300",
                                    "external_id": "master-node",
                                    "attributes": {
                                        "xpack.installed": "true"
                                    },
                                    "roles": [
                                        "data",
                                        "master"
                                    ]
                                }
                            }
                        ],
                        "current_master": {
                            "Bi7HMyQ8QmSk12bJk6RNiA": {
                                "name": "master-node",
                                "ephemeral_id": "n4ww22m9SHe-QC-xxenhGQ",
                                "transport_address": "127.0.0.1:9300",
                                "external_id": "master-node",
                                "attributes": {
                                    "xpack.installed": "true"
                                },
                                "roles": [
                                    "data",
                                    "master"
                                ]
                            }
                        }
                    },
                    "impacts": [
                        {
                            "severity": 1,
                            "description": "The cluster cannot create, delete, or rebalance indices, and cannot insert or update documents.",
                            "impact_areas": [
                                "ingest"
                            ]
                        },
                        {
                            "severity": 1,
                            "description": "Scheduled tasks such as Watcher, ILM, and SLM will not work. The _cat APIs will not work.",
                            "impact_areas": [
                                "deployment_management"
                            ]
                        },
                        {
                            "severity": 3,
                            "description": "Snapshot and restore will not work.",
                            "impact_areas": [
                                "backup"
                            ]
                        }
                    ],
                    "user_actions": [
                        {
                            "message": "The Elasticsearch cluster does not have a stable master node. This almost always requires expert assistance. Please contact Elastic support to resolve the problem."
                        }
                    ]
                }
            }
        },
        "data": {
            "status": "unknown",
            "indicators": {
                "shards_availability": {
                    "status": "unknown",
                    "summary": "Could not determine indicator state. Cluster state is not stable. Check details for critical issues keeping this indicator from running.",
                    "details": {
                        "reasons": {
                            "stable_master": "yellow"
                        }
                    }
                },
                "ilm": {
                    "status": "unknown",
                    "summary": "Could not determine indicator state. Cluster state is not stable. Check details for critical issues keeping this indicator from running.",
                    "details": {
                        "reasons": {
                            "stable_master": "yellow"
                        }
                    }
                }
            }
        },
        "snapshot": {
            "status": "unknown",
            "indicators": {
                "repository_integrity": {
                    "status": "unknown",
                    "summary": "Could not determine indicator state. Cluster state is not stable. Check details for critical issues keeping this indicator from running.",
                    "details": {
                        "reasons": {
                            "stable_master": "yellow"
                        }
                    }
                },
                "slm": {
                    "status": "unknown",
                    "summary": "Could not determine indicator state. Cluster state is not stable. Check details for critical issues keeping this indicator from running.",
                    "details": {
                        "reasons": {
                            "stable_master": "yellow"
                        }
                    }
                }
            }
        }
    }
}
```
And here is an example response when the master has been stable (not that other indicators are evaluated) :
```
$ curl localhost:9200/_internal/_health | python3 -mjson.tool
{
    "status": "green",
    "cluster_name": "master-test",
    "components": {
        "cluster_coordination": {
            "status": "green",
            "indicators": {
                "stable_master": {
                    "status": "green",
                    "summary": "The cluster has a stable master node",
                    "details": {
                        "recent_masters": [
                            {
                                "Bi7HMyQ8QmSk12bJk6RNiA": {
                                    "name": "master-node",
                                    "ephemeral_id": "n4ww22m9SHe-QC-xxenhGQ",
                                    "transport_address": "127.0.0.1:9300",
                                    "external_id": "master-node",
                                    "attributes": {
                                        "xpack.installed": "true"
                                    },
                                    "roles": [
                                        "data",
                                        "master"
                                    ]
                                }
                            }
                        ],
                        "current_master": {
                            "Bi7HMyQ8QmSk12bJk6RNiA": {
                                "name": "master-node",
                                "ephemeral_id": "n4ww22m9SHe-QC-xxenhGQ",
                                "transport_address": "127.0.0.1:9300",
                                "external_id": "master-node",
                                "attributes": {
                                    "xpack.installed": "true"
                                },
                                "roles": [
                                    "data",
                                    "master"
                                ]
                            }
                        }
                    }
                }
            }
        },
        "data": {
            "status": "green",
            "indicators": {
                "shards_availability": {
                    "status": "green",
                    "summary": "This cluster has all shards available.",
                    "details": {
                        "creating_primaries": 0,
                        "started_primaries": 1,
                        "unassigned_primaries": 0,
                        "initializing_replicas": 0,
                        "started_replicas": 1,
                        "initializing_primaries": 0,
                        "restarting_replicas": 0,
                        "restarting_primaries": 0,
                        "unassigned_replicas": 0
                    }
                },
                "ilm": {
                    "status": "green",
                    "summary": "ILM is running",
                    "details": {
                        "ilm_status": "RUNNING",
                        "policies": 15
                    }
                }
            }
        },
        "snapshot": {
            "status": "green",
            "indicators": {
                "repository_integrity": {
                    "status": "green",
                    "summary": "No repositories configured."
                },
                "slm": {
                    "status": "green",
                    "summary": "No policies configured",
                    "details": {
                        "slm_status": "RUNNING",
                        "policies": 0
                    }
                }
            }
        }
    }
}
```


masseyke:
Relates #85941

server/src/test/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorServiceTests.java
```diff
@@ -0,0 +1,362 @@
...
+
+    }
+
+    public void testMasterGoesNullWithRemoteException() throws Exception {
+        /*
+         * On the local node:
+         *   node1 -> null -> node1 -> null -> node1 -> null -> node1 -> null -> node1
+         * Connecting to the master node throws an exception
+         * In this case, the master identity changed 0 times as seen from the local node. The same master went null 4 times as seen from
+         * my the local node. So we check the remote history. The remote history throws an exception, so the status is
```
andreidan, line 193:
```suggestion
         * the local node. So we check the remote history. The remote history throws an exception, so the status is
```

server/src/test/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorServiceTests.java
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    public void testMasterGoesNullLocallyButRemotelyChangesIdentity() throws Exception {
+        /*
+         * On the local node:
+         *   node1 -> null -> node1 -> null -> node1 -> null -> node1 -> null -> node1
+         * On the master node:
+         *   node1 -> null -> node1 -> node2 -> node3 -> node2 -> node3
+         * In this case, the master identity changed 0 times as seen from the local node. The same master went null 4 times as seen from
+         * my the local node. So we check the remote history. The master only went null here one time, but it changed identity 4 times. So
```
andreidan, line 230:
```suggestion
         * the local node. So we check the remote history. The master only went null here one time, but it changed identity 4 times. So
```

server/src/test/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorServiceTests.java
```diff
@@ -0,0 +1,362 @@
...
+         * we still get a status of YELLOW. (Note: This scenario might not be possible in the real world for a couple of reasons7, but it
```
andreidan, line 231:
```suggestion
         * we still get a status of YELLOW. (Note: This scenario might not be possible in the real world for a couple of reasons, but it
```

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,433 @@
...
+ * confirmed by checking with the last-known master), then this will report YELLOW.
+ * If we have not had a master within the last 30 seconds, then this will will report RED with one exception. That exception is when:
+ * (1) no node is elected master, (2) this node is not master eligible, (3) some node is master eligible, (4) we ask a master-eligible node
+ * to run this indicator, and (5) it comes back with a result that is not RED.
+ * Since this indicator needs to be able to run when there is no master at all, it does not depend on the dedicated health node (which
+ * requires the existence of a master).
+ */
+public class StableMasterHealthIndicatorService implements HealthIndicatorService, ClusterStateListener {
+
+    public static final String NAME = "stable_master";
```
andreidan, line 59: As this is user facing, should we be a bit more verbose - I believe we referred to this as `master_is_stable` generally. Should we rename this to `master_is_stable`?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,433 @@
...
+    private static final String DETAILS_EXCEPTION_FETCHING_HISTORY_STACK_TRACE = "exception_fetching_history_stack_trace";
+
+    // Impacts of having an unstable master:
+    private static final String UNSTABLE_MASTER_INGEST_IMPACT = "The cluster cannot create, delete, or rebalance indices, and cannot "
+        + "insert or update documents.";
+    private static final String UNSTABLE_MASTER_DEPLOYMENT_MANAGEMENT_IMPACT = "Scheduled tasks such as Watcher, ILM, and SLM will not "
+        + "work. The _cat APIs will not work.";
+    private static final String UNSTABLE_MASTER_BACKUP_IMPACT = "Snapshot and restore will not work.";
+
+    public static final Setting<TimeValue> VERY_RECENT_PAST_SETTING = Setting.timeSetting(
```
andreidan, line 105: Is this history not already configured for the master service? I wouldn't think we'd need this configuration again? (as the master history service has a definition of "recent" already via `maxHistoryAge` https://github.com/elastic/elasticsearch/blob/f69c7396d9334d16c2b6a1614d5f32fb8a250d1c/server/src/main/java/org/elasticsearch/cluster/coordination/MasterHistory.java#L37 )

Do we envision these 2 settings to ever diverge? What would the consequence of that be and why would we want to?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,433 @@
...
+    /**
+     * This method returns the only user action that is relevant when the master is unstable -- contact support.
+     * @param explain If true, the returned list includes a UserAction to contact support, otherwise an empty list
+     * @return a single UserAction instructing users to contact support.
+     */
+    private List<UserAction> getContactSupportUserActions(boolean explain) {
+        if (explain) {
+            UserAction.Definition contactSupport = new UserAction.Definition(
+                "contact_support",
+                "The Elasticsearch cluster does not have a stable master node. This almost always requires expert assistance. Please "
```
andreidan, line 237:
```suggestion
                "The Elasticsearch cluster does not have a stable master node. Please "
```

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,433 @@
...
+                    + "contact Elastic support to resolve the problem.",
```
andreidan, line 238:
```suggestion
                    + "contact Elastic Support (https://support.elastic.co) to discuss available options.",
```

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,433 @@
...
+                 */
+                if (master != null && clusterService.localNode().equals(master) == false) {
+                    masterHistoryService.refreshRemoteMasterHistory(master);
+                }
+            }
+        }
+    }
+
+    /**
+     * XContentBuilder doesn't deal well with ToXContentFragments (which is what DiscoveryNodes are). Also XContentBuilder doesn't do well
```
andreidan, line 415: Can you maybe elaborate what doesn't work with `ToXContentFragments` when it comes to nulls ?

Maybe I'm missing something but wouldn't we handle the nulls when printing the `details` map? I don't think we should be needing another entity to wrap the `DiscoveryNode` as we don't print it directly but via a map that's in our control (the `details` map)

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,433 @@
...
+    private final MasterHistoryService masterHistoryService;
+    /**
+     * This is the amount of time we use to make the initial decision -- have we seen a master node in the very recent past?
+     */
+    private final TimeValue veryRecentPast;
+    /**
+     * This is the number of times that it is OK for the master history to show a transition from a non-null master to a null master before
+     * it starts impacting the health status.
+     */
+    private final int acceptableNullTransitions;
```
andreidan, line 71: This field is something we always read by `+1`-ing it. That seems a bit error prone to me. Should we have a field that is has the effective value of `acceptableNullTransitions + 1` and just use that? (ie. maybe we define a threshold for when things go wrong and name it accordingly, as opposed to what's acceptable? ie `witnessedNullTransitionsThreshold` or `masterWentNullThreshold` that would default to 4 ? or something better named :) )

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,433 @@
...
+    /**
+     * This is the number of times that it is OK for the master history to show a transition one non-null master to a different non-null
+     * master before it starts impacting the health status.
+     */
+    private final int acceptableIdentityChanges;
```
andreidan, line 76: Same as before. Should we have this as a threshold so we don't have to +1 it everywhere?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,433 @@
...
+     * @param masterChanges The number of times that the local machine has seen the master identity change in the last 30 minutes
+     * @param explain Whether to calculate and include the details in the result
+     * @return The HealthIndicatorResult for the given localMasterHistory
+     */
+    private HealthIndicatorResult calculateWhenMasterHasChangedIdentity(
+        MasterHistory localMasterHistory,
+        int masterChanges,
+        boolean explain
+    ) {
+        logger.trace("Have seen {} master changes in the last {}}", masterChanges, localMasterHistory.getMaxHistoryAge());
```
andreidan, line 188:
```suggestion
        logger.trace("Have seen {} master changes in the last {}", masterChanges, localMasterHistory.getMaxHistoryAge());
```

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,433 @@
...
+        }
+    }
+
+    /**
+     * Returns the health result for the case when we have seen a master recently (at some point in the last 30 seconds).
+     * @param localMasterHistory The master history as seen from the local machine
+     * @param explain Whether to calculate and include the details and user actions in the result
+     * @return The HealthIndicatorResult for the given localMasterHistory
+     */
+    private HealthIndicatorResult calculateWhenHaveSeenMasterRecently(MasterHistory localMasterHistory, boolean explain) {
```
andreidan, line 161: The name of this method indicates it's looking for the time something happened - the *when*. Which is not correct AFAICS. Should we name it differently? ie. `onSeenMaster` / `checkIfCurrentMasterIsStable` ? Or something along those lines?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,433 @@
...
+        return createIndicator(stableMasterStatus, summary, new SimpleHealthIndicatorDetails(details), impacts, userActions);
+    }
+
+    /**
+     * Returns the health result for the case when we have NOT seen a master recently (at some point in the last 30 seconds).
+     * @param localMasterHistory The master history as seen from the local machine
+     * @param explain Whether to calculate and include the details in the result
+     * @return The HealthIndicatorResult for the given localMasterHistory
+     */
+    private HealthIndicatorResult calculateWhenHaveNotSeenMasterRecently(MasterHistory localMasterHistory, boolean explain) {
```
andreidan, line 361: As above, I think this should be named according to what it does/what it looks for as opposed to when it's called (the *when* in the name is confusing as it points to calculating the time when an event happened)

Maybe `onNotSeenMaster` / `onNoLocalMaster` or something similar?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,433 @@
...
+         * though if the remote history has not arrived yet.
+         * If the local node is not master and the master history from the master itself reports that the master has gone null repeatedly
+         *  or changed identity repeatedly, then we have a problem (the master has confirmed what the local node saw).
+         */
+        boolean masterConfirmedUnstable = localNodeIsMaster
+            || remoteHistory == null
+            || MasterHistory.hasMasterGoneNullAtLeastNTimes(remoteHistory, acceptableNullTransitions + 1)
+            || MasterHistory.getNumberOfMasterIdentityChanges(remoteHistory) > acceptableIdentityChanges;
+        if (masterConfirmedUnstable) {
+            if (localNodeIsMaster == false && remoteHistory == null) {
```
andreidan, line 300: if remoteHistory is null but there's an in-flight request to get it I believe the status should still be GREEN?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,433 @@
...
+         * If the local node is master, then we have a confirmed problem (since we now know that from this node's point of view the
+         * master is unstable).
+         * If the local node is not master but the remote history is null then we have a problem (since from this node's point of view the
+         * master is unstable, and we were unable to get the master's own view of its history). It could just be a short-lived problem
+         * though if the remote history has not arrived yet.
+         * If the local node is not master and the master history from the master itself reports that the master has gone null repeatedly
+         *  or changed identity repeatedly, then we have a problem (the master has confirmed what the local node saw).
+         */
+        boolean masterConfirmedUnstable = localNodeIsMaster
+            || remoteHistory == null
```
andreidan, line 296: if remoteHistory is null but there's an in-flight request to get it I believe the status should still be GREEN?

andreidan (commented):

Wasn't sure if you'd want a review for this as no reviewer was selected but went through it and added some thoughts.

andreidan:
@DaveCTurner what do you think about using the disruption tests infrastructure to add some integration tests for this indicator?

evilwon22:
Check it out look very nice. Keep working have a good day bye-bye

DaveCTurner:
Yes we should have some integ tests for this.

I also suggest trying to add some tests that use the `CoordinatorTests` framework since that would let you simulate various disruptions plus the passage of time (which is important for this indicator) as well as being single-threaded and repeatable.

server/src/test/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorServiceTests.java
masseyke: Oops, good catch

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
masseyke: I can remove this class and inline the logic.

masseyke: Oh, the reason is that this class does not actually control the printing -- SimpleHealthIndicatorDetails does. I'm not sure how best to handle it in there. We don't want to inspect the type in there and do special handling for different types (I don't think). So I have to change it from DiscoveryNode to something that won't fail. I think the options are to either convert it to a non-fragment ToXContentObject (which is what is being done now) or maybe just convert it to a String? What do you think?

masseyke: Or another option is that I could move this logic into the toXContent method of SimpleHealthIndicatorDetails for any ToXContent value where isFragment() is true

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
masseyke: For this one we actually don't `+1` it anywhere, so it's probably good to stay as-is right?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
andreidan: Ah, that's correct, my bad. Thanks for the clarification!

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
andreidan: Could we take full control over the representation of the details (and avoid the `SimpleHealthIndicatorDetails` altogether) like we do in the `InstanceHasMasterIndicator`? https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/cluster/coordination/InstanceHasMasterHealthIndicatorService.java#L73

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
andreidan: Ah, that's true. For consistency would it make sense to go with the threshold pattern and adapt the checks accordingly?
ie. have the condition 
```
if (masterChanges >= identityChangeThreshold) {
```

What do you think?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
masseyke: We had multiple offline conversations about this and decided to make two changes:

1. We're introducing the notion of a remote history becoming stale (#86936) 
2. If the remote master history is unset or stale we'll return the status as GREEN

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
masseyke: I can do that.

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
masseyke: OK I switched over to custom HealthIndicatorDetails in this class (not counting the one for the "haven't seen a master recently" branch that is just a stub in this PR), and I've added several unit tests around it. It seems to be working well.

andreidan (commented):

I left some comments.

I am wondering if it'd make sense to separate the "master stability diagnosis" part of this into its own service (that'll listen to cluster change events, and perform the needed diagnosis when requested).

I've recommended something similar for the shards_availability indicator and we should revise it there too (// cc @jbaiera )

Currently, the indicator is mixing the diagnosis logic with the representation of the diagnosis. 
Modularising it a bit might aid with future requirements where we'd want (possibly) the diagnosing logic to be used in other transports/services.

What do you think?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,449 @@
...
+                    builder.nullField("node_id");
+                    builder.nullField("name");
+                }
+            });
+            List<DiscoveryNode> recentMasters = localMasterHistory.getNodes();
+            builder.array(DETAILS_RECENT_MASTERS, arrayXContentBuilder -> {
+                for (DiscoveryNode recentMaster : recentMasters) {
+                    if (recentMaster != null) {
+                        builder.startObject();
+                        builder.field("node_id", recentMaster.getId());
```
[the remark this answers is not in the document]
masseyke: There's no notion of timestamp here, mostly so that we don't leak out the notion of the machine's relative timestamps that don't mean anything outside of the JVM. You think we ought to add some public notion of time to https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/cluster/coordination/MasterHistory.java? That might be best in a couple of follow-up PRs?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,449 @@
...
+    private final int unacceptableIdentityChanges;
+
+    private static final Logger logger = LogManager.getLogger(StableMasterHealthIndicatorService.class);
+
+    // This is the default amount of time we look back to see if we have had a master at all, before moving on with other checks
+    private static final TimeValue DEFAULT_VERY_RECENT_PAST = new TimeValue(30, TimeUnit.SECONDS);
+    private static final TimeValue SMALLEST_ALLOWED_VERY_RECENT_PAST = new TimeValue(1, TimeUnit.SECONDS);
+
+    // This is the default number of times that it is OK to have a master go null. Any more than this will be reported as a problem
+    private static final int DEFAULT_ACCEPTABLE_NULL_TRANSITIONS = 3;
```
dakrone: I don't think we need all of these as private static variables, we can put the comment and the actual value (3 in this case) in the setting itself, which is already public and static?

I prefer to have the hard value with the setting itself, so if you look at the `Setting` you don't have to do another reference jump to see the default value.

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,449 @@
...
+
+    // Impacts of having an unstable master:
+    private static final String UNSTABLE_MASTER_INGEST_IMPACT = "The cluster cannot create, delete, or rebalance indices, and cannot "
+        + "insert or update documents.";
+    private static final String UNSTABLE_MASTER_DEPLOYMENT_MANAGEMENT_IMPACT = "Scheduled tasks such as Watcher, ILM, and SLM will not "
+        + "work. The _cat APIs will not work.";
+    private static final String UNSTABLE_MASTER_BACKUP_IMPACT = "Snapshot and restore will not work.";
+
+    public static final Setting<TimeValue> VERY_RECENT_PAST_SETTING = Setting.timeSetting(
+        "health.master_history.very_recent_past",
```
dakrone: I don't think this is a very descriptive name (just "very_recent_past"), I'd prefer something like "stability_window" to (hopefully?) clarify a little bit more what it does from the name alone.

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,449 @@
...
+        DEFAULT_VERY_RECENT_PAST,
+        SMALLEST_ALLOWED_VERY_RECENT_PAST,
+        Setting.Property.NodeScope
+    );
+
+    public static final Setting<Integer> ACCEPTABLE_NULL_TRANSITIONS_SETTING = Setting.intSetting(
+        "health.master_history.acceptable_null_transitions",
```
dakrone: I don't think we should "leak" the word/concept of 'null' here, perhaps "`acceptable_no_master_transitions`" or "`acceptable_none_transitions`"?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,449 @@
...
+    private HealthIndicatorResult calculateOnMasterHasChangedIdentity(
+        MasterHistory localMasterHistory,
+        int masterChanges,
+        boolean explain
+    ) {
+        logger.trace("Have seen {} master changes in the last {}", masterChanges, localMasterHistory.getMaxHistoryAge());
+        HealthStatus stableMasterStatus = HealthStatus.YELLOW;
+        String summary = String.format(
+            Locale.ROOT,
+            "The master has changed %d times in the last %s",
```
dakrone, line 198:
```suggestion
            "The elected master node has changed %d times in the last %s",
```

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,449 @@
...
+                    if (recentMaster != null) {
+                        builder.startObject();
+                        builder.field("node_id", recentMaster.getId());
+                        builder.field("name", recentMaster.getName());
+                        builder.endObject();
+                    }
+                }
+            });
+            return builder.endObject();
+        } : HealthIndicatorDetails.EMPTY;
```
dakrone, line 244: I think we should limit ternary operators to single-line statements, since it's difficult to parse as a human when it spans tens of lines.

Perhaps we could do something like:

```java
if (explain == false) {
    return HealthIndicatorDetails.EMPTY;
} else {
    return <this big thing>;
}
```

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,449 @@
...
+
+    /**
+     * Returns the list of the impacts of an unstable master node.
+     * @return The list of the impacts of an unstable master node
+     */
+    private List<HealthIndicatorImpact> getUnstableMasterImpacts() {
+        List<HealthIndicatorImpact> impacts = new ArrayList<>();
+        impacts.add(new HealthIndicatorImpact(1, UNSTABLE_MASTER_INGEST_IMPACT, List.of(ImpactArea.INGEST)));
+        impacts.add(new HealthIndicatorImpact(1, UNSTABLE_MASTER_DEPLOYMENT_MANAGEMENT_IMPACT, List.of(ImpactArea.DEPLOYMENT_MANAGEMENT)));
+        impacts.add(new HealthIndicatorImpact(3, UNSTABLE_MASTER_BACKUP_IMPACT, List.of(ImpactArea.BACKUP)));
```
dakrone, line 275: We might as well hard-code this as a `static` list instead of constructing it every time, right? Since it doesn't change based on anything else.

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,449 @@
...
+     * If the master transitions from one non-null master to a different non-null master at least this many times it starts impacting the
+     * health status.
+     */
+    private final int unacceptableIdentityChanges;
+
+    private static final Logger logger = LogManager.getLogger(StableMasterHealthIndicatorService.class);
+
+    // This is the default amount of time we look back to see if we have had a master at all, before moving on with other checks
+    private static final TimeValue DEFAULT_VERY_RECENT_PAST = new TimeValue(30, TimeUnit.SECONDS);
+    private static final TimeValue SMALLEST_ALLOWED_VERY_RECENT_PAST = new TimeValue(1, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
masseyke: Done.

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,449 @@
...
+            || remoteHistoryException != null
+            || (remoteHistory != null
+                && (MasterHistory.hasMasterGoneNullAtLeastNTimes(remoteHistory, unacceptableNullTransitions)
+                    || MasterHistory.getNumberOfMasterIdentityChanges(remoteHistory) >= unacceptableIdentityChanges));
+        if (masterConfirmedUnstable) {
+            logger.trace("The master node {} thinks it is unstable", master);
+            final HealthStatus stableMasterStatus = HealthStatus.YELLOW;
+            String summary = String.format(
+                Locale.ROOT,
+                "The cluster's master has alternated between %s and no master multiple times in the last %s",
```
dakrone, line 324:
```suggestion
                "The cluster's elected master node has alternated between %s and no elected master node multiple times in the last %s",
```

(I clarify here because I don't want to confuse a user between there being no `master` nodes in the cluster, and no *elected* master node in the cluster)

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,449 @@
...
+     * @param localMasterHistory The localMasterHistory
+     * @param remoteHistoryException An exception that was found when retrieving the remote master history. Can be null
+     * @return The HealthIndicatorDetails
+     */
+    private HealthIndicatorDetails getHealthIndicatorDetailsOnMasterHasFlappedNull(
+        boolean explain,
+        MasterHistory localMasterHistory,
+        @Nullable Exception remoteHistoryException
+    ) {
+        return explain ? (builder, params) -> {
```
dakrone, line 356: Same comment here about keeping ternary things to single-line statements.

dakrone (commented):
I left some comments but this generally looks good!

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
masseyke: I changed the setting name and the setting variable name, but left internal variable names and method names as they were b/c using the word "null" really makes it easier to follow internally. Let me know if you disagree.

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
masseyke: Done.

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
masseyke: Changed to NODE_HAS_MASTER_LOOKUP_TIMEFRAME_SETTING based on feedback from Andrei.

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -0,0 +1,449 @@
...
+     */
+    private HealthIndicatorResult calculateOnHaveNotSeenMasterRecently(MasterHistory localMasterHistory, boolean explain) {
+        // NOTE: The logic in this method will be implemented in a future PR
+        HealthStatus stableMasterStatus = HealthStatus.RED;
+        String summary = "No master has been observed recently";
+        Map<String, Object> details = new HashMap<>();
+        Collection<HealthIndicatorImpact> impacts = getUnstableMasterImpacts();
+        List<UserAction> userActions = getContactSupportUserActions(explain);
+        return createIndicator(
+            stableMasterStatus,
```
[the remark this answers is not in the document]
masseyke: > I am wondering if it'd make sense to separate the "master stability diagnosis" part of this into its own service (that'll listen to cluster change events, and perform the needed diagnosis when requested).
> 
> I've recommended something similar for the shards_availability indicator and we should revise it there too (// cc @jbaiera )
> 
> Currently, the indicator is mixing the diagnosis logic with the representation of the diagnosis. Modularising it a bit might aid with future requirements where we'd want (possibly) the diagnosing logic to be used in other transports/services.
> 
> What do you think?

It sounds reasonable, but likely a very big change. What do you think about considering that for a follow-up PR (or PRs)?

masseyke:
> I am wondering if it'd make sense to separate the "master stability diagnosis" part of this into its own service (that'll listen to cluster change events, and perform the needed diagnosis when requested).
> 
> I've recommended something similar for the shards_availability indicator and we should revise it there too (// cc @jbaiera )
> 
> Currently, the indicator is mixing the diagnosis logic with the representation of the diagnosis. Modularising it a bit might aid with future requirements where we'd want (possibly) the diagnosing logic to be used in other transports/services.
> 
> What do you think?

@andreidan it sounds reasonable, but likely a very big change. What do you think about considering that for a follow-up PR (or PRs)?

andreidan (approved):
LGTM, thanks for iterating on this Keith

Left a few nits but this looks great. Thanks for adding all the tests too ! 🚀

Issue #56305, opened 2020-05-06
date_histogram's offset doesn't like daylight savings time


nik9000:
`date_histogram`'s [offset](https://www.elastic.co/guide/en/elasticsearch/reference/7.8/search-aggregations-bucket-datehistogram-aggregation.html#search-aggregations-bucket-datehistogram-offset) advertises itself as shifting around the start of the histogram interval. It says:

```
For example, when using an interval of day, each bucket runs from midnight to midnight. Setting the offset parameter to +6h changes each bucket to run from 6am to 6am
```

This is mostly true but not actually true. What it really does is shift each document's time back the offset, perform the rounding, and then shift it forward by the offset. If it weren't for daylight savings time this would do exactly as advertised. But, daylight savings time is hard. Check this out:
```
ES='curl -s -uelastic:password'
ES_DELETE="$ES -XDELETE"
ES_POST="$ES -XPOST -HContent-Type:application/json"
ES_PUT="$ES -XPUT -HContent-Type:application/json"

$ES_DELETE localhost:9200/test
$ES_PUT localhost:9200/test -d'{
  "settings": {
    "number_of_replicas": 0
  },
  "mappings": {
    "properties": {
      "timestamp": { "type": "date" }
    }
  }
}'

$ES_POST 'localhost:9200/test/_doc?pretty&refresh' -d'{"date": "2020-03-09T04:00:00Z"}'
$ES_POST 'localhost:9200/test/_doc?pretty&refresh' -d'{"date": "2020-03-09T05:00:00Z"}'
$ES_POST 'localhost:9200/test/_doc?pretty&refresh' -d'{"date": "2020-03-09T06:00:00Z"}'
$ES_POST 'localhost:9200/test/_doc?pretty&refresh' -d'{"date": "2020-03-09T07:00:00Z"}'
$ES_POST 'localhost:9200/test/_doc?pretty&refresh' -d'{"date": "2020-03-09T08:00:00Z"}'

$ES_POST 'localhost:9200/test/_search?size=0&pretty' -d'{
  "aggs": {
    "oh_no": {
      "date_histogram": {
        "field": "date",
        "calendar_interval": "1d",
        "time_zone": "America/New_York",
        "offset": "+1h"
      }
    }
  }
}'
```

That results in:
```
  "aggregations" : {
    "oh_no" : {
      "buckets" : [
        {
          "key_as_string" : "2020-03-08T00:00:00.000-05:00",
          "key" : 1583643600000,
[...]

PR #86526, opened 2022-05-06
test: add tests to date_histogram with and without offset and timezone

salvatore-campagna:
The first of these test was meant to reproduce issue #56305. Anyway the result looks correct to me.
The second test is there to have the same test but without the `offset`.

### DIFF
--- a/rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.aggregation/360_date_histogram.yml
+++ b/rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.aggregation/360_date_histogram.yml
@@ -44,6 +44,17 @@ setup:
                 type: date_nanos
 
   - do:
+      indices.create:
+        index: timezone_daylight_test
+        body:
+          settings:
+            number_of_replicas: 0
+          mappings:
+            properties:
+              date:
+                type: date
+
+  - do:
       bulk:
         index: test_date_hist
         refresh: true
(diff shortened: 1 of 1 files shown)


rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.aggregation/360_date_histogram.yml
```diff
@@ -356,3 +385,58 @@ setup:
 
   - match: { hits.total.value: 1 }
   - length: { aggregations.datehisto.buckets: 3 }
+
+---
+"Daylight with offset date_histogram test":
+  - skip:
+      version: "- 8.2.99"
```
salvatore-campagna, line 392: I know this is incorrect...I will try to see if I can find when this was fixed (assuming the test results are correct)

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.aggregation/360_date_histogram.yml
```diff
@@ -356,3 +385,58 @@ setup:
...
+  - match: { aggregations.date_histogram_daily.buckets.0.key_as_string: "2020-03-08T01:00:00.000-05:00" }
+  - match: { aggregations.date_histogram_daily.buckets.0.key: 1583647200000 }
+  - match: { aggregations.date_histogram_daily.buckets.0.doc_count: 2 }
+  - match: { aggregations.date_histogram_daily.buckets.1.key_as_string: "2020-03-09T01:00:00.000-04:00" }
+  - match: { aggregations.date_histogram_daily.buckets.1.key: 1583730000000 }
+  - match: { aggregations.date_histogram_daily.buckets.1.doc_count: 4 }
+
+---
+"Daylight without offset date_histogram test":
+  - skip:
```
not-napoleon, line 419: Without DST, this probably doesn't need a skip version at all.

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.aggregation/360_date_histogram.yml
not-napoleon: As noted, I think 7.17 is good enough.

not-napoleon (approved):
LGTM

nik9000 (approved):
OK then. I don't know how or when we fixed this, but yay.

PR #86528, opened 2022-05-06
Clarify risks around ordering terms aggregation

not-napoleon:
I found myself confused by the docs on ordering terms aggregation by sub aggregations, and took a pass at cleaning it up.

### DIFF
--- a/docs/reference/aggregations/bucket/terms-aggregation.asciidoc
+++ b/docs/reference/aggregations/bucket/terms-aggregation.asciidoc
@@ -212,11 +212,12 @@ By default, the `terms` aggregation orders terms by descending document
 `_count`.  This produces a bounded <<terms-agg-doc-count-error,document count>>
 error that {es} can report.
 
-You can use the `order` parameter to specify a different sort order.  It is extremely
-easy to create a terms ordering that will just return wrong results, and not
-obvious to see when you have done so.  Change this only with caution.
+You can use the `order` parameter to specify a different sort order, but we
+don't recommend it.  It is extremely easy to create a terms ordering that will
+just return wrong results, and not obvious to see when you have done so.
+Change this only with caution.
 
-WARNING: Avoid using `"order": { "_count": "asc" }`. If you need to find rare
+WARNING: Especially avoid using `"order": { "_count": "asc" }`. If you need to find rare
 terms, use the
 <<search-aggregations-bucket-rare-terms-aggregation,`rare_terms`>> aggregation
 instead. Due to the way the `terms` aggregation
@@ -253,11 +254,17 @@ WARNING: Sorting by a sub aggregation generally produces incorrect ordering, due
 <<search-aggregations-bucket-terms-aggregation-shard-size,gets results from
 shards>>.
 
-There are two cases when sub-aggregation sorting is safe.  Sorting by a maximum
-in descending order, or a minimum in ascending order will be correct.  This
-works only because the sort order aligns with the behavior of the sub
-aggregations, i.e. sorting maxes by largest value first or mins by lowest value
-first.
+There are two cases when sub-aggregation ordering is safe and returns correct
+results: sorting by a maximum in descending order, or sorting by a minimum in
+ascending order. These approaches work because they align with the behavior of
+sub aggregations. That is, if you're looking for the largest maximum or the
+smallest minimum, the global answer (from combined shards) must be included in
+one of the local shard answers. Conversely, the smallest maximum and largest
+minimum wouldn't be accurately computed.
+
+Note also that in these cases, the ordering is correct but the doc counts and
+non-ordering sub aggregations may still have errors (and {es} does not calculate a
+bound for those errors).
 
 Ordering the buckets by single value metrics sub-aggregation (identified by the aggregation name):
 
--- a/docs/reference/aggregations/bucket/terms-aggregation.asciidoc
+++ b/docs/reference/aggregations/bucket/terms-aggregation.asciidoc
@@ -210,7 +210,7 @@ determined and is given a value of -1 to indicate this.
 
 By default, the `terms` aggregation orders terms by descending document
 `_count`.  This produces bounded error, which elasticsearch can report (see
-<<terms-aggregation-doc-count-error>>.
+<<terms-agg-doc-count-error>>.
 
 You can use the `order` parameter to specify a different sort order.  It is extremely
 easy to create a terms ordering that will just return wrong results, and not
--- a/docs/reference/aggregations/bucket/terms-aggregation.asciidoc
+++ b/docs/reference/aggregations/bucket/terms-aggregation.asciidoc
@@ -209,8 +209,8 @@ determined and is given a value of -1 to indicate this.
 ==== Order
 
 By default, the `terms` aggregation orders terms by descending document
-`_count`.  This produces bounded error, which elasticsearch can report (see
-<<terms-agg-doc-count-error>>.)
+`_count`.  This produces a bounded <<terms-agg-doc-count-error,document count>>
+error that {es} can report.
 
 You can use the `order` parameter to specify a different sort order.  It is extremely
 easy to create a terms ordering that will just return wrong results, and not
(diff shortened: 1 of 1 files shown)


docs/reference/aggregations/bucket/terms-aggregation.asciidoc
```diff
@@ -208,8 +208,13 @@ determined and is given a value of -1 to indicate this.
 [[search-aggregations-bucket-terms-aggregation-order]]
 ==== Order
 
-By default, the `terms` aggregation orders terms by descending document `_count`.
-Use the `order` parameter to specify a different sort order.
+By default, the `terms` aggregation orders terms by descending document
+`_count`.  This produces bounded error, which elasticsearch can report (see
+<<terms-agg-doc-count-error>>.
```
nik9000, line 213: missing right parenthesis.

nik9000 (approved):

docs/reference/aggregations/bucket/terms-aggregation.asciidoc
```diff
@@ -208,8 +208,13 @@ determined and is given a value of -1 to indicate this.
...
 
-By default, the `terms` aggregation orders terms by descending document `_count`.
-Use the `order` parameter to specify a different sort order.
+By default, the `terms` aggregation orders terms by descending document
+`_count`.  This produces bounded error, which elasticsearch can report (see
+<<terms-agg-doc-count-error>>.)
+
+You can use the `order` parameter to specify a different sort order.  It is extremely
+easy to create a terms ordering that will just return wrong results, and not
+obvious to see when you have done so.  Change this only with caution.
```
lockewritesdocs, line 217: Do we want to discourage use of the `order` parameter altogether? Given that it's very easy to create a terms ordering that returns inaccurate results, I wonder if we want to expand on the **WARNING** message on line 219 and just say "don't use this". 

WARNING: Avoid using the `order` parameter to specify a different sort order. Due to the way
the `terms` aggregation
<<search-aggregations-bucket-terms-aggregation-shard-size,gets terms from
shards>>, sorting by ascending doc count often produces inaccurate results. If you need to
find rare terms, use the <<search-aggregations-bucket-rare-terms-aggregation,`rare_terms`>>
aggregation instead.

lockewritesdocs, line 217: Do we want to discourage use of the `order` parameter altogether? Given that it's very easy to create a terms ordering that returns inaccurate results, I wonder if we want to expand on the **WARNING** message on line 219 and just say "don't use this". 

> WARNING: Avoid using the `order` parameter to specify a different sort order. Due to the way
> the `terms` aggregation
> <<search-aggregations-bucket-terms-aggregation-shard-size,gets terms from
> shards>>, sorting by ascending doc count often produces inaccurate results. If you need to
> find rare terms, use the <<search-aggregations-bucket-rare-terms-aggregation,`rare_terms`>>
> aggregation instead.

docs/reference/aggregations/bucket/terms-aggregation.asciidoc
```diff
@@ -208,8 +208,13 @@ determined and is given a value of -1 to indicate this.
 [[search-aggregations-bucket-terms-aggregation-order]]
 ==== Order
 
-By default, the `terms` aggregation orders terms by descending document `_count`.
-Use the `order` parameter to specify a different sort order.
+By default, the `terms` aggregation orders terms by descending document
+`_count`.  This produces bounded error, which elasticsearch can report (see
+<<terms-agg-doc-count-error>>.)
```
lockewritesdocs, line 213:
```suggestion
`_count`.  This produces a bounded <<terms-agg-doc-count-error,document count>>
error that {es} can report.
```

docs/reference/aggregations/bucket/terms-aggregation.asciidoc
```diff
@@ -218,24 +223,14 @@ instead. Due to the way the `terms` aggregation
...
-      }
-    }
-  }
-}
---------------------------------------------------
+===== Ordering by the term value
+In this case, the buckets are ordered by the actual term values, e.g.
+Lexicographically for keywords or numerically for numbers.  This sorting is
+safe in both ascending and descending directions, and will always produce
+correct results.
```
lockewritesdocs, line 231:
```suggestion
In this case, the buckets are ordered by the actual term values, such as
lexicographic order for keywords or numerically for numbers. This sorting is
safe in both ascending and descending directions, and produces accurate
results.
```

lockewritesdocs: We typically want to avoid phrases that include "always", which might not always be true 💡 . Instead, I think we can say that this sorting "produces accurate results".

docs/reference/aggregations/bucket/terms-aggregation.asciidoc
```diff
@@ -252,13 +247,17 @@ GET /_search
...
-shards>>, sorting on a `max` sub-aggregation in _ascending_ order often produces
-inaccurate results. However, sorting on a `max` sub-aggregation in _descending_
-order is typically safe.
+shards>>.
+
+There are two cases when sub-aggregation sorting is safe.  Sorting by a maximum
+in descending order, or a minimum in ascending order will be correct.  This
+works only because the sort order aligns with the behavior of the sub
+aggregations, i.e. sorting maxes by largest value first or mins by lowest value
+first.
```
lockewritesdocs, line 260:
```suggestion
There are two cases when sub-aggregation sorting is safe and returns correct results:
sorting by a maximum in descending order, or a sorting by a minimum in ascending order. 
Because the sort order aligns with the behavior of the sub aggregations, you can sort
`max` buckets by largest value first or `min` buckets by lowest value first.

```

lockewritesdocs: I think that you meant `max` and `min` buckets, but I might have misinterpreted your intention. Let me know if I'm off base here.

docs/reference/aggregations/bucket/terms-aggregation.asciidoc
```diff
@@ -384,6 +383,27 @@ their `doc_count` in descending order.
 NOTE: In the event that two buckets share the same values for all order criteria the bucket's term value is used as a
 tie-breaker in ascending alphabetical order to prevent non-deterministic ordering of buckets.
 
+===== Ordering by count ascending
+
+Ordering by count ascending has unbounded error, and elasticsearch is unable to
+accurately report the error in the doc counts.  We strongly recommend not doing
+this.  If you really must, this is how you sort on doc count ascending:
```
lockewritesdocs, line 390:
```suggestion
Ordering terms by ascending document `_count` produces an unbounded error that
{es} can't accurately report. We therefore strongly recommend against using
`"order": { "_count": "asc" }` as shown in the following example:
```

lockewritesdocs (commented):
@not-napoleon, I left a few comments and suggested changes. I'm not very familiar with aggregations, so I might have gotten something wrong. If so, let me know and we can iterate further.

docs/reference/aggregations/bucket/terms-aggregation.asciidoc
not-napoleon: Nik and I spent like 20 minutes proving it before I typed "always", but yeah, as Knuth said "be careful, I've only proven it correct, I haven't actually tested it".

docs/reference/aggregations/bucket/terms-aggregation.asciidoc
lockewritesdocs: Thanks for the context @not-napoleon. I think your instincts are correct here. We might want to add a little more context about why it's so easy to create a terms ordering that won't return correct results. What can we replace as the [reason] in the following sentence? Is this easily explained?

> You can use the `order` parameter to specify a different sort order, but we don't recommend it. Because [reason], it's easy to create a terms ordering that returns incorrect results. Use this parameter with caution.

docs/reference/aggregations/bucket/terms-aggregation.asciidoc
lockewritesdocs: Tricky, indeed! Here's another rendering that hopefully is more accurate:

> There are two cases when sub-aggregation ordering is safe and returns correct results: sorting by a maximum in descending order, or sorting by a minimum in ascending order. These approaches work because they align with the behavior of sub aggregations. That is, if you're looking for the largest maximum or the smallest minimum, the global answer (from combined shards) must be included in one of the local shard answers. Conversely, the smallest maximum and largest minimum wouldn't be accurately computed.

not-napoleon: Thanks, that reads pretty well to me.  I'll make the change.

lockewritesdocs (approved):
LGTM 🦖 Thanks for iterating @not-napoleon.

PR #86537, opened 2022-05-08
AbstractAnalyzerProvider does not need to extend AbstractIndexComponent

original-brownbear:
Remove the inheritance here to make instances smaller and speed up many-shards benchmarks a little (setting up the unused loggers turns out to be costly, we can fix this for other inheritors of `AbstractIndexComponent` as well).
Did not remove the dead arguments from the constructors in this PR as that would have been a
very noisy change.


original-brownbear:
Jenkins run elasticsearch-ci/part-1 (unrelated)

dnhatn (approved):
LGTM.

original-brownbear:
Thanks Nhat!

PR #86538, opened 2022-05-08
Remove Redundant IndexComponent Interface

original-brownbear:
This interface is pointless as it's (with one exception only inherited and used by `AbstractIndexComponent` in production code.


grcevski (approved):
LGTM!

PR #86543, opened 2022-05-09
User Profile - add caching for hasPrivileges check

ywangd:
This PR adds a new cache for hasPrivileges checks so that the same check
for the same role is short-circuited for performance. The new cache is
attached to Role. Since role itself is cached, we avoid having to manage
the new cache lifecyle separately, e.g. exposing new clear cache
endpoint etc.

As discussed, the cache is currently only applied to simple named roles,
i.e. it works for regular users but not API keys. This is acceptable for
now since API keys cannot have user profiles.


n1v0lg (approved):
LGTM

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/SimpleRole.java
```diff
@@ -155,4 +163,38 @@ public int hashCode() {
         return result;
     }
 
+    public static final Setting<Integer> CACHE_SIZE_SETTING = Setting.intSetting(
+        "xpack.security.authz.store.role.has_privileges.cache.max_size",
+        1000,
+        Setting.Property.NodeScope
+    );
```
albertzaharovits, line 170: Nit: I noticed the other `xpack.security.authz.store.*` settings use "roles" (plural) instead of "role". I think it would be more consistent that instead. Also, maybe you could move this declaration to the top of the class.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/SimpleRole.java
```diff
@@ -155,4 +163,38 @@ public int hashCode() {
...
+    public PrivilegesCheckResult checkPrivilegesWithCache(PrivilegesToCheck privilegesToCheck) {
+        final Cache<PrivilegesToCheck, PrivilegesCheckResult> cache = hasPrivilegesCacheReference.get();
+        if (cache == null) {
+            return null;
+        }
+        return cache.get(privilegesToCheck);
+    }
+
+    // package private for testing
+    public Cache<PrivilegesToCheck, PrivilegesCheckResult> getHasPrivilegesCache() {
```
albertzaharovits, line 197: nit: drop public

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/profile/ProfileIntegTests.java
```diff
@@ -594,6 +604,84 @@ public void testSetEnabled() {
         );
     }
 
+    public void testHasPrivileges() {
```
albertzaharovits, line 607: 👍 Nice test!

albertzaharovits (approved):
LGTM

No action to take here and now but I want to mention some ideas for possible improvements in the future:

When/if we document this cache sizing setting we should point out that the size is per user (ie each user has its quota of 1000 cached "has privilege" check results), rather than a global count.
Maybe later we investigate if it is possible to share the "weight" across multiple caches so that the weight adjustment setting is for a global check result count.
In addition, `PrivilegesToCheck` is not great as a caching key... Ideally the member privileges should be a `Set` rather than an array. I think this could be achieved relatively easy, I think I had it like that in some draft version (for this reason as well, the cache sizing parameter is even more arcane).

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/SimpleRole.java
ywangd: As commented, the cache is per `role`. So I intentionally changed `roles` to `role`. But maybe it is more confusing than helpful. I changed it back to be `role` again. Also moved the line to the top.

PR #86424, opened 2022-05-04
[Test] Replace all mock authentication with concrete objects

ywangd:
Changes in this PR enables marking the Authentication class to final in
a future PR as part of the overall plan for closing down Authentication.

Relates: #86206

PR #86544, opened 2022-05-09
The Authentication class is now final

ywangd:
This PR marks the Authentication class to be final which is the last
step of locking down the Authentication class. The Authentication class
is basically a record class and it has internal logics on what values
can or cannot be used together. We don't expect it to be extended by any
subclasses and a concrete object should always be created for tests
instead of mocking.

Relates: #86424
Relates: #86206


albertzaharovits (approved):
LGTM

Well done pushing this through 🎊

ywangd:
@elasticmachine update branch

PR #86547, opened 2022-05-09
simplify immutable map creation

idegtiarenko:
This commit replaces the static block that wraps map with immutable
wrapper with immutable map collector.


original-brownbear (approved):

jbaiera (approved):
LGTM

PR #86548, opened 2022-05-09
Port build dependencies to gradle version catalogue

breskeby:
We introduce the use of a Gradle version catalogue for handling build related dependencies. 

This provides type safe accessors for dependencies and allow centralised version definitions. 
Later we want to move all our dependency handling to version catalogues. 
Since this is a Gradle feature we remove  long term maintenance cost for custom version handling in 
our build and make centralised version handling more straight forward and support better tooling 
based on version catalogues 

Fixes asm version alignment on the way using 9.3 everywhere in our build logic.

### DIFF
--- a/build-tools-internal/settings.gradle
+++ b/build-tools-internal/settings.gradle
@@ -1,6 +1,6 @@
 dependencyResolutionManagement {
     versionCatalogs {
-        create("buildLibs") {
+        buildLibs {
             from(files("../gradle/build.versions.toml"))
         }
     }
--- a/build-tools-internal/settings.gradle
+++ b/build-tools-internal/settings.gradle
@@ -1,0 +1,7 @@
+dependencyResolutionManagement {
+    versionCatalogs {
+        create("buildLibs") {
+            from(files("../gradle/build.versions.toml"))
+        }
+    }
+}
--- a/build-tools/build.gradle
+++ b/build-tools/build.gradle
@@ -122,7 +122,6 @@ dependencies {
     testFixturesApi gradleApi()
     testFixturesApi gradleTestKit()
     testFixturesApi buildLibs.junit
-    testFixturesApi buildLibs.junit
     testFixturesApi buildLibs.randomized.runner
     testFixturesApi buildLibs.wiremock
     testFixturesApi platform(buildLibs.spock.platform)
--- a/build-tools/build.gradle
+++ b/build-tools/build.gradle
@@ -135,14 +135,14 @@ dependencies {
     }
 
     integTestImplementation(platform(buildLibs.junit5.platform))
-    integTestImplementation("org.junit.jupiter:junit-jupiter") {
+    integTestImplementation(buildLibs.junit5.jupiter) {
         because 'allows to write and run Jupiter tests'
     }
-    integTestRuntimeOnly("org.junit.vintage:junit-vintage-engine") {
+    integTestRuntimeOnly(buildLibs.junit5.vintage) {
         because 'allows JUnit 3 and JUnit 4 tests to run'
     }
 
-    integTestRuntimeOnly("org.junit.platform:junit-platform-launcher") {
+    integTestRuntimeOnly(buildLibs.junit5.platform.launcher) {
         because 'allows tests to run from IDEs that bundle older version of launcher'
     }
 }
--- a/build-tools/build.gradle
+++ b/build-tools/build.gradle
@@ -130,7 +130,7 @@ dependencies {
         exclude module: "groovy"
     }
 
-    integTestImplementation(buildLibs.spock.junit) {
+    integTestImplementation(buildLibs.spock.junit4) {
         because 'required as we rely on junit4 rules'
     }
 
--- a/build-tools/build.gradle
+++ b/build-tools/build.gradle
@@ -106,34 +106,35 @@ repositories {
 dependencies {
     constraints {
         // ensuring brought asm version brought in by spock is up-to-date
-        testFixturesApi 'org.ow2.asm:asm:9.3'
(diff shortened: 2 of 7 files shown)


build-tools/build.gradle
```diff
@@ -106,42 +106,43 @@ repositories {
...
+    implementation buildLibs.asm.tree
+    implementation buildLibs.asm
+
     testFixturesApi gradleApi()
     testFixturesApi gradleTestKit()
-    testFixturesApi 'com.github.tomakehurst:wiremock-jre8-standalone:2.23.2'
-    testFixturesApi platform("org.spockframework:spock-bom:2.1-groovy-3.0")
-    testFixturesApi("org.spockframework:spock-core") {
+    testFixturesApi buildLibs.junit
+    testFixturesApi buildLibs.junit
```
pugnascotia, line 125: Duplicate of the previous line?

pugnascotia (approved):

build-tools/build.gradle
breskeby: fixed. good catch!

build-tools-internal/settings.gradle
```diff
@@ -0,0 +1,7 @@
+dependencyResolutionManagement {
+    versionCatalogs {
+        create("buildLibs") {
```
mark-vieira, line 3: Couldn't this just be `buildLibs { }`? Any benefit from explicitly calling `create()` here?

gradle/build.versions.toml
```diff
@@ -0,0 +1,42 @@
...
+hamcrest = "org.hamcrest:hamcrest:2.1"
+httpcore = "org.apache.httpcomponents:httpcore:4.4.12"
+httpclient = "org.apache.httpcomponents:httpclient:4.5.10"
+idea-ext = "gradle.plugin.org.jetbrains.gradle.plugin.idea-ext:gradle-idea-ext:1.1.1"
+json-schema-validator = "com.networknt:json-schema-validator:1.0.49"
+jackson-dataformat-yaml = { group = "com.fasterxml.jackson.dataformat", name="jackson-dataformat-yaml", version.ref="jackson" }
+jackson-platform = { group = "com.fasterxml.jackson", name="jackson-bom", version.ref="jackson" }
+jna = "net.java.dev.jna:jna:5.10.0"
+junit = "junit:junit:4.12"
+junit5-platform = { group = "org.junit", name="junit-bom", version.ref="junit5" }
```
mark-vieira, line 28: If we're using the platform bom shouldn't we just omit version numbers from the dependencies below? Also they don't seem to match. We're using version 5.7.1 of the BOM but 5.8.1 of the individual dependencies.

gradle/build.versions.toml
```diff
@@ -0,0 +1,42 @@
...
+junit5-jupiter = "org.junit.jupiter:junit-jupiter:5.8.1"
+junit5-platform-launcher = "org.junit.platform:junit-platform-launcher:1.8.0"
+junit5-vintage = "org.junit.vintage:junit-vintage-engine:5.8.1"
+maven-model = "org.apache.maven:maven-model:3.6.2"
+mockito-core = "org.mockito:mockito-core:1.9.5"
+nebula-info = "com.netflix.nebula:gradle-info-plugin:11.3.3"
+randomized-runner = "com.carrotsearch.randomizedtesting:randomizedtesting-runner:2.7.7"
```
mark-vieira, line 35: I think we can ditch this altogether. We only reference this class a couple of places for `assumeFalse()` which we can easily replace with `Assumptions` from JUnit.

gradle/build.versions.toml
```diff
@@ -0,0 +1,42 @@
+[versions]
+asm = "9.3"
+jackson = "2.13.2"
+junit5 = "5.7.1"
+spock = "2.1-groovy-3.0"
+
+[libraries]
+ant = "org.apache.ant:ant:1.10.8"
```
mark-vieira, line 8: Is there a reason we can just use what's provided by `gradleApi()`?

mark-vieira (approved):
Couple comments but otherwise LGTM.

gradle/build.versions.toml
breskeby: there's is gap in version catalogue support that does not let allow you to declare a dependency without a version. therefore we I needed to add the version there too. I think that's fine as the interesting bit of platforms is the handling of transitive deps their versions.

Interestingly I took the dependencies that are 5.8.1 as they we're actually used in our build already. turns out the Spock dependency brings in a newer version of the `junit5-bom`. 🤯 I've upgraded version we use to 5.8.1 but referenced the junit5 version used in the other dependency declarations with that junit5 version.

gradle/build.versions.toml
breskeby: there was a time where we needed a newer version of ant than that was shipped with gradle.

breskeby: Ah yes. our UnzipTransforms rely on `org.apache.tools.zip` that is not part the gradle api

gradle/build.versions.toml
breskeby: GradleUnitTestCase and BaseTestCase still rely on plenty of that randomizedtesting stuff and we still have plenty of test classes based on those in build-tools and build-tools-internal

build-tools-internal/settings.gradle
breskeby: true. fixed. looked at the Kotlin DSL example :)

gradle/build.versions.toml
mark-vieira: We should look to ditch those if we can. Not sure how much of a refactoring effort that would be.

PR #86553, opened 2022-05-09
test desired balance service concurrency

idegtiarenko:
This change introduces the test for DesiredBalanceShardsAllocator that verifies:
* all listeners are called
* all listeners are called after corresponding reroute is complete

and fixes the implementation to pass the test


server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
```diff
@@ -333,6 +336,62 @@ public ShardAllocationDecision decideShardAllocation(ShardRouting shard, Routing
         assertThat(listenersCalled.get(), equalTo(1));
     }
 
+    public void testConcurrency() {
```
idegtiarenko, line 339: This is not passing just yet. I will add the fix in the same pr

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
```diff
@@ -333,6 +336,73 @@ public ShardAllocationDecision decideShardAllocation(ShardRouting shard, Routing
...
+
+        var createdIndices = new CopyOnWriteArraySet<String>();
+        var allocatedIndices = new CopyOnWriteArraySet<String>();
+        var reroutedIndices = new CopyOnWriteArraySet<String>();
+
+        var threadPool = new TestThreadPool(getTestName());
+        RerouteService rerouteService = (r, p, l) -> threadPool.executor(ThreadPool.Names.CLUSTER_COORDINATION).submit(() -> {
+            reroutedIndices.addAll(createdIndices);
+            l.onResponse(null);
+        });
```
DaveCTurner, line 349: Think it's slightly stronger to capture the created indices earlier here? Also `ThreadPool.Names.CLUSTER_COORDINATION` seems like the wrong pool (and it only has one thread), this should work with `GENERIC` too:

```
        RerouteService rerouteService = (r, p, l) -> {
            final var localCreatedIndices = Set.copyOf(createdIndices);
            threadPool.executor(ThreadPool.Names.GENERIC).submit(() -> {
                reroutedIndices.addAll(localCreatedIndices);
                l.onResponse(null);
            });
        };
```

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
```diff
@@ -333,6 +336,73 @@ public ShardAllocationDecision decideShardAllocation(ShardRouting shard, Routing
...
+
+        var discoveryNode = createDiscoveryNode();
+        var clusterState = ClusterState.builder(ClusterName.DEFAULT)
+            .nodes(DiscoveryNodes.builder().add(discoveryNode).localNodeId(discoveryNode.getId()).masterNodeId(discoveryNode.getId()))
+            .build();
+
+        var indexNameGenerator = new AtomicInteger();
+        var listenersCalled = new AtomicInteger();
+
+        var iterations = 1_000;
```
DaveCTurner, line 377: Think we can randomise the number of iterations, and also use a `CountDownLatch` to avoid the `assertBusy()` at the end?

```suggestion
        var iterations = between(1, 1000);
        var listenersCountdown = new CountDownLatch(iterations);
```

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
```diff
@@ -333,6 +336,73 @@ public ShardAllocationDecision decideShardAllocation(ShardRouting shard, Routing
...
+                clusterState = ClusterState.builder(clusterState)
+                    .metadata(Metadata.builder(clusterState.metadata()).put(indexMetadata, true))
+                    .routingTable(RoutingTable.builder(clusterState.routingTable()).addAsNew(indexMetadata))
+                    .build();
+                createdIndices.add(indexName);
+            }
+            var indexName = "index-" + indexNameGenerator;
+            desiredBalanceShardsAllocator.allocate(createAllocationFrom(clusterState), ActionListener.wrap(() -> {
+                assertThat("Listener should be called after index is rerouted", reroutedIndices, hasItem(indexName));
+                listenersCalled.incrementAndGet();
```
DaveCTurner, line 392: and here...

```suggestion
                listenersCalled.incrementAndGet();
                listenersCountdown.countDown();
```

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
```diff
@@ -333,6 +336,73 @@ public ShardAllocationDecision decideShardAllocation(ShardRouting shard, Routing
...
+            }));
+        }
+
+        assertBusy(() -> {
+            assertThat(createdIndices.size(), equalTo(indexNameGenerator.get()));
+            assertThat(allocatedIndices.size(), equalTo(indexNameGenerator.get()));
+            assertThat(reroutedIndices.size(), equalTo(indexNameGenerator.get()));
+            assertThat(listenersCalled.get(), equalTo(iterations));
+        });
```
DaveCTurner, line 401: And then this should work right?

```suggestion
        assertTrue(listenersCountdown.await(10, TimeUnit.SECONDS));
        assertThat(createdIndices.size(), equalTo(indexNameGenerator.get()));
        assertThat(allocatedIndices.size(), equalTo(indexNameGenerator.get()));
        assertThat(reroutedIndices.size(), equalTo(indexNameGenerator.get()));
        assertThat(listenersCalled.get(), equalTo(iterations));
```

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
```diff
@@ -333,6 +336,73 @@ public ShardAllocationDecision decideShardAllocation(ShardRouting shard, Routing
...
+        var allocator = new ShardsAllocator() {
+            @Override
+            public void allocate(RoutingAllocation allocation) {// TODO attempt to count up-to-date allocations
+                final var dataNodeId = allocation.nodes().getDataNodes().values().iterator().next().getId();
+                final var unassignedIterator = allocation.routingNodes().unassigned().iterator();
+                while (unassignedIterator.hasNext()) {
+                    String indexName = unassignedIterator.next().getIndexName();
+                    allocatedIndices.add(indexName);
+                    unassignedIterator.initialize(dataNodeId, null, 0L, allocation.changes());
+                }
```
DaveCTurner, line 359: Can we simulate the case where the allocator can't allocate everything in one go here? I think this will allow for more interesting coverage of cases where the input goes stale mid-allocation.

```suggestion
                var madeProgress = false;
                while (unassignedIterator.hasNext()) {
                    final var indexName = unassignedIterator.next().getIndexName();
                    if (randomBoolean() || (madeProgress == false && unassignedIterator.hasNext() == false)) {
                        allocatedIndices.add(indexName);
                        unassignedIterator.initialize(dataNodeId, null, 0L, allocation.changes());
                        madeProgress = true;
                    } else {
                        unassignedIterator.removeAndIgnore(UnassignedInfo.AllocationStatus.NO_ATTEMPT, allocation.changes());
                    }
                }
```

DaveCTurner (commented):
Just looked at the test so far, will get to the prod code in the next round.

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
```diff
@@ -167,23 +154,20 @@ public void afterPrimariesBeforeReplicas(RoutingAllocation allocation) {}
         final var indexMetadata = createIndex(TEST_INDEX);
         var clusterState = createClusterState(discoveryNode, indexMetadata);
 
-        var listener = new TestActionListener();
```
DaveCTurner, line 170: I would prefer to keep the assertions that this listener is called (eventually) although I would not be surprised if we need to move those assertions later.

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
```diff
@@ -231,117 +213,149 @@ public void afterPrimariesBeforeReplicas(RoutingAllocation allocation) {}
         rerouteService.assertNoPendingReroute();
     }
 
-    public void testCallListenersOnlyAfterProducingFreshInput() {
```
DaveCTurner, line 234: This commit removes `testCallListenersOnlyAfterProducingFreshInput` and `testFailListenersOnNoLongerMasterException` (replacing them with `testConcurrency`). I think the removed tests are still valid and useful aren't they? Could we bring them back?

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
```diff
@@ -231,117 +213,149 @@ public void afterPrimariesBeforeReplicas(RoutingAllocation allocation) {}
...
 
-        var discoveryNode = createDiscoveryNode();
-        var index = createIndex(UUIDs.randomBase64UUID());
+        rerouteServiceSupplier.set((r, p, l) -> {
+            clusterService.submitUnbatchedStateUpdateTask("test-desired-balance-reroute", new ClusterStateUpdateTask() {
+                @Override
+                public ClusterState execute(ClusterState currentState) {
+                    logger.info("Performing reroute [{}]", r);
+                    reroutedIndices.addAll(createdIndices);
+                    return allocationService.reroute(currentState, "test", ActionListener.noop());
```
DaveCTurner, line 285: Should this complete the listener `l`?

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
```diff
@@ -231,117 +213,149 @@ public void afterPrimariesBeforeReplicas(RoutingAllocation allocation) {}
...
+        }, desiredBalanceShardsAllocator, () -> ClusterInfo.EMPTY, () -> SnapshotShardSizeInfo.EMPTY);
 
-        var discoveryNode = createDiscoveryNode();
-        var index = createIndex(UUIDs.randomBase64UUID());
+        rerouteServiceSupplier.set((r, p, l) -> {
+            clusterService.submitUnbatchedStateUpdateTask("test-desired-balance-reroute", new ClusterStateUpdateTask() {
+                @Override
+                public ClusterState execute(ClusterState currentState) {
+                    logger.info("Performing reroute [{}]", r);
+                    reroutedIndices.addAll(createdIndices);
```
DaveCTurner, line 284: Is `createdIndices` equal to the the list of indices in `currentState`? If so can we assert that?

DaveCTurner (commented):
Test looks good, I left a few small requests.

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocatorTests.java
idegtiarenko: I am now considering replacing this with a proper BatchedRerouteService..

DaveCTurner (approved):
LGTM

PR #86554, opened 2022-05-09
Switch to OpenJDK and upgrade to 18.0.1

pugnascotia:
Adoptium releases are very slow to release across multiple platforms.
Switch back to Oracle's OpenJDK releases and upgrade.

Also tweak our version parsing and URL construction, since Oracle's
URLs have changed again.


ChrisHegarty (approved):

mark-vieira (approved):
I'm good with pushing forward with this. I have no reason to believe the Adoptium situation is going to improve in the near term.

PR #86562, opened 2022-05-09
Port xpack qa projects away from elasticsearch rest test gradle plugin

breskeby:
This deprecates the `elasticsearch.rest-test` plugin and `elasticsearch.standalone-rest-test` and ports 
all usages of them in `x-pack/plugins`. Other usages will be removed in a few upcoming PRs to not have one >300file PR

When all usages have been addressed we're going to remove those gradle plugins from the codebase.

This PR is a subset of https://github.com/elastic/elasticsearch/pull/85491 which got just too big to handle IMO


mark-vieira (approved):
LGTM. Since this is build related, we would ideally want to backport this to 7.17 to make our lives easier down the road. Given the nature of these changes, that's likely to be a nasty backport but I think it's worth doing to reduce pain in the future.

breskeby:
👍

breskeby:
Manual backport at https://github.com/elastic/elasticsearch/pull/86592

Issue #86496, opened 2022-05-05
[CI] AssignmentPlannerTests testTooManyNodesAndModels_DoesNotThrowOOM_GivenNodesJustUnderLimit failing


slobodanadamovic:
**Build scan:**
https://gradle-enterprise.elastic.co/s/jjsb3opg5n7iy/tests/:x-pack:plugin:ml:test/org.elasticsearch.xpack.ml.inference.assignment.planning.AssignmentPlannerTests/testTooManyNodesAndModels_DoesNotThrowOOM_GivenNodesJustUnderLimit

**Reproduction line:**
`./gradlew ':x-pack:plugin:ml:test' --tests "org.elasticsearch.xpack.ml.inference.assignment.planning.AssignmentPlannerTests.testTooManyNodesAndModels_DoesNotThrowOOM_GivenNodesJustUnderLimit" -Dtests.seed=110A9DC5977D6165 -Dbuild.snapshot=false -Dtests.jvm.argline="-Dbuild.snapshot=false" -Dtests.locale=es-MX -Dtests.timezone=GB -Druntime.java=17`

**Applicable branches:**
master

**Reproduces locally?:**
Yes

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.xpack.ml.inference.assignment.planning.AssignmentPlannerTests&tests.test=testTooManyNodesAndModels_DoesNotThrowOOM_GivenNodesJustUnderLimit

**Failure excerpt:**
```
java.lang.OutOfMemoryError: Java heap space

  at __randomizedtesting.SeedInfo.seed([110A9DC5977D6165:1B9089CC55E53785]:0)
  at org.ojalgo.array.Primitive64Array.<init>(Primitive64Array.java:112)
  at org.ojalgo.matrix.store.Primitive64Store.<init>(Primitive64Store.java:413)
  at org.ojalgo.matrix.store.Primitive64Store$1.make(Primitive64Store.java:173)
  at org.ojalgo.matrix.store.Primitive64Store$1.make(Primitive64Store.java:70)
  at org.ojalgo.structure.Factory2D.make(Factory2D.java:115)
  at org.ojalgo.optimisation.linear.SimplexTableau$DenseTableau.<init>(SimplexTableau.java:61)
  at org.ojalgo.optimisation.linear.SimplexTableau.make(SimplexTableau.java:764)
[...]

PR #86566, opened 2022-05-09
[ML] Upgrade ojalgo to 51.2.0

dimitris-athanasiou:
Ojalgo 51.2.0 brings significant improvements to memory utilization
and speed for the linear solver we use to determine assignment plans
for trained models.

In addition to updating the library, we now make use of the ability
to use the solver in sparse mode which uses much less memory at the cost
of speed. Thus we determine an intermediate threshold where we switch
from dense to sparse solver.

Testing showed a significant decrease in memory use. In particular,
the worse cases regardless of dense/sparse solving should now not be using
more than ~40MB which is much less than what we allowed previously.

The dense solver is faster than before. Even when switching to the sparse solver,
it is not much slower than what the dense solver achieved previously.

Relates #86496


przemekwitek (approved):
LGTM

PR #86571, opened 2022-05-09
[ML] Express lane inference 

davidkyle:
Not all inference requests are equal, requests for searches should have a higher priority than those generated by ingest. 

3 types of request are expected each with a different priority:
- Normal inference requests executed at normal priority
- Higher priority inference requests (e.g. for search) that skip ahead of the queue of normal requests
- Control messages that change the number of allocations/threads. These are highest priority and are always executed even if the number of items in the queue has reached the capacity value.

An unbounded priority queue is used to queue the requests which are sorted first by priority then my the requests id (long) to preserve order. 

A `skipQueue` option is added to the internal inference request, this is not exposed to the REST API or indeed used anywhere as it will be used by the model allocation service.

### DIFF
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/pytorch/PriorityProcessWorkerExecutorService.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/pytorch/PriorityProcessWorkerExecutorService.java
@@ -83,10 +83,14 @@ public class PriorityProcessWorkerExecutorService extends AbstractProcessWorkerE
             return;
         }
 
-        if (queue.size() == queueCapacity
-            || queue.add(new OrderedRunnable(priority, tieBreaker, contextHolder.preserveContext(command))) == false) {
+        // highest priority requests are always accepted even if the queue is full
+        if (queue.size() >= queueCapacity && priority != RequestPriority.HIGHEST) {
             command.onRejection(new EsRejectedExecutionException(processName + " queue is full. Unable to execute command", false));
+            return;
         }
+
+        // PriorityBlockingQueue::offer always returns true
+        queue.offer(new OrderedRunnable(priority, tieBreaker, contextHolder.preserveContext(command)));
     }
 
     @Override
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/pytorch/PriorityProcessWorkerExecutorService.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/pytorch/PriorityProcessWorkerExecutorService.java
@@ -17,9 +17,8 @@ import java.util.concurrent.PriorityBlockingQueue;
 
 /**
  * An {@link AbstractProcessWorkerExecutorService} where the runnables
- * are have a priority and are executed in priority order with a tie breaker
- * value for requests of equal priority. The tie breaker can be used to
- * preserve insertion order.
+ * are executed in priority order with a tie breaker value for requests
+ * of equal priority. The tie breaker can be used to preserve insertion order.
  *
  * Order is maintained by a PriorityQueue
  */
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/pytorch/PriorityProcessWorkerExecutorService.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/pytorch/PriorityProcessWorkerExecutorService.java
@@ -76,13 +76,14 @@ public class PriorityProcessWorkerExecutorService extends AbstractProcessWorkerE
      * @param priority Request priority
      * @param tieBreaker For sorting requests of equal priority
      */
-    public void executeWithPriority(AbstractRunnable command, RequestPriority priority, long tieBreaker) {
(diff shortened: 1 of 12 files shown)


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/pytorch/PriorityProcessWorkerExecutorService.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Insert a runnable into the executor queue.
+     *
+     * @param command The runnable
+     * @param priority Request priority
+     * @param tieBreaker For sorting requests of equal priority
+     */
+    public void executeWithPriority(AbstractRunnable command, RequestPriority priority, long tieBreaker) {
```
benwtrent, line 79: This needs to be synchronized.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/pytorch/PriorityProcessWorkerExecutorService.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * @param contextHolder the thread context holder
+     * @param processName the name of the process to be used in logging
+     * @param queueCapacity the capacity of the queue holding operations. If an operation is added
+     *                  for execution when the queue is full a 429 error is thrown.
+     */
+    @SuppressForbidden(reason = "properly rethrowing errors, see EsExecutors.rethrowErrors")
+    public PriorityProcessWorkerExecutorService(ThreadContext contextHolder, String processName, int queueCapacity) {
+        super(contextHolder, processName, queueCapacity, PriorityBlockingQueue::new);
```
benwtrent, line 68: The other option here is to create a queue that wraps three separate queues (normal, high, highest). It would take a "prioritized" object and feed it to the appropriate queue. Poll, etc. would then pull from the appropriate queue and it would keep track of its size as a sum of the inner queues. 

I am not sure if that is any cleaner/faster or nicer than simply using a priority queue like you got here. But it would prevent having to reshuffle the order on offer every time.

benwtrent: @davidkyle I dont think you have to sit there and poll like that. I would poll from all three, if its null, wait on a lock condition that is signaled when something is added to your queue wrapper. Let me see if I can write up a quick example...

```java
static class MyQueues<E> implements BlockingQueue<E> {

        private final Queue<E> queueNormal;
        private final Queue<E> queueHigh;
        private final Queue<E> queueHighest;
        private final int sizeLimit;

        // lock for queueing
        private final ReentrantLock lock = new ReentrantLock();

        // For waiting to not be empty
        private final Condition notEmpty = lock.newCondition();

        MyQueues(int sizeLimit) {
            queueNormal = new ArrayDeque<>();
            queueHigh = new ArrayDeque<>();
            queueHighest = new ArrayDeque<>();
            this.sizeLimit = sizeLimit;
        }

        public boolean offerWithPriority(E item, RequestPriority priority) {
            lock.lock();
            try {
                if (queueNormal.size() + queueHigh.size() + queueHighest.size() > sizeLimit) {
                    return false;
                }
                switch (priority) {
                    case HIGH -> queueHigh.add(item);
                    case NORMAL -> queueNormal.add(item);
                    case HIGHEST -> queueHighest.add(item);
                }
                notEmpty.signal();
                return true;
            } finally {
                lock.unlock();
            }
        }

        @Override
        public E poll(long timeout, TimeUnit unit) throws InterruptedException {
            long nanos = unit.toNanos(timeout);
            lock.lockInterruptibly();
            try {
                E item = null;
                while ((item = pollUnsafe()) == null && nanos > 0) {
                    nanos = notEmpty.awaitNanos(nanos);
                }
                return item;
            } finally {
                lock.unlock();
            }
        }

        @Override
        public E poll() {
            lock.lock();
            try {
                return pollUnsafe();
            } finally {
                lock.unlock();
            }
        }

        private E pollUnsafe() {
            return Optional.ofNullable(queueHighest.poll())
                .or(() -> Optional.ofNullable(queueHigh.poll()))
                .or(() -> Optional.ofNullable(queueNormal.poll()))
                .orElse(null);
        }
// <snip> all the other methods for blocking queue </snip>
    }
```

davidkyle: ++ looks good I think that would work.

I paused before implementing this- as usual the best thing to do is run a micro benchmark first. 

The concern is that offers into the priority queue will be slow. For the common use case we expect requests to have normal priority and be naturally ordered by the request Id which does not cause a reshuffling of the priority queue. 

For this case I compared offering 1000 items in their natural order into a `LinkedBlockingQueue` and `PriorityBlockingQueue`. Surprisingly the priority queue was faster 17807.750 ns vs 26797.762 ns on average.

Inserting 1000 items into the priority queue where every 100th item has a higher priority was still faster then inserting into the linked queue at 20959.939 ns (average)

benwtrent (approved):

davidkyle:
The benchmark mentioned above did not take into account the extra cost of polling from a priority queue. Once both `offer`ing and `poll`ing  items was benchmarked the `LinkedBlockingQueue` was found to be ~ twice as fast as `PriorityBlockingQueue`

PR #86574, opened 2022-05-09
Reroute after migrating to data tiers routing

### LINKED ISSUE #86572: Migrate to data tiers routing API doesn't reroute
### Elasticsearch Version

7.17, 8.x

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

Darwin

### Problem Description

The migrate to data tiers routing API changes the allocation configuration for 
the indices that need migrating to data tiers routing. 
After such changes we should do a cluster re-route so the allocation changes
take effect immediately. 

### Steps to Reproduce

1. On a one node `data_hot,data_content,master` node create the following index
```
PUT datatiers_problem_index
{
    "settings": {
        "index.routing.allocation.include._tier_preference": "data_hot",
        "index.routing.allocation.require.data": "hot"
    }
}
```
2. Migrate to data tiers and observe the shard is not allocated afterwards
`POST _ilm/migrate_to_data_tiers`

### Logs (if relevant)

_No response_

andreidan:
Fixes #86572


x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/action/TransportMigrateToDataTiersAction.java
```diff
@@ -138,6 +136,8 @@ public void onFailure(Exception e) {
             @Override
             public void clusterStateProcessed(ClusterState oldState, ClusterState newState) {
                 super.clusterStateProcessed(oldState, newState);
```
DaveCTurner, line 138: I think this does nothing, may as well remove it.

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/action/TransportMigrateToDataTiersAction.java
```diff
@@ -138,6 +136,8 @@ public void onFailure(Exception e) {
...
+                clusterService.getRerouteService().reroute("cluster migrated to data tiers routing", Priority.NORMAL,
+                    ActionListener.noop());
```
DaveCTurner, line 140: Does it matter that the listener completes without waiting for the reroute? I expect there to be some tests that need a reroute to have happened - if they're passing today then are they waiting artificially on something else, and if so can the wait be removed?

DaveCTurner (commented):
I would expect some test changes to go with this, maybe removing some explicit reroutes or other waiting?

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/action/TransportMigrateToDataTiersAction.java
andreidan: Thanks for bringing this up. 

IMO the API got the job done and shouldn't necessarily wait for a successful reroute - I guess we can have a listener that logs the error? 

Good point about the tests - we tested the effect of the API (ie. some settings got removed, some migrated) - I'll add one that waits for a green cluster after migration. Thanks for the suggestion.

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/action/TransportMigrateToDataTiersAction.java
```diff
@@ -137,7 +139,16 @@ public void onFailure(Exception e) {
...
             public void clusterStateProcessed(ClusterState oldState, ClusterState newState) {
-                super.clusterStateProcessed(oldState, newState);
+                clusterService.getRerouteService()
+                    .reroute("cluster migrated to data tiers routing", Priority.NORMAL, new ActionListener<ClusterState>() {
+                        @Override
+                        public void onResponse(ClusterState clusterState) {}
+
+                        @Override
+                        public void onFailure(Exception e) {
+                            logger.warn("unsuccessful reroute after migration to data tiers routing", e);
```
DaveCTurner, line 149: Nit: if `MasterService#isPublishFailureException(e)` this message should just be at `DEBUG`, no need to make more noise on a master failure than necessary.

DaveCTurner (approved):
LGTM, one nit

andreidan:
@elasticmachine update branch

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/action/TransportMigrateToDataTiersAction.java
andreidan: TIL

PR #85298, opened 2022-03-24
added the edge case in the related document

ghost:
This is a PR for issue #128194

- Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
- Have you followed the [contributor guidelines](https://github.com/elastic/elasticsearch/blob/master/CONTRIBUTING.md)?
- If submitting code, have you built your formula locally prior to submission with `gradle check`?
- If submitting code, is your pull request against master? Unless there is a good reason otherwise, we prefer pull requests against master and will backport as needed.
- If submitting code, have you checked that your submission is for an [OS and architecture that we support](https://www.elastic.co/support/matrix#show_os)?
- If you are submitting this code for a class then read our [policy](https://github.com/elastic/elasticsearch/blob/master/CONTRIBUTING.md#contributing-as-part-of-a-class) for that.

PR #86575, opened 2022-05-09
[DOCS] Document when security auto configuration is skipped

### LINKED ISSUE #83306: Document the cases where security auto-configuration is not triggered
In 8.0.0 , with Security on by default, we introduce security auto-configuration when a node starts up for the first time. There are certain cases though, where we either can't proceed with auto-configuration or we determine/assume that the user is intending to configure security on their own and we proceed with starting the node without performing the security auto-configuration steps. We should document the cases where this happens and link it in appropriate places in our Getting started docs so that folks can easily find them / we can point them to these.

##### Environment related
- When the data dir exists and is not empty. This is a **strong** indication that this is not the first time that the node starts and we don't want to run autoconfiguration in this case. 
- When `elasticsearch.yml` doesn't exist, or we can't read it. If we can't read it, then we can't know if elasticsearch security features are already configured or not. This is also an indication that we;re running as a different user ( or a not privileged enough user ) and we can't handle this case. 
- When we can't read `elasticsearch.keystore` for the same reasons as above
- When the configuration directory is not writeable. This means that it is made read only by the admin, or the user starting is not the user that installed elasticsearch. 
- When indices are read-only ( ie, because of low disk space availability, https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-cluster.html#disk-based-shard-allocation ) because we can't write the `elastic` user password and the API key to the security index

##### Configuration related
- When `xpack.security.autoconfiguration.enabled` is set to false. This defaults to true and it's purpose is for users to be able to explicitly disable auto-configuration
- When `xpack.security.enabled` is set ( independent of its value ). We assume that the user wants to configure security features themselves.
- When any `xpack.security.transport.ssl.*` or `xpack.security.http.ssl.*` settings are set. We assume that the users want to configure TLS themselves
- When any of the discovery ( `discovery.type`, `discovery.seed_hosts` ) or initial master nodes (`cluster.initial_master_nodes` ) settings are set. There are two exceptions to this: 
  - When `discovery.type: single-node`
  - When `cluster.initial_master_nodes` exists but contains only the name of the current node. 
- If the current node cannot become a master node ( i.e. node.roles is configured with other values ).
- If the current node cannot have a copy of the security index ( i.e. only has role that can't contain data )  

#### Cases where initial `elastic` password generation doesn't happen

- All the above cases 
- When `bootstrap.password` is set in the `elasticsearch.keystore`
- When `autoconfiguration.password_hash` is set in the `elasticsearch.keystore` ( this is an implementation detail of the installation process for DEB/RPM ) 

#### Cases where an enrollment token for nodes is not generated ( default behavior for anything else than docker ) 

- When elasticsearch is only bound to localhost.

lockewritesdocs:
Adds documentation for cases where security auto-configuration is skipped.

* Adds a [new section](https://elasticsearch_86575.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/configuring-stack-security.html#stack-skip-auto-configuration) documenting when auto configuration is skipped
* Adds a note in the [overview](https://elasticsearch_86575.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/configuring-stack-security.html#configuring-stack-security) that links to the new section
* Adds a note indicating when [interactive mode](https://elasticsearch_86575.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/configuring-stack-security.html#stack-start-with-security) won't start for Kibana (cc: @jportner to review)
* Adds a description for the `xpack.security.autoconfiguration.enabled` parameter in the [secure settings](https://elasticsearch_86575.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/security-settings.html#general-security-settings) page

Closes #83306 
Closes #85298
Closes https://github.com/elastic/kibana/issues/128194

### DIFF
--- a/x-pack/docs/en/security/configuring-stack-security.asciidoc
+++ b/x-pack/docs/en/security/configuring-stack-security.asciidoc
@@ -132,9 +132,9 @@ Certificate Authority (CA) for the HTTP layer.
 When you start {es} for the first time, we try to automatically configure
 security for you. We run some checks to determine whether this is the first time
 that the node is starting. If any of those checks fail, there's a good
-indication that you already configured security, or don't want security to be
-configured automatically. In these cases, we'll respect your existing
-configuration and skip the auto-configuration process. 
+indication that you <<<<manually-configure-security,manually configured security>>,
+or don't want security to be configured automatically. In these cases, we'll
+respect your existing configuration and skip the auto-configuration process. 
 
 [discrete]
 [[stack-existing-environment-detected]]
@@ -151,12 +151,13 @@ The existence of this directory is a strong indicator that the node was started
 previously, and we don't want to overwrite your existing configuration.
 
 The `elasticsearch.yml` file doesn't exist (or isn't readable), or the `elasticsearch.keystore` isn't readable::
-If either of these file aren't readable, we can't determine whether {es} security 
+If either of these files aren't readable, we can't determine whether {es} security 
 features are already enabled. This state also indicates that we're probably not 
 running as a user who has sufficient privileges.
 
 The {es} configuration directory isn't writable::
-This state indicates that an administrator made this directory read-only, or that the user who is starting {es} is not the user that installed {es}.
+This state likely indicates that an administrator made this directory read-only,
+or that the user who is starting {es} is not the user that installed {es}.
 
 Indices are read-only::
 It's likely that there's not enough disk space, so we can't write the `elastic`
@@ -165,9 +166,8 @@ user password and API key to the `.security` index.
 [discrete]
 [[stack-existing-settings-detected]]
 ==== Existing settings detected
-If any of the following settings have a specific value, or have a value
-configured, then we'll assume that you want to configure security on your own
-and we won't configure security automatically.
+If {es} detects any of the following settings, then we'll assume that you want
+to configure security on your own and we won't configure security automatically.
 
 * {ref}/modules-node.html#node-roles[`node.roles`] is set to a value other than 
 `master`
--- a/x-pack/docs/en/security/configuring-stack-security.asciidoc
+++ b/x-pack/docs/en/security/configuring-stack-security.asciidoc
@@ -16,8 +16,9 @@ cluster, authenticates to {es} with the built-in `kibana` service account, and w
 security configuration to `kibana.yml`.
 
 NOTE: There are <<stack-skip-auto-configuration,some cases>> where security
-can't be configured automatically because an existing environment or existing
-settings are detected when starting {es}.
+can't be configured automatically because the node startup process detects that
+the node is already part of a cluster, or that security is already configured or
+explicitly disabled.
 
 [discrete]
 === Prerequisites
@@ -129,12 +130,18 @@ Certificate Authority (CA) for the HTTP layer.
 [discrete]
 [[stack-skip-auto-configuration]]
 === Cases when security auto configuration is skipped
-When you start {es} for the first time, we try to automatically configure
-security for you. We run some checks to determine whether this is the first time
-that the node is starting. If any of those checks fail, there's a good
-indication that you <<<<manually-configure-security,manually configured security>>,
-or don't want security to be configured automatically. In these cases, we'll
-respect your existing configuration and skip the auto-configuration process. 
+When you start {es} for the first time, the node startup process tries to
+automatically configure security for you. The process runs some checks to
+determine: 
+
+* If this is the first time that the node is starting
+* Whether security is already configured
+* If the startup process can modify the node configuration
+
+If any of those checks fail, there's a good indication that you
+<<manually-configure-security,manually configured security>>, or don't want
+security to be configured automatically. In these cases, the node starts
+normally using the existing configuration. 
 
 [discrete]
 [[stack-existing-environment-detected]]
@@ -148,29 +155,27 @@ automatically.
 
 The {es} `/data` directory exists and isn't empty::
 The existence of this directory is a strong indicator that the node was started 
-previously, and we don't want to overwrite your existing configuration.
+previously, and might already be part of a cluster.
 
 The `elasticsearch.yml` file doesn't exist (or isn't readable), or the `elasticsearch.keystore` isn't readable::
 If either of these files aren't readable, we can't determine whether {es} security 
-features are already enabled. This state also indicates that we're probably not 
-running as a user who has sufficient privileges.
+features are already enabled. This state can also indicate that the node startup
+process isn't running as a user with sufficient privileges to modify the
+node configuration.
 
 The {es} configuration directory isn't writable::
 This state likely indicates that an administrator made this directory read-only,
 or that the user who is starting {es} is not the user that installed {es}.
 
-Indices are read-only::
-It's likely that there's not enough disk space, so we can't write the `elastic`
-user password and API key to the `.security` index.
-
 [discrete]
 [[stack-existing-settings-detected]]
 ==== Existing settings detected
-If {es} detects any of the following settings, then we'll assume that you want
-to configure security on your own and we won't configure security automatically.
+The following settings are incompatible with security auto configuration. If any
+of these settings exist, the node startup process skips configuring security
+automatically and the node starts normally.
 
-* {ref}/modules-node.html#node-roles[`node.roles`] is set to a value other than 
-`master`
+* {ref}/modules-node.html#node-roles[`node.roles`] is set to a value where the
+node can't be elected as `master`, or if the node can't hold data
 * {ref}/security-settings.html#general-security-settings[`xpack.security.autoconfiguration.enabled`] is set to `false`
 * {ref}/security-settings.html#general-security-settings[`xpack.security.enabled`] has a value set
 * Any of the
--- a/x-pack/docs/en/security/configuring-stack-security.asciidoc
+++ b/x-pack/docs/en/security/configuring-stack-security.asciidoc
@@ -181,7 +181,8 @@ node can't be elected as `master`, or if the node can't hold data
 * Any of the
 {ref}/security-settings.html#transport-tls-ssl-settings[`xpack.security.transport.ssl.*`] or 
 {ref}/security-settings.html#http-tls-ssl-settings[`xpack.security.http.ssl.*`]
-settings have a value set
+settings have a value set in the `elasticsearch.yml` configuration file or in
+the `elasticsearch.keystore`
 * Any of the `discovery.type`, `discovery.seed_hosts`, or
 `cluster.initial_master_nodes`
 {ref}/modules-discovery-settings.html[discovery and cluster formation settings] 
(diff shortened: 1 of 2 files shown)


x-pack/docs/en/security/configuring-stack-security.asciidoc
```diff
@@ -15,6 +15,10 @@ minutes. This token automatically applies the security settings from your {es}
 cluster, authenticates to {es} with the built-in `kibana` service account, and writes the
 security configuration to `kibana.yml`.
 
+NOTE: There are <<stack-skip-auto-configuration,some cases>> where security
+can't be configured automatically because an existing environment or existing
+settings are detected when starting {es}.
```
albertzaharovits, line 20:
```suggest
will not be configured automatically because it appears that the node is already part of a cluster, Security (or just TLS) is already configured or explicitly disabled.
```  

It's only a suggestion feel free to adapt it, as you see fit.

x-pack/docs/en/security/configuring-stack-security.asciidoc
```diff
@@ -112,3 +125,78 @@ include::{es-ref-dir}/setup/install/security-files-reference.asciidoc[leveloffse
 Additionally, when you use the enrollment token to connect {kib} to a secured {es} cluster, the HTTP layer CA certificate is retrieved from {es} and stored in the
 {kib} `/data` directory. This file establishes trust between {kib} and the {es}
 Certificate Authority (CA) for the HTTP layer.
+
+[discrete]
+[[stack-skip-auto-configuration]]
+=== Cases when security auto configuration is skipped
+When you start {es} for the first time, we try to automatically configure
+security for you. We run some checks to determine whether this is the first time
+that the node is starting. If any of those checks fail, there's a good
```
albertzaharovits, line 134:
```suggest
that the node is starting, if Security is not otherwise configured, and if the node configuration can be modified by the node startup process . If any of those checks fail
```
I'm not too happy with the above proposal, though. It's too verbose. I'll leave it up to you to tailor it.

x-pack/docs/en/security/configuring-stack-security.asciidoc
```diff
@@ -112,3 +125,78 @@ include::{es-ref-dir}/setup/install/security-files-reference.asciidoc[leveloffse
...
+indication that you <<<<manually-configure-security,manually configured security>>,
+or don't want security to be configured automatically. In these cases, we'll
+respect your existing configuration and skip the auto-configuration process.
```
albertzaharovits, line 137: "we'll respect your existing configuration" -> the node starts up normaly using the existing configuration

x-pack/docs/en/security/configuring-stack-security.asciidoc
```diff
@@ -112,3 +125,78 @@ include::{es-ref-dir}/setup/install/security-files-reference.asciidoc[leveloffse
...
+If certain directories already exist, there's a strong indication that the node
+was started previously. Similarly, if certain files _don't_ exist, or we can't 
+read or write to specific files or directories, then we're likely not running as 
+the user who installed {es} or an administrator imposed restrictions. If any of 
+the following environment checks are true, security isn't configured 
+automatically.
+
+The {es} `/data` directory exists and isn't empty::
+The existence of this directory is a strong indicator that the node was started 
+previously, and we don't want to overwrite your existing configuration.
```
albertzaharovits, line 151: "and we don't want to overwrite your existing configuration." -> "and might already be part of a cluster"

x-pack/docs/en/security/configuring-stack-security.asciidoc
```diff
@@ -112,3 +125,78 @@ include::{es-ref-dir}/setup/install/security-files-reference.asciidoc[leveloffse
...
+
+The `elasticsearch.yml` file doesn't exist (or isn't readable), or the `elasticsearch.keystore` isn't readable::
+If either of these files aren't readable, we can't determine whether {es} security 
+features are already enabled. This state also indicates that we're probably not 
+running as a user who has sufficient privileges.
```
albertzaharovits, line 156: "we're probably not running as a user who has suggicient privileges" -> the node process is not running as an user that has access to change the configuration.

x-pack/docs/en/security/configuring-stack-security.asciidoc
```diff
@@ -112,3 +125,78 @@ include::{es-ref-dir}/setup/install/security-files-reference.asciidoc[leveloffse
...
+
+The {es} configuration directory isn't writable::
+This state likely indicates that an administrator made this directory read-only,
+or that the user who is starting {es} is not the user that installed {es}.
+
+Indices are read-only::
+It's likely that there's not enough disk space, so we can't write the `elastic`
+user password and API key to the `.security` index.
```
albertzaharovits, line 164: I would suggest we drop this item.
Security might already be auto-configured by the time the code gets to the point to encounter this condition. If this condition is encountered there is a message shown on the console with the instructions to follow for remediation.

x-pack/docs/en/security/configuring-stack-security.asciidoc
```diff
@@ -112,3 +125,78 @@ include::{es-ref-dir}/setup/install/security-files-reference.asciidoc[leveloffse
...
+
+[discrete]
+[[stack-existing-settings-detected]]
+==== Existing settings detected
+If {es} detects any of the following settings, then we'll assume that you want
+to configure security on your own and we won't configure security automatically.
+
+* {ref}/modules-node.html#node-roles[`node.roles`] is set to a value other than 
+`master`
```
albertzaharovits, line 173: is set such that the node cannot be elected master or if it cannot hold data

x-pack/docs/en/security/configuring-stack-security.asciidoc
```diff
@@ -112,3 +125,78 @@ include::{es-ref-dir}/setup/install/security-files-reference.asciidoc[leveloffse
...
+
+Indices are read-only::
+It's likely that there's not enough disk space, so we can't write the `elastic`
+user password and API key to the `.security` index.
+
+[discrete]
+[[stack-existing-settings-detected]]
+==== Existing settings detected
+If {es} detects any of the following settings, then we'll assume that you want
+to configure security on your own and we won't configure security automatically.
```
albertzaharovits, line 170: The following settings are incompatible with the security auto-configuration. If any are present security auto-configuratin is skipped and the node starts normally.

x-pack/docs/en/security/configuring-stack-security.asciidoc
```diff
@@ -112,3 +125,78 @@ include::{es-ref-dir}/setup/install/security-files-reference.asciidoc[leveloffse
...
+====
+Exceptions are when `discovery.type` is set to `single-node`, or when
+`cluster.initial_master_nodes` exists but contains only the name of the current
+node.
+====
+--
+
+[discrete]
+[[stack-existing-keystore-detected]]
+==== Keystore settings detected
```
albertzaharovits, line 196: TLDR I would drop this entire section. I think it is too much of an implementation detail.

> we can't generate a password for the `elastic` user

That's not quire true, that is not the reason.
If there are TLS settings in the keystore we don't overwrite them and the node will fail to start.

albertzaharovits (commented):
I would like to have another pass at it, after you've considered my feedback.

Sincerely appreciate you tackling this. This is not an easy task! Thank you.

x-pack/docs/en/security/configuring-stack-security.asciidoc
lockewritesdocs: I'll change this to read:

>...because the node startup process detects that the node is already part of a cluster, or that security is already configured or explicitly disabled.

x-pack/docs/en/security/configuring-stack-security.asciidoc
lockewritesdocs: I'm going to turn this sentence into a bulleted list, such as:

> When you start {es} for the first time, the node startup process tries to automatically configure security for you. The process runs some checks to determine: 
> 
> * If this is the first time that the node is starting
> * Whether security is already configured
> * If the startup process can modify the node configuration

x-pack/docs/en/security/configuring-stack-security.asciidoc
```diff
@@ -112,3 +125,78 @@ include::{es-ref-dir}/setup/install/security-files-reference.asciidoc[leveloffse
...
 {kib} `/data` directory. This file establishes trust between {kib} and the {es}
 Certificate Authority (CA) for the HTTP layer.
+
+[discrete]
+[[stack-skip-auto-configuration]]
+=== Cases when security auto configuration is skipped
+When you start {es} for the first time, we try to automatically configure
+security for you. We run some checks to determine whether this is the first time
+that the node is starting. If any of those checks fail, there's a good
+indication that you <<<<manually-configure-security,manually configured security>>,
```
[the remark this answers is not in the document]
lockewritesdocs: Good catch! 🎣

x-pack/docs/en/security/configuring-stack-security.asciidoc
lockewritesdocs: I'll change this to read:

>This state also indicates that the node startup process isn't running as a user with sufficient privileges to modify the node configuration.

x-pack/docs/en/security/configuring-stack-security.asciidoc
lockewritesdocs: Works for me -- I'll remove this section.

x-pack/docs/en/security/configuring-stack-security.asciidoc
lockewritesdocs: I'll change to read:

>The following settings are incompatible with security auto configuration. If any of these settings exist, the node startup process skips configuring security automatically and the node starts normally.

x-pack/docs/en/security/configuring-stack-security.asciidoc
lockewritesdocs: I'll change to read:

>...is set to a value where the node can't be elected as `master`, or if the node can't hold data

x-pack/docs/en/security/configuring-stack-security.asciidoc
lockewritesdocs: Is it worth changing this section to say:

>If the node startup process detects existing TLS settings in the Elasticsearch keystore, the process halts and the node fails to start.

Do users need to know even that? Or is it still too low-level to matter?

lockewritesdocs:
Thanks for your input @albertzaharovits! I addressed your comments, so sending this back to you for review.

x-pack/docs/en/security/configuring-stack-security.asciidoc
albertzaharovits: > If the node startup process detects existing TLS settings in the Elasticsearch keystore, the process halts and the node fails to start.

That's a good way to phrase it, I think.

The underlying reason to skip the auto-configuration in this case is similar to skipping it when the regular .yml TLS ones are encountered, ie. "it appears that the user already has configured Security to some extent".
An alternative suggestion would be to squeeze the keystore mention  in the point above 
> `xpack.security.transport.ssl.*` or`xpack.security.http.ssl.*`
settings have a value set (in the `elasticsearch.yml` or the `elasticsearch.keystore` files).

albertzaharovits (approved):
LGTM

I think this is very well phrased overall. Thank you, Adam!

albertzaharovits:
NB: I commented on https://github.com/elastic/elasticsearch/pull/86575#discussion_r873985929 , it needs to be sorted out, but I don't think it needs my involvement anymore, hence approving.

lockewritesdocs:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.2|[](https://github.com/elastic/elasticsearch/pull/86839)|
|✅|8.1|[](https://github.com/elastic/elasticsearch/pull/86840)|
|✅|8.0|[](https://github.com/elastic/elasticsearch/pull/86841)|

### Questions ?

PR #86577, opened 2022-05-09
Use desired nodes during data tier allocation decisions

fcofdez:
Today when the `DataTierAllocationDecider` checks if a shard can be allocated
in a node or not it uses the current cluster topology. This works fine in most cases
when the data flows from hotter to colder data nodes as it was designed. But 
in certain situations users might want to remove a tier (i.e. the cold tier) that
allows having shards in hotter tiers, but with the current implementation once
a shard is allocated into the highest preference tier and there are nodes of that
tier, the shard won't move. This commit introduces a change that allows to use
the desired nodes (when available) to compute the preferred tier based on the
cluster topology provided by the desired nodes (only taking into account desired
members that are already members of the cluster).


fcofdez:
Moving this back to draft as @henningandersen as identified a case we should take into account.

server/src/main/java/org/elasticsearch/cluster/metadata/InternalDesiredNodesMembershipService.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    public static InternalDesiredNodesMembershipService create(ClusterService clusterService) {
+        var tracker = new InternalDesiredNodesMembershipService();
+        clusterService.addListener(tracker);
+        return tracker;
+    }
+
+    @Override
+    public synchronized void clusterChanged(ClusterChangedEvent event) {
```
dakrone, line 34: My understanding is that `clusterChanged` events are along the same lines as cluster state operations, in that they are serialized to a single thread. That would make the `synchronized` keyword unnecessary here. Am I correct in my understanding?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -375,7 +376,11 @@ private IndexMetadata indexMetadata(ShardRouting shard, RoutingAllocation alloca
             return allocation.metadata().getIndexSafe(shard.index());
         }
 
-        private Optional<String> highestPreferenceTier(List<String> preferredTiers, DiscoveryNodes unused) {
+        private Optional<String> highestPreferenceTier(
+            List<String> preferredTiers,
+            DiscoveryNodes unused,
+            DesiredNodes.ClusterMembers desiredNodes
```
dakrone, line 382: This parameter is never used, is that intentional?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierAllocationDecider.java
```diff
@@ -135,6 +138,29 @@ private static Decision debugYesAllowed(RoutingAllocation allocation, List<Strin
      * exist. If no nodes for any of the tiers are available, returns an empty
      * {@code Optional<String>}.
      */
+    public static Optional<String> preferredAvailableTier(
```
dakrone, line 141: Can you update the javadoc for this method please?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierAllocationDecider.java
```diff
@@ -135,6 +138,29 @@ private static Decision debugYesAllowed(RoutingAllocation allocation, List<Strin
...
+        final var desiredNodesPreferredTier = preferredAvailableTier(prioritizedTiers, desiredNodesMembers);
+
+        if (desiredNodesPreferredTier.isPresent()) {
+            return desiredNodesPreferredTier;
+        }
+
+        return preferredAvailableTier(prioritizedTiers, nodes);
+    }
+
+    public static Optional<String> preferredAvailableTier(List<String> prioritizedTiers, DesiredNodes.ClusterMembers nodes) {
```
dakrone, line 155: Can you add javadocs for this method please?

dakrone (commented):
I left a couple of small comments, but the data tier portion looks good to me otherwise.

One kind of larger meta comment: as this is the first time I've started reading through the desired nodes work, **please please please** I plead with you to add javadoc to these (and other desired node) classes! None of the desired nodes classes I could find had any javadoc, and I imagine myself thumbing through this code in response to a support request sometime in the future and think about how helpful it would be to have that documentation.

fcofdez:
> One kind of larger meta comment: as this is the first time I've started reading through the desired nodes work, please please please I plead with you to add javadoc to these (and other desired node) classes! None of the desired nodes classes I could find had any javadoc, and I imagine myself thumbing through this code in response to a support request sometime in the future and think about how helpful it would be to have that documentation.

Thanks for the feedback @dakrone, I'll open a separate PR documenting all the desired nodes related classes 👍

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
fcofdez: Yes, we use this method as a parameter in:
```
                    node -> DataTierAllocationDecider.shouldFilter(
                        indexMetadata,
                        node.node().getRoles(),
                        this::highestPreferenceTier,
                        allocation
                    )
```

fcofdez:
Moving this to ready for review cc @henningandersen

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierAllocationDecider.java
```diff
@@ -133,9 +136,51 @@ private static Decision debugYesAllowed(RoutingAllocation allocation, List<Strin
...
+
+        return getPreferredAvailableTierFromClusterMembers(prioritizedTiers, nodes);
+    }
+
+    /**
+     * Given a list of tiers in descending order, return the tier that's present
+     * in the desired nodes with the highest priority, if none is present returns an
+     * {@code Optional.empty()}.
+     */
+    public static Optional<String> getPreferredTierFromDesiredNodes(
```
henningandersen, line 162: I wonder if we should wait for all seeing all expected nodes in higher prioritized tiers before we consider a tier removed? I worry that a user removes warm tier and upsizes hot tier in one orchestration change. And we immediately start pushing too much data to hot tier, making it run out of space before we see the new node(s)? Might not be super realistic (takes time to move data) but feels like a problem we should consider tackling.

server/src/main/java/org/elasticsearch/cluster/metadata/InternalDesiredNodesMembershipService.java
```diff
@@ -0,0 +1,78 @@
...
+                }
+            }
+        } else if (event.changedCustomMetadataSet().contains(DesiredNodesMetadata.TYPE) || latestHistoryId == null) {
+            if (desiredNodes.historyID().equals(latestHistoryId) == false) {
+                members.clear();
+            }
+            latestHistoryId = desiredNodes.historyID();
+
+            final Set<DesiredNode> removedDesiredNodes = new HashSet<>(members);
+            desiredNodes.nodes().forEach(removedDesiredNodes::remove);
```
henningandersen, line 55: I think we may want to refine this to work on just the external id? The idea being that a node that was a member in the prior desired nodes is also a member with the new desired nodes even if it is temporarily out of the cluster while applying this. We can leave that for later I think.

henningandersen (commented):
Left some more comments to consider here.

server/src/main/java/org/elasticsearch/cluster/metadata/InternalDesiredNodesMembershipService.java
```diff
@@ -0,0 +1,78 @@
...
+
+            members.removeAll(removedDesiredNodes);
+        }
+    }
+
+    @Override
+    public synchronized DesiredNodes.MembershipInformation getMembershipInformation() {
+        if (members.isEmpty()) {
+            return DesiredNodes.MembershipInformation.EMPTY;
+        }
```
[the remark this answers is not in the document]
fcofdez: That was a mistake, I was trying to avoid allocating in that case, it wasn't a good idea. The scenario that you mention makes total sense.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
```diff
@@ -439,6 +440,7 @@ private static RoutingAllocation createRoutingAllocation(ClusterState state, All
             state,
             createClusterInfo(state),
             null,
+            DesiredNodes.MembershipInformation.EMPTY,
```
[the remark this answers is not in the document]
fcofdez: I think we should rely on the desired nodes membership information here. `DataTierAllocationDecider` needs to get a valid `DesiredNodes.MembershipInformation` instance, unless we want to add a check in the decider and just fallback to the `DiscoveryNodes` when `DesiredNodes.MembershipInformation` is null, wdyt?

x-pack/plugin/core/src/internalClusterTest/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierAllocationDeciderIT.java
```diff
@@ -67,6 +84,108 @@ public void testDefaultIndexAllocateToContent() {
...
+
+        client().admin()
+            .indices()
+            .prepareUpdateSettings(index)
+            .setSettings(
+                Settings.builder()
+                    .put("index.routing.allocation.exclude._name", String.join(",", masterNode.getName(), coldNodeName))
+                    .build()
+            )
+            .get();
```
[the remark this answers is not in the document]
fcofdez: That's a good idea. I think it shouldn't be too costly, right?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -80,7 +83,7 @@ public RoutingAllocation(
         SnapshotShardSizeInfo shardSizeInfo,
         long currentNanoTime
     ) {
-        this(deciders, null, clusterState, clusterInfo, shardSizeInfo, currentNanoTime);
+        this(deciders, null, clusterState, clusterInfo, shardSizeInfo, DesiredNodes.MembershipInformation.EMPTY, currentNanoTime);
```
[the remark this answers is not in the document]
fcofdez: I totally missed these use cases 🤦. I'll add a few more tests to at least cover the explain allocation.

server/src/main/java/org/elasticsearch/cluster/metadata/InternalDesiredNodesMembershipService.java
fcofdez: I used this approach for simplicity, but it's true that if we end up relying on this more it can become costly. I will move the `members` map to a `volatile` field to avoid such synchronization.

x-pack/plugin/core/src/internalClusterTest/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierAllocationDeciderIT.java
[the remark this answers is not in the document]
henningandersen: Correct. We would expect many more reroutes throughout an orchestrated change of the cluster, adding one here should not matter.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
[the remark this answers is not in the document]
henningandersen: I see. Let us leave this as is then.

fcofdez:
@henningandersen as we agreed I've moved the desired nodes membership status to the cluster state, I think this is ready for a first round of reviews. Let me know what you think about this approach.

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNodes.java
```diff
@@ -180,4 +230,128 @@ private static Map<String, DesiredNode> toMap(final List<DesiredNode> desiredNod
...
+        final var desiredNodes = DesiredNodes.latestFromClusterState(clusterState);
+        return desiredNodes == null
+            || clusterState.nodes()
+                .stream()
+                .map(node -> desiredNodes.find(node.getExternalId()))
+                .filter(Objects::nonNull)
+                .allMatch(DesiredNode::isMember);
+    }
+
+    public static class MembershipInformation {
```
henningandersen, line 312: This class looks unused.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierAllocationDecider.java
```diff
@@ -133,9 +137,69 @@ private static Decision debugYesAllowed(RoutingAllocation allocation, List<Strin
...
+        DiscoveryNodes discoveryNodes,
+        DesiredNodes desiredNodes
+    ) {
+        if (desiredNodes == null) {
+            return Optional.empty();
+        }
+
+        for (int i = 0; i < prioritizedTiers.size(); i++) {
+            final var tier = prioritizedTiers.get(i);
+            final var nextTier = i + 1 == prioritizedTiers.size() ? null : prioritizedTiers.get(i + 1);
```
henningandersen, line 170: I think we should loop through all possible lower tiers? Suppose we have cold,warm,hot preference but warm is unused. Then we would want to do the check here against hot too?

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNode.java
```diff
@@ -95,6 +107,26 @@ private static Version parseVersion(String version) {
         return Version.fromString(version);
     }
 
+    public enum MembershipStatus {
+        UNKNOWN((short) 0),
```
henningandersen, line 111: I wonder if UNKNOWN should be "NEW" instead? It is a new node that we have not seen yet.

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNodes.java
```diff
@@ -161,6 +200,17 @@ public List<DesiredNode> nodes() {
         return List.copyOf(nodes.values());
     }
 
+    public List<DesiredNode> members() {
+        assert members.stream().allMatch(DesiredNode::isMember);
+        return members;
+    }
+
+    public List<DesiredNode> notMembers() {
```
henningandersen, line 208: Like for the enum, I wonder if `newDesiredNodes` is better? I am not truly happy about that either though.

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNodes.java
```diff
@@ -180,4 +230,128 @@ private static Map<String, DesiredNode> toMap(final List<DesiredNode> desiredNod
...
+        }
+
+        if (modified) {
+            return new DesiredNodes(historyID, version, updatedStateDesiredNodes);
+        } else {
+            return this;
+        }
+    }
+
+    public static boolean knownDesiredNodesAreCorrect(ClusterState clusterState) {
```
henningandersen, line 302: I am not entirely sure I correctly understand `known` in this method name.

henningandersen (commented):
Thanks Francisco, I gave this an initial read and left a few detailed comments. I have 2 higher level comments:

1. Terminology: I am not sure of the membership terminology. Also, sometimes it is called "known" in the code. I think we need to find a good terminology here. Incubating, seen, known, unknown, member are some vague suggestions, perhaps you can work on a list of suggestions?
2. Tracking the state inside the `DesiredNode` object has benefits but also makes the code that updates desired nodes quite complex. I wonder if you considered tracking this outside the `DesiredNode` instead? Would take slightly more space but I would not consider that a concern.

henningandersen:
We discussed this offline and decided to use "status=pending" for nodes that we have not seen yet and "status=actualized" (or materialized) for nodes that we have seen at least once.

The status will move to a wrapper `DesiredNodeWithStatus`, with two fields `status` and `desiredNode`.

fcofdez:
Closing this in favor of https://github.com/elastic/elasticsearch/pull/87735

PR #70625, opened 2021-03-22
Total data set size in stats

henningandersen:
With shared cache searchable snapshots we have shards that have a size
in S3 that differs from the locally occupied disk space. This commit
introduces `store.total_data_set_size` to node and indices stats, allowing to
differ between the two.

Relates #69820

PR #86580, opened 2022-05-09
Correctly calculate disk usage for frozen data tier telemetry

### LINKED ISSUE #86055: Data tiers usage is 0 for partially mounted indices
### Elasticsearch Version

8.1.2

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

ESS on Azure

### Problem Description

The `_xpack/usage` API reports the size of the data set per data tier. However, it uses the local disk size as the metric it reports on and this is 0 (zero) for partially mounted indices, hiding the size of the data set mounted on the frozen tier.

We should switch to using the `total_data_set_size` metric that was introduced when we added the frozen tier/partially mounted indices.

### Steps to Reproduce

Mount a partially mounted index on a dedicated frozen tier. Use the `_xpack/usage` API and see that no (or very little) data is reported on frozen tier.

### Logs (if relevant)

_No response_

dakrone:
The telemetry for data tiers was using the size in bytes, however, for the frozen tier using
searchable snapshots, this was the disk usage rather than the size of the actual data.

This commit changes the telemetry to use `total_data_set_size` as introduced in #70625 so that the
telemetry is correct.

Resolves #86055


dakrone:

gmarouli (approved):
LGTM :)

dakrone:
@elasticmachine update branch

PR #86584, opened 2022-05-09
Removing extra spaces from message

masseyke:
Some messages in the ShardsAvailabilityHealthIndicatorService had an extra space. This commit removes them.


dakrone (approved):
:+1: assuming CI is happy

PR #86589, opened 2022-05-10
Hide ImmutableOpenMap in IndexMetadata

rjernst:
IndexMetadata has several ImmutableOpenMaps in its implementation.
However, the getters for those members expose ImmutableOpenMap when Map
works just fine. This commit changes the return types of these methods
from IndexMetadata, so that the IndexMetadata implementation can be
changed independently.

relates #86239


martijnvg (approved):
👍 LGTM

PR #86590, opened 2022-05-10
[ML] Autoscaling fixes and additional debug logging

droberts195:
Fixes the following problems in autoscaling:

1. Divide by zero error when scaling up from zero
2. Given JVM size rounding, we'd sometimes choose a node that
   was 1-3MB bigger than necessary, potentially leading to an
   unnecessarily large scale-up
3. The per-node overhead was being added to required size twice
4. Improve the reason when a cluster with no ML nodes decides
   it still needs no ML nodes
5. Now consider free memory at current scale when scaling up

Also adds more debug logging so we can get visibility of
what happens in deployed Cloud clusters when necessary.

Followup to #86399


droberts195:
Jenkins run `elasticsearch-ci/part-2`

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java
```diff
@@ -367,9 +367,9 @@ public AutoscalingDeciderResult scale(Settings configuration, AutoscalingDecider
         if (isMaster == false) {
```
dimitris-athanasiou, line 367: Not relevant to this PR but this is one scary long method. Should we raise an issue on looking to refactor this? Or is it just not possible to do much better here due to the nature of what it's doing?

dimitris-athanasiou (approved):
LGTM

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java
```diff
@@ -449,7 +449,7 @@ public AutoscalingDeciderResult scale(Settings configuration, AutoscalingDecider
                     reasonBuilder.setSimpleReason(
                         String.format(
                             Locale.ROOT,
-                            "Passing currently perceived capacity as down scale delay has not been satisfied; configured delay [%s]"
+                            "Passing currently perceived capacity as down scale delay has not been satisfied; configured delay [%s] "
```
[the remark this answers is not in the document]
droberts195: The rest of the sentence gets appended on the line below and that doesn't start with a space.  So we need one somewhere.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java
droberts195: It's probably possible to reduce the size.

I still need to iterate on this PR first though.  There's at least one more bug, maybe two, and these need fixing before 8.3 feature freeze.

dimitris-athanasiou: Certainly, I wasn't suggesting to do anything as part of this PR.

x-pack/plugin/ml/src/test/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderServiceTests.java
```diff
@@ -665,14 +668,20 @@ public void testScaleUp_withWaitingJobsAndRoomInNodes() {
                 analytics,
                 List.of(),
                 null,
-                NativeMemoryCapacity.ZERO,
+                currentScale,
```
droberts195, line 671: Several of the tests in this file weren't setting `currentScale` appropriately for the test scenario, and were then asserting results that didn't make sense given how things work in production. I have tried to make all the `testScaleUp_blah` tests in this file call `service.checkForScaleUp()` with coherent arguments so that we will be exercising that method in the same way as production in the future.

droberts195:
This is `>non-issue` because the major fixes (items 1-3) are logically part of #86399.

PR #86592, opened 2022-05-10
[7.17] Port xpack plugins projects away from elasticsearch rest test gradle plugin (#86562)

breskeby:
This deprecates the elasticsearch.rest-test plugin and elasticsearch.standalone-rest-test and ports
all usages of them in x-pack/plugins. Other usages will be removed in a few upcoming PRs to not have one >300file PR

When all usages have been addressed we're going to remove those gradle plugins from the codebase.

This PR is a subset of https://github.com/elastic/elasticsearch/pull/85491 which got just too big to handle IMO


breskeby:
@elasticmachine test this

Issue #81210, opened 2021-12-01
Snapshots as simple archives


ywelsch:
### Goal

The goal of this effort is to provide access to older Elasticsearch data, for compliance or regulatory reasons, the occasional lookback or investigation, or to rehydrate parts of it. Access to the data is expected to be very infrequent, and can therefore happen with limited performance and query capabilities. Running old versions of Elasticsearch to access the old data is not practical as it would require running outdated and unsupported software.

A non-goal of this effort is to fully solve the major version upgrade problem. *"Snapshots as simple archives"* is an important first step however towards longer term data retention and access. It will allow some users to refrain from upgrading their archived data, and refraining from upgrading is probably the simplest upgrade option. 

### Solution

Snapshots have long been used for backup purposes. With this new feature, they can be used for archival purposes as well now. Elasticsearch will have the ability to access older snapshot repositories and the data therein. In addition, some basic query and aggregation capabilities are available, and it allows reindexing the data into newer Elasticsearch clusters without having the old cluster present. It provides the guarantee that the data put into Elasticsearch (and stored in snapshots) does not have an EOL, but can be accessed for a long time into the future (even if at reduced speed). The data can either be restored with read-only access, or the data can be accessed via searchable snapshots so that the archived data won't even need to fully reside on local disks for access.

#### Phase 0: Prototype

- [x] Basic implementation showing feasibility (#77542)

#### Phase 1: MVP (target release: 8.3)

Allow Elasticsearch 8 nodes to access snapshot repositories written by previous Elasticsearch versions going back to Elasticsearch 5.0. Allow restoring indices from snapshots in the old repository into the Elasticsearch 8 cluster as well as mounting them as searchable snapshots. Allow basic query and aggregation capabilities based on postings / doc values as well as runtime fields on these indices.

##### Supported field types

Old mappings are imported as much "as-is" as possible into Elasticsearch 8, but only provide regular query / aggregation capabilities on a select subset of fields:

- Numeric types
- `boolean` type
- `ip` type
- `geo_point` type
- `date` types: the date `format` setting on date fields is supported in so far as it behaves similarly across these versions. In case it is not, this field can be updated on legacy indices so that it can be changed by a user if need be.
- `keyword` type: the `normalizer` setting on keyword fields is supported in so far as it behaves similarly across these versions. In case it is not, this field can be updated on legacy indices if need be.
[...]

PR #86594, opened 2022-05-10
Disable get API on legacy indices

ywelsch:
The get API relies under the hood on accessing postings to lookup the _id and retrieve the corresponding document. Guaranteeing this access via postings is not something we would like to guarantee on archive indices. While we are adding "[text field support](https://github.com/elastic/elasticsearch/pull/86591)" for archive indices, we reserve the flexibility to eventually swap that out with a "runtime-text field" variant, and hence only provide those capabilities that can be emulated via a runtime field. Doing the same for "get" would mean doing a full scan of the index (using stored fields), which is counterintuitive to what the get API is meant to be used for (quick lookup of document). We would therefore rather not have the API accessible on archive indices.

Relates #81210

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/shard/IndexShard.java
+++ b/server/src/main/java/org/elasticsearch/index/shard/IndexShard.java
@@ -1185,6 +1185,9 @@ public class IndexShard extends AbstractIndexShardComponent implements IndicesCl
         if (mappingLookup.hasMappings() == false) {
             return GetResult.NOT_EXISTS;
         }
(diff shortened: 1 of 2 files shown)


romseygeek:
Can you add something to the description about why we want to do this? The code change looks correct to me but I don't understand the motivation for it yet :)

ywelsch:
> Can you add something to the description about why we want to do this? The code change looks correct to me but I don't understand the motivation for it yet :)

Of course! I've updated the PR description.

romseygeek (approved):
LGTM! Thanks @ywelsch

ywelsch:
Thanks @romseygeek!

PR #86597, opened 2022-05-10
[ML] Rename threading params in _start trained model deployment API

dimitris-athanasiou:
When starting a trained model deployment the user can tweak performance
by setting the `model_threads` and `inference_threads` parameters.
These parameters are hard to understand and cause confusion.

This commit renames these as well as the fields where their values are
reported in the stats API.

- `model_threads` => `number_of_allocations`
- `inference_threads` => `threads_per_allocation`

Now the terminology is as follows.

A model deployment starts with a requested `number_of_allocations`.
Each allocation means the model gets another thread for executing
parallel inference requests. Thus, more allocations should increase
throughput. In its turn, each allocation is may be using a number
of threads to parallelize each individual inference request.
This is the `threads_per_allocation` setting and increases inference
speed (which might also result in improved throughput).

### DIFF
--- a/docs/reference/ml/trained-models/apis/get-trained-models-stats.asciidoc
+++ b/docs/reference/ml/trained-models/apis/get-trained-models-stats.asciidoc
@@ -120,17 +120,9 @@ The sum of `error_count` for all nodes in the deployment.
 (integer)
 The sum of `inference_count` for all nodes in the deployment.
 
-`inference_threads`:::
-(integer)
-The number of threads used by the inference process.
-
 `model_id`:::
 (string)
 include::{es-repo-dir}/ml/ml-shared.asciidoc[tag=model-id]
-
-`model_threads`:::
-(integer)
-The number of threads used when sending inference requests to the model.
 
 `nodes`:::
 (array of objects)
@@ -157,21 +149,9 @@ The number of errors when evaluating the trained model.
 (integer)
 The total number of inference calls made against this node for this model.
 
-`inference_threads`:::
-(integer)
-The number of threads used by the inference process.
-This value is limited by the number of hardware threads on the node;
-it might therefore differ from the `inference_threads` value in the <<start-trained-model-deployment>> API.
-
 `last_access`:::
 (long)
 The epoch time stamp of the last inference call for the model on this node.
-
-`model_threads`:::
-(integer)
-The number of threads used when sending inference requests to the model.
-This value is limited by the number of hardware threads on the node;
-it might therefore differ from the `model_threads` value in the <<start-trained-model-deployment>> API.
 
 `node`:::
 (object)
@@ -199,6 +179,12 @@ include::{es-repo-dir}/ml/ml-shared.asciidoc[tag=node-id]
 (string)
 include::{es-repo-dir}/ml/ml-shared.asciidoc[tag=node-transport-address]
 ========
+
+`number_of_allocations`:::
+(integer)
+The number of allocations assigned to this node.
+This value is limited by the number of hardware threads on the node;
+it might therefore differ from the `number_of_allocations` value in the <<start-trained-model-deployment>> API.
 
 `number_of_pending_requests`:::
 (integer)
@@ -240,6 +226,12 @@ queue was full.
 (long)
 The epoch timestamp when the allocation started.
 
+`threads_per_allocation`:::
+(integer)
+The number of threads by each allocation during inference.
+This value is limited by the number of hardware threads on the node;
+it might therefore differ from the `threads_per_allocation` value in the <<start-trained-model-deployment>> API.
+
 `timeout_count`:::
 (integer)
 The number of inference requests that timed out before being processed.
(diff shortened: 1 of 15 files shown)


docs/reference/ml/trained-models/apis/get-trained-models-stats.asciidoc
```diff
@@ -240,6 +226,12 @@ queue was full.
 (long)
 The epoch timestamp when the allocation started.
 
+`threads_per_allocation`:::
+(integer)
+The number of threads by each allocation during inference.
```
davidkyle, line 231:
```suggestion
The number of threads for each allocation during inference.
```

docs/reference/ml/trained-models/apis/start-trained-model-deployment.asciidoc
```diff
@@ -68,6 +60,15 @@ Every machine learning node in the cluster where the model can be allocated
 has a queue of this size; when the number of requests exceeds the total value,
 new requests are rejected with a 429 error. Defaults to 1024.
 
+`threads_per_allocation`::
+(Optional, integer)
+Sets the number of threads used by each model allocation during during inference. This generally increases
```
davidkyle, line 65:
```suggestion
Sets the number of threads used by each model allocation during inference. This generally increases
```

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/action/StartTrainedModelDeploymentAction.java
```diff
@@ -254,22 +254,28 @@ public static boolean mayAssignToNode(DiscoveryNode node) {
 
         public static final Version VERSION_INTRODUCED = Version.V_8_0_0;
         private static final ParseField MODEL_BYTES = new ParseField("model_bytes");
-        public static final ParseField MODEL_THREADS = new ParseField("model_threads");
-        public static final ParseField INFERENCE_THREADS = new ParseField("inference_threads");
+        public static final ParseField NUMBER_OF_ALLOCATIONS = new ParseField("number_of_allocations");
```
davidkyle, line 257: Why can't a `ParseField` with a deprecated name be used here as you have done with the request object

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/action/StartTrainedModelDeploymentAction.java
```diff
@@ -279,24 +285,42 @@ public static TaskParams fromXContent(XContentParser parser) {
...
+            Integer threadsPerAllocation,
+            int queueCapacity,
+            Integer legacyModelThreads,
+            Integer legacyInferenceThreads
+        ) {
+            this(
+                modelId,
+                modelBytes,
+                threadsPerAllocation == null ? legacyInferenceThreads : threadsPerAllocation,
+                numberOfAllocations == null ? numberOfAllocations : legacyModelThreads,
```
davidkyle:
```suggestion
                numberOfAllocations == null ? legacyModelThreads : numberOfAllocations,
```

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/action/StartTrainedModelDeploymentAction.java
dimitris-athanasiou: `TaskParams` is stored in cluster state as part of `TrainedModelAssignment`. If we have a running deployment we want to parse the old names successfully without deprecation warnings. That is why we're trying to read them separately with both the legacy and the new names.

benwtrent: I think this is good. The user has no control of "un-deprecating" the previously stored information outside of stopping/starting. I think just allowing it internally like this is fine.

benwtrent (approved):

docs/reference/ml/trained-models/apis/start-trained-model-deployment.asciidoc
```diff
@@ -38,26 +38,18 @@ include::{es-repo-dir}/ml/ml-shared.asciidoc[tag=model-id]
...
-greater than the number of available hardware threads on the machine does not increase the
-inference speed. If this setting is greater than the number of hardware threads
-it will automatically be changed to a value less than the number of hardware threads.
-Defaults to 1.
-
-`model_threads`::
-(Optional, integer)
-The number of threads used when sending inference requests to the model.
+The number of allocations a model gets on each node it is assigned to.
+For each allocation there is a separate thread used for sending inference requests to the model.
```
davidkyle, line 44:
```suggestion
The number of model allocations on each node where the model is deployed.
All allocations on a node share the same copy of the model in memory but use
a separate set of threads to evaluate the model. 
```

I think this ^ can be clarified.

davidkyle (approved):
LGTM

Left a comment on the docs that might need another iteration

PR #86602, opened 2022-05-10
Add a Gradle Plugin for compiling modular source projects.

ChrisHegarty:
This PR upstreams a custom Gradle Plugin for compiling modular source
projects, that we've been using for a while now over on the modules
branch. The Plugin infers which dependencies should be on the module
path and which dependencies should be on the class path, when compilng
the main java sourceset. This supports the "bottom up" modularization of
Elasticsearch, where parent projects are modularized before child
projects. Allowing to modularize the core of the server independently of
the extension components, that can themselves be modularized later.

With this plugin various subprojects can be modularized (by adding a
module-info.java in the source root) with minimal fuss - no custom
gradle logic is required by the subproject, beyond a couple of
subproject specific compiler lint suppressions.

The inference in the plugin is driven by walking the project
dependencies, searching for the presence of a module-info.java,
afterwhich all subprojects will be put on the module path. Again, this 
supports "bottom up" modularization.


build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaModulePathPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+
+        var moduleCompileClasspath = configurations.create("moduleCompileClasspath", it -> {
+            it.extendsFrom(compileClasspath);
+            it.setCanBeConsumed(false); // We don't want this configuration used by dependent projects
+            it.attributes(
+                attrs -> attrs.attribute(
+                    LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE,
+                    project.getObjects().named(LibraryElements.class, LibraryElements.CLASSES)
+                )
+            );
```
breskeby, line 78: Why was this exactly necessary? The java / java library plugin configures this too. Why do we need to set this explicitly to classes and not using jars? It shouldn't really matter should it?

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaModulePathPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+                FileCollection trimmedClasspath = classpath.minus(moduleCompileClasspath);
+                task.setClasspath(project.files(trimmedClasspath));
+            }
+            task.doLast(new Action<Task>() {
+                @Override
+                public void execute(Task task2) {
+                    // System.out.println("%s, Module path args: %s".formatted(project, argsToString(argumentProvider.asArguments())));
+                    // System.out.println("%s, Classpath: %s".formatted(project, pathToString(task.getClasspath().getAsPath())));
+                }
+            });
```
breskeby, line 105: Can be removed I guess

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaModulePathPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        var isModuleProject = hasModuleInfoDotJava(project);
+        var configurations = project.getConfigurations();
+        var compileClasspath = configurations.getByName(JavaPlugin.COMPILE_CLASSPATH_CONFIGURATION_NAME);
+
+        var moduleCompileClasspath = configurations.create("moduleCompileClasspath", it -> {
+            it.extendsFrom(compileClasspath);
+            it.setCanBeConsumed(false); // We don't want this configuration used by dependent projects
+            it.attributes(
```
breskeby, line 73: Lets add an explicit `it.setCanBeResolved(true);` to make the purpose of this configuration explicit

breskeby (approved):
I've added a few remarks but in general this looks good to me.

Regarding test coverage we might want to have some basic coverage for dealing with third party plugins (modules, non modules). I can help you with that if you want.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaModulePathPlugin.java
ChrisHegarty: I found this to be useful on numerous occasions, and having it in comments allowed me to quickly enable it.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaModulePathPlugin.java
mark-vieira: Generally just to avoid building the jar in most cases. Using the classes dir is just more efficient.

breskeby (changes requested):
Some more remarks. One blocker (the check for isIdea)

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaModulePathPlugin.java
rjernst: Or change this to use the task logger at info level?

ChrisHegarty: Great idea. Done.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaModulePathPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+        }).collect(Collectors.joining("\n  ", "[\n  ", "]"));
+    }
+
+    static String pathToString(String path) {
+        return Arrays.stream(path.split(File.pathSeparator)).sorted().collect(Collectors.joining("\n  ", "[\n  ", "]"));
+    }
+
+    static boolean isIdea() {
+        return System.getProperty("idea.active", "false").equals("true");
+    }
```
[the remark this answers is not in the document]
ChrisHegarty: Unfortunately `idea.sync.active:true` is not sufficient - it results in a per module dependency of `Empty Library - Provided, which in turn leads to compilation failures.  I agree that "having to do this" is weird, but similar has been seen elsewhere too, e.g. Lucene.  (it's a current limitation of the IDEA modelling of modular projects)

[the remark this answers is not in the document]
ChrisHegarty: Here's a link so similar in Lucene - https://issues.apache.org/jira/browse/LUCENE-10304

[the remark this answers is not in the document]
ChrisHegarty: @breskeby and I went through the various workflows with `idea.sync.active`, and it works perfectly. Updated.  ( I did try this yesterday, but likely fat fingered the property name or something )

build-tools-internal/src/integTest/groovy/org/elasticsearch/gradle/internal/ElasticsearchJavaModulePathPluginFuncTest.groovy
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    private def assertModulePathClasspath(List<String> expectedEntries, String output) {
+        def allArgs = output.find(/(?<=COMPILE_JAVA_COMPILER_ARGS ).*\n/).trim()
+        if(allArgs.isEmpty()) {
+            assert expectedEntries.size() == 0
+        } else {
+            def modulePathEntries = OperatingSystem.current.isWindows() ?
+                allArgs.find(/(?<=.*--module-path=)[^;]*(?=;)?/) :
+                allArgs.find(/(?<=.*--module-path=).*/).minus(";--module-version=${ES_VERSION}")
```
breskeby, line 167: sorry I think this needs to be reverted. meaning

```
 modulePathEntries = OperatingSystem.current.isWindows() ?
                 allArgs.find(/(?<=.*--module-path=).*/).minus(";--module-version=${ES_VERSION}") :
                 allArgs.find(/(?<=.*--module-path=)[^;]*(?=;)?/)
```

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaModulePathPlugin.java
breskeby: FYI: true on linux, on windows things might differ as the bloated input calculation is more expensive on windows.

breskeby (approved):
lgtm. one remark regarding the workaround for our tests on windows

build-tools-internal/src/integTest/groovy/org/elasticsearch/gradle/internal/ElasticsearchJavaModulePathPluginFuncTest.groovy
ChrisHegarty: D'oh! fixing

ChrisHegarty:
@elasticmachine update branch

rjernst (commented):
One minor question

build-tools-internal/src/integTest/groovy/org/elasticsearch/gradle/internal/ElasticsearchJavaModulePathPluginFuncTest.groovy
```diff
@@ -0,0 +1,203 @@
...
+
+    public static final GString JAVA_BASE_MODULE = "java.base:${System.getProperty("java.version")}"
+
+    public static final String ES_VERSION = VersionProperties.getElasticsearch()
+
+    def setup() {
+        javaMainClass()
+        subProject("some-lib") << """
+            plugins {
+                id 'java-library'
```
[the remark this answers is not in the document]
ChrisHegarty: The java-library plugin adds the `api` configuration that the test uses.

PR #86604, opened 2022-05-10
Role APIs allow internal role names

### LINKED ISSUE #86480: Internal role names are (also) not special
### Description

Internal users have internal role names associated with them (e.g., `_system` and `_xpack`). Currently, these role names are "reserved" in that end-users can't create roles with clashing names. We should remove this restriction. The reasoning is similar to #86326:

For one, internal role names are not directly used during authorization. Authorization for internal users follows a separate path, leveraging the role descriptor that each internal user defines directly (e.g. see [SecurityProfileUser](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/user/SecurityProfileUser.java#L22-L36)). As such, internal roles and end-user defined roles do not interfere with each other.

By removing the restriction, we further decouple the implementation details of internal users from our end-user APIs, which allows us to evolve these details without introducing breaking changes.

To guard against the role name of an internal user resolving to an end-user defined-role erroneously, we should modify the internal user classes to only use the role name in the role descriptor, but exclude it from role names associated with the user instance.

n1v0lg:
Internal users have internal role names associated with them (e.g.,
`_system` and `_xpack`). Currently, these role names are "reserved" in
that end-users can't create roles with clashing names. This PR removes
this restriction. The reasoning is similar to #86326:

For one, internal role names are not directly used during
authorization. Authorization for internal users follows a separate
path, leveraging the role descriptor that each internal user defines
directly. As such, internal roles and end-user defined roles do not
interfere with each other.

By removing the restriction, the PR decouples the implementation
details of internal users from our end-user APIs, which allows us to
evolve these details without introducing breaking changes.

Closes #86480


x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/AuthorizationServiceTests.java
```diff
@@ -479,6 +480,79 @@ public void testActionsForSystemUserIsAuthorized() {
         verifyNoMoreInteractions(auditTrail);
     }
 
+    public void testActionsForUserMatchingSystemUserRoleNameDenied() {
+        final Authentication authentication = createAuthentication(new User(SystemUser.NAME, SystemUser.ROLE_NAME));
+        final IndexRequest request = mock(IndexRequest.class);
+        final String requestId = AuditUtil.getOrGenerateRequestId(threadContext);
+
+        RoleDescriptor role = new RoleDescriptor(SystemUser.ROLE_NAME, new String[] {}, null, null, null, null, null, null);
+        roleMap.put(SystemUser.ROLE_NAME, role);
```
ywangd, line 489: I think this is not the right test for the changes. The change is to (1) allow user-defined roles to use internal names (CRUD operations) and also (2) a role name like `_system` should be resolved correctly to the user-define role, not the internal one. 

I believe the test here is more for (2). But the critical piece of "resolving role names to roleDescriptors" is mocked out with `roleMap` and `buildRole(...)`. So what it really exercises is the orchestration logic and also RBACEngine with given roleDescriptor. The better place to test "resolving role names to roleDescriptors" is `CompositeRolesStoreTests`

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/AuthorizationServiceTests.java
```diff
@@ -479,6 +480,79 @@ public void testActionsForSystemUserIsAuthorized() {
...
+            SystemUser.ROLE_NAME,
+            new String[] { ClusterPrivilegeResolver.ALL.name() },
+            null,
+            null,
+            null,
+            null,
+            null,
+            null
+        );
+        roleMap.put(SystemUser.ROLE_NAME, role);
```
ywangd, line 536: Similar to the above. But in this case, such a test already exists in `CompositeRolesStoreTests` as `testDoesNotUseRolesStoreForXPacAndAsyncSearchUser` and `testGetRolesForSystemUserThrowsException`.

ywangd (commented):
I have a different view on some of the tests. Please see below for details.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/AuthorizationServiceTests.java
n1v0lg: The system user is special, in that there is additional logic in `AuthorizationService` that deals with it, before we even get to the role store:

https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java#L255-L258

and

https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java#L615-L625

I intended this and the second test to cover that logic, and the tests I added (or extended) in `CompositeRolesStoreTests` to handle the other resolving role name scenarios.

I agree that this test is less focused on role resolution and more on internal users and would have been better placed in my previous PR. Since it does give us additional coverage of a path that is closely related to role resolution I think it still makes sense to keep. WYT?

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/esnative/NativeRealmIntegTests.java
```diff
@@ -283,9 +291,9 @@ public void testAuthWithInternalUsernameFailsWithoutCorrespondingUser() {
         assertThat(e.status(), is(RestStatus.UNAUTHORIZED));
     }
 
-    private void testAddUserAndRoleThenAuth(String username) {
+    private void testAddUserAndRoleThenAuth(String username, String roleName) {
```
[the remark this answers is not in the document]
n1v0lg: Absolutely, happy to add a negative test 👍

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/AuthorizationServiceTests.java
ywangd: > I agree that this test is less focused on role resolution and more on internal users and would have been better placed in my previous PR. Since it does give us additional coverage of a path that is closely related to role resolution I think it still makes sense to keep. WYT?

OK. It's indeed more in context for #86398. I have no problem for you to keep it here. My main point is to make sure there is a test for CompositeRolesStore that checks:
(1) A role name like `_xpack_security` will _not_ be resolved to the internal role definition
(2) An internal user such as `XpackSecurityUser` can still get its role resolved correctly regardless whether there is an user-defined role called `_xpack_security`.

Some of those probably already exist. But it would be great if you could verify.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/esnative/NativeRealmIntegTests.java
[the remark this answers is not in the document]
n1v0lg: I've added a negative assertion (`assertClusterHealthOnlyAuthorizesWhenAnonymousRoleActive`) which was already conveniently used in other tests. The assertions checks that the user can't execute a health check. _system, _xpack, and _xpack_security all grant that role, `_security_profile` and `_async_search` do not. Let me know if I should go more granular than that!

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/AuthorizationServiceTests.java
n1v0lg: Makes sense! I've added added explicit tests for (1). 

For (2), `testDoesNotUseRolesStoreInternalUsers` verifies that there were no interactions with the stores, and `testGetRoleDescriptorsListForInternalUsers` does not even mock them; this should implicitly cover the assertion that resolution works regardless whether there is an user-defined role with a matching name.

ywangd (approved):
LGTM 

Thanks for the great work on the tests.

n1v0lg:
@elasticmachine update branch

n1v0lg:
@elasticmachine run elasticsearch-ci/rest-compatibility

PR #84941, opened 2022-03-14
add privileges for kIbana_system user to serve cloud security posture…

CohenIdo:
Add privileges for `kibana_system` user to allow support in Cloud Security Posture's feature.

PR #84279, opened 2022-02-23
[Osquery]  Extend kibana_system role with an access to osquery_manager indices

patrykkopycinski:
The PR adds proper permissions for the kibana_system user to two indices specific to Osquery actions.

The two indices are .logs-osquery_manager.actions-* and .logs-osquery_manager.action.responses-* .

The kibana_system user will create, write, and read from the .logs-osquery_manager.actions-* index to log new actions and visualize them for users.

The kibana_system user will read from .logs-osquery_manager.action.responses-*` to visualize responses from the Osquery for the user.

This change is needed as Osquery plugins currently are using raw Fleet actions, but in order to provide a better user experience, we need to store additional metadata about the actions that are taken. We need to ensure kibana_system can properly interact with these indices.

Adds access to indices added here https://github.com/elastic/integrations/pull/2563
Follows https://github.com/elastic/elasticsearch/pull/84279

PR #86609, opened 2022-05-10
[Osquery] Extend kibana_system role with an access to osquery_manager…

patrykkopycinski:
In order to support transforms we need to extend the current user with a privilege to create `responses` index https://github.com/elastic/elasticsearch/pull/84941#issuecomment-1081974019

Follow up for https://github.com/elastic/elasticsearch/pull/84279

```
The PR adds proper permissions for the kibana_system user to two indices specific to Osquery actions.

The two indices are .logs-osquery_manager.actions-* and .logs-osquery_manager.action.responses-* .

The kibana_system user will create, write, and read from the .logs-osquery_manager.actions-* index to log new actions and visualize them for users.

The kibana_system user will read from .logs-osquery_manager.action.responses-*` to visualize responses from the Osquery for the user.

This change is needed as Osquery plugins currently are using raw Fleet actions, but in order to provide a better user experience, we need to store additional metadata about the actions that are taken. We need to ensure kibana_system can properly interact with these indices.

Adds access to indices added here https://github.com/elastic/integrations/pull/2563
Follows https://github.com/elastic/elasticsearch/pull/84279
```

### DIFF
--- a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/store/ReservedRolesStore.java
+++ b/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/store/ReservedRolesStore.java
@@ -770,7 +770,7 @@ public class ReservedRolesStore implements BiConsumer<Set<String>, ActionListene
                     .privileges("auto_configure", "read", "write")
                     .build(),
                 // Osquery manager specific action responses. Kibana reads from these to display responses to the user.
-                RoleDescriptor.IndicesPrivileges.builder().indices(".logs-osquery_manager.action.responses-*").privileges("read").build(),
+                RoleDescriptor.IndicesPrivileges.builder().indices(".logs-osquery_manager.action.responses-*").privileges("create_index", "read", "index", "delete").build(),
                 // Osquery manager specific actions. Kibana reads and writes to this index to track new actions and display them.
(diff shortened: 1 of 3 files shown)


ywangd:
ping @elastic/kibana-security

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/store/ReservedRolesStore.java
```diff
@@ -770,7 +770,10 @@ public static RoleDescriptor kibanaSystemRoleDescriptor(String name) {
                     .privileges("auto_configure", "read", "write")
                     .build(),
                 // Osquery manager specific action responses. Kibana reads from these to display responses to the user.
-                RoleDescriptor.IndicesPrivileges.builder().indices(".logs-osquery_manager.action.responses-*").privileges("read").build(),
+                RoleDescriptor.IndicesPrivileges.builder()
```
azasypkin, line 773: __nit:__ would make sense to update the comment as well to reflect new privileges and what are they used for.

azasypkin (approved):
LGTM from the Kibana Security perspective.

ywangd:
@patrykkopycinski Next time please also wait for a "LGTM" from the @elastic/es-security team before merging. Thanks!

Issue #84478, opened 2022-03-01
Lightweight Logging API


ChrisHegarty:
### Description

This is a meta-issue tracking subtasks required to introduce a lightweight logging API for use by plugins.

The lightweight logging API offers plugins a stable API upon which to depend. The logging API will facilitate the logging needs of plugins without those plugins depending on any particular third-party library, or server internals. Moreover, any third-party logging library, like say Log4J, will be removed from the class loader hierarchy available to plugins.

The logging API is a small lightweight facade that locates a concrete implementation through an SPI. The logging API is agnostic of any other underlying logging library, like say, Log4J. The implementation however may use Log4J, or another popular logging library. The logging API has no dependency on server, it may however have a dependency on other libs like say, core or x-content (but care must be taken to reduce any dependencies as much as possible).

Stable plugin API
- [ ] Introduce basic usage for stable plugin API https://github.com/elastic/elasticsearch/pull/86612
- [ ] Logging api with Messages (ESLogMessage)
- [ ] DeprecationLogger 
- [ ] Prefix logger
- [ ] Move log4j off the system classloader (Implement subproject impl which has the log4j dependency)

ES dedicated logging api
- [ ] Logging Bootstrap
- [ ] Creating appenders
- [ ] Examine Logging audit trail

Refactorings
- [x] Remove context sensitive `getLogger` calls that do a stack walk. (#84480)
- [ ] Remove usage of `ParameterizedMessage` https://github.com/elastic/elasticsearch/issues/86549
- [ ] log4j utilities that are not used in logging context - Strings, Throwables 
- [ ] Refactor logging bootstrap and Loggers into ES dedicated api

Testing
- [ ] testing - ideally to not rely on log4j. More time consuming tests - EvilLoggerTests, CapturingLogger, JsonLoggerTests
- [ ]  LoggerUsageChecker - possibly a refactor and expanding to cover String.format https://github.com/elastic/elasticsearch/issues/87166
- [ ] MockLogAppender - widely used in our testing. An appender with additional assertion capabilities. THe refactoring can be performed where the log4j appender code is moved into `internal` implementation and the expectation(asertion) code is within the core/testing package which is part of the exported API.

Bridges
- [ ] SLF4J -> ES_Logging API bridge
- [ ] Log4j -> ES_Logging API bridge (or we can decide to use the log4j<->sl4j bridge

More:
- [ ] Refactor and extrapolate logging configuration in server

PR #86612, opened 2022-05-10
Stable logging API - the basic use case

pgomulka:
Introducing a stable logging API under libs/logging.
This change covers the most common use cases for logging: fetching a logger with LogManager, emitting a log messages with Logger and Level. 
It is influenced by log4j2-api, but do not include Marker and LogBuilder methods.
Also methods using org.apache.logging.log4j.util.Supplier are replaced with java.util.Supplier<String>

The basic implementation is present in server and injected statically in LogConfigurator

relates https://github.com/elastic/elasticsearch/issues/84478


pgomulka:
should we mark this as experimental?

libs/logging/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+tasks.named('forbiddenApisMain').configure {
+  // :libs:elasticsearch-core does not depend on server
+  // TODO: Need to decide how we want to handle for forbidden signatures with the changes to server
+  replaceSignatureFiles 'jdk-signatures'
+}
+tasks.named("compileJava").configure {
+  options.compilerArgs.add("-Xlint:-requires-automatic,-requires-transitive-automatic")
+  options.compilerArgs.add("-Xlint:-module") // qualified exports
+  options.compilerArgs.add("-Xlint:-exports") // implements Message!!
+}
```
ChrisHegarty, line 51: These compiler suppressions should not be needed (since we now do the suppressions in the Gradle Module path plugin)

libs/logging/src/main/java/org/elasticsearch/logging/Level.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /**
+     * Returns the name of this level.
+     */
+    public String name() {
+        return name;
+    }
+
+    public int getSeverity() {
```
ChrisHegarty, line 131: We should have at least some minimal Javadoc on each of the public members.

libs/logging/src/main/java/org/elasticsearch/logging/Level.java
```diff
@@ -0,0 +1,171 @@
...
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, severity);
+    }
+
+    /**
+     * Used for mapping between the API and the implementation
+     */
+    public static class StandardLevels {
+
```
ChrisHegarty, line 152: Do thees standard levels need to be part of the public API?

libs/logging/src/main/java/org/elasticsearch/logging/LogManager.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.elasticsearch.logging;
+
+import org.elasticsearch.logging.spi.LogManagerFactory;
+
+/**
+ * A class used for creating loggers.
+ */
+public class LogManager {
+
```
ChrisHegarty, line 17: minor; this class could be final, right?

ChrisHegarty:
> should we mark this as experimental?

Is there a standard way of doing that ?

libs/logging/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+  testImplementation(project(":test:framework")) {
+    exclude group: 'org.elasticsearch', module: 'elasticsearch-logging'
+  }
+}
+
+
+
+tasks.named("thirdPartyAudit").configure {
+  ignoreMissingClasses(
+      'org.osgi.framework.Bundle',
```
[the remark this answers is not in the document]
pgomulka: fair point. I previously had a log4j dependency which I now removed

libs/logging/src/main/java/org/elasticsearch/logging/Level.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Retrieves an existing Level or creates on if it didn't previously exist.
+     *
+     * @param name The name of the level.
+     * @param severity The integer value for the Level. If the level was previously created this value is ignored.
+     * @return The Level.
+     * @throws java.lang.IllegalArgumentException if the name is null or intValue is less than zero.
+     */
+    public static Level forName(String name, int severity) {
```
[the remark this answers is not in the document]
pgomulka: for custom levels like DeprecationLogger.CRITICAL https://github.com/elastic/elasticsearch/blob/76d7d583d7e4a8b9ca5e27eb6bb9880a6dc1eae1/server/src/main/java/org/elasticsearch/common/logging/DeprecationLogger.java#L41

libs/logging/src/main/java/org/elasticsearch/logging/spi/LogManagerFactory.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * SPI for creating new loggers
+ */
+public interface LogManagerFactory {
+
+    LogManagerFactory INSTANCE = loadProvider();
+
+    static LogManagerFactory loadProvider() {
+        ServiceLoader<LogManagerFactory> sl = ServiceLoader.load(LogManagerFactory.class, ClassLoader.getSystemClassLoader());
```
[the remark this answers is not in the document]
pgomulka: I was thinking that this class will change when we move the log4j off the system classloader. Something to what we have here https://github.com/ChrisHegarty/elasticsearch/blob/logging-module/libs/logging/src/main/java/org/elasticsearch/logging/spi/LogManagerFactory.java

libs/logging/src/main/java/org/elasticsearch/logging/spi/StandardLevels.java
```diff
@@ -0,0 +1,33 @@
...
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.logging.spi;
+
+/**
+ * Used for mapping between the API and the implementation
+ */
+public class StandardLevels {
```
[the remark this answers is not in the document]
pgomulka: combining this with https://github.com/elastic/elasticsearch/pull/86612/files#r880704533 ?
I am not sure how would this work. We want to initialise those constants statically.
This is hidden in spi, so should not be visible

libs/logging/src/main/java/org/elasticsearch/logging/Level.java
```diff
@@ -0,0 +1,154 @@
...
+        Objects.requireNonNull(name);
+        final String levelName = name.trim().toUpperCase(Locale.ROOT);
+        final Level level = LEVELS.get(levelName);
+        if (level != null) {
+            return level;
+        }
+        throw new IllegalArgumentException("Unknown level constant [" + levelName + "].");
+    }
+
+    public static Collection<Level> values() {
```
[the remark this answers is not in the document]
pgomulka: we don't have usages of org.apache.logging.log4j.Level#values in our codebase. Unlikely that plugins have too.
I only use this in testing. The only usages I could think of is to prevent exception when trying to get Level.forName
```
if(Level.values().contains("CustomName")) 
Level.forName("CustomName")

File: libs/logging/src/main/java/org/elasticsearch/logging/spi/LogManagerFactory.java (hunk 7 above)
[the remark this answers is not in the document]
Comment (ChrisHegarty) [reply libs/logging/src/main/java/org/elasticsearch/logging/spi/LogManagerFactory.java:23]: How about we look up the implementation using the service loader variant that accepts a module layer? Or is the question more related to the use of service loader at all?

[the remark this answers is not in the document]
Comment (rjernst) [reply libs/logging/src/main/java/org/elasticsearch/logging/spi/LogManagerFactory.java:23]: The latter. While eventually I think service loader will be used when we can move log4j off the system classloader, I was suggesting starting by loading this in server and setting statically here.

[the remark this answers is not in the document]
Comment (pgomulka) [reply libs/logging/src/main/java/org/elasticsearch/logging/spi/LogManagerFactory.java]: What is the benefit of setting this statically from server? We would have to remember to set this with LogConfigurator, and once we move implementation off the system classloader to remove it. 
I personally think this is an easier to read solution, but don't have strong opinion against setting as well.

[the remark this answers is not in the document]
Comment (rjernst) [reply libs/logging/src/main/java/org/elasticsearch/logging/spi/LogManagerFactory.java]: For stable plugin api, we are thinking about changing the way the system classloader is setup. Right now, everything from server is in the system classloader. For stable plugin api, we want to first load the stable api jars, then create a separate loader, parented by the stable api jars, that is everything from server. That separation would break this ServerLoader lookup, because the logging api impl would not be on the system classloader, and this code could not statically find the loader wtih the impl (since it will not yet have been created). Thus, setting this statically as an init of the logging api allows us to delay until server has actually been loaded.

File: libs/logging/src/main/java/org/elasticsearch/logging/Level.java (hunk 6 above)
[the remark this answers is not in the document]
Comment (pgomulka) [reply libs/logging/src/main/java/org/elasticsearch/logging/Level.java:89]: as we discussed offline. We do not need to allow plugin developers to use custom levels and deprecation logger can be hidden in the implementation.

File: libs/logging/src/main/java/org/elasticsearch/logging/Level.java (hunk 9 above)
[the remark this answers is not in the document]
Comment (pgomulka) [reply libs/logging/src/main/java/org/elasticsearch/logging/Level.java]: removed, because Level is now an enum

**Comment by pgomulka** [comment]
@elasticmachine run elasticsearch-ci/example-plugins

File: libs/logging/src/main/java/org/elasticsearch/logging/spi/StandardLevels.java (hunk 8 above)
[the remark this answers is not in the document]
Comment (pgomulka) [reply libs/logging/src/main/java/org/elasticsearch/logging/spi/StandardLevels.java]: StandardLevel removed. Level refactored into enum

File: libs/logging/src/main/java/org/elasticsearch/logging/Level.java [hunk 10]
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * A fatal error that will prevent the application from continuing.
+     */
+    FATAL,
+    /**
+     * An error in the application, possibly recoverable.
+     */
+    ERROR,
+    /**
+     * An event that might possible lead to an error.
```
ChrisHegarty, line 29: typo? "An event that might possiblY lead to an error."

libs/logging/src/main/java/org/elasticsearch/logging/LogManager.java
```diff
@@ -0,0 +1,39 @@
...
+public final class LogManager {
+
+    /**
+     * Returns a Logger with the specified name.
+     *
+     * @param name The logger name.
+     * @return The Logger.
+     */
+    public static Logger getLogger(final String name) {
+        return LogManagerFactory.provider().getLogger(name);
```
ChrisHegarty, line 25: I assume null in results in a NPE? If so, maybe a class-level comment, e.g.

  `Unless otherwise noted, passing a null argument to a method of any class will cause a NullPointerException`

libs/logging/src/main/java/org/elasticsearch/logging/LogManager.java
```diff
@@ -0,0 +1,39 @@
...
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.logging;
+
+import org.elasticsearch.logging.internal.spi.LogManagerFactory;
+
+/**
+ * A class used for creating loggers.
```
ChrisHegarty, line 14: Maybe expand this a little, e.g. 
  "A Manager of {@code Loggers}. This class consists of factory methods for creating and retrieving Loggers."

libs/logging/src/main/java/org/elasticsearch/logging/Logger.java
```diff
@@ -0,0 +1,718 @@
...
+     */
+    void log(Level level, String message);
+
+    /**
+     * Logs a String message which is only to be constructed if the logging level is the specified level.
+     *
+     * @param level           the logging level
+     * @param messageSupplier A function, which when called, produces the desired log String message;
+     */
+    void log(Level level, java.util.function.Supplier<String> messageSupplier, Throwable throwable);
```
ChrisHegarty, line 32: Suggest, highlighting the LAZY aspect. e.g. 

"Logs a lazily supplied String message associated with a given throwable. If the logger is currently enabled for the specified log message level, then a message is logged that is the result produced by the given supplier function."

Note: missing at param for `throwable`

libs/logging/src/main/java/org/elasticsearch/logging/LogManager.java
pgomulka: actually this depends on log4j. and if a null is passed as a `name` or `clazz` parameter, it will take caler's class 
`StackLocatorUtil.getCallerClass(2)`

ChrisHegarty: We should clarify this in the javadoc, no ? Or alternatively, just disallow null (throw NPE)

ChrisHegarty (approved):

libs/logging/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+apply plugin: 'elasticsearch.publish'
+apply plugin: 'elasticsearch.build'
+
+
+tasks.named("loggerUsageCheck").configure {enabled = false }
+
+dependencies {
+  testImplementation "com.carrotsearch.randomizedtesting:randomizedtesting-runner:${versions.randomizedrunner}"
```
rjernst, line 20: these should all be pulled in by the test:framework dep below right?

libs/logging/src/main/java/org/elasticsearch/logging/LogManager.java
rjernst: Has this been address in the javadoc? I'm not seeing it

libs/logging/src/main/java/org/elasticsearch/logging/Logger.java
```diff
@@ -0,0 +1,732 @@
...
+ * This is interface heavily influenced by org.apache.logging.log4j2.Logger.
+ * The most notable difference is lack of methods with Marker, LogBuilder and
+ * the message supplier is {@code java.util.function.Supplier<String>}
+ */
+public interface Logger {
+    /**
+     * Logs a message String with the given level.
+     *
+     * @param level   the logging level
+     * @param message the message CharSequence to log.
```
rjernst, line 22: String not CharSequence?

libs/logging/src/main/java/org/elasticsearch/logging/Logger.java
```diff
@@ -0,0 +1,732 @@
...
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.logging;
+
+/**
+ * Main interface for logging. Most operations are done through this interface.
+ * This is interface heavily influenced by org.apache.logging.log4j2.Logger.
```
rjernst, line 13: reversed words: `is interface` -> `interface is`

libs/logging/src/main/java/org/elasticsearch/logging/Logger.java
```diff
@@ -0,0 +1,732 @@
...
+
+    // -- error
+
+    /**
+     * Logs a lazily supplied String message. If the logger is currently enabled for
+     * {@link org.elasticsearch.logging.Level#ERROR ERROR} level, then a message is logged
+     * that is the result produced by the given supplier function.
+     * @param messageSupplier A function, which when called, produces the desired log String message;
+     */
+    void error(java.util.function.Supplier<String> messageSupplier);
```
rjernst, line 215: Is the fully qualified name for Supplier needed?

libs/logging/src/main/java/org/elasticsearch/logging/Logger.java
```diff
@@ -0,0 +1,732 @@
...
+     */
+    void trace(String message, Throwable throwable);
+
+    /**
+     * Logs a message with parameters at trace level.
+     *
+     * @param message the message to log
+     * @param p0      parameter to the message.
+     */
+    void trace(String message, Object p0);
```
rjernst, line 667: I wonder if we should omit the single arg variants of these methods for all levels. The reason is it can be ambiguous with the general method taking an array of args. I'm not sure if there is a performance hit for that, but we are talking about logging so I can't imagine a single implicit array creation is going to make a huge difference.

libs/logging/src/main/java/org/elasticsearch/logging/internal/spi/LogManagerFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.logging.internal.spi;
+
+import org.elasticsearch.logging.Logger;
+
+/**
+ * SPI for creating new loggers
```
rjernst, line 14: not really SPI anymore, it's just a static holder for the factory

server/src/main/java/org/elasticsearch/common/logging/internal/LogManagerFactoryImpl.java
```diff
@@ -0,0 +1,26 @@
...
+package org.elasticsearch.common.logging.internal;
+
+import org.apache.logging.log4j.LogManager;
+import org.elasticsearch.logging.Logger;
+import org.elasticsearch.logging.internal.spi.LogManagerFactory;
+
+public class LogManagerFactoryImpl extends LogManagerFactory {
+    @Override
+    public Logger getLogger(String name) {
+        // TODO PG logger impl instance caching
```
rjernst, line 18: Can you create an issue for this?

rjernst (commented):
This looks good overall. I have a few more nits.

I also think it would be good to generate the javadocs and see what they look like. I think some formatting is missing in several methods.

Also, this will need to be followed up by adding the jar to the release configuration.

libs/logging/src/main/java/org/elasticsearch/logging/LogManager.java
pgomulka: I did not add the javadoc. This depends on how log4j does this and I did not want to mention this in the API's javadoc. I think is is rare that a name is null.

libs/logging/src/main/java/org/elasticsearch/logging/Logger.java
pgomulka: good point, I will remove them

libs/logging/src/main/java/org/elasticsearch/logging/internal/spi/LogManagerFactory.java
pgomulka: but doesn't it still serve a purpose of SPI? the implementation has to provide the concrete getLogger methods that this class declares

rjernst: SPI normally has the connotation of being loaded by ServiceProvider, but this is not.

server/src/main/java/org/elasticsearch/common/logging/internal/LogManagerFactoryImpl.java
pgomulka: https://github.com/elastic/elasticsearch/issues/87511

rjernst: Thanks! Please add a link to the issue in the comment :)

libs/logging/src/main/java/org/elasticsearch/logging/Logger.java
rjernst: Yeah I think we should. We can always add them back.

server/src/main/java/org/elasticsearch/common/logging/internal/LoggerImpl.java
```diff
@@ -0,0 +1,228 @@
...
+@SuppressLoggerChecks(reason = "safely delegates to logger")
+public final class LoggerImpl implements Logger {
+    private final org.apache.logging.log4j.Logger log4jLogger;
+
+    public LoggerImpl(org.apache.logging.log4j.Logger log4jLogger) {
+        this.log4jLogger = log4jLogger;
+    }
+
+    private org.apache.logging.log4j.util.Supplier<?> mapSupplier(Supplier<String> msgSupplier) {
+        return (org.apache.logging.log4j.util.Supplier<Object>) () -> msgSupplier.get();
```
rjernst, line 26: is the cast necessary given the return type?

rjernst (approved):
LGTM

PR #86621, opened 2022-05-10
[ML] Add missing params to start deployment api REST spec

dimitris-athanasiou:
Updates the REST spec for the start deployment api.
In particular, adds params: `number_of_allocations`, `threads_per_allocation`,
and `queue_capacity`.


sethmlarson (approved):
This is great, thanks!

Could you also open a corresponding PR that adds these parameters to the [Elasticsearch specification](https://github.com/elastic/elasticsearch-specification) so we can start generating these parameters in clients? [This is the file where the change would be made](https://github.com/elastic/elasticsearch-specification/blob/main/specification/ml/start_trained_model_deployment/MlStartTrainedModelDeploymentRequest.ts). Reach out if you have questions :)

PR #86622, opened 2022-05-10
Expand jar hell to include modules

ChrisHegarty:
Expands jar hell to include modules from the system class loader.


libs/core/src/main/java/org/elasticsearch/jdk/JarHell.java
```diff
@@ -279,4 +319,12 @@ private static void checkClass(Map<String, Path> clazzes, String clazz, Path jar
             }
         }
     }
+
+    private static URL toURL(URI uri) {
+        try {
+            return uri.toURL();
+        } catch (MalformedURLException e) {
+            throw new RuntimeException(e);
```
rjernst, line 327: Since we always pass actual file URIs here, this should be impossible right? Can we throw AssertionError then, so we are less likely to hide a failure through a bare try/catch(exception)?

rjernst (approved):
LGTM

libs/core/src/main/java/org/elasticsearch/jdk/JarHell.java
ChrisHegarty: Done.

PR #86624, opened 2022-05-10
Remove ImmutableOpenMap from GetIndexResponse

rjernst:
This commit converts the maps exposed from GetIndexResponse to Map.

relates #86239


dakrone (approved):
LGTM

PR #86625, opened 2022-05-10
Update painless lookup for modules

ChrisHegarty:
When we move to modules, public lookup has only access to public members
of exported packages. For painless itself this is not sufficient to
lookup public members within its own module, since painless need not
export all of the packages that it requires access too.

The solution is to use a lookup that holds the MODULE mode for self 
introspection.


rjernst (approved):
LGTM

stu-elastic (approved):
lgtm

Issue #86549, opened 2022-05-09
 Remove usage of ParameterizedMessage in server


ChrisHegarty:
A common use of ParameterizedMessage in the ES codebase is to facilitate lazy evaluation of message arguments. ParameterizedMessage predates some of the newer lambda shaped Supplier-accepting log variants. Quite often using a String or argument Supplier is just as good, more readably, and likely slightly more efficient. 

Within server there are the following usages:

```
$ find server/src/main/java/ -name "*.java" -exec grep org.apache.logging.log4j.message.ParameterizedMessage {} \; -print | grep -e "\.java$" | sed "s/\/[a-zA-Z0-9]*.java$//" | sort | uniq | sed "s/^server/- [ ] server/"
```

for majority of use case a structured replace refactoring from intellij will be good option

relates #84478

pgomulka:
the refactoring is now done

PR #86631, opened 2022-05-10
Convert no arg parameterized message to direct strings

rjernst:
This commit converts all ParameterizedMessages which take no arguments
to direct logging messages. There is no need for lazy evaluation since
these strings are static.

relates #86549


pgomulka (approved):
LGTM

ChrisHegarty (approved):
LGTM

PR #86632, opened 2022-05-10
[ML] Make autoscaling and task assignment use same memory staleness definition

### LINKED ISSUE #86616: [ML] Oscillation of persistent task assignment can block ML autoscaling
It's possible that in certain situations the ML tier in Cloud does not autoscale up as it should.

This happens because persistent task assignment and autoscaling both refuse to run if the memory tracker is "stale", but both use different definitions of staleness.  Persistent task assignment considers the memory tracker stale if it wasn't refreshed in the last 90 seconds (by default), and autoscaling considers the memory tracker stale if it wasn't refreshed in the last 421 seconds (= learnt interval between autoscaling calls, which is usually 301 seconds, plus 2 different 60 second tolerances).

Autoscaling checks here: https://github.com/elastic/elasticsearch/blob/3da04514a570e0eb8103b06d7b6c192c40993ef7/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java#L477

Persistent task assignment checks here: https://github.com/elastic/elasticsearch/blob/1f24ffdacd2415b7b9eb6a188ab47ce084186203/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/task/AbstractJobPersistentTasksExecutor.java#L187

But the nasty problem is that if persistent task assignment finds the memory tracker stale then it changes the tasks assignment explanation to say this.  If autoscaling runs while the task has this assignment then it won't consider the task to be "waiting" here: https://github.com/elastic/elasticsearch/blob/3da04514a570e0eb8103b06d7b6c192c40993ef7/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java#L388

For example, instead of `{"executor_node":null,"explanation":"persistent task is awaiting node assignment."}` we get `{"executor_node":null,"explanation":"Not opening job [my_job] because job memory requirements are stale - refresh requested"}`.  In fact, the assignment explanation oscillates between these two states, which is visible in the allocation ID counting upwards despite no changes in the cluster: `"allocation_id":222`.  For a job that had never been successfully assigned to a node in an unchanging cluster we'd hope that that allocation ID would be 1 or 2, not 222.

If autoscaling runs while the assignment explanation relates to stale memory then it will bypass the scale up code, and send us into the "weird" code block here: https://github.com/elastic/elasticsearch/blob/3da04514a570e0eb8103b06d7b6c192c40993ef7/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java#L591

The safest way to avoid this discrepancy would seem to be to use the same definition of staleness everywhere.

droberts195:
Previously autoscaling used a staleness definition of ~7 minutes
and task assignment used a staleness definition of ~90 seconds.

This could lead to the assignment explanation of tasks not being
"awaiting lazy assignment" at the moment when autoscaling ran,
thus preventing an autoscaling decision being made.

The solution is to always use the same definition of staleness in
the memory tracker. This is set to the maximum of what the two
interested parties suggest.

Fixes #86616

### DIFF
--- a/docs/changelog/86632.yaml
+++ b/docs/changelog/86632.yaml
@@ -1,0 +1,6 @@
+pr: 86632
+summary: Make autoscaling and task assignment use same memory staleness definition
+area: Machine Learning
+type: bug
+issues:
+ - 86616
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java
@@ -367,13 +367,10 @@ public class MlAutoscalingDeciderService implements AutoscalingDeciderService, L
         if (isMaster == false) {
             throw new IllegalArgumentException("request for scaling information is only allowed on the master node");
         }
-        final Duration memoryTrackingStale;
-        long previousTimeStamp = this.lastTimeToScale;
-        this.lastTimeToScale = this.timeSupplier.getAsLong();
-        if (previousTimeStamp == 0L) {
-            memoryTrackingStale = DEFAULT_MEMORY_REFRESH_RATE;
-        } else {
-            memoryTrackingStale = Duration.ofMillis(TimeValue.timeValueMinutes(1).millis() + this.lastTimeToScale - previousTimeStamp);
+        long previousTimeStamp = lastTimeToScale;
+        lastTimeToScale = timeSupplier.getAsLong();
+        if (previousTimeStamp > 0L && lastTimeToScale > previousTimeStamp) {
+            mlMemoryTracker.setAutoscalingCheckInterval(Duration.ofMillis(lastTimeToScale - previousTimeStamp));
         }
 
         final ClusterState clusterState = context.state();
@@ -466,8 +463,11 @@ public class MlAutoscalingDeciderService implements AutoscalingDeciderService, L
             );
         }
 
-        if (mlMemoryTracker.isRecentlyRefreshed(memoryTrackingStale) == false) {
-            logger.debug("view of job memory is stale given duration [{}]. Not attempting to make scaling decision", memoryTrackingStale);
+        if (mlMemoryTracker.isRecentlyRefreshed() == false) {
+            logger.debug(
+                "view of job memory is stale given duration [{}]. Not attempting to make scaling decision",
+                mlMemoryTracker.getStalenessDuration()
+            );
             return buildDecisionAndRequestRefresh(reasonBuilder);
         }
         // We need the current node loads to determine if we need to scale up or down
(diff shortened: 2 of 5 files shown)


droberts195:

benwtrent (approved):
The change is probably OK from an autoscaling side of things. I am more worried about jobs allowing more stale memory views. But, if you are cool with that, this PR seems fine.

It does seem weird to me that this would occur frequently. 

It seems:

 - Job sees stale memory, asks for refresh doesn't get assigned yet
 - autoscaling doesn't see job as awaiting assignment, doesn't consider it for autoscaling
 - Either autoscaling completes the scaling request (up down or none) or sees memory needs refreshed
 - Job then gets assigned or awaiting assignment
 - autoscaling gets it on the next pass

droberts195:
What was happening was something like this:
* Initial open job API call refreshes memory
* Job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due stale memory, explanation set to stale memory and refresh requested
* 30 seconds later, job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due stale memory, explanation set to stale memory and refresh requested
* 20 seconds later, autoscaling call made and finds no jobs awaiting lazy assignment
* 10 seconds later, job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due stale memory, explanation set to stale memory and refresh requested
* 30 seconds later, job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due to insufficient space, explanation set to awaiting lazy assignment
* 30 seconds later, job cannot be assigned due stale memory, explanation set to stale memory and refresh requested
* 20 seconds later, autoscaling call made and finds no jobs awaiting lazy assignment

It might have stopped eventually, as the two cycles aren't as perfectly aligned as that example makes out.  But I observed a case where this went on for over 200 cycles, so it's certainly possible to be unlucky.

PR #86638, opened 2022-05-10
TSDB: Test `create_doc` permission

nik9000:
This adds a test for the `create_doc` and `index` permissions into tsdb
indices.

### DIFF
--- a/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/security/authz/70_tsdb.yml
+++ b/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/security/authz/70_tsdb.yml
@@ -198,3 +198,237 @@ document level security on tsid is not possible:
             tsids:
               terms:
                 field: _tsid
+
+---
+create_doc permission can create:
+  - skip:
+      version: " - 8.0.99"
+      reason: _tsid support introduced in 8.1.0
+      features: headers
+
+  - do:
+      security.put_role:
+        name: "createonly"
+        body:  >
+          {
+            "indices": [
+              {
+                "names": ["*"],
+                "privileges": ["create_doc"]
+              }
+            ]
+          }
+
+  - do:
+      security.put_user:
+        username: "limited"
+        body:  >
+          {
+            "password" : "x-pack-test-password",
+            "roles" : [ "createonly" ],
+            "full_name" : "user who can only create docs"
+          }
+
+  - do:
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
LGTM, thanks for adding a test!

nik9000:
@elasticmachine update branch

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/security/authz/70_tsdb.yml
```diff
@@ -198,3 +198,237 @@ document level security on tsid is not possible:
...
+          - '{"@timestamp": "2021-04-28T18:50:04.467Z", "metricset": "pod", "k8s": {"pod": {"name": "cat", "uid":"947e4ced-1786-4e53-9e0c-5c447e959507", "ip": "10.10.55.1", "network": {"tx": 2001818691, "rx": 802133794}}}}'
+  - match: { items.0.index.error.reason: "/is\\ unauthorized\\ for\\ user\\ \\[limited\\]/" }
+
+  - do:
+      headers: { Authorization: "Basic bGltaXRlZDp4LXBhY2stdGVzdC1wYXNzd29yZA==" } # limited - user
+      catch: "/is\ unauthorized\ for\ user\ \\[limited\\]/"
+      index:
+        refresh: true
+        index: test
+        op_type: index
```
tvernum, line 306: I'm curious what happens without this?
I assume that `AutoIdHandler` will set the op_type to `create` and then TSDB will reject the request because a document with the (auto generated) id already exists.

Is that the intended semantics of the `{ts-index}/_doc` endpoint - if there's a possibility that there might be an existing doc for the same tsid, then you need to set the op_type or be prepared to handle a failure?

tvernum (approved):
The test LGTM, but I have a question.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/security/authz/70_tsdb.yml
martijnvg: I think the `AutoIdHandler` will only set the op_type to create if no op_type has been specified. So it will overwrite a document?

nik9000: Yeah. It's a bit weird to me, but that's what happens. Without this you get create and overwrites stop in their tracks. I think it's a good thing that this isn't index dependent, but it *is* weird. And for _bulk we don't see it because folks say `create` or `index`. It's a quirk that'll have to be documented for tsdb.

PR #86642, opened 2022-05-10
[Refactor] Flexible general actions over plugins in PluginsService

williamrandolph:
PluginsService had a handful of methods that simply called a method from the Plugin superclass for every plugin in the plugin service. Here, we refactor this logic out to the call sites by adding flexible, generic methods for applying actions to all plugins.

We also find places in the Node class that used filterPlugin(Plugin.class) to accomplish the same thing, and replace those usages with these new methods.

### DIFF
--- a/server/src/main/java/org/elasticsearch/plugins/PluginsService.java
+++ b/server/src/main/java/org/elasticsearch/plugins/PluginsService.java
@@ -207,7 +207,7 @@ public class PluginsService implements ReportingService<PluginsAndModules> {
     }
 
     /**
-     * FlatMap a function over all plugins
+     * Map a function over all plugins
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -214,6 +206,34 @@ private static void logPluginInfo(final List<PluginInfo> pluginInfos, final Stri
         }
     }
 
+    /**
+     * FlatMap a function over all plugins
```
pugnascotia, line 210:
```suggestion
     * Map a function over all plugins
```

rjernst (approved):
LGTM

ChrisHegarty (approved):

PR #86647, opened 2022-05-10
TSDB: Initial reindex fix 

nik9000:
This teaches reindex the smallest thing that it needs to know about
tsdb: `_id` is automatically generated. Armed with that knowledge
reindex now doesn't attempt to copy the `_id` when writing to a tsdb
index.

Important: If the index doesn't yet exist it will *assume* that the
index will be created in `standard` mode. We can detect what mode it
*should* be created with in a follow up change.


modules/reindex/src/main/java/org/elasticsearch/reindex/Reindexer.java
```diff
@@ -285,7 +300,7 @@ protected RequestWrapper<IndexRequest> buildRequest(ScrollableHitSource.Hit doc)
             }
 
             // id and source always come from the found doc. Scripts can change them but they operate on the index request.
-            index.id(doc.getId());
+            index.id(destinationIndexIdMapper.reindexId(doc.getId()));
```
martijnvg, line 303: this will probably also speed reindex up

docs/changelog/86647.yaml
```diff
@@ -0,0 +1,5 @@
+pr: 86647
```
martijnvg, line 1: Should this be mentioned here? I don't think this changes anything with the tsdb feature flag in place? If so then we should label this PR also as non-issue.

martijnvg (approved):
LGTM

docs/changelog/86647.yaml
nik9000: Silly generated thing....

modules/reindex/src/main/java/org/elasticsearch/reindex/Reindexer.java
nik9000: I don't *think* so. It's a noop for standard indices. For tsdb indices it saves us some bits on the wire and a comparison. We still have to regenerate the id.

nik9000:
run elasticsearch-ci/packaging-tests-windows-sample

PR #86648, opened 2022-05-10
Clarify environments with strict firewalls and GEOIP (RE: #85637)

zembtach:
This reverts commit 1c52081b1f1adfb3292232bd1a744352e5f10c66.

There has been a misunderstanding somewhere. You cannot access the internet across an air gap. **That's the whole point of the air gap.**

If you connect your network to the internet you no longer have an air gap.

Please see the most well-established definition of an air gap in [RFC 4949](https://datatracker.ietf.org/doc/html/rfc4949#section-4).
```
   $ air gap
      (I) An interface between two systems at which (a) they are not
      connected physically and (b) any logical connection is not
      automated (i.e., data is transferred through the interface only
      manually, under human control). (See: sneaker net. Compare:
      gateway.)

      Example: Computer A and computer B are on opposite sides of a
      room. To move data from A to B, a person carries a disk across the
      room. If A and B operate in different security domains, then
      moving data across the air gap may involve an upgrade or downgrade
      operation.
```

Also, see [Wikipedia](https://en.wikipedia.org/wiki/Air_gap_(networking)).

> An air gap, air wall, air gapping or disconnected network is a network security measure employed on one or more computers to ensure that a secure computer network is physically isolated from unsecured networks, such as the public Internet or an unsecured local area network.

### DIFF
--- a/docs/reference/ingest/processors/geoip.asciidoc
+++ b/docs/reference/ingest/processors/geoip.asciidoc
@@ -327,6 +327,12 @@ a secure proxy. You can then specify the proxy endpoint URL in the
 <<ingest-geoip-downloader-endpoint,`ingest.geoip.downloader.endpoint`>> setting
 of each node’s `elasticsearch.yml` file.
 
+In a strict setup the following domains may need to be added to the allowed
+domains list:
+
+* `geoip.elastic.co`
+* `storage.googleapis.com`
+
 [[use-custom-geoip-endpoint]]
 **Use a custom endpoint**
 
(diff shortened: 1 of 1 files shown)


dakrone:
@wallrik thanks for opening this. I agree that this is not actually an "air-gap". I think, however, that instead of reverting the commit wholesale, it should be changed to say something like "In environments behind a strict firewall, the ES nodes require access to XYZ and ABC".

Would you be able to make the changes to the wording for this?

wallrik:
Yes, sure. However, if I'm adding new text, I also don't agree with putting the information in an _important_ box. I've made a new commit with a suggested edit that instead continues the proxy section with another paragraph. The section now reads as follows:

> **Use a proxy endpoint**
>
> If you can't connect directly to the Elastic GeoIP endpoint, consider setting up a secure proxy. You can then specify the proxy endpoint URL in the `ingest.geoip.downloader.endpoint` setting of each node’s `elasticsearch.yml` file.
>
> In a strict setup the following domains may need to be added to the allowed domains list:
>
> * `geoip.elastic.co`
> * `storage.googleapis.com`

Would you like for me to squash the commits @dakrone, or is this fine?

dakrone:
@elasticmachine ok to test

dakrone (approved):

PR #86649, opened 2022-05-10
Convert most single arg parameterized message to direct strings

rjernst:
This commit converts most ParameterizedMessages which take one argument
to direct logging messages. This was done with regex search/replace in
IntelliJ. It does omits if those single arguments have any parenthesis
(cast or method call). Those will be done in a followup.

relates #86549


ChrisHegarty (approved):
Nice!

pgomulka (approved):
LGTM

PR #86655, opened 2022-05-11
Add points metadata support for archive indices

ywelsch:
[Archive indices](https://github.com/elastic/elasticsearch/issues/81210) appear just as regular indices in a cluster, and can be part of index patterns when queried. To allow searches to quickly skip shards of archive indices that might not have relevant data, we're adding support for skipping shards of archive indices here that don't have data falling in the time range being queried. This is critical for the Kibana experience which relies on the date range picker to quickly skip some of the indices in an index pattern.

Doing the actual time-range query on the archive index is much more expensive as on a regular index (as it's using doc-values instead of points to run the query, equating to a full scan of the columnar data). The solution here is to make points metadata available in archive indices, so that the minimum and maximum value can be retrieved in constant time (only a tiny fraction of the full points capabilities).


ywelsch:

dnhatn (approved):
LGTM. Thanks Yannick.

ywelsch:
Thanks @dnhatn!

jpountz (commented):
LGTM too. (Sorry for being late to the party!)

PR #86656, opened 2022-05-11
[7.17] Keep build dependencies and gradle plugins up-to-date (#86611)

breskeby:
 - Keep build dependencies and gradle plugins up-to-date (#86611)


breskeby:
Backports the following commits to 7.17 but kept the docker compose plugin as is due to error reports on windows

Keep build dependencies and gradle plugins up-to-date (https://github.com/elastic/elasticsearch/pull/86611)

PR #86657, opened 2022-05-11
Avoid breaking add/clear voting exclusions

henningandersen:
Adding and clearing voting exclusions is part of the steps an operator/orchestration
will take to change topology of a cluster. It is therefore important that as much as
possible these do not fail, and we therefore now bypass the circuit breaker for the
associated rest and transport actions.


DaveCTurner (approved):
LGTM

idegtiarenko:
Should this also target 8.2.x?

idegtiarenko (approved):

henningandersen:
Failure looks like #86666, already fixed.

henningandersen:
@elasticmachine update branch

henningandersen:
Thanks David and Ievgen!

PR #73524, opened 2021-05-28
Change rest status code for TaskCancelledException to 400

iverase:
Currently when a rest action receives a`TaskCancelledException` , the response is translated to a HTTP 500 error (INTERNAL_SERVER_ERROR). This in turn gets logged as a warning in the Elasticsearch logs.

Until now this has not been a problem, but with the introduction of https://github.com/elastic/elasticsearch/pull/72688, this has become more probable and in application where there is a lot of cancellation it can lead to a very noise log with lots of warning messages.

Arguably, a task cancelation should not be consider a internal error as it is more normally related to something done by the client. There is no a good mapping between HTTP errors and task cancellation but after discussing with David and Armin, we think a 400 (BAD_REQUEST) is a better code for such an error.

This in turn will make  the error only be log in DEBUG mode.I have label the issue as a bug.

PR #72157, opened 2021-04-23
Add support for task cancellation to TransportMasterNodeAction

fcofdez:
This commit adds support for cancellation on TransportMasterNodeAction.
This is useful to provide cancellation for requests such as `TransportGetMappingsAction`

PR #86659, opened 2022-05-11
Use TaskCancelledException in TMNA

DaveCTurner:
In #72157 we made it so that cancelling the task that a
`TransportMasterNodeAction` was executing causes the action to fail with
a `java.util.concurrent.CancellationException`. This exception is
treated as a `500 Internal Server Error` which results in `WARN`-level
logs, but cancelling a task is normal and expected behaviour and should
not be logged (see #73524).

This commit fixes this by using a `TaskCancelledException` instead,
since this exception maps to a `400 Bad Request` and generates no logs.

### DIFF
--- a/docs/changelog/86659.yaml
+++ b/docs/changelog/86659.yaml
@@ -1,0 +1,5 @@
+pr: 86659
+summary: Use `TaskCancelledException` in TMNA
+area: Task Management
+type: bug
+issues: []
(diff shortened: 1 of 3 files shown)


fcofdez (approved):
LGTM, good catch

DaveCTurner:
:) I blame the reviewers...

DaveCTurner:
@elasticmachine please run elasticsearch-ci/part-1 (see https://github.com/elastic/elasticsearch/issues/86664)

PR #84242, opened 2022-02-22
[ML] allow autoscaling to work when vertical scaling is possible

benwtrent:
When an NLP model is deployed, or a DFA/Anomaly job is assigned, we have historically relied only on the `xpack.ml.max_lazy_ml_nodes` to determine if scaling is possible. But, in certain scenarios, it may be that scaling is available when `xpack.ml.max_lazy_ml_nodes` is fully satisfied. 

`xpack.ml.max_ml_node_size` is now checked to see if the current ML nodes exceed this size. If not, we assume vertical scaling is possible and allow the tasks to be created.

closes https://github.com/elastic/elasticsearch/issues/84198

PR #86660, opened 2022-05-11
[ML] Fix max_model_memory_limit reported by _ml/info with autoscaling

droberts195:
Previously the _ml/info endpoint took account of lazy nodes that
were not added to the cluster, but didn't take account of the
possibility of lazy nodes already in the cluster being able to
scale up to a larger size. This meant that the UI could incorrectly
cap the model_memory_limit that could be configured for new jobs
in autoscaling Cloud clusters where the tier limit meant only one
node per availability zone was possible. This PR fixes that case.

Fixes elastic/kibana#131954


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/utils/NativeMemoryCalculator.java
```diff
@@ -199,22 +199,25 @@ public static long dynamicallyCalculateJvmSizeFromNativeMemorySize(long nativeMa
...
         // to the cluster, in which case take those into account too
         long maxMlNodeSize = clusterSettings.get(MAX_ML_NODE_SIZE).getBytes();
         int maxLazyNodes = clusterSettings.get(MAX_LAZY_ML_NODES);
-        if (maxMlNodeSize > 0 && numMlNodes < maxLazyNodes) {
+        // Even if all the lazy nodes have been added to the cluster, we make
+        // the assumption that if any were configured they'll be able to grow
+        // to the maximum ML node size. (We assuming that lazy nodes always
+        // behave like they do with Elastic Cloud autoscaling, where vertical
+        // scaling is possible.)
+        if (maxMlNodeSize > 0 && maxLazyNodes > 0) {
```
droberts195, line 220: The fix here is consistent with the way #84242 was fixed.  That was a similar problem.

droberts195:
@elasticmachine update branch

droberts195:
@elasticmachine update branch

benwtrent (approved):

PR #86670, opened 2022-05-11
Add master_timeout support to voting config exclusions APIs

DaveCTurner:
Today the add/clear voting config exclusions APIs route a request to the
master node but do not expose the usual `?master_timeout` parameter
allowing to change the timeout for this phase of execution. This commit
adds the missing parameter.

### DIFF
--- a/server/src/main/java/org/elasticsearch/rest/action/admin/cluster/RestClearVotingConfigExclusionsAction.java
+++ b/server/src/main/java/org/elasticsearch/rest/action/admin/cluster/RestClearVotingConfigExclusionsAction.java
@@ -34,12 +34,16 @@ public class RestClearVotingConfigExclusionsAction extends BaseRestHandler {
 
     @Override
     protected RestChannelConsumer prepareRequest(final RestRequest request, final NodeClient client) throws IOException {
-        final var req = new ClearVotingConfigExclusionsRequest();
-        req.masterNodeTimeout(request.paramAsTime("master_timeout", req.masterNodeTimeout()));
-        req.setTimeout(req.masterNodeTimeout());
-        if (request.hasParam("wait_for_removal")) {
-            req.setWaitForRemoval(request.paramAsBoolean("wait_for_removal", true));
-        }
+        final var req = resolveVotingConfigExclusionsRequest(request);
         return channel -> client.execute(ClearVotingConfigExclusionsAction.INSTANCE, req, new RestToXContentListener<>(channel));
     }
+
+    static ClearVotingConfigExclusionsRequest resolveVotingConfigExclusionsRequest(final RestRequest request) {
+        final var resolvedRequest = new ClearVotingConfigExclusionsRequest();
+        resolvedRequest.masterNodeTimeout(request.paramAsTime("master_timeout", resolvedRequest.masterNodeTimeout()));
+        resolvedRequest.setTimeout(resolvedRequest.masterNodeTimeout());
+        resolvedRequest.setWaitForRemoval(request.paramAsBoolean("wait_for_removal", resolvedRequest.getWaitForRemoval()));
+        return resolvedRequest;
+    }
+
 }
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/elasticsearch/rest/action/admin/cluster/RestClearVotingConfigExclusionsAction.java
```diff
@@ -34,7 +34,9 @@ public String getName() {
 
     @Override
     protected RestChannelConsumer prepareRequest(final RestRequest request, final NodeClient client) throws IOException {
-        ClearVotingConfigExclusionsRequest req = new ClearVotingConfigExclusionsRequest();
+        final var req = new ClearVotingConfigExclusionsRequest();
+        req.masterNodeTimeout(request.paramAsTime("master_timeout", req.masterNodeTimeout()));
+        req.setTimeout(req.masterNodeTimeout());
```
henningandersen, line 39: Should we simply remove the timeout field instead and just use `masterNodeTimeout`?

server/src/test/java/org/elasticsearch/rest/action/admin/cluster/RestAddVotingConfigExclusionActionTests.java
```diff
@@ -56,4 +59,47 @@ public void testResolveVotingConfigExclusionsRequestNodeNames() {
         assertArrayEquals(expected, addVotingConfigExclusionsRequest.getNodeNames());
     }
 
+    public void testResolveVotingConfigExclusionsRequestTimeout() {
+        Map<String, String> params = new HashMap<>();
+        params.put("node_names", "node-1,node-2,node-3");
+        params.put("timeout", "60s");
+        RestRequest request = new FakeRestRequest.Builder(xContentRegistry()).withMethod(RestRequest.Method.PUT)
```
henningandersen, line 66: It would feel more intuitive to use `POST`, but I also see how it does not matter.

server/src/test/java/org/elasticsearch/rest/action/admin/cluster/RestAddVotingConfigExclusionActionTests.java
```diff
@@ -56,4 +59,47 @@ public void testResolveVotingConfigExclusionsRequestNodeNames() {
...
+            .withPath("/_cluster/voting_config_exclusions")
+            .withParams(params)
+            .build();
+
+        AddVotingConfigExclusionsRequest addVotingConfigExclusionsRequest = action.resolveVotingConfigExclusionsRequest(request);
+        assertEquals(TimeValue.timeValueSeconds(30), addVotingConfigExclusionsRequest.getTimeout());
+        assertEquals(TimeValue.timeValueMinutes(1), addVotingConfigExclusionsRequest.masterNodeTimeout());
+    }
+
+    public void testResolveVotingConfigExclusionsRequestTimeoutAndMasterTimeout() {
```
henningandersen, line 90: Ideally we would want a similar test for the clear voting exclusion case?

henningandersen (approved):
LGTM, consider all my comments here optional.

server/src/test/java/org/elasticsearch/rest/action/admin/cluster/RestAddVotingConfigExclusionActionTests.java
DaveCTurner: Good point, didn't spot that. Must have been left like that from a much earlier iteration.

server/src/main/java/org/elasticsearch/rest/action/admin/cluster/RestClearVotingConfigExclusionsAction.java
DaveCTurner: I did wonder about that but it is doing something a bit complicated that I'd rather leave alone for now.

server/src/test/java/org/elasticsearch/rest/action/admin/cluster/RestAddVotingConfigExclusionActionTests.java
DaveCTurner: Yes ok, see bc105beb25f58bc4ce9b4959080c84777c43360a.

PR #86673, opened 2022-05-11
Use ImmutableOpenMap for DatastreamMetadata to speed up DS operations

original-brownbear:
This is mainly motivated by the many-shards benchmark performance.
Working with `ImmutableOpenMap` is orders of magnitude faster than the immutable JDK immutable-map copy-on-write manipulations. This translates into about a 20% speedup for bootstrapping the 50k data streams cluster in the many shards benchmark.
The trade-off vs the JDK's immmutable maps is a trivial amount of memory overhead because the immutable open maps aren't filled to 100% capacity and a (barely measurable) reduction in iteration performance.

Before:
image

After:

(hint: the adding of a DS to the metadata is simply gone in this one and accounts for less than 0.1% of the overall profile vs 25% before)

image

relates #77466

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/metadata/DataStreamMetadata.java
+++ b/server/src/main/java/org/elasticsearch/cluster/metadata/DataStreamMetadata.java
@@ -13,9 +13,9 @@ import org.elasticsearch.cluster.Diff;
 import org.elasticsearch.cluster.DiffableUtils;
 import org.elasticsearch.cluster.NamedDiff;
 import org.elasticsearch.common.Strings;
+import org.elasticsearch.common.collect.ImmutableOpenMap;
 import org.elasticsearch.common.io.stream.StreamInput;
 import org.elasticsearch.common.io.stream.StreamOutput;
-import org.elasticsearch.common.util.Maps;
 import org.elasticsearch.xcontent.ConstructingObjectParser;
 import org.elasticsearch.xcontent.ParseField;
 import org.elasticsearch.xcontent.XContentBuilder;
@@ -23,7 +23,6 @@ import org.elasticsearch.xcontent.XContentParser;
 
 import java.io.IOException;
 import java.util.EnumSet;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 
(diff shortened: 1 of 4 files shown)


original-brownbear:
relates https://github.com/elastic/elasticsearch/issues/86239 which this shows we have to be careful with when it comes to large maps that we frequently update, `Map.copyOf` etc. are no alternative in this case performance wise because of the high cost of copying.

server/src/main/java/org/elasticsearch/cluster/metadata/DataStreamMetadata.java
```diff
@@ -34,69 +33,70 @@ public class DataStreamMetadata implements Metadata.Custom {
...
-        } else if (existing.equals(datastream)) {
+        if (datastream.equals(existing)) {
             return this;
         }
-        return new DataStreamMetadata(Maps.copyMapWithAddedOrReplacedEntry(dataStreams, name, datastream), dataStreamAliases);
+        return new DataStreamMetadata(ImmutableOpenMap.builder(dataStreams).fPut(name, datastream).build(), dataStreamAliases);
     }
 
-    public Map<String, DataStream> dataStreams() {
+    public ImmutableOpenMap<String, DataStream> dataStreams() {
```
martijnvg, line 95: Can the return type remain `Map`? I think we should at leat try to make `ImmutableOpenMap` an implementation detail of this class.

original-brownbear: I did it like this for now because we need to update the map in some of the methods in `Metadata`. I agree though, we should do something via copy-constructors here and just make the whole handling of this map internal to this class like we already did for adding a data stream. Maybe ok to push that into a follow-up to keep this change isolated?

martijnvg: > Maybe ok to push that into a follow-up to keep this change isolated?

If we do this in a follow up change then I'm good with this change.

original-brownbear: I'll handle it, promise :)

martijnvg (approved):
LGTM

original-brownbear:
Thanks Martijn + Joe (who reviewed this on another channel :))!

dakrone:
@original-brownbear I think we might need some kind of comment for this, as @rjernst is currently working on removing instances of `ImmutableOpenMap` and if there isn't communication about it, this change is likely to be undone sometime in the future.

PR #84336, opened 2022-02-24
Restrict run-as to realm and api_key authentication types

ywangd:
This PR removes run-as support for authentication types other than realm
and API key. The change essentially makes the behaviour closer to
the existing one (in released versions) except for API keys. This is not
to say that the existing behaviour is the best. But we need more time to
agree on the new behaviour.

Relates: #79809

PR #86411, opened 2022-05-03
[DOCS] JWT doc enhancements

lockewritesdocs:
This PR:

* Expands documentation for the `run_as` privilege page
* Includes an explanation of `run_as` in the JWT realm
* Adds more examples for JWT, such as `curl` commands and header requests
* Streamlines the steps for configuring ES to use a JWT realm

Note: The changes for the `run_as` page should be backported to 8.1 and 8.0 separately.

Preview links:

* [Configure Elasticsearch to use a JWT realm](https://elasticsearch_86411.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/jwt-realm.html#jwt-realm-configuration)
* [Applying the `run_as` privilege in the JWT realm](https://elasticsearch_86411.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/jwt-realm.html#_applying_the_run_as_privilege_to_roles)
* [Applying the `run_as` privilege to roles](https://elasticsearch_86411.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/run-as-privilege.html#run-as-privilege-apply)

PR #86680, opened 2022-05-11
Add run-as support for OAuth2 tokens

ywangd:
Authentication with an OAuth2 token can now perform run-as with changes
in this PR. This is in addition to the existing run-as support for realm
and API key authentication.

NB there are additional constraints on whether an OAuth2 token is
indeed qualified for run-as:
1. The token cannot itself already be a run-as (this can happen when the token is created using run-as and `client_credentials` grant)
2. The token cannot be created by anonymous or internal users

Service accounts cannot create tokens which meant run-as is not
supported for service accounts (none of existing service accounts
requires it anyway).

Relates: #84336

### DIFF
--- a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
+++ b/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
@@ -391,7 +391,7 @@ public class Authentication implements ToXContentObject {
     /**
      * Whether the authentication can run-as another user
      */
-    public boolean supportsRunAs(AnonymousUser anonymousUser) {
+    public boolean supportsRunAs(@Nullable AnonymousUser anonymousUser) {
         // Chained run-as not allowed
         if (isRunAs()) {
             return false;
--- a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
+++ b/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
@@ -242,12 +242,12 @@ public class Authentication implements ToXContentObject {
      * The security {@code RealmRef#Domain} of the resulting {@code Authentication} is that of the run-as user's realm.
      */
     public Authentication runAs(User runAs, @Nullable RealmRef lookupRealmRef) {
-        Objects.requireNonNull(runAs);
+        assert supportsRunAs(null);
         assert false == runAs instanceof RunAsUser;
         assert false == runAs instanceof AnonymousUser;
-        assert false == getUser() instanceof RunAsUser;
         assert false == hasSyntheticRealmNameOrType(lookupRealmRef) : "should not use synthetic realm name/type for lookup realms";
-        assert AuthenticationType.REALM == getAuthenticationType() || AuthenticationType.API_KEY == getAuthenticationType();
+
+        Objects.requireNonNull(runAs);
         return new Authentication(
             new RunAsUser(runAs, getUser()),
             getAuthenticatedBy(),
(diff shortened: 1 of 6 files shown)


justincr-elastic:
Can you clarify this in the description?
> The token cannot itself cannot already be a run-as

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
```diff
@@ -388,6 +388,55 @@ public boolean isApiKey() {
...
+        //
+        // This problem is at least partly due to we don't have special serialisation for the AnonymousUser class.
+        // As a result, it is serialised just as a normal user. At deserializing time, it is impossible to reliably
+        // tell the difference. This is what happens when AnonymousUser creates a token.
+        // Also, if anonymous access is disabled or anonymous username, roles are changed after the token is created.
+        // Should we still consider the token being created by an anonymous user which is now different from the new
+        // anonymous user?
+        if (getUser().equals(anonymousUser)) {
+            assert ANONYMOUS_REALM_TYPE.equals(getAuthenticatingSubject().getRealm().getType())
+                && ANONYMOUS_REALM_NAME.equals(getAuthenticatingSubject().getRealm().getName());
```
justincr-elastic, line 426: Should this be `||` based on your comment that a custom realm can return ANONYMOUS_REALM_NAME?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/AuthenticatorChain.java
```diff
@@ -203,15 +202,8 @@ void maybeLookupRunAsUser(Authenticator.Context context, Authentication authenti
...
-        // Both realm user and api_key can create tokens. They can also run-as another user and create tokens.
-        // In both cases, the created token will have a TOKEN authentication type and hence does not support run-as.
-        if (Authentication.AuthenticationType.REALM != authentication.getAuthenticationType()
-            && Authentication.AuthenticationType.API_KEY != authentication.getAuthenticationType()) {
-            logger.info(
-                "ignore run-as header since it is currently not supported for authentication type [{}]",
-                authentication.getAuthenticationType().name().toLowerCase(Locale.ROOT)
-            );
+        if (false == authentication.supportsRunAs(anonymousUser)) {
+            logger.info("ignore run-as header since it is currently not supported for authentication [{}]", authentication);
```
justincr-elastic, line 206: Log parameter changed from `authentication.getAuthenticationType().name().toLowerCase(Locale.ROOT)` to `authentication`. Is that design intent?

justincr-elastic:
I left a couple of drive by comments.

justincr-elastic:
Are there any OAuth2 tests that use Access Tokens formatted as JWTs?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
ywangd: The intention is to make sure we only treat ES's own anonymous user as the true anonymous user. So the user's realm must match both of the name and type of ES anonymous realm.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/AuthenticatorChain.java
ywangd: With the changes here, determine whether an authentication supports run-as is more involved and not just based on the authenticationType. So I changed it to log the whole authentication object, which will be easier for us to diagnose if the logging message ever comes to us in a SDH.

Technically we can log multiple different message but more precise about exactly which part of authentication preventing run-as. But I am not sure if it's worth it. Let me know if you feel strongly about it.

ywangd:
> Are there any OAuth2 tests that use Access Tokens formatted as JWTs?

Not intentionally. But practically the `JwtRestIT` can be considered as one because for a JWT realm to work successfully, the ES OAuth2 token authenticator must have decided that it should yield from the authentication chain.

lockewritesdocs:
@ywangd, as part of #86411, I'm clarifying that:

>Tokens that are created with the get token API are incompatible with the `run_as` privilege because OAuth2 tokens do not support this privilege.

This is true until your PR is merged. As part of this PR, I think that we should remove that sentence and then provide examples in the `run_as` section about how to generate OAuth2 tokens and use them with the `run_as` privilege.

ywangd:
> As part of your PR, I think that we should remove that sentence and then provide examples in the `run_as` section about how to generate OAuth2 tokens and use them with the `run_as` privilege.

Makse sense. I'll get to it when your PR is merged. Thanks for the reminder!

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java
```diff
@@ -388,6 +388,55 @@ public boolean isApiKey() {
         return effectiveSubject.getType() == Subject.Type.API_KEY;
     }
 
+    /**
+     * Whether the authentication can run-as another user
+     */
+    public boolean supportsRunAs(AnonymousUser anonymousUser) {
```
tvernum, line 394:
```suggestion
    public boolean supportsRunAs(@Nullable AnonymousUser anonymousUser) {
```

tvernum (approved):
LGTM

ywangd:
@lockewritesdocs This PR is now ready to be merged. Just need  a round of CI to pass. In this case, would it make more sense for you to update your PR on the docs? I am happy to help out with providing technical details if needed. Thanks!

lockewritesdocs:
@ywangd, go ahead and merge your PR and I'll remove the note about OAauth2. Please let me know if you have examples that we can use for `run_as`.

ywangd:
> go ahead and merge your PR and I'll remove the note about OAauth2. Please let me know if you have examples that we can use for `run_as`.

Thanks. I'll message you for examples in a separate channel.

PR #86681, opened 2022-05-11
[7.17] [ML] Fix max_model_memory_limit reported by _ml/info with autoscaling

droberts195:
Previously the _ml/info endpoint took account of lazy nodes that
were not added to the cluster, but didn't take account of the
possibility of lazy nodes already in the cluster being able to
scale up to a larger size. This meant that the UI could incorrectly
cap the model_memory_limit that could be configured for new jobs
in autoscaling Cloud clusters where the tier limit meant only one
node per availability zone was possible. This PR fixes that case.

Backport of #86660


droberts195:

PR #86685, opened 2022-05-11
Improve "Has Privilege" performance for boolean-only response

albertzaharovits:
Boolean-only privilege checks, i.e. the ones currently used in the "profile has privilege" API,
now benefit from a performance improvement, because the check will now stop upon first
encountering a privilege that is NOT granted over a resource (and return `false` overall).
Previously, all the privileges were always checked over all the resources in order to assemble
a comprehensive response with all the privileges that are not granted.


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/AuthorizationEngine.java
```diff
@@ -211,12 +212,15 @@ void validateIndexPermissionsAreSubset(
...
      * @param privilegesToCheck the object that contains the privileges to check for the Subject
+     * @param runDetailedCheck whether to return the details on which privileges are granted and which are not.
+     *                         {@code PrivilegesCheckResult#details} must be {@code null} only if this parameter is {@code false}.
      * @param applicationPrivilegeDescriptors a collection of application privilege descriptors
      * @param listener the listener to be notified of the check privileges response
      */
     void checkPrivileges(
         AuthorizationInfo authorizationInfo,
         PrivilegesToCheck privilegesToCheck,
+        boolean runDetailedCheck,
```
ywangd, line 223: I understand this PR is still draft, but I strongly think `runDetailedCheck` should be part of `PrivilegesToCheck` class instead of a separate parameter. A few reasons:
* In a single request flow, there is always just one way to check the privileges, either complete or short-circuited. It is by nature part of the check request itself, not a separate thing. I don't think we would ever support running both complete and short-circuited checks for a single request.
* The values for `privilegesToCheck` and `runDetailedCheck` should be determined at REST layer. We talked about that in future ProfileHasPrivileges can also run the complete check on an optional parameter. This parameter is from the REST layer and should be part of the privilegeCheck request itself. 
* Conceptually, a 1-to-1 relationship between check and result is much easier to understand. So I'd keep the variation of the check inside the check itself instead of keeping it outside as a separate variable.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/Role.java
```diff
@@ -90,12 +90,16 @@ public interface Role {
...
+     * @return {@code true} when all the privileges are granted over all the resources, or {@code false} otherwise
      */
-    ResourcePrivilegesMap checkIndicesPrivileges(
+    boolean checkIndicesPrivileges(
         Set<String> checkForIndexPatterns,
         boolean allowRestrictedIndices,
-        Set<String> checkForPrivileges
+        Set<String> checkForPrivileges,
+        @Nullable ResourcePrivilegesMap.Builder resourcePrivilegesMapBuilder
     );
```
albertzaharovits, line 103: I would like to take some space to explain my interface decisions for: `Role#checkIndicesPrivilegeges`, `Role#checkApplicationResourcePrivileges`, `IndicesPermission#checkREsourcePrivileges`, and `ApplicationPermission#checkResourcePrivileges`.

These methods must be aware of the mode of operation, ie whether to short-circuit or not, because they must return early in the short-circuit mode (in addition, the response "type" is different between the 2 modes) . So, one of the parameters of these methods must express this operation mode.

In this case, I think there are two implementation options to go about it:
* pass a dedicated parameter, to signal the short-circuit mode, and make the return value contain an optional field which is `null` when in the short-circuit mode; eg `RBACEngine#checkPrivileges(..., boolean runDetailedCheck, ..., ActionListener<PrivilegesCheckResult> listener)` where `PrivilegesCheckResult` has two fields: `boolean allChecksSuccess` and an optional `Details details`.
* return 2 values, rather than a single one with an optional `null` field. The first return value is the `allChecksSuccess` flag. The second returned value is via an out-parameter 👻 . When the out-parameter is non-`null`, it is used to return the privileges details, ie it works in the regular, non-short-circuit mode; otherwise, if `null`, no check details are returned and the operation mode is short-circuit. 

I had 2 reasons to choose the second option for these methods (unlike `RBACEngine#checkPrivileges`):
* I disliked having a `boolean` parameter going from the REST handler all the layers through to the permission level inside a role.
* `ResourcePrivilegesMap` has a boolean field `allowAll`. Its value is currently driven by the individual privilege check results. But In the short-circuit mode, it becomes separated from the check results, as it can be `true`/`false` when the individual privilege check results are `null`. With such a separated `allowAll` parameter, keeping it still a member of the `ResourcePrivilegesMap` felt janky (what does it mean to update its value in the non-short-circuit mode, for every privilege result, when its value can technically also be set externally - the encapsulation was gone).

I hope the scare of an out-parameter is not too great.
I think this is the more elegant implementation of the two. I like that these `check*` methods always return a simple boolean. Also, the out parameter works OK in the application check section in `RBACEngine#checkPrivileges` and in the `LimitedRole` class, I think.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/AuthorizationEngine.java
ywangd: Thanks for taking my suggestion. I really appreciate it. Here are some clarifications on my reasonings for posterity.

> I backtracked on it because the PrivilegesToCheck existing fields are parsed directly from the request body (for the "profile has privilege" endpoint) and it felt hacky to sneak in an extra argument in the constructor there.

My thinking is that the `mode` parameter may as well also be parsed from the request body. For now, we will supply it to the parsing method. But when we support optional full check for profileHasPrivilege in future, this `mode` will be something that users provide in the request body.

> I actually feel differently on this. I see it as a single "collection" of privileges that can be checked in 2 different ways. The class is used to group the "things" to check and the new parameter is about the "mode" to check.

I agree your way of viewing it as 'a single "collection" of privileges that can be checked in 2 different ways'. But my conclusion is different (in whether they should be kept together). My thinking is that there should be a single class that represent the entire check request. Maybe our different views have something to do with the class name. Imagine if it is called HasPrivilegeRequest (not proposing this name), I guess it would feel more natural to have it enclose both pieces of information?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/Role.java
ywangd: I noticed the out parameters on my first reading of the PR. I would probably approach it differently. But what is proposed in this PR works as well. Plus you already considered the other option and ruled it out with reasons. Also your reasoning about the standalone `allowAll` field and the individual check results of `ResourcePrivilegesMap` is a good one. 

In summary, I am OK with the out parameters. 

PS: In future, we may need support caching for LimitedRole (now it is only supported for SimpleRole). For that I am thinking that the result intersecting between `role` and `limited-by-role` should be done at the overall `PrivilegesCheckResult` level rather than the individual checks level (and also move most `RBCEngine#checkPrivileges` to `Role` level). This has nothing to do with the current PR. Just thinking out loud in case it is something you also thought about.

albertzaharovits: > In summary, I am OK with the out parameters.

🙇 

> PS: In future, we may need support caching for LimitedRole (now it is only supported for SimpleRole). For that I am thinking that the result intersecting between role and limited-by-role should be done at the overall PrivilegesCheckResult level rather than the individual checks level (and also move most RBCEngine#checkPrivileges to Role level). This has nothing to do with the current PR. Just thinking out loud in case it is something you also thought about.

No I haven't considered something like that. I'm not sure I follow either. My first approach would be to investigate whether the `LimitedRole` can be cached like the `Role` can.

albertzaharovits:
@ywangd this is ready for review

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/RBACEngine.java
```diff
@@ -548,61 +548,100 @@ public void checkPrivileges(
                 listener.onResponse(result);
                 return;
             }
+            maybeCachingListener = listener.delegateFailure((delegateListener, privilegesCheckResult) -> {
+                try {
+                    simpleRole.cacheHasPrivileges(settings, privilegesToCheck, privilegesCheckResult);
+                } catch (ExecutionException e) {
```
ywangd, line 554: I think we should be more liberal here and catch `Exception` instead of just `ExecutionException`. We don't want any future possible error to slip through this catch block.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/RBACEngine.java
```diff
@@ -527,14 +528,13 @@ public void checkPrivileges(
...
+                "Check whether role [{}] has privileges [{}]",
                 Strings.arrayToCommaDelimitedString(userRole.names()),
-                Arrays.toString(privilegesToCheck.cluster()),
-                Arrays.toString(privilegesToCheck.index()),
-                Arrays.toString(privilegesToCheck.application())
+                privilegesToCheck
             )
         );
 
+        final ActionListener<PrivilegesCheckResult> maybeCachingListener;
```
ywangd, line 537: I'd suggestion we rename this variable to `listener` and the current `listener` to something like `originalListener`. There is a tendency that people would just call `listener` in future since it is the norm. So I think renaming it can help.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/RBACEngine.java
```diff
@@ -548,61 +548,100 @@ public void checkPrivileges(
...
+        boolean allMatch = true;
+
+        final Map<String, Boolean> clusterPrivilegesCheckResults = new HashMap<>();
         for (String checkAction : privilegesToCheck.cluster()) {
-            cluster.put(checkAction, userRole.grants(ClusterPrivilegeResolver.resolve(checkAction)));
+            boolean privilegeGranted = userRole.grants(ClusterPrivilegeResolver.resolve(checkAction));
+            allMatch = allMatch && privilegeGranted;
+            if (privilegesToCheck.runDetailedCheck()) {
+                clusterPrivilegesCheckResults.put(checkAction, privilegeGranted);
+            } else if (false == privilegeGranted) {
```
ywangd, line 574: Nit: I feel it is more natural to test `allMatch` instead. It is after all the actual value we want to return (details or not).

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/RBACEngine.java
```diff
@@ -548,61 +548,100 @@ public void checkPrivileges(
...
                 Sets.newHashSet(check.getIndices()),
                 check.allowRestrictedIndices(),
-                Sets.newHashSet(check.getPrivileges())
+                Sets.newHashSet(check.getPrivileges()),
+                combineIndicesResourcePrivileges
             );
-            allMatch = allMatch && resourcePrivileges.allAllowed();
-            combineIndicesResourcePrivileges.addResourcePrivilegesMap(resourcePrivileges);
+            allMatch = allMatch && privilegesGranted;
+            if (false == privilegesToCheck.runDetailedCheck() && false == privilegesGranted) {
```
ywangd, line 591: Same here for testing `allMatch`.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/RBACEngine.java
```diff
@@ -548,61 +548,100 @@ public void checkPrivileges(
...
                         applicationName,
                         Sets.newHashSet(p.getResources()),
                         Sets.newHashSet(p.getPrivileges()),
-                        applicationPrivileges
+                        applicationPrivileges,
+                        resourcePrivilegesMapBuilder
                     );
-                    builder.addResourcePrivilegesMap(appPrivsByResourceMap);
+                    allMatch = allMatch && privilegesGranted;
+                    if (false == privilegesToCheck.runDetailedCheck() && false == privilegesGranted) {
```
ywangd, line 617: Here as well for `allMatch`.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/RBACEngine.java
```diff
@@ -548,61 +548,100 @@ public void checkPrivileges(
...
+                    allMatch,
+                    new PrivilegesCheckResult.Details(
+                        clusterPrivilegesCheckResults,
+                        combineIndicesResourcePrivileges.build().getResourceToResourcePrivileges(),
+                        privilegesByApplication
+                    )
+                )
+            );
+        } else {
+            maybeCachingListener.onResponse(PrivilegesCheckResult.ALL_CHECKS_SUCCESS_NO_DETAILS);
```
ywangd, line 643: Should we assert `allMatch` is indeed `true`?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/RBACEngine.java
```diff
@@ -548,61 +548,100 @@ public void checkPrivileges(
...
+            maybeCachingListener.onResponse(
+                new PrivilegesCheckResult(
+                    allMatch,
+                    new PrivilegesCheckResult.Details(
+                        clusterPrivilegesCheckResults,
+                        combineIndicesResourcePrivileges.build().getResourceToResourcePrivileges(),
+                        privilegesByApplication
+                    )
+                )
+            );
```
ywangd, line 641: I wouldn't mind to assert `combineIndicesResourcePrivileges` is not `null`.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/AuthorizationServiceTests.java
```diff
@@ -2780,21 +2782,19 @@ public void testAuthorizationEngineSelectionForCheckPrivileges() throws Exceptio
...
         );
         doAnswer(i -> {
             assertThat(i.getArguments().length, equalTo(4));
             final Object arg1 = i.getArguments()[0];
             assertThat(arg1, instanceOf(AuthorizationInfo.class));
             AuthorizationInfo authorizationInfoArg = (AuthorizationInfo) arg1;
-            final Object arg4 = i.getArguments()[3];
-            assertThat(arg4, instanceOf(ActionListener.class));
+            final Object arg5 = i.getArguments()[3];
+            assertThat(arg5, instanceOf(ActionListener.class));
```
ywangd, line 2795: Just curious: why does this have to be `arg5` now?

ywangd (approved):
LGTM

Some comments nothing major. Up to you how to deal with them. Thanks!

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/AuthorizationServiceTests.java
albertzaharovits: It should be `arg4`.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/RBACEngine.java
albertzaharovits: Yep, I made it so.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/RBACEngine.java
albertzaharovits: Okay, I made it so.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/ResourcePrivileges.java
```diff
@@ -69,7 +69,7 @@ public static Builder builder(String resource) {
 
     public static final class Builder {
         private final String resource;
-        private Map<String, Boolean> privileges = new HashMap<>();
+        private Map<String, Boolean> privileges = new TreeMap<>();
```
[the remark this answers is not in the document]
albertzaharovits: Yes, for consistent order. There were some tests in the `RBACEngine` that fail otherwise, because I changed them to randomize the privileges-to-check order (also add more privileges to check).

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/ResourcePrivilegesMap.java
```diff
@@ -20,25 +19,19 @@
...
 
-    public ResourcePrivilegesMap(boolean allAllowed, Map<String, ResourcePrivileges> resToResPriv) {
-        this.allAllowed = allAllowed;
+    public ResourcePrivilegesMap(Map<String, ResourcePrivileges> resToResPriv) {
         this.resourceToResourcePrivileges = Collections.unmodifiableMap(Objects.requireNonNull(resToResPriv));
     }
 
-    public boolean allAllowed() {
-        return allAllowed;
-    }
```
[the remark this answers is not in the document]
albertzaharovits: Part of the ethos of this change was to keep the `ResourcePrivilegesMap` reserved for the check details only, so not include the authorization result (because they can exist separately). The `allAllowed` result can be confounded with the authorization result. Overall, I think it is better we don't expose it.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/ProfileHasPrivilegesRequest.java
```diff
@@ -36,7 +36,7 @@ public class ProfileHasPrivilegesRequest extends ActionRequest {
         PARSER.declareStringArray(constructorArg(), Fields.UIDS);
         PARSER.declareField(
             constructorArg(),
-            parser -> RoleDescriptor.parsePrivilegesToCheck("profile_has_privileges_request", parser),
+            parser -> RoleDescriptor.parsePrivilegesToCheck("profile_has_privileges_request", false, parser),
```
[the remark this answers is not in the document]
albertzaharovits: I don't think so, because there's nothing downstream relying on the value to have a certain value (ie part of the reason to make it a class member was to not code assumptions like that).

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportHasPrivilegesAction.java
```diff
@@ -80,21 +85,21 @@ protected void doExecute(Task task, HasPrivilegesRequest request, ActionListener
...
-                            privilegesCheckResult.allMatch(),
-                            privilegesCheckResult.cluster(),
-                            privilegesCheckResult.index().values(),
-                            privilegesCheckResult.application()
-                        )
-                    )
+                            privilegesCheckResult.allChecksSuccess(),
+                            privilegesCheckResult.getDetails() != null ? privilegesCheckResult.getDetails().cluster() : Map.of(),
+                            privilegesCheckResult.getDetails() != null ? privilegesCheckResult.getDetails().index().values() : List.of(),
+                            privilegesCheckResult.getDetails() != null ? privilegesCheckResult.getDetails().application() : Map.of()
```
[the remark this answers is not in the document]
albertzaharovits: Yeah, we can do that later, I have no issue with null checks in cases like this (comparted with the "instance of" checks of the alternative).

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/AuthorizationEngine.java
```diff
@@ -329,16 +346,61 @@ public String toString() {
...
 
-    record PrivilegesCheckResult(
-        boolean allMatch,
-        Map<String, Boolean> cluster,
-        Map<String, ResourcePrivileges> index,
-        Map<String, Collection<ResourcePrivileges>> application
-    ) {}
+    /**
+     * The result of a (has) privilege check. This is not an Elasticsearch authorization result (though clients can base their authorization
+     * decisions on this response). The {@link #allChecksSuccess} field tells if all the privileges are granted over all the resources.
```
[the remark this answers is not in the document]
albertzaharovits: I mean that this not to be used/interpreted as the Elasticsearch authorization result.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/authz/permission/LimitedRoleTests.java
```diff
@@ -459,242 +529,325 @@ public void testGetApplicationPrivilegesByResource() {
...
+                allowRestrictedIndices,
+                checkForPrivileges,
+                resourcePrivilegesMapBuilder
+            );
+            assertThat(privilegesCheck, is(expectedCheckResult));
+        } else {
+            privilegesCheck = role.checkIndicesPrivileges(checkForIndexPatterns, allowRestrictedIndices, checkForPrivileges, null);
+            assertThat(privilegesCheck, is(expectedCheckResult));
+        }
+        assertThat(resourcePrivilegesMapBuilder.build(), equalTo(expectedAppPrivsByResource));
```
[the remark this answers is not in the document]
albertzaharovits: I prefer we keep it. I think there's a high chance the order can make a difference when we evolve the caching.

albertzaharovits:

albertzaharovits:

albertzaharovits:

PR #86700, opened 2022-05-11
[8.0] DOC Reinstate deprecation for ingest privileges with mapping update

albertzaharovits:
We missed, in v8.0, to actually break the deprecated behavior from #60024 .
Consequently, the deprecated behavior still works in all 8.* and it still generates
deprecation logs and headers.
This PR reinstates the doc side of the deprecation notice,
the plan being that we're now going to break the behavior in v9.

Co-authored-by: Adam Locke <adam.locke@elastic.co>


albertzaharovits:
@elasticmachine update this

PR #86701, opened 2022-05-11
Move static plugin loading methods from PluginsService to util class

williamrandolph:
We had a set of utility methods in PluginsService that were public and static so that the plugin cli tool could use them. Here, we move those methods to a utility class so that the plugin CLI doesn't need to import PluginsService. For the same reason, we also move "PluginsService.Bundle" to a top-level class called "PluginBundle." Finally, we move tests to the appropriate test classes.

Arguably these classes could go to a new plugin package in `org.elasticsearch.common`, but that would have given this diff a bigger footprint, and I would want to save it for a purely mechanical PR rather than one where I've used judgment in moving tests around.


server/src/main/java/org/elasticsearch/plugins/PluginBundle.java
```diff
@@ -0,0 +1,77 @@
...
+import java.net.URL;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.LinkedHashSet;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * A "bundle" is a group of jars that will be loaded in a single classloader
```
rjernst, line 21: This isn't entirely true: spi jars may be loaded in a separate loader.

rjernst (approved):
LGTM, one nit on the javadoc.

server/src/main/java/org/elasticsearch/plugins/PluginBundle.java
williamrandolph: I changed it to "A "bundle" is a group of jars for an Elasticsearch plugin. Ultimately, the PluginsService will load these jars, creating any necessary classloaders."

PR #86702, opened 2022-05-11
Grants `kibana_system` reserved role access to Endpoint event streams.

pjhampton:
To build event timelines from Endpoint alerts the Kibana internal user needs to access user data to construct the tree.

### DIFF
--- a/x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/authz/store/ReservedRolesStoreTests.java
+++ b/x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/authz/store/ReservedRolesStoreTests.java
@@ -686,20 +686,12 @@ public class ReservedRolesStoreTests extends ESTestCase {
             assertThat(kibanaRole.indices().allowedIndicesMatcher("indices:foo").test(mockIndexAbstraction(index)), is(false));
             assertThat(kibanaRole.indices().allowedIndicesMatcher("indices:bar").test(mockIndexAbstraction(index)), is(false));
             assertThat(kibanaRole.indices().allowedIndicesMatcher(GetIndexAction.NAME).test(mockIndexAbstraction(index)), is(true));
-            assertThat(kibanaRole.indices().allowedIndicesMatcher(CreateIndexAction.NAME).test(mockIndexAbstraction(index)), is(false));
             assertThat(kibanaRole.indices().allowedIndicesMatcher(IndexAction.NAME).test(mockIndexAbstraction(index)), is(false));
             assertThat(kibanaRole.indices().allowedIndicesMatcher(DeleteAction.NAME).test(mockIndexAbstraction(index)), is(false));
             assertThat(kibanaRole.indices().allowedIndicesMatcher(SearchAction.NAME).test(mockIndexAbstraction(index)), is(true));
             assertThat(kibanaRole.indices().allowedIndicesMatcher(MultiSearchAction.NAME).test(mockIndexAbstraction(index)), is(true));
             assertThat(kibanaRole.indices().allowedIndicesMatcher(GetAction.NAME).test(mockIndexAbstraction(index)), is(true));
             assertThat(kibanaRole.indices().allowedIndicesMatcher(READ_CROSS_CLUSTER_NAME).test(mockIndexAbstraction(index)), is(false));
-
-            // Privileges needed for Fleet package upgrades
-            assertThat(kibanaRole.indices().allowedIndicesMatcher(UpdateSettingsAction.NAME).test(mockIndexAbstraction(index)), is(true));
-            assertThat(kibanaRole.indices().allowedIndicesMatcher(PutMappingAction.NAME).test(mockIndexAbstraction(index)), is(true));
-            assertThat(kibanaRole.indices().allowedIndicesMatcher(RolloverAction.NAME).test(mockIndexAbstraction(index)), is(true));
-            // Privileges needed for installing current ILM policy with delete action
-            assertThat(kibanaRole.indices().allowedIndicesMatcher(DeleteIndexAction.NAME).test(mockIndexAbstraction(index)), is(true));
         });
 
         Arrays.asList(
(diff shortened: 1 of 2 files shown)


donaherc (approved):
LGTM! Refer to https://github.com/elastic/security-team/issues/2223 for more context on why the permission is needed

stevewritescode:
FYI @joe-desimone

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/authz/store/ReservedRolesStoreTests.java
```diff
@@ -681,6 +681,27 @@ public void testKibanaSystemRole() {
...
+            assertThat(kibanaRole.indices().allowedIndicesMatcher(MultiSearchAction.NAME).test(mockIndexAbstraction(index)), is(true));
+            assertThat(kibanaRole.indices().allowedIndicesMatcher(GetAction.NAME).test(mockIndexAbstraction(index)), is(true));
+            assertThat(kibanaRole.indices().allowedIndicesMatcher(READ_CROSS_CLUSTER_NAME).test(mockIndexAbstraction(index)), is(false));
+
+            // Privileges needed for Fleet package upgrades
+            assertThat(kibanaRole.indices().allowedIndicesMatcher(UpdateSettingsAction.NAME).test(mockIndexAbstraction(index)), is(true));
+            assertThat(kibanaRole.indices().allowedIndicesMatcher(PutMappingAction.NAME).test(mockIndexAbstraction(index)), is(true));
+            assertThat(kibanaRole.indices().allowedIndicesMatcher(RolloverAction.NAME).test(mockIndexAbstraction(index)), is(true));
+            // Privileges needed for installing current ILM policy with delete action
+            assertThat(kibanaRole.indices().allowedIndicesMatcher(DeleteIndexAction.NAME).test(mockIndexAbstraction(index)), is(true));
```
gwbrown, line 702: These tests currently fail; you'll need to add some privileges in addition to `read` for these to work. My recommendation would be `manage`; we might be able to find a more minimal set of privileges that allows these actions but it seems as though we do want the Kibana system user to be able to manage those indices (just not write data to them), so `manage` makes sense semantically.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/authz/store/ReservedRolesStoreTests.java
```diff
@@ -681,6 +681,27 @@ public void testKibanaSystemRole() {
             assertThat(kibanaRole.indices().allowedIndicesMatcher(DeleteIndexAction.NAME).test(mockIndexAbstraction(index)), is(true));
         });
 
+        // read-only indices for Endpoint events (to build timelines)
+        Arrays.asList(".logs-endpoint.events.process-default-" + randomAlphaOfLength(randomIntBetween(0, 13))).forEach((index) -> {
+            assertThat(kibanaRole.indices().allowedIndicesMatcher("indices:foo").test(mockIndexAbstraction(index)), is(false));
+            assertThat(kibanaRole.indices().allowedIndicesMatcher("indices:bar").test(mockIndexAbstraction(index)), is(false));
+            assertThat(kibanaRole.indices().allowedIndicesMatcher(GetIndexAction.NAME).test(mockIndexAbstraction(index)), is(true));
+            assertThat(kibanaRole.indices().allowedIndicesMatcher(CreateIndexAction.NAME).test(mockIndexAbstraction(index)), is(false));
```
gwbrown, line 689: I'm not sure we can assert this if we need the Kibana system user to be able to roll over the index as tested below. Historically, rollover internally called `CreateIndexAction` so the ability to create the next index was an inherent requirement to be able to use rollover. That's not the case anymore and it would be technically possible to distinguish the two today, but we haven't (yet) added a privilege that just allows rollover and disallows other index creation.

My recommendation would be to drop this line from the test and use `manage` as I describe in my other comment, but I'm less familiar with the contents of these indices and any particular requirements we have around them.

ywangd:
Ping @elastic/kibana-security

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/authz/store/ReservedRolesStoreTests.java
pjhampton: Thanks, Gordon. 
`read` permissions will suffice for this use case.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/authz/store/ReservedRolesStoreTests.java
pjhampton: Got it, thanks!

pjhampton:
For the Kibana Security gang, this functionality is required for the backref'd issue in Security.

The Kibana backend builds telemetry records periodically and transmits them back to us to improve our Endpoint and SIEM protections. The current functionality we are hooking into uses the ES Client `asCurrentUser` functionality, to make sure the user logged into Kibana can themselves read from this index. We need to use `asInternalUser` and plan to program to the interface to make this work. This PR gives the `kibana_system` user access to the endpoint streams, so I can read it out and build the graph.

For any other details feel free to reach out to us in our slack channel regarding the need / scope of this.

azasypkin (approved):
> For the Kibana Security gang, this functionality is required for the backref'd issue in Security.
> The Kibana backend builds telemetry records periodically and transmits them back to us to improve our Endpoint and SIEM protections. The current functionality we are hooking into uses the ES Client asCurrentUser functionality, to make sure the user logged into Kibana can themselves read from this index. We need to use asInternalUser and plan to program to the interface to make this work. This PR gives the kibana_system user access to the endpoint streams, so I can read it out and build the graph. We wouldn't let the user access their cluster data via the internal user.

Sounds good to me from the Kibana Security perspective. Thanks for the detailed explanation about how it's being used in Kibana! I wish everyone who's adding new privileges to `kibana_system` did this :slightly_smiling_face:

PR #86703, opened 2022-05-11
Port QA projects to use javaRestTest and yamlRestTest

breskeby:
Do not use deprecated elasticsearch.rest-test plugin


qa/ccs-unavailable-clusters/build.gradle
```diff
@@ -5,16 +5,13 @@
...
  * Side Public License, v 1.
  */
-apply plugin: 'elasticsearch.internal-testclusters'
-apply plugin: 'elasticsearch.standalone-rest-test'
-apply plugin: 'elasticsearch.rest-test'
-apply plugin: 'elasticsearch.test-with-dependencies'
+apply plugin: 'elasticsearch.internal-java-rest-test'
 
-testClusters.matching { it.name == "integTest" }.configureEach {
+testClusters.matching { it.name == "javaRestTest" }.configureEach {
```
mark-vieira, line 10: Not sure why we used`matching` here when there's only ever one. Wouldn't `testClusters.named()` be more concise?

qa/system-indices/build.gradle
```diff
@@ -23,3 +23,5 @@ testClusters.configureEach {
   setting 'xpack.security.autoconfiguration.enabled', 'false'
   user username: 'rest_user', password: 'rest-user-password'
 }
+
+tasks.named('test').configure { enabled = false}
```
mark-vieira, line 27: Why is this necessary?

mark-vieira (approved):
LGTM

qa/ccs-unavailable-clusters/build.gradle
breskeby: this is because we create those cluster lazy as part of the task creation. That means that taskClusters.named would not find the cluster yet when hitting this config block initially

qa/system-indices/build.gradle
breskeby: there were no tests we did this in other places. But you're right we should just rely on gradle skipping those tasks anyhow.

breskeby: removed

PR #86704, opened 2022-05-11
Reindex support for TSDB creating indices

nik9000:
This teaches reindex to infer if the destination index is a tsdb index
if it doesn't yet exist. It looks at the templates that apply to the
index and checks what settings they use. So you should be able to
reindex into a non-existant index and automatically get the tsdb
support.

### DIFF
--- a/modules/reindex/src/test/java/org/elasticsearch/reindex/ReindexIdTests.java
+++ b/modules/reindex/src/test/java/org/elasticsearch/reindex/ReindexIdTests.java
@@ -31,7 +31,7 @@ import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.nullValue;
 
 /**
- * Reindex
+ * Reindex tests for picking ids.
  */
 public class ReindexIdTests extends AbstractAsyncBulkByScrollActionTestCase<ReindexRequest, BulkByScrollResponse> {
     public void testEmptyStateCopiesId() throws Exception {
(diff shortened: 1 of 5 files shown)


modules/reindex/src/test/java/org/elasticsearch/reindex/ReindexIdTests.java
```diff
@@ -53,6 +70,16 @@ private ClusterState stateWithIndex(Settings.Builder settings) {
         return ClusterState.builder(ClusterState.EMPTY_STATE).metadata(Metadata.builder(Metadata.EMPTY_METADATA).put(meta)).build();
     }
 
+    private Settings.Builder standardSettings() {
+        return Settings.builder().put(IndexSettings.MODE.getKey(), IndexMode.STANDARD);
```
martijnvg, line 74: maybe randomly return an empty setting? Since standard is the default.

martijnvg (approved):
👍 LGTM

modules/reindex/src/test/java/org/elasticsearch/reindex/ReindexIdTests.java
nik9000: :+1:

modules/reindex/src/test/java/org/elasticsearch/reindex/ReindexIdTests.java
```diff
@@ -34,15 +39,27 @@ public void testEmptyStateCopiesId() throws Exception {
...
+        metadata.put("c", new ComponentTemplate(new Template(standardSettings().build(), null, null), null, null));
+        metadata.put("c", new ComposableIndexTemplate(List.of("dest_index"), null, List.of("c"), null, null, null));
+        ClusterState state = ClusterState.builder(ClusterState.EMPTY_STATE).metadata(metadata).build();
+        assertThat(action(state).buildRequest(doc()).getId(), equalTo(doc().getId()));
+    }
+
+    public void testMissingIndexWithTsdbTemplateClearsId() throws Exception {
+        Metadata.Builder metadata = Metadata.builder();
+        metadata.put("c", new ComponentTemplate(new Template(tsdbSettings().build(), null, null), null, null));
+        metadata.put("c", new ComposableIndexTemplate(List.of("dest_index"), null, List.of("c"), null, null, null));
```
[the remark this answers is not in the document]
nik9000: :+1:

PR #86706, opened 2022-05-11
Add annotation for disabling security manager in tests

rjernst:
The Elasticsearch test framework initializes a test security manager, so
that tests run in a similar environment to that which we run in
production. However, some tests need to run without security manager
because they do "evil" things, like setting adding and removing jvm wide
shutdown hooks. Currently these tests must exist in separate projects,
mostly under qa/evil-tests, so that the security manager can be disabled
for the entire jvm with the tests.security.manager system property. The
separation between these tests and the other tests for the server makes
development of these tests more difficult, especially the inability to run
these within IntelliJ without manually adding additional sysprops.

This commit adds a new mechanism to disable security manager for these
tests. A new `@NoSecurityManager` annotation can be added to a test
suite. ESTestCase detects this annotation and removes the security
manager for the duration of that test suite. We could potentially allow
this per test as well, but I started with the simple suite wide
strategy. One example test is converted. The rest will be converted in
followups.

### DIFF
--- a/qa/evil-tests/src/test/java/org/elasticsearch/bootstrap/EvilBootstrapChecksTests.java
+++ b/qa/evil-tests/src/test/java/org/elasticsearch/bootstrap/EvilBootstrapChecksTests.java
@@ -12,7 +12,6 @@ import org.apache.logging.log4j.Logger;
 import org.elasticsearch.core.SuppressForbidden;
 import org.elasticsearch.node.NodeValidationException;
 import org.elasticsearch.test.AbstractBootstrapCheckTestCase;
-import org.elasticsearch.test.ESTestCase;
 import org.elasticsearch.test.ESTestCase.WithoutSecurityManager;
 import org.hamcrest.Matcher;
 import org.junit.After;
--- a/qa/evil-tests/src/test/java/org/elasticsearch/bootstrap/EvilBootstrapChecksTests.java
+++ b/qa/evil-tests/src/test/java/org/elasticsearch/bootstrap/EvilBootstrapChecksTests.java
@@ -29,7 +29,7 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
 
-@ESTestCase.NoSecurityManager
+@ESTestCase.WithoutSecurityManager
 public class EvilBootstrapChecksTests extends AbstractBootstrapCheckTestCase {
 
     private String esEnforceBootstrapChecks = System.getProperty(ES_ENFORCE_BOOTSTRAP_CHECKS);
--- a/qa/evil-tests/src/test/java/org/elasticsearch/bootstrap/EvilBootstrapChecksTests.java
+++ b/qa/evil-tests/src/test/java/org/elasticsearch/bootstrap/EvilBootstrapChecksTests.java
@@ -12,6 +12,7 @@ import org.apache.logging.log4j.Logger;
 import org.elasticsearch.core.SuppressForbidden;
 import org.elasticsearch.node.NodeValidationException;
 import org.elasticsearch.test.AbstractBootstrapCheckTestCase;
+import org.elasticsearch.test.ESTestCase;
 import org.hamcrest.Matcher;
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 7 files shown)


qa/evil-tests/src/test/java/org/elasticsearch/bootstrap/EvilBootstrapChecksTests.java
```diff
@@ -28,6 +29,7 @@
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
 
+@ESTestCase.NoSecurityManager
```
mark-vieira, line 32: Why don't we import the annotation itself here so this isn't so verbose?

test/framework/src/main/java/org/elasticsearch/test/ESTestCase.java
```diff
@@ -349,6 +355,34 @@ protected void afterIfFailed(List<Throwable> errors) {}
     /** called after a test is finished, but only if successful */
     protected void afterIfSuccessful() throws Exception {}
 
+    /**
+     * Marks a test suite or a test method that should run without security manager enabled.
+     */
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target({ ElementType.TYPE })
+    @Inherited
+    public @interface NoSecurityManager {
```
mark-vieira, line 364: I wonder if we should place this on `ESClientYamlSuiteTestCase` since there's not point in _any_ yaml rest tests using the security manager. You could probably make an argument for java rest tests as well but those are a bit more of the wild west.

mark-vieira (approved):

test/framework/src/main/java/org/elasticsearch/test/ESTestCase.java
rjernst: Are ESClientYamlSuiteTestCases ever runnable in the IDE? I think it's likely better to just decouple those from ESTestCase, since as you say, they don't ever need to run with security manager.

qa/evil-tests/src/test/java/org/elasticsearch/bootstrap/EvilBootstrapChecksTests.java
rjernst: Just saves an extra import :)

test/framework/src/main/java/org/elasticsearch/test/ESTestCase.java
breskeby: invoked via the gradle runner, those tests should run from the IDE

ChrisHegarty (approved):
Nice!

test/framework/src/main/java/org/elasticsearch/test/ESTestCase.java
```diff
@@ -349,6 +355,34 @@ protected void afterIfFailed(List<Throwable> errors) {}
...
+    /**
+     * Marks a test suite or a test method that should run without security manager enabled.
+     */
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target({ ElementType.TYPE })
+    @Inherited
+    public @interface WithoutSecurityManager {
+    }
+
+    private static Closeable securityManagerRestorer;
```
ChrisHegarty, line 367: I'm not exactly sure where these before/after methods run, but we might want to make this field volatile, or use a lock to ensure visibility?

mark-vieira: I'm quite certain all test lifecycle methods execute within a single thread, and we don't perform any intra-JVM parallelization of test execution (parallel classes run in individual forked JVMs) so I don't think this is a concern.

rjernst: Yes that is right, all before*/after* methods are run in the same test thread, as if they were part of each test method execution. You can see things set in other tests by before/after do not concern themselves with threads.

test/framework/src/main/java/org/elasticsearch/test/ESTestCase.java
rjernst: I will look at what classes can utilize this annotation more in followups, I'd like to keep this PR as small as possible just for the infrastructure.

Issue #82596, opened 2022-01-14
Incorrect response when manually creating security indices


ywangd:
In ES 8.0, when creating the security token index manually with
```
PUT .security-tokens-42
```

We get the following response:
```
{
  "acknowledged": true,
  "shards_acknowledged": true,
  "index": ".security-tokens-42"
}
```

The response says `.security-tokens-42` is created. But that is not true and ES actually created `.security-tokens-7` instead. This happens because the requested index name matches name pattern of the security token system index. The system index framwork  thus intercepts the request and makes sure the index creation follows what is defined by the system index descriptor. 

Though it is arguable that the request interception and the index creation are desired behaviours, the response is incorrect and potentially misleading. The example uses security token index. But I suspect the same issue can happen for any system indices that claims index name namespaces with wildcards.

ywangd:
Similarly, if the security token index alreay exists, you get the following error when trying to create `.security-tokens-42`. I think it is not ideal either.

```
PUT .security-tokens-42

===
{
  "error": {
    "root_cause": [
      {
        "type": "resource_already_exists_exception",
        "reason": "index [.security-tokens-7/vsuxZJTkTUu3jVsCfkVGxQ] already exists",
        "index_uuid": "vsuxZJTkTUu3jVsCfkVGxQ",
        "index": ".security-tokens-7"
      }
    ],
    "type": "resource_already_exists_exception",
    "reason": "index [.security-tokens-7/vsuxZJTkTUu3jVsCfkVGxQ] already exists",
    "index_uuid": "vsuxZJTkTUu3jVsCfkVGxQ",
    "index": ".security-tokens-7"
  },
  "status": 400
}
```

gwbrown:
We discussed this in the Core/Infra sync yesterday.

Part of this behavior is that when a System Index is registered, it can (sometimes must) register a "primary index", and if any index creation request comes in for that system index's pattern (`.security-tokens-[0-9]+*`), it gets replaced with the primary index (in this case, `.security-tokens-7`).

This behavior seems counterintuitive to me and I don't recall why it's this way - maybe @pugnascotia can fill in some of the background (if he remembers)?

Anyway, assuming this is the behavior we want, if we see a request trying to create an index in the pattern other than the primary index, we should throw an exception rather than silently replace it. If for some reason that's a bad idea, we should make sure the response is correct.

I'm not sure exactly when we'll have someone with bandwidth to address this, we've got some other system index related issues we're working through.

[...]

PR #86707, opened 2022-05-11
Error on direct creation of non-primary system index

williamrandolph:
Currently, a "create index" request for a managed system index will always try to create the primary index, rather than the index contained in the request. The result is not something that users expect (see for example, https://github.com/elastic/elasticsearch/issues/82596).

It probably makes more sense to throw an error in this case.


grcevski (approved):
LGTM! Should we document this change anywhere? It's essentially a behaviour change even though it's very likely an uncommon usage.

AthenaEryma (approved):
LGTM as well, thanks for tracking this down!

williamrandolph:
@grcevski This ought to get listed as a bugfix in the 8.3 release notes. As far as I know, we don't have any documentation of the existing behavior because our plan has been to completely prevent users from doing this in the REST API. Judging by what I've seen in tests and bug reports, everyone finds the previous behavior very strange. I'm hoping to get this behavior written down in Javadoc and in internal docs in the next week and a half.

PR #86709, opened 2022-05-12
Use pipes for spawned controllers

rjernst:
Controller processes currently inherit stdout/stderr streams from
the main Elasticsearch process. This isn't horrible when we run in the
foreground, but when we daemonize, the open pipe causes the underlying
file descriptor to stay open, even though we have closed the stream
within Elasticsearch.

This commit changes the controller spawner to pipe stdout/stderr, and
creates pump threads.

relates #85758

### DIFF
--- a/server/src/main/java/org/elasticsearch/bootstrap/Spawner.java
+++ b/server/src/main/java/org/elasticsearch/bootstrap/Spawner.java
@@ -50,9 +50,7 @@ final class Spawner implements Closeable {
         closeables.addAll(pumpThreads.stream().map(t -> (Closeable) () -> {
             try {
                 t.join(); // wait for thread to complete now that the spawned process is destroyed
-            } catch (InterruptedException e) {
-                throw new AssertionError(e); // we don't interrupt these threads
-            }
+            } catch (InterruptedException e) { /* best effort, ignore*/ }
         }).toList());
         IOUtils.close(closeables);
     }
@@ -110,7 +108,7 @@ final class Spawner implements Closeable {
                     logger.warn(line);
                 }
             } catch (IOException e) {
-                logger.error("Error while pumping " + streamName + " for " + componentName + " controller", e);
+                logger.error(e);
             }
         }, loggerName + "-pump");
         t.start();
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/elasticsearch/bootstrap/Spawner.java
```diff
@@ -34,22 +40,28 @@ final class Spawner implements Closeable {
...
 
     @Override
     public void close() throws IOException {
-        IOUtils.close(() -> processes.stream().map(s -> (Closeable) s::destroy).iterator());
+        List<Closeable> closeables = new ArrayList<>();
+        closeables.addAll(processes.stream().map(s -> (Closeable) s::destroy).toList());
+        closeables.addAll(pumpThreads.stream().map(t -> (Closeable) () -> {
+            try {
+                t.join(); // wait for thread to complete now that the spawned process is destroyed
+            } catch (InterruptedException e) { /* best effort, ignore*/ }
```
droberts195, line 53: I think it would be best practice to call `Thread.currentThread().interrupt();` in the `catch` block here, because if the interruption occurred when joining the first thread we wouldn't want the join of the second thread to block.

server/src/main/java/org/elasticsearch/bootstrap/Spawner.java
```diff
@@ -75,16 +87,39 @@ void spawnNativeControllers(final Environment environment, final boolean inherit
...
+        final Logger logger = LogManager.getLogger(loggerName);
+        Thread t = new Thread(() -> {
+            try (var br = new BufferedReader(new InputStreamReader(stream, StandardCharsets.UTF_8))) {
+                String line;
+                while ((line = br.readLine()) != null) {
+                    // since we do not expect native controllers to ever write to stdout/stderr, we always log at warn level
+                    logger.warn(line);
+                }
+            } catch (IOException e) {
+                logger.error("error while reading " + streamName, e);
```
droberts195, line 111:
```suggestion
                logger.error("error while reading " + loggerName, e);
```

Otherwise just "error while reading stderr" might be a bit unclear about which stderr to the average person reading the log file.

ChrisHegarty (approved):
LGTM.

droberts195 (approved):
LGTM apart from a couple of minor suggestions.

Thanks for fixing this.

grcevski (approved):
LGTM!

server/src/main/java/org/elasticsearch/bootstrap/Spawner.java
rjernst: The logger name itself is descriptive. I only added the message at all because there is no error method which takes only a throwable.

server/src/main/java/org/elasticsearch/bootstrap/Spawner.java
rjernst: Fair enough, I pushed [f68b57f](https://github.com/elastic/elasticsearch/pull/86709/commits/f68b57faa341dc622eaafcd30fd64e602499ceeb)

PR #86710, opened 2022-05-12
Upgrade procrun executables to 1.3.1

rjernst:
This commit updates the procrun manager and service exes to 1.3.1. There
are a few bugfixes, but most notably there was an unfiled bug that seems
to be fixed where stderr could get gigabytes of null characters in at
least one case.

relates #85758


williamrandolph:
It looks like our tests assume that the `remove` command will succeed as a no-op if there's no service to remove, but that this is no longer true in `1.3.1`.

ChrisHegarty (approved):

mark-vieira (commented):
I notice that we don't have any license or notice information for procrun anywhere in the project. This looks to be an oversight, mostly because we're not consuming this as a external binary dependency and instead check these into source control.

rjernst:
Re license/notice: I opened https://github.com/elastic/elasticsearch/pull/86773

rjernst:
@elasticmachine run elasticsearch-ci/rest-compatibility

PR #86715, opened 2022-05-12
Remove remaining single arg ParameterizedMessages

rjernst:
This commit removes the remaining ParameterizedMessages that take a
single argument, this time where the argument contains method calls.
This was again done almost entirely through find/replace with regex in
IntelliJ.

relates #86549


pgomulka (approved):
LGTM

PR #86718, opened 2022-05-12
Move DataStreamMetadata update code into DatastreamMetadata class

original-brownbear:
Follow up to #86673 hiding the immutable open map implementation details from the public interface
of this class.


martijnvg (approved):
Thanks for the quick follow up! LGTM when CI is happy.

PR #86727, opened 2022-05-12
[Transform] improve CCS test to cover the previous minor version

hendrikmuhs:
add testing against the previous minor version


davidkyle (approved):
LGTM

PR #86728, opened 2022-05-12
Make user and role name constraint consistent with max document ID.

### LINKED ISSUE #66020: Max length constaint for role/user name should be more consistent
The security related [validation](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/support/Validation.java#L22) mandates a maximum length of 1024 characters for username and role name. This limitation is also documented in following places:
* [Defining roles](https://www.elastic.co/guide/en/elasticsearch/reference/master/defining-roles.html)
* [Put user API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-put-user.html#security-api-put-user-path-params)
* [`elasticsearch-users` command](https://www.elastic.co/guide/en/elasticsearch/reference/current/users-command.html#_description_10)*

However, when the user and role are stored in the security index, their names are used to form the document ID, for which the indexing code has a max length [constraint](https://github.com/elastic/elasticsearch/blob/20d6fbcd551cc9b9cf1dbb45fac062b101dfa4a8/server/src/main/java/org/elasticsearch/action/index/IndexRequest.java#L215-L218) of 512 bytes. And this 512 bytes constraint is what takes the ultimate effect. Also because the document IDs are prefixed by either `user-` or `role-`, the actual possible max length is a bit less than 512 bytes.

When the users/roles are defined in files, they are no longer limited by the document ID and in this case, they can be up to 1024 characters. So this means the above documentation for `elasticsearch-users` command is accurate. However, I wonder whether this difference between file and index is necessary or intended.

PS: role mapping has the same issue with document ID. But security validation does not touch it (probably because the length is enforced at HTTP layer as request URL). Also we don't have documentation about its length limit. So it is less of an issue.

slobodanadamovic:
This change tries to make user and role name constraints consistent with
what is currently allowed to store in the native realm (512).
Because the document IDs are prefixed by either `user-` or `role-`,
the actual possible max length is a funky looking 507 chars.
If we choose (in the future) to allow more than 507 chars we should
either consider increasing the max allowed size for document ID or
consider hashing names longer than 507 in order to fit into document ID.

Note: File realm validation is left as it was and allows max 1024 chars. 
I'm not sure if this is something we should reflect in the documentation.

Closes #66020


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/support/Validation.java
```diff
@@ -175,18 +178,23 @@ public static final class Users {
          * @param username the username to validate
          * @param allowReserved whether or not to allow reserved user names
          * @param settings the settings which may contain information about reserved users
+         * @param maxLength the maximal allowed length a username can have
          * @return {@code null} if valid
          */
-        public static Error validateUsername(String username, boolean allowReserved, Settings settings) {
-            if (isValidUserOrRoleName(username) == false) {
-                return new Error(String.format(Locale.ROOT, INVALID_NAME_MESSAGE, "User"));
+        public static Error validateUsername(String username, boolean allowReserved, Settings settings, int maxLength) {
```
ywangd, line 184: This works. But I'd prefer hiding the `maxLength` parameter like how you did with the `FileRealmValidationUtil`. The length parameter is really an internal thing. We end up having 2 possible values due to mainly historical reasons, not because it is truly designed to accept multiple different length values. Therefore more specialised classes or methods seem to be a better fit instead of exposing the length as an argument for a public method. For an example, maybe the method can be private (or package private) and have a new public method `validateNativeUsername` which does not need the maxLength parameter. 

This likely implies more changes like relocating `FileRealmValidationUtil` so it can access the private (or package private) method. Since it's not critical, maybe we can let it pass in this PR and get back to it in future when we need to do something about roleMapping name validation (we don't have any today, but will likley need it). I'll let you make the call. Thanks!

ywangd (approved):
LGTM

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/SetEnabledRequest.java
```diff
@@ -40,9 +40,13 @@ public SetEnabledRequest(StreamInput in) throws IOException {
...
-        if (error != null) {
-            validationException = addValidationError(error.toString(), validationException);
+        if (username == null) {
+            validationException = addValidationError("username must be set", validationException);
+        } else {
+            Error error = Validation.Users.validateUsername(username, true, Settings.EMPTY);
+            if (error != null) {
+                validationException = addValidationError(error.toString(), validationException);
+            }
         }
```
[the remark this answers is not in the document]
slobodanadamovic: Good question. I had same thought when I was checking `PutUserRequest` and `TransportPutUserAction`. I was confused that `null` check and validation were split. I thought that there was a good reason for this. Now that I look at it again, I don't see a reason why not. It would make `PutUserRequest` and `SetEnabledRequest` more consistent.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/support/Validation.java
```diff
@@ -17,10 +17,11 @@
 
 public final class Validation {
 
-    static final int MIN_NAME_LENGTH = 1;
-    static final int MAX_NAME_LENGTH = 1024;
+    public static final int MIN_NAME_LENGTH = 1;
 
-    static final Set<Character> VALID_NAME_CHARS = Set.of(
+    public static final int DEFAULT_MAX_NAME_LENGTH = 507;
```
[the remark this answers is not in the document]
slobodanadamovic: Sure. I will add it.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/support/Validation.java
```diff
@@ -132,8 +131,12 @@ public final class Validation {
         + "and at most 256 characters that are alphanumeric (A-Z, a-z, 0-9) or hyphen (-) or underscore (_). "
         + "It must not begin with an underscore (_).";
 
-    private static boolean isValidUserOrRoleName(String name) {
-        if (name.length() < MIN_NAME_LENGTH || name.length() > MAX_NAME_LENGTH) {
+    private static boolean isValidUserOrRoleName(String name, int maxLength) {
+        if (maxLength < MIN_NAME_LENGTH) {
+            throw new IllegalArgumentException("maxLength cannot be less than " + MIN_NAME_LENGTH);
+        }
```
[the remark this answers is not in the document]
slobodanadamovic: Makes sense. I will change it to `assert`.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/support/Validation.java
slobodanadamovic: It's a valid point. I constrained myself (for no particular reason) not to make changes which affect native usages of this method. I'll adjust it since it will make the code cleaner and avoid misusage.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/support/Validation.java
```diff
@@ -175,18 +178,23 @@ public static final class Users {
...
          * @return {@code null} if valid
          */
-        public static Error validateUsername(String username, boolean allowReserved, Settings settings) {
-            if (isValidUserOrRoleName(username) == false) {
-                return new Error(String.format(Locale.ROOT, INVALID_NAME_MESSAGE, "User"));
+        public static Error validateUsername(String username, boolean allowReserved, Settings settings, int maxLength) {
+            if (isValidUserOrRoleName(username, maxLength) == false) {
+                return new Error(String.format(Locale.ROOT, INVALID_NAME_MESSAGE, "User", maxLength));
             }
             if (allowReserved == false && ClientReservedRealm.isReserved(username, settings)) {
```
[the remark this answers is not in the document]
slobodanadamovic: I see your point. I will think on this and come up with a proposal in a separate PR.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/RoleDescriptor.java
```diff
@@ -300,7 +301,7 @@ public static RoleDescriptor parse(String name, BytesReference source, boolean a
 
     public static RoleDescriptor parse(String name, XContentParser parser, boolean allow2xFormat) throws IOException {
         // validate name
-        Validation.Error validationError = Validation.Roles.validateRoleName(name, true);
+        Validation.Error validationError = NativeRealmValidationUtil.validateRoleName(name, true);
```
slobodanadamovic, line 304: @ywangd I made some changes to address review comments and came across an issue with validation in `RoleDescriptor.parse` method. This method is called from `FileRolesStore, `NativeRolesStore, `ApiKeyService` and `PutRoleRequestBuilder`. This means that using here the "native" validator is wrong.
Since this lowers my confidence in this change, I will mark this PR as draft until I fix this issue and write some more tests for role name validation.

ywangd: @slobodanadamovic This is a good find. I completely missed it when reviewing it previously. My thinking is that we should keep the scope of changes in this PR to the original intention, i.e. fix the role name validation for roles stored in the security index. Based on this, I'd suggest we leave all the current usages of `validateRoleName` and only add the new validation to `PutRoleRequest` and maybe also `NativeRoleStore#putRole`. For any other places, you can come up with a theorectical scenario where it could be a "breaking" change to alter the validation. For example, `ApiKeyService` can take role descriptors with long names because they are not used for document ID, or `TransportPutUserAction` can also take long role names because these roles can from file. Happy to sync if it helps sorting out the issue.

slobodanadamovic: > Happy to sync if it helps sorting out the issue.
@ywangd That would help. Let me know what time would be convenient for you.

slobodanadamovic:
Tests failing due to https://github.com/elastic/elasticsearch/issues/86877.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/support/ValidationTests.java
```diff
@@ -133,38 +133,40 @@ public class ValidationTests extends ESTestCase {
...
     );
 
+    private static final int MAX_NAME_LENGTH = Validation.MAX_NAME_LENGTH;
+
     public void testUsernameValid() throws Exception {
-        int length = randomIntBetween(Validation.MIN_NAME_LENGTH, Validation.MAX_NAME_LENGTH);
+        int length = randomIntBetween(Validation.MIN_NAME_LENGTH, MAX_NAME_LENGTH);
         String name = new String(generateValidName(length));
-        assertThat(Users.validateUsername(name, false, Settings.EMPTY), nullValue());
+        assertThat(Users.validateUsername(name, false, Settings.EMPTY, MAX_NAME_LENGTH), nullValue());
```
ywangd, line 141: Do we still have a test that is for this method signature?
```
Users.validateUsername(String, boolean, Settings)
```
If not, I think we should have one.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/support/ValidationTests.java
```diff
@@ -178,40 +180,40 @@ public void testUsersValidatePassword() throws Exception {
     }
 
     public void testRoleNameValid() throws Exception {
-        int length = randomIntBetween(Validation.MIN_NAME_LENGTH, Validation.MAX_NAME_LENGTH);
+        int length = randomIntBetween(Validation.MIN_NAME_LENGTH, MAX_NAME_LENGTH);
         String name = new String(generateValidName(length));
-        assertThat(Roles.validateRoleName(name), nullValue());
+        assertThat(Roles.validateRoleName(name, false, MAX_NAME_LENGTH), nullValue());
```
ywangd, line 185: Similarly, do we have a test for 
```
Roles.validateRoleName(String, boolean)
```

ywangd (approved):
LGTM

Thanks for being thorough.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/support/ValidationTests.java
slobodanadamovic: We don't. I'm happy to add them.

PR #86729, opened 2022-05-12
[8.2][Transform] workaound exception on CCS

hendrikmuhs:
implement a workaound for security exception caught when CCS 8.2 talks to a 8.1 node.


hendrikmuhs:
closing as the workaround does not work. I will open a new backport PR for the system test

Issue #68743, opened 2021-02-09
Prevent md5 from being used for cache hashing when in FIPS mode


ywangd:
SP 800-52 approves that MD5 can be used with TLS and it is the only scenario where it is allowed. However, the BCFIPS library (and maybe other libraries) does not prevent it from being used in other places (there really is no reliable way to detect it). In our case, user can choose to use it as the caching algorithm. The usage includes [CachingUsernamePasswordRealm](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/support/CachingUsernamePasswordRealm.java), [ApiKeyService](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java) and maybe some other places. It also can be used in the [fingerprint ingest processor](https://github.com/elastic/elasticsearch/pull/68415). We should fix them by either throwing error or fail to start.

In all the aforementioned places we should be checking the value of `XPackSettings.FIPS_MODE_ENABLED` in the current loaded settings ( i.e. by checking `Environment#settings` ) and throw an exception that would cause the component and the node to exit or print a warning in the logs if we want to be more tolerant.

n1v0lg:
As discussed on slack, [FingerprintProcessor](https://github.com/elastic/elasticsearch/blob/master/modules/ingest-common/src/main/java/org/elasticsearch/ingest/common/FingerprintProcessor.java) does not use the hash function in a security context, i.e., does not require a cryptographically secure hash function. Instead the hash function is used for content fingerprinting. As such this is outside of scope for FIPS and it does not make sense to include a warning here.

PR #86740, opened 2022-05-12
Log warning when cache hasher not compliant with FIPS

n1v0lg:
Currently, it's possible to choose a hash function for various cache
hashers (e.g., in `ApiKeyService`) that is not compliant with FIPS 140
(e.g., MD5). This PR logs a warning on node start if a non-compliant
hashing algorithm is used in FIPS mode.

Note that there are other usages of non-FIPS compliant hash functions,
which are not configured through settings (e.g.
`FingerprintProcessor`). I plan to address these in a separate PR.

Relates #68743


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/Security.java
```diff
@@ -1419,6 +1419,22 @@ static void validateForFips(Settings settings) {
...
+        cacheHashAlgoSettings.keySet().forEach((key) -> {
+            final var setting = cacheHashAlgoSettings.get(key);
+            assert setting != null;
+            final var hashAlgoName = setting.toLowerCase(Locale.ROOT);
+            if (hashAlgoName.equals("ssha256") == false
+                && hashAlgoName.equals("sha1") == false
+                && hashAlgoName.startsWith("pbkdf2") == false) {
+                logger.warn(
+                    "Only SSHA256, SHA1, or PBKDF2 are allowed for secure hashing in a FIPS 140 JVM. "
+                        + "Please change the [{}] setting from [{}] to an appropriate value.",
```
ywangd, line 1432: I have a few comments on this message:
1. Since this is a warning, I'd tone down it a bit. So instead of saying ".. are allowed ...", I suggest we change it to "...  are recommended ..." so it does _not_ read as strong as other messages that would cause node start failure.
2. I don't think we should recommend SHA1. Strictly speaking we should not use unsalted hash for password like information. In practice, it's probably not too bad since cache is relatively short-lived (24 hours for API keys). But as a general message, it's better to avoid it.
3. I don't it's wise to use PBKDF2 for cache hashing because of performance. But it is not wrong. So I am OK with it.

ywangd (commented):
Haven't look into the tests yet. But have a few suggestions around the warning message.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/Security.java
n1v0lg: Also wondering if we should go with something like "in-memory credential hashing" instead of "secure hashing" if we kick out SHA1. Wydt?

ywangd: I am no FIPS expert but I don't think it explicitly recommends any password hashing algorithms. PBKDF2 is the "semi"-standard because it is a clear FIPS choice for key derivation which happens to work like a hash. So there is some wiggle room here. But there are published attacks for SHA-1. So I find it hard to recommend it for password hashing, even when it is in memory.

> Also wondering if we should go with something like "in-memory credential hashing" instead of "secure hashing" for the message
This is a good point. I think we can be more explicit here because people may even question ssha256 without more context.

n1v0lg:
Build failures not related to this PR.

n1v0lg:
@elasticmachine update branch

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/Security.java
```diff
@@ -1419,6 +1419,20 @@ static void validateForFips(Settings settings) {
                     + " ] setting."
             );
         }
+        final var cacheHashAlgoSettings = settings.filter(k -> k.endsWith("cache.hash_algo"));
```
ywangd, line 1422: Nit: does it make sense to have a bit more strick check?
```suggestion
        final var cacheHashAlgoSettings = settings.filter(k -> k.endsWith(".cache.hash_algo"));
```

We have cache called `doc_cache`, but it does not have a `hash_algro` field. So it is not really an issue. But I wonder whether stricter is better?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/Security.java
```diff
@@ -1419,6 +1419,20 @@ static void validateForFips(Settings settings) {
...
+        cacheHashAlgoSettings.keySet().forEach((key) -> {
+            final var setting = cacheHashAlgoSettings.get(key);
+            assert setting != null;
+            final var hashAlgoName = setting.toLowerCase(Locale.ROOT);
+            if (hashAlgoName.equals("ssha256") == false && hashAlgoName.startsWith("pbkdf2") == false) {
+                logger.warn(
+                    "Only SSHA256 or PBKDF2 is recommended for in-memory credential hashing in a FIPS 140 JVM. "
+                        + "Please change the [{}] setting from [{}] to an appropriate value.",
+                    key,
+                    hashAlgoName
```
ywangd, line 1432: Nit:
```suggestion
                    setting
```
to better preserve what user exactly configures?

ywangd (approved):
LGTM

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/Security.java
n1v0lg: Yup, good point!

ywangd:
I think we should amend the doc. The wording is likely tricky to justify `ssha256` is sufficient for in-memory hashing in FIPS mode. But I find it hard to recommend `pbkdf2` as `cache.hash_algo`. Though technically correct, this recommendation is set users to a likely path of failure. Especially with API keys, the secret is already of 20 byte length. Default pbkdf2 takes about 70 ms to compute. This means a single core can only authenticate 15 requests per second. It will not work for a deployment with something like Fleet. 

Taking a step back, my original intention was to log a warning message of "xxx is **not** recommended" which intentionaly avoid having to mention `ssha256` explicitly (gray area as I see it). But negative message is not very actionable (because users do not know what would be recommended). So I accepted that a positive message is better, but I wasn't aware our docs recommend pbkdf2 for cache hasing. Now I think we need retrace our decisions. My current preference would be (1) *not* recommend pbkdf2 for performance reasons or recommend it but also explain the performance implication, and (2) *not* warn againet ssha256. But let's bring it to the team for broader discussion. Thanks!

n1v0lg:
> This means a single core can only authenticate 15 requests per second.

This is indeed not realistic. 

> But negative message is not very actionable

The steps that users would likely take is either ignore the warning altogether, read our/NIST's docs, try out options blindly until they hit the right one, or reach out to support. We'd essentially still need to make a recommendation. It also gets complicated with, for example, the NOOP hasher. 

I'm with you on your preferences regarding docs and allowing `ssha256`.

As for interpreting NIST recommendations, I found two applicable resources so far:

1. [800-63B Digital Identity Guidelines](https://pages.nist.gov/800-63-3/sp800-63b.html#sec5)
1. [800-132 Recommendation for Password-Based Key
Derivation](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-132.pdf)

(1) Mentions:

> Verifiers SHALL store memorized secrets in a form that is resistant to offline attacks. Memorized secrets SHALL be salted and hashed using a suitable one-way key derivation function. ... Examples of suitable key derivation functions include Password-based Key Derivation Function 2 (PBKDF2) [[SP 800-132]](https://pages.nist.gov/800-63-3/sp800-63b.html#SP800-132) and Balloon [[BALLOON]](https://pages.nist.gov/800-63-3/sp800-63b.html#balloon).

(2) essentially specifies PBKDFs and recommends their use for "the
protection of electronically-stored data or for the protection of data protection keys".

Both resources deal with _stored_ data. 

If in-memory data does not fall under that category, then as far as NIST recommendations (and FIPS 140-2) go, perhaps it's more about the hash function itself. If it's approved as a secure hash function (e.g. it's not `md5`) then there is no violation. This doesn't mean we should recommend using an unsalted (or broken!) hash function for cache hashing, we can still recommend `ssha256`, but at least we now aren't stuck with only `pbkdf2`. Anyway, I just wanted to write this up somewhere, I'm not sure this is the route we should go and we'll discuss more in sec team meeting.

One final note: [FingerprintProcessor](https://github.com/elastic/elasticsearch/blob/master/modules/ingest-common/src/main/java/org/elasticsearch/ingest/common/FingerprintProcessor.java#L218) does not support `pbkdf2` at all (which is very sensible from a performance perspective).

ywangd (changes requested):
I am tagging this as "request changes" until the team agrees on what message we want to show to the end-users.

tvernum:
My advice would be to make 1 recommendation
For example:

```
The [{}] password hasher is not recommended when running in FIPS mode. The recommended hasher is SSHA256. 
```
 
We could also say `Other FIPS compatible options are ...`, but then the question is which ones we want to list.

n1v0lg:
@tvernum I like your suggestion. 

Wondering if `Another FIPS compatible option is PBKDF, however it may significantly bottle-neck cache performance.` is a concise but clear enough way to explain how PBKDF fits it.

ywangd:
My suggestions are:
1. Only recommend `ssha256` in the warning message (this PR)
2. Amend the docs to clarify the choice of cache hashing algorithm and add details about why `ssha256` is recommended and why `pbkdf2` is not recommended even though it is "more" compatible and why it is not really necessary for in-memory credentials.
3. Update the warning message (separate PR) to add an extra sentence like `please refer to the (link-to-the-doc) for more information about choosing the cache hashing algorithm`.

tvernum:
@ywangd's suggestion looks good to me. Step 3 would be contingent on us (ES dev) reaching a conclusion on how we want to embed URLs in messages, but we can defer the work until we have an answer there.

n1v0lg:
Sounds good to me! @ywangd thank you for the suggestion.

ywangd:
> Step 3 would be contingent on us (ES dev) reaching a conclusion on how we want to embed URLs in messages, but we can defer the work until we have an answer there.

I have seen a few places where we already embed short URLs in responses or logs. One recent example is:
https://github.com/elastic/elasticsearch/blob/519f7aaacb3ce03677c1d1ab5b7527b97810b991/x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/slm/SlmHealthIndicatorService.java#L40

Not sure whether this is considered "official" though.

ywangd (approved):
LGTM

Thanks for being thorough!

PR #86741, opened 2022-05-12
Ensure authentication is wire compatible when setting user

### LINKED ISSUE #86716: [Transform] 8.2 transform can not read from < 8.2.0 CCS source
**Affected version**: 8.2.0
**Fixed with**: 8.2.1 (TBD)

After upgrading the ML QA long running cluster from 8.1.3 to 8.2.0, transforms with CCS source <8.2.0 (8.1.3 and 8.1.1 in our case) are failing with
```
Failed to retrieve checkpointing info for transform [ccs_monitoring_cluster_data_stream_with_ilm_on] org.elasticsearch.xpack.transform.checkpoint.CheckpointException: Failed to create checkpoint
[...]
Caused by: org.elasticsearch.ElasticsearchSecurityException: failed to verify signed authentication information
```
The remote cluster log shows related entries:
```
caught exception while trying to read authentication from request [transport request action [indices:monitor/stats]] java.io.EOFException: null
[...]
Authentication using apikey failed - unable to find apikey with id <REDACTED>
```
Running an `_update` request on the transforms to update the `headers._xpack_security_authentication` didn't fix it.

**Mitigation**

- upgrade the CCS remote to >= 8.2.0

- TBD: upgrade the cluster to >= 8.2.1

ywangd:
The SecurityServerTransportInterceptor class is responsible for writing
authentication header in a wire compatible format before the request
leaving the local node. However, a bug made it ignore the wire version
when setting user based on the action origin. This PR fixes it and adds
relevant tests.

It is an old bug but never manifested itself previously because (1) the
code path is rare enuough and (2) authentication didn't have any version
difference till 8.2.

Resolves: #86716

PS: I believe @hendrikmuhs will add BWC cross cluster tests on the ML side 
where the issue was originally discovered. Those tests will effectively also
test change in this PR since they would fail without it.


ywangd:
Sorry Analytics folks. If you recieved message about this PR, it is a false alarm.

hendrikmuhs:
> PS: I believe @hendrikmuhs will add BWC cross cluster tests on the ML side
where the issue was originally discovered. Those tests will effectively also
test change in this PR since they would fail without it.

I have merged #86727 earlier today, it enhances an existing CCS test to check the current branch against a previous minor. I will prepare the backport for 8.2 (to be merged after rebase).

nik9000:
> Sorry Analytics & Geo folks. If you recieved message about this PR, it is a false alarm.

:heart:

tvernum (approved):
LGTM

docs/changelog/86741.yaml
```diff
@@ -0,0 +1,5 @@
+pr: 86741
+summary: Ensure authentication is wire compatible when setting user
+area: Authentication
+type: bug
+issues: []
```
ywangd, line 5:
```suggestion
issues:
  - 86716
```

ywangd:
@elasticmachine update branch

PR #86757, opened 2022-05-12
[ML] [Transforms] prefer secondary auth headers for transforms

### LINKED ISSUE #86731: [ML] Prefer secondary authentication headers when creating/updating transforms
This requirement is similar to #53801 but for transforms rather than ML jobs.  The implementation can probably be similar to #54121.

If a transform is created or updated and a secondary authentication header is present then the security headers we persist should be based on that secondary authentication header.

If no secondary authentication header is present when a transform is created or updated then everything should work as it does today.

The reason for adding this functionality is to provide extra options when installing Fleet packages.  We need to do a better job of setting the user that transforms embedded in solutions will run as, and being able to use different credentials to those of the user installing the Fleet package will be an important building block.

benwtrent:
When creating and updating transforms, it is possible for clients to provide secondary headers.

When PUT, _preview, _update is called with secondary authorization headers, those are then used or stored with the transform.

closes: https://github.com/elastic/elasticsearch/issues/86731


docs/changelog/86757.yaml
```diff
@@ -0,0 +1,5 @@
+pr: 86757
+summary: "[Transforms] prefer secondary auth headers for transforms"
```
droberts195, line 2:
```suggestion
summary: "Prefer secondary auth headers for transforms"
```

The script that creates these strips out the `[ML]` prefix, but not `[Transforms]` for some reason. But I am sure we don't want either as a prefix in the final docs end users see.

docs/reference/transform/apis/put-transform.asciidoc
```diff
@@ -24,6 +24,12 @@ Requires the following privileges:
 * source indices: `read`, `view_index_metadata`
 * destination index: `read`, `create_index`, `index`. If a `retention_policy` is configured, the `delete` privilege is
```
droberts195, line 25: You could also modify the start transform docs page to remove the index permission requirements if they're no longer needed.

droberts195 (commented):
> This commit also fixes a bug with `_start` transform that didn't utilize the stored headers to create the destination index.

Shouldn't there be a change to `TransportStartTransformAction.java` for this part?

Have you got a commit locally that made that part of the change but isn't pushed to GitHub?

benwtrent:
> Have you got a commit locally that made that part of the change but isn't pushed to GitHub?

The x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/persistence/TransformIndex.java contains the fix.

Its about the creating destination index as part of `_start` so the bug fix is indeed for start but not in that particular file :).

droberts195:
> The x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/persistence/TransformIndex.java contains the fix.

Ah, OK, thanks. It’s getting late for me today. I’ll have a proper look tomorrow.

hendrikmuhs (approved):
LGTM

docs/reference/transform/apis/start-transform.asciidoc
```diff
@@ -22,8 +22,6 @@ Requires the following privileges:
 * cluster: `manage_transform` (the `transform_admin` built-in role grants this
   privilege)
 * source indices: `read`, `view_index_metadata`.
```
droberts195, line 24: It's interesting that start still needs privileges on the source indices.

It's because the call to validate is also made without using the stored headers: https://github.com/elastic/elasticsearch/blob/fd0e2522b18570c641e16cdb38986bda63a69037/x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/action/TransportStartTransformAction.java#L253-L257

I think this is suboptimal, because we're requiring that both the user who creates the transform _and_ the user who starts the transform has access to the source indices.

It would be better if we had a clear demarcation between permissions needed to start/stop and permissions needed to create/update.  Keeping this clear will become more important as we see more transforms being installed in packages for solutions.

droberts195 (approved):
LGTM on the secondary credentials.

I think there's another problem in the code with permissions for start vs create.  That _could_ be fixed in a separate PR as it's really nothing to do with secondary credentials, which is the subject of this PR.

Changing how permissions work for transforms is a more noticeable change for end users than introducing the secondary credentials option (which is more of an internal technical detail).  So ideally we'd have a release note that emphasises this point.

Given that the PR is mixing two things, please can you do one of the following (whichever is less effort):

1. Move the permissions fix for the destination index permissions into a separate PR and also fix the permissions required on the source index by the start API in that PR
2. Open a new PR to fix the permissions required on the source index by the start API but release note it as fixing both the permissions problems

Either way that means we'll have something in the release notes that alerts users to the change.  I don't see how anybody can complain because we're making the distinction in permissions required for different operations clearer and no operation will require more permissions than it did before.

przemekwitek (approved):
LGTM

PR #86759, opened 2022-05-12
Batch ILM move to retry step task update

original-brownbear:
This one was still not batched, same approach to making it batched as with the other step update tasks and would create tons of tasks for retries in some cases.

relates #77466


joegallo (approved):
LGTM!

PR #86760, opened 2022-05-12
Synethetic Source: Fix scaled float

nik9000:
This causes scaled float values that entirely saturate their numeric
range to continue saturating their range on round trip.

### DIFF
--- a/modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/ScaledFloatFieldMapper.java
+++ b/modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/ScaledFloatFieldMapper.java
@@ -680,6 +680,29 @@ public class ScaledFloatFieldMapper extends FieldMapper {
                 b.value(convert(value));
             }
 
+            /**
+             * Convert the scaled value back into it's {@code double} representation,
+             * attempting to undo {@code Math.round(value * scalingFactor)}. It's
+             * important that "round tripping" a value into the index, back out, and
+             * then back in yields the same index. That's important because we use
+             * the synthetic source produced by these this for reindex.
+             * <p>The tricky thing about undoing {@link Math#round} is that it
+             * "saturates" to {@link Long#MAX_VALUE} when the {@code double} that
+             * it's given is too large to fit into a {@code long}. And
+             * {@link Long#MIN_VALUE} for {@code double}s that are too small. But
+             * {@code Long.MAX_VALUE / scalingFactor} doesn't always yield a value
+             * that would saturate. In other words, sometimes:
+             * <pre>{@code
+             *   long scaled1 = Math.round(BIG * scalingFactor);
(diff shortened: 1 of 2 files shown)


modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/ScaledFloatFieldMapper.java
```diff
@@ -677,7 +677,54 @@ public SourceLoader.SyntheticFieldLoader syntheticFieldLoader() {
...
+                if (scaledValue == Long.MIN_VALUE) {
+                    double min = Long.MIN_VALUE / scalingFactor;
+                    if (Math.round(min * scalingFactor) != Long.MIN_VALUE) {
+                        double v = min - Math.ulp(min);
+                        assert Math.round(v * scalingFactor) == Long.MIN_VALUE;
+                        return v;
+                    }
+                    return min;
+                }
+                return scaledValue / scalingFactor;
```
romseygeek, line 727: Given this logic is fairly gnarly, is it worth extracting it into a helper and then throwing some heavy-duty testing on it? Even something quick-check like, to test boundary conditions and then a solid random selection of points.

nik9000: We get that from the round trip tests but I could certainly see wanting it to be tested more directly. I'll yank it around a bit.

nik9000:
@romseygeek I've pushed some more tests. Let me know if you think I should add even more. This is a tricky thing.

romseygeek (approved):
LGTM. Let's get it in and see what CI makes of it.

modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/ScaledFloatFieldMapper.java
```diff
@@ -677,8 +681,55 @@ public SourceLoader.SyntheticFieldLoader syntheticFieldLoader() {
...
+     * {@link Long#MIN_VALUE} for {@code double}s that are too small. But
+     * {@code Long.MAX_VALUE / scalingFactor} doesn't always yield a value
+     * that would saturate. In other words, sometimes:
+     * <pre>{@code
+     *   long scaled1 = Math.round(BIG * scalingFactor);
+     *   assert scaled1 == Long.MAX_VALUE;
+     *   double decoded = scaled1 / scalingFactor;
+     *   long scaled2 = Math.round(decoded * scalingFactor);
+     *   assert scaled2 != Long.MAX_VALUE;
+     * }</pre>
```
not-napoleon, line 708: Good explanation, thanks for including it.

not-napoleon (approved):
LGTM.  Or, well, looks as good as this much floating point jank can look ;)  Nice work.

nik9000:
Be free PR! Go and mingle with your brothers and sisters in CI and find bugs.

Issue #84379, opened 2022-02-25
Support splitting Groups list in SAML


tvernum:
### Description

Our SAML realm has a setting `attribute.groups` that tells the realm which SAML attribute contains the list of "groups" that the user belongs to.
(Note: "groups" here is a generic term, carried over from the LDAP realm. The SAML IdP might refer to this attribute as "roles" or "teams" or something else entirely).

Because SAML supports multi-valued attributes (example below) we assume that each group will be listed as a distinct value.
```
 <saml:Attribute NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:uri"
       Name="http://idp.example.org/attributes/groups" FriendlyName="groups">
       <saml:AttributeValue>engineering</saml:AttributeValue>
       <saml:AttributeValue>elasticsearch-admins</saml:AttributeValue>
       <saml:AttributeValue>employees</saml:AttributeValue>
</saml:Attribute>
```
However, it appears some IdPs are incapable of sending the groups list as a list of values and will only ever send it as a comma separated string.
```
 <saml:Attribute NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:uri"
       Name="http://idp.example.org/attributes/groups" FriendlyName="groups">
       <saml:AttributeValue>engineering,elasticsearch-admins,employees</saml:AttributeValue>
</saml:Attribute>
```

We currently require that the cluster security admin create a regex based role mapping rule to match on something like `^(.*,)?elasticsearch-admins(,.*)?$` which is non-obvious, cumbersome and error-prone.

An alternative would be to add a setting like `attribute.groups_split: ","` to automatically split a comma separated string into a
real list.  
This would be a reflection that real world IdPs don't always do a good job of conforming with the intent of the SAML spec, and our need to work within their limitations.

bytebilly:
Thanks Tim, I like the approach to make UX better. Do you have a feeling on how common the single-value scenario is, or if it affects common IdP?

Related question: is `,` an allowed character for group names? If not, should we automatically detect and support comma-separated values without requiring the admin to configure a new option?

jkakavas:
> Do you have a feeling on how common the single-value scenario is, or if it affects common IdP?

I don't know of this to be a limitation of any well-known (or even just commonly used) IdP implementations. I think this is mostly a configuration issue on the IdP administration side, rather than a software limitation. I agree with the arguments around doing something about this, nevertheless. 

[...]

PR #86770, opened 2022-05-12
[DOCS] SAML groups can use wildcards

### LINKED ISSUE #86409: [DOCS] SAML role mapping of groups does not parse comma separated lists to string array
### Description

Following the documentation for setting up SAML integration with Elastic, it mentions using a "groups" attribute so the SAML server can pass a list of groups back to ElasticSearch/Kibana: https://www.elastic.co/guide/en/elasticsearch/reference/current/saml-guide-stack.html

You will notice here that the role mapping is specifying the group name "finance-team".
```
PUT /_security/role_mapping/saml-finance
{
  "roles": [ "finance_data" ],
  "enabled": true,
  "rules": { "all": [
        { "field": { "realm.name": "saml1" } },
        { "field": { "groups": "finance-team" } }
  ] }
}
```

In my case, we were unable to get this to work because the SAML server is actually passing a string that contains a comma separated list of security groups that the user belongs to and not just one single group.

Therefore, it appears that Elastic is not converting this string to an array. I do not know if this is by design but the purpose of opening this feature request is to document that you can actually use wildcards in the role mapping:
image

To document this better, I think these questions will need to be answered:
- Is Elastic supposed to parse the "groups" string from SAML into an array if it detects it is a comma separated string?
- Are the wildcards actually supported on the "text" type for the user field "groups". It works, but I didn't see it documented and wondering if it could stop working in the future.

This is what the SAML server is returning to Elastic/Kibana (example):
```
"rows": [
            {
                "name": "mail",
                "value": "jdoe@domain.com",
                "claimType": null,
                "attrNameFormat": "urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified",
                "responsePosition": "BODY",
                "checkRequest": false,
                "returnResponse": true,
                "getInput": false
            },
            {
                "name": "nameid",
                "value": "jdoe@domain.com",
                "claimType": null,
                "attrNameFormat": "urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified",
                "responsePosition": "BODY",
                "checkRequest": false,
                "returnResponse": true,
                "getInput": false
            },
            {
                "name": "groups",
                "value": "Organization Management,Discovery Management,Mailbox Import and Export Role,Domain Admins,Denied RODC Password Replication Group,VPN Users,Offer Remote Assistance Helpers,MSSQL - Admins,Systems Analyst",
                "claimType": null,
                "attrNameFormat": "urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified",
                "responsePosition": "BODY",
                "checkRequest": false,
                "returnResponse": true,
                "getInput": false
            }
        ],
```

lockewritesdocs:
As described in #84379, Elasticsearch can't split comma-separated groups passed into a SAML realm from an IdP. In #86409, we decided to add this limitation to the docs and also describe that wildcards can be used in role mapping for `groups`. This PR adds that documentation.

Closes #86409

### DIFF
--- a/x-pack/docs/en/security/authentication/saml-guide.asciidoc
+++ b/x-pack/docs/en/security/authentication/saml-guide.asciidoc
@@ -342,7 +342,7 @@ matches only on the `elasticsearch-admins` group:
 ----
 
 These regular expressions are based on Lucene’s
-{ref}/regexp-syntax.html[regexp automaton syntax], and can match more complex 
+{ref}/regexp-syntax.html[regular expression syntax], and can match more complex 
 patterns. All regular expressions must start and end with a forward slash.
 ====
 
--- a/x-pack/docs/en/security/authentication/saml-guide.asciidoc
+++ b/x-pack/docs/en/security/authentication/saml-guide.asciidoc
@@ -326,10 +326,24 @@ groups:: _(Recommended)_
 +
 [NOTE]
 ====
-Some IdPs are only able to send the `groups` list as a comma-separated string.
-{es} can't parse this string into an array of groups. However, you can use a
-wildcard when <<saml-role-mapping,configuring role mappings>> to map this SAML
-attribute to the `attributes.groups` setting in the {es} realm.
+Some IdPs are configured to send the `groups` list as a comma-separated string, 
+but {es} can't parse this string into an array of groups. To map this SAML 
+attribute to the `attributes.groups` setting in the {es} realm, a cluster 
+security administrator can use a wildcard when
+<<saml-role-mapping,configuring role mappings>>. While flexible, wildcards are
+less accurate and can match on unwanted patterns. Instead, a cluster security 
+administrator can use a regular expression to create a role mapping rule that 
+matches only a single group. For example, the following regular expression
+matches only on the `elasticsearch-admins` group:
+
+[source,sh]
+----
+/^(.*,)?elasticsearch-admins(,.*)?$/
+----
+
+These regular expressions are based on Lucene’s
+{ref}/regexp-syntax.html[regexp automaton syntax], and can match more complex 
+patterns. All regular expressions must start and end with a forward slash.
 ====
 
 name:: _(Optional)_ The user's full name.
@@ -718,7 +732,8 @@ PUT /_security/role_mapping/saml-finance
 }
 --------------------------------------------------
 <1> The `groups` attribute supports using wildcards (`*`). Refer to the
-<<saml-es-user-properties,{es} user properties>> for more information.
+<<security-api-put-role-mapping-example,create or update role mappings API>> for 
+more information.
 
 If your users also exist in a repository that can be directly accessed by {es}
 (such as an LDAP directory) then you can use
--- a/x-pack/docs/en/security/authentication/saml-guide.asciidoc
+++ b/x-pack/docs/en/security/authentication/saml-guide.asciidoc
@@ -322,7 +322,15 @@ groups:: _(Recommended)_
     If you wish to use your IdP's concept of groups or roles as the basis for a
     user's {es} privileges, you should map them with this attribute.
     The `groups` are passed directly to your
-    <<saml-role-mapping, role mapping rules>>
+    <<saml-role-mapping, role mapping rules>>.
++
+[NOTE]
+====
+Some IdPs are only able to send the `groups` list as a comma-separated string.
+{es} can't parse this string into an array of groups. However, you can use a
+wildcard when <<saml-role-mapping,configuring role mappings>> to map this SAML
+attribute to the `attributes.groups` setting in the {es} realm.
+====
 
 name:: _(Optional)_ The user's full name.
 mail:: _(Optional)_ The user's email address.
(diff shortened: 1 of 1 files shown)


x-pack/docs/en/security/authentication/saml-guide.asciidoc
```diff
@@ -705,10 +713,12 @@ PUT /_security/role_mapping/saml-finance
   "enabled": true,
   "rules": { "all": [
         { "field": { "realm.name": "saml1" } },
-        { "field": { "groups": "finance-team" } }
+        { "field": { "groups": "finance-team" } } <1>
   ] }
 }
 --------------------------------------------------
+<1> The `groups` attribute supports using wildcards (`*`). Refer to the
+<<saml-es-user-properties,{es} user properties>> for more information.
```
ywangd, line 721: I think it would be better to link the [doc page for role mapping](https://elasticsearch_86770.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/security-api-put-role-mapping.html#security-api-put-role-mapping-example) since it gives more examples of using not only wildcard, but also template. We don't currently document the usage of regex. But if we do, it will be part of the role mapping doc page as well.

x-pack/docs/en/security/authentication/saml-guide.asciidoc
```diff
@@ -322,7 +322,15 @@ groups:: _(Recommended)_
...
-    <<saml-role-mapping, role mapping rules>>
+    <<saml-role-mapping, role mapping rules>>.
++
+[NOTE]
+====
+Some IdPs are only able to send the `groups` list as a comma-separated string.
+{es} can't parse this string into an array of groups. However, you can use a
+wildcard when <<saml-role-mapping,configuring role mappings>> to map this SAML
+attribute to the `attributes.groups` setting in the {es} realm.
+====
```
ywangd, line 333: Here is some more information that may help paraphrase these sentences a bit more: It's mostly likely a configuration issue rather than idPs "not able to send". That said, sometimes our customres have no control over their idPs. In those situations, the end result is the same for them, i.e. they recieve a comma-separate string that is meant to be a list. ES does not (for now) attempt to split it. 

It is recommended to fix it on the idP side. But as said above, customers may not have control. For now the workaround is to use wildcard (regex is more accurate but we haven't really document its general usage other than [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/defining-roles.html#CO545-5)).

ywangd (commented):
I left two comments for a bit more contexts. Thanks!

x-pack/docs/en/security/authentication/saml-guide.asciidoc
lockewritesdocs: That change sounds good to me -- I'll update the link.

x-pack/docs/en/security/authentication/saml-guide.asciidoc
lockewritesdocs: Thanks for the context @ywangd! Per our conversation, we can also add a suggestion about regex here with an example from #84379. Let me know what you think.

> Some IdPs are configured to send the `groups` list as a comma-separated string, but {es} can't parse this string into an array of groups. To map this SAML attribute to the `attributes.groups` setting in the {es} realm, a cluster security administrator can use a wildcard when <<saml-role-mapping,configuring role mappings>>. Alternatively, a cluster security administrator can use a regular expression to create a role mapping rule such as `/^(.*,)?elasticsearch-admins(,.*)?$/` that matches on multiple groups. These regular expressions are based on Lucene’s regexp automaton syntax, and can match more complex patterns. All regular expressions must start and end with a forward slash.

ywangd: Thanks Adam.
* We can link to the doc page of [lucene regex syntax](https://www.elastic.co/guide/en/elasticsearch/reference/current/regexp-syntax.html)
* "Alternatively ... matches on _multiple_ groups" - It actually only matches a single group which is `elasticsearch-admins`. The extra stuff around it is to make sure this name can appear anywhere in the comma separate list, e.g. it can be in the beginning where it is _not_ preceded with a comma; or it can be at the end of the string where it is _not_ followed by a comma; or it can be in the middle where it is both preceded and followed by a comma; or it can be standalone where it is the only member of the string. Using only wildcard is not possible to achieve this level of accuracy and flexibility. For example, `*elasticsearch-admins*` can also match `my-elasticsearch-admins` which is probably not what the user wants. Or `*,elasticsearch-admin,*`, but it wont' match if `elasticsearch-admins` is at the begining of the string.

lockewritesdocs: Thanks Yang. I updated the note in [this recent push](https://github.com/elastic/elasticsearch/pull/86770/commits/fc46081becdd582ad56701e7c0edf97611aa0d6d).

x-pack/docs/en/security/authentication/saml-guide.asciidoc
```diff
@@ -322,7 +322,29 @@ groups:: _(Recommended)_
...
+matches only a single group. For example, the following regular expression
+matches only on the `elasticsearch-admins` group:
+
+[source,sh]
+----
+/^(.*,)?elasticsearch-admins(,.*)?$/
+----
+
+These regular expressions are based on Lucene’s
+{ref}/regexp-syntax.html[regexp automaton syntax], and can match more complex
```
ywangd, line 345:
```suggestion
{ref}/regexp-syntax.html[regular expression syntax], and can match more complex 
```

ywangd (approved):
LGTM 

Thanks for the quick turnaround!

PR #86773, opened 2022-05-12
Add license and notice files for commons daemon

rjernst:
Elasticsearch uses Apache procrun (part of the commons-daemon project)
to launching as a service on Windows. This commit adds attribution for
the project by including license and notice files. These were missed in
the past because the dependency is just two binary exes that are
embedded in the elasticseach codebase.


mark-vieira (approved):
Thanks, Ryan. LGTM.

PR #86777, opened 2022-05-13
Fix the test framework codebase name

rjernst:
A side effect of https://github.com/elastic/elasticsearch/pull/86706 was
unintentionally renaming the test framework jar. This commit reverts the
jar name to the horrible named "framework", and adjusts the test
security manager bootstrap to account for the name.


mark-vieira (approved):
LGTM

PR #86781, opened 2022-05-13
Update gradle JavaPlugin to support generating javadoc for modules

ChrisHegarty:
This change updates the JavaPlugin to support generating javadoc for modules, i.e. ensure that modular dependencies can be found on the module path.


breskeby (changes requested):
That should probably life in the ElasticsearchJavadocPlugin

ChrisHegarty:
> That should probably life in the ElasticsearchJavadocPlugin

Good idea. Moved.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavadocPlugin.java
```diff
@@ -46,6 +47,9 @@ public void apply(Project project) {
             // see https://discuss.gradle.org/t/add-custom-javadoc-option-that-does-not-take-an-argument/5959
             javadoc.getOptions().setEncoding("UTF8");
             ((StandardJavadocDocletOptions) javadoc.getOptions()).addStringOption("Xdoclint:all,-missing", "-quiet");
+
+            // ensure that modular dependencies can be found on the module path
+            javadoc.getOptions().modulePath(Util.getJavaMainSourceSet(project).get().getCompileClasspath().getFiles().stream().toList());
```
breskeby, line 52: Why do we always put the whole compileclasspath on the modulePath here?

Using `getCompileClasspath().getFiles()` here is problematic, It materializes all files of `getCompileClasspath` here at configuration time. That has two major flaws:

1. At that point it is not guaranteed that the dependent classpath is already fully available. e.g. when you have a project dependency, that project has likely not been built yet when this logic here is triggered. 
2. In general that slows down the configuration time as its quite an expensive operation given we do this for every project (in builds that generate javadoc).

I guess we cannot use `javadoc#modularity#inferClasspath` here to avoid this (as we don't infer classpaths when compiling)?

A proper fix would be that Gradle supports lazy file collections that are evaluated at execution time when really needed. I will discuss this with the gradle team and maybe raise a PR for this. Its a shame they introduced this with 6.4 without taking this into account. 

As a workaround for now, we can do the following: 

```
javadoc.doFirst(new Action<Javadoc>() {
                @Override
                public void execute(Javadoc task) {
                    task.getOptions().modulePath(task.getClasspath().getFiles().stream().toList());
                }
            });
```

This snippet assumes that the javadoc tasks classpath input as it is configured by the gradle built-in java plugin is basically the same as what we'd get with `Util.getJavaMainSourceSet(project).get().getCompileClasspath()`

Also by we aim for making our build logic compatible with gradles `configuration cache`. Referencing the mutual `project` object in a task action is breaking this compatibility (I know I know we have other places where we still do that but we do not want to introduce new ones).

In general since I just added this nice `JavadocPluginFuncTest` we should add some basic coverage there for all this.

breskeby (changes requested):

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaPlugin.java
```diff
@@ -43,6 +43,7 @@ public class ElasticsearchJavaPlugin implements Plugin<Project> {
     public void apply(Project project) {
         project.getPluginManager().apply(ElasticsearchJavaBasePlugin.class);
         project.getPluginManager().apply(JavaLibraryPlugin.class);
+        project.getPluginManager().apply(ElasticsearchJavaModulePathPlugin.class);
```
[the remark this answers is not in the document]
ChrisHegarty: It is not. I'll remove it.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavadocPlugin.java
ChrisHegarty: >I guess we cannot use javadoc#modularity#inferClasspath here to avoid this (as we don't infer classpaths when compiling)?

We cannot use Gradle's inference here for the same reason that we cannot use it when compiling source - it will not put all dependencies on the module path, e.g. jars without either a module-info.class or a Automatic-Module-Name attribute in their manifest. One such example is jsr305-3.0.2.jar, which is a dependency of :libs:core.

ChrisHegarty:
A minimal test has been added.

breskeby (approved):
lgtm

PR #86782, opened 2022-05-13
Disable High Level Rest Client javadoc task

ChrisHegarty:
This change disables generating the javadoc for the High Level Rest Client.

We no long publish the HLRC, and this Javadoc task requires some work to get working with Java modules, so just disable it if we don't need it.


rjernst (approved):
LGTM

PR #86784, opened 2022-05-13
Move RestWrapper to interceptor.RestInterceptor

ChrisHegarty:
Refactor the REST handler wapper into a separate interface.

Moving this to its own package will allow to share with just xpack security when we integrate the modules branch (by means of a qualified export). That way the runtime check will not be strictly required - since the interface will only be accessible to the xpack security module.

Trivially, wrapper is replaced with interceptor.


rjernst (approved):
LGTM. Just to confirm, once the modules branch has landed and this is a qualified export, we can remove this error handling altogether because plugins would not have access to the class at all, right?

ChrisHegarty:
>LGTM. Just to confirm, once the modules branch has landed and this is a qualified export, we can remove this error handling altogether because plugins would not have access to the class at all, right?

Exactly.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/SecurityTests.java
```diff
@@ -686,10 +686,10 @@ public void testSecurityRestHandlerWrapperCanBeInstalled() throws IllegalAccessE
             // We will throw later if another wrapper is already installed
```
ywangd, line 686: Nit:
```suggestion
            // We will throw later if another interceptor is already installed
```

I cannot comment on the line above this one. But it also uses the word `wrapper` instead of `interceptor`. The same goes for the name of this method and the previous one.

server/src/main/java/org/elasticsearch/action/ActionModule.java
```diff
@@ -477,24 +478,26 @@ public ActionModule(
...
-                logger.debug("Using REST wrapper from plugin " + plugin.getClass().getName());
-                if (plugin.getClass().getCanonicalName() == null
-                    || plugin.getClass().getCanonicalName().startsWith("org.elasticsearch.xpack") == false) {
-                    throw new IllegalArgumentException(
-                        "The "
-                            + plugin.getClass().getName()
-                            + " plugin tried to install a custom REST "
-                            + "wrapper. This functionality is not available anymore."
-                    );
+            if (plugin instanceof RestInterceptor riplugin) {
```
ywangd, line 483: With this change, there is no inherent relationship between `ActionPlugin` and `RestInterceptor`. So how can we be sure this `instanceof` check is only needed inside a loop for ActionPlugins? It is probably not an issue at all and mainly due to my lack of understanding of future modules change. I appreicate if you could help explaining it. Thanks!

ywangd (approved):
LGTM

> Moving this to its own package will allow to share with just xpack security when we integrate the modules branch (by means of a qualified export). That way the runtime check will not be strictly required - since the interface will only be accessible to the xpack security module.

This is really for my own knowledge: When you say "will not be **strictly** required", does it mean there are still ways to overcome "qualified export" restrictions?

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/SecurityTests.java
ChrisHegarty: Good catch. Fixed.

server/src/main/java/org/elasticsearch/action/ActionModule.java
ChrisHegarty: Yeah, this is a good point. I think it would be better to make RestInterceptor extend ActionPlugin (since it is really a specialised action plugin, rather an a standalone entity )

ChrisHegarty:
> This is really for my own knowledge: When you say "will not be **strictly** required", does it mean there are still ways to overcome "qualified export" restrictions?

Modules and qualified exports are part of the Java language, JVM, and runtime, but .... just like with package-private members, there are ways that one can somewhat bypass restrictions, if one has access to the all powerful command line. For example, putting a custom user jar on the class path that contains classes from org.elasticsearch.XXX, will allow the custom user code effective package access to members of org.elasticsearch.XXX. Similarly, adding a `--add-exports` command line flag could export the o.e.plugins.interceptor package to code beyond that of pack security. (But one has to have access to the command line! )

PR #86785, opened 2022-05-13
Save redundant singleton maps in field mappers

original-brownbear:
How about this relatively straight-forward win for the mapper size? :)

In the many-shards benchmarks the singleton maps storing just a single
analyzer for each keyword field mapper cost around 5% of the total heap
usage on data nodes (700MB for ~15k indices which themselves contain ~16M instances
of keyword field mapper for Beats mappings, so the few bytes that each of these maps 
costs adds up enormously).
Creating specific implementations for the zero, one or many analyzers
use cases that already have their own specialized constructors eliminates this
overhead completely.

Also, IMO this refactoring makes the code a little easier to follow by moving the analyzers map implementation into those mappers that actually need it. Technically one could make this even drier by creating new parents for the couple duplicate cases that came up here, but I elected not to, to keep things simple for now. Even in the current form we're saving 100+ LOC here anyway.


romseygeek (approved):
LGTM, thanks @original-brownbear

original-brownbear:
np + thanks Alan!

PR #86790, opened 2022-05-15
Auto generate index.routing_path from mapping

martijnvg:
If no `index.routing_path` has been specified in the matching template or the component template that the template refers to then attempt to generate the `index.routing_path` index setting from the matching template's mapping or component template the template refers to.

Relates #74660


modules/data-streams/src/main/java/org/elasticsearch/datastreams/DataStreamIndexSettingsProvider.java
```diff
@@ -92,4 +118,54 @@ public Settings getAdditionalIndexSettings(
         return Settings.EMPTY;
     }
 
+    // Find fields in mapping that are of type keyword and time_series_dimension enabled.
```
nik9000, line 121: I'd use javadoc for this. It'll show up when you mouseover the method in your IDE which is kind of nice.

nik9000 (commented):
It makes sense to me. I'm not particularly worried about the overhead involved when validating templates. I *think* it's a fairly infrequent action - just on template PUT. Is that true?

martijnvg:
>  I think it's a fairly infrequent action - just on template PUT. Is that true?

This will be used when creating a template (as part of validating), when simulating a template via simulate api and creation of a backing index of data stream. The latter will be the most frequent compared to the other two use cases and happens as part of creating a new data stream or rolling over a data stream. I still think that this is okay.

martijnvg:
Pinging @jsoriano for awareness.

nik9000 (approved):

PR #86791, opened 2022-05-15
Simplify Netty4HttpPipeliningHandler

original-brownbear:
No need to collect responses into an intermediary list, we can write them directly
as they become writable.
Also, we can be more efficient for the common case where the response written is of the
right sequence number and write it out directly.

follow up to #86133


modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -109,11 +109,30 @@ public void write(final ChannelHandlerContext ctx, final Object msg, final Chann
...
+                // response is not at the current sequence number so we add it to the outbound queue and return
+                outboundHoldingQueue.add(new Tuple<>(response, promise));
+                success = true;
+                return;
             }
+
+            // response is at the current sequence number and does not need to wait for any other response to be written so we write
+            // it out directly
+            doWrite(ctx, response, promise);
             success = true;
```
idegtiarenko, line 129: Should this be set after the loop below?

original-brownbear: No in fact I consciously moved it here. Once we resolve the promise in `doWrite` (which we will do one way or another) we must not touch it again, that would always be a bug.

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -109,11 +109,30 @@ public void write(final ChannelHandlerContext ctx, final Object msg, final Chann
...
         boolean success = false;
         try {
-            List<Tuple<Netty4HttpResponse, ChannelPromise>> readyResponses = write((Netty4HttpResponse) msg, promise);
-            for (Tuple<Netty4HttpResponse, ChannelPromise> readyResponse : readyResponses) {
-                doWrite(ctx, readyResponse.v1(), readyResponse.v2());
+            final Netty4HttpResponse response = (Netty4HttpResponse) msg;
+            if (response.getSequence() != writeSequence) {
+                if (outboundHoldingQueue.size() >= maxEventsHeld) {
+                    int eventCount = outboundHoldingQueue.size() + 1;
+                    throw new IllegalStateException(
```
henningandersen, line 116: This PR slightly changes semantics for when the queue is nearly full in that if the event that triggers that the queue is full has the right seq-no, it no longer fails. That sounds fine to me, but wanted to ensure other reviewers consider it.

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
henningandersen: Can we add a test that provokes that failure and then see that it is fixed by this PR? Seems we do not have tests that verify that we do not double notify failures and we used to sometimes do so?

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -109,11 +109,30 @@ public void write(final ChannelHandlerContext ctx, final Object msg, final Chann
         assert msg instanceof Netty4HttpResponse : "Invalid message type: " + msg.getClass();
         boolean success = false;
         try {
-            List<Tuple<Netty4HttpResponse, ChannelPromise>> readyResponses = write((Netty4HttpResponse) msg, promise);
-            for (Tuple<Netty4HttpResponse, ChannelPromise> readyResponse : readyResponses) {
-                doWrite(ctx, readyResponse.v1(), readyResponse.v2());
+            final Netty4HttpResponse response = (Netty4HttpResponse) msg;
+            if (response.getSequence() != writeSequence) {
+                if (outboundHoldingQueue.size() >= maxEventsHeld) {
```
henningandersen, line 114: Can we assert `response.getSequence() > writeSequence`?

henningandersen (approved):
LGTM.

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
original-brownbear: I don't think we can test this without creating an unrealistic situation. We should never be throwing here unless we exceed the outbound queue capacity to begin with. I really just fixed this as a cleanup to make it more obviously correct, unless we have a bug somewhere this doesn't change anything.

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
original-brownbear: sure done :)

PR #86802, opened 2022-05-16
[ML] [Transforms] fix transform _start permissions to use stored headers in the config

benwtrent:
It was previously required that the `_start` API caller required the same roles as the `create` API caller. 

This does not make sense as when the transform is actually running (after `_start`) we rely solely on the roles of the caller who created the transform. 

Consequently, this commit does the permission validations and various checks with the roles of user who created the transform, not the one calling `_start`

### DIFF
--- a/docs/changelog/86802.yaml
+++ b/docs/changelog/86802.yaml
@@ -1,5 +1,5 @@
 pr: 86802
-summary: "[Transforms] fix transform `_start` permissions to use stored headers in\
+summary: "Fix transform `_start` permissions to use stored headers in\
   \ the config"
 area: Transform
 type: bug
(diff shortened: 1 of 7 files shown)


docs/changelog/86802.yaml
```diff
@@ -0,0 +1,6 @@
+pr: 86802
+summary: "[Transforms] fix transform `_start` permissions to use stored headers in\
```
droberts195, line 2:
```suggestion
summary: "Fix transform `_start` permissions to use stored headers in\
```

droberts195 (approved):
LGTM if you could just tweak the release note text

przemekwitek (commented):
LGTM

PR #86806, opened 2022-05-16
Simplify CellIdSource classes used in GeoGrid aggregations

iverase:
Currently we have three classes that shares a fair amount of code. This Change extract a base class called `CellIdSource` in order to simplify the current implementation.


x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/bucket/geogrid/GeoHexCellIdSource.java
```diff
@@ -6,111 +6,67 @@
...
-                    return bbox.left() < lon || bbox.right() > lon;
-                } else {
-                    return bbox.left() < lon && bbox.right() > lon;
-                }
-            }
-            return false;
-        }
-
-        private boolean validHex(long hex) {
+        public boolean validHex(long hex) {
```
craigtaverner, line 69: Should be private

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/bucket/geogrid/GeoHexCellIdSource.java
```diff
@@ -145,5 +101,16 @@ private boolean intersects(double minLon, double maxLon, double minLat, double m
             }
             return false;
         }
+
+        public boolean validPoint(double lat, double lon) {
```
craigtaverner, line 105: Should be private

craigtaverner: Also, this method, and the one in `GeoTileCellIdSource`, could move to `CellIdSource` parent class, which can also have `crossesDateline`. Since two of the three child classes use the same function, perhaps it makes sense. However, if `GeoTileCellIdSource`  stops using `validPoint`, then there is no point in moving this.

craigtaverner (approved):
Looks good, although I think two functions should be made private. And I had some other questions that are not important, but I'm curious.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/geogrid/GeoHashCellIdSource.java
```diff
@@ -7,100 +7,55 @@
...
-            if (validPoint(target.getLon(), target.getLat()) || predicate.validHash(hash)) {
-                values[valuesIdx] = Geohash.longEncode(hash);
-                return valuesIdx + 1;
+    protected CellValues boundedCellValues(MultiGeoPointValues values, GeoBoundingBox boundingBox) {
+        final GeoHashBoundedPredicate predicate = new GeoHashBoundedPredicate(precision(), boundingBox);
+        return new CellValues(values, precision()) {
+            @Override
+            protected int advanceValue(org.elasticsearch.common.geo.GeoPoint target, int valuesIdx) {
+                final String hash = Geohash.stringEncode(target.getLon(), target.getLat(), precision);
+                if (validPoint(boundingBox, target.getLon(), target.getLat()) || predicate.validHash(hash)) {
```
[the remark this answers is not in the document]
iverase: It is a performance optimization. Creating a rectangle for every point can be expensive so if the point is inside of the bounding box we can return true without creating a new object.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/bucket/geogrid/GeoHexCellIdSource.java
iverase: I like the idea of moving this method to `CellIdSource`, I have a go

PR #86808, opened 2022-05-16
Fix Terminal flushing and make api consistent

rjernst:
This commit fixes the PrintWriters created by Terminal to use
autoflushing. It also fixes prompting to flush the error stream after
writing. Finally, it makes the APIs more consistent by adding getters
for verbosity and reader so that subclasses of Terminal can fully wrap
and existing terminal.

relates #85758


rjernst:
Note that the auto flushing issue was introduced recently and so is not a bug visible to users.

rjernst:
@elasticmachine run elasticsearch-ci/eql-correctness

williamrandolph (approved):
LGTM!

PR #86809, opened 2022-05-16
Improve debug logging in packaging tests

rjernst:
When the Elasticsearch log file cannot be moved during packaging test
cleanup, it usually means the Windows Elasticsearch process is still
running. This commit moves the code to dump all available log
information out to a separate dumpDebug() method and utilizes it when
a failure ocurrs moving the log file.

relates #85758


williamrandolph (approved):
LGTM!

PR #86820, opened 2022-05-16
Guard cursor for journalctl command in packaging tests

rjernst:
It is possible for the cursor from journalctl to be empty, perhaps if
the log is empty. This commit guards the getLogs command line to only
add the --after-cursor if a non empty cursor exists, otherwise
journalctl raises an error.


mark-vieira (approved):
LGTM

PR #86821, opened 2022-05-16
Dump a truncated file if over limit in packaging tests

rjernst:
The packaging tests run many shell commands, and capture the stdout and
stderr of those commands into files. If the file is extremely large, the
shell omits reading the file. That makes debugging any issue with large
output difficult. This commit adds handling to read up to the max file
size worth of characters from the file.

### DIFF
--- a/qa/os/src/test/java/org/elasticsearch/packaging/util/Shell.java
+++ b/qa/os/src/test/java/org/elasticsearch/packaging/util/Shell.java
@@ -202,8 +202,14 @@ public class Shell {
     private String readFileIfExists(Path path) throws IOException {
         if (Files.exists(path)) {
             long size = Files.size(path);
-            if (size > 100 * 1024) {
-                return "<<Too large to read: " + size + " bytes>>";
+            final int maxFileSize = 100 * 1024;
+            if (size > maxFileSize) {
+                // file is really big, truncate
+                try (var br = Files.newBufferedReader(path, StandardCharsets.UTF_8)) {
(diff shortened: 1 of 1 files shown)


mark-vieira (commented):
What do we end up doing with the returned output? Maybe we should provide a way to grab these things as streams instead?

rjernst:
We have various assertions on the output/error string values. It would be a lot of work to convert these all to read the output themselves. I think the approach here is pragmatic: if we have  more than 100k output, something is wacky, let's just get what we can, otherwise the test is impossible to debug (we don't know what the file is filled with).

mark-vieira:
> if we have more than 100k output, something is wacky, let's just get what we can

Would it probably make more sense then to _tail_ the output? Perhaps we'll see how it goes in practice and see if just getting the head is enough to debug these issues.

rjernst:
I don't think tail or head particularly matters. In the case that I experienced this issue, procrun was outputting gigabytes of null characters (ie not printable). But I had no idea, I only saw that the file size was 8GB, and didn't know anything about the content.

rjernst:
@mark-vieira Do you have any other thoughts here? I don't have that strong of an opinion here, I can drop this PR, but I do think it's wrong to just print the existing cryptic "the file was too long". If that ever fails, as it did for me, it makes debugging extremely difficult, and ultimately I had to add this to see what was going on.

mark-vieira (approved):
This is certainly an improvement. Let's push forward with this.

PR #86824, opened 2022-05-17
[DOCS] Updates example output for start trained model deployment API

lcawl:
This PR updates the example output in the [start trained model deployment API](https://www.elastic.co/guide/en/elasticsearch/reference/master/start-trained-model-deployment.html) docs to include the `threads_per_allocation`, `number_of_allocations`, and `queue_capacity` values. It also removes a reference to `inference_threads`, which was replaced in https://github.com/elastic/elasticsearch/pull/86597.


dimitris-athanasiou (approved):
LGTM Thanks Lisa!

PR #86829, opened 2022-05-17
Skip redundant name collision check in metadata build

original-brownbear:
No need to check for name collisions when we are reusing the indices lookup
and have checked that we didn't change any naming.
-> saves us another percent on CS updates around reroute

relates #77466


martijnvg (approved):
👍 LGTM

original-brownbear:
Thanks Martijn!

Issue #85955, opened 2022-04-18
flag to disable forward on individual yaml tests


nik9000:
### Description

Right now we disable forwards compatibility testing with something like:
```
task.skipTest("search.aggregation/20_terms/string profiler via global ordinals filters implementation", "The profiler results aren't backwards compatible.")
task.skipTest("search.aggregation/20_terms/string profiler via global ordinals native implementation", "The profiler results aren't backwards compatible.")
```
This *works* but it's kind of inconvenient both for writing the tests and for reading them. It'd be nice if we had a flag we could put on the tests in the yaml file that'd disable them in the forwards compatibility testing. Mostly for things like profiler output - useful to have tests for but that aren't controlled by semver.

mark-vieira:
The confusing thing about this is these tests actually live in 7.17 so we'd have to annotate them _there_. The benefit of configuring this at the task level is at least this all lives in the branch which you are actually testing, and you don't have to go look in a different branch to see if/why something is being ignored. There's also the question of associating these things with versions. The way it is now the skips and transforms are forward-compatibile version specific, but if this stuff lives in the 7.17 it would apply _everywhere_ which may not the the correct thing to do.

The compatibility tests are... awkward, but I'm not sure moving config into the tests (in the 7.17 branch) would necessarily alleviate any issues and may cause more. I'd be curious what specific workflow this would improve. At this point, the existing 7.17 test suites shouldn't be changing much, so we shouldn't really need to be adjusting these exclusions unless we actually break forward compatibility in a later dev branch, in which case it certainly makes sense that we should make the exclusions there.

nik9000:
> he confusing thing about this is these tests actually live in 7.17 so we'd have to annotate them _there_.

Ooof. Yeah. That's rough. I suppose the reason this comes up for me is that there are some tests that I know really shouldn't be forwards tested when I write them and I don't have a good way to mark them. I suppose I could add them to the filter in build when I write them. But then renames lose it. But, I guess, that's not the end of the world.

mark-vieira:
> I suppose the reason this comes up for me is that there are some tests that I know really shouldn't be forwards tested when I write them and I don't have a good way to mark them.

Yeah, this is awkward and I'm not sure there's a perfect solution here. As I understand it though, new tests being added to 7.17 should be pretty minimal as backports to that branch are kept to critical bug fixes.

nik9000:
[...]

PR #86831, opened 2022-05-17
[Rest Api Compatibility] Enhance error message with ref to dev guide

pgomulka:
When a rest api compatibility test fails a message should be enriched
to include a ref to a developers guide.
The developers guide contains more information on how to troubleshoot
the test failure, where to look for transform tests etc.

relates #85955


test/framework/src/main/java/org/elasticsearch/test/junit/listeners/ReproduceInfoPrinter.java
```diff
@@ -91,9 +91,20 @@ public void testFailure(Failure failure) throws Exception {
...
+            b.append(
+                "This is a Rest Api Compatibility Test. "
+                    + "See the developers guide for details how to troubleshoot - REST_API_COMPATIBILITY.md"
+            );
+        }
         printToErr(b.toString());
     }
 
+    private boolean isRestApiCompatibilityTest(String task) {
+        return task.matches(".*yamlRestTestV\\dCompatTest");
```
mark-vieira, line 105: This is a bit brittle. Perhaps we can use an explicit system property here instead and then inject it in the rest compatibilty gradle plugin.

test/framework/src/main/java/org/elasticsearch/test/junit/listeners/ReproduceInfoPrinter.java
```diff
@@ -91,9 +91,20 @@ public void testFailure(Failure failure) throws Exception {
         GradleMessageBuilder gradleMessageBuilder = new GradleMessageBuilder(b);
         gradleMessageBuilder.appendAllOpts(failure.getDescription());
 
+        if (isRestApiCompatibilityTest(task)) {
+            b.append(System.lineSeparator());
+            b.append(
+                "This is a Rest Api Compatibility Test. "
+                    + "See the developers guide for details how to troubleshoot - REST_API_COMPATIBILITY.md"
```
mark-vieira, line 98: Should we include an actual URL?

pgomulka: you mean a github link ? https://github.com/elastic/elasticsearch/blob/master/REST_API_COMPATIBILITY.md
In that case it should be branch aware if we at some point change it?

test/framework/src/main/java/org/elasticsearch/test/junit/listeners/ReproduceInfoPrinter.java
pgomulka: do we do this for some other tasks?
You mean just setting with System.setProperty in YamlRestCompatTestPlugin ? 
we have to remember to remove it when the tasks finishes. Does this not prevent concurrent task runs?

test/framework/src/main/java/org/elasticsearch/test/junit/listeners/ReproduceInfoPrinter.java
mark-vieira: It's probably not worth being branch aware.

test/framework/src/main/java/org/elasticsearch/test/junit/listeners/ReproduceInfoPrinter.java
mark-vieira: Yes, we use the same pattern for [identifying if this is a BWC test](https://github.com/elastic/elasticsearch/blob/0832232b997fa0e50f8dcf1911307a2e8771a77f/test/framework/src/main/java/org/elasticsearch/test/junit/listeners/ReproduceInfoPrinter.java#L67). We can just add something like a `tests.restCompat` system property when we setup the test task [here](https://github.com/elastic/elasticsearch/blob/0ec229050e5bebdb82a0a273076aae52e56ba520/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/rest/compat/YamlRestCompatTestPlugin.java#L198-L216).

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/rest/compat/YamlRestCompatTestPlugin.java
```diff
@@ -198,6 +198,7 @@ public void apply(Project project) {
         // setup the test task
         Provider<RestIntegTestTask> yamlRestCompatTestTask = RestTestUtil.registerTestTask(project, yamlCompatTestSourceSet, testTaskName);
         project.getTasks().withType(RestIntegTestTask.class).named(testTaskName).configure(testTask -> {
+            testTask.getSystemProperties().put("tests.restCompat", true);
```
mark-vieira, line 201: This can be simplified to just `testTask.systemProperty("tests.restCompat", "true")`

mark-vieira (approved):

pgomulka:
@elasticmachine update branch

Issue #86834, opened 2022-05-17
geo_distance does not support GeoJSON for points


craigtaverner:
### Description

Defining points using GeoJSON has been supported in the geo_shape field type all along, but recently we added GeoJSON support for geo_point and point (geographic and cartesian respectively):

* https://github.com/elastic/elasticsearch/pull/85120
* https://github.com/elastic/elasticsearch/pull/85442

This means that all three field types now consistently support both GeoJSON and WKT for points. However, it turns out that this support does not extend to the `geo_distance` query, which supports all the same point formats as `geo_point` except for GeoJSON. Interestingly, the `geo_bounding_box` query does support GeoJSON for the `top_left` and `bottom_right` points (as well as WKT and other point formats).

Adding GeoJSON to `geo_distance` should close this consistency gap.

PR #86835, opened 2022-05-17
Soft-deprecation of point/geo_point formats

### LINKED ISSUE #47535: Asymmetry between supported point format in geo_point and geo_shape field types
Right now `geo_point` can be defined in multiple formats:

- `"POINT(0 0)"`, WKT
- `{ "lat": 0, "lon": 0}`, lat/lon object
- `[0, 0]`, [lon, lat] array
- `"0, 0"`, (lat, lon) string
- `"gbsuv"`, geohash

Since `geo_shape` supports point types, it may make sense to also support `geo_point` formats as field values for this field type.

relates to #47243, where it would be nice to match either geo_point or geo_shape values.

craigtaverner:
GeoJSON and WKT are now common formats for all three types: `geo_shape`, `geo_point` and `point`

We decided to soft-deprecate the other point formats by prioritizing WKT and GeoJSON in the documentation examples and re-ordering the explicitly defined support for `point` and `geo_point` with GeoJSON and WKT first:

* GeoJSON (object with keys `type` and `coordinates`)
* WKT `POINT(x y)`
* Object with keys `lat` and `lon` (or `x` and `y` for point)
* Array [lon,lat]
* String `"lat,lon"` (or `"x,y"` in point)
* String with geohash (only in `geo_point`)

The geohash is last because it is only supported by one field type `geo_point`.
The string version is second last because it is the most controversial, being the only version to reverse the coordinate order from all other formats (for `geo_point` only, since the coordinates are not reversed in `point`).

In addition we replaced many examples in both documentation and tests to prioritize WKT over the plain string format.

Many remaining examples of array format or object with keys still exist and could be replaced by, for example, GeoJSON, if we feel the need, either in the PR or as a future improvement.

In particular the documentation for the `geo_distance` query could be improved, but that is probably best done as part of the work in https://github.com/elastic/elasticsearch/issues/86834.

Closes https://github.com/elastic/elasticsearch/issues/47535

### DIFF
--- a/docs/reference/aggregations/bucket/composite-aggregation.asciidoc
+++ b/docs/reference/aggregations/bucket/composite-aggregation.asciidoc
@@ -502,8 +502,8 @@ GET /_search
                 "field": "location",
                 "precision": 22,
                 "bounds": {
-                  "top_left": "52.4, 4.9",
-                  "bottom_right": "52.3, 5.0"
+                  "top_left": "POINT (4.9 52.4)",
+                  "bottom_right": "POINT (5.0 52.3)"
                 }
               }
             }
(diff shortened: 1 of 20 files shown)


iverase (approved):
LGTM

sethmlarson (approved):

PR #86837, opened 2022-05-17
[7.17] Use internal-java-rest-test in high level rest client

breskeby:
This is 7.17 specific change before we can backport the deletion of the elasticsearch.rest-test plugin


breskeby:
@elasticsearchmachine test this

PR #86842, opened 2022-05-17
Don't download geoip databases if geoip system index is blocked.

martijnvg:
For example in the case that the a cluster is running out of disk
space and indices reject writes.

This can otherwise load to unnecessary error logs being printed and
just add to more instability. Instead, the GeoIpDownloader should
just try to download the files the next time it runs.


modules/ingest-geoip/src/main/java/org/elasticsearch/ingest/geoip/GeoIpDownloader.java
```diff
@@ -125,6 +126,18 @@ public void setPollInterval(TimeValue pollInterval) {
 
     // visible for testing
     void updateDatabases() throws IOException {
+        var clusterState = clusterService.state();
+        var geoipIndex = clusterState.getMetadata().getIndicesLookup().get(GeoIpDownloader.DATABASES_INDEX);
+        if (geoipIndex != null) {
+            if (clusterState.getRoutingTable().index(geoipIndex.getWriteIndex()).allPrimaryShardsActive() == false) {
+                throw new RuntimeException("not all primary shards of [" + DATABASES_INDEX + "] index are active");
```
dakrone, line 133: Perhaps `ElasticsearchException` instead of a generic `RuntimeException`?

dakrone (approved):
LGTM, left one optional question

PR #86860, opened 2022-05-17
Speed up FieldMapper construction/parsing/serialization

original-brownbear:
Speeding this up some more as it's now 50% of the bootstrap time of the many shards benchmarks.
Iterating an array here in all cases is quite a bit faster than iterating various kinds of lists
and doesn't complicate the code. Also removes a redundant call to `getValue()` for each parameter
during serialization.

relates #77466


javanna (approved):
makes sense to me, not sure why we used a list here. Thanks!

original-brownbear:
Thanks Luca!

PR #86862, opened 2022-05-17
Do not hold the lock during flush in BulkProcessor

### LINKED ISSUE #68468: Generic threads deadlock related to ILMHistoryStore indexing

fcofdez:
This commit releases the BulkProcessor lock after a new bulk request
is created during flush instead of waiting until the request is
dispatched. This prevents a deadlock when the GENERIC thread pool
is saturated with tasks adding new items to the BulkProcessor
waiting to obtain the lock, this might prevent the original
BulkRequest of making progress in certain scenarios, as some
retry operations might be dispatched in the GENERIC thread pool.

Closes #68468

### DIFF
--- a/docs/changelog/86862.yaml
+++ b/docs/changelog/86862.yaml
@@ -1,0 +1,6 @@
+pr: 86862
+summary: Do not hold the lock during flush in `BulkProcessor`
+area: Java High Level REST Client
+type: bug
+issues:
+ - 68468
--- a/server/src/internalClusterTest/java/org/elasticsearch/action/bulk/BulkProcessorIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/action/bulk/BulkProcessorIT.java
@@ -20,10 +20,17 @@ import org.elasticsearch.cluster.metadata.IndexMetadata;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.common.unit.ByteSizeUnit;
 import org.elasticsearch.common.unit.ByteSizeValue;
+import org.elasticsearch.common.util.concurrent.EsThreadPoolExecutor;
 import org.elasticsearch.core.TimeValue;
+import org.elasticsearch.plugins.Plugin;
 import org.elasticsearch.test.ESIntegTestCase;
+import org.elasticsearch.test.transport.MockTransportService;
+import org.elasticsearch.threadpool.ThreadPool;
+import org.elasticsearch.transport.ConnectTransportException;
+import org.elasticsearch.transport.TransportService;
 
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -43,6 +50,11 @@ import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.lessThanOrEqualTo;
 
 public class BulkProcessorIT extends ESIntegTestCase {
+
+    @Override
+    protected Collection<Class<? extends Plugin>> nodePlugins() {
+        return List.of(MockTransportService.TestPlugin.class);
+    }
 
     public void testThatBulkProcessorCountIsCorrect() throws Exception {
         final CountDownLatch latch = new CountDownLatch(1);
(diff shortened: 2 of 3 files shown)


fcofdez:
I'm not sure if these are the correct labels, I was reviewing this comment https://github.com/elastic/elasticsearch/pull/85490#discussion_r868421363 and I decided to try to reproduce this issue and fix it. Maybe it makes sense to solve it differently, but I guess that having the test reproducing the issue is still valuable.

DaveCTurner (commented):
One effect of the lock here is to apply backpressure on the writing threads if indexing is slow. I think this change removes that backpressure which could lead to building up an unbounded number of pending items.

In the case of the ILM history store there was some discomfort about dropping entries from history in the face of such backpressure, but the only alternative is to slow down ILM itself which also significant drawbacks. I don't know that we made a decision on this topic yet.

fcofdez:
> One effect of the lock here is to apply backpressure on the writing threads if indexing is slow. I think this change removes that backpressure which could lead to building up an unbounded number of pending items.

I see your point but that backpressure is not guaranteed if the bulk request is flushed after the maximum size / number of docs is reached as the lock is released before sending the request. Additionally, in the ILM history context we're effectively queing an unlimited number of documents since we dispatch the `BulkProcessor#add` to the `GENERIC` thread pool, right? I guess it would make sense to clarify what's the expected behaviour of this class and rewrite it to handle better these scenarios.

DaveCTurner:
Yes that's right, we are not really respecting backpressure with ILM history entries anyway thanks to the unbounded queue on the `GENERIC` threadpool. It'd be great to make that queue bounded too, but that's a separate (and much bigger) project. Still, I think it would be better not to introduce a new (and hidden) unbounded queue here.

jakelandis:
> Still, I think it would be better not to introduce a new (and hidden) unbounded queue here.

Which queue is unbounded here ? It is kinda hard to follow since the scheduler is a `org.elasticsearch.threadpool.Scheduler` and could find any docs that guarantee (for/against) what happens if two schedules overlap. (i.e. flush 1 takes a long time and flush 2 needs to kick off). Is it the scheduler's queue that backs up or does the scheduler allow concurrent schedules meaning that multiple GENERIC threads would be inflight and blocked by the semaphor of BulkRequestHandler ? (Also the GENERIC thread pool is bounded, just a very high bound ... if I am reading correctly)

If we were using Java's `ScheduledThreadPoolExecutor#scheduleAtFixedRate` the docs there are explicit state "If any execution of this task takes longer than its period, then subsequent executions may start late, but will not concurrently execute." ...so at worse we would back up the scheduler and have more control over the queue and rejections for this specific flush case. Do we need to use the custom scheduler for thread context stuff for security ? 

At a higher level...what happens when the writes are very slow the flushes are very frequent ?

jakelandis (commented):
changes look good, just a couple nitpics.

DaveCTurner:
> Which queue is unbounded here ? It is kinda hard to follow since the scheduler is a org.elasticsearch.threadpool.Scheduler and could find any docs that guarantee (for/against) what happens if two schedules overlap. (i.e. flush 1 takes a long time and flush 2 needs to kick off). Is it the scheduler's queue that backs up or does the scheduler allow concurrent schedules meaning that multiple GENERIC threads would be inflight and blocked by the semaphor of BulkRequestHandler ?

Our `Scheduler` is really [a `ScheduledThreadPoolExecutor` with a pool size of 1](https://github.com/elastic/elasticsearch/blob/12ad399c488f0cc60e19b5e1b29c6d569cb4351a/server/src/main/java/org/elasticsearch/threadpool/Scheduler.java#L45-L49); [these docs](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/ScheduledThreadPoolExecutor.html) indicate that this executor does not create extra threads but does have an unbounded queue:

> While this class inherits from [ThreadPoolExecutor](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/concurrent/ThreadPoolExecutor.html), a few of the inherited tuning methods are not useful for it. In particular, because it acts as a fixed-sized pool using `corePoolSize` threads and an unbounded queue, adjustments to `maximumPoolSize` have no useful effect.

However the scheduled flush tasks all immediately fork to the `GENERIC` pool so there can be (boundedly) many of them running at once, and (unboundedly) many of them enqueued.

> (Also the GENERIC thread pool is bounded, just a very high bound ... if I am reading correctly)

It comes from `EsExecutors#newFixed` so the number of threads is bounded but its queue is unbounded.

> At a higher level...what happens when the writes are very slow the flushes are very frequent and the concurrency of BulkRequestHandler is 1?

Right, this is exactly the point. Today if writes get too far behind we block the threads calling `add()` which slows down the incoming work. With this change we remove that backpressure mechanism and allow the work to build up in memory without bound.

fcofdez:
> However the scheduled flush tasks all immediately fork to the GENERIC pool so there can be (boundedly) many of them running at once, and (unboundedly) many of them enqueued.

Just as a note, this depends on how `BulkProcessor` is configured. If the `ThreadPool` is used it dispatches the flush task to the `GENERIC` pool, otherwise the flush task is executed in the scheduler thread that is created for the `BulkProcessor`.

DaveCTurner:
Oh right I see, it ignores the `executor` parameter in that case. Tricky. Still, it's the same situation just with fewer threads involved right?

PR #86863, opened 2022-05-17
[Refactor] Make PluginBundle package-private

williamrandolph:
Our plugin loading code uses the `PluginBundle` class to describe a plugin on disk. This should be an implementation detail of the plugin loading code, and not part of the API shared with the plugin CLI. This refactoring moves static methods from the plugin CLI to the `PluginsUtils` class, makes `PluginBundle` package-private, and reduces the visibility of `PluginsUtils` static methods where possible.


rjernst (approved):
LGTM

PR #86868, opened 2022-05-17
Script: Fix setter shortcut for unbridged setters

stu-elastic:
The setter shortcut resolution worked for bridged setter methods, in `PainlessLookupBuilder.cacheRuntimeHandles`, but did not work for other setters.


rjernst (approved):
LGTM

PR #86883, opened 2022-05-18
[TEST] Accept warning about multiple matching templates

javanna:
This PR fixes a recent test failure caused by multiple matching legacy templates: https://gradle-enterprise.elastic.co/s/usz3awka53gii/console-log?task=:rest-api-spec:yamlRestTest .


craigtaverner (approved):
LGTM

javanna:
@dakrone are you ok with this approach or can you point me to an alternative solution?

dakrone:
@javanna yep, this is normally the way we handle this, if you search the codebase for "`index template .* [global]`" you can see we've added a lot of them for the occasionally injected `global` template, which is a similar situation.

PR #86885, opened 2022-05-18
installation on windows must be done as admin

magorbalassy:
As reported by users, the `.\bin\elasticsearch.bat` command must be executed as an administrator, but this is not 100% clear from the documentation. Thus submitting this pull request adding this detail.

### DIFF
--- a/docs/reference/setup/install/zip-windows-start.asciidoc
+++ b/docs/reference/setup/install/zip-windows-start.asciidoc
@@ -1,6 +1,6 @@
 ==== Run {es} from the command line
 
-Run the following command to start {es} from the command line:
+Run as an administrator user the following command to start {es} from the command line:
 
 [source,sh]
 ----


magorbalassy:
When the batch file is executed as a normal user, the person will be unable to create an enrollment token., getting the following  errors:

1. Enrollment with user account
Enduser tried to enroll with a user account but they get this **error** when starting kibana:
```
[2022-05-02T11:00:04.615+02:00][ERROR][elasticsearch-service] Unable to retrieve version information from Elasticsearch nodes. security_exception: [security_exception] Reason: unable to authenticate user [elasticsearchuser] for REST request [/_nodes?filter_path=nodes.*.version%2Cnodes.*.http.publish_address%2Cnodes
```

2. When trying to use the `elasticsearch-create-enrollment-token` , Call `C:\elastic\elasticsearch-8.1.3\bin>elasticsearch-create-enrollment-token -s kibana --url "http://localhost:9200/`
User getting the following error:
```
**ERROR**: Unable to create an enrollment token. Elasticsearch node HTTP layer SSL configuration is not configured with a keystore 
```
When installed as admin, these errors aren't present.

When these above steps are performed as a non-admin user, the TLS is not configured and no enrollment token is generated and the only notification is a warning about access to the folder.

mark-vieira:
Perhaps someone from @elastic/es-security can provide some insight here. Should admin rights be required for creating an enrollment token? Is this something we should add to the documentation as suggested?

evilwon22:
SBI. Are investigating now.

ywangd:
> the only notification is a warning about access to the folder.

The initial security auto-configuration needs write permission to the `config` folder (so not necessary adminstrator) Otherwise, you should see a message that says
> Could not create auto-configuration directory

But it's not a warning. Maybe we could make this message stand out a bit more.

When that happens,  the node will start with no TLS configured. That seems to match the provided error about "SSL is not configured with a keystore".

@albertzaharovits for more insights.

mark-vieira:
So the bring this around, this isn't really Windows-specific at all. You'd get similar errors on any platform if the user running Elasticsearch cannot write to the `config` directory. Should we make that more obvious in the error messages if that's not already the case?

mark-vieira:
So I think we have a few things here:

1. When we can't auto-configure security we continue on and it's easy to miss that SSL wasn't setup.
2. When creating an enrollment token after auto-security fails the failure message is pretty confusing.
  ```
  **ERROR**: Unable to create an enrollment token. Elasticsearch node HTTP layer SSL configuration is not configured with a keystore 
  ```
3. Starting ES with an enrollment token requires write permissions to the config directory.

There's probably some usability/documentation improvements here that could be done. Perhaps we should capture these in separate issues and close out this PR since running as an admin isn't strictly necessary or even the best solution here.

leemthompo:
> [!IMPORTANT]  
> Elastic documentation is migrating to Markdown for version 9.0+. See the [migration guide](https://elastic.github.io/docs-builder/migration/freeze/index.html) for details.


## ℹ️ **What's happening?**
- Starting January 29, we will start closing all unmerged documentation PRs targeting main/master
- We're migrating from AsciiDoc to Markdown for 9.0+
- 9.0 docs will be frozen from January 29 until February 20 2024
- **NOTE**: PRs that include both code and documentation changes will remain open

### **What do I need to do?**

#### For **<=8.x docs**:
 1. Rebase your PR to target the relevant 8.x branch instead
 2. The content can remain in AsciiDoc format

#### For **9.0+ docs**:

_Option 1_:
 - Draft docs in Markdown
 - Once migration freeze ends, find the relevant page in the new docs system and use the edit options to submit your changes


_Option 2_:
- Open an issue in [elastic/docs-content](https://github.com/elastic/docs-content)

### 💡 **Need help?**
1. For Elasticians: Ask in **#docs** Slack channel
2. For external contributors: Open an issue in [elastic/docs-content](https://github.com/elastic/docs-content)

PR #86891, opened 2022-05-18
[Refactor] Move plugin setting merge method from PluginsService to Node

williamrandolph:
The `PluginsService#updatedSettings` method was only used once in the `Node` constructor, so it makes sense to make that a utility method on the `Node` class. We make it testable by factoring out the PluginsService, instead providing a map of names to plugins as an argument.


rjernst:
This seems very similar to the map and flatMap methods you added recently. The reason for exposing the raw map of name to plugin seems to be just for error cases, so that we can give a helpful error message. I wonder if map and flatMap should be reworked to include the plugin name somehow, so that similar error conditions can be better logged?

grcevski (approved):
LGTM!

williamrandolph:
@rjernst That's a very good point. I'm going to merge this PR so we don't have to re-review the test code here, but I can put out a PR that lets us use `BiFunction` or `BiConsumer` to access plugin names.

PR #67705, opened 2021-01-19
Support max_single_primary_size in Resize Action and exposed in ILM

gaobinlong:
Relates to #65714.

The main changes are:
1. Add a `max_single_primary_size` parameter in Resize Action, it's used to calculate an appropriate shards number of the target index when executing the Shrink API.
2. Expose the `max_single_primary_size` parameter in ILM's Shrink Action.
3. Support the parameter in high level rest client.
4. Add some description for the parameter in the documents.
5. Add some tests for the code change.

PR #86897, opened 2022-05-18
Fix max_primary_shard_size resize factor math

joegallo:
There's an off-by-one bug in #67705, such that the logic doesn't handle all factors correctly in some circumstances.

For example, `calculateAcceptableNumberOfShards(9, 2)` would return 9 rather than 3, because it would evaluate all potential factors up to but not including 3 (`<`), and not finding such a factor, it would return 9. The correct logic should evaluate all potential factors up to and including 3 (`<=`), and return 3.


dakrone (approved):
LGTM, thanks for fixing this.

joegallo:

PR #86873, opened 2022-05-17
Revert "Enable HLRC compatibility mode by default (#86517)"

sethmlarson:
This PR reverts https://github.com/elastic/elasticsearch/pull/86517 which set the compatibility mode to always be enabled for the Java HLRC pending a deeper investigation into the impacts.

It looks like there were some documentation changes that *could* have been left but in the interest of getting a clean timely revert we'll go this route and add the documentation changes separately.

PR #86900, opened 2022-05-18
[docs] Explain and advertise HLRC compatibility mode

swallez:
Adds documentation about compatibility mode in HLRC that allows it to work with ES 8.x, and add a note linking to that documentation to the index page to make sure users are aware of it.

This PR is the docs part of PR #86517 that we decided to revert in PR #86873.


sethmlarson (approved):
LGTM

PR #79837, opened 2021-10-26
Speed up ReplicationTracker Logic on Data Nodes a Little

original-brownbear:
This stuff shows up a little in profiling when the data-node holds a lot of shards.
Mainly just avoiding extra loops in `calculateReplicationGroup` (biggest cost in profiling), allocating streams to iterate 2 steps and some other indirection that starts to hurt if executed over thousands of shards.
Also, we were submitting a lot of empty tasks to the generic pool at times, which temporarily showed up in profiling.

I think we can make bigger improvements here and run this logic more selectively, it seems a non-trivial number of noops are executed here.

relates #77466

PR #86903, opened 2022-05-18
Speed up building RoutingTable from RoutingNodes

original-brownbear:
We do the routing nodes -> routing table step during reroute and it's a significant contributor
to the runtime of reroute.
The PR speeds it up by more than double in the common one shard + one primary case by saving the expensive
and slow-to-iterate hashmap keyed by shard id, the needless pre-computation of the allocation ids set (it's only used in an assertion and in the replica tracker but it was the most expensive part in that constructor) and a redundant round of building `IndexShardRoutingTable` in `addShard`.

This gives us another 2-3% speedup on the initial indices bootstrap in the many-shards benchmark.
Relates #77466

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
@@ -418,7 +418,7 @@ public class IndexRoutingTable implements SimpleDiffable<IndexRoutingTable> {
             shards = new IndexShardRoutingTable.Builder[indexMetadata.getNumberOfShards()];
             for (int shardNumber = 0; shardNumber < indexMetadata.getNumberOfShards(); shardNumber++) {
                 ShardId shardId = new ShardId(index, shardNumber);
-                IndexShardRoutingTable.Builder indexShardRoutingBuilder = new IndexShardRoutingTable.Builder(shardId);
+                IndexShardRoutingTable.Builder indexShardRoutingBuilder = IndexShardRoutingTable.builder(shardId);
                 for (int i = 0; i <= indexMetadata.getNumberOfReplicas(); i++) {
                     boolean primary = i == 0;
                     if (asNew && ignoreShards.contains(shardNumber)) {
@@ -469,7 +469,7 @@ public class IndexRoutingTable implements SimpleDiffable<IndexRoutingTable> {
                     // a freshly created index with no restriction
                     primaryRecoverySource = EmptyStoreRecoverySource.INSTANCE;
                 }
-                IndexShardRoutingTable.Builder indexShardRoutingBuilder = new IndexShardRoutingTable.Builder(shardId);
+                IndexShardRoutingTable.Builder indexShardRoutingBuilder = IndexShardRoutingTable.builder(shardId);
                 for (int i = 0; i <= indexMetadata.getNumberOfReplicas(); i++) {
                     boolean primary = i == 0;
                     indexShardRoutingBuilder.addShard(
@@ -487,13 +487,9 @@ public class IndexRoutingTable implements SimpleDiffable<IndexRoutingTable> {
         }
 
         public Builder addReplica() {
-            if (shards == null) {
-                return this;
-            }
+            assert shards != null;
             for (IndexShardRoutingTable.Builder existing : shards) {
-                if (existing == null) {
-                    continue;
-                }
+                assert existing != null;
                 // version 0, will get updated when reroute will happen
                 existing.addShard(
                     ShardRouting.newUnassigned(
@@ -508,21 +504,17 @@ public class IndexRoutingTable implements SimpleDiffable<IndexRoutingTable> {
         }
 
         public Builder removeReplica() {
-            if (shards == null) {
-                return this;
-            }
+            assert shards != null;
             for (int shardId = 0; shardId < shards.length; shardId++) {
                 IndexShardRoutingTable.Builder found = shards[shardId];
-                if (found == null) {
-                    continue;
-                }
+                assert found != null;
                 final IndexShardRoutingTable indexShard = found.build();
                 if (indexShard.replicaShards().isEmpty()) {
                     // nothing to do here!
                     return this;
                 }
                 // re-add all the current ones
-                IndexShardRoutingTable.Builder builder = new IndexShardRoutingTable.Builder(indexShard.shardId());
+                IndexShardRoutingTable.Builder builder = IndexShardRoutingTable.builder(indexShard.shardId());
                 for (int copy = 0; copy < indexShard.size(); copy++) {
                     ShardRouting shardRouting = indexShard.shard(copy);
                     builder.addShard(shardRouting);
@@ -565,14 +557,14 @@ public class IndexRoutingTable implements SimpleDiffable<IndexRoutingTable> {
             ensureShardArray(shardId);
             IndexShardRoutingTable.Builder indexShard = shards[shardId];
             if (indexShard == null) {
-                shards[shardId] = new IndexShardRoutingTable.Builder(shard.shardId()).addShard(shard);
+                shards[shardId] = IndexShardRoutingTable.builder(shard.shardId()).addShard(shard);
             } else {
                 indexShard.addShard(shard);
             }
             return this;
         }
 
-        private void ensureShardArray(int shardId) {
+        void ensureShardArray(int shardId) {
             if (shards == null) {
                 shards = new IndexShardRoutingTable.Builder[shardId + 1];
             } else if (shards.length < shardId + 1) {
--- a/server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
@@ -298,7 +298,7 @@ public class IndexRoutingTable implements SimpleDiffable<IndexRoutingTable> {
         Builder builder = new Builder(index);
 
         int size = in.readVInt();
-        builder.ensureShardArray(size - 1);
+        builder.ensureShardArray(size);
         for (int i = 0; i < size; i++) {
             builder.addIndexShard(IndexShardRoutingTable.Builder.readFromThin(in, index));
         }
@@ -542,7 +542,7 @@ public class IndexRoutingTable implements SimpleDiffable<IndexRoutingTable> {
             assert indexShard.shardId().getIndex().equals(index)
                 : "cannot add shard routing table for " + indexShard.shardId() + " to index routing table for " + index;
             final int sid = indexShard.shardId().id();
-            ensureShardArray(sid);
+            ensureShardArray(sid + 1);
             shards[sid] = indexShard;
             return this;
         }
@@ -554,7 +554,7 @@ public class IndexRoutingTable implements SimpleDiffable<IndexRoutingTable> {
         public Builder addShard(ShardRouting shard) {
             assert shard.index().equals(index) : "cannot add [" + shard + "] to routing table for " + index;
             int shardId = shard.id();
-            ensureShardArray(shardId);
+            ensureShardArray(shardId + 1);
             IndexShardRoutingTable.Builder indexShard = shards[shardId];
             if (indexShard == null) {
                 shards[shardId] = IndexShardRoutingTable.builder(shard.shardId()).addShard(shard);
@@ -564,11 +564,11 @@ public class IndexRoutingTable implements SimpleDiffable<IndexRoutingTable> {
             return this;
         }
 
-        void ensureShardArray(int shardId) {
+        void ensureShardArray(int shardCount) {
             if (shards == null) {
-                shards = new IndexShardRoutingTable.Builder[shardId + 1];
-            } else if (shards.length < shardId + 1) {
-                IndexShardRoutingTable.Builder[] updated = new IndexShardRoutingTable.Builder[shardId + 1];
+                shards = new IndexShardRoutingTable.Builder[shardCount];
+            } else if (shards.length < shardCount) {
+                IndexShardRoutingTable.Builder[] updated = new IndexShardRoutingTable.Builder[shardCount];
                 System.arraycopy(shards, 0, updated, 0, shards.length);
                 shards = updated;
             }
--- a/server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
@@ -28,10 +28,8 @@ import org.elasticsearch.index.shard.ShardId;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.function.Predicate;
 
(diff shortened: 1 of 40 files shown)


server/src/main/java/org/elasticsearch/cluster/routing/IndexShardRoutingTable.java
```diff
@@ -477,6 +471,15 @@ public ShardRouting getByAllocationId(String allocationId) {
     }
 
     public Set<String> getAllAllocationIds() {
+        Set<String> allAllocationIds = new HashSet<>();
+        for (ShardRouting shard : shards) {
```
original-brownbear, line 475: The master node doesn't use this field anyway ... if we want faster here on the data node I have an open PR for that  https://github.com/elastic/elasticsearch/pull/79837 :)

original-brownbear:
Jenkins run elasticsearch-ci/part-1 (unrelated)

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -481,29 +481,56 @@ private Builder initializeEmpty(IndexMetadata indexMetadata, UnassignedInfo unas
...
+                if (existing == null) {
+                    continue;
+                }
+                if (existing instanceof IndexShardRoutingTable) {
+                    shards[shardNumber] = new IndexShardRoutingTable.Builder((IndexShardRoutingTable) existing).addShard(
+                        newUnassignedReplica(((IndexShardRoutingTable) existing).shardId())
+                    ).build();
+                } else {
+                    assert existing instanceof IndexShardRoutingTable.Builder;
+                    // version 0, will get updated when reroute will happen
```
henningandersen, line 504: I'd think this comment also applies in the case of `IndexShardRoutingTable`?

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -328,7 +326,7 @@ public static Builder builder(Index index) {
     public static class Builder {
 
         private final Index index;
-        private final Map<Integer, IndexShardRoutingTable> shards = new HashMap<>();
+        private Object[] shards;
```
henningandersen, line 329: Could we make this a `IndexShardRoutingTable.Builder[]` if we made deserialize not use the builder? The other usage I found was `RoutingTable.updateNumberOfReplicas`, but that seems to create builders anyway.

That would avoid the instanceof and casting introduced here.

server/src/main/java/org/elasticsearch/cluster/routing/IndexShardRoutingTable.java
henningandersen: Can we assert that it is not called on the master thread with a suitable comment? That could document this optimization.

server/src/main/java/org/elasticsearch/cluster/routing/RoutingTable.java
```diff
@@ -385,6 +384,31 @@ public void writeTo(StreamOutput out) throws IOException {
         }
     }
 
+    public static RoutingTable of(long version, RoutingNodes routingNodes) {
```
henningandersen, line 387: Can we add a test of this new method?

henningandersen (commented):
This looks mostly good to me though I'd like to avoid the instanceof and casting in this central code.

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -481,29 +481,56 @@ private Builder initializeEmpty(IndexMetadata indexMetadata, UnassignedInfo unas
...
+                return this;
+            }
+            for (int shardId = 0; shardId < shards.length; shardId++) {
+                Object found = shards[shardId];
+                if (found == null) {
+                    continue;
+                }
+                final IndexShardRoutingTable indexShard = found instanceof IndexShardRoutingTable
+                    ? (IndexShardRoutingTable) found
+                    : ((IndexShardRoutingTable.Builder) found).build();
```
idegtiarenko, line 533: It looks like below this is converted back to the builder. Is it possible to re-use existing builder instead?

server/src/main/java/org/elasticsearch/cluster/routing/RoutingTable.java
original-brownbear: Sure added :)

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
original-brownbear: This complexity is now all gone thanks to moving to an array of builders :)

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
original-brownbear: Good point and this turned out to be really trivial to do actually (apart from some diff noise in tests) because the deserialization was creating builders anyway.
=> done :)

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
original-brownbear: Only the builder version is left :) -> problem solved

original-brownbear:
Thanks Henning + Ievgen, all points addressed now I think :)

server/src/main/java/org/elasticsearch/cluster/routing/IndexShardRoutingTable.java
original-brownbear: Sure, added that assertion.

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -481,29 +481,42 @@ private Builder initializeEmpty(IndexMetadata indexMetadata, UnassignedInfo unas
...
-                shards.put(shardNumber, indexShardRoutingBuilder.build());
+                shards[shardNumber] = indexShardRoutingBuilder;
             }
             return this;
         }
 
         public Builder addReplica() {
-            for (var shardNumber : shards.keySet()) {
-                ShardId shardId = new ShardId(index, shardNumber);
+            if (shards == null) {
```
henningandersen, line 490: AFAICS, we could instead

`assert shards != null`?

Likewise for the null check in the loop?

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -481,29 +481,42 @@ private Builder initializeEmpty(IndexMetadata indexMetadata, UnassignedInfo unas
...
                 );
-                shards.put(shardNumber, new IndexShardRoutingTable.Builder(shards.get(shard.id())).addShard(shard).build());
             }
             return this;
         }
 
         public Builder removeReplica() {
-            for (var shardId : shards.keySet()) {
-                IndexShardRoutingTable indexShard = shards.get(shardId);
+            if (shards == null) {
```
henningandersen, line 511: Like for `addReplica`, I think we always have this filled in here and can instead `assert shards != null`?

server/src/main/java/org/elasticsearch/cluster/routing/RoutingTable.java
```diff
@@ -470,7 +464,7 @@ public Builder updateNumberOfReplicas(final int numberOfReplicas, final String[]
                 IndexRoutingTable.Builder builder = new IndexRoutingTable.Builder(indexRoutingTable.getIndex());
                 // re-add all the shards
                 for (int i = 0; i < indexRoutingTable.size(); i++) {
-                    builder.addIndexShard(indexRoutingTable.shard(i));
+                    builder.addIndexShard(new IndexShardRoutingTable.Builder(indexRoutingTable.shard(i)));
```
henningandersen, line 467: Should we do `builder.ensureShardArray(indexRoutingTable.size()-1)` before this loop?

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/store/NativeRolesStoreTests.java
```diff
@@ -300,10 +300,17 @@ private ClusterState getClusterStateWithSecurityIndex() {
...
         );
-        IndexShardRoutingTable table = new IndexShardRoutingTable.Builder(new ShardId(index, 0)).addShard(
-            shardRouting.initialize(randomAlphaOfLength(8), null, shardRouting.getExpectedShardSize()).moveToStarted()
-        ).build();
-        RoutingTable routingTable = RoutingTable.builder().add(IndexRoutingTable.builder(index).addIndexShard(table).build()).build();
+        RoutingTable routingTable = RoutingTable.builder()
+            .add(
+                IndexRoutingTable.builder(index)
+                    .addIndexShard(
+                        new IndexShardRoutingTable.Builder(new ShardId(index, 0)).addShard(
```
henningandersen, line 307: Can we add `IndexShardRoutingTable.builder(ShardId)` and use instead? This looks a bit odd here given the two different styles.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/support/SecurityIndexManagerTests.java
```diff
@@ -142,11 +142,20 @@ public void testIndexWithoutPrimaryShards() {
...
-            shardRouting.initialize(nodeId, null, shardRouting.getExpectedShardSize())
-                .moveToUnassigned(new UnassignedInfo(UnassignedInfo.Reason.ALLOCATION_FAILED, ""))
-        ).build();
-        clusterStateBuilder.routingTable(RoutingTable.builder().add(IndexRoutingTable.builder(index).addIndexShard(table).build()).build());
+        clusterStateBuilder.routingTable(
+            RoutingTable.builder()
+                .add(
+                    IndexRoutingTable.builder(index)
+                        .addIndexShard(
+                            new IndexShardRoutingTable.Builder(new ShardId(index, 0)).addShard(
```
henningandersen, line 150: Also use `IndexShardRoutingTable.builder(ShardId)` here.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/test/SecurityTestUtils.java
```diff
@@ -74,10 +74,17 @@ public static RoutingTable buildIndexRoutingTable(Index index) {
...
         String nodeId = ESTestCase.randomAlphaOfLength(8);
-        IndexShardRoutingTable table = new IndexShardRoutingTable.Builder(new ShardId(index, 0)).addShard(
-            shardRouting.initialize(nodeId, null, shardRouting.getExpectedShardSize()).moveToStarted()
-        ).build();
-        return RoutingTable.builder().add(IndexRoutingTable.builder(index).addIndexShard(table).build()).build();
+        return RoutingTable.builder()
+            .add(
+                IndexRoutingTable.builder(index)
+                    .addIndexShard(
+                        new IndexShardRoutingTable.Builder(new ShardId(index, 0)).addShard(
```
henningandersen, line 81: Also use `IndexShardRoutingTable.builder(ShardId)` here.

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -328,7 +326,7 @@ public static Builder builder(Index index) {
     public static class Builder {
 
         private final Index index;
-        private final Map<Integer, IndexShardRoutingTable> shards = new HashMap<>();
+        private IndexShardRoutingTable.Builder[] shards;
```
henningandersen, line 329: I think the primary reason we use an array rather than a List here is to size it properly, in particular for the 1 shard case, perhaps we should comment on that?

henningandersen (commented):
Thanks for improving this, looks great. I added a number of more detailed comments.

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
original-brownbear: ++ done

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
original-brownbear: ++ done

server/src/main/java/org/elasticsearch/cluster/routing/RoutingTable.java
original-brownbear: Yea that sounds like it should save an allocation here and there :) Added it

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authz/store/NativeRolesStoreTests.java
original-brownbear: ++ done

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/support/SecurityIndexManagerTests.java
original-brownbear: ++ done

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/test/SecurityTestUtils.java
original-brownbear: ++ done

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
henningandersen: I am not sure I follow. You have `ensureShardArray` anyway, which could be translated directly into `ArrayList.ensureCapacity`? It could avoid the potential degenerate case of copying data over and over if growing by one too many times.

But let us leave this out, was a minor point for sure.

original-brownbear: > You have ensureShardArray anyway, which could be translated directly into ArrayList.ensureCapacity?

It doesn't translate like that. `ensureCapacity` still doesn't change the list size. Even if I do `ensureCapacity(2)` I don't automatically get the ability to add at index `1`, I still need to put a `null` into index `0` to be able to do that without an index out of bounds exception :)

henningandersen (approved):
LGTM.

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -546,18 +553,38 @@ public Builder addIndexShard(IndexShardRoutingTable indexShard) {
...
         public IndexRoutingTable build() {
-            return new IndexRoutingTable(index, shards);
+            final IndexShardRoutingTable[] res;
+            if (shards != null) {
+                res = new IndexShardRoutingTable[shards.length];
+                for (int i = 0; i < shards.length; i++) {
+                    res[i] = shards[i].build();
+                }
+            } else {
+                res = new IndexShardRoutingTable[0];
```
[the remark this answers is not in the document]
original-brownbear: I'm not sure, I don't really see how it would be. But annoyingly we have a couple of tests that trigger this with an empty builder randomly. I put down a todo to look into cleaning this up. Maybe we can become a little cleverer here still anyway with how we handle the array :)

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -546,18 +553,38 @@ public Builder addIndexShard(IndexShardRoutingTable indexShard) {
...
+                shards[shardId] = IndexShardRoutingTable.builder(shard.shardId()).addShard(shard);
             } else {
-                indexShard = new IndexShardRoutingTable.Builder(indexShard).addShard(shard).build();
+                indexShard.addShard(shard);
             }
-            shards.put(indexShard.shardId().id(), indexShard);
             return this;
         }
 
+        void ensureShardArray(int shardId) {
```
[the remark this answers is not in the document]
original-brownbear: Fair point moved this to using size as the parameter now :)

original-brownbear:
Thanks Ievgen+Henning!

PR #86904, opened 2022-05-18
Health api add indicator doc links

### LINKED ISSUE #86892: Add indicator level troubleshooting URL when the status is RED/YELLOW
This is to include a generic troubleshooting link for the indicators in case drill down is NOT performed 
(and initially even if the drill down is performend but we don't have user actions - the initial release only the shards_availability indicator will have user actions)

The following links should be advertised:
https://ela.st/fix-shard-allocation for `shards_availability`
https://ela.st/fix-ilm for `ilm`
https://ela.st/fix-slm for `slm`
[ela.st/fix-master](http://ela.st/fix-master) for `instance_has_master`
https://ela.st/fix-repository-integrity for `repository_integrity`

jbaiera:
This PR adds a number of documentation links to health indicators. An indicator's help URL will be displayed when a health indicator returns a non-GREEN result and no user actions (either due to none being available or because they are not generated due to lack of drill down).

Closes #86892


andreidan (approved):
Thanks for working on this Jimmy.

Left a few very minor suggestions. Thanks!

jbaiera:
@elasticmachine update branch

PR #86906, opened 2022-05-18
[Refactor] Extract method for mandatory plugins check

williamrandolph:
We had some unit tests that checked whether classpath plugins matched a list of mandatory plugins. Since we shouldn't be using classpath plugins in production, I don't know how useful this is as a check, but this PR factors out a `checkMandatoryPlugins` method so we can keep test coverage without using the classpath plugin code path, which will soon be moved to `MockNode`.

I also couldn't resist using set logic instead of for-loop list building for the check itself.


rjernst (approved):
LGTM

PR #86907, opened 2022-05-18
New testcases for SearchSortValues

SSSSam141:
This PR is to add new test cases for SearchSortValues. The existing testcases does not cover all the functions in SearchSortValues and this PR tries to add test cases to test for the getRawSortValues() the return value and also test for the override equal() methods for SearchSortValues.


jtibshirani:
Hello @SSSSam141, could you please explain the motivation for this change (and include it in the PR description)? Also, it looks like the PR contains some accidental/ extra changes that should be removed.

SSSSam141:
> Hello @SSSSam141, could you please explain the motivation for this change (and include it in the PR description)? Also, it looks like the PR contains some accidental/ extra changes that should be removed.

SSSSam141:
Extra changes have been removed.

PR #86909, opened 2022-05-18
Fix clearing of lastSuccessfulAuthCache when clear all realm cache API is called

### LINKED ISSUE #86650: Clear all realm cache should also invalidate lastSuccessfulAuthCache
The [ClearRealmCache API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-clear-cache.html) evicts users from caching realms. It should also clear the [lastSuccessfulAuthCache](https://github.com/elastic/elasticsearch/blob/fcfb04392cfdf6f3ae35587ce0f4ed6522cef05a/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/AuthenticationService.java#L97) in the process. 

But it fails to do that if the realm is specified as a wildcard, i.e. the following request clear all realm caches, but not the lastSuccessfulAuthCache ([relevant code](https://github.com/elastic/elasticsearch/blob/fcfb04392cfdf6f3ae35587ce0f4ed6522cef05a/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/realm/TransportClearRealmCacheAction.java#L85-L90)):

```
POST _security/realm/*/_clear_cache
```

But it does clear lastSuccessfulAuthCache if concrete realm name is specified, e.g. the following request clear both the reserved realm cache and the lastSuccessfulAuthCache ([relevant code](https://github.com/elastic/elasticsearch/blob/fcfb04392cfdf6f3ae35587ce0f4ed6522cef05a/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/realm/TransportClearRealmCacheAction.java#L92-L100)):

```
POST _security/realm/reserved/_clear_cache
```

#### Steps to reproduce the issue
1. Enable both file and native realm on a node and make sure file is before native in the realm chain
2. Create a native user `foo` and authenticate with it.
3. Add a file user `foo` with the same password and wait for the node to auto-load the file
4. Clear realm cache with `POST _security/realm/*/_clear_cache`
5. Authenticate with `foo`, but the realm is still `native` (_incorrect behaviour_)
6. Clear the realm cache with `POST _security/realm/reserved/_clear_cache`
7. Authenticate with `foo`, this time the realm is `file` (_correct behaviour_)

slobodanadamovic:
This PR adds missing call to clear last successful authentication cache,
when clear all realm cache action is called.

Fixes #86650

### DIFF
--- a/x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/realm/TransportClearRealmCacheActionTests.java
+++ b/x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/realm/TransportClearRealmCacheActionTests.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0; you may not use this file except in compliance with the Elastic License
+ * 2.0.
+ */
+
+package org.elasticsearch.xpack.security.action.realm;
+
+import org.elasticsearch.Version;
+import org.elasticsearch.action.support.ActionFilters;
+import org.elasticsearch.cluster.node.DiscoveryNode;
+import org.elasticsearch.cluster.service.ClusterService;
+import org.elasticsearch.tasks.Task;
+import org.elasticsearch.test.ESTestCase;
+import org.elasticsearch.threadpool.ThreadPool;
+import org.elasticsearch.transport.TransportService;
+import org.elasticsearch.xpack.core.security.action.realm.ClearRealmCacheRequest;
+import org.elasticsearch.xpack.core.security.action.realm.ClearRealmCacheResponse;
+import org.elasticsearch.xpack.core.security.authc.Realm;
+import org.elasticsearch.xpack.core.security.authc.RealmConfig;
+import org.elasticsearch.xpack.core.security.authc.support.CachingRealm;
+import org.elasticsearch.xpack.security.authc.AuthenticationService;
+import org.elasticsearch.xpack.security.authc.Realms;
+import org.junit.Before;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static org.hamcrest.Matchers.notNullValue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class TransportClearRealmCacheActionTests extends ESTestCase {
+
+    private AuthenticationService authenticationService;
+    private TransportClearRealmCacheAction action;
+    private TestCachingRealm nativeRealm;
+    private TestCachingRealm fileRealm;
+
+    @Before
+    public void setup() {
+        nativeRealm = mockRealm("native");
+        fileRealm = mockRealm("file");
+        final Realms realms = mockRealms(List.of(nativeRealm, fileRealm));
+
+        authenticationService = mock(AuthenticationService.class);
+        action = mockClearRealmCacheAction(realms, authenticationService);
+    }
+
+    public void testSingleUserCacheCleanupForAllRealms() {
+
+        final String user = "test";
+
+        // When no realm is specified we should clear all realms.
+        // This is equivalent to using a wildcard (*) instead of specifying realm name in query.
+        final String[] realmsToClear = new String[] {};
+        final String[] usersToClear = new String[] { user };
+        ClearRealmCacheRequest.Node clearCacheRequest = mockClearCacheRequest(realmsToClear, usersToClear);
+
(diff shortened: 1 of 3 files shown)


x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/realm/TransportClearRealmCacheActionTests.java
```diff
@@ -0,0 +1,179 @@
...
+        action = mockClearRealmCacheAction(realms, authenticationService);
+    }
+
+    public void testSingleUserCacheCleanupForAllRealms() {
+
+        final String user = "test";
+
+        // When no realm is specified we should clear all realms.
+        // This is equivalent to using a wildcard (*) instead of specifying realm name in query.
+        final String[] realmsToClear = new String[] {};
```
ywangd, line 61: I think we can random from either empty string array or null, e.g.
```suggestion
        final String[] realmsToClear = randomFrom(Strings.EMPTY_ARRAY, null);
```

Same applies to other empty realm/user array

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/realm/TransportClearRealmCacheActionTests.java
```diff
@@ -0,0 +1,179 @@
...
+        ClearRealmCacheRequest.Node clearCacheRequest = mockClearCacheRequest(realmsToClear, usersToClear);
+
+        ClearRealmCacheResponse.Node response = action.nodeOperation(clearCacheRequest, mock(Task.class));
+        assertThat(response.getNode(), notNullValue());
+
+        // We expect that caches of all realms are cleared for the given user,
+        // including last successful cache in the authentication service.
+        verify(fileRealm).expire(user);
+        verify(nativeRealm).expire(user);
+        verify(authenticationService).expire(user);
```
ywangd, line 72: We could also assert `expireAll` is not called.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/realm/TransportClearRealmCacheActionTests.java
```diff
@@ -0,0 +1,179 @@
...
+        ClearRealmCacheRequest.Node clearCacheRequest = mockClearCacheRequest(realmsToClear, usersToClear);
+
+        ClearRealmCacheResponse.Node response = action.nodeOperation(clearCacheRequest, mock(Task.class));
+        assertThat(response, notNullValue());
+
+        // We expect that only native cache is cleared,
+        // including last successful cache in the authentication service.
+        verify(nativeRealm).expire(user);
+        verify(fileRealm, times(0)).expire(user);
+        verify(authenticationService).expire(user);
```
ywangd, line 91: Same here, negative assertion for `expireAll` could be helpful.

ywangd (approved):
LGTM

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/realm/TransportClearRealmCacheActionTests.java
```diff
@@ -0,0 +1,179 @@
...
+            realms,
+            authenticationService
+        );
+    }
+
+    private TestCachingRealm mockRealm(String name) {
+        TestCachingRealm realm = mock(TestCachingRealm.class);
+        when(realm.name()).thenReturn(name);
+        return realm;
+    }
```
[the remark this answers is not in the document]
slobodanadamovic: Unfortunately, we can't. Reason is because `Realms` holds `Realm` objects not `CachingRealm`. 
I thought of using `CachingUsernamePasswordRealm` instead, but this abstract class has made `expire` and `expireAll` methods final. Hence why I decided to introduce `TestCachingRealm`.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/realm/TransportClearRealmCacheActionTests.java
slobodanadamovic: Good suggestion. Thanks.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/realm/TransportClearRealmCacheActionTests.java
slobodanadamovic: Makes total sense. Will do.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/realm/TransportClearRealmCacheActionTests.java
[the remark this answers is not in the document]
ywangd: All good. Thanks for the explanation.

PR #86914, opened 2022-05-19
[Refactor] Move classpath plugins to MockNode

### LINKED ISSUE #86635: Only load classpath plugins in MockNode
For a very long time, the `PluginsService` constructor has taken a collection argument that contains plugin classes to load from the classpath, rather than from a plugin bundle on disk. This argument is always an empty collection in production code, but the `MockNode` class passes in a list for tests.

Since then, several mock services have been added to the `MockNode` class. We should also make a mock PluginsService so that we can move all of the code for loading classpath plugins out of the production path and into testing classes.

williamrandolph:
We only use classpath plugins in tests, so we should move code handling them to our test framework. This PR introduces a new MockPluginsService in our test framework. The MockPluginsService is constructed with a list of classpath plugins that it will load by reflection. As before, these classpath plugins won't have their own classloaders.

We have to construct the pluginsService early in the `Node` class constructor so that `MockNode` methods can access it, but we have to do it after `Node` updates its `Settings` argument. Therefore, instead of passing in classpath plugins as a paramater, we are passing in a function that takes a `Settings` argument and returns a `PluginsService`. This lets us remove the classpath plugin list from the production code entirely.

Closes https://github.com/elastic/elasticsearch/issues/86635


williamrandolph:
@elasticsearchmachine please run elasticsearch-ci/part-1

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -79,91 +79,52 @@ record LoadedPlugin(PluginInfo info, Plugin instance, ClassLoader loader) {}
...
+                modules.stream().map(b -> b.plugin).forEach(modulesList::add);
                 seenBundles.addAll(modules);
             } catch (IOException ex) {
                 throw new IllegalStateException("Unable to initialize modules", ex);
             }
         }
 
-        // now, find all the ones that are in plugins/
+        // load plugins
+        List<PluginInfo> pluginsList = new ArrayList<>();
```
grcevski, line 106: Does it make sense for us to switch this to a Set instead of a list? It seems that PluginInfo has proper equals and hashcode, although the equals seems to also check version at the moment? It might be good to figure out if the comment on equals in PluginInfo should taken care of now as well. Should we be allowing multiple versions of equally named plugins? Same for modules, I think we shouldn't be allowing duplicates, perhaps even error out if we find some?

grcevski (approved):
LGTM! I left a comment around checking for duplicates on PluginInfo and perhaps looking into the comment in the PluginInfo equals method.

server/src/test/java/org/elasticsearch/plugins/PluginsServiceTests.java
```diff
@@ -187,11 +191,12 @@ public MultiplePublicConstructorsPlugin(final Settings settings) {
         final Settings settings = Settings.builder().put(Environment.PATH_HOME_SETTING.getKey(), home).build();
         final IllegalStateException e = expectThrows(
             IllegalStateException.class,
-            () -> newPluginsService(settings, MultiplePublicConstructorsPlugin.class)
+            () -> PluginsService.loadPlugin(MultiplePublicConstructorsPlugin.class, settings, home)
         );
         assertThat(e, hasToString(containsString("no unique public constructor")));
     }
 
+    // TODO: avoid classpath plugins
```
williamrandolph, line 199: Missed TODO?

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
williamrandolph: Looks like Version got added to PluginInfo a long time ago: abf9a866788b2fab855b50728670e2aa2dd42c5a

I don't see any reason in that commit for it to be there. I'll put out a quick fix for it on a different PR in case there's some sneaky dependence on it that we'd catch in tests.

We catch duplicate-named plugins and modules in our plugin-loading logic a little ways down (triggered by `loadBundles(seenBundles)`, and I added a test to catch this in https://github.com/elastic/elasticsearch/pull/86891/files#diff-099587538f96b160a9033106117540c99ffb60e00f8e4dfe8f54a2091a262943 , so I think for the moment we're good there.

williamrandolph: Ah, PluginInfo's equality check was briefly commented on in PR review, but not really discussed: https://github.com/elastic/elasticsearch/pull/4378/files#r9292810

server/src/main/java/org/elasticsearch/node/Node.java
```diff
@@ -291,21 +291,26 @@ public class Node implements Closeable {
...
-        this(environment, Collections.emptyList(), true);
+        this(environment, PluginsService.getCreatePluginsServiceFunction(environment), true);
     }
 
     /**
      * Constructs a node
      *
      * @param initialEnvironment         the initial environment for this node, which will be added to by plugins
-     * @param classpathPlugins           the plugins to be loaded from the classpath
+     * @param pluginsServiceFunction     a function that takes a {@link Settings} object and returns a {@link PluginsService}
```
rjernst, line 307: Instead of the generic "function" let's call this the ctor? ie `pluginServiceCtor`

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -79,91 +79,52 @@ record LoadedPlugin(PluginInfo info, Plugin instance, ClassLoader loader) {}
...
             try {
                 // TODO: remove this leniency, but tests bogusly rely on it
                 if (isAccessibleDirectory(pluginsDirectory, logger)) {
                     PluginsUtils.checkForFailedPluginRemovals(pluginsDirectory);
                     Set<PluginBundle> plugins = PluginsUtils.getPluginBundles(pluginsDirectory);
-                    for (final PluginBundle bundle : plugins) {
-                        pluginsList.add(bundle.plugin);
-                        pluginsNames.add(bundle.plugin.getName());
-                    }
+                    plugins.stream().map(b -> b.plugin).forEach(pluginsList::add);
```
rjernst, line 113: nit: the map can be `PluginBundle::plugin`

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -79,91 +79,52 @@ record LoadedPlugin(PluginInfo info, Plugin instance, ClassLoader loader) {}
...
-        pluginsLoaded.addAll(loaded);
-
         this.info = new PluginsAndModules(pluginsList, modulesList);
-        this.plugins = Collections.unmodifiableList(pluginsLoaded);
+        this.plugins = loadBundles(seenBundles);
 
-        checkMandatoryPlugins(pluginsNames, MANDATORY_SETTING.get(settings));
+        checkMandatoryPlugins(
+            pluginsList.stream().map(PluginInfo::getName).collect(Collectors.toSet()),
+            new HashSet<>(MANDATORY_SETTING.get(settings))
```
rjernst, line 126: Let's change the setting to be a Set<String> instead of List<String>?

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -205,8 +166,8 @@ private static void logPluginInfo(final List<PluginInfo> pluginInfos, final Stri
      * @return A stream of results
      * @param <T> The generic type of the result
      */
-    public <T> Stream<T> map(Function<Plugin, T> function) {
-        return plugins.stream().map(LoadedPlugin::instance).map(function);
+    public final <T> Stream<T> map(Function<Plugin, T> function) {
+        return this.plugins().stream().map(LoadedPlugin::instance).map(function);
```
rjernst, line 170: style nit: why add the `this`? it should be assumed for any method calls?

test/framework/src/main/java/org/elasticsearch/plugins/MockPluginsService.java
```diff
@@ -0,0 +1,78 @@
...
+        List<LoadedPlugin> pluginsLoaded = new ArrayList<>();
+
+        for (Class<? extends Plugin> pluginClass : classpathPlugins) {
+            Plugin plugin = loadPlugin(pluginClass, settings, configPath);
+            PluginInfo pluginInfo = new PluginInfo(
+                pluginClass.getName(),
+                "classpath plugin",
+                "NA",
+                Version.CURRENT,
+                "1.8",
```
rjernst, line 51: nit: we should set this to current java version?

rjernst (approved):
LGTM. I left a few nits.

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
williamrandolph: Doesn't seem to work because we've been directly accessing the public final `plugin` field, but it's simple to add a getter.

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
williamrandolph: Not sure how that got in there... but I'll take it out.

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
williamrandolph: It would be great if I could, but it looks like we need to add some capability to `Setting` before this would be simple.

test/framework/src/main/java/org/elasticsearch/plugins/MockPluginsService.java
williamrandolph: I'm going to use `JvmInfo.jvmInfo().version()`, but let me know if there's a better way.

rjernst: Yeah there is a simpler way: `Runtime.version().feature()` should work.

PR #86916, opened 2022-05-19
User Profile - hash only username for profile uid

ywangd:
This PR changes the profile uid calculation to hash only the username
and removes any use of domain/realm information. This means any users
with the same username now has the same base uid, which will still be
differentiated by the numerical suffix.

The reasons for this change are:
1. Make profile uid generation resistant to domain name or composition
   changes. Initially the predictable uid is only necessary for the
   moment of creation. But we now agree that there are benefits for this
   UID to be predictable in longer terms.
2. Allow a ESS-like experience where username is always unique and hence
   a 1-to-1 relationship between username and profile uid is desirable.
   If users can provide the same unique username guarantee with their
   own identity management system, they will be able to have the
   ESS-like experience.


x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/profile/ProfileDocumentTests.java
```diff
@@ -26,27 +27,25 @@ public class ProfileDocumentTests extends ESTestCase {
...
-            } else {
-                final RealmConfig.RealmIdentifier realmIdentifier = randomFrom(realmDomain.realms());
-                realmRef = new Authentication.RealmRef(
-                    realmIdentifier.getName(),
-                    realmIdentifier.getType(),
-                    randomAlphaOfLengthBetween(3, 8),
-                    realmDomain
-                );
-            }
-            realmRef = maybeMutateRealmRef(realmRef);
```
n1v0lg, line 48: Can we also nuke `maybeMutateRealmRef`?

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/profile/ProfileDocumentTests.java
```diff
@@ -26,27 +27,25 @@ public class ProfileDocumentTests extends ESTestCase {
...
+                randomAlphaOfLengthBetween(3, 18),
+                randomAlphaOfLengthBetween(8, 18),
+                randomMap(
+                    0,
+                    5,
+                    () -> new Tuple<>(randomAlphaOfLengthBetween(3, 8), randomFrom(randomAlphaOfLengthBetween(3, 8), randomInt()))
+                ),
+                randomBoolean()
+            );
+            final ProfileDocument profileDocument = ProfileDocument.fromSubject(new Subject(user, realmRef));
```
n1v0lg, line 48: I'm wondering if we should also add a test that validates that two users with different usernames produce different uids. Please ignore if we already have a test like that elsewhere.

n1v0lg (approved):
LGTM

For my own understanding, how and when will the numerical suffix be changed?

albertzaharovits (approved):
LGTM

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/profile/ProfileDocumentTests.java
```diff
@@ -26,27 +27,25 @@ public class ProfileDocumentTests extends ESTestCase {
 
     // Same uid should be used for the same username and domain definition (if domain is configured) or realm
```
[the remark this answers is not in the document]
ywangd: Good points. Updated as suggested.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/profile/ProfileDocumentTests.java
ywangd: Yes absolutely.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/profile/ProfileDocumentTests.java
ywangd: Makes sense. I add `testDifferentUids`.

ywangd:
> For my own understanding, how and when will the numerical suffix be changed?

The numerical suffix is to differentiate profile UIDs if they happen to have the same base UID. This can happen when:
1. A genuine hash collision (unlikely to happen, but not impossible)
2. Same username but from different realms/domains.
3. Username changes (quite feasible for a realm like SAML). The envisioned workflow is that the `username` will also be changed in the profile document so it matches the new username (but profile uid is guaranteed to be stable, i.e. not changed, let's say it's `u_0a1e2i3o4u_0`). The old username may be later taken by a different new user and the suffix is auto-increment by 1 (`u_0a1e2i3o4u_1`) when a new profile is generated for the new user.

ywangd:
@elasticmachine update branch

PR #86922, opened 2022-05-19
Implement write throttling in Netty4HttpPipeliningHandler

original-brownbear:
A step towards chunked rest encoding that only serializes chunks that
can be written right away and an optimization in isolation. A follow-up to this change will be able to use the flush + writability hooks introduced here to do serialization lazily and in chunks. This change was extracted into its own PR because it is somewhat complex for one but also because it provides performance benefits in isolation.

Introduces the concept of queuing writes in the handler once a channel becomes unwritable
to the `Netty4HttpPipeliningHandler`. Doing this reduces the memory consumption of compressed
responses as well as their first byte latency by only compressing chunks of a message
that can be written to the channel directly. Prior to this change, all chunks would be compressed
first before any writes to the channel, wasting buffers that can't be written yet as well as
making the write to the channel wait until compression completed.
A further advantage of this change is that a large REST response will not block the transport thread
for a long compression operation any longer as it does today.

The implementation of writability and flush hooks in `Netty4HttpPipeliningHandler` has been intentionally coded in a similar fashion to that in `Netty4WriteThrottlingHandler` (albeit it is much simpler because we do not chunk individual messages on the byte level in it) to allow for some drying up of the logic in a follow-up.


modules/transport-netty4/src/main/java/org/elasticsearch/transport/netty4/Netty4WriteThrottlingHandler.java
```diff
@@ -43,7 +46,54 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+        } else {
+            queueWrite(buf, promise);
+        }
+    }
+
+    /**
+     * Writes slices of up to the max write size until the channel stops being writable or the message has been written in full.
+     */
+    private void writeInSlices(ChannelHandlerContext ctx, ChannelPromise promise, ByteBuf buf) {
+        while (true) {
```
original-brownbear, line 66: I guess we don't really need a loop here since this will always just run a single time because the outbound buffer on the Netty end is less than `MAX_BYTES_PER_WRITE` in all cases today. We do have the same kind of loop in `doFlush` though, so I made it a loop here as well for consistency. Also, this makes the logic stable in case the outbound buffer size ever changes.

modules/transport-netty4/src/main/java/org/elasticsearch/transport/netty4/Netty4WriteThrottlingHandler.java
```diff
@@ -43,7 +46,54 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+            ctx.write(writeBuffer).addListener(forwardFailureListener(ctx, promise));
+            if (ctx.channel().isWritable() == false) {
+                // channel isn't writable any longer -> move to queuing
+                queueWrite(buf, promise);
+                return;
+            }
+        }
+    }
+
+    private static GenericFutureListener<Future<? super Void>> forwardFailureListener(ChannelHandlerContext ctx, ChannelPromise promise) {
```
original-brownbear, line 86: Just a note on why the future handling here is different from that in the http handler:

this listener is a little dirty and we do it nicer (but more expensive) in `Netty4HttpPipeliningHandler` where we use a proper promise combiner. We could align this to use a promise combiner as well or use this optimization that does not combine promises in the REST layer as well (it's not really necessary to use the combiner when all the messages are handled in order in all downstream handlers).

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -170,15 +184,72 @@ private void doWrite(ChannelHandlerContext ctx, Netty4HttpResponse readyResponse
...
+        }
+    }
+
+    @Override
+    public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+        doFlush(ctx);
+        super.channelInactive(ctx);
+    }
+
+    private boolean doFlush(ChannelHandlerContext ctx) {
```
henningandersen, line 216: Can we add javadoc on the return value here?

henningandersen (commented):
This looks fine to me, but I would like to see tests added for the new throttling as well as the sliced writing?

Also, I wonder if we do need two-level throttling or if just having the outer-level throttling could suffice?

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
original-brownbear: ++ added

original-brownbear:
> This looks fine to me, but I would like to see tests added for the new throttling as well as the sliced writing?

We cover this path in `org.elasticsearch.http.netty4.Netty4HttpServerTransportTests#testLargeCompressedResponse` at least and have various tests covering large messages passing through the `Netty4WriteThrottlingHandler`. What kind of test would you like to see in addition to these?

> Also, I wonder if we do need two-level throttling or if just having the outer-level throttling could suffice?

Yea I was wondering the same. Maybe not in most cases now actually. My thinking was mainly this:
1. We still have the no split option that benefits from the throttling in `Netty4WriteThrottlingHandler`.
2. The overhead of `Netty4WriteThrottlingHandler` is hardly measurable
3. The fixes I added to `Netty4WriteThrottlingHandler` help the transport pipeline as well even if we wouldn't need them for the HTTP path.

-> didn't see a big downside to keeping the throttling at the lower level for now.

henningandersen:
> We cover this path in `org.elasticsearch.http.netty4.Netty4HttpServerTransportTests#testLargeCompressedResponse` at least and have various tests covering large messages passing through the `Netty4WriteThrottlingHandler`. What kind of test would you like to see in addition to these?

I see that test checking that a write is passed through, but not that the throttling actually works. This PR adds throttling but does not verify that it works. I would like to see a test that verifies that we do throttle when necessary, also capturing this for future changes to the code here.

Also, we now extend `Netty4WriteThrottlingHandler` to write in slices to throttle earlier, but this is not captured in tests.

> Yea I was wondering the same. Maybe not in most cases now actually ... no split option
Besides the no split option, is there any cases where we will be benefiting from this second layer throttling?

> The fixes I added to Netty4WriteThrottlingHandler help the transport pipeline as well even if we wouldn't need them for the HTTP path.
Absolutely, not questioning the usefulness of the change, only that we have 2 layers of throttling in http handling now.

original-brownbear:
> I see that test checking that a write is passed through, but not that the throttling actually works. This PR adds throttling but does not verify that it works. I would like to see a test that verifies that we do throttle when necessary, also capturing this for future changes to the code here.

Got it. I agree this would be a nice test to have indeed. I'll have to think about how we can best build this in a not too specific way now that we have further changes incoming in this pipeline.

> Besides the no split option, is there any cases where we will be benefiting from this second layer throttling?

It depends and I don't have the benchmarks to judge how much it helps but. We currently split into 1M chunks in the REST handler but do 256kB writes downstream. So in the uncompressed case we save some buffers for TLS due to it. You could argue that this is a 0.75M saving per connection under heavy load but only so long as you are actually writing large responses on all of them.
-> But to be clear, if it weren't for the no-split option. I think I would have removed the downstream throttling handler here to keep things simple.

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpServerTransport.java
```diff
@@ -300,7 +299,6 @@ protected HttpChannelHandler(final Netty4HttpServerTransport transport, final Ht
         protected void initChannel(Channel ch) throws Exception {
             Netty4HttpChannel nettyHttpChannel = new Netty4HttpChannel(ch);
             ch.attr(HTTP_CHANNEL_KEY).set(nettyHttpChannel);
-            ch.pipeline().addLast("chunked_writer", new Netty4WriteThrottlingHandler(transport.getThreadPool().getThreadContext()));
```
original-brownbear, line 303: @henningandersen I thought about this a little more and I don't see the point in double-throttling outside of the no_split case. I think the no_split case is sufficiently fringe to ignore here and not optimize for. 

=> dropped the low level throttling in favor of just having the high level throttling now for the REST path.

henningandersen (commented):
I still miss the test that the new higher level throttling works.

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -196,6 +272,20 @@ public void close(ChannelHandlerContext ctx, ChannelPromise promise) {
         ctx.close(promise);
     }
 
+    private Future<Void> enqueueWrite(HttpObject msg, ChannelHandlerContext ctx) {
+        final ChannelPromise p = ctx.newPromise();
+        enqueueWrite(msg, p, ctx);
+        return p;
+    }
+
+    private void enqueueWrite(HttpObject msg, ChannelPromise promise, ChannelHandlerContext ctx) {
```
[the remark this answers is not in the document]
original-brownbear: ++ reordered :)

original-brownbear:
Thanks @henningandersen !

> I still miss the test that the new higher level throttling works.

Sorry about that, I added a new one for the low level throttling and then decided to remove it here and forgot all about a test so 🤦 . I think I found a neat way of testing this now that just asserts that we're not writing into unwritable channels. Let me know what you think :)

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpServerTransportTests.java
```diff
@@ -343,14 +356,33 @@ public void dispatchBadRequest(final RestChannel channel, final ThreadContext th
...
+                @Override
+                public ChannelHandler configureServerChannelHandler() {
+                    return new HttpChannelHandler(this, handlingSettings) {
+                        @Override
+                        protected void initChannel(Channel ch) throws Exception {
+                            super.initChannel(ch);
+                            ch.pipeline().addBefore("pipelining", "assert-throttling", new ChannelOutboundHandlerAdapter() {
+                                @Override
+                                public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
+                                    assertTrue("handler should throttle to only write into writable channels", ctx.channel().isWritable());
```
henningandersen, line 369: This verifies one side of the equation, that we do not write into non-writeable channels. Does the test also ensure that the channel indeed goes non-writeable and can we verify that somewhere, perhaps by adding another handling after the pipelining handler that snoops this?

henningandersen (commented):
Thanks for the test, I hope we can do a slight improvement to it, otherwise this looks good.

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpServerTransportTests.java
henningandersen: Looks good, thanks.

henningandersen (approved):
LGTM.

original-brownbear:
Thanks Henning!

PR #86923, opened 2022-05-19
Add explanation for persistent task assignment failure due to node shutdown

### LINKED ISSUE #84195: Persistent task assignment explanations can be confusing when nodes are shutting down
A side effect of the changes made in #72260 is that explanations for failures to assign persistent tasks do not mention nodes which are shutting down.

As an example, this is the assignment explanation from a cluster that contains 24 data/transform nodes that are shutting down and 21 master/ml/ingest nodes that are not shutting down:

image

The explanation seems bizarre in that it tells you that the transform is not going to be assigned to any of the _non_-transform nodes but does not tell you why it's not going to be assigned to any of the transform nodes.

We need to find a way to include explanations for why persistent tasks are not being assigned to nodes that are shutting down into the overall explanation of why a persistent task was not assigned.

pgomulka:
When a node is shutting down we should indicate that a node cannot run
persistent task for transform job. 
The information about nodes being shut down is present in cluster state metadata. 
This commit is passing down metadata into a method that creates explanations.

closes #84195


server/src/main/java/org/elasticsearch/cluster/node/DiscoveryNodes.java
```diff
@@ -702,6 +702,10 @@ private String validateAdd(DiscoveryNode node) {
             return null;
         }
 
+        public Map<String, DiscoveryNode> getNodes() {
```
rjernst, line 705: This seems to be added just for a test. I think that test can call build() to get the DiscoveryNodes instance and call getNodes() on that?

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/TransformNodes.java
```diff
@@ -245,7 +249,24 @@ public static boolean nodeCanRunThisTransform(
...
+                explain.put(node.getId(), "node is shutting down");
+            }
+            return false;
+        }
+
         // we found no reason that the transform can not run on this node
         return true;
     }
+
+    private static boolean isNodeShuttingDown(Metadata metadata, String nodeId) {
```
rjernst, line 263: Perhaps this should be a utility method in node shutdown. I assume we need this (or may already have it) in other code that uses tasks and needs better explanations about why tasks are rejected.

rjernst, line 263: Perhaps this should be a utility method in node shutdown. I assume we need this (or may already have it) in other code that uses tasks and needs better explanations about why tasks are rejected.

rjernst (commented):
This looks ok, but I wonder if this is something that should be more general purpose within the tasks infrastructure. Having to tweak every bit of code to be shutdown aware is likely to be error prone.

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/TransformNodes.java
```diff
@@ -245,7 +249,24 @@ public static boolean nodeCanRunThisTransform(
...
+        NodesShutdownMetadata nodesShutdownMetadata = metadata != null ? metadata.custom(NodesShutdownMetadata.TYPE) : null;
+        if (nodesShutdownMetadata == null || nodesShutdownMetadata.getAllNodeMetadataMap() == null) {
+            // There are no nodes in the process of shutting down, return null.
+            return false;
+        }
+
+        return nodesShutdownMetadata.getAllNodeMetadataMap().get(nodeId) != null;
```
gwbrown, line 270: Super nitpicky, but `containsKey()` might be clearer.

AthenaEryma (commented):
The node shutdown piece of this makes sense, but it looks like a lot of the changes are to Transform code, so I'm less certain of that context.

server/src/main/java/org/elasticsearch/cluster/node/DiscoveryNodes.java
pgomulka: sure, this requires a slight change to other tests using buildNodes but agree that it should look better

rjernst:
> I think it could do this in a separate PR though.

Consolidating as a followup is fine, I just want to make sure we aren't duplicating this effort later.

> do we have another example to use it?

Any other persistent task? I assume all these can/should fail for the same reason?

pgomulka:
>> I think it could do this in a separate PR though.

>Consolidating as a followup is fine, I just want to make sure we aren't duplicating this effort later.

so there is around 10+ subclasses of PersistentTasksExecutor. They all have their own way of deciding if a node can be assigned and creating an explanation if non can be.
It would be a bigger work to refactor PersistentTasksExecutor::getAssigment and use some kind of Predicate which can test and generate explanations.

I recon we merge this PR and create an issue to track this. Better to have this fixed in one executor than nowhere.

droberts195 (commented):
The problem with this change is that the approach of fixing the individual persistent task executors is inconsistent with the higher level approach for node shutdown that was implemented in #72260.  In that PR code was added to filter the candidate nodes before the individual persistent task executors even see them: https://github.com/elastic/elasticsearch/blob/2fc917265fd0a145ba587cc2b520785a8d45661c/server/src/main/java/org/elasticsearch/persistent/PersistentTasksClusterService.java#L342-L348

This means that the method `nodeCanRunThisTransform` will never get passed a node that was shutting down, so will never add its new fragment to the explanation.

#72260 _could_ have been implemented by having the checks for shutting-down nodes in every persistent task executor. Then the approach of this PR would work for adding the explanation fragments. So one option for this PR would be to change everything to work that way. However, that would be quite a lot of code churn, and also I imagine the approach of #72260 was designed to prevent accidents where a bug in an individual `getAssignment` implementation forgets to exclude shutting down nodes.

Another option would be to change the method signature for `PersistentTasksExecutor.getAssignment` so that as well as a list of candidate nodes it also takes a list of shutting down nodes. Then the sub-classes can optionally use the shutting down nodes to build their explanation.

Or the final option I can think of is for the calling code in `PersistentTasksClusterService` to generate an explanation fragment like, "Not considering assignment to the following nodes that are shutting down [node1, node6, node12]". Then pass this when calling `PersistentTasksExecutor.getAssignment`, and then it's up to each `PersistentTasksExecutor` to decide whether or not to append that to the explanation it sends back.

Issue #73276, opened 2021-05-20
MatrixStats NPE when technically misused


benwtrent:
**Elasticsearch version** (`bin/elasticsearch --version`):
All
**Steps to reproduce**:

 1. Two or more indices
 2. Try to execute matrix stats on two fields in the two indices (but the fields are EXCLUSIVELY in each index)
 3. Witness the NPE

**Provide logs (if relevant)**:

```
[2021-05-20T08:48:36,621][WARN ][r.suppressed             ] [runTask-0] path: /correlate_latency*/_search, params: {pretty=true, size=0, index=correlate_latency*}
org.elasticsearch.action.search.SearchPhaseExecutionException:
        at org.elasticsearch.action.search.AbstractSearchAsyncAction.onPhaseFailure(AbstractSearchAsyncAction.java:661) [elasticsearch-8.0.0-SNAPSHOT.jar:8.0.0-SNAPSHOT]
        at org.elasticsearch.action.search.FetchSearchPhase$1.onFailure(FetchSearchPhase.java:89) [elasticsearch-8.0.0-SNAPSHOT.jar:8.0.0-SNAPSHOT]
        at org.elasticsearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:28) [elasticsearch-8.0.0-SNAPSHOT.jar:8.0.0-SNAPSHOT]
        at org.elasticsearch.common.util.concurrent.TimedRunnable.doRun(TimedRunnable.java:33) [elasticsearch-8.0.0-SNAPSHOT.jar:8.0.0-SNAPSHOT]
        at org.elasticsearch.common.util.concurrent.ThreadContext$ContextPreservingAbstractRunnable.doRun(ThreadContext.java:728) [elasticsearch-8.0.0-SNAPSHOT.jar:8.0.0-SNAPSHOT]
        at org.elasticsearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:26) [elasticsearch-8.0.0-SNAPSHOT.jar:8.0.0-SNAPSHOT]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
        at java.lang.Thread.run(Thread.java:834) [?:?]
Caused by: java.lang.NullPointerException
[...]

PR #86926, opened 2022-05-19
fix: check field existence before trying to merge running stats

salvatore-campagna:
All fields must co-hexist in the same docs for the stats to be
calculated correctly. As a result, we need to check existence of
the fields before trying to merge running stats.

Instead of collecting all field names into a collection we decided
to extract the field names from one of the maps. This saves us from
serializing/deserializing another collection and some backward
compatibility pain. This means also that we assume all maps
to hold the same set of keys and just chose one of them as the
candidate to extract keys (field names) from.

This is a fix for issue #73276.


modules/aggs-matrix-stats/src/main/java/org/elasticsearch/search/aggregations/matrix/stats/InternalMatrixStats.java
```diff
@@ -7,6 +7,8 @@
  */
 package org.elasticsearch.search.aggregations.matrix.stats;
 
+import joptsimple.internal.Strings;
```
nik9000, line 10: I don't think we mean for jopt to be used outside of the cli bits.

modules/aggs-matrix-stats/src/main/java/org/elasticsearch/search/aggregations/matrix/stats/RunningStats.java
```diff
@@ -309,6 +333,12 @@ public RunningStats clone() {
         }
     }
 
+    public Set<String> getAllFieldNames() {
+        // NOTE: here we assume all maps used to store values hold the same set of keys (field names).
+        // As a result, we just need to read the set of keys from one of them.
```
nik9000, line 338: Instead of writing this comment you can `assert` that they all have the same keyset. That might be a little easier to scan for a reviewer and a little more paranoid for testing.

nik9000 (approved):

ineo00048:
no problem with key won't work for us at this moment but I will be sure to
keep you posted on what happens to be the area installed on newest
construction project for front door repair.

nik9000:
> front door repair.

Can't tell if accidental or new technical jargon.

modules/aggs-matrix-stats/src/main/java/org/elasticsearch/search/aggregations/matrix/stats/InternalMatrixStats.java
salvatore-campagna: 🤦

PR #86928, opened 2022-05-19
Use larger string literal deduplicator for field names

original-brownbear:
Beats mappings are almost 5k fields, lets size this larger to make mapper parsing
faster. The extra memory cost of 5k vs 1k really doesn't matter much here
in practice anyway but is a nice couple of percent speedup for bootstrapping the many shards benchmarks cluster.

### DIFF
--- a/server/src/main/java/org/elasticsearch/common/util/StringLiteralDeduplicator.java
+++ b/server/src/main/java/org/elasticsearch/common/util/StringLiteralDeduplicator.java
@@ -18,11 +18,19 @@ import java.util.Map;
  */
 public final class StringLiteralDeduplicator {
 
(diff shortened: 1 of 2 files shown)


joegallo:
Any sense in making the size constants be some multiple of a power of two? Perhaps 1024 for DEFAULT_MAX_SIZE and 6144 (4096+2048) for the mapper constructor call? I imagine there's no harm in keeping them as multiples of powers of ten, but I'm thinking the underlying map has some power of two size affinity.

In the same vein, does it make sense to pass an `initialCapacity` to `ConcurrentCollections.newConcurrentMapWithAggressiveConcurrency()`? 5000 seems pretty far from 16, so I'm imagining a lot of internal resizes which seems like a waste.

PR #86932, opened 2022-05-19
Avoid building duplicate FieldTypeLookup in MappingLookup Constructor

original-brownbear:
If we don't have runtime fields we don't need to build this twice.
These are huge for something like Beats mappings and this change saves
about 400MB of heap per data node in the many shards benchmarks.

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/mapper/MappingLookup.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/MappingLookup.java
@@ -170,8 +170,14 @@ public final class MappingLookup {
             }
         }
 
-        this.fieldTypeLookup = new FieldTypeLookup(mappers, aliasMappers, mapping.getRoot().runtimeFields());
-        this.indexTimeLookup = new FieldTypeLookup(mappers, aliasMappers, Collections.emptyList());
+        final Collection<RuntimeField> runtimeFields = mapping.getRoot().runtimeFields();
+        this.fieldTypeLookup = new FieldTypeLookup(mappers, aliasMappers, runtimeFields);
+        if (runtimeFields.isEmpty()) {
+            // without runtime fields this is the same as the field type lookup
+            this.indexTimeLookup = fieldTypeLookup;
+        } else {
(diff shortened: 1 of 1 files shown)


cbuescher (approved):
LGTM

original-brownbear:
Thanks Christoph!

server/src/main/java/org/elasticsearch/index/mapper/MappingLookup.java
```diff
@@ -170,21 +170,22 @@ private MappingLookup(
...
 
-        this.fieldTypeLookup = new FieldTypeLookup(mappers, aliasMappers, mapping.getRoot().runtimeFields());
-        this.indexTimeLookup = new FieldTypeLookup(mappers, aliasMappers, Collections.emptyList());
+        final Collection<RuntimeField> runtimeFields = mapping.getRoot().runtimeFields();
+        this.fieldTypeLookup = new FieldTypeLookup(mappers, aliasMappers, runtimeFields);
+        if (runtimeFields.isEmpty()) {
+            // without runtime fields this is the same as the field type lookup
+            this.indexTimeLookup = fieldTypeLookup;
+        } else {
+            this.indexTimeLookup = new FieldTypeLookup(mappers, aliasMappers, Collections.emptyList());
```
javanna, line 179: Good catch. Would it make sense to have a test that verifies that these two are the same instance when there are no runtime fields, just to verify that we don't duplicate it when we can avoid it? Or is this overkill?

original-brownbear: Thanks :) I'm not sure a test is all that helpful here. It locks down this specific instance but wouldn't have prevented it initially. Maybe we should have a bigger conversation about somehow tracking the memory usage of the objects we create for each index/shard and testing the output of that. This kind of thing is somewhat unfortunate for large frozen nodes in particular and regressions are almost impossible to spot in benchmarks.
This may not be the right place to raise this topic though :) I'll start a conversation on a more appropriate channel.

PR #86934, opened 2022-05-19
Move pidfile handling to server cli

rjernst:
Now that the server cli is in java, we can do more system level things
inside it. This commit moves validating the pidfile into the
server cli, and moves writing and cleaning it up into the main
Elasticsearch class. One benefit is we get validation of directory/file
problems up front before even trying to start the ES process.


grcevski (commented):
This looks good! I'm wondering if we are missing the clean up of the pid file on exit?

rjernst:
> I'm wondering if we are missing the clean up of the pid file on exit?
Yes, I missed that before. I just pushed a commit that moves the writing back into the main process, but moves it up to the main startup, so we no longer need to pass it around in Settings. We still do the validation in the CLI, so what is passed in ServerArgs must be good to write to.

server/src/main/java/org/elasticsearch/bootstrap/Elasticsearch.java
```diff
@@ -171,6 +172,31 @@ private static void startCliMonitorThread(InputStream stdin) {
...
+        }
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            if (Files.exists(pidFile)) {
+                try {
+                    Files.delete(pidFile);
+                } catch (IOException e) {
+                    // ignore, nothing we can do because are shutting down
+                }
+            }
+        }, "elasticsearch[pidfile-cleanup]"));
```
grcevski, line 192: Good catch here to name this thread! I think there's one test in `SimpleThreadPoolIT` that might need to know about this name now.

grcevski (approved):
LGTM!

server/src/main/java/org/elasticsearch/bootstrap/Elasticsearch.java
rjernst: I think it's not necessary because the thread isn't started until shutdown.

PR #86936, opened 2022-05-19
Timing out stale remote master history

masseyke:
This change makes it so that the remote master history that has been fetched will time out after 5 minutes (by default). The MasterHistoryService will return its value as `null` if it is older than the configured timeout. This way MasterHistoryService clients don't report status based on out of date information.


server/src/main/java/org/elasticsearch/cluster/coordination/MasterHistoryService.java
```diff
@@ -60,14 +79,23 @@ public MasterHistory getLocalMasterHistory() {
...
     public List<DiscoveryNode> getRemoteMasterHistory() throws Exception {
         // Grabbing a reference to the object in case it is replaced in another thread during this method:
         RemoteHistoryOrException remoteHistoryOrExceptionCopy = remoteHistoryOrException;
+        /*
+         * If the remote history we have is too old, we just return null with the assumption that it is stale and the new one has not
+         * come in yet.
+         */
+        long acceptableRemoteHistoryTime = currentTimeMillisSupplier.getAsLong() - acceptableRemoteHistoryAge.getMillis();
+        if (remoteHistoryOrExceptionCopy.creationTimestamp < acceptableRemoteHistoryTime) {
+            return null;
```
andreidan, line 97: Would we want to also at some point null-ify `remoteHistoryOrException` ?  (we'd likely need an AtomicReference or such to do a compare and swap) 

What do you think?

andreidan (commented):

This generally LGTM - just left a few questions

server/src/main/java/org/elasticsearch/cluster/coordination/MasterHistoryService.java
masseyke: My thinking was that there was no harm in holding onto a single one, and it avoids having to synchronize access to it. Synchronizing it would not be super expensive (we'd need to read it, do the time calculation, and set it to null in an atomic action), but it seemed like an unnecessary complication. Is it the memory that you're concerned about? We cap these things to have at most 50 entries, so I'd expect at most a few KB here.

masseyke:

server/src/main/java/org/elasticsearch/cluster/coordination/MasterHistoryService.java
```diff
@@ -33,17 +36,33 @@ public class MasterHistoryService {
...
 
+    private static final TimeValue DEFAULT_MAX_USABLE_REMOTE_HISTORY_AGE = new TimeValue(5, TimeUnit.MINUTES);
+
+    /**
+     * This is the amount of time that can pass after a RemoteHistoryOrException is returned from the remote master until it is
+     * considered stale and not usable.
+     */
+    public static final Setting<TimeValue> MAX_USABLE_REMOTE_HISTORY_AGE_SETTING = Setting.timeSetting(
+        "master_history.max_usable_remote_history_age",
+        DEFAULT_MAX_USABLE_REMOTE_HISTORY_AGE,
```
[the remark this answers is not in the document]
andreidan: Shall we use `positiveTimeSetting` to at least not worry about negative values?

server/src/main/java/org/elasticsearch/cluster/coordination/MasterHistoryService.java
andreidan: I was mostly worried about the `remoteHistoryOrException` leaking outside the service by other means than `getRemoteMasterHistory` or the service evolving and using the `remoteHistoryOrException` in a private way (assuming it's `null` when it's stale, which would not be a wild assumption given what the getter returns) - @DaveCTurner would you have a strong opinion here?

DaveCTurner: I am inclined to leave it in place too. You never know, it might even end up being useful in a heap dump. Maybe add a comment to the field indicating that it might be stale.

For future reference it wouldn't be a big deal to clear it after the timeout either:

```
        transportService.getThreadPool().schedule(() -> {/* clear field*/}, acceptableRemoteHistoryAge, Names.SAME);
```

(also there's no need for a `ThreadPool` constructor argument if you have a `TransportService`)

andreidan: ++ let's leave it as is then

server/src/main/java/org/elasticsearch/cluster/coordination/MasterHistoryService.java
```diff
@@ -132,26 +164,27 @@ public void onFailure(Exception e) {
...
-                    remoteHistoryOrException = new RemoteHistoryOrException(e);
+                    remoteHistoryOrException = new RemoteHistoryOrException(e, currentTimeMillisSupplier.getAsLong());
                 }
             }
         );
     }
 
-    record RemoteHistoryOrException(List<DiscoveryNode> remoteHistory, Exception exception) { // non-private for testing
+    // non-private for testing
+    record RemoteHistoryOrException(List<DiscoveryNode> remoteHistory, Exception exception, long creationTimestamp) {
```
andreidan, line 174:
```suggestion
    record RemoteHistoryOrException(List<DiscoveryNode> remoteHistory, Exception exception, long creationTimeMillis) {
```

server/src/main/java/org/elasticsearch/cluster/coordination/MasterHistoryService.java
```diff
@@ -132,26 +164,27 @@ public void onFailure(Exception e) {
...
 
         public RemoteHistoryOrException {
             if (remoteHistory != null && exception != null) {
                 throw new IllegalArgumentException("Remote history and exception cannot both be non-null");
             }
         }
 
-        RemoteHistoryOrException(List<DiscoveryNode> remoteHistory) {
-            this(remoteHistory, null);
+        RemoteHistoryOrException(List<DiscoveryNode> remoteHistory, long creationTimestamp) {
```
andreidan, line 182: Can we rename the creationTimestamp to include the unit everywhere? ( https://github.com/elastic/elasticsearch/pull/86936/files#r880273628 )

andreidan (approved):
LGTM, with a minor rename suggestion

Thanks Keith

masseyke:

masseyke:

PR #86938, opened 2022-05-19
Rename NodeEnvironment.NodePath to DataPath

rjernst:
The NodePath inner class of NodeEnvironment represents path.data entries
of the node. The name however is sometimes confusing since these are the
data paths, and there is normally no singular concept of "node path" (an
installation has several different paths). This commit renames NodePath
to DataPath, as well as all methods and variables referring to it, so
that the more general term "node paths" can be utilized for something
node wide: the paths in environment (in a future PR).


grcevski (approved):
LGTM!

rjernst:
Since this is purely a rename, I will backport to make future backports in the touched code easier.

PR #86939, opened 2022-05-19
Fix merging with empty results

### LINKED ISSUE #84622: Wrong aggregation result on date field.
### Elasticsearch Version

8.0.0

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

MacOS

### Problem Description

When aggregating over 2 indices (each one containing a date field but **not the same** field), the aggregation result for one of the fields is lacking `value_as_string` field.

### Steps to Reproduce

1. Create first index

```
PUT sample_index_1
{
  "mappings": {
    "properties": {
      "date_field_1": {
        "type": "date"
      }
    }
  }
}
```

2. Create second index

```
PUT sample_index_2
{
  "mappings": {
    "properties": {
      "date_field_2": {
        "type": "date"
      }
    }
  }
}
```

Notice that these 2 indices have mappings for **different** fields.

3. Index a doc into the first index

```
POST sample_index_1/_doc
{
  "date_field_1": "2021-05-01T00:00:00.000Z"
}
```

4. Index a doc into the second index

```
POST sample_index_2/_doc
{
  "date_field_2": "2021-05-02T00:00:00.000Z"
}
```

5. Perform an aggregation

```
GET sample_index_1,sample_index_2/_search
{
  "size": 0,
  "aggs": {
    "date_field_1": {
      "max": {
        "field": "date_field_1"
      }
    },
    "date_field_2": {
      "max": {
        "field": "date_field_2"
      }
    }
  }
}
```

The result is:
```
{
  "took" : 14,
  "timed_out" : false,
  "_shards" : {
    "total" : 2,
    "successful" : 2,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 2,
      "relation" : "eq"
    },
    "max_score" : null,
    "hits" : [ ]
  },
  "aggregations" : {
    "date_field_2" : {
      "value" : 1.6199136E12
    },
    "date_field_1" : {
      "value" : 1.6198272E12,
      "value_as_string" : "2021-05-01T00:00:00.000Z"
    }
  }
}
```

The aggregation result for `date_field_1` has `value_as_string` but `date_field_2` doesn't.

I took some time to debug the code and this is what I saw in `Max.java`:
```
        if (hasValue && format != DocValueFormat.RAW) {
            builder.field(CommonFields.VALUE_AS_STRING.getPreferredName(), format.format(max).toString());
        }
```

Indeed, after adding logging it can be seen that for one aggregation the condition `format != DocValueFormat.RAW` doesn't hold and `value_as_string` field is not emitted.
Unfortunately, I don't know why it is the case and why it is different between `date_field_1` and `date_field_2` where the situation seems perfectly symmetrical.

### Logs (if relevant)

_No response_

tmgordeeva:
If we try to merge responses with an empty response, we might take the RAW
format from an empty response over a format from a non-empty response.

Closes #84622


server/src/test/java/org/elasticsearch/action/search/SearchResponseMergerTests.java
```diff
@@ -452,6 +455,49 @@ public void testMergeCompletionSuggestionsTieBreak() throws InterruptedException
         }
     }
 
+    // Test merging results where one result has a raw format, for instance if
+    // searching over multiple indexes where the field isn't mapped in all indexes.
+    public void testMergeEmptyFormat() throws InterruptedException {
```
nik9000, line 460: If one field isn't mapped I think we use the mapped field as the "lead" for the merge. Does this happen in production?

not-napoleon: Bit of an off the cuff read, so I could be wrong, but I'm pretty sure we don't do that.  We've had problems around this with terms over IP fields that were unmapped in some indices, where the unmapped ones came back as keywords (the default).  Often we use `value_type` to nudge it, but I don't think anything in the merge picks the right one naturally.

nik9000: It's quite possible for two aggs to have a different type and for that to have an empty type. But the "real" one will be first. Sorry for the confusion.

nik9000:
> The current version takes advantage of the logic from InternalAggregations reduce which is meant to move mapped fields to the top in merges. It does this by making empty Maxes claim to be unmapped, when they might be actually empty maxes on mapped fields, so I'm a little hesitant to take care of it this way.

That `isUnmapped` method is really only used for this so far as I can tell. @not-napoleon what would you think of renaming it?

I see other aggregations performing pretty similar routines. It also looks common to build "empty" aggregations for uncollected ordinals. My guess is we'd be better off renaming `isUnmapped` to `isEmpty` or `canLeadReduction`.

tmgordeeva:
This is failing on a cross version check on the REST API test that's expected. I'm not sure how to disable that specifically here.

server/src/main/java/org/elasticsearch/search/aggregations/metrics/Max.java
```diff
@@ -22,10 +22,21 @@
...
     private final double max;
+    private final boolean mapped;
 
     public Max(String name, double max, DocValueFormat formatter, Map<String, Object> metadata) {
         super(name, formatter, metadata);
         this.max = max;
+        this.mapped = true;
+    }
+
+    public static Max EmptyMax(String name, DocValueFormat formatter, Map<String, Object> metadata) {
```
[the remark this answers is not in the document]
nik9000: could you rename this method?

server/src/test/java/org/elasticsearch/action/search/SearchResponseMergerTests.java
```diff
@@ -452,6 +455,47 @@ public void testMergeCompletionSuggestionsTieBreak() throws InterruptedException
...
```
nik9000, line 460: Maybe javadoc? It's not important but may as well be consistent.

nik9000 (changes requested):

nik9000 (approved):

PR #86941, opened 2022-05-19
Extract transport cluster settings/ilm execute logic

grcevski:
Extract the execute logic from the transport actions for cluster update settings and ILM put/delete to support future reuse for operator file based updates.

This PR serves as a base for the follow-up work to actually implement file based settings. There's no particular reason why these two were chosen, other than diversity of operations: settings vs entities and elasticsearch core vs plugin.

Relates to #86224


dakrone (approved):
LGTM

DaveCTurner (approved):

PR #86946, opened 2022-05-19
Propagate MapperBuilderContext across merge calls

javanna:
With #86166 we have added support for leaf fields that have dots in their names, without needing to expand their path to intermediate objects. That means that for instance a host.name keyword field mapper can be held directly by the root object mapper if the root has subobjects set to false.

This opens up for situations where a field name containing dots can't necessarily be split into a leaf name and a parent path made of intermediate object mappers. In the field mapper merge code, we build the full path of the merged field making that now incorrect assumption, which causes the result of merged leaf fields to have incorrect full path. This is a bug although it got unnoticed so far as the full path of a field mapper is not so widely used compared to its leaf name (returned by the simpleName method). One place where the full path is used is when sorting the child mappers in ObjectMapper#serializeMappers which was causing MapperService#assertRefreshIsNotNeeded to trip as the result of the merge has fields in a different order compared to the incoming mappings, due to the different name between the original mapper and the merged mapper.

With this fix, we add a MapperBuilderContext argument to the different merge methods, propagate the MapperBuilderContext in all the merge calls and create a child context when needed, meaning whenever merging object mappers, so that their children mappers are created with the proper full path.

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/mapper/DocumentParserContext.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/DocumentParserContext.java
@@ -364,6 +364,14 @@ public abstract class DocumentParserContext {
 
     public abstract ContentPath path();
 
+    public final MapperBuilderContext createMapperBuilderContext() {
+        String p = path().pathAsText("");
+        if (p.endsWith(".")) {
+            p = p.substring(0, p.length() - 1);
+        }
+        return new MapperBuilderContext(p);
+    }
+
     public abstract XContentParser parser();
 
     public abstract LuceneDocument rootDoc();
--- a/server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
@@ -1142,8 +1142,9 @@ public abstract class FieldMapper extends Mapper implements Cloneable {
             for (Parameter<?> param : getParameters()) {
                 param.merge(in, conflicts);
             }
+            MapperBuilderContext childContext = mapperBuilderContext.createChildContext(in.simpleName());
             for (FieldMapper newSubField : in.multiFields.mappers) {
-                multiFieldsBuilder.update(newSubField, mapperBuilderContext);
+                multiFieldsBuilder.update(newSubField, childContext);
             }
             this.copyTo.reset(in.copyTo);
             validate();
--- a/server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
@@ -321,7 +321,7 @@ public abstract class FieldMapper extends Mapper implements Cloneable {
     public abstract Builder getMergeBuilder();
 
     @Override
-    public final FieldMapper merge(Mapper mergeWith) {
+    public final FieldMapper merge(Mapper mergeWith, MapperBuilderContext mapperBuilderContext) {
         if (mergeWith == this) {
             return this;
         }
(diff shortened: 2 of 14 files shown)


server/src/main/java/org/elasticsearch/index/mapper/DocumentParserContext.java
```diff
@@ -364,6 +364,14 @@ public DocumentDimensions getDimensions() {
 
     public abstract ContentPath path();
 
+    public final MapperBuilderContext createMapperBuilderContext() {
```
javanna, line 367: This is not perfect, but more isolated now to the only place where it's needed. I am not entirely sure if we can remove it from here hence I removed the TODO.

server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
```diff
@@ -321,7 +321,7 @@ private static void checkNestedScopeCompatibility(String source, String target)
     public abstract Builder getMergeBuilder();
 
     @Override
-    public final FieldMapper merge(Mapper mergeWith) {
+    public final FieldMapper merge(Mapper mergeWith, MapperBuilderContext mapperBuilderContext) {
```
javanna, line 324: I went for propagating the MapperBuilderContext instead of adding a String path argument, because in some cases we already have a mapper builder context instance that we can just propagate, and in general it seems like a good idea to start merging at the root level with a ROOT mapper builder context and then build a child context for each level, in the caller of the merge methods.

server/src/test/java/org/elasticsearch/index/mapper/ObjectMapperMergeTests.java
```diff
@@ -109,29 +109,119 @@ public void testMergeDisabledRootMapper() {
...
+        final ObjectMapper merged = mergeInto.merge(mergeWith, MapperBuilderContext.ROOT);
+
+        ObjectMapper foo = (ObjectMapper) merged.getMapper("foo");
+        ObjectMapper metrics = (ObjectMapper) foo.getMapper("metrics");
+        final TextFieldMapper textFieldMapper = (TextFieldMapper) metrics.getMapper("host.name");
+        assertEquals("foo.metrics.host.name", textFieldMapper.name());
+        assertEquals("host.name", textFieldMapper.simpleName());
+        FieldMapper fieldMapper = textFieldMapper.multiFields.iterator().next();
+        assertEquals("foo.metrics.host.name.keyword", fieldMapper.name());
+        assertEquals("keyword", fieldMapper.simpleName());
```
javanna, line 213: I am not too sure that these additional merge tests belong to this class. I don't know whether there's an existing place for specific merge tests. Also, these are weird merge tests because they don't really test merging different mapper but rather merge the same mappers and verify the resulting simpleName and name. I can see how that will look weird in a couple of years: "why did we write these tests again?". Thoughts?

romseygeek: Maybe add a comment to the effect that we're testing that leaf fields with dots in their names merge correctly?

romseygeek: Really I think we should rewrite all these tests so that they extend MapperServiceTestCase and test the actual merge paths that MapperService calls, but let's do that as a followup.

romseygeek (approved):
LGTM.  If we ever manage to move merging onto the Builders then I think we can remove a lot of this again, but for now we're stuck with it.

server/src/test/java/org/elasticsearch/index/mapper/ObjectMapperMergeTests.java
javanna: interesting, do they now test a different codepath? I mean they call root.merge directly, instead of going through mapper service, is that what you mean?

PR #86948, opened 2022-05-20
[Test] Increase length of test password for FIPS

ywangd:
Password must be at least 114 bits in FIPS mode. This PR fixes the password length
in the new ServerCliTests so it passes in FIPS mode.

Relates: #85758 

PS: The test [failed](https://gradle-enterprise.elastic.co/s/mrlw6o27onxee/tests/:distribution:tools:server-cli:test/org.elasticsearch.server.cli.ServerCliTests/testKeystorePassword) on my PR CI.


tvernum (approved):
LGTM,
But I'm surprised the line above works. Password-less keystores shouldn't work in FIPS either.

ywangd:
> LGTM, But I'm surprised the line above works. Password-less keystores shouldn't work in FIPS either.

The test has logic to bypass manipulating keyword if the password is `null` or empty. It works in the sense that it does not try to test anything about keystore when there is no password. The test is just to ensure the command prompt sequence. So I didn't bother to do anything extra.

ywangd:

PR #86950, opened 2022-05-20
Rename PluginInfo to PluginDescriptor

rjernst:
The class PluginInfo represents the plugin-descriptor.properties file
that each plugin must have. This commit renames the class to more
closely match what it represents: the plugin descriptor.


distribution/tools/plugin-cli/src/test/java/org/elasticsearch/plugins/cli/InstallPluginActionTests.java
```diff
@@ -581,7 +581,8 @@ public void testJarHell() throws Exception {
         assumeTrue("real filesystem", isReal);
         Path pluginDirectory = createPluginDir(temp);
         writeJar(pluginDirectory.resolve("other.jar"), "FakePlugin");
-        PluginDescriptor pluginZip = createPluginZip("fake", pluginDirectory); // adds plugin.jar with FakePlugin
+        InstallablePlugin pluginZip = createPluginZip("fake", pluginDirectory); // adds plugin.jar with
```
williamrandolph, line 584: not even a nit: Did Spotless reformat this line-ending comment? I don't think I've seen a comment at the end of a line spill over to the next line in our code base. (This was the only surprise in the diff, but it's fine with me if it stays.)

williamrandolph (approved):
LGTM! I very much agree that this is consistent with how we use "Descriptor" in the rest of our code.

distribution/tools/plugin-cli/src/test/java/org/elasticsearch/plugins/cli/InstallPluginActionTests.java
rjernst: I think what happened was in an initial commit for this, I had not renamed this PluginDescriptor, and so it had a fully qualified name, which pushed the line length out, and did this odd comment splitting. Then when I renamed this to InstallablePlugin, it was no longer too long. I've fixed it now and spotless seems happy.

rjernst:
Because this is purely a rename, I'm going to backport this to make future backports in this code easier.

PR #86952, opened 2022-05-20
User Profile - Support literal username as base of profile UID

ywangd:
Some identity system can provide strong guarantee on username
uniqueness. In this case, it is desirable that the unique Profile ID is
transparently related to the username, i.e. the profile UID and the
username are tranlatable in both ways. This means hashing the username
is not a viable option since hashing is one way.

This PR adds support for using the literal username as the base of a
profile UID with a fix suffix string of admin's choosing (configured on
the security domain level). This mode of UID generation also means that
ES will *not* try to differentiate users when they have the same name
because this should be guaranteed by the identity system. ES will
instead throw an error on duplication which is the safe behaviour.


x-pack/plugin/security/qa/profile/build.gradle
```diff
@@ -23,6 +26,13 @@ testClusters.matching { it.name == 'javaRestTest' }.configureEach {
 
   setting 'xpack.security.authc.realms.ldap.ldap1.enabled', 'false'
   setting 'xpack.security.authc.domains.my_domain.realms', 'default_file,ldap1'
+
+  // randomize between hashed and literal username for profile uid
+  if (literalUsername) {
+    setting 'xpack.security.authc.domains.my_domain.uid_generation.literal_username', 'true'
+    setting 'xpack.security.authc.domains.my_domain.uid_generation.suffix', 'es'
+  }
```
ywangd, line 34: This just test the new settings. How the settings work is tested with `ProfileUidIntegTests`.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/Security.java
```diff
@@ -776,6 +776,7 @@ Collection<Object> createComponents(
             client,
             systemIndices.getProfileIndexManager(),
             clusterService,
+            realms::getDomainConfig,
```
ywangd, line 779: I'd like to refactor Domains into its own class in future. So for now I am passing a method reference here instead of `Realms` because I don't think `ProfileService` needs to know `Realms` which is quite heavy.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -470,7 +491,34 @@ void searchVersionedDocumentForSubject(Subject subject, ActionListener<Versioned
...
+    private ValidationException validateUsername(Subject subject) {
+        final RealmDomain realmDomain = subject.getRealm().getDomain();
+        assert realmDomain != null;
+        assert domainConfigLookup.apply(realmDomain.name()) != null;
+        assert domainConfigLookup.apply(realmDomain.name()).literalUsername();
+
+        final String username = subject.getUser().principal();
+        assert username != null;
+
+        if (username.length() < 1 || username.length() > 490) {
```
ywangd, line 508: The value `490` is somewhat artificial. Strictly speaking it is `491` when you take out the prefix (`profile_u_`) and the suffix (`_xx`). But `491` is rather an odd number. So I rounded it to `490`. It is still odd, but I don't want to contraint it to something like 100 without good reasons.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -594,6 +663,21 @@ void incrementDifferentiatorAndCreateNewProfile(Subject subject, ProfileDocument
...
 
+    private DomainConfig getDomainConfigForSubject(Subject subject) {
+        final RealmDomain realmDomain = subject.getRealm().getDomain();
+        if (realmDomain != null) {
+            final DomainConfig domainConfig = domainConfigLookup.apply(realmDomain.name());
+            if (domainConfig == null) {
+                throw new ElasticsearchException(
+                    "subject realm is under a domain [" + realmDomain.name() + "], but no associated domain config is found"
+                );
+            }
```
ywangd, line 674: Many adjustments to the tests are due to this check here. Previously it is OK to call the methods with arbitrary authentication object with domain. But it is no longer possible because this piece code checks that if the authentication has a domain, the domain must be configured in the settings.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/RealmSettings.java
```diff
@@ -119,39 +145,68 @@ public static Map<RealmConfig.RealmIdentifier, Settings> getRealmSettings(Settin
...
-            Setting<List<String>> realmsByDomainSetting = DOMAIN_TO_REALM_ASSOC_SETTING.getConcreteSettingForNamespace(domainName);
-            for (String realmName : realmsByDomainSetting.get(globalSettings)) {
-                realmToDomainsMap.computeIfAbsent(realmName, k -> new TreeSet<>()).add(domainName);
+
+            final Setting<List<String>> domainRealmsSetting = DOMAIN_TO_REALM_ASSOC_SETTING.getConcreteSettingForNamespace(domainName);
+            if (false == domainRealmsSetting.exists(globalSettings)) {
+                throw new IllegalArgumentException("[" + domainRealmsSetting.getKey() + "] must exist for security domain configuration");
+            }
+            final List<String> memberRealmNames = domainRealmsSetting.get(globalSettings);
+            // TODO: Does it make sense to have empty realms for a domain?
```
ywangd, line 163: So far we allow an empty list of realms to be defined for a domain. But it seems too much leniency. Such domain is also not tracked in current code because this method builds a map of realm to domain. If the domain has no realm, it is not included in the map. It still works because other parts of the code should never need to know about the domain. But it still feels weird to some extent. I didn't change it in this PR. But we can discuss whether this should be fixed as a bug before 8.3 is out.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -552,16 +600,18 @@ void getOrCreateProfileWithBackoff(
...
             }
         }, listener::onFailure));
     }
 
     // Package private for tests
-    void incrementDifferentiatorAndCreateNewProfile(Subject subject, ProfileDocument profileDocument, ActionListener<Profile> listener)
+    void maybeIncrementDifferentiatorAndCreateNewProfile(Subject subject, ProfileDocument profileDocument, ActionListener<Profile> listener)
         throws IOException {
         final String uid = profileDocument.uid();
         final int index = uid.lastIndexOf("_");
```
albertzaharovits, line 617: I think this is a problem because the uid suffix can contain "_".

ywangd: Good point! I think the simplest thing to do is to not allow it. I will tighten it to be just alphanumeric.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/Security.java
albertzaharovits: Personally, this doesn't bother me.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
albertzaharovits: This doesn't leave us with any room to change the profile_id generation in the future (for this generation scheme).
I don't think it is important, though I would've made it 256.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -470,7 +491,34 @@ void searchVersionedDocumentForSubject(Subject subject, ActionListener<Versioned
         });
     }
 
-    private void createNewProfile(Subject subject, String uid, ActionListener<Profile> listener) throws IOException {
+    private static final String INVALID_USERNAME_MESSAGE = "Security domain [%s] is configured to use literal username. "
+        + "As a result, creating new user profile requires the username to be at least 1 and no more than 490 characters. "
+        + "The username can contain alphanumeric characters (a-z, A-Z, 0-9), spaces, punctuation, "
+        + "and printable symbols in the Basic Latin (ASCII) block.";
+
+    private ValidationException validateUsername(Subject subject) {
```
albertzaharovits, line 499: I think this validation format (to return an exception) is ugly and should be confined to request body validation. I would prefer we just throw exceptions.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -575,6 +625,25 @@ void incrementDifferentiatorAndCreateNewProfile(Subject subject, ProfileDocument
             listener.onFailure(new ElasticsearchException("profile uid [{}] does not contain a differentiator", uid));
             return;
         }
+
+        final DomainConfig domainConfig = getDomainConfigForSubject(subject);
+        if (domainConfig != null && differentiatorString.equals(domainConfig.suffix())) {
```
albertzaharovits, line 630: I think this should be:
`if (domainConfig != null && domainConfig.suffix() != null)`

This is the case where a user that is part of a domain that doesn't do the auto-increment logic, tries to activate an existing profile of another user, which is an error (irrespective of the suffix value).

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/DomainConfig.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public DomainConfig {
+        assert (literalUsername && suffix != null) || (false == literalUsername && suffix == null);
+    }
+
+    @Override
+    public int compareTo(DomainConfig o) {
+        return name.compareTo(o.name);
+    }
+}
```
albertzaharovits, line 24: Some minor, "nit" suggestions:
* Make the `List<String> memberRealmNames` a `Set<String>` (it bothers me that the `equals` method is not intuitive if we use the List)
* I don't think implementing `Comparable` is useful.
* I think it is possible to move the `literalUsername` and `suffix` validations inside the constructor.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -470,7 +491,34 @@ void searchVersionedDocumentForSubject(Subject subject, ActionListener<Versioned
...
     }
 
-    private void createNewProfile(Subject subject, String uid, ActionListener<Profile> listener) throws IOException {
+    private static final String INVALID_USERNAME_MESSAGE = "Security domain [%s] is configured to use literal username. "
+        + "As a result, creating new user profile requires the username to be at least 1 and no more than 490 characters. "
+        + "The username can contain alphanumeric characters (a-z, A-Z, 0-9), spaces, punctuation, "
+        + "and printable symbols in the Basic Latin (ASCII) block.";
+
+    private ValidationException validateUsername(Subject subject) {
+        final RealmDomain realmDomain = subject.getRealm().getDomain();
```
albertzaharovits, line 500: I wonder if it would be possilble to have the two new domain settings inside the `RealmDomain`.
I think it would simplify some stuff. 
This is a suggestion for when we get to spend some time on refactorings.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
ywangd: In practice, the `differentiatorString` should always equal to `domainConfig.suffix()` because when the code reaches here, it is because a profile with UID ending with `domainConfig.suffix()` is owned by the user.

But I think you are also correct in that the logic should be strict and independant here. I'll update as suggested.

albertzaharovits: Yeah, it should be an internal error if they're not equal.

albertzaharovits (approved):
LGTM

Food for thought: have we considered adding a node version (or any version) to the profile id?  I think when strings start to get a format it is a good idea to add a version tag. But maybe profile ids are special.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
ywangd: I changed to throw `ElasticsearchException`

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
ywangd: I thought about that and decided against it. It could simplify the check in ProfileService. But the benefit isn't significant. Also these information becomes something sent across the wire which feels weird because:
1. They are not needed outside of the local node.
2. There are some similarity between`RealmRef`/`RealmConfig` and `RealmDomain`/`DomainConfig` in that the Config class is a separate thing from the reference class. If we ever add more settings under domain, it helps to separate them earlier.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/DomainConfig.java
ywangd: * I changed `memberRealmNames` to be a `Set`
* It is needed because the value of `realmToDomainsMap` is a [`TreeSet`](https://github.com/elastic/elasticsearch/blob/8c2f4292ec4b78e8c4eaec44443f0a32532cdf38/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/RealmSettings.java#L133). I inherited this code and it seems implementing `Comparable` is the least effort.
* I moved them into the constructor. I also tried to further beefy up `DomainConfig` to make it handle `Settings` in its constructor (similar to `RealmConfig`). But this causes cascading changes in tests and is a bit out of hand for this PR. Fortunately, it's all internal code. We can definitely refactor this later.

ywangd:
> Food for thought: have we considered adding a node version (or any version) to the profile id? I think when strings start to get a format it is a good idea to add a version tag. But maybe profile ids are special.

Adding version to the profile uid will likely violate the requirement of two-way translatable between username and uid. Because version is meant to change. Therefore it is not possible to tell the uid of a given username because it could be of any version.

But I think it could be an useful thing to have in the profile document. We have a task to finalise the index mappings before release. This seems to fit it there.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
ywangd: I think you have a point here. I changed it to 256. Thanks!

PR #86954, opened 2022-05-20
Fix needless capturing lambda in analyzer wrapper

original-brownbear:
Just a random thing I found when inspecting a heap dump that had ~100M
of these in it.


romseygeek (approved):
LGTM

original-brownbear:
Thanks Alan!

PR #86970, opened 2022-05-20
Simplify master task failure handling

DaveCTurner:
Today a batching executor must submit an `ActionListener<ClusterState>`
for every task in the batch that succeeds. This listener allows to
consume the published cluster state, or any failure that occurs during the
publication following the successful task execution.

There's no real reason for an executor to customise the failure handling
here, and in practice these listeners always defer failures to the
task's `onFailure()` method. This commit removes the ability for
executors to handle failures differently here by replacing the
`ActionListener<ClusterState>` with a simple `Consumer<ClusterState>`.

Moreover we consider it an antipattern to actually consume the published
`ClusterState`, so this commit introduces variants that ignore the
`ClusterState` and accept a `Runnable` instead.


idegtiarenko (approved):

DaveCTurner:
Thanks Ievgen!

PR #86973, opened 2022-05-20
Fix ElasticsearchJavaModulePathPluginFuncTest on Windows

ChrisHegarty:
Re-enable temporarily disabled ElasticsearchJavaModulePathPluginFuncTest on Windows.

We can simplify the regex by just flipping between logical separators on Windows versus non-Windows platforms - since Windows uses `;` for the path-separator character.


rjernst (approved):
LGTM

PR #86979, opened 2022-05-20
Don't output default values in text query xcontent representations

romseygeek:
This removes default values from the xcontent representations of the
following queries:
* intervals
* combined_fields
* match_bool_prefix
* match_phrase_prefix
* match_phrase
* multi_match
* query_string
* simple_query_string

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/query/SimpleQueryStringBuilder.java
+++ b/server/src/main/java/org/elasticsearch/index/query/SimpleQueryStringBuilder.java
@@ -433,12 +433,18 @@ public class SimpleQueryStringBuilder extends AbstractQueryBuilder<SimpleQuerySt
             builder.field(ANALYZER_FIELD.getPreferredName(), analyzer);
         }
 
-        builder.field(FLAGS_FIELD.getPreferredName(), flags);
-        builder.field(DEFAULT_OPERATOR_FIELD.getPreferredName(), defaultOperator.name().toLowerCase(Locale.ROOT));
+        if (flags != DEFAULT_FLAGS) {
+            builder.field(FLAGS_FIELD.getPreferredName(), flags);
+        }
+        if (defaultOperator != DEFAULT_OPERATOR) {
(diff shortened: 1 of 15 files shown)


server/src/main/java/org/elasticsearch/index/query/SimpleQueryStringBuilder.java
```diff
@@ -433,24 +433,38 @@ protected void doXContent(XContentBuilder builder, Params params) throws IOExcep
...
+            builder.field(GENERATE_SYNONYMS_PHRASE_QUERY.getPreferredName(), settings.autoGenerateSynonymsPhraseQuery());
+        }
+        if (settings.fuzzyPrefixLength() != DEFAULT_FUZZY_PREFIX_LENGTH) {
+            builder.field(FUZZY_PREFIX_LENGTH_FIELD.getPreferredName(), settings.fuzzyPrefixLength());
+        }
+        if (settings.fuzzyMaxExpansions() != DEFAULT_FUZZY_MAX_EXPANSIONS) {
+            builder.field(FUZZY_MAX_EXPANSIONS_FIELD.getPreferredName(), settings.fuzzyMaxExpansions());
+        }
+        if (settings.fuzzyTranspositions() != DEFAULT_FUZZY_TRANSPOSITIONS) {
+            builder.field(FUZZY_TRANSPOSITIONS_FIELD.getPreferredName(), settings.fuzzyTranspositions());
```
javanna, line 465: do we need some common method somewhere that given the default value does the if internally and eventually prints out the field? It is a bit of a shame to have to repeat these ifs everywhere, maybe?

romseygeek: I thought about that but typing makings things difficult, because the type of the value itself doesn't necessarily reflect the type of the `builder.field()` initial parameter.

javanna: I see!

javanna (approved):
LGTM thanks @romseygeek

PR #86981, opened 2022-05-20
Fix null message in output

grcevski:
When launching Elasticsearch, the node auto configuration code might throw an exception because security is disabled or already setup. If this node is not in enrolment mode, the exception is thrown with null message, which we should ignore and not print.


distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ServerCli.java
```diff
@@ -153,7 +153,9 @@ private Environment autoConfigureSecurity(
             };
             if (options.has(enrollmentTokenOption) == false && okCode) {
                 // we still want to print the error, just don't fail startup
-                terminal.errorPrintln(e.getMessage());
+                if (e.getMessage() != null) {
```
rjernst, line 156: Can you add a test for this in ServerCliTests? The mock autoconfig should be able to replicate the conditions.

rjernst (approved):
Looks good, one ask

PR #86984, opened 2022-05-20
Improve error message for search API url parameters

### LINKED ISSUE #79719: Improve error message for search API query parameters that require other parameters
Some search API query parameters, such as `analyzer`, require the `q` query parameter. Similarly, the `suggest_mode` query parameter requires both the `suggest_field` and `suggest_text` query parameters.

If you attempt to use these parameters without the required parameters, Elasticsearch returns the following message:

```
request [/<index>/_search] contains unrecognized parameter: [<parameter>]
```

This suggests the parameter is invalid, which isn't true. It'd be more helpful to list the other required parameters.

Relates to #79674 and #79716.

cbuescher:
The search API uses url query parameters that require other parameters to be
present. When one of them is ommited, the other depending parameters might not
be consumed, leading to an error message about unconsumed parameters from the
BaseRestHandler that are not actionable by the user.
This change adds improvements for url query parameters that require the
existance of the 'q' query string parameter and also adds better error message
for suggestion url parameter handling.

Closes #79719

### DIFF
--- a/docs/changelog/86984.yaml
+++ b/docs/changelog/86984.yaml
@@ -1,0 +1,6 @@
+pr: 86984
+summary: Improve error message for search API url parameters
+area: Search
+type: enhancement
+issues:
+ - 79719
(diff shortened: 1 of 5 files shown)


romseygeek (approved):
LGTM, nice cleanup.

cbuescher:
@romseygeek thanks a lot for the review.

PR #86985, opened 2022-05-20
Remove default values from term and span query xcontent

romseygeek:
Mostly this is just removing boosts, but it also simplifies span_not
slightly.  It also removes the boost parameter from `term`, as that
shares an implementation with `span_term`.

Relates to #76515


cbuescher (approved):
LGTM

romseygeek:

PR #86986, opened 2022-05-20
Fix versions in serial diff tests

nik9000:
This adds some missing skips to make the tests for the serial_diff
always pass and test the current version. The tests contain a
`fixed_interval` which is only supported in 7.2+. That's fine probably
ok - we don't expect to make backwards breaking changes at all, much
less breaking with 7.1 and *not* 7.2


nik9000:
Hey label bot I've fixed them.

PR #86989, opened 2022-05-20
[ML] adds start and end params to _preview and excludes cold/frozen tiers from unbounded previews

benwtrent:
In larger clusters with complicated datafeed requirements, being able to preview only a specific window of time is important. Previously, datafeed previews would always start at `0` (or from the beginning of the data). This causes issues if the index pattern contains indices on slower hardware, but when the datafeed is actually started, the "start" time is set to more recent data (and thus on faster hardware).

Additionally, when _preview is unbounded (as before), it attempts to only preview indices that are NOT frozen or cold. This is done through a query against the `_tier` field. Meaning, it only effects newer indices that actually have that field set.

### DIFF
--- a/docs/changelog/86989.yaml
+++ b/docs/changelog/86989.yaml
@@ -1,0 +1,6 @@
+pr: 86989
+summary: Adds start and end params to `_preview` and excludes cold/frozen tiers from
+  unbounded previews
+area: Machine Learning
+type: enhancement
+issues: []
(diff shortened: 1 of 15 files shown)


droberts195 (approved):
LGTM

lcawl (commented):
Docs LGTM, thanks for adding!

droberts195:
Kibana was changed to make use of this in https://github.com/elastic/kibana/pull/134073

PR #86990, opened 2022-05-20
Adjust osprobe assertion for burst cpu

rjernst:
When gathering v1 cgroup cpu stats, the file is expected to have 3
lines. However, if the burst feature of cpu accounting is enabled, it
will actually have 5 lines. This commit adjusts the assertion to allow
for the extra lines. We ignore these lines already in the parsing of the
calling method. Unfortunately this is not currently easy to test, and no
tests exist. Give it is just an assertion that is relaxing, I think it's
ok to fix this issue, and separately work on revamping the OsProbe
tests.

### DIFF
--- a/docs/changelog/86990.yaml
+++ b/docs/changelog/86990.yaml
@@ -1,0 +1,5 @@
+pr: 86990
+summary: Adjust osprobe assertion for burst cpu
+area: Infra/Core
+type: bug
+issues: []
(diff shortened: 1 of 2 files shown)


grcevski (approved):
LGTM!

javanna:
thanks a lot for addressing this, it will definitely make my life easier :)

PR #86997, opened 2022-05-20
[8.2] Adjust osprobe assertion for burst cpu (#86990)

rjernst:
 - Adjust osprobe assertion for burst cpu (#86990)


rjernst:
@elasticmachine run 
elasticsearch-ci/rest-compatibility

PR #87004, opened 2022-05-20
[Refactor] Use Lists instead of Maps for SystemIndices features

williamrandolph:
This PR gets rid of a lot of crufty code in `SystemIndices` and related classes. Basically, we held Feature objects in maps of feature name to feature, implicitly expecting that the map key would match the value of Feature#getName. This was easy to mess up in tests.

The history here is that `Feature#getName` was added to the Feature class after all those maps had been written, but I never cleaned up the old code.

One place we did take advantage of the map structure was in the snapshot/restore code, so I've added a `SystemIndices#getFeature` method to replace it.

I also renamed some variables as I was going to better reflect what was actually being passed around.

If approved, I'm planning to backport this so that we can more easily apply bugfixes in the system indices code.

### DIFF
--- a/server/src/main/java/org/elasticsearch/indices/SystemIndices.java
+++ b/server/src/main/java/org/elasticsearch/indices/SystemIndices.java
@@ -84,11 +84,10 @@ public class SystemIndices {
 
     /**
      * Initialize the SystemIndices object
-     * @param pluginAndModulesDescriptors A map of this node's feature names to
-     *                                    feature objects.
+     * @param features A list of features from which we will load system indices
      */
-    public SystemIndices(Map<String, Feature> pluginAndModulesDescriptors) {
-        featureDescriptors = buildSystemIndexDescriptorMap(pluginAndModulesDescriptors);
+    public SystemIndices(List<Feature> features) {
+        featureDescriptors = buildSystemIndexDescriptorMap(features);
         indexDescriptors = featureDescriptors.values()
             .stream()
             .flatMap(f -> f.getIndexDescriptors().stream())
@@ -577,9 +576,9 @@ public class SystemIndices {
         return Operations.isEmpty(Operations.intersection(a1Automaton, a2Automaton)) == false;
     }
 
-    private static Map<String, Feature> buildSystemIndexDescriptorMap(Map<String, Feature> featuresMap) {
-        final Map<String, Feature> map = Maps.newMapWithExpectedSize(featuresMap.size() + SERVER_SYSTEM_INDEX_DESCRIPTORS.size());
-        map.putAll(featuresMap);
+    private static Map<String, Feature> buildSystemIndexDescriptorMap(List<Feature> features) {
+        final Map<String, Feature> map = Maps.newMapWithExpectedSize(features.size() + SERVER_SYSTEM_INDEX_DESCRIPTORS.size());
+        features.forEach(f -> map.put(f.getName(), f));
         // put the server items last since we expect less of them
         SERVER_SYSTEM_INDEX_DESCRIPTORS.forEach((source, feature) -> {
             if (map.putIfAbsent(source, feature) != null) {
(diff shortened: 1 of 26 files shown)


server/src/main/java/org/elasticsearch/snapshots/RestoreService.java
```diff
@@ -336,9 +336,8 @@ private void startRestore(
             .flatMap(Collection::stream)
             .collect(Collectors.toSet());
 
-        final Map<String, SystemIndices.Feature> featureSet = systemIndices.getFeatures();
         final Set<String> featureStateDataStreams = featureStatesToRestore.keySet().stream().filter(featureName -> {
-            if (featureSet.containsKey(featureName)) {
+            if (Objects.nonNull(systemIndices.getFeature(featureName))) {
```
williamrandolph, line 340: This class has most of the cases where I used the new `SystemIndices#getFeature` method.

server/src/main/java/org/elasticsearch/indices/SystemIndices.java
```diff
@@ -334,18 +329,21 @@ public Predicate<String> getProductSystemIndexNamePredicate(ThreadContext thread
         return automaton::run;
     }
 
-    public Map<String, Feature> getFeatures() {
-        return featureDescriptors;
+    public Feature getFeature(String name) {
+        return getFeatures().stream().filter(f -> Objects.equals(name, f.getName())).findFirst().orElse(null);
```
rjernst, line 333: This turns looking up a feature by name into a linear operation. Is that ok? Seems like we should still perhaps have the map internally but not expose it? It would seem trivial to build the structure and keep it inside SystemIndices as a private member, but still expose the list of features elsewhere? Alternatively, we could keep the list, but sort it, and then use Collections.binarySearch. It won't be as fast as a hash lookup, but at least it would not be linear, and would probably be good enough for the limited use of looking up a specific feature.

server/src/main/java/org/elasticsearch/indices/SystemIndices.java
```diff
@@ -577,22 +568,26 @@ private static boolean overlaps(String pattern1, String pattern2) {
...
+
+        features.forEach(f -> map.put(f.getName(), f));
+
+        List<String> providedFeatureNames = features.stream().map(Feature::getName).toList();
+
         // put the server items last since we expect less of them
-        SERVER_SYSTEM_INDEX_DESCRIPTORS.forEach((source, feature) -> {
-            if (map.putIfAbsent(source, feature) != null) {
+        SERVER_SYSTEM_INDEX_DESCRIPTORS.forEach(feature -> {
+            if (providedFeatureNames.contains(feature.getName())) {
```
rjernst, line 580: This is going to be linear complexity. It looks like this method is only called during initialization, so maybe it is ok to still pass in a Map, and then reduce that to a list of features later cone all the building/validation is complete?

rjernst (changes requested):
How many feature objects does the system have? I'm slightly concerned about making lookups for a specific feature linear.

server/src/main/java/org/elasticsearch/indices/SystemIndices.java
williamrandolph: Right now I think there are fifteen or sixteen features at most, but of course features can be added by plugins, so I take your point. I'll go back to using a map internally.

server/src/main/java/org/elasticsearch/indices/SystemIndices.java
williamrandolph: I've reverted to the old code that uses the `Map#put` method to detect feature name clashes.

williamrandolph:
@rjernst I think I've backed out the changes that raised your performance concerns. I still like having the constructor take a `List` parameter because it ensures that we don't have a mismatch between the key and the feature name.

server/src/main/java/org/elasticsearch/indices/SystemIndices.java
```diff
@@ -65,10 +66,9 @@ public class SystemIndices {
 
     private static final Automaton EMPTY = Automata.makeEmpty();
 
-    private static final Map<String, Feature> SERVER_SYSTEM_INDEX_DESCRIPTORS = Map.of(
-        TASKS_FEATURE_NAME,
+    private static final Map<String, Feature> SERVER_SYSTEM_INDEX_DESCRIPTORS = Stream.of(
         new Feature(TASKS_FEATURE_NAME, "Manages task results", List.of(TASKS_DESCRIPTOR))
-    );
+    ).collect(Collectors.toMap(Feature::getName, Function.identity()));
```
rjernst, line 71: toUnmodifiableMap?

rjernst (approved):
LGTM

Issue #78744, opened 2021-10-06
Modularization Investigation


ChrisHegarty:
This issue has been logged to help track various aspects of the effort to modularize the core of elasticsearch. This issue will likely not deliver any specific changes, but rather serve to identify and coordinate other issues / PRs that will deliver the actual changes. Tasks will be added and expended over time.

[ ] Resolve split packages (#78166)
[x] Start the server with the module-path (prototype in #78711)
[ ] Testing
  [ ] Black/White box test infrastructure, --patch-module
  [ ] Majority of tests pass
  [ ] All tests pass
[ ] Declare explicit modules

ChrisHegarty:
Some preliminary test results from the module-path branch in #78711.

```
$ ./gradlew :server:test
...
12108 tests completed, 126 failed, 54 skipped
```

ChrisHegarty:
Some preliminary test results from the module-path2 branch in #81066 (supersedes  #78711). This, module-path2, version contains much less changes, since most of the split package issues have been merged into the mainline branch.

PR #87005, opened 2022-05-20
Modularize the analysis-common component

ChrisHegarty:
This change modularizes analysis-common, by adding a module-info.java

The project only requires painless SPI to compile, so that was fixed along the way ( so that the compile module path can be inferred directly from the dependencies ).

relates #78744

### DIFF
--- a/modules/analysis-common/build.gradle
+++ b/modules/analysis-common/build.gradle
@@ -22,7 +22,7 @@ restResources {
 }
 
 dependencies {
-  compileOnly project(':modules:lang-painless')
+  compileOnly project(':modules:lang-painless:spi')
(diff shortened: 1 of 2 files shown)


rjernst (approved):
LGTM

rjernst:
It looks like something is off still with the painless extension, the test failure looks related.

ChrisHegarty:
> It looks like something is off still with the painless extension, the test failure looks related.

Yeah, painless needs to introspect. Adding exports and opens.  (the exports should be qualified, but we need to add a little support for cross-layer sibling exports/opens to the code controlling the module layers - will do this soon)

Issue #26220, opened 2017-08-15
Comprehensive aggregation REST tests


polyfractal:
There are only a few REST tests that verify the body of aggregation results.  We should have a more comprehensive suite of tests that check the body under a variety of conditions and options.

### Metrics:
- [ ] Avg Aggregation
- [ ] Cardinality Aggregation
- [ ] Extended Stats Aggregation
- [ ] Geo Bounds Aggregation
- [ ] Geo Centroid Aggregation
- [ ] Max Aggregation
- [ ] Min Aggregation
- [ ] Percentiles Aggregation
- [ ] Percentile Ranks Aggregation
- [ ] Scripted Metric Aggregation
- [ ] Stats Aggregation
- [ ] Sum Aggregation
- [ ] Top hits Aggregation
- [ ] Value Count Aggregation

### Bucketing
- [ ] Children Aggregation
- [ ] Date Range Aggregation
- [ ] Diversified Sampler Aggregation
- [ ] Filters Aggregation
- [ ] Geo Distance Aggregation
- [ ] GeoHash grid Aggregation
- [ ] Global Aggregation
- [ ] Histogram Aggregation
- [ ] IP Range Aggregation
- [ ] Missing Aggregation
- [ ] Nested Aggregation
- [ ] Reverse nested Aggregation
- [ ] Sampler Aggregation

### Pipeline
- [ ] Avg Bucket Aggregation
- [ ] Derivative Aggregation
- [ ] Max Bucket Aggregation
- [ ] Min Bucket Aggregation
- [ ] Sum Bucket Aggregation
- [ ] Stats Bucket Aggregation
- [ ] Extended Stats Bucket Aggregation
- [ ] Percentiles Bucket Aggregation
- [ ] Moving Average Aggregation
- [ ] Cumulative Sum Aggregation
- [ ] Bucket Script Aggregation
- [ ] Bucket Selector Aggregation
- [ ] Serial Differencing Aggregation

### Matrix
- [ ] Matrix Stats

### Review existing tests
- [ ] Terms Aggregation
- [ ] Date Histogram Aggregation
- [ ] Significant Terms Aggregation
- [ ] Range Aggregation
- [ ] Adjacency Matrix Aggregation
- [ ] Filter Aggregation

javanna:
Question: do we want to have these as yaml tests so the language clients can run them too? Or shall we consider making them java tests that se the low-level REST client instead?

polyfractal:
[...]

PR #87009, opened 2022-05-20
REST tests for avg/max/min/sum_bucket aggs

nik9000:
Adds REST layer tests for the `avg_bucket`, `max_bucket`, `min_bucket`,
and `sum_bucket` pipeline aggregations. This gives us forwards and
backwards compatibility tests for these aggs as well as mixed version
cluster tests for these aggs.

Relates to #26220

### DIFF
--- a/rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.aggregation/500_avg_bucket.yml
+++ b/rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.aggregation/500_avg_bucket.yml
@@ -1,0 +1,149 @@
+setup:
+  - do:
+      bulk:
+        index: no_gaps
+        refresh: true
+        body:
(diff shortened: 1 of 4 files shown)


csoulios (approved):
LGTM

I think tests cover the happy path and I wonder if we should test edge or failing cases, such as:

- The path defined `buckets_path`does not exist
- [Dealing with dots in bucket names](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-pipeline.html#dots-in-agg-names)
- bucket agg returns no results (eg we use a query to filter docs)

On the other hand, testing the happy path is much better than what we had. We can add more tests to cover edge cases with another PR.

nik9000:
> On the other hand, testing the happy path is much better than what we had. We can add more tests to cover edge cases with another PR.

Yeah. That'd be good too.

PR #87011, opened 2022-05-20
Fix status code when open point in time without keep_alive

### LINKED ISSUE #87003: Creating a PIT without `?keep_alive` parameter returns 500 rather than 4xx
### Elasticsearch Version

8.2.0

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

Cloud

### Problem Description

Creating a PIT without the required `?keep_alive` parameter throws a NPE which maps to HTTP status code `500 Internal Server Error` whereas really this isn't a server-side error, it's a bad request, so the status code should be something like `400 Bad Request`

### Steps to Reproduce

```
# Request:
POST /_pit

# Response:
# 500 Internal Server Error
# {
#   "status": 500,
#   "error": {
#     "reason": "[keep_alive] parameter must be non null",
#     "root_cause": [
#       {
#         "reason": "[keep_alive] parameter must be non null",
#         "type": "null_pointer_exception"
#       }
#     ],
#     "type": "null_pointer_exception"
#   }
# }
```

### Logs (if relevant)

_No response_

dnhatn:
Closes #87003


ywelsch (approved):
LGTM. I'm just wondering if it is linking to the right issue?

dnhatn:
@ywelsch Thanks for reviews. Yes, the issue should be #87003.

PR #87012, opened 2022-05-20
Improve server process tests

rjernst:
When the server process dies, the cli should reflect the unexpected exit
of the process in its return from waitFor. The test for this closed the
input side of the pipe, but that is where the buffer lives. This means
there is a race condition in the cli testcase fully reading the final
message expected. This commit removes some complexity from the test by
using a single latch, and closing the pipe from the process side to
mimic the break.


williamrandolph (approved):
LGTM!

PR #87015, opened 2022-05-21
87014 Changed to correct default value for number_of_shards setting

amitmbm:


DaveCTurner:
The original code is what we want. With this change we lose the ability to detect cases where the setting is missing such as here:

https://github.com/elastic/elasticsearch/blob/ab5cb89c172e0136d7d594b58e81ea311c634370/server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetadata.java#L1543-L1545

Since no change is needed in this area I am closing this.

PR #87016, opened 2022-05-21
Remove some blocking in CcrRepository

DaveCTurner:
Today `CcrRepository#getRepositoryData` and
`CcrRepository#getSnapshotInfo` block the calling thread pending receipt
of the full metadata of the remote cluster. It would be preferable if
they didn't get the full cluster metadata at all, but we can at least
remove the blocking waits here.


x-pack/plugin/ccr/src/main/java/org/elasticsearch/xpack/ccr/repository/CcrRepository.java
```diff
@@ -259,44 +258,41 @@ public IndexMetadata getSnapshotIndexMetaData(RepositoryData repositoryData, Sna
...
-            for (String indexName : remoteMetadata.getConcreteAllIndices()) {
-                // Both the Snapshot name and UUID are set to _latest_
-                SnapshotId snapshotId = new SnapshotId(LATEST, LATEST);
-                copiedSnapshotIds.put(indexName, snapshotId);
-                final long nowMillis = threadPool.absoluteTimeInMillis();
-                snapshotsDetails.put(
-                    indexName,
-                    new RepositoryData.SnapshotDetails(SnapshotState.SUCCESS, Version.CURRENT, nowMillis, nowMillis, "")
+        try {
+            getRemoteClusterClient().admin().cluster().prepareState().clear().setMetadata(true).execute(listener.map(response -> {
```
DaveCTurner, line 262: NB this moves the work of iterating over all the indices in the remote metadata and building the `remoteIndices` map onto a transport worker thread. We could fork back to `SNAPSHOT` to do this, not sure if it's worth it tho.

It also completes the outer listener on a transport worker, but that's ok according to the Javadoc on `getRepositoryData()`.

Finally it removes the timeout on getting the remote state. I don't think we really need an application-level timeout here if we're not blocking, since the transport layer ensures we will get a response eventually anyway.

x-pack/plugin/ccr/src/main/java/org/elasticsearch/xpack/ccr/repository/CcrRepository.java
```diff
@@ -258,44 +257,41 @@ public IndexMetadata getSnapshotIndexMetaData(RepositoryData repositoryData, Sna
...
+                Metadata remoteMetadata = response.getState().getMetadata();
+
+                Map<String, SnapshotId> copiedSnapshotIds = new HashMap<>();
```
henningandersen, line 264: Maybe we can update the hashmap sizing in the next two lines to be based on `remoteIndices.size()` instead? Given that it runs on transport thread, that sort of makes sense to me.

x-pack/plugin/ccr/src/main/java/org/elasticsearch/xpack/ccr/repository/CcrRepository.java
henningandersen: > could fork back to SNAPSHOT to do this, not sure if it's worth it tho.

Agree, we must have just deserialized on transport thread anyway.

> removes the timeout on getting the remote state. I don't think we really need an application-level timeout here if we're not blocking

It does slightly change the behavior of put-follow under such circumstances - I think it would fail after the timeout rather than wait today. But then again, that seems unlikely to be the intention and actually succeeding if we can seems better to me too.

henningandersen (approved):
LGTM.

PR #87019, opened 2022-05-22
Temporarily skip server module tests on JDK 19 until a JDK bug is resolved

ChrisHegarty:
Temporarily skip server module tests on JDK 19, until this issue is resolved https://bugs.openjdk.java.net/browse/JDK-8287097


rjernst (approved):
LGTM

PR #87020, opened 2022-05-22
[Refactor] Convert ExecutorNames POJO to Record object

williamrandolph:
ExecutorNames was a very simple data-holding class that easily works as a Java record.


rjernst (approved):
LGTM

PR #87030, opened 2022-05-23
Allow the modify data stream api to remove broken reference to a backing index

martijnvg:
This adds a `force_remove_backing_index` action to the modify data stream api to allow removing broken reference to backing indices from a data stream.

```
POST _data_stream/_modify
{
  "actions": [
    {
      "force_remove_backing_index": {
        "data_stream": "my-logs",
        "index": ".ds-my-logs-2099.01.01-000001"
      }
    }
  ]
}
```


DaveCTurner (commented):
Sorry if this is a silly question but why introduce a new command rather than making `removeBackingIndex` robust to this? Put differently, why would users run `removeBackingIndex` when `forceRemoveBackingIndex` does the same thing and succeeds in more cases?

martijnvg:
> Sorry if this is a silly question but why introduce a new command rather than making removeBackingIndex robust to this? 

I added a new command here, because the normal existing remove_backing_index is build on the assumption that the cluster state is in a valid state and validates that the backing index being removed actually exists. I think under normal circumstances we want this checks to be done? For example when a user just wants to add a restored index to a data stream? I added `force_remove_backing_index` to be used only in case of emergency. But I can change the pr that `remove_backing_index` does what `force_remove_backing_index` does. I don't have a very strong opinion about this.

DaveCTurner (approved):
LGTM thanks @martijnvg

dakrone (approved):
LGTM

ineo48 (commented):
See changed

PR #87031, opened 2022-05-23
Cleanup dead variable in restore task

original-brownbear:
This is unused and it's confusing to have a reference to this in the restore task
because these things hold the repository generation.


pxsalehi:
While I've just started looking into the code, that variable is indeed not used, and as the tests are green, LGTM!

pxsalehi (approved):

original-brownbear:
Thanks Pooya!

PR #87044, opened 2022-05-23
[TEST] assert busy waiting for dynamic update

javanna:
Dynamic updates are not guaranteed to be applied on the master node after an index operation, hence we need to wrap the get mappings call with an assertBusy and retry until the updated mappings made it to the master.


ywelsch (approved):
LGTM

PR #87047, opened 2022-05-23
Change Metadata indices to return Map

rjernst:
The indices/getIndices methods of Metadata expose the impl detail of
using an ImmutableOpenMap. This commit changes the return type to Map.
For the one case in rollup where the map was fed back to an
ImmutableOpenMap builder, the type for the builder ctor was relaxed to
Map with special casing an existing ImmutableOpenMap.

relates #86239


original-brownbear (approved):
LGTM!

server/src/main/java/org/elasticsearch/common/collect/ImmutableOpenMap.java
```diff
@@ -368,8 +368,13 @@ public static <KType, VType> Builder<KType, VType> builder(int size) {
         return new Builder<>(size);
     }
 
-    public static <KType, VType> Builder<KType, VType> builder(ImmutableOpenMap<KType, VType> map) {
-        return new Builder<>(map);
+    public static <KType, VType> Builder<KType, VType> builder(Map<KType, VType> map) {
+        if (map instanceof ImmutableOpenMap<KType, VType> iom) {
+            return new Builder<>(iom);
```
original-brownbear, line 373: nice :)

PR #87049, opened 2022-05-23
[8.2] [Refactor] Use Lists instead of Maps for SystemIndices features (#87004)

williamrandolph:
The SystemIndices constructor should take a list instead of a map as an
argument so that we can guarantee that the map we use for feature lookups is
keyed on the feature name.

We also provide some new getter methods so that calling code does not have to
handle the map directly.


williamrandolph:
Manual backports: #87063, #87065, #87067

PR #87051, opened 2022-05-23
Guard for adding null value tags to vector tiles

iverase:
It is possible for the fields API to return null objects when a field is set to ignore_malformed values and the values are actually malformed. This causes a NPE on vector tiles if trying to add the value as the value of a tag. This PR adds a guard so tags with null values are ignored.


craigtaverner (approved):
LGTM - even the with_labels code path should use the same `addPropertyToFeature` method.

PR #87052, opened 2022-05-23
Add Resolve Index API to the "read" permission for an index

### LINKED ISSUE #86977: Permissions issue for the resolve_index api
### Elasticsearch Version

latest

### Installed Plugins

_No response_

### Java Version

18.0.1.1

### OS Version

Linux 5.17.6-1-MANJARO

### Problem Description

We created a `hasData` service which utilizes the `resolve_index` api to check whether ES data exists. When making a call to `internal/index-pattern-management/resolve_index` to query whether data exists or not, in certain scenarios it fails with a 500 Internal Server Error.

The ES error being - 

> security_exception: [security_exception] Reason: action [indices:admin/resolve/index] is unauthorized for user [test_user] with roles [test_dashboard_user,reporting_user], this
action is granted by the index privileges [view_index_metadata,manage,all]

The scenarios being:
Scenario one:
1. There are 3 indices
2. User has `read` access to 1 of the indices, but not to the other two
3. The call fails with a 500 Internal Server Error. The user should be able to see the index they have access to.

Scenario two:
1. This refers to a scenario where the user has `read` access to the index, but `no view_index_metadata` privileges.

Original issue ticket in Kibana for reference - [#132478](https://github.com/elastic/kibana/issues/132478)

### Steps to Reproduce

The user has `read` access to the index, but `no view_index_metadata` privileges.
The api fails with a 500 due to the missing permissions.

### Logs (if relevant)

security_exception: [security_exception] Reason: action [indices:admin/resolve/index] is unauthorized for user [test_user] with roles [test_dashboard_user,reporting_user], this
action is granted by the index privileges [view_index_metadata,manage,all]

dakrone:
This commit allows the index "read" permission to also allow using the resolve index Action.

Resolves #86977


dakrone:
@tvernum from the DM side I don't have any problems with this, but I would like to have someone on the Security team side give a :heavy_check_mark: before merging any change like this in.

tvernum:
> Also, from the security side, should this go to 7.17.x also?

I don't think so. From our side it isn't a bug - it works as designed it's just that the intended behaviour isn't a good fit for Kibana.
If backporting to 7.17 would provide substantial value to Kibana it's probably OK to do that, but it's not a priority from our side.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/security/authz/60_resolve_index.yml
```diff
@@ -122,13 +122,6 @@ teardown:
   - match: { data_streams.0.backing_indices.0: $idx0name }
   - match: { data_streams.0.timestamp_field: "@timestamp" }
 
-  # should fail because no_authz_user is not authorized for view_index_metadata
-  - do:
-      catch: forbidden
-      headers: { Authorization: "Basic bm9fYXV0aHpfdXNlcjp4LXBhY2stdGVzdC1wYXNzd29yZA==" } # no_authz_user
-      indices.resolve_index:
-        name: '*'
-
```
tvernum, line 131: I'd prefer we change the `data_stream_role2` role to use `monitor` instead of `read` and then kept this test.

tvernum (approved):
LGTM, but I'd like to keep the negative test (either here or in followup PR)

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/security/authz/60_resolve_index.yml
dakrone: Sounds good, I pushed a change that does this. I just can't figure out why it fails on the v7 compatibility yet.

dakrone:
@elasticmachine update branch

dakrone:
@elasticmachine update branch

PR #87053, opened 2022-05-23
Added ability to not index a wildcard field

not-napoleon:
This is some kludgery to make it so we can optionally not index wildcard fields.  This *SHOULD NOT* be merged to production, it's only intended as a prototype for testing.  Largely untested.


jpountz (approved):

not-napoleon:
Since this is just for proof of concept prototyping, I'm going to merge this with the failing test.  It's not going to master.

PR #87054, opened 2022-05-23
Remove "Push back excessive requests for stats (#83832)"

joegallo:

This is a continuation of #85504. We don't want to ship #83832 (stats request rate limiting) with 8.3, because that rate limiting ends up causing problems for ILM with many indices (#85333). For the time being, we're going to revert #83832 from master (via this PR), so that we can re-approach #51992 with #85333 in mind.


dakrone (approved):
LGTM if CI is happy, since I assume this is a straightforward revert?

joegallo:
> since I assume this is a straightforward revert?

Indeed -- there were two very minor conflicts due to changes in the surrounding context, nothing fancy.

PR #87058, opened 2022-05-23
Remove startup sleep time from archive tests

rjernst:
The packaging tests had an env var that would sleep for some seconds to
allow Elasticsearch to startup. This commit removes the env var as it is
no longer used or necessary.


mark-vieira (approved):
LGTM

PR #87059, opened 2022-05-23
Assert that some getters are not accessed for unmanaged indices

williamrandolph:
There are many properties in SystemIndexDescriptor that should only be accessed when dealing with managed system indices. Any code that accesses the getters for these properties should be inside of a block that checks whether an index is managed or not. Here, we add assertions to verify that this is the case, or to point us to places where our code may be buggy.


rjernst (approved):
LGTM

PR #87061, opened 2022-05-23
Add unstable plugin API telemetry

rjernst:
This commit adds some new information to the node info and cluster stats responses regarding plugins. For each plugin returned, in addition to the plugin descriptor, a boolean indicating whether the plugin is an official plugin, as well as info on the particular interfaces and methods that the plugin implements are returned. This will allow measuring the effectiveness over time of the stable plugin API.


ChrisHegarty (approved):
LGTM. This telemetry is going to be so helpful.

bytebilly:
@rjernst could listing legacy items for official plugins be useful to internal development?
From a telemetry perspective, we can use the `is_official` value to include those results or not, so it would not create noise.

rjernst:
I think it would just create noise. We have the code for official plugins to go through and find uses.

bytebilly:
> I think it would just create noise. We have the code for official plugins to go through and find uses.

Thanks, makes sense to me

PR #87064, opened 2022-05-23
Ensure Java compile module path is tracked as a task input

mark-vieira:
This is a follow up to #81066 which ensures that we track the module path as a input on Java compilation tasks. Since we truncate the compile classpath and pass the module path as separate CLI arguments we need to track those separately. We do that by simply annotating a getter on the `CommandLineArgumentProvider`.


ChrisHegarty (approved):
LGTM.

PR #86811, opened 2022-05-16
Moving deprecation info API checks off the transport_worker thread

masseyke:
Doing a lot of work on the tranport_worker thread can cause the cluster to become unstable because nodes can't communicate quickly enough. With a lot of indices, the deprecation info API check can take tens of seconds. This change moves those checks off of the transport_worker thread.
Closes #86764

PR #87066, opened 2022-05-23
Moving deprecation info API checks off the transport_worker thread (#86811)

masseyke:
Doing a lot of work on the tranport_worker thread can cause the cluster to become unstable because nodes can't 
communicate quickly enough. With a lot of indices, the deprecation info API check can take tens of seconds. This change 
moves those checks off of the transport_worker thread.


masseyke:

PR #87068, opened 2022-05-23
Add an option to _search to force synthetic source

nik9000:
This adds `?force_synthetic_source` to, well, force running the fetch
phase with synthetic source. If the mapping is incompatible with
synthetic source it'll throw a 400 error. This should be useful for
folks looking to evaluate synthetic source.

Folks can use it to check if their current mapping supports synthetic
source. Given how picky synthetic source is, their mapping probably
*doesn't* support it now, but this gives us a very quick way to test
it. Which should help shorten the testing cycle.

If their mapping does support synthetic source it'll fetch the _source
as thought synthetic source were enabled so folks can have a look at
the result and see if it's done a good job. Synthetic source makes a
lot of assumptions about the original layout of the _source and some
of those assumptions are destructive to, well, the layout. If you
need the layout you can try to use the results this produces.

Finally this gives you an upper bound on the performance hit of synthetic
source. This feature will *still* load and decompress the original
_source bytes, so you won't get any of the fetch-time performance
advantages to synthetic source - but it throws out the _source and
rebuilds it from doc values - do you get all of the fetch time
performance disadvanteages. It the fethc performance is still
acceptable then you know synthetic source is ok for you. If it is
marginal then you'd need a full scale test, reindexing everything
with synthetic source. If the fetch performance is nowhere near ok
then you know synthetic source isn't good for you, at least not yet.


nik9000:
This actually won't "turn off" the _source - it'll still be decompressed. So this will show a "worst case" for latency - it'll include the synthetic source. It'll just load the `_source` from stored fields and throw it away.

javanna:
Heya @nik9000 can you expand on what the usecase is for this feature?

nik9000:
> Heya @nik9000 can you expand on what the usecase is for this feature?

I've edited the description to add it - the short version is that it's a way to evaluate synthetic source without reindexing. I don't think it's a "forever" thing, but I've talked to quite a few folks that would find it useful.

sethmlarson:
> I don't think it's a "forever" thing

Is this option going to be deprecated+removed in the future? If so might be good to document it this way. Also is this option for human debugging only or are we planning on users using it in applications?

nik9000:
> Is this option going to be deprecated+removed in the future?

I think so.

> Also is this option for human debugging only or are we planning on users using it in applications?

Debugging. Though maybe not just humans.

sethmlarson:
@nik9000 Gotcha! In that case can we document that this parameter is experimental and will be removed in a future version? I think that's enough warning for a parameter that's behind a feature flag and undocumented elsewhere.

server/src/main/java/org/elasticsearch/rest/action/search/RestSearchAction.java
```diff
@@ -208,6 +209,9 @@ public static void parseSearchRequest(
                 request.paramAsBoolean("ccs_minimize_roundtrips", searchRequest.isCcsMinimizeRoundtrips())
             );
         }
+        if (IndexSettings.isTimeSeriesModeEnabled() && request.paramAsBoolean("force_synthetic_source", false)) {
```
romseygeek, line 212: Just to check: `IndexSettings.isTimeSeriesModeEnabled()` is the feature flag, yes?

server/src/main/java/org/elasticsearch/search/internal/ShardSearchRequest.java
```diff
@@ -300,13 +320,15 @@ public ShardSearchRequest(ShardSearchRequest clone) {
...
         this.waitForCheckpointsTimeout = clone.waitForCheckpointsTimeout;
+        this.forceSyntheticSource = clone.forceSyntheticSource;
     }
 
     @Override
     public void writeTo(StreamOutput out) throws IOException {
         super.writeTo(out);
         innerWriteTo(out, false);
         OriginalIndices.writeOriginalIndices(originalIndices, out);
+
```
romseygeek, line 331: nit: no new line here

server/src/main/java/org/elasticsearch/search/internal/ShardSearchRequest.java
```diff
@@ -357,6 +379,13 @@ protected final void innerWriteTo(StreamOutput out, boolean asKey) throws IOExce
                     + "] or greater."
             );
         }
+        if (out.getVersion().onOrAfter(Version.V_8_4_0)) {
+            out.writeBoolean(forceSyntheticSource);
+        } else {
+            if (forceSyntheticSource) {
+                throw new IllegalArgumentException("force_synethic_source is not supported before 8.3.0");
```
romseygeek, line 386: nit: spelling `force_synethic_source`

romseygeek (commented):
Looks good in general.  I wonder if it would make more sense to make this flag state on SearchExecutionContext and then choose whether or not to build a synthetic source in SEC.getSourceLoader(), rather than pushing this down into MappingLookup though?

nik9000:
> I wonder if it would make more sense to make this flag state on SearchExecutionContext and then choose whether or not to build a synthetic source in SEC.getSourceLoader(), rather than pushing this down into MappingLookup though?

`SearchExecutionContext` can't see the `root` object mapper that's required. I don't think we want to make that public for other reasons.

romseygeek:
> SearchExecutionContext can't see the root object mapper that's required. I don't think we want to make that public for other reasons.

It can via `mappingLookup.getMapping().getRoot()`

nik9000:
> It can via `mappingLookup.getMapping().getRoot()`

"'getRoot()' is not public in 'org.elasticsearch.index.mapper.Mapping'. Cannot be accessed from outside package"

romseygeek:
Fair point...

But! SourceFieldMapper is in the same package, and we can change the signature of `SourceFieldMapper.newSourceLoader()` to take the Mapping instead of the root object; and indeed, everywhere that `newSourceLoader()` is called, it's called with a call to `getMapping().getRoot()`.  So if we rearrange things a bit then we can keep package-privacy and avoid adding that boolean to the mapping APIs.

nik9000:
> But! SourceFieldMapper is in the same package, and we can change the signature of `SourceFieldMapper.newSourceLoader()` to take the Mapping instead of the root object; and indeed, everywhere that `newSourceLoader()` is called, it's called with a call to `getMapping().getRoot()`. So if we rearrange things a bit then we can keep package-privacy and avoid adding that boolean to the mapping APIs.

I'll have a look

nik9000:
> I'll have a look

Yeah, that works. @romseygeek, have another look when  you get a chance.

romseygeek (approved):
LGTM, thanks @nik9000!

nik9000:
@sethmlarson I've added a flag to the spec to mark this parameter as behind the feature flag. Is that OK? Am I going to break a everyone?

rest-api-spec/src/main/resources/rest-api-spec/api/search.json
```diff
@@ -77,6 +77,10 @@
         "type":"number",
         "description":"Starting offset (default: 0)"
       },
+      "force_synthetic_source":{
+        "type":"boolean",
+        "description":"Should we force synthetic source for all fetches from this index?"
+      },
```
nik9000: Woops - dupes. One moment.

sethmlarson (approved):
The change to the rest-api-spec schema.json LGTM.

server/src/main/java/org/elasticsearch/search/internal/ShardSearchRequest.java
```diff
@@ -357,6 +381,13 @@ protected final void innerWriteTo(StreamOutput out, boolean asKey) throws IOExce
                     + "] or greater."
             );
         }
+        if (out.getVersion().onOrAfter(Version.V_8_4_0)) {
+            out.writeBoolean(forceSyntheticSource);
+        } else {
+            if (forceSyntheticSource) {
+                throw new IllegalArgumentException("force_synthetic_source is not supported before 8.3.0");
```
javanna, line 388: you mention 8.3 in the error message, but 8.4 in the conditional above. I suspect you did that on purpose but I am not following: is there a way to force synthetic source on 8.3 which does not support the flag?

nik9000: A mistake!

server/src/main/java/org/elasticsearch/action/search/SearchRequest.java
```diff
@@ -308,6 +321,9 @@ public void writeTo(StreamOutput out) throws IOException {
                     + "] or greater."
             );
         }
+        if (out.getVersion().onOrAfter(Version.V_8_4_0)) {
+            out.writeBoolean(forceSyntheticSource);
+        }
```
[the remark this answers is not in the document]
nik9000: Yikes. Sorry, yeah. I should have had it.

server/src/main/java/org/elasticsearch/search/internal/ShardSearchRequest.java
javanna: Unbelievable, I could not believe my eyes hence I convinced myself you must have done it on purpose :)

PR #87070, opened 2022-05-23
Make ImmutableOpenMap hash code the same as HashMap

rjernst:
ImmutableOpenMap implements Map so that we can slowly convert uses to
regular Maps. However, many serialization tests check equal and hashcode
is the same with a roundtrip. In order for that to be true, the hashcode
for ImmutableOpenMap needs to be the same as HashMap. This commit
makes ImmutableOpenMap extend AbstractMap, and then changes the each
entry hashCode to the be the same as HashMap.Entry.

relates #86239


original-brownbear (approved):
LGTM

PR #87071, opened 2022-05-24
Release bytes flushed to network

Tim-Brooks:
This commit modifies the transport serialization mechanism to expose
individual byte components. Those components can optionally have their
ownership be taken by a transport layer implementation and released
at will. This commit implements incremental release by the netty module
as bytes are no longer needed.


tbrooks8:
This is probably ready for a first look. I still want to add a few tests and check that things are working at the netty level how I expect it.

tbrooks8:
Don't review this yet as I need to go over a few things. I'll ping when it is ready.

PR #87073, opened 2022-05-24
Add log4j.core to requires for the security module

ywangd:
org.apache.logging.log4j.core is referenced in the LoggingAuditTrail
class. This PR adds it to the requires list of the security module.

Relates: #81066


ChrisHegarty (approved):
LGTM.

PR #87076, opened 2022-05-24
Fix CCR following a datastream with closed indices on the follower corrupting the datastream

### LINKED ISSUE #87048: CCR AutoFollowCoordinator can trigger put-follow actions that can corrupt datastreams in an edge case
A transport put follow-action can be triggered for an index that is part of a data stream and already exists.
Normally this is not an issue because the restore operation will just fail because the index already exists.
If however, the index is currently closed (for example due to ILM doing a force-merge with codec change step on the index),
this causes the currently closed index to be restored into. As a result we run into a bug in the following code in

`org.elasticsearch.xpack.ccr.action.TransportPutFollowAction#createFollowerIndex`:

```java
BiConsumer<ClusterState, Metadata.Builder> updater = (currentState, mdBuilder) -> {
   DataStream localDataStream = currentState.getMetadata().dataStreams().get(remoteDataStream.getName());
   Index followerIndex = mdBuilder.get(followerIndexName).getIndex();
   assert followerIndex != null;

   DataStream updatedDataStream = updateLocalDataStream(followerIndex, localDataStream, remoteDataStream);
      mdBuilder.put(updatedDataStream);
   };
   restoreService.restoreSnapshot(restoreRequest, delegatelistener, updater);
 }
```

This code updates the local data stream when following an index without checking if that index already is part of the data stream, thus now creating two entries for the same `Index` in the data stream's indices list. Code like `org.elasticsearch.cluster.metadata.DataStream#removeBackingIndex` however assumes no duplicates and will when deleting an index, only remove it from the data stream once. Thus leaving a dangling `Index` instance that does not point to any existing index any more, corrupting the cluster state.

cc @martijnvg thanks for the help with reasoning this out

original-brownbear:
Reproducer and fix for #87048.

Reproduces the edge case by closing follower index that is part of a datastream and then recreating and re-adding that same
index on the leader to make it get picked up by the auto-follower again. Using stats call in the test mainly to reproduce
the exact issue that motivated #87048 and to show that the datastream is correctly resolved by the index name expression resolver.

closes #87048


x-pack/plugin/ccr/src/main/java/org/elasticsearch/xpack/ccr/action/TransportPutFollowAction.java
```diff
@@ -328,14 +328,19 @@ static DataStream updateLocalDataStream(Index backingIndexToFollow, DataStream l
...
+                backingIndices = new ArrayList<>(localDataStream.getIndices());
+                backingIndices.add(backingIndexToFollow);
+                // When following an older backing index it should be positioned before the newer backing indices.
+                // Currently the assumption is that the newest index (highest generation) is the write index.
+                // (just appending an older backing index to the list of backing indices would break that assumption)
+                // (string sorting works because of the naming backing index naming scheme)
+                backingIndices.sort(Comparator.comparing(Index::getName));
+            } else {
+                // edge case where the index was closed on the follower and was already in the datastream's index list
+                backingIndices = localDataStream.getIndices();
```
martijnvg, line 342: So in this case, the new index being followed is not added to the data stream, but is being replicated by ccr? And the original, but closed backing index remains to exist and data stream remains to refer to the closed backing index?

original-brownbear: The "new" index is already part of the data stream, we merely reopened it due to the restore action => must not add it again.


->
 
> And the original, but closed backing index remains to exist and data stream remains to refer to the closed backing index?

no its' reopened like with a normal snapshot restore :)

martijnvg: ah, got it. Thx

martijnvg:
I think this change should be back ported to 8.2 and 7.17 branches as well?

server/src/main/java/org/elasticsearch/cluster/metadata/DataStream.java
```diff
@@ -121,7 +122,18 @@ public DataStream(
         this.system = system;
         this.allowCustomRouting = allowCustomRouting;
         this.indexMode = indexMode;
+        assert assertConsistent(this.indices);
+
```
martijnvg, line 126: unneeded newline?

x-pack/plugin/ccr/src/main/java/org/elasticsearch/xpack/ccr/action/TransportPutFollowAction.java
```diff
@@ -328,14 +328,19 @@ static DataStream updateLocalDataStream(Index backingIndexToFollow, DataStream l
...
-            List<Index> backingIndices = new ArrayList<>(localDataStream.getIndices());
-            backingIndices.add(backingIndexToFollow);
-
-            // When following an older backing index it should be positioned before the newer backing indices.
-            // Currently the assumption is that the newest index (highest generation) is the write index.
-            // (just appending an older backing index to the list of backing indices would break that assumption)
-            // (string sorting works because of the naming backing index naming scheme)
-            backingIndices.sort(Comparator.comparing(Index::getName));
+            final List<Index> backingIndices;
+            if (localDataStream.getIndices().contains(backingIndexToFollow) == false) {
```
henningandersen, line 332: I would find it more explicit to use the `currentState` above to detect that the index already existed and thus should not be added to the data stream? (and perhaps rename `currentState` to `previousState` for readability).

We should then assert that it already exists there. And here we can assert that it cannot exist already (or rely on the new assertion in the `DataStream` constructor).

x-pack/plugin/ccr/src/internalClusterTest/java/org/elasticsearch/xpack/ccr/AutoFollowIT.java
```diff
@@ -621,6 +634,89 @@ public void testAutoFollowExclusion() throws Exception {
...
+        assertLongBusy(() -> {
+            AutoFollowStats autoFollowStats = getAutoFollowStats();
+            assertThat(autoFollowStats.getNumberOfSuccessfulFollowIndices(), equalTo(3L));
+        });
+
+        assertAcked(leaderClient().admin().indices().prepareDelete(indexInDatastream).get());
+        assertAcked(followerClient().admin().indices().prepareDelete(indexInDatastream).setMasterNodeTimeout(TimeValue.MAX_VALUE).get());
+        ensureFollowerGreen("*");
+        final IndicesStatsResponse stats = followerClient().admin().indices().prepareStats(datastream).get();
+        assertThat(stats.getIndices(), not(anEmptyMap()));
```
henningandersen, line 717: Can we assert that we have the exact right set of indices here?

x-pack/plugin/ccr/src/internalClusterTest/java/org/elasticsearch/xpack/ccr/AutoFollowIT.java
```diff
@@ -621,6 +634,89 @@ public void testAutoFollowExclusion() throws Exception {
...
+        leaderClient().execute(
+            ModifyDataStreamsAction.INSTANCE,
+            new ModifyDataStreamsAction.Request(List.of(DataStreamAction.addBackingIndex(datastream, indexInDatastream)))
+        ).get();
+
+        assertLongBusy(() -> {
+            AutoFollowStats autoFollowStats = getAutoFollowStats();
+            assertThat(autoFollowStats.getNumberOfSuccessfulFollowIndices(), equalTo(3L));
+        });
+
```
henningandersen, line 712: Can we also assert that the index is now part of the data stream on the follower and is opened?

henningandersen (commented):
This looks mostly good though I'd like it to be a bit more explicit when we add to the backing indices.

x-pack/plugin/ccr/src/main/java/org/elasticsearch/xpack/ccr/action/TransportPutFollowAction.java
original-brownbear: I'd rather not use the `currentState` here to be honest. It will work now, but we go through super complicated code when restoring the datastream in the restore service, I find it much easier to reason about taking the information from the builder.
If I use the `currentState` then the whole logic becomes predicated on a bunch of assumptions (like that we won't ever remove an index in the restore action) and as a result (at least to me) much less obviously correct.

x-pack/plugin/ccr/src/internalClusterTest/java/org/elasticsearch/xpack/ccr/AutoFollowIT.java
original-brownbear: Jup added

x-pack/plugin/ccr/src/internalClusterTest/java/org/elasticsearch/xpack/ccr/AutoFollowIT.java
original-brownbear: ++ added

martijnvg (approved):
LGTM, thx for putting this up!

x-pack/plugin/ccr/src/main/java/org/elasticsearch/xpack/ccr/action/TransportPutFollowAction.java
henningandersen: I'd hope we can find a better way to signal this then, perhaps expand the updater contract to include this boolean information. I am OK with merging as is, but would like to see this in a follow-up.

The current way seems to be somewhat lenient and could "repair" something inadvertently.

henningandersen (approved):
LGTM2, thanks for fixing this.

original-brownbear:
Thanks Henning + Martijn! Will look into a follow-up to make this a little nicer as requested :)

PR #87077, opened 2022-05-24
Replace new ParametrizedMessage() with java.util.Supplier<String> 

pgomulka:
This is a result of structural search/replace in intellij. This only affects log methods with a signature
logger.info(ParametrizedMessage)
logger.info(ParametrizedMessage, Throwable)


pgomulka:
to reviewers: take a look at this commit 6aac324a1c60c62da4bfaef7d92f6aa44c0d39ef
this is where I had to manually fix compilation errors due to use of non final variable in lambda

pgomulka:
@elasticmachine update branch

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/process/logging/CppLogMessageHandler.java
```diff
@@ -348,17 +349,11 @@ private void parseMessage(XContent xContent, BytesReference bytesRef) {
         } catch (IOException e) {
             if (jobId != null) {
                 LOGGER.warn(
-                    new ParameterizedMessage(
-                        "[{}] IO failure receiving C++ log message: {}",
-                        new Object[] { jobId, bytesRef.utf8ToString() }
-                    ),
+                    () -> format("[%s] IO failure receiving C++ log message: %s", new Object[] { jobId, bytesRef.utf8ToString() }),
```
grcevski, line 352: Any reason we wrap this in Object[]?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ldap/PoolingSessionFactory.java
```diff
@@ -201,8 +200,8 @@ static LDAPConnectionPool createConnectionPool(
                     pool.setHealthCheckIntervalMillis(healthCheckInterval);
                 } else {
                     logger.warn(
-                        new ParameterizedMessage(
-                            "[{}] and [{}} have not been specified or are not valid distinguished names,"
+                        () -> format(
+                            "[%s] and [%s} have not been specified or are not valid distinguished names,"
```
grcevski, line 204: I think the closing `}` in the second parameter is a typo. We should probably fix it.

grcevski (approved):
LGTM! 

I left few comments and questions. I think we should definitely try/catch/assert to prevent logger messages from creating new failures while debugging, but change in format can come in a follow-up PR.

To test better the lambda logging I think we need to introduce a CI job with log=TRACE to cover all log messages. It doesn't have to run on every PR, but perhaps something that runs on merge?

pgomulka:
@elasticmachine update branch

PR #87079, opened 2022-05-24
Return a default user action if no actions could be determined

jbaiera:
If a shard is unassigned for a reason other than `NO_VALID_SHARD_COPY` or the deciders returning a `NO` response, then no user actions are returned. If the deciders return no, but we cannot determine a root cause in the indicator we return a default user action. If we could not generate any user actions for a shard being unassigned, we should always add one.


andreidan (approved):
LGTM thanks for fixing this

PR #87080, opened 2022-05-24
Add support for geo_point in top_metrics aggregation

iverase:
This PR adds the possibility of adding fields of type geo_point as a metric in the top_metrics aggregation. /It currently just output the first point in case of an array of points with the format "lat, lon".


nik9000 (commented):
I think that's how to do it, yeah.

I do think our lack of support for fetching all of the values of a field is a problem - there's an issue for it that came up many months after we wrote the feature. And I imagine it's a bigger problem with geo because arrays of points is not uncommon.

iverase:
> And I imagine it's a bigger problem with geo because arrays of points is not uncommon.

I don't think so as when I see points as metrics I imagine more a device reporting its position so in general point metrics should be single valued. 

At the moment the biggest issue (and it is not specific to this aggregation) is the format of geo_point doc values. They come as a string"lat, lon" which is totally different to the formats supported for the fields API.

nik9000:
> At the moment the biggest issue (and it is not specific to this aggregation) is the format of geo_point doc values. They come as a string"lat, lon" which is totally different to the formats supported for the fields API.

Ah! When I implements `terms` for `_tsid` I had it come out as a map which felt natural but required a bit of special casing, sadly. I think it's worth a more general clean up.

PR #87085, opened 2022-05-24
[7.17] Allow the modify data stream api to remove broken reference to a backing index (#87030)

dakrone:
 - Allow the modify data stream api to remove broken reference to a backing index (#87030)


martijnvg:
In master and 8.2 the test tear down is different than in 7.17 branch. This requires adding a cleanup method to remove data streams and templates after each test to `DataStreamSecurityIT`. 

The `DataStreamSecurityIT#testRemoveGhostReference()` failed, because data streams were not removed and the regular tear down tried to remove all indices. The write index of a data stream can't be removed. The added teardown logic use the delete data stream api to delete the data stream, so that the regular teardown does't encounter backing indices of data streams.

PR #87086, opened 2022-05-24
Fix expand_wildcards default in docs

rjernst:
The get settings api has accepts the expand_wildcards option. The docs
state the default value is `all`, but it is actually now `open` (which
does not include hidden or closed indices by default). This commit
changes the docs to match the existing behavior.


williamrandolph (approved):
LGTM! If we backport this to other 8.x branches, does it show up in docs for those releases?

PR #87087, opened 2022-05-24
Fix DataStreamSecurityIT#testRemoveGhostReference() test

martijnvg:
If the modify data stream api can find the IndexMetadata of
an index it tries to remove from a data stream then it can
unhide the index as well, otherwise it can't.

This determines whether the expected index stats is either 1 or 2
in the last get indices stats call.


dakrone (approved):
LGTM

PR #87093, opened 2022-05-24
[Refactor] Remove nearly-redundant method from SystemIndices

williamrandolph:
The getProductSystemIndexMetadataPredicate was only used once, and could be replaced by the very similar method getProductSystemIndexNamePredicate. This not only simplifies the public API, but removes the cognitive load of understanding how the two methods differed.


stu-elastic (approved):
`getProductSystemIndexMetadataPredicate` just grabbed the name anyway.  Makes sense.

williamrandolph:
@elasticmachine update branch

It looks like previous failures had to do with release process turbulence.

PR #87096, opened 2022-05-25
Read official plugins list non-statically

rjernst:
New plugin telemetry keeps track of whether a plugin is in the list of
official plugins. This requires reading the plugins list resource file.
Reading that statically depends on the jar being built, but that is not
available during tests. It's also not necessary to keep this list in
memory for the lifetime of the node. This commit moves reading the file
to occur only when constructing the plugin runtime info.


ywangd (approved):
LGTM

PR #87097, opened 2022-05-25
User Profile - audit support for security domain

ywangd:
Security domain allows sharing resources across different realms which
means it is now part of the authorization process. Hence we should
capture it in relevant audit entries for completeness.

In theory, the domain name should be captured in all the places realm
name is captured. However, it is not necessary in practice because:

1. The field `realm` in authentication_success event is for BWC purpose.
   Hence we don't need pair it with domain.
2. API Keys cannot be under any domain.
3. No need to audit domain for authentication_failed events because no
   authorization is possible in these cases.

### DIFF
--- a/docs/changelog/87097.yaml
+++ b/docs/changelog/87097.yaml
@@ -1,0 +1,5 @@
+pr: 87097
+summary: User Profile - audit support for security domain
+area: Audit
+type: enhancement
+issues: []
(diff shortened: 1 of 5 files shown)


albertzaharovits (approved):
LGTM

ywangd:
> It is cumbersome, and I'm not sure we should, but new audit fields are usually documented, https://www.elastic.co/guide/en/elasticsearch/reference/current/audit-event-types.html#_extra_audit_event_attributes_for_specific_events .

Sorry I meant to comment on this but forgot. I intentionally left it out because there are a few nuisances which made me think a separate PR is better:
* Currently we still have the feature flag. So adding contents into existing page leads to many littering of `ifeval ...` conditions. 
* The feature flag should be removed soon since we agreed to it is no longer needed in 8.4. So adding `ifeval ...` as described above is rather redundant.
* But I haven't removed the feature flag because there are some doc changes that I still want to be part of 8.3.
* Even with feature flag removed, the feature is still in beta. I am not sure what is the best to notify that on an existing page. 
* This is for 8.4 so we still have time to work with doc experts to solve it.

PR #87098, opened 2022-05-25
Handle unexpected exit code from server process

rjernst:
When running Elasticsearch in the foreground, the cli process waits
indefinitely on the server. If the server dies unexpectedly, the
ServerProcess throws an exception. However, the exit code is hidden in
the exception message. This commit changes waitFor to return the exit
code, so it can be propagated to the cli main. Note that when stopping
in a shutdown hook the exit code must be ignored because calling
System.exit from a shutdownhook results in a deadlock.


grcevski (approved):
LGTM!

PR #87100, opened 2022-05-25
Give doc-value-only mappings to numeric fields on metrics templates.

jpountz:
It doesn't make much sense to filter samples depending on the values of some
metric fields, as these metrics mostly make sense in aggregate and over time.
This change proposes to give doc-value-only mappings to numeric fields in the
default metrics template. This helps reduce index-time CPU usage and disk usage
while not impacting the search-time experience.

Another improvement that this change brings is mapping floating-point numbers
as doubles instead of floats by default, since double precision is more
commonly used across other systems that are designed at handling metrics. I
could revert that part and switch back to floats if it proved controversial.

These doc-value-only dynamic templates might catch fields that are not metrics
at times, which is probably ok.


dakrone (approved):
LGTM from the ES side, but since these mappings came from the observability side I'd prefer someone from that team sign off before merging.

nik9000 (approved):
It's fine by me too.

ruflin:
Change LGTM. On the double change, will this have any storage or performance impact for numbers with only 1 digit for example?

From an Observability perspective, by default this will not have an effect on all the data we load packages with templates for as these have higher priority. It will mainly affect the data without predefined mappings. If the above change is unexpected for a user, I assume our recommendation is to set a specific template in place for these data streams with higher priority to revert back to the old behaviour.

@axw @felixbarny I assume this will not really have an affect on APM?

@joshdover This change of defaults should also be applied to Fleet and the default mappings we load for each data stream. We should have again a discussion with @jpountz around reusing or not the "base" templates from Elasticsearch for these.

felixbarny:
+1 to index: false and rely only on doc values. This is how I used to [map metrics indices](https://www.elastic.co/blog/elasticsearch-as-a-time-series-data-store) in the past.

Last time I checked, which was quite a while ago, doubles took up more space than floats but if the industry standard is to use doubles that that's what we should do, too. But maybe compression has improved in the meantime in ES?

> a discussion with @jpountz around reusing or not the "base" templates from Elasticsearch for these.

Is there any capability to "inherit" from the base template or will a higher priority template always override and not merge with other templates?

ruflin:
> Is there any capability to "inherit" from the base template or will a higher priority template always override and not merge with other templates?

Yes. If we use component templates, I think the last (or first) component wins.

axw:
> @axw @felixbarny I assume this will not really have an affect on APM?

APM has its own dynamic templates. We should probably follow suit though: https://github.com/elastic/apm-server/issues/8261

ruflin (approved):
Approving this change. The assumption is that at some point we likely will optimise double vs float usage with TSDB and on the reuse (or not) of component templates, we will have a separate discussion.

ruflin:
@joshdover These are the ones I was referring to and we should likely add some numeric values. But lets split the discussion from this PR as there is more to it. I'll ping you.

ruflin (approved):

jpountz:

PR #87108, opened 2022-05-25
Faster mapper parsing when merging multiple templates

original-brownbear:
This speeds up the many-shards benchmark significantly.
When merging multiple templates (as is often the case for creating
indices as part of a datastream) we don't need to create a
full document parser instance for each template individually.
The only change required to make this work is that we need to pass
the `metadataMappers` map to the mapping parse function explicitly
instead of implicitly taking it from `this.mapper` via the parse context.
(IMO this is easier to reason about anyway but maybe I'm missing something here)

This saves the expensive serialization of the intermediary merged
mappings and as a result makes the mapping parsing+merging about 15-25%
faster end-to-end in the many-shards benchmark (which is a huge win
considering mapping parsing accounts for more than a third of the
time it takes to bootstrap the larger cluster at the start of these
benchmarks).

relates #77466


ywelsch (commented):
MapperService is a cyclic mess of dependencies, making it near impossible to understand whether these changes are safe or not. It's impossible to know where `this.mapper` is being accessed in the update methods. It reminds me a bit of `ClusterApplierService` and its `state()` method (which should not be called by `ClusterStateAppliers`).

Is there a way to refactor these classes to statically ensure (or via runtime checks) that the mutation methods are not accessing forbidden state?

PR #87109, opened 2022-05-25
Reword #87076 changelog entry

DaveCTurner:
The phrase "corrupting the datastream" sounds like data loss (or worse)
but in fact it's just bad metadata and the situation is recoverable.


original-brownbear (approved):
LGTM, thanks David!

PR #87062, opened 2022-05-23
Fix incorrect term in AD/LDAP docs

hulkk:
The dc container in this context is not domain controller, it is domain component. Schema can be checked from e.g. RFC 4519.

PR #87114, opened 2022-05-25
[8.0] Fix incorrect term in AD/LDAP docs (#87062)

ywangd:
Backports the following commits to 8.0:
 - Fix incorrect term in AD/LDAP docs (#87062)


lockewritesdocs:
@elasticmachine update branch

Issue #72983, opened 2021-05-12
top_metrics and bucket_script params should automagically work


philippkahr:
**Elasticsearch version** (`bin/elasticsearch --version`): 7.12

**OS version** (`uname -a` if on a Unix-like system): Running Elastic Cloud.

**Description of the problem including expected versus actual behavior**:

Combining `bucket_script` with `top_metrics` should just work.

In the tutorials and/or docs on elastic.co the `bucket_script params` are often referred to child / grandparents syntax. However that does not work.

```
current-price>top
current-price>top>metrics
current-price>top>metrics>price
```

or (change everything to dot-walking)

```
current-price.top.metrics.price
```

I get either that error

```
"reason" : "buckets_path must reference either a number value or a single value numeric metric aggregation, but [current-price] contains multiple values. Please specify which to use."
```

or

```
"reason" : "Validation Failed: 1: No aggregation [top] found for path [current-price>top];"
```

or (when dot walking)

```
"reason" : "path not supported for [current-price]: [top]"
```

**How should it work**:

1. The `top_metrics` looks like this:

```json
"current-price" : {
    "top" : [
      {
        "sort" : [
          "+51361822-08-28T19:06:40.000Z"
        ],
        "metrics" : {
          "price" : 4.94372760037313
        }
      }
    ]
 }
```

2. Elasticsearch should *magically* detect that the *current-price* top only contains *one* valid value.
3. I should be able to write `buckets_path: { "currentprice":"current-price"}` and *not* `current-price.price`.

**Working version**:

<details>
  <summary>Query</summary>

```json
{
  "query": {
    "range": {
      "@timestamp": {
        "gte": "now-1y/d",
        "lte": "now/d"
      }
    }
  },
  "size": 0,
  "aggs": {
    "terms_agg": {
      "terms": {
        "field": "pair"
    },
    "aggs":{
      "sum-of-expenses":{
        "sum": {
          "script": """
            if (doc['type'].contains('buy')) {
              return doc['cost']
            }
          """
          }
        },
        "amount-of-coin":{
          "sum": {
            "script": """
              if (doc['type'].contains('buy')) {
                return doc['vol']
              }
            """
          }
        },
        "current-price":{
         "top_metrics":{
            "metrics": { "field": "price" },
            "sort": { "@timestamp": "desc" },
            "size": 1
            }
        },
        "value-of-depot": {
          "bucket_script": {
            "buckets_path": {
              "coins": "amount-of-coin",
[...]

PR #87117, opened 2022-05-25
Make top metrics field name optional when used in bucket script

salvatore-campagna:
Before this change using top metrics inside a bucket script required
the full metric field path to be specified in the params section.
With this change the metric field name becomes optional. If top
metrics includes more than one metric the field is still required.
If, instead, only one metric is available that single metric is used
just specifying the top metrics name (and no metric field name
is required).

The old notation is still supported for backward compatibility.

This is a fix for #72983.


server/src/main/java/org/elasticsearch/search/aggregations/pipeline/BucketHelpers.java
```diff
@@ -207,6 +208,9 @@ public static Double resolveBucketValue(
                     value = ((Number) propertyValue).doubleValue();
                 } else if (propertyValue instanceof InternalNumericMetricsAggregation.SingleValue) {
                     value = ((InternalNumericMetricsAggregation.SingleValue) propertyValue).value();
+                } else if (propertyValue instanceof NumericMetricsAggregation.MultiValue) {
+                    final String metricName = aggPathAsList.get(aggPathAsList.size() - 1);
+                    value = ((NumericMetricsAggregation.MultiValue) propertyValue).value(metricName);
```
salvatore-campagna, line 213: Looks like this breaks all other aggregations implementing `NumericMetricsAggregation.MultiValue`...

salvatore-campagna: I think I need to implement `SingleValue` not `MultiValue`.

salvatore-campagna:
Closing since I had to rebase the branch because of an issue with the current version being changed. Will do another PR.

PR #87121, opened 2022-05-25
Rest API Spec: Wraps YAML tests values with colon in quotes

picandocodigo:
The Ruby client's YAML parser parses these values as Dates if they're not wrapped in quoted to be parsed as Strings.


grcevski (commented):
I see that the problem is present in the 'main' branch too. I think it would be best if we created a PR for master and then backport all the way to 7.17.

grcevski:
I think we can safely close this PR, a backport PR will be created from the 'master' branch PR, tracking the original PR number.

PR #79922, opened 2021-10-27
Reinterpret dots in field names as object structure

romseygeek:
DocumentParser parses documents by following their object hierarchy, and
using a parallel hierarchy of ObjectMappers to work out how to map leaf fields.
Field names that contain dots complicate this, meaning that many methods 
need to reverse-engineer the object hierarchy to check that the current parent
object mapper is the correct one; this is particularly complex when objects
are being created dynamically.

To simplify this logic, this commit introduces a DotExpandingXContentParser, 
which will wrap another XContentParser and re-interpret a field name containing
dots as a series of objects.  So for example, `"foo.bar.baz":{ ... }` is passed
to the DocumentParser as `"foo":{"bar":{"baz":{...}}}`. The central parsing
logic of DocumentParser detects when a fieldname has this form and temporarily
swaps out its context for a context with a wrapped parser.

PR #87123, opened 2022-05-25
Don't run include_in_parent when in copy_to context

### LINKED ISSUE #87036: bad indexing performance in elasticsearch 8.2.0
### Elasticsearch Version

8.2.0

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

docker-image = docker.elastic.co/elasticsearch/elasticsearch:8.2.0

### Problem Description

We are trying to migrate from Elasticsearch 6.7.1 to Elasticsearch 8.2.0.

With the same amount of documents per bulk request (100 docs per request), we always get a socket timeout in Elasticsearch 8.2.0. Everything works fine in Elasticsearch 6.7.1.

I noticed that indexing the same document in Elasticsearch 8.2.0 takes 44152ms. In Elasticsearch 6.7.1, indexing takes 50ms.

### Steps to Reproduce

[elasticsearch_8.2.0.txt](https://github.com/elastic/elasticsearch/files/8754299/elasticsearch_8.2.0.txt)
[elasticsearch_6.7.1.txt](https://github.com/elastic/elasticsearch/files/8754300/elasticsearch_6.7.1.txt)

### Logs (if relevant)

_No response_

romseygeek:
We changed how copy_to is implemented in #79922, which moved
the handling of dots in field names into a specialised parser.  Unfortunately,
while doing this we added a bug whereby every time a copy_to directive
is processed for a nested field, the nested field's include_in_parent logic
would be run, meaning that the parent would end up with multiple copies
of the nested child's fields.

This commit fixes this by only running include_in_parent when the parser
is not in a copy_to context.  It also fixes another bug that meant the parent
document would contain multiple copies of the ID field.

Fixes #87036


server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
```diff
@@ -363,7 +370,7 @@ private static void nested(DocumentParserContext context, NestedObjectMapper nes
     private static void addFields(Version indexCreatedVersion, LuceneDocument nestedDoc, LuceneDocument rootDoc) {
         String nestedPathFieldName = NestedPathFieldMapper.name(indexCreatedVersion);
         for (IndexableField field : nestedDoc.getFields()) {
-            if (field.name().equals(nestedPathFieldName) == false) {
+            if (field.name().equals(nestedPathFieldName) == false && field.name().equals(IdFieldMapper.NAME) == false) {
```
ywelsch, line 373: Should we move the logic for adding _id fields to nested docs to the `IdFieldMapper.postParse` method, similar as we have it for other metadata fields? Then we wouldn't need this exception here?

server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
```diff
@@ -345,6 +345,13 @@ private static void throwEOF(ObjectMapper mapper, String currentFieldName) {
     }
 
     private static void nested(DocumentParserContext context, NestedObjectMapper nested) {
```
ywelsch, line 347: This method name does not suggest at all that this method is about adding fields to parent/root doc. Should this be renamed to make the purpose clearer?

ywelsch (commented):
Thanks for looking into this, @romseygeek! I've left one question, looking good o.w.

server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
romseygeek: That sounds like a good plan, but I think it might end up being a fairly big change as there are a few places that refer directly to the id field scattered around the place.  It also interacts with TSDB in a slightly complicated way - are you OK with doing this as a follow-up?

server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
romseygeek: ++ have renamed to `copyNestedFields`

romseygeek:
Thanks for the review @ywelsch, this is ready for another go-around.

server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
ywelsch: sure, follow-up is fine.

ywelsch (approved):
LGTM. Thank you for digging into this bug, and the fix!

romseygeek:
@elasticmachine update branch

PR #87124, opened 2022-05-25
Hide ImmutableOpenMap in Metadata public methods

rjernst:
This commit converts the remaining methods in Metadata that return an
ImmutableOpenMap to return Map, thus hiding the fact that Metadata uses
ImmutableOpenMap internally.

relates #86239


grcevski (approved):
LGTM!

Issue #85697, opened 2022-04-05
[CI] TokenAuthIntegTests testRefreshingMultipleTimesWithinWindowSucceeds failing


albertzaharovits:
Failed only 7 times in the last 28 days, and started failing in the last week or so:
https://gradle-enterprise.elastic.co/scans/tests?search.startTimeMax=1649167149683&search.startTimeMin=1646085600000&search.timeZoneId=Europe/Bucharest&tests.container=org.elasticsearch.xpack.security.authc.TokenAuthIntegTests&tests.sortField=FAILED&tests.test=testRefreshingMultipleTimesWithinWindowSucceeds&tests.unstableOnly=true

And sometimes the assertion is:
```
java.lang.AssertionError: 	
Expected: <1>	
     but: was <3>
```

**Build scan:**
https://gradle-enterprise.elastic.co/s/oit7vnoh6abfs/tests/:x-pack:plugin:security:internalClusterTest/org.elasticsearch.xpack.security.authc.TokenAuthIntegTests/testRefreshingMultipleTimesWithinWindowSucceeds

**Reproduction line:**
`./gradlew ':x-pack:plugin:security:internalClusterTest' --tests "org.elasticsearch.xpack.security.authc.TokenAuthIntegTests.testRefreshingMultipleTimesWithinWindowSucceeds" -Dtests.seed=F7EB4530E2A3CA68 -Dbuild.snapshot=false -Dtests.jvm.argline="-Dbuild.snapshot=false" -Dtests.locale=es-PA -Dtests.timezone=America/Rankin_Inlet -Druntime.java=17`

**Applicable branches:**
master, 8.2

**Reproduces locally?:**
No

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.xpack.security.authc.TokenAuthIntegTests&tests.test=testRefreshingMultipleTimesWithinWindowSucceeds

**Failure excerpt:**
```
java.lang.AssertionError: 
Expected: <1>
     but: was <2>

  at __randomizedtesting.SeedInfo.seed([F7EB4530E2A3CA68:D6E525F2AF664A1C]:0)
  at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:18)
  at org.junit.Assert.assertThat(Assert.java:956)
  at org.junit.Assert.assertThat(Assert.java:923)
[...]

PR #87125, opened 2022-05-25
Hide system indices and their aliases in upgraded clusters

williamrandolph:
We have code that is supposed to make sure existing system indices are made into hidden indices if they are visible, and to hide aliases for system indices. However, that code was in a ClusterStateUpdateTask, and that task was only submitted in the case when we needed to toggle an index's "system" setting from true to false or false to true.

Here, we update the checks so that the task is submitted in the cases where an index is correctly marked as a system index, but is not itself hidden or has aliases that are not hidden. We also add test coverage for the method that checks whether system index metadata should be updated.

This should fix problems where system indices created in 7.x are not hidden in 8.x.


server/src/main/java/org/elasticsearch/cluster/metadata/SystemIndexMetadataUpgradeService.java
```diff
@@ -72,6 +74,35 @@ public void clusterChanged(ClusterChangedEvent event) {
...
+        }
+
+        if (shouldBeSystem) {
+            return shouldBeHidden(indexMetadata) || shouldHideAliases(indexMetadata);
+        }
+
+        return false;
+    }
+
+    private boolean shouldBeHidden(IndexMetadata indexMetadata) {
```
grcevski, line 93: Since these boolean logic methods are now nicely extracted, I think it would be really good if we added unit tests for the true false conditions. `shouldBeHidden` is perhaps trivial, but the others have certain level of control flow in them.

grcevski (commented):
This looks great to me, I just have minor comment about adding some additional unit tests for the extracted boolean logic.

server/src/main/java/org/elasticsearch/cluster/metadata/SystemIndexMetadataUpgradeService.java
williamrandolph: Good idea! I added some tests that check the methods and whether they're interacting with requiresUpdate as we expect.

williamrandolph:

The test failure was unrelated.

williamrandolph:
@elasticmachine please run elasticsearch-ci/2

different unrelated test failure this time: https://github.com/elastic/elasticsearch/issues/85697

williamrandolph:

different unrelated test failure this time: https://github.com/elastic/elasticsearch/issues/85697

grcevski (approved):
LGTM!

PR #87128, opened 2022-05-25
Remove leftover debugging statement on error

rjernst:
A debugging statement was left being printed when ES exits with a
non-zero status. This commit removes the debug statement.

relates #85758


grcevski (approved):
LGTM!

PR #87130, opened 2022-05-25
Updated elasticsearch version to 8.4.0

craigtaverner:
The branch for 8.3.0 has been made, so we move master onto 8.4.0.


.ci/bwcVersions
```diff
@@ -1,49 +1,4 @@
 BWC_VERSION:
-  - "7.0.0"
```
rjernst, line 2: I don't think all of these should be removed?

rjernst (commented):
Looks good, but I dont' think the bwc versions for 7.x should be removed...

.ci/bwcVersions
craigtaverner: It was the gradle updateCIBwcVersions command that removed them. I do not know how it makes those decisions, but was curious about this.

breskeby (approved):

PR #87132, opened 2022-05-25
Catch an exception due to incorrect pattern in Strings.format

pgomulka:
Strings.format method, which is used heavily in logging with
Supplier<String> should handle exceptions when a format is incorrect.
This will prevent a hard to catch mistakes to blow up in server.
Those mistakes are especially hard to detect in logging when a
code to create a message might be only executed when logger is debug
or trace. Which is not always the case in CI.

relates https://github.com/elastic/elasticsearch/pull/87077#discussion_r881794589

relates #86549

### DIFF
--- a/libs/core/src/main/java/org/elasticsearch/core/Strings.java
+++ b/libs/core/src/main/java/org/elasticsearch/core/Strings.java
@@ -21,8 +21,14 @@ public class Strings {
      *
      * This method calls {@link String#format(Locale, String, Object...)}
      * with Locale.ROOT
+     * If format is incorrect the function will return format without populating
+     * its variable placeholders.
      */
     public static String format(String format, Object... args) {
-        return String.format(Locale.ROOT, format, args);
+        try {
+            return String.format(Locale.ROOT, format, args);
+        } catch (Throwable e) {
+            return format;
+        }
     }
 }
--- a/libs/core/src/main/java/org/elasticsearch/core/Strings.java
+++ b/libs/core/src/main/java/org/elasticsearch/core/Strings.java
@@ -18,7 +18,7 @@ public class Strings {
     /**
      * Returns a formatted string using the specified format string and
      * arguments.
-     *
+     * <p>
      * This method calls {@link String#format(Locale, String, Object...)}
      * with Locale.ROOT
      * If format is incorrect the function will return format without populating
@@ -27,7 +27,8 @@ public class Strings {
     public static String format(String format, Object... args) {
         try {
             return String.format(Locale.ROOT, format, args);
-        } catch (Throwable e) {
+        } catch (Exception e) {
+            assert false : "Exception thrown when formatting [" + format + "]. " + e.getClass().getCanonicalName() + ". " + e.getMessage();
             return format;
         }
     }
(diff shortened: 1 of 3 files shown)


libs/core/src/main/java/org/elasticsearch/core/Strings.java
```diff
@@ -21,8 +21,14 @@ public class Strings {
...
      * This method calls {@link String#format(Locale, String, Object...)}
      * with Locale.ROOT
+     * If format is incorrect the function will return format without populating
+     * its variable placeholders.
      */
     public static String format(String format, Object... args) {
-        return String.format(Locale.ROOT, format, args);
+        try {
+            return String.format(Locale.ROOT, format, args);
+        } catch (Throwable e) {
```
rjernst, line 30: Why throwable? This should probably be just Exception. Let OOM/etc propagate.

libs/core/src/main/java/org/elasticsearch/core/Strings.java
```diff
@@ -21,8 +21,14 @@ public class Strings {
...
+            return format;
```
rjernst, line 31: Can we not fail silently? We should somehow capture the actual problem, embedding it in the message. Also, add an assert here so that if we hit this in tests we can fail the test.

rjernst (commented):
A couple suggestions

libs/core/src/main/java/org/elasticsearch/core/Strings.java
pgomulka: Exception is definitely better

libs/core/src/main/java/org/elasticsearch/core/Strings.java
pgomulka: how should we capture it? Maybe we should log it (with additional Strings's logger)

rjernst: Yes, logging sounds good.

rjernst: Actually, not sure we can do logging, because this is in the core lib? It doesn't have any deps. Perhaps we could embed it in the message?

pgomulka: I am not sure.. This would be a good solution if we only use this in logging (we do, but the method is in core with no indication that this is only for logging).
I wonder if just returning the original pattern is not good enough. This is what log4j is doing when a pattern has more place holders. 
I wonder if we could rather just try to use some static analysis to detect this
https://rules.sonarsource.com/java/RSPEC-2275 
this looks promising but would have to be adopted to our needs

grcevski: I'm OK with leaving it as is if it matches what log4j does.

grcevski (approved):
LGTM!

libs/core/src/main/java/org/elasticsearch/core/Strings.java
pgomulka: I raised an issue to look for static analysis solution this https://github.com/elastic/elasticsearch/issues/87166

rjernst (approved):
LGTM, we can iterate in the future.

PR #87133, opened 2022-05-25
Use all lucene index compatible versions for bwc version generation

rjernst:
The updateCIBwcVersions task regenerates the list of versions that are
tested in CI. This task internally calls a filter method which filters
out incompatible versions for the local system. That means it can change
depending on the type of system the task is run on. This commit adds a
new variant of the internal method to return all actual index compatible
versions, and renames the existing method to make it clear it is for
running actual tests on the local system.


build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/BwcVersions.java
```diff
@@ -229,18 +229,23 @@ private List<Version> getReleased() {
         return versions.stream().map(v -> v.elasticsearch).filter(v -> unreleased.containsKey(v) == false).toList();
     }
 
-    public List<Version> getIndexCompatible() {
-        return filterSupportedVersions(
-            versions.stream().filter(v -> v.lucene.getMajor() >= (currentVersion.lucene.getMajor() - 1)).map(v -> v.elasticsearch).toList()
-        );
+    public List<Version> getLocalIndexCompatible() {
```
mark-vieira, line 232: I think I'd prefer us keep `getIndexCompatible` with the same behavior and just _add_ the method to get "unfiltered" results given the only scenario that is used is for this specific task. This avoid the awkwardness of having to use `localIndexCompatible` all over.

mark-vieira (changes requested):

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/BwcVersions.java
rjernst: > so we're adding verbosity and confusion to the common case

I see it as adding clarity to the common case. I guess I'm more used to developing in IDEs where I would see both methods and not know which to choose. Having the two clear names is better IMO than having one that is ambiguous (why would getIndexCompatible not already be "all" of the versions, since there is no qualification?).

mark-vieira: If we are going to have mutliple then we should also add JavaDocs here to explain the difference. It's not clear what "local" means here. I still prefer the default method be the common case and we add a method for "unfiltered". We mostly use these in build scripts and it's just confusing to have `wireCompatible` next to `localIndexCompatible`. Why the difference?

rjernst: Alright, I've reverted the naming, and also added javadocs to these two methods.

mark-vieira (approved):
Nope. LGTM.

PR #87134, opened 2022-05-25
Make pidfile deletion more resilient

rjernst:
Deleting the pidfile when Elasticsearch is shutting down was moved in #86934.
However, the delete SM permission is still needed, since SM is installed
by the time we shutdown. This commit adds back to the pidfile delete
permission, and also rethrows any io exception so it can be logged by
our uncaught exception handler.


rjernst:
Note that it was painful to pass the pidfile through, but I don't think we should add it back to Environment, it's not necessary there. This can get better in future refactorings to collapse all the different init methods.

grcevski (approved):
LGTM!

PR #87145, opened 2022-05-25
Use 'monitor' instead of 'read' permission for resolve index test

dakrone:
This commit changes no production code, but uses the `monitor` instead of `read` permission for the negative index resolve test (testing that a request is rejected). This is required because we have no way to handle the transform for a test like this in order to merge #87052


jakelandis (approved):
LGTM

PR #87147, opened 2022-05-26
Make KeywordFieldMapper more compact

original-brownbear:
Removing some redundant fields that we store both on the mapper and the field type 
to make these objects smaller.
This change saves almost 200M on heap per data node in the many-shards
benchmark by saving an estimated 14b-16b (didn't do the alignment math)
per instance.

relates #77466


romseygeek (approved):
LGTM

original-brownbear:
Thanks Alan!

PR #87148, opened 2022-05-26
Rest API Spec: Wraps YAML tests values with colon in quotes

picandocodigo:
The Ruby client's YAML parser parses these values as Dates if they're not wrapped in quoted to be parsed as Strings.

Related: #87121

### DIFF
--- a/rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.aggregation/500_serial_diff.yml
+++ b/rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.aggregation/500_serial_diff.yml
@@ -25,9 +25,9 @@ basic:
                 d: {serial_diff: {buckets_path: v}}
   - match: { hits.total.value: 3 }
   - length: { aggregations.@timestamp.buckets: 3 }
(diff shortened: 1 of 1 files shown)


grcevski (approved):
LGTM!

I added the labels to backport this fix to earlier versions, going all the way to 7.17. With the auto-backport-and-merge label, backport PRs will automatically be created for you and merged if there are no conflicts and CI success. In case of conflicts, you'll need to manually update the PRs to make the CI green, but no separate merge approval is required.

If your intent was not to fix this in 7.17, just remove the label please before merging.

picandocodigo:
@grcevski that's great, I wondered how the backporting was handled, but all the way to 7.17 is exactly what I need :+1: 
I'll keep an eye out in case any of the automatic PR's fail and update them if needed. Thanks!

grcevski:
It looks like 8.2 would need to be done manually. Simply follow the format in the PR of what was done for 8.3 and it should be fine.

picandocodigo:
@grcevski it looks like this test is not present in 8.2, so the backport is not necessary.

grcevski:
All good then! Thanks for checking.

PR #87149, opened 2022-05-26
Temporarily provide SystemPropertiesPropertySource

ChrisHegarty:
Temporarily provide SystemPropertiesPropertySource to workaround a bug in Log4J where it does not declare the provider implementations in its module-info.

This has been resolved in log4J 2.18, as part of https://issues.apache.org/jira/browse/LOG4J2-3427

### DIFF
--- a/docs/changelog/87149.yaml
+++ b/docs/changelog/87149.yaml
@@ -1,0 +1,5 @@
+pr: 87149
+summary: Temporarily provide `SystemPropertiesPropertySource`
+area: Infra/Logging
+type: bug
+issues: []
--- a/server/src/main/java/module-info.java
+++ b/server/src/main/java/module-info.java
@@ -353,4 +353,5 @@ module org.elasticsearch.server {
         with
             org.elasticsearch.cluster.coordination.NodeToolCliProvider,
             org.elasticsearch.index.shard.ShardToolCliProvider;
+    provides org.apache.logging.log4j.util.PropertySource with org.elasticsearch.common.logging.ESSystemPropertiesPropertySource;
 }
--- a/server/src/main/java/org/elasticsearch/common/logging/ESSystemPropertiesPropertySource.java
+++ b/server/src/main/java/org/elasticsearch/common/logging/ESSystemPropertiesPropertySource.java
@@ -1,0 +1,17 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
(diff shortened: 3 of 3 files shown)


ChrisHegarty:
Without this change the `log4j2.disable.jmx` system property cannot be successfully read when running with the security manager, resulting in the below exception when the server is shutting down (since the log4J implementation thinks that jmx is enabled).

Note: by facilitating the location the SystemPropertiesPropertySource provider, the log4J implementation, when starting up, will read and cache the system properties before the restrictive ES security manager is installed.

```
Unable to unregister MBeans java.security.AccessControlException: access denied ("javax.management.MBeanServerPermission" "createMBeanServer")
  at java.base/java.security.AccessControlContext.checkPermission(AccessControlContext.java:485)
  at java.base/java.security.AccessController.checkPermission(AccessController.java:1068)
  at java.base/java.lang.SecurityManager.checkPermission(SecurityManager.java:411)
  at java.management/java.lang.management.ManagementFactory.getPlatformMBeanServer(ManagementFactory.java:481)
  at org.apache.logging.log4j.core@8.4.0-SNAPSHOT/org.apache.logging.log4j.core.jmx.Server.unregisterLoggerContext(Server.java:248)
  at org.apache.logging.log4j.core@8.4.0-SNAPSHOT/org.apache.logging.log4j.core.LoggerContext.stop(LoggerContext.java:379)
  at org.apache.logging.log4j.core@8.4.0-SNAPSHOT/org.apache.logging.log4j.core.AbstractLifeCycle.stop(AbstractLifeCycle.java:136)
  at org.apache.logging.log4j.core@8.4.0-SNAPSHOT/org.apache.logging.log4j.core.config.Configurator.shutdown(Configurator.java:417)
  at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.bootstrap.Bootstrap$4.run(Bootstrap.java:196)
```

ChrisHegarty:
To further compound the mystery around the root cause of this issue - why does log4J not have permission to read the `log4j2.disable.jmx` system property, when the ES policy grants all code permission to read all properties? Well... the log4J code that performs the "fallback" property is as follows [1]:

```
private static boolean hasSystemProperty(final String key) {
  try {
    return System.getProperties().containsKey(key);
  } catch (final SecurityException ignored) {
    return false;
  }
}
```

`System.getProperties()` requires both _read_ and _write_ property permission, for all properties. While `System.getProperty(..)` requires only  _read_ for the given permission. It would seem that the latter is a more security manager "friendly" way to determine if the property is set or not.

This log4J code has been refactored post 2.17.1, will check what it is doing now separately.

[1] https://github.com/apache/logging-log4j2/blob/rel/2.17.2/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java#L475

rjernst (approved):
LGTM

PR #87151, opened 2022-05-26
SQL: Fix date range checks

### LINKED ISSUE #77179: SQL: Conjunction with comparison of datetime field to date string incorrectly yields no results
The following query incorrectly yields no results:

```
select hire_date from test_emp where hire_date > 'now-25y' and hire_date < '2021-09-01'

   hire_date
---------------
```

Expected result is equal to the result returned by only using the first condition:

```
select hire_date from test_emp where hire_date > 'now-25y'

       hire_date
------------------------
1999-04-30T00:00:00.000Z
1997-05-19T00:00:00.000Z
1996-11-05T00:00:00.000Z
```

The second condition matches all records in the test_emp table.

Interestingly, not all conjunctions with this kind of comparison fail in this way:

```
select hire_date from test_emp where hire_date > 'now-25y' and hire_date < 'now/y'

       hire_date
------------------------
1999-04-30T00:00:00.000Z
1997-05-19T00:00:00.000Z
1996-11-05T00:00:00.000Z
```

luigidellaquila:
Fixes #77179


x-pack/plugin/ql/src/test/java/org/elasticsearch/xpack/ql/expression/predicate/RangeTests.java
```diff
@@ -0,0 +1,116 @@
...
+                (Boolean) test[4],
+                l(test[5], (DataType) test[6]),
+                (Boolean) test[7],
+                ZoneId.systemDefault()
+            );
+            assertEquals("failed on test n " + i + ": " + Arrays.toString(test), test[8], range.areBoundariesInvalid());
+        }
+    }
+
+    private static Literal l(Object value, DataType type) {
```
astefan, line 105: There's already a function having the same name and doing the same thing in `FunctionTestUtils.l`.

x-pack/plugin/ql/src/test/java/org/elasticsearch/xpack/ql/expression/predicate/RangeTests.java
```diff
@@ -0,0 +1,116 @@
...
+            Range range = new Range(
+                Source.EMPTY,
+                l(test[0], (DataType) test[1]),
+                l(test[2], (DataType) test[3]),
+                (Boolean) test[4],
+                l(test[5], (DataType) test[6]),
+                (Boolean) test[7],
+                ZoneId.systemDefault()
+            );
+            assertEquals("failed on test n " + i + ": " + Arrays.toString(test), test[8], range.areBoundariesInvalid());
```
astefan, line 101: Either don't add `n` before the test number or use a more complete word for it.

x-pack/plugin/ql/src/test/java/org/elasticsearch/xpack/ql/expression/predicate/RangeTests.java
```diff
@@ -0,0 +1,116 @@
...
+    private static final DataType T = DataTypes.TEXT;
+    private static final DataType DT = DataTypes.DATETIME;
+
+    public void testAreBoundariesInvalid() {
+        // value, value type, lower, lower type, lower included, higher, higher type, higher included, boundaries invalid
+        Object[][] tests = {
+            // dates
+            { d("2021-01-01"), DT, "2021-01-01", K, randomBoolean(), "2022-01-01", K, randomBoolean(), false },
+            { d("2021-01-01"), DT, "2022-01-01", K, randomBoolean(), "2021-01-01", K, randomBoolean(), true },
+            { d("2021-01-01"), DT, "now-10y", K, randomBoolean(), "2022-01-01", K, randomBoolean(), false },
```
astefan, line 33: It doesn't matter if the text field type is keyword or text, so why not use everywhere `randomFrom(TEXT, KEYWORD)`? Extract this in a method `randomTextType()` and use that everywhere.

x-pack/plugin/ql/src/test/java/org/elasticsearch/xpack/ql/expression/predicate/RangeTests.java
```diff
@@ -0,0 +1,116 @@
...
+import org.elasticsearch.xpack.ql.type.DataTypes;
+import org.elasticsearch.xpack.ql.type.DateUtils;
+
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.Arrays;
+
+public class RangeTests extends ESTestCase {
+
+    private static final DataType K = DataTypes.KEYWORD;
```
astefan, line 23: Not a big fan of this approach. `import static DataTypes.KEYWORD` and all other data types and use `KEYWORD`, `TEXT` and `DATETIME` everywhere.

x-pack/plugin/ql/src/test/java/org/elasticsearch/xpack/ql/expression/predicate/RangeTests.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private static Literal l(Object value, DataType type) {
+        return new Literal(Source.EMPTY, value, type);
+    }
+
+    private static ZonedDateTime d(String date) {
+        return DateUtils.asDateTime(date);
+    }
+
+    private static DataType randNType() {
```
astefan, line 113: Use a fuller name for this method. `randomNumericType`, for example.

x-pack/plugin/ql/src/test/java/org/elasticsearch/xpack/ql/expression/predicate/RangeTests.java
```diff
@@ -0,0 +1,116 @@
...
+        return randomFrom(DataTypes.INTEGER, DataTypes.SHORT, DataTypes.LONG, DataTypes.UNSIGNED_LONG, DataTypes.FLOAT, DataTypes.DOUBLE);
```
astefan, line 114: Same here about using `import static` instead of explicit references.

x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/expression/predicate/Range.java
```diff
@@ -108,10 +110,28 @@ public Object fold() {
 
     /**
      * Check whether the boundaries are invalid ( upper &lt; lower) or not.
-     * If they do, the value does not have to be evaluate.
+     * If they do, the value does not have to be evaluated.
      */
-    private boolean areBoundariesInvalid() {
-        Integer compare = BinaryComparison.compare(lower.fold(), upper.fold());
+    protected boolean areBoundariesInvalid() {
+        Object lowerValue = lower.foldable() ? lower.fold() : null;
```
astefan, line 116: Is this ternary operator necessary here? The previous code just folded the values directly.

x-pack/plugin/sql/qa/server/src/main/resources/datetime.csv-spec
```diff
@@ -1570,3 +1570,173 @@ SELECT CAST (CAST (birth_date AS VARCHAR) AS TIMESTAMP) a FROM test_emp WHERE YE
 ---------------
 1965-01-03T00:00:00Z
 ;
+
+
+
+// datetime ranges
+
+dateRange0
```
astefan, line 1578: Can you try and give more meaningful names to these tests?
`dateRangeTwoDateStrings` for example. The next one `dateRange_DateMathAndDate`. And for those that just change the sign (from `<` to `>`) for example, you can add a number at the end: `dateRange_Date_MathAndDate_2`.

astefan (approved):
LGTM
Left some minor/cosmetics comments.

x-pack/plugin/sql/qa/server/src/main/resources/datetime.csv-spec
```diff
@@ -1570,3 +1570,173 @@ SELECT CAST (CAST (birth_date AS VARCHAR) AS TIMESTAMP) a FROM test_emp WHERE YE
 ---------------
 1965-01-03T00:00:00Z
 ;
+
+
+
+// datetime ranges
```
Luegg, line 1576: All the tests here are only using `AND`. I'd assume date math would also work with `OR` and `NOT`?

Luegg (approved):
Lgtm (apart from the outstanding comments).

x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/expression/predicate/Range.java
luigidellaquila: Too defensive probably. I'm removing it.

x-pack/plugin/sql/qa/server/src/main/resources/datetime.csv-spec
luigidellaquila: Good point, adding them

bpintea (approved):
This LGTM, but not sure if we shouldn't expand a fix to cover https://github.com/elastic/elasticsearch/issues/87268 as well.

x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/expression/predicate/Range.java
```diff
@@ -108,10 +110,28 @@ public Object fold() {
 
     /**
      * Check whether the boundaries are invalid ( upper &lt; lower) or not.
-     * If they do, the value does not have to be evaluate.
+     * If they do, the value does not have to be evaluated.
```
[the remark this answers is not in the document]
luigidellaquila: Thanks @bpintea 
I checked https://github.com/elastic/elasticsearch/issues/87268, it seems to be a bit out of the scope of this fix (see my comment on the issue). 
I'd suggest to fix it in a separate PR

costin (approved):
While the fix is up and to the point, there's a subtle behavior that we're introducing here that has ramifications: automatic type conversion.
The date to string comparison works when pushed down to Elastic - which is always. However that's a black box to SQL since we don't understand the string expression and thus cannot reason about it:
`date > 'now/y'` vs `date > 'someString'`

The proper way to correct this is to promote the string to a date (though this will have implications over the ES evaluation).
That is `date > 'now'` should be translated into `date > NOW()`, etc...

Currently we're half pushing/hiding the date promotion of string inside the range operator.
We should have a dedicated rule that picks this patterns up and promotes the special date strings accordingly.
I'm okay with merging this PR as long as a dedicated issue is created for promoting date expression strings into proper date times.

x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/expression/predicate/Range.java
```diff
@@ -108,10 +110,28 @@ public Object fold() {
...
      * Check whether the boundaries are invalid ( upper &lt; lower) or not.
-     * If they do, the value does not have to be evaluate.
+     * If they are, the value does not have to be evaluated.
      */
-    private boolean areBoundariesInvalid() {
-        Integer compare = BinaryComparison.compare(lower.fold(), upper.fold());
+    protected boolean areBoundariesInvalid() {
+        Object lowerValue = lower.fold();
+        Object upperValue = upper.fold();
+        if (DataTypes.isDateTime(value.dataType()) || DataTypes.isDateTime(lower.dataType()) || DataTypes.isDateTime(upper.dataType())) {
```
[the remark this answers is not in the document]
luigidellaquila: Thank you @costin, I created a new issue about this https://github.com/elastic/elasticsearch/issues/87500

About the check on `value`, actually the value is never involved in this comparison (not in this method at least), so there is no risk to have false positives due to it (on the other hand, false negatives are tolerated, eg. for date maths).

PR #87156, opened 2022-05-26
Refactor ParameterizedMessage used in lambda and casted to Supplier

pgomulka:
This is a result of structural search/replace in intellij. This only affects log methods with a signature
logger.info((Supplier<?>) ()-> ParametrizedMessage)
logger.info((Supplier<?>) ()-> ParametrizedMessage, Throwable)

relates #86549


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/watcher/transform/chain/ExecutableChainTransform.java
```diff
@@ -47,7 +45,7 @@ public ChainTransform.Result execute(WatchExecutionContext ctx, Payload payload)
         try {
             return doExecute(ctx, payload, results);
         } catch (Exception e) {
-            logger.error((Supplier<?>) () -> new ParameterizedMessage("failed to execute [{}] transform for [{}]", TYPE, ctx.id()), e);
+            logger.error(() -> org.elasticsearch.core.Strings.format("failed to execute [%s] transform for [%s]", TYPE, ctx.id()), e);
```
grcevski, line 48: Nit: might be a bit clearer if we didn't have the full class name here, I don't see another `Strings` class imported.

grcevski (approved):
LGTM! I just left one minor pedantic comment.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/watcher/transform/chain/ExecutableChainTransform.java
pgomulka: there is `import static org.elasticsearch.common.logging.LoggerMessageFormat.format` which probably deserves a refactoring on its own

PR #87157, opened 2022-05-26
Modularize the data-streams component

ChrisHegarty:
This is change modularizes data-streams, by adding a module-info.java

relates #78744

### DIFF
--- a/modules/data-streams/src/main/java/module-info.java
+++ b/modules/data-streams/src/main/java/module-info.java
@@ -1,0 +1,17 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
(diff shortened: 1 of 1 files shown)


modules/data-streams/src/main/java/module-info.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+module org.elasticsearch.datastreams {
+    requires org.elasticsearch.base;
+    requires org.elasticsearch.server;
+    requires org.elasticsearch.xcontent;
+    requires org.apache.logging.log4j;
+    requires org.apache.lucene.core;
+
+    exports org.elasticsearch.datastreams.action to org.elasticsearch.server;
```
rjernst, line 16: I assume this is because of some reflective access in server? Should we document this type of export so that other developers know what errors to look for that require this export?

rjernst (approved):
LGTM

PR #87158, opened 2022-05-26
Make the bucket_path full path optional if the target multi value aggregation has a single value

salvatore-campagna:
Before this change having `bucket_path` fully qualify the target 
field path in the params section was required.
With this change the field full path becomes optional. If the path points
to a multi value aggregation including more than one metric the field
is still required. If, instead, only one value is available that single value
is used just specifying the multi value aggregation name.

The old notation is still supported for backward compatibility.


salvatore-campagna:

PR #87168, opened 2022-05-26
Convert o.e.cluster to use Map in public methods

rjernst:
Many classes that use ImmutableOpenMap expose the maps directly, when
the consumers actually only use methods from the base Map interface.
This commit converts public methods in subpackages of o.e.cluster to
return Map instead of ImmutableOpenMap. Note that o.e.cluster classes
are not yet converted because they require more involved changes.

relates #86239


grcevski (approved):
LGTM!

PR #87172, opened 2022-05-26
[ML] Split node load counts by job type

dimitris-athanasiou:
This refactoring is in preparation to remove native inference jobs
from the jobs that account towards the max open jobs limit.

### DIFF
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/job/NodeLoad.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/job/NodeLoad.java
@@ -25,9 +25,9 @@ public class NodeLoad {
     private final String nodeId;
     private final boolean useMemory;
     private final String error;
-    private int numAssignedAnomalyDetectorJobs;
-    private int numAssignedDataFrameAnalyticsJobs;
-    private int numAssignedNativeInferenceJobs;
+    private final int numAssignedAnomalyDetectorJobs;
+    private final int numAssignedDataFrameAnalyticsJobs;
+    private final int numAssignedNativeInferenceJobs;
     private final long assignedNativeCodeOverheadMemory;
     private final long assignedAnomalyDetectorMemory;
     private final long assignedDataFrameAnalyticsMemory;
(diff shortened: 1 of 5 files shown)


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/job/NodeLoad.java
```diff
@@ -25,7 +25,9 @@ public class NodeLoad {
     private final String nodeId;
     private final boolean useMemory;
     private final String error;
-    private final int numAssignedJobs;
+    private int numAssignedAnomalyDetectorJobs;
+    private int numAssignedDataFrameAnalyticsJobs;
+    private int numAssignedNativeInferenceJobs;
```
benwtrent, line 30:
```suggestion
    private final int numAssignedAnomalyDetectorJobs;
    private final int numAssignedDataFrameAnalyticsJobs;
    private final int numAssignedNativeInferenceJobs;
```

Everything in NodeLoad should be final.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/job/NodeLoad.java
```diff
@@ -261,15 +275,15 @@ public long getFreeMemory() {
         }
 
         public int remainingJobs() {
-            return Math.max(maxJobs - numAssignedJobs, 0);
+            return Math.max(maxJobs - getNumAssignedJobs(), 0);
```
benwtrent, line 278: I thought we were going to remove pytorch from the job count on the node because its using different executors.

dimitris-athanasiou: Yes, but in the PR where we wire in the planner.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/job/NodeLoad.java
dimitris-athanasiou: 🤦

benwtrent (approved):

PR #87179, opened 2022-05-26
Update targz install instructions to use curl instead of wget

kilfoyle:
I don't believe WGET is included automatically on macOS. This updates the Elasticsearch [install instructions for macOS](https://www.elastic.co/guide/en/elasticsearch/reference/current/targz.html#install-macos) to match the [similar instructions for Kibana](https://www.elastic.co/guide/en/kibana/current/targz.html#install-darwin64), which use curl.


kilfoyle:
Closing. I should have targeted `master` so rather than rebase I'll open a new PR.

PR #87188, opened 2022-05-26
How to increase node capacity docs

masseyke:
This adds troubleshooting documentation for the case when the ShardsAvailabilityHealthIndicatorService
reports that there are not enough nodes in the data tier (user action "increase_node_capacity_for_allocations" or
"increase_tier_capacity_for_allocations_<tier>". This covers both the cloud and self-managed environments. For
cloud we first recommend increasing the number of availability zones (because you cannot directly add nodes), and
decreasing `index.number_of_replicas` if that is not possible. For self-managed, we first recommend adding nodes,
and decreasing `index.number_of_replicas` if that is not possible.


docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc
```diff
@@ -0,0 +1,251 @@
...
+targeting the `warm` tier and more nodes with the `data_warm` role are needed in
+the {es} cluster.
+
+//end::kibana-api-ex-2[]
+
+//tag::increase-azs[]
+[increase-azs]
+Now that you know the tier, increase the number of availability zones.
+
+. Go back to the deployment's landing page and click the
```
andreidan, line 90: Shall we mention how to get back here? (Mentioning the top left button/link in Kibana that does this)

docs/reference/troubleshooting/data/increase-node-capacity.asciidoc
```diff
@@ -0,0 +1,18 @@
+[[increase-node-capacity]]
+== Not enough nodes to place all shard replicas
+
+Elasticsearch tries to keep copies of the data (index shard replicas) on different nodes so that
+if one node fails the data is still available on another node. If the index has a preferred tier,
+Elasticsearch will only try to place the copies of the data for that index on nodes in that tier.
```
andreidan, line 6:
```suggestion
Elasticsearch will only try to place the copies of the data for that index on nodes in the target tier.
```

andreidan (commented):
Thanks for putting this together Keith. 

I left a few minor comments. Can you please link a preview to the PR too?

docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc
```diff
@@ -0,0 +1,251 @@
...
+<1> Represents a comma separated list of data tier node roles this index is allowed
+to be allocated on, the first one in the list being the one with the higher priority
+i.e. the tier the index is targeting.
+e.g. in this example the tier preference is `data_warm,data_hot` so the index is
+targeting the `warm` tier and more nodes with the `data_warm` role are needed in
+the {es} cluster.
+
+
+Alternatively, if adding more nodes to the {es} cluster is not desired,
+inspect the `cluster.routing.allocation.total_shards_per_node`
```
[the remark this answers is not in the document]
masseyke: Good catch!

docs/reference/troubleshooting/data/increase-node-capacity.asciidoc
```diff
@@ -0,0 +1,18 @@
+[[increase-node-capacity]]
+== Not enough nodes to place all shard replicas
+
+Elasticsearch tries to keep copies of the data (index shard replicas) on different nodes so that
+if one node fails the data is still available on another node. If the index has a preferred tier,
```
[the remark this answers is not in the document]
masseyke: That (and more) will have been in the impact on the indicator that got them here, right? This part is more focused on just how to fix it isn't it?

docs/reference/troubleshooting/data/increase-node-capacity.asciidoc
```diff
@@ -0,0 +1,18 @@
+[[increase-node-capacity]]
```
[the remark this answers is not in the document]
masseyke: I read "increase node capacity" as "increase the number of nodes" but I guess it is ambiguous. Are you suggesting changing it everywhere (including file names) to "increase-tier-capacity"? I can do that.

Leaf-Lin:
Can we also add somewhere on this page to describe `the number-of-replicas` should <= `the number-of-eligible-nodes-in-the-tier` - 1?

docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc
```diff
@@ -0,0 +1,251 @@
...
+the {es} cluster.
+
+//end::kibana-api-ex-2[]
+
+//tag::increase-azs[]
+[increase-azs]
+Now that you know the tier, increase the number of availability zones.
+
+. Go back to the deployment's landing page by clicking on the three horizontal bars on the top left of the screen
+and choosing **Manage this deployment**. On that page click the **Manage** button, and choose **Edit deployment**.
```
Leaf-Lin, line 91: I believe this requires the user authenticated can have a certain privilege.  Can we add some text to the effect?

docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc
```diff
@@ -0,0 +1,251 @@
...
+. In the {es} section, find the tier where the replica shards could not be assigned. Select a higher number
+of availability zones for that tier.
++
+
+[role="screenshot"]
+image::images/data-tiers/ess-advanced-config-data-tiers.png[{kib} Console,align="center"]
+//end::increase-azs[]
+
+If it is not possible to increase the number of availability zones, you can reduce the number of replicas of your index
+data. We'll achieve this by inspecting the <<dynamic-index-number-of-replicas,`index.number_of_replicas`>> index setting
```
Leaf-Lin, line 102: I found it's a bit strange to talk about increasing the availabitiy zone which is related to fault tolerance.  Shouldn't we be talking about adjusting the `size per zone` from the dropdown list? 

We can also link to this ESS page: https://www.elastic.co/guide/en/cloud/current/ec-customize-deployment-components.html#ec-cluster-size

Leaf-Lin: Having thought about it more, adding node capacity is related to the max size of nodes in the tier. 
--> If max size of the node in the tier >= 58GB, users can use either dropdown `size per zone` or `availability zone` to add more nodes.
--> If max size of the node < 58GB, users should use `availability zone` to add more nodes.

docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc
Leaf-Lin: Users authenticated in Kibana are different from users authenticated in cloud console. :sigh:

- All users logged into the cloud console do have the same permission (for now) and will be allowed for `Manage this deployment`.
- Unfortunately, users logged into Kibana may only have a subset of permission and may not see the `Manage this deployment` link.

I understand if they follow the steps from the cloud console --> Kibana, they will be able to see the link. It just feels safer to add a warning that if they do not see this link, they can start from cloud.elastic.co and navigate to the deployment management page.

docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc
```diff
@@ -0,0 +1,299 @@
...
+----
+// TESTRESPONSE[skip:the result is for illustrating purposes only]
++
+You can count the rows containing the letter representing the target tier to know how many nodes you have. See
+<<cat-nodes-api-query-params>> for details. The example above has two rows containing `h`, so there are two
+nodes in the hot tier.
++
+. <<indices-update-settings, Decrease>> the
+value for the total number of replica shards required for this index. The new value needs to be less
+than the number of nodes found above minus one. Since the example above found 2 nodes in the hot tier,
```
Leaf-Lin, line 159:
```suggestion
. <<indices-update-settings, Decrease>> the value for the total number of 
replica shards required for this index. As replica shards cannot reside on the
same node as primary shards for <<high-availability-cluster-design,high
availability>>, the new value needs to be less than the number of nodes found
above minus one. Since the example above found 2 nodes in the hot tier,
```

I meant to link the high availability text to the following link, is that correct? 
https://www.elastic.co/guide/en/elasticsearch/reference/current/high-availability-cluster-small-clusters.html

docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc
```diff
@@ -0,0 +1,299 @@
...
+  }
+}
+----
+// TEST[continued]
+
++
+<1> The new value for the `index.number_of_replicas` index configuration
+is decreased from the previous value of `2` to `1`. It can be set as low as 0 but
+configuring it to 0 may lead to temporary availability loss during node restarts
+or permanent data loss in case of data corruption.
```
Leaf-Lin, line 177:
```suggestion
is decreased from the previous value of `2` to `1`. It can be set as low as 0 
but configuring it to 0 for indices other than <<searchable-snapshots,
searchable snapshot indices>> may lead to temporary availability loss during 
node restarts or permanent data loss in case of data corruption.```

File: docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc [hunk 10]
```diff
@@ -0,0 +1,299 @@
...
+----
+// TESTRESPONSE[skip:the result is for illustrating purposes only]
++
+You can count the rows containing the letter representing the target tier to know how many nodes you have. See
+<<cat-nodes-api-query-params>> for details. The example above has two rows containing `h`, so there are two
+nodes in the hot tier.
++
+. <<indices-update-settings, Decrease>> the
+value for the total number of replica shards required for this index. The new value needs to be less
+than the number of nodes found above minus one. Since the example above found 2 nodes in the hot tier,
```
Leaf-Lin, line 278:
```suggestion
value for the total number of replica shards required for this index. The new value needs to be less
than or equal to the number of nodes found above minus one. Since the example above found 2 nodes in the hot tier,
```

docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc
```diff
@@ -0,0 +1,299 @@
...
+    "number_of_replicas" : 1 <1>
+  }
+}
+----
+// TEST[continued]
+
++
+<1> The new value for the `index.number_of_replicas` index configuration
+is decreased from the previous value of `2` to `1`. It can be set as low as 0 but
+configuring it to 0 may lead to temporary availability loss during node restarts
```
Leaf-Lin, line 295:
```suggestion
is decreased from the previous value of `2` to `1`. It can be set as low as 0 
but configuring it to 0 for indices other than <<searchable-snapshots,
searchable snapshot indices>> may lead to temporary availability loss during 
node restarts
```

docs/reference/troubleshooting/data/increase-node-capacity.asciidoc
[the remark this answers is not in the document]
Leaf-Lin: I also think `increase-tier-capacity` makes more sense here.

docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc
masseyke: Are you saying that in cloud we use at most 58 GB boxes, so if you ask for more than 58 GB of RAM you're implicitly asking for more than one node? That seems like an implementation detail that could change at any moment doesn't it?

masseyke: Maybe you're saying something else though? I just changed my "size per zone" to 64 GB, with 2 zones. But when I use `_cat/nodes`, I still only have 2 nodes. Is there an option I'm missing?

docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc
masseyke: I put in a sentence about needing to be logged in to https://cloud.elastic.co/

docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc
Leaf-Lin: I've modified my comment above to adjust the threshold to 64GB for all regions and cloud providers. I don't think this value is likely to change as it is very much depends on the [magical 32GB heap boundary](https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html#compressed_oops)

docs/reference/troubleshooting/data/increase-node-capacity.asciidoc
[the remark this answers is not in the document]
andreidan: Oh sure, I was proposing we rephrase a bit this: 
> Elasticsearch tries to keep copies of the data (index shard replicas) on different nodes so that
if one node fails the data is still available on another node.

To something along the lines of:
>Elasticsearch tries to keep copies of the data (index shard replicas) on different nodes to protect against
hardware failure and increase the capacity to serve requests like searching for documents.

andreidan:
@masseyke 

>That will require adding sections on how to get the number of eligible nodes in the tier. I'll try to find a way to do that without making this too bloated.

I think Leaf suggested just describing that behaviour. Maybe also a mention to `_cat/nodes` that will show how many nodes belong to each tier ?

docs/reference/tab-widgets/troubleshooting/data/increase-node-capacity.asciidoc
masseyke: OK I see what you're saying now. At 64 GB RAM per zone, you still only get 1 node/box, but every 64 GB you go up from there you get another node. I confirmed it in the _cat/nodes API.

masseyke: Let me know what you think of the way I phrased this. Maybe I should add something like "figure out which way is cheapest for you, and do that." I didn't want to put all of that math into this document though.

Leaf-Lin:
Link the preview link:
https://elasticsearch_87188.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/increase-tier-capacity.html

docs/reference/tab-widgets/troubleshooting/data/increase-tier-capacity.asciidoc
```diff
@@ -0,0 +1,317 @@
...
+}
+----
+// TEST[continued]
+
++
+<1> The new value for the `index.number_of_replicas` index configuration
+is decreased from the previous value of `2` to `1`. It can be set as low as 0
+but configuring it to 0 for indices other than <<searchable-snapshots,
+searchable snapshot indices>> may lead to temporary availability loss during
+node restarts or permanent data loss in case of data corruption.```
```
Leaf-Lin, line 193:
```suggestion
node restarts or permanent data loss in case of data corruption.
```

Remove extra ```

docs/reference/troubleshooting/data/increase-tier-capacity.asciidoc
```diff
@@ -0,0 +1,18 @@
...
+== Not enough nodes to allocate all shard replicas
+
+Elasticsearch tries to keep copies of the data (index shard replicas) on different nodes to protect against
+hardware failure and increase the capacity to serve requests like searching for documents. If the index has
+a preferred tier, Elasticsearch will only try to place the copies of the data for that index on nodes in
+the target tier.
+
+Users might want to influence this behavior by adding more nodes to the cluster (or tier if tiers are
+in use), or by reducing the
+<<dynamic-index-number-of-replicas,`index.number_of_replicas`>> index setting.
```
Leaf-Lin, line 11:
```suggestion
Distributing copies of the data (index shard replicas) on different nodes can 
parallelize processing requests thus speeding up search queries. This can be 
achieved by increasing the number of replica shards up to the maximum value
(total number of nodes minus one) which also serves the purpose to protect
against hardware failure. If the index has a preferred tier, Elasticsearch will
only place the copies of the data for that index on nodes in the target tier.

If a warning is encountered with not enough nodes to allocate all shard 
replicas, you can influence this behavior by adding more nodes to the cluster
(or tier if tiers are in use), or by reducing the
<<dynamic-index-number-of-replicas,`index.number_of_replicas`>> index setting.
```

Reworking on the opening a bit to add the reasons why replica shards are needed and strengthen the connection between cause and action.

Leaf-Lin (commented):
A few more minor points, but the new section on ESS tier vs size per node options looks good!

andreidan (approved):
LGTM, thanks Keith

PR #87189, opened 2022-05-27
Fix the ingest attachment license

rjernst:
This commit fixesx the license/notice files for ingest attachment
dependency to account for both tika-langdetect and tika-langdetect-tika
jars when building dependency info.


rjernst:
Given the release manager is broken right now because of this, I’m going to merge.

PR #87209, opened 2022-05-30
Small speedup in FieldMapper.Builder.parse

original-brownbear:
Creating the params map is quite hot in benchmarks.
Presizing it correctly helps a little and speeds up `FieldMapper.Builder.parse` measurably. Also replaced a needless series
of conditionals with a switch to make the massive loop a little cheaper.

relates #77466


original-brownbear:

server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
```diff
@@ -1203,9 +1204,10 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
          * @param fieldNode         the root node of the map of mappings for this field
          */
         public final void parse(String name, MappingParserContext parserContext, Map<String, Object> fieldNode) {
-            Map<String, Parameter<?>> paramsMap = new HashMap<>();
+            final Parameter<?>[] params = getParameters();
+            Map<String, Parameter<?>> paramsMap = Maps.newHashMapWithExpectedSize(params.length);
             Map<String, Parameter<?>> deprecatedParamsMap = new HashMap<>();
```
javanna, line 1209: the params map is hotter than the deprecated params map, because many params are added to it? Leave a comment around this to explain why we create the map this way?

javanna (approved):
I am fine with this change, it is contained enough, although I would prefer to have a test for it (but I could not come up with one that makes sense) or a clear indication of the performance improvement and some guarantee that it will not be lost over time. Let's have this discussion separately

server/src/main/java/org/elasticsearch/index/mapper/FieldMapper.java
original-brownbear: Yea, the deprecated names are mostly empty I think + we don't know the size up front without iterating all parameters -> added comment about that.

original-brownbear:
Thanks Luca!

PR #87216, opened 2022-05-30
[TEST] Expand tests for subobjects false

javanna:
This commit adds more tests that cover edge cases around setting subobjects to false for a given object. The new tests focus on dynamic mapping behaviour, for instance when dynamic is set to false on the parent mapper, or the mapping for the object itself comes from a dynamic template. Also, a new test is added to cover geo_point parsing, meaning to verify that objects are accepted in a document when for fields that are mapped (either in the mapping or through dynamic template) and support parsing objects natively.


ywelsch (approved):
LGTM

javanna:
thanks @ywelsch for the speedy review!

PR #87217, opened 2022-05-30
[TEST] Rename DynamicTemplateTests to DynamicTemplateParseTests

javanna:
We have had until today DynamicTemplateTests as well as DynamicTemplatesTests. When adding a new test for dynamic templates, naming does not help understanding what the right place is. This commit addressed this by renaming DynamicTemplateTests to DynamicTemplateParseTests as it holds only tests around the static parse method that parse a dynamic template into its corresponding object.


ywelsch (approved):
LGTM

PR #87218, opened 2022-05-30
[TEST] Verify that nested type is not supported when subobjects is false 

javanna:
We have tests that verify that objects can't be created in the mapping under an object that has subobjects set to false. The existing checks work also for the nested type, as NestedObjectMapper extends ObjectMapper, yet this commit adds specific tests for this scenario.

This helps clarify that we do need to check this in two places, once at parse time and once in buildMappers, for objects that are dynamically mapped while parsing incoming documents.

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/mapper/ObjectMapper.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/ObjectMapper.java
@@ -152,7 +152,7 @@ public class ObjectMapper extends Mapper implements Cloneable {
                     throw new IllegalArgumentException(
                         "Object ["
                             + context.buildFullName(name)
-                            + "] has subobjects set to false hence it does not support inner object or nested ["
+                            + "] has subobjects set to false hence it does not support inner object ["
                             + mapper.simpleName()
                             + "]"
                     );
--- a/server/src/main/java/org/elasticsearch/index/mapper/ObjectMapper.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/ObjectMapper.java
@@ -149,7 +149,7 @@ public class ObjectMapper extends Mapper implements Cloneable {
             for (Mapper.Builder builder : mappersBuilders) {
                 Mapper mapper = builder.build(mapperBuilderContext);
                 if (subobjects.value() == false && mapper instanceof ObjectMapper) {
-                    throw new IllegalStateException(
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/elasticsearch/index/mapper/ObjectMapper.java
```diff
@@ -280,14 +282,23 @@ protected static void parseProperties(
...
                                 + "] has subobjects set to false hence it does not support inner object ["
                                 + fieldName
                                 + "]"
                         );
                     }
+                    if (objBuilder.subobjects.value() == false && type.equals(NestedObjectMapper.CONTENT_TYPE)) {
+                        throw new MapperParsingException(
+                            "Object ["
+                                + objBuilder.name()
+                                + "] has subobjects set to false hence it does not support nested object ["
```
ywelsch, line 297: here you create a separate exception message for nested (vs object) fields, whereas in buildMappers you don't?

ywelsch (commented):
I've left one question, o.w. looking good.

server/src/main/java/org/elasticsearch/index/mapper/ObjectMapper.java
javanna: yea this is odd, it's a consequence of how the check is performed, the two have a different type, but NestedObjectMapper extends ObjectMapper hence one instanceof check covers them both. Another reason is that I found no way to test a possible separate error for nested, because nested type never gets dynamically mapped. If it does it comes from a dynamic template but that does go through parsing and fails earlier during validation (test that I added to DynamicTemplatesTests). All in all I thought it's not wort going the extra mile of adding a separate error message for nested here. I am curious if you have other ideas though.

ywelsch: Can we assert then that we never have a NestedObjectMapper there?

javanna: yea that could work. That should help figuring out how to test that scenario if it ends up being possible :)

ywelsch: right, it might hold a surprise :)

ywelsch (approved):
LGTM

javanna:
Thanks for the review Yannick!

PR #87219, opened 2022-05-30
Modularize the ingest.common component (as well as dissect and grok dependent libs)

ChrisHegarty:
This is change modularizes the ingest.common component, by adding a module-info.java. As well as two dependent libs.

The project only requires painless SPI to compile, so that was fixed along the way ( so that the compile module path can be inferred directly from the dependencies ).

relates #78744


rjernst (approved):
LGTM

PR #87224, opened 2022-05-30
Cancellable Profile Has Privilege check 

albertzaharovits:
The "profile has privileges" endpoint is designed to check the privileges of
many users, given their profile ids, in a single call.
This commit runs the privileges check under a "cancellable" task.
When the task is cancelled, the check is interrupted for the remaining
profiles to-be-checked. The task can also be cancelled when the HTTP
client disconnected.


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/AuthorizationEngine.java
```diff
@@ -291,6 +292,16 @@ public ActionRequestValidationException validate(ActionRequestValidationExceptio
...
+                for (int i = 0; i < index.length; i++) {
+                    BytesReference query = index[i].getQuery();
+                    if (query != null) {
+                        validationException = addValidationError(
+                            "may only check index privileges without any DLS query [" + query.utf8ToString() + "]",
+                            validationException
+                        );
+                    }
+                }
             }
```
albertzaharovits, line 305: This is a fix outside of the scope for this PR. I've included it because it is minor.
The DLS verification was part of the transport action not the proper request validation.
Consequently the new "profile has privilege" check missed to also run the same verification.
I have hence moved the verification as part of the request validation, where it's harder to miss.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/profile/ProfileCancellationIntegTests.java
```diff
@@ -61,9 +84,19 @@ protected boolean addMockHttpTransport() {
     protected Collection<Class<? extends Plugin>> nodePlugins() {
         final List<Class<? extends Plugin>> plugins = new ArrayList<>(super.nodePlugins());
         plugins.add(SearchBlockPlugin.class);
+        {
+            // replace the security plugin with the security plugin with the dummy authorization engine extension
+            plugins.remove(LocalStateSecurity.class);
+            plugins.add(LocalStateWithDummyAuthorizationEngineExtension.class);
+        }
```
albertzaharovits, line 91: I need this in order to block the "check privileges" invocation. This adds and dummy authorization engine (which authz everything) in this IT suite.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/profile/TransportProfileHasPrivilegesAction.java
```diff
@@ -82,6 +88,11 @@ protected void doExecute(Task task, ProfileHasPrivilegesRequest request, ActionL
                 request,
                 ActionListener.wrap(applicationPrivilegeDescriptors -> threadPool.generic().execute(() -> {
                     for (Map.Entry<String, Subject> profileUidToSubject : profileUidAndSubjects) {
+                        // return the partial response if the "has privilege" task got cancelled in the meantime
+                        if (((CancellableTask) task).isCancelled()) {
+                            listener.onResponse(new ProfileHasPrivilegesResponse(hasPrivilegeProfiles, errorProfiles));
+                            return;
+                        }
```
albertzaharovits, line 95: I think we could discuss whether to verify the cancellation status before every profile is checked, or every _n_ such checks.
We could also discuss whether an error is more suitable than the partial result.

I don't think the ES JVM node process would notice the performance impovement of every _n_ checks, and partial results is nice and comes for free (there's no headache to implement it).

ywangd (commented):
We mostly just need agree on the partial result handling (see below).

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/AuthorizationEngine.java
albertzaharovits: Aaargs there's validation in parse as well!!
I agree we should contain all of these in a single place, which is probably the _parse_ method, as you suggest.
I'm going to leave this for another time. I've had the fair share of "side quests" in the PR.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/profile/TransportProfileHasPrivilegesAction.java
albertzaharovits: Roger! I'll keep the cancellation status check once every profile, but return the `TaskCancelledException` rather than the partial result.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/profile/TransportProfileHasPrivilegesActionTests.java
```diff
@@ -180,10 +189,86 @@ public void testNoProfileSubjectsFound() throws Exception {
...
+        }).when(profileService).getProfileSubjects(anyCollection(), anyActionListener());
+
+        doAnswer(invocation -> {
+            final ActionListener<Collection<ApplicationPrivilegeDescriptor>> listener = (ActionListener<
+                Collection<ApplicationPrivilegeDescriptor>>) invocation.getArguments()[2];
+            listener.onResponse(List.of());
+            return null;
+        }).when(nativePrivilegeStore).getPrivileges(anyCollection(), any(), anyActionListener());
+
+        final int partialResultSize = randomIntBetween(1, profileUids.size());
```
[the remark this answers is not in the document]
albertzaharovits: You mean that the cancellation logic is not exercised in that case, right?

It's probably not ideal from a test design perspective, but it's also not completely irrelevant because, if the API supports partial results, there's still this case where the task is cancelled after the code last check the cancel status before returning a non-error response.
IMO it meets the bar for a randomization case.

PS: I initially wrote this test to mock the `CancellableTask#isCancelled`, but that is a final method, so I recurred to the currently proposed alternative (so this test makes it look like it is testing that every partial result is possible, but the goal is to cover that the task can be canceled at any point).

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/profile/TransportProfileHasPrivilegesAction.java
albertzaharovits: I've pushed [e918fd8](https://github.com/elastic/elasticsearch/pull/87224/commits/e918fd81112d596d1e9d79d71d141769fca4e11a) .

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/profile/TransportProfileHasPrivilegesActionTests.java
```diff
@@ -180,10 +190,84 @@ public void testNoProfileSubjectsFound() throws Exception {
...
+            listener.onResponse(List.of());
+            return null;
+        }).when(nativePrivilegeStore).getPrivileges(anyCollection(), any(), anyActionListener());
+
+        final AtomicInteger cancelAfter = new AtomicInteger(randomIntBetween(1, profileUids.size()));
+        final CancellableTask cancellableTask = new CancellableTask(0, "type", "action", "description", TaskId.EMPTY_TASK_ID, Map.of());
+
+        if (cancelAfter.decrementAndGet() == 0) {
+            TaskCancelHelper.cancel(cancellableTask, "reason");
+        }
```
ywangd, line 256: I am a little unsure about the intention of this new `if` block. Is it to address my previous comment about how `isCancelled` is `not` checked in the main code when `partialResultSize == profileUids.size()`. I thought we agreed it is a valid randomisation scenario and we should keep it? 

Or are you trying to cover the scenario where the task is cancelled before even the 1st check (I believe that's what happens when `profileUids.size() == 1`? This is a valid scenario as well. But it makes the previous scenario (cancelled _after_ last check is performed) no longer tested. Would it be better to wrap this with a `randomBoolean`, i.e. `if (randomBoolean() && cancelAfter.decrementAndGet() == 0) { ... }`, to maximise its coverage for both scenarios?

ywangd (approved):
LGTM

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/profile/TransportProfileHasPrivilegesActionTests.java
albertzaharovits: I have pushed [a39a476](https://github.com/elastic/elasticsearch/pull/87224/commits/a39a4768d975869cfed7626b11d8a35e8a4a7b57) .

The task in the test scenario is now cancelled before any `checkPrivileges` as well as after all the `checkPrivileges` .

PR #87225, opened 2022-05-30
Replace log4j-1.2 with log4j2 API

pgomulka:
Some of the logging usages are incorrectly using log4j v 1.2
instead of log4j2-api.
This commit fixes this.


modules/repository-url/src/main/java/org/elasticsearch/common/blobstore/url/http/RetryingHttpInputStream.java
```diff
@@ -149,7 +148,7 @@ private void ensureOpen() {
     private void maybeThrow(IOException e) throws IOException {
         if (retryCount >= maxRetries || e instanceof NoSuchFileException) {
             logger.debug(
-                new ParameterizedMessage(
+                () -> format(
                     "failed reading [{}] at offset [{}], retry [{}] of [{}], giving up",
```
grcevski, line 152: I think we need to use %s here inside the `format` instead of {}

modules/repository-url/src/main/java/org/elasticsearch/common/blobstore/url/http/RetryingHttpInputStream.java
```diff
@@ -162,7 +161,7 @@ private void maybeThrow(IOException e) throws IOException {
         }
 
         logger.debug(
-            new ParameterizedMessage(
+            () -> format(
```
grcevski, line 164: Same here.

grcevski (changes requested):
The package changes look good, but I think there's a typo in the format usage.

grcevski (approved):
LGTM!

PR #68749, opened 2021-02-09
Add read permissions for `apm_user` role to APM fleet indices

sorenlouv:
Related: https://github.com/elastic/kibana/issues/87501

**Permissions to `apm_user`**
Starting in 7.12 APM Server will be available under fleet. In this mode data will no longer be ingested to `apm-*` indices but to `logs-apm*, metrics-apm*, traces-apm*`.

This PR ensures that users upgrading to APM Server under fleet can still access APM data.

~**Permissions to `kibana_user`**
In Kibana there is a background task that collects telemetry. We want ensure that this background task can access APM indices and upload telemetry based on this.~

Note to self: 
```sh
# run tests
./gradlew :x-pack:plugin:core:test --tests 'org.elasticsearch.xpack.core.security.authz.store.*'
```

cc @elastic/apm-ui

PR #87233, opened 2022-05-31
Remove apm_user role

sorenlouv:
Meta issue: https://github.com/elastic/kibana/issues/116760
The `apm_user` role was [marked as deprecated](https://github.com/elastic/elasticsearch/pull/68749) in 7.13 and was supposed to be removed in 8.0.
All mentions of `apm_user` role was finally removed in https://github.com/elastic/kibana/pull/132790 and this PR removes the actual role

### DIFF
--- a/docs/changelog/87233.yaml
+++ b/docs/changelog/87233.yaml
@@ -1,0 +1,12 @@
+pr: 87233
+summary: Remove `apm_user` role
+area: Authorization
+type: breaking
+issues: []
+breaking:
+  title: Remove `apm_user` role
+  area: Authorization
+  details: Please describe the details of this change for the release notes. You can
+    use asciidoc.
+  impact: Please describe the impact of this change to users
+  notable: false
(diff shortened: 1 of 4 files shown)


sqren:
I added the `>breaking` label since this is technically a breaking change. OTOH the role was deprecated over a year ago, and the role was supposed to have been removed in 8.0. Finally removing it can be seen as an implementation detail.

ywangd:
@sqren IIUC, the `apm_user` role was meant to be removed in 8.0, but we forgot to do so? If that's true, I think now it needs to go through the breaking change committee first for future removal in (possibly) 9.0.

ywangd:
> Does it still need to go through the breaking change committee if we move the removal to 9.0?

Yes. We have a similar case and it needed to go through the committe. I'll tag you on the issue.

> I'll update the PR to do that. 

It's hard to do that now since we don't have a 9.0 branch yet.

sqren:
Replaced by https://github.com/elastic/elasticsearch/pull/87674

PR #87238, opened 2022-05-31
Add assertion that all field mapper names are interned

original-brownbear:
Follow up to #86301. Assert that all field names are interned so we don't
lose this optimization and fix `FieldAliasMapper.name` for it not to trip.

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/mapper/MappingLookup.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/MappingLookup.java
@@ -186,6 +186,22 @@ public final class MappingLookup {
         this.indexTimeScriptMappers = List.copyOf(indexTimeScriptMappers);
 
         runtimeFields.stream().flatMap(RuntimeField::asMappedFieldTypes).map(MappedFieldType::name).forEach(this::validateDoesNotShadow);
+        assert assertMapperNamesInterned(this.fieldMappers, this.objectMappers);
+    }
+
+    private static boolean assertMapperNamesInterned(Map<String, Mapper> mappers, Map<String, ObjectMapper> objectMappers) {
+        mappers.forEach(MappingLookup::assertNamesInterned);
+        objectMappers.forEach(MappingLookup::assertNamesInterned);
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/elasticsearch/index/mapper/MappingLookup.java
```diff
@@ -186,6 +186,22 @@ private MappingLookup(
...
+        return true;
+    }
+
+    private static void assertNamesInterned(String name, Mapper mapper) {
+        assert name == name.intern();
+        assert mapper.name() == mapper.name().intern();
+        assert mapper.simpleName() == mapper.simpleName().intern();
+        if (mapper instanceof ObjectMapper) {
+            ((ObjectMapper) mapper).mappers.forEach(MappingLookup::assertNamesInterned);
+        }
```
javanna, line 204: thanks a lot for working on this. Maybe silly question, but could this be instead a new test method added to MapperTestCase instead of an assertion?

javanna, line 204: thanks a lot for working on this. Maybe silly question, but could this be instead a new test method added to MapperTestCase instead of an assertion?

original-brownbear: I figured that a test will always only cover a portion of our parsing (and wouldn't have caught the alias field case fixed here for example) while an assertion gives us full coverage?

javanna: @romseygeek do you have opinions here? Do you mind giving this a quick look?

romseygeek: I think an assertion is fine, all of the MapperTestCase tests will build a MappingLookup so this assertion will trip there as well.

javanna (approved):
Thanks for your patience @original-brownbear

original-brownbear:
Thanks Luca!

PR #87239, opened 2022-05-31
Broadcast replication action support build response from shard level response

rhluo:
The broadcast replication action has been widely used in existing consistency tasks, suck as bulk , global checkpoint sync. However current broadcast replication action not support use shard level response to build broadcast response. see in org.elasticsearch.action.support.replication.TransportBroadcastReplicationAction#newResponse.
Additionally, the broadcast by node action can support use shard level response to build broadresponse. see in org.elasticsearch.action.support.broadcast.node.TransportBroadcastByNodeAction#newResponse.

This  feature will be helpful for additional consistency tasks develop such as decoupling storage and compute.


DaveCTurner:
Thanks for the suggestion @rhluo. This commit just adds an unused parameter to this method signature so I don't think it's something we'd want to merge as-is. Instead, we'd accept this as part of a bigger change that also uses the parameter you're adding.

As such, I hope it's ok if I close this.

PR #87247, opened 2022-05-31
Fix ClusterHealthIT

idegtiarenko:
This change treats no desired nodes as `NO_ATTEMPT` and expects a followup reroute to allocate the shard.

### DIFF
--- a/server/src/internalClusterTest/java/org/elasticsearch/cluster/ClusterHealthIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/cluster/ClusterHealthIT.java
@@ -258,18 +258,14 @@ public class ClusterHealthIT extends ESIntegTestCase {
 
     public void testHealthOnIndexCreation() throws Exception {
         final AtomicBoolean finished = new AtomicBoolean(false);
-        Thread clusterHealthThread = new Thread() {
-            @Override
-            public void run() {
-                while (finished.get() == false) {
-                    ClusterHealthResponse health = client().admin().cluster().prepareHealth().get();
-                    assertThat(health.getStatus(), not(equalTo(ClusterHealthStatus.RED)));
-                }
+        Thread clusterHealthThread = new Thread(() -> {
+            while (finished.get() == false) {
+                assertThat(client().admin().cluster().prepareHealth().get().getStatus(), not(equalTo(ClusterHealthStatus.RED)));
             }
-        };
+        });
         clusterHealthThread.start();
         for (int i = 0; i < 10; i++) {
-            createIndex("test" + i);
+            createIndex("test-" + i);
         }
         finished.set(true);
         clusterHealthThread.join();
--- a/server/src/internalClusterTest/java/org/elasticsearch/cluster/ClusterHealthIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/cluster/ClusterHealthIT.java
@@ -258,14 +258,18 @@ public class ClusterHealthIT extends ESIntegTestCase {
 
     public void testHealthOnIndexCreation() throws Exception {
         final AtomicBoolean finished = new AtomicBoolean(false);
-        Thread clusterHealthThread = new Thread(() -> {
-            while (finished.get() == false) {
-                assertThat(client().admin().cluster().prepareHealth().get().getStatus(), not(equalTo(ClusterHealthStatus.RED)));
+        Thread clusterHealthThread = new Thread() {
+            @Override
+            public void run() {
+                while (finished.get() == false) {
+                    ClusterHealthResponse health = client().admin().cluster().prepareHealth().get();
+                    assertThat(health.getStatus(), not(equalTo(ClusterHealthStatus.RED)));
+                }
             }
-        });
+        };
         clusterHealthThread.start();
         for (int i = 0; i < 10; i++) {
-            createIndex("test-" + i);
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceReconciler.java
```diff
@@ -228,8 +228,9 @@ private void allocateUnassigned() {
                 }
 
                 final var allocationStatus = UnassignedInfo.AllocationStatus.fromDecision(
-                    isThrottled ? Decision.Type.THROTTLE : Decision.Type.NO
+                    isThrottled || desiredNodeIds.isEmpty() ? Decision.Type.THROTTLE : Decision.Type.NO
```
idegtiarenko, line 231: Just realized that this breaks `ClusterAllocationExplainIT#testAllocationFilteringOnIndexCreation` with:
```
expected:<DECIDERS_NO> but was:<DECIDERS_THROTTLED>
Expected :DECIDERS_NO
Actual   :DECIDERS_THROTTLED
```

So we need to distinguish between `desiredNodeIds` not computed / computed but empty. I am working on that.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceReconciler.java
```diff
@@ -228,8 +228,9 @@ private void allocateUnassigned() {
                 }
 
                 final var allocationStatus = UnassignedInfo.AllocationStatus.fromDecision(
-                    isThrottled ? Decision.Type.THROTTLE : Decision.Type.NO
+                    isThrottled || desiredBalance.isBalanceComputed(shard.shardId()) == false ? Decision.Type.THROTTLE : Decision.Type.NO
```
DaveCTurner, line 231: Hmm if we haven't computed the balance for a shard yet I would expect the right status to be `AllocationStatus.NO_ATTEMPT`. Does that not work?

server/src/internalClusterTest/java/org/elasticsearch/cluster/ClusterHealthIT.java
```diff
@@ -258,18 +258,14 @@ public void testHealthWithClosedIndices() {
...
     public void testHealthOnIndexCreation() throws Exception {
         final AtomicBoolean finished = new AtomicBoolean(false);
-        Thread clusterHealthThread = new Thread() {
-            @Override
-            public void run() {
-                while (finished.get() == false) {
-                    ClusterHealthResponse health = client().admin().cluster().prepareHealth().get();
-                    assertThat(health.getStatus(), not(equalTo(ClusterHealthStatus.RED)));
-                }
+        Thread clusterHealthThread = new Thread(() -> {
```
DaveCTurner, line 261: Are these changes meaningful or just cosmetic? If they're not important here then it'd be best to keep them off the feature branch. It'd be fine to make them in `master` tho.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceService.java
```diff
@@ -239,7 +239,7 @@ boolean updateDesiredBalanceAndReroute(DesiredBalanceInput desiredBalanceInput,
         }
 
         final var unassigned = new HashMap<ShardId, Integer>();
-        for (var ignored : routingAllocation.routingNodes().unassigned().ignored()) {
+        for (var ignored : routingNodes.unassigned().ignored()) {
```
DaveCTurner, line 242: I think we use `routingAllocation.routingNodes()` in other places above too - may as well fix them all?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceReconciler.java
idegtiarenko: Unfortunately that still produces RED status

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceReconciler.java
```diff
@@ -227,9 +227,15 @@ private void allocateUnassigned() {
                     logger.trace("No eligible node found to assign shard [{}] amongst [{}]", shard, desiredNodeIds);
                 }
 
-                final var allocationStatus = UnassignedInfo.AllocationStatus.fromDecision(
-                    isThrottled ? Decision.Type.THROTTLE : Decision.Type.NO
-                );
+                final UnassignedInfo.AllocationStatus allocationStatus;
+                if (isThrottled) {
+                    allocationStatus = UnassignedInfo.AllocationStatus.DECIDERS_THROTTLED;
+                } else if (desiredBalance.isBalanceComputed(shard.shardId()) == false) {
```
DaveCTurner, line 233: Kind of a nit but I'm wondering why this check isn't first. I think this would override the throttled state. Can we be throttled without having the balance computed? That seems strange.

idegtiarenko: Good point

DaveCTurner (approved):
LGT<

PR #87248, opened 2022-05-31
Do not retry client requests when failing with ContentTooLargeException

### LINKED ISSUE #86041: Low level rest client retrying scrolls, skipping pages
### Elasticsearch Version

Doesn't matter

### Installed Plugins

_No response_

### Java Version

all

### OS Version

doesn't matter

### Problem Description

When you send a `scroll` request to the low level rest client it uses it's standard retry mechanism which catches all errors and just retries against another ES instance. The trouble with that is that you *can't* retry a scroll - it generally consumes state on the server and moves to the next page.

### Steps to Reproduce

Scroll with low level rest client.

### Logs (if relevant)

_No response_

swallez:
The default configuration of the Low Level Rest Client is to buffer the response body up to 100 MB. If that threshold is exceeded, a `ContentTooLargeException` is thrown in the lower layers of LLRC.

When such an exception occurs, LLRC will retry the request, which has some bad side effects:
- for idempotent requests like a regular search, there's no point in retrying since the response will always be too large.
- for scroll searches, retrying will move to the next scroll slice. With a cluster big enough or a number of scroll slices small enough, retrying may reach the last scroll slice, which is usually be smaller and may fit into the buffer limit. From an application perspective, the request will have succeeded with the last slice as result, despite having effectively skipped a large amount of results.

This PR fixes that, by ensuring that no retry happens if a request fails with a `ContentTooLargeException`.

Fixes #86041.

I added the `v6.8.24` label to backport to 6.8 as this was found by a customer still using 6.8 and even if there's no additional 6.8 release, they may want to use a snapshot version with the fix.

### DIFF
--- a/client/rest/src/main/java/org/elasticsearch/client/RestClient.java
+++ b/client/rest/src/main/java/org/elasticsearch/client/RestClient.java
@@ -21,6 +21,7 @@ package org.elasticsearch.client;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.http.ConnectionClosedException;
+import org.apache.http.ContentTooLongException;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHost;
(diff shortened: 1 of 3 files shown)


client/rest/src/main/java/org/elasticsearch/client/RestClient.java
```diff
@@ -563,6 +564,19 @@ private static boolean isSuccessfulResponse(int statusCode) {
         return statusCode < 300;
     }
 
+    /**
+     * Should an exception cause retrying the request?
+     */
+    private static boolean isRetryableException(Throwable e) {
+        if (e instanceof ExecutionException) {
+            e = e.getCause();
+        }
```
nik9000, line 573: I was going to say it'd be better to use the `try` in the caller, but seeing this thing, I get why you didn't.

nik9000 (approved):

PR #87254, opened 2022-05-31
Fork after calling getRepositoryData from StoreRecovery

### LINKED ISSUE #87237: StoreRecovery#restore may invoke Repository#restoreShard on transport thread
This code involves `getRepositoryData` which may complete its listener on a transport worker:

https://github.com/elastic/elasticsearch/blob/898d84998b081d47ec6a9c2305f9bf4486d69be3/server/src/main/java/org/elasticsearch/index/shard/StoreRecovery.java#L528-L547

However the listener in question calls `Repository#restoreShard` which involves blocking calls and should not run on a transport worker.

DaveCTurner:
`Repository#getRepositoryData` will occasionally complete its listener
on a blocking-free thread (a transport worker or the cluster applier
thread) so it's not valid to call `Repository#restoreShard` directly.
This commit fixes this by using a forking listener.

Closes #87237

### DIFF
--- a/docs/changelog/87254.yaml
+++ b/docs/changelog/87254.yaml
@@ -1,0 +1,6 @@
+pr: 87254
+summary: Fork after calling `getRepositoryData` from `StoreRecovery`
+area: Snapshot/Restore
+type: bug
+issues:
+ - 87237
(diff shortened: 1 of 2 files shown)


original-brownbear (approved):
LGTM nice find!

DaveCTurner:
This was reverted, trying again in #87264.

PR #87261, opened 2022-05-31
Synthetic source: tests for disabling subobjects

nik9000:
This adds some paranoid tests for synthetic source with disabling
subobjects, as added by #86166. It turns out that synthetic source does
exactly what you'd expect with disabling subobjects - it creates fields
with dots in their names. This adds tests for that.


javanna (approved):
LGTM thanks!

nik9000:
> LGTM thanks!

I was super happy that it "just works". I didn't realize `subobjects` was a concept that only lived at the auto-mapping level.

Issue #86801, opened 2022-05-16
Upgrade to 8.0: Reindexing `.security` index fails if `security-index-template-v6` is still present


JohannesMahne:
### Elasticsearch Version

7.17

### Java Version

_bundled_

### OS Version

Any

### Problem Description

# Symptoms

This problem can occur during the running of the `7.17` Upgrade Assistant `Migrate system indices` action:
- If `System indices migration failed. An error occurred while migrating system indices for security: exception` error is reported.
- The cluster logs will report `"reason": "mapping set to strict, dynamic introduction of [allow_restricted_indices] within [indices] is not allowed"`, and other similar messages for the new `.security-7-reindexed-for-8` index.
- The same error will also be reported in the output of `GET /_migration/system_features`.
- There will also be a legacy index template (e.g. `security-index-template-v6`), that matches the `.security-*` index pattern.

# Fix/Workaround

The following should be done to fix the migration issue:
1. [Delete the index template](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-delete-template.html) `security-index-template-v6`.
2. [Delete](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-delete-index.html) partially reindexed `.security-7-reindexed-for-8` index.
3. Rerun the `Migrate system indices` action form the Upgrade Assistant.

### Steps to Reproduce

The required reindexing step of the `.security-7` index may fail, if older index templates matching the `.security-*` index pattern are still present. This may occur on clusters that were originally created on version 5.x.

albertzaharovits:
I was able to reproduce this. I have tried migrating a Security index created in version 6.

If there is a template (I've only tested a legacy templates, but I think the composite templates behave the same), and the `.security*` index needs to be migrated, the settings and mappings from the template will be merged in the resulting `.security-6-reindexed-for-8`. This is a problem because most likely Security will become unusable at this point (hence the cluster will be inaccessible).

I don't think we should apply templates when we migrate the .security index.
We don't apply them when the .security index is created automatically, and we also do not apply them if the user tries to manually create it.
What's your take on this @elastic/es-data-management ?

williamrandolph:
The unexpected effects of templates on important internal indices were a big motivation for creating system indices in the first place. Unfortunately, we didn't manage to get to #42508 before the 8.0 release, so now anything we do needs to be backwards-compatible. But it's a big problem and I need to work on a solution for it.

[...]

PR #87262, opened 2022-05-31
Don't migrate system indices that match user templates

williamrandolph:
In 7.x, legacy and composable templates may apply to system indices. We don't want this behavior during feature migration. The destination index should get its settings, mappings, and aliases from its system index descriptor. A user who gets this warning should modify their templates so that they do not match system indices, at least during the migration.

Fixes https://github.com/elastic/elasticsearch/issues/86801

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/metadata/IndexTemplateMetadata.java
+++ b/server/src/main/java/org/elasticsearch/cluster/metadata/IndexTemplateMetadata.java
@@ -162,6 +162,11 @@ public class IndexTemplateMetadata extends AbstractDiffable<IndexTemplateMetadat
     }
 
     @Override
+    public String toString() {
+        return name;
+    }
+
+    @Override
     public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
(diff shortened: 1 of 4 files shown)


williamrandolph:

looks like it was some kind of timeout on an unrelated test

server/src/main/java/org/elasticsearch/cluster/metadata/IndexTemplateMetadata.java
```diff
@@ -161,6 +161,11 @@ public static Builder builder(String name) {
         return new Builder(name);
     }
 
+    @Override
+    public String toString() {
```
grcevski, line 165: Convenience method so that the errors we return actual contain the template name we failed on.

grcevski:
I think this PR needs to check for unmanaged indices and not block the migration, since the Kibana security-session template will definitely be flagged as a problem template https://github.com/elastic/kibana/issues/134897

server/src/main/java/org/elasticsearch/upgrades/SystemIndexMigrator.java
```diff
@@ -381,6 +383,41 @@ private void migrateSingleIndex(ClusterState clusterState, Consumer<BulkByScroll
         }
         Index oldIndex = imd.getIndex();
         String newIndexName = migrationInfo.getNextIndexName();
+
+        final String v2template = MetadataIndexTemplateService.findV2Template(clusterState.metadata(), newIndexName, false);
```
gwbrown, line 387: I thought v2 templates already don't match system indices, and the problem was only caused by legacy templates - I'll try to double check this, but I'll also ask: Am I just straight-up wrong about that?

grcevski: I think both apply, is this legacy or not template usage from kibana: https://github.com/elastic/kibana/blob/3730dd0779ed8efd74aee88f57422781ec1ac122/x-pack/plugins/security/server/session_management/session_index.ts#L61-L92

grcevski:
Superseded by https://github.com/elastic/elasticsearch/pull/87979.

PR #85348, opened 2022-03-24
Support fields option on source-less indices

nik9000:
This adds support for fetching `fields` from doc values for a few field
types:
* long
* int
* short
* byte
* double
* short
* date
* boolean
* ip
* keyword
* unsigned long
* version string
* wildcard
* scaled float

We've had support for fetching from doc values forever using the
`docvalue_fields` option and this just plugs into that when the source
is disabled. It's possible that it's faster to plug into it in other
cases but we don't know those cases yet.

It does that by creating a `ValueFetcherSource` which is a place where
we can attach additional fetch machinery. At this point it mostly just
has the machinery for the `fields` fetch API and highlighters. But we
believe we'll want this spot for TSDB's synthetic source project. And so
we can make the fields API more efficient when fetching small numbers of
fields later.

PR #85185, opened 2022-03-21
Support `fields` option on source-less indices

nik9000:
This adds support for fetching `fields` from doc values for a few field
types:
* long
* int
* short
* byte
* double
* short
* date
* boolean
* ip
* keyword
* unsigned long
* version string
* wildcard
* scaled float

We've had support for fetching from doc values forever using the
`docvalue_fields` option and this just plugs into that when the source
is disabled. It's possible that it's faster to plug into it in other
cases but we don't know those cases yet.

PR #87267, opened 2022-05-31
Fields API to allow fetching values when _source is disabled

### LINKED ISSUE #87072: Runtime fields and _source disabled
### Elasticsearch Version

8.2

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

Ubuntu 20.04.4 LTS

### Problem Description

Currently when using runtime fields it is not possible to search using `fields` when the index has `_source` disabled.

It will fail with `Unable to retrieve the requested [fields] since _source is disabled in the mappings for index [example-disabled]` even if `_source` isn't being used in a runtime field. Steps below.

Ideally this would fail only if the runtime field referenced `_source` fields.

Usecase is that certain logging platforms configure Elasticsearch indices with `_source` disabled (even though this is not recommended) and `store: true` for fields. Having more flexibility in using runtime fields in this instance would make migrations easier.

### Steps to Reproduce

Create indices with `_source` enabled or disabled.

```
PUT example-enabled
{
  "mappings": {
    "_source": {
      "enabled": true
    }, 
    "runtime": {
      "age": {
        "type": "long",
        "script": {
          "source": "emit(1)"
        }
      }
    }
  }
}

PUT example-disabled
{
  "mappings": {
    "_source": {
      "enabled": false
    }, 
    "runtime": {
      "age": {
        "type": "long",
        "script": {
          "source": "emit(1)"
        }
      }
    }
  }
}

PUT example-disabled/_doc/1
{}

PUT example-enabled/_doc/1
{}
```

Search over both indices:

```
GET example-*/_search
{
  "fields": [
    "age"
  ]
}
```

Response

```
{
  "took" : 459,
  "timed_out" : false,
  "_shards" : {
    "total" : 2,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 1,
    "failures" : [
      {
        "shard" : 0,
        "index" : "example-disabled",
        "node" : "fNuQNb81QsC6Yb0jAnUATw",
        "reason" : {
          "type" : "illegal_argument_exception",
          "reason" : "Unable to retrieve the requested [fields] since _source is disabled in the mappings for index [example-disabled]"
        }
      }
    ]
  },
  "hits" : {
    "total" : {
      "value" : 2,
      "relation" : "eq"
    },
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : "example-enabled",
        "_id" : "1",
        "_score" : 1.0,
        "_source" : { },
        "fields" : {
          "age" : [
            1
          ]
        }
      }
    ]
  }
}
```

### Logs (if relevant)

_No response_

javanna:
Back when we introduced the fields parameter to the search API, it could only fetch values from _source, hence
the corresponding sub-fetch phase fails early whenever _source is disabled. Today though runtime fields can
be retrieved from a separate value fetcher that reads from fielddata, and metadata fields can be retrieved
from stored fields. These two scenarios currently throw an unnecessary error whenever _source is disabled.

This commit removes the check for disabled _source, so that runtime fields and metadata fields can be retrieved even when _source is disabled. Fields that need to be loaded from _source are simply skipped whenever _source is disabled, similar to when a field is not found in _source.

Closes #87072


nik9000 (approved):
LGTM.  Feels a lot like the work I started on in #85348 and #85185 but that stalled out. I'm happy to see this in though.

server/src/main/java/org/elasticsearch/search/fetch/subphase/FetchFieldsPhase.java
```diff
@@ -31,15 +32,6 @@ public FetchSubPhaseProcessor getProcessor(FetchContext fetchContext) {
...
 
-        if (fetchContext.getSearchExecutionContext().isSourceEnabled() == false) {
-            throw new IllegalArgumentException(
-                "Unable to retrieve the requested [fields] since _source is disabled "
-                    + "in the mappings for index ["
-                    + fetchContext.getIndexName()
-                    + "]"
-            );
-        }
-
```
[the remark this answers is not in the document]
nik9000: Yeah, it misuses it. I'm mostly ok with it because all they really want is the doc id. But there is a lot of baggage here. A cleaner interface could be much nicer.

javanna:
> Feels a lot like the work I started on in https://github.com/elastic/elasticsearch/pull/85348 and https://github.com/elastic/elasticsearch/pull/85185 but that stalled out. I'm happy to see this in though.

Yes, I did verify that you also had a similar fix in your PRs. It feels good that this goes in separately, and I will revive discussions on fetching from doc_values and possibly from stored fields too.

sethmlarson (approved):
LGTM from an API perspective

server/src/main/java/org/elasticsearch/search/fetch/subphase/FetchFieldsPhase.java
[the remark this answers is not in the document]
jtibshirani: It'd be great to clean up this interface a bit. I don't think we had concrete ideas, would be happy to discuss in a follow-up.

jtibshirani (commented):
Thanks for fixing this issue!

Stepping back, I'm not sure how helpful this check is in practice. For example it doesn't cover the case where a field has been excluded from _source (but _source is not entirely disabled). Maybe it could also be annoying if you specify `"fields": "*"` (as an analogy to `"_source": true`), just wanting all fields possible, but then we fail the request?

javanna:
> I'm not sure how helpful this check is in practice

I think I agree. It would be better to return what we can retrieve instead of failing the whole request, yet silently failing to fetch does not sound like a good idea either, or should we consider doing that?

jtibshirani:
Right, I was suggesting that we remove the check altogether. I think it's already documented that if you don't have `_source` enabled (on non-runtime fields) then the fields will be missing from the response. I can see arguments both ways, but to me this approach feels the simplest and most flexible.

javanna:
@jtibshirani I removed the check for _source disabled, and fixed failing tests here and there. Could you have another look please?

server/src/main/java/org/elasticsearch/index/mapper/ArraySourceValueFetcher.java
```diff
@@ -38,7 +39,7 @@ public ArraySourceValueFetcher(String fieldName, SearchExecutionContext context)
      * @param nullValue A optional substitute value if the _source value is 'null'.
      */
     public ArraySourceValueFetcher(String fieldName, SearchExecutionContext context, Object nullValue) {
-        this.sourcePaths = context.sourcePath(fieldName);
+        this.sourcePaths = context.isSourceEnabled() ? context.sourcePath(fieldName) : Collections.emptySet();
```
jtibshirani, line 42: Could you explain the motivation for these checks `context.isSourceEnabled()`? Is it a small optimization to avoid extra work?

I saw that `FieldFetcherTests#testFetchFromSourceWithSourceDisabled` can fail without this check, but it seems like the test logic just needs to be adapted to not pass a null or empty `SourceLookup`?

jtibshirani (commented):
Thanks! I left one last comment.

server/src/main/java/org/elasticsearch/index/mapper/ArraySourceValueFetcher.java
javanna: yes the check helps the new test work without providing a source lookup, but mostly I was annoyed that we do extra work when we can avoid it: why call sourcePath and iterate over what it returns when we already know that we have no _source?> And why even require a non-null SourceLookup in that case? Also, the way the test is currently setup, if I provide a non null _source, fields are going to be retrieved from it no matter whether source is disabled or not, because source is not retrieved from the index but rather provided manually to the fetch method.

I guess I took a shortcut to avoid having to deal with all of the above in this PR but I'd expect things to change once SearchLookup is no longer the only required argument. With this in mind this change did not feel so bad after all.

jtibshirani (approved):

server/src/main/java/org/elasticsearch/index/mapper/ArraySourceValueFetcher.java
javanna: ++ thanks I will keep this in mind

PR #87269, opened 2022-05-31
TSDB: Implement Downsampling ILM Action for time-series indices

### LINKED ISSUE #68609: [CI] RollupActionIT testRollupIndexAndSetNewRollupPolicy failing
This is a new test case which has failed a few times now since being added.

**Build scan**:
https://gradle-enterprise.elastic.co/s/nckptifuyp4v2/tests/:x-pack:plugin:ilm:qa:multi-node:javaRestTest/org.elasticsearch.xpack.ilm.actions.RollupActionIT/testRollupIndexAndSetNewRollupPolicy#1

**Repro line**:
`./gradlew ':x-pack:plugin:ilm:qa:multi-node:javaRestTest' --tests "org.elasticsearch.xpack.ilm.actions.RollupActionIT.testRollupIndexAndSetNewRollupPolicy" -Dtests.seed=F0A4BF22F8036B7E -Dtests.security.manager=true -Dtests.locale=sk-SK -Dtests.timezone=Pacific/Truk -Druntime.java=11`

**Reproduces locally?**:
Nope.

**Applicable branches**:
`master` and `7.x` (error on the `7.x` build looked a bit different though)

**Failure history**:
https://gradle-enterprise.elastic.co/scans/tests?search.relativeStartTime=P7D&search.timeZoneId=America/Los_Angeles&tests.container=org.elasticsearch.xpack.ilm.actions.RollupActionIT&tests.sortField=FAILED&tests.unstableOnly=true

**Failure excerpt**:
```
org.elasticsearch.xpack.ilm.actions.RollupActionIT > testRollupIndexAndSetNewRollupPolicy FAILED
    java.lang.AssertionError
        at __randomizedtesting.SeedInfo.seed([F0A4BF22F8036B7E:7FF7996A3D79E9A]:0)
```

csoulios:
This PR adds support for an ILM action that downsamples a time-series index by invoking the `_rollup` endpoint (#85708)

A  policy that includes the rollup action will look like the following

```json
PUT _ilm/policy/my_policy
{
  "policy": {
    "phases": {
      "warm": {
        "actions": {
          "rollup": {
  	    "fixed_interval": "24h"
  	  }
  	}
      }
    }
  }
}
```

Fixes https://github.com/elastic/elasticsearch/issues/68609

Finally, this PR is based on the code written for https://github.com/elastic/elasticsearch/pull/68184


x-pack/plugin/ilm/qa/multi-node/src/javaRestTest/java/org/elasticsearch/xpack/ilm/actions/RollupActionIT.java
```diff
@@ -7,82 +7,236 @@
...
+        String rollupIndex = getRollupIndexName(originalIndex);
 
-        assertBusy(() -> assertNotNull(getRollupIndexName(index)));
-        String rollupIndex = getRollupIndexName(index);
-        assertBusy(() -> assertTrue(indexExists(rollupIndex)));
-        assertBusy(() -> assertThat(getOnlyIndexSettings(client(), rollupIndex).get(LifecycleSettings.LIFECYCLE_NAME), equalTo(policy)));
-        assertBusy(() -> assertTrue(indexExists(index)));
+        assertBusy(() -> assertTrue("Rollup index does not exist", indexExists(rollupIndex)), 30, TimeUnit.SECONDS);
+        assertBusy(() -> assertFalse("Source index should have been deleted", indexExists(originalIndex)), 30, TimeUnit.SECONDS);
+        // assertBusy(() -> assertThat(getStepKeyForIndex(client(), rollupIndex), equalTo(PhaseCompleteStep.finalStep("hot").getKey())));
```
csoulios, line 205: Not sure if we should test this too. Right now RollupILMAction does not set ILM phase in the index. should this be done?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/RollupILMAction.java
```diff
@@ -104,31 +91,21 @@ public boolean isSafeAction() {
...
-            Step rollupStep = new RollupStep(rollupKey, updateRollupIndexPolicyStepKey, client, config);
-            Step updateRollupIndexPolicyStep = new UpdateRollupIndexPolicyStep(
-                updateRollupIndexPolicyStepKey,
-                nextStepKey,
-                client,
-                rollupPolicy
-            );
-            return List.of(checkNotWriteIndexStep, readOnlyStep, generateRollupIndexNameStep, rollupStep, updateRollupIndexPolicyStep);
-        }
+        Step rollupStep = new RollupStep(rollupKey, nextStepKey, client, config);
```
csoulios, line 107: I wonder if there are more steps, that should be added before or after the rollup step.

dakrone: How will this work with TSDB data streams writing to indices *after* they have been rolled over? I can imagine that we'd want to wait until the index is out of the "valid" time range before doing the downsample, correct? Otherwise we risk documents getting rejected because the index is marked as read-only.

dakrone: Also, it doesn't appear that ILM re-adds the new index to the abstraction that it is currently in (the alias or data stream). Does the rollup action automatically do that, or do we need to add a step to do that somewhere?

dakrone (changes requested):
How will this work with an ILM policy? I looked at the implementation in `TransportRollupAction`, but it doesn't look like all of the metadata pieces get copied over. Right now that would mean that the down-sampling would happen but the new index is subject to templates for its settings.

That means that the new index will need to have a separate template with a separate policy (since no `rollover` is needed), and that the new index will essentially "restart" its ILM lifecycle. This seems a little complicated to try and understand for an end user. I would imagine the simpler behavior is to pretend that rolling up the index is transparent and happens to the existing index (kind of like `shrink`) and then it goes its merry way along the rest of its lifecycle.

In the old version we explicitly call out a new policy for the index to follow (and I think we all agreed that was very complicated), this is still kind of that, but even more confusing to a user since they'll need to set up a template beforehand.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/RollupILMAction.java
csoulios: > How will this work with TSDB data streams writing to indices after they have been rolled over? I can imagine that we'd want to wait until the index is out of the "valid" time range before doing the downsample, correct? Otherwise we risk documents getting rejected because the index is marked as read-only.

We leave this up to the user to configure the ILM policy in a way that there is enough time for late arrivals. After the index has been downsampled, it becomes read-only and no delayed entries are allowed. 


> Also, it doesn't appear that ILM re-adds the new index to the abstraction that it is currently in (the alias or data stream). Does the rollup action automatically do that, or do we need to add a step to do that somewhere?

If the source index is a backing index of a data stream, it is replaced by the rollup index. This happens at the [TransportRollupAction](https://github.com/elastic/elasticsearch/blob/f64c75dca608760c0470896316c3aa81ade93845/x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java#L567-L571). We chose this approach instead of doing so at the ILM action, because we wanted all the rollup process to be available as an API endpoint that can be invoked by the user.

As for an alias, we assume that the index is read-only before being downsampled. So, there should not be any write alias pointing to this index. Do you think we should cover the case of a read alias?

csoulios:
@dakrone thanks  a lot for the feedback.

> How will this work with an ILM policy? I looked at the implementation in TransportRollupAction, but it doesn't look like all of the metadata pieces get copied over. Right now that would mean that the down-sampling would happen but the new index is subject to templates for its settings.

Source indices are always deleted at the end of the rollup action and they are replaced by the rollup index. When an index belongs to a data stream, the souce index is replaced by the rollup index in the data stream too. This all happens at the [`TransportRollupAction`](https://github.com/elastic/elasticsearch/blob/f64c75dca608760c0470896316c3aa81ade93845/x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java#L567-L584).

You are right that not all the metadata pieces get copied over. We only copy the metadata required for the rollup process. 

Do you think that we should copy all metadata or copying only `index.lifecycle.name` is enough? 

>That means that the new index will need to have a separate template with a separate policy (since no rollover is needed), and that the new index will essentially "restart" its ILM lifecycle. This seems a little complicated to try and understand for an end user. I would imagine the simpler behavior is to pretend that rolling up the index is transparent and happens to the existing index (kind of like shrink) and then it goes its merry way along the rest of its lifecycle.

I absolutely agree with you. So, the rollup index should inherit the ILM policy of the source index (like a shrinked index inherits the ILM policy of the source index).

> In the old version we explicitly call out a new policy for the index to follow (and I think we all agreed that was very complicated), this is still kind of that, but even more confusing to a user since they'll need to set up a template beforehand.

That's correct. We want to avoid all complications of a separate ILM policy. This is why I removed the `rollup_policy` configuration param from the `rollup` action.

Finally, I was hoping I could copy the ILM related metadata in a different ILM step. However, this seems to be impossible since the source index has been deleted. Is there any way we can read index metadata before the `RollupStep` and somehow apply them after the `RollupStep` has completed? If that's not possible, I will probably need to copy metadata in the `RollupTransportAction`

dakrone (commented):
> Source indices are always deleted at the end of the rollup action and they are replaced by the rollup index. When an index belongs to a data stream, the souce index is replaced by the rollup index in the data stream too. This all happens at the TransportRollupAction.

> You are right that not all the metadata pieces get copied over. We only copy the metadata required for the rollup process.

> Do you think that we should copy all metadata or copying only index.lifecycle.name is enough?

I think we will need to copy the metadata. We already have a precedent for this with `shrink` as well as the `searchable_snapshot` ILM actions. For example, with an index called "myindex" the shrink action creates a new shrunken index called "shrink-myindex", then a separate ILM step copies the necessary metadata. After that, the index abstraction (either a data stream or alias) is updated to atomically "swap" the index in, and the old index is deleted.

I know that having duplicate data was originally a concern (due to wildcards used by the UI for example), but this is one of the reasons we *prepend* the modifier (`shrink-` for shrink, `restored-` for mounted, or `partial-` for partially mounted) to the index name. We haven't had anyone run into issues with our prepending and seeing duplicate data at all.

> I absolutely agree with you. So, the rollup index should inherit the ILM policy of the source index (like a shrinked index inherits the ILM policy of the source index).

> Finally, I was hoping I could copy the ILM related metadata in a different ILM step. However, this seems to be impossible since the source index has been deleted. Is there any way we can read index metadata before the RollupStep and somehow apply them after the RollupStep has completed? If that's not possible, I will probably need to copy metadata in the RollupTransportAction

I think this is different than our other ILM actions. We already have a pretty strong set of existing behavior for actions like this. As an example, here is the way that I would implement this:

- check for write index step (currently does this)
- wait for no followers (currently does this)
- set the index to read-only (currently does this)
- generate a unique index name with the `rollup-` prefix (currently does this)
- calls the rollup API
  - the rollup API creates the new index and does the required reindex, but *does not* delete the source index (or else ILM passes in a flag that keeps the source index around)
- the original index copies the execution state necessary from the source to the new index (we have a step for this already, `CopyExecutionStateStep`)
- do the index abstraction swap-and-delete (it's one step for an alias and two steps for a data stream, see `ReplaceDataStreamBackingIndexStep`)

This allows us to treat the new index exactly like the old index, and still have an atomic swap for the data stream. It also means that the rollup logic doesn't have to know about any of the ILM internals to make sure to copy them correctly (ILM does it itself).

I also notice that we don't copy all the settings from the original index to the downsampled index, which I think may end up biting us in the future also (in particular, I'm thinking of what would happen if someone set `index.lifecycle.origination_date` or `index.lifecycle.parse_origination_date` on the source index but then it was not copied over to the downsampled index).

Hopefully that helps, let me know if I made things confusing :)

martijnvg:
> Well, I think it's okay to not support aliases directly, considering that an index will already be in TSDB-mode anyway, and that requires a data stream. What I am concerned about is if a user manually adds an index into a set of aliases, then when it is downsampled it gets automatically removed from those aliases. If the Rollup API is going to treat the downsample as a 1-to-1 replacement, it seems like it should also copy over the alias configuration so that it matches the previous index exactly.

I think this is not a problem? This feature only works with data streams and it is not allowed to create an alias that refers to a backing index of a data stream. A validation error should occur when such alias is attempted to be created. So I think we don't need to copy over the alias metadata.

csoulios:
@dakrone 

My initial intention was to make the `/<index/_rollup/<rollup-index>` action perform all required steps to rollup the index, delete the source and update the data stream. We (the team) have discussed your approach again in the past. However, thinking of it again, I start to like this solution for the following reasons:

- It keeps rollup behavior consistent with the shrink action
- We had a feeling that `TransportRollupAction` does too much and may have trouble with random failures. ILM seems to deal with failures better.
- We allow users to rollup an index without deleting the original source. This has been requested more than a few times already.


Let me discuss this approach with the team and I will get back to you

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/CleanupTargetIndexStep.java
```diff
@@ -0,0 +1,129 @@
...
+
+import java.util.Objects;
+import java.util.function.Function;
+
+/**
+ * Deletes the target index created by an operation such as shrink or rollup and
+ * identified the target index name stored in the lifecycle state of the managed
+ * index (if any was generated)
+ */
+public class CleanupTargetIndexStep extends AsyncRetryDuringSnapshotActionStep {
```
csoulios, line 29: I can modify `ShrinkAction` so that it uses `CleanupTargetIndexStep` instead of `CleanupShrinkIndexStep` in a subsequent PR

csoulios:
> > I think this is not a problem? This feature only works with data streams and it is not allowed to create an alias that refers to a backing index of a data stream. A validation error should occur when such alias is attempted to be created. So I think we don't need to copy over the alias metadata.
> 
> So, I don't need to add the `SwapAliasesAndDeleteSourceIndexStep` to make sure that aliases to the source index are transfered to the rollup index. Do I?

I had some doubts about this statement: "A validation error should occur when such alias is attempted to be created.", so I tested it and it looks that it's perfectly fine to create a concrete index (not belonging to a data stream) and assign an alias to it. And to be honest, I don't see the reason why why we should not allow it. 

So, in light of those findings, I should add the `SwapAliasesAndDeleteSourceIndexStep` to the list of actions.

I will get back with the change and the code for this.

dakrone (changes requested):
Thanks for the changes Christos! I think this generally looks good to me, I left a few more comments and I think once we handle the rollup failure this will be ready.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/RollupILMAction.java
```diff
@@ -104,31 +102,115 @@ public boolean isSafeAction() {
...
+            nextStepKey
+        );
+
+        // Copy the index.lifecycle.name setting to the rollup index settings
+        CopySettingsStep copySettingsStep = new CopySettingsStep(
+            copyLifecyclePolicySettingKey,
+            dataStreamCheckBranchingKey,
+            (indexName, lifecycleState) -> lifecycleState.rollupIndexName(),
+            LifecycleSettings.LIFECYCLE_NAME
+        );
```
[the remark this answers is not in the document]
csoulios: This step will go away once #88565 get merged in

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/RollupILMAction.java
```diff
@@ -104,31 +102,115 @@ public boolean isSafeAction() {
...
+
+        // Here is where the actual rollup action takes place
+        RollupStep rollupStep = new RollupStep(rollupKey, waitForGreenRollupIndexKey, client, fixedInterval);
+        // Wait until the Rollup index health is green
+        WaitForIndexColorStep waitForGreenIndexHealthStep = new WaitForIndexColorStep(
+            waitForGreenRollupIndexKey,
+            copyMetadataKey,
+            ClusterHealthStatus.GREEN,
+            (indexName, lifecycleState) -> lifecycleState.rollupIndexName()
+        );
```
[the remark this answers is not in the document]
csoulios: It makes sense. I only wonder what would happen if rollup index is recovered (status turns yellow), we delete the source index and then we lose one of the nodes before the rollup index is replicated?
Is this scenario possible? Will it lead to data loss?

csoulios:
@dakrone After merging #88565, I have removed the copy-settings step. 
Also, I have addressed all the other comments you have left. 

Can you please have another look?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/CopySettingsStep.java
```diff
@@ -10,34 +10,41 @@
...
 public class CopySettingsStep extends ClusterStateActionStep {
     public static final String NAME = "copy-settings";
 
     private static final Logger logger = LogManager.getLogger(CopySettingsStep.class);
 
     private final String[] settingsKeys;
-    private final String indexPrefix;
 
-    public CopySettingsStep(StepKey key, StepKey nextStepKey, String indexPrefix, String... settingsKeys) {
+    private final BiFunction<String, LifecycleExecutionState, String> targetIndexNameSupplier;
```
csoulios, line 36: Now that I have removed the copy-settings step from rollup ILM action, I can revert the `CopySettingsStep` so that it `indexPrefix`. 

I only thought that making it more generic using an index name supplier function may be needed in the future. I don't have a strong opinion about keeping it this way or reverting it.

dakrone: I think it's fine to keep for now

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/RollupStep.java
```diff
@@ -45,9 +53,13 @@ public void performAction(
         ClusterStateObserver observer,
         ActionListener<Void> listener
     ) {
+        LifecycleExecutionState lifecycleState = indexMetadata.getLifecycleExecutionState();
+        if (lifecycleState.lifecycleDate() == null) {
+            throw new IllegalStateException("source index [" + indexMetadata.getIndex().getName() + "] is missing lifecycle date");
+        }
```
dakrone, line 59: Out of curiosity, why do we need to ensure that the lifecycle date is set here?

x-pack/plugin/ilm/qa/multi-node/src/javaRestTest/java/org/elasticsearch/xpack/ilm/actions/RollupActionIT.java
```diff
@@ -7,82 +7,259 @@
...
+            .field("type", "double")
+            .field("time_series_metric", "gauge")
+            .endObject()
+            .endObject()
+            .endObject();
+        String mapping = Strings.toString(builder);
+        createIndexWithSettings(client(), index, alias, settings, mapping);
+    }
+
+    public void testRollupIndex() throws Exception {
```
dakrone, line 140: Should these tests also perform a simple query to ensure that they can query the data stream without any problems after the rollup?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/RollupILMAction.java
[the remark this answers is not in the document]
dakrone: > we delete the source index and then we lose one of the nodes before the rollup index is replicated? Is this scenario possible? Will it lead to data loss?

That is possible I think, but waiting for green would mean we are subject to waiting forever for single-node clusters, which is also out I think.

dakrone (approved):
LGTM, I left a few really trivial comments/questions

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/RollupStep.java
csoulios: I don't have a good answer for this. I only copied it from the [`ShrinkStep`](https://github.com/elastic/elasticsearch/blob/be7c7415627377a1b795400fb8dfcc6cbdf0e322/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/ShrinkStep.java#L62-L65) thinking this may be needed later since`ShrinkStep` and `RollupStep` are pretty similar in their operation.

Do you think I should remove this check?

x-pack/plugin/ilm/qa/multi-node/src/javaRestTest/java/org/elasticsearch/xpack/ilm/actions/RollupActionIT.java
csoulios: I added an assertion that the `IndexMetadata.INDEX_ROLLUP_STATUS` index setting is `RollupTaskStatus.SUCCESS` at the end of the process: [7345a57](https://github.com/elastic/elasticsearch/pull/87269/commits/7345a57ff42b461904e0722171298f935fad92f7)

For queries on rollup indexes there is the [`RollupActionSingleNodeTests` class](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/rollup/src/test/java/org/elasticsearch/xpack/rollup/v2/RollupActionSingleNodeTests.java) that validates the result of the rollup process.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/RollupStep.java
dakrone: No I don't think it's necessarily a bad check, I was just curious where it came from.

PR #87273, opened 2022-05-31
Move declarative plugin sync to server cli

rjernst:
When running in Docker, the elasticsearch-plugins.yml allows configuring
plugins that should be installed in the system. Upon Elasticsearch
starting up, plugins are installed/removed to match the configured
plugins. However, this happens late in startup, and it would be nice to
keep the main Elasticsearch process from ever writing outside the
configured data directories. Now that the server cli has been moved to
Java, this is possible.

This commit moves invocation of the plugins sync command into the server
cli. Note that the sync plugins action should probably be reworked as it
can be implement Command directly now. However, this commit tries to be
the minimal change possible to remove plugin cli knowledge from server.


server/src/main/java/org/elasticsearch/bootstrap/plugins/PluginsManager.java
```diff
@@ -1,75 +0,0 @@
...
-     * @throws Exception if anything goes wrong
-     */
-    public static void syncPlugins(Environment env) throws Exception {
-        ClassLoader classLoader = buildClassLoader(env);
-
-        @SuppressWarnings("unchecked")
-        final Class<PluginsSynchronizer> synchronizerClass = (Class<PluginsSynchronizer>) classLoader.loadClass(SYNC_PLUGINS_ACTION);
-
-        final PluginsSynchronizer provider = synchronizerClass.getConstructor(Terminal.class, Environment.class)
-            .newInstance(LoggerTerminal.getLogger(SYNC_PLUGINS_ACTION), env);
```
pugnascotia, line 47: You can remove the `LoggerTerminal` class now too.

pugnascotia (approved):

PR #87229, opened 2022-05-31
Support exists query for API key query

ywangd:
Exists query is useful for testing whether expiration date is configured
for an API key or whether a metadata field exists. This PR adds support
for it.

PR #87275, opened 2022-06-01
[Doc] Add exists to support query types of QueryApiKey API

ywangd:
The exists query type is supported for QueryApiKey API since #87229.
This PR adds it to relevant doc page.

Relates: #87229


x-pack/docs/en/rest-api/security/query-api-key.asciidoc
```diff
@@ -334,3 +336,43 @@ The response contains a list of matched API keys along with their sort values:
...
+      ],
+      "minimum_should_match": 1
+    }
+  }
+}
+----
+// NOTCONSOLE
+
+<1> Matching API keys must not be invalidated
+<2> Matching API keys must be either not expired or does not have an expiration data
```
tvernum, line 378:
```suggestion
<2> Matching API keys must be either not expired or does not have an expiration date
```

tvernum (approved):

PR #87276, opened 2022-06-01
Persist vm.max_map_count for WSL2

kyanha:

### DIFF
--- a/docs/reference/setup/install/docker.asciidoc
+++ b/docs/reference/setup/install/docker.asciidoc
@@ -399,7 +399,29 @@ sudo sysctl -w vm.max_map_count=262144
 
 ====== Windows with https://docs.docker.com/docker-for-windows/wsl[Docker Desktop WSL 2 backend]
 
-The `vm.max_map_count` setting must be set in your %USERPROFILE%\.wslconfig as follows:
+The `vm.max_map_count` setting must be set in the "docker-desktop" WSL instance before the
+ElasticSearch container will properly start. There are several ways to do this, depending
+on your version of Windows and your version of WSL.
+
+If you are on Windows 10 before version 22H2, or if you are on Windows 10 version 22H2 using the
+built-in version of WSL, you must either manually set it every time you restart Docker before starting
+your ElasticSearch container, or (if you do not wish to do so on every restart) you must globally set
+every WSL2 instance to have the `vm.max_map_count` changed. This is because these versions of WSL
+do not properly process the /etc/sysctl.conf file.
+
+To manually set it every time you reboot, you must run the following commands in a command prompt
+or PowerShell window every time you restart Docker:
+
+[source]
+--------------------------------------------
+wsl -d docker-desktop -u root
+sysctl -w vm.max_map_count=262144
+exit
+--------------------------------------------
+
+If you are on these versions of WSL and you do not want to have to run those commands every
+time you restart Docker, you can globally change every WSL distribution with this setting
+by modifying your %USERPROFILE%\.wslconfig as follows:
 
 [source]
 --------------------------------------------
@@ -407,7 +429,23 @@ The `vm.max_map_count` setting must be set in your %USERPROFILE%\.wslconfig as f
 kernelCommandLine = "sysctl.vm.max_map_count=262144"
 --------------------------------------------
 
-This will cause all WSL2 VMs to have that setting assigned at boot time.
+This will cause all WSL2 VMs to have that setting assigned when they start.
+
+If you are on Windows 11, or Windows 10 version 22H2 and have installed the Microsoft Store
+version of WSL, you can modify the /etc/sysctl.conf within the "docker-desktop" WSL
+distribution, perhaps with commands like this:
+
+[source]
+--------------------------------------------
+wsl -d docker-desktop -u root
+vi /etc/sysctl.conf
+--------------------------------------------
+
+and appending a line which reads:
+[source]
+--------------------------------------------
+vm.max_map_count = 262144
+--------------------------------------------
 
 ===== Configuration files must be readable by the `elasticsearch` user
 
--- a/docs/reference/setup/install/docker.asciidoc
+++ b/docs/reference/setup/install/docker.asciidoc
@@ -399,14 +399,15 @@ sudo sysctl -w vm.max_map_count=262144
 
 ====== Windows with https://docs.docker.com/docker-for-windows/wsl[Docker Desktop WSL 2 backend]
 
-The `vm.max_map_count` setting must be set in the docker-desktop container:
+The `vm.max_map_count` setting must be set in your %USERPROFILE%\.wslconfig as follows:
 
-[source,sh]
+[source]
 --------------------------------------------
-wsl -d docker-desktop
-sysctl -w vm.max_map_count=262144
+[wsl2]
+kernelCommandLine = "sysctl.vm.max_map_count=262144"
 --------------------------------------------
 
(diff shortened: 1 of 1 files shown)


kyanha:
I submitted the CLA after creating this pull request, but I have no idea how to update this.

I also don't know who to assign this to, or what labels would apply. It's just a documentation update that makes it so our Docker Desktop-hosted elasticsearch Docker images can automatically restart with Windows 10 and Windows 11 WSL2.

kyanha:
I submitted this when I realized that every time I restarted Windows, I had to manually redo the change to the docker-desktop WSL container. (WSL uses a custom-written init binary, which does not read or process /etc/sysctl.conf.)

While researching it, I came across this question and answer on stackoverflow: https://stackoverflow.com/questions/69214301/using-docker-desktop-for-windows-how-can-sysctl-parameters-be-configured-to-per

I tested it and it works on Windows 10. (Well, 'works', in that the Elasticsearch container no longer aborts on reboot with the message about vm.max_mem_count not being set high enough, which necessitated the manual kernel retuning on every reboot and WSL restart.)

In the spirit of open source, I have submitted this documentation update so that nobody else needs to beat their heads against the same wall I did.

Also, see https://github.com/microsoft/WSL/issues/4232 for the upstream issue about MS's custom-written init not doing any of the systemd processing for sysctl (which itself points to other WSL issues about not actually doing any systemd-like processing, which prevents all the other workarounds we'd use inside Linux from working).

I hope this gives a bit more context, to help evaluate the proposed update.

Thanks for your time!

grcevski:
Hi @kyanha, thanks for this PR, your proposed way to set the WSL setting is much better. However, I can see how certain users would not want to make the setting permanent in their WSL configuration, so they might like to manually set the setting each time.

Can you please change your PR to keep our existing way of configuring WSL, and then add your proposed way to make this a permanent setting. After our existing WSL documentation, something like:

```
If you would like to make the `vm.max_map_count` setting permanent across reboots, you can set the `vm.max_map_count` setting in your %USERPROFILE%\.wslconfig as follows:

 [source]
 --------------------------------------------
 [wsl2]
 kernelCommandLine = "sysctl.vm.max_map_count=262144"
 --------------------------------------------

 This will cause all WSL2 VMs to have that setting assigned at boot time.
```

rjernst:
@kyanha Are you still interested in getting this change in? If so, can you please apply Nikola's suggestion?

kyanha:
> @kyanha Are you still interested in getting this change in? If so, can you please apply Nikola's suggestion?

I can't do that until Microsoft's Patch Tuesday in December, because WSL included in Windows 10 doesn't have the ability to execute anything at WSL machine startup (it doesn't support systemd), and it doesn't have the ability to set different kernel parameters for different machines (it doesn't support /etc/wsl.conf inside the VMs).

However, WSL from the Microsoft Store supports both systemd and /etc/wsl.conf.

Unfortunately, on Windows 10, the version of WSL from the Microsoft Store requires a particular OS update (KB5020030). This is currently in Preview, and on the Retail update channel it's not offered on Windows Update. (It's available from the Update Catalog, but it's still not considered production-ready.)

Patch Tuesday in December 2022 will include this update for the Retail channel.

Until this update lands, there is no actual way on Windows 10 to do what you request. Windows 11, yes, but not Windows 10.

kyanha:
I hope I've managed to address all identified issues with this latest commit. I don't know if there need to be any additional headings?

docs/reference/setup/install/docker.asciidoc
```diff
@@ -399,14 +399,53 @@ sudo sysctl -w vm.max_map_count=262144
...
+
+To manually set it every time you reboot, you must run the following commands in a command prompt
+or PowerShell window every time you restart Docker:
+
+[source]
 --------------------------------------------
-wsl -d docker-desktop
+wsl -d docker-desktop -u root
 sysctl -w vm.max_map_count=262144
+exit
```
grcevski, line 419: Do we need `exit` here?

grcevski (commented):
This looks great, thanks for this write-up. I just had one minor question of adding `exit` to our existing instructions.

grcevski:

docs/reference/setup/install/docker.asciidoc
```diff
@@ -399,14 +399,53 @@ sudo sysctl -w vm.max_map_count=262144
...
+If you are on Windows 10 before version 22H2, or if you are on Windows 10 version 22H2 using the
+built-in version of WSL, you must either manually set it every time you restart Docker before starting
+your ElasticSearch container, or (if you do not wish to do so on every restart) you must globally set
+every WSL2 instance to have the `vm.max_map_count` changed. This is because these versions of WSL
+do not properly process the /etc/sysctl.conf file.
+
+To manually set it every time you reboot, you must run the following commands in a command prompt
+or PowerShell window every time you restart Docker:
+
+[source]
```
grcevski, line 415: This should be `[source,sh]`.

docs/reference/setup/install/docker.asciidoc
```diff
@@ -399,14 +399,53 @@ sudo sysctl -w vm.max_map_count=262144
...
+kernelCommandLine = "sysctl.vm.max_map_count=262144"
 --------------------------------------------
 
+This will cause all WSL2 VMs to have that setting assigned when they start.
+
+If you are on Windows 11, or Windows 10 version 22H2 and have installed the Microsoft Store
+version of WSL, you can modify the /etc/sysctl.conf within the "docker-desktop" WSL
+distribution, perhaps with commands like this:
+
+[source]
```
grcevski, line 438: This should be `[source,sh]`.

docs/reference/setup/install/docker.asciidoc
```diff
@@ -399,14 +399,53 @@ sudo sysctl -w vm.max_map_count=262144
...
+wsl -d docker-desktop -u root
 sysctl -w vm.max_map_count=262144
+exit
+--------------------------------------------
+
+If you are on these versions of WSL and you do not want to have to run those commands every
+time you restart Docker, you can globally change every WSL distribution with this setting
+by modifying your %USERPROFILE%\.wslconfig as follows:
+
+[source]
```
grcevski, line 426: This should be `[source,text]`.

grcevski:
The checks failed because of missing source language. I've updated the review comments with the appropriate source language.

```
reference/setup/install/docker.asciidoc[416:420]: Snippet missing a language. This is required by Elasticsearch's doc testing infrastructure so we be sure we don't accidentally forget to test a snippet.
```

docs/reference/setup/install/docker.asciidoc
kyanha: As much as I dislike mentioning this, many Windows users and administrators prefer simply closing the window they're working in to exiting -- because they usually can't figure out how to get back to the mode they had been in before. I view it as a potential place where they can learn about how to exit WSL, the same way many users [new to UNIX] need help understanding how to exit out of insert mode in vi.

However, I am not the one who's responsible for the tone of the documentation. That would be you and your organization. I'm simply stating my reason for it, and I won't be particularly sad if you decide it shouldn't be there.

kyanha:
@elasticmachine test this please, if you'll allow an external contributor to request such a check.

grcevski:

grcevski (approved):
LGTM! Thanks for your contribution.

grcevski:
I didn't realize this was targeting 8.2, I'll forward port the changes to main.

kyanha:
It's actually good for everything that mentions Docker Desktop. I do
recognize you are trying to freeze old documentation versons, though.

PR #87277, opened 2022-06-01
Deduplicate pipeline name when reading index requests

original-brownbear:
When not using any pipeline we were wasting 96 bytes per index request for
two freshly read `_none` strings per index request. This came out to
~100M in additional heap used in a recent real world heap investigation.
When using an actual pipeline the name tends to be longer than 5 characters
so the problem could be even more pronounced.

-> deduplicate all the pipeline names to save heap for large bulks


carlosdelest (approved):
LGTM 👍

original-brownbear:
Thanks Carlos!

PR #87278, opened 2022-06-01
Pre-parse tsdb index settings

martijnvg:
The index.mode, index.time_series.start_time and index.time_series.end_time index settings are
parsed in order to determine what write index should be selected when indexing into a tsdb data stream.

Parsing a setting has a certain overhead, which normally isn't noticeable. However, these settings
are parsed every time we need to determine the write index for a document and then parsing these
settings has a noticeable overhead (takes around 10% in captured flame graphs).

This change pre-parses the mentioned settings and records the respected parsed values with
the IndexMetadata instance.

Relates to #74660


server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetadata.java
```diff
@@ -845,6 +863,21 @@ public boolean isPartialSearchableSnapshot() {
         return isPartialSearchableSnapshot;
     }
 
+    @Nullable
+    public IndexMode getIndexMode() {
```
nik9000, line 867: Wants javadoc I think.

nik9000: Just because this class is, like, *everywhere* and it's convenient to be able to see it on mouseover and the like.

martijnvg: I will add javadoc.

nik9000 (approved):

PR #87280, opened 2022-06-01
Assignment record

idegtiarenko:
This change combines shard assignments and number of unassigned shards into a single record so that it is easier to compute and log the difference when tracing.

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalance.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalance.java
@@ -12,24 +12,19 @@ import org.elasticsearch.index.shard.ShardId;
 
 import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 
 /**
  * The desired balance of the cluster, indicating which nodes should hold a copy of each shard.
  *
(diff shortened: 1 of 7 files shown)


server/src/internalClusterTest/java/org/elasticsearch/cluster/ClusterHealthIT.java
```diff
@@ -258,18 +258,14 @@ public void testHealthWithClosedIndices() {
...
     public void testHealthOnIndexCreation() throws Exception {
         final AtomicBoolean finished = new AtomicBoolean(false);
-        Thread clusterHealthThread = new Thread() {
-            @Override
-            public void run() {
-                while (finished.get() == false) {
-                    ClusterHealthResponse health = client().admin().cluster().prepareHealth().get();
-                    assertThat(health.getStatus(), not(equalTo(ClusterHealthStatus.RED)));
-                }
+        Thread clusterHealthThread = new Thread(() -> {
```
DaveCTurner, line 261: I think these changes are unrelated to the rest of the PR, could you back them out here?

DaveCTurner (commented):
Makes sense to me. I left one small request.

DaveCTurner (approved):
LGTM

PR #87285, opened 2022-06-01
Stop versioning Checkstyle IDE config

### LINKED ISSUE #86945: IntelliJ IDE continually modifies source controlled checkstyle plugin configuration
We commit configuration for the Idea Checkstyle plugin in [source control](https://github.com/elastic/elasticsearch/blob/master/.idea/checkstyle-idea.xml) so that developers have this auto-configured for them. One issue is that due to differing versions of the plugin, the IDE will often modify this file, meaning that git changeset now include this file by mistake.

I'm not sure we can ever completely control the fact that the Checkstyle plugin likes to modify this file, so the alternative is _not_ to keep it in source control, and instead generate the file (or copy it from another location) so that if it is modified, it's not a change tracked by git.

pugnascotia:
Closes #86945.

The IntelliJ Checkstyle plugin likes to modify its configuration file
unecessarily, which creates friction for developers because they have
the revert these changes.

Instead, stop versioning the config file and instead update the
`configureIdeCheckstyle` task to copy the required content into place.


build-tools-internal/src/main/groovy/elasticsearch.ide.gradle
```diff
@@ -25,15 +25,28 @@ tasks.register('configureIdeCheckstyle') {
...
 
   inputs.files(file(checkstyleConfig), file(checkstyleIdeFragment))
   outputs.files(file(checkstyleIdeConfig))
 
   doLast {
+    // Configure the IntelliJ Checkstyle plugin by copying a standard file. We don't simply commit
+    // the result to version control, because the plugin has a habit of modifying the file and
+    // replacing the `$PROJECT_DIR$` placeholders, which developers must then revert.
+    Files.copy(
+      Paths.get(file("${resources}/checkstyle-idea.xml").getPath()),
```
mark-vieira, line 45: We should add these as inputs/outputs of this task.

mark-vieira (approved):
One minor comment, otherwise LGTM.

PR #87292, opened 2022-06-01
Use Map in indices responses

rjernst:
This commit converts GetAliasesResponse and GetSettingsResponse to use
Map instead of ImmutableOpenMap.

relates #86239


dakrone (approved):
LGTM

PR #87297, opened 2022-06-01
[ML] adjust change_point to not throw with too few buckets

benwtrent:
It can be frustrating for agg consumers to handle failure modes when we should just indicate that there were too few buckets to calculate the change.

Other aggs simply return NaN or empty results when too few docs are calculated. We should follow suit for continuity of experience.


benwtrent:
https://github.com/elastic/elasticsearch/pull/87297

przemekwitek (approved):
LGTM

benwtrent:
@elasticmachine update branch

benwtrent:
run elasticsearch-ci/part-2

PR #87298, opened 2022-06-01
Allow cluster start with unreachable remote clusters

dnhatn:
If the local cluster has a remote_cluster setting defined in YAML file and the remote cluster is not reachable, then the local cluster won't start. This PR relaxes that constraint as it's too restricted and requires manual intervene.

I hit this constraint while setting two clusters using Rally.

### DIFF
--- a/docs/changelog/87298.yaml
+++ b/docs/changelog/87298.yaml
@@ -1,0 +1,5 @@
+pr: 87298
+summary: Allow start cluster with unreachable remote clusters
+area: Network
+type: enhancement
+issues: []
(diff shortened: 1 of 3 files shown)


original-brownbear (approved):
LGTM

dnhatn:
Thanks Armin!

javanna:
FYI This addressed a long standing issue that was bugging many users for a long time :) We have discussed in the past to maybe honor skip_unavailable but I don't think it's needed. Thanks for taking care of this!

PR #87299, opened 2022-06-01
Using the correct connection to fetch remote master history

masseyke:
Previously MasterHistoryService was using `openConnection` to open a connection to the master in order to fetch the remote master history. However it was not passing that connection to `sendRequest` so we might have been using another one. This change uses `connectToNode` instead.


masseyke:

DaveCTurner (approved):
One nit but otherwise LGTM

server/src/main/java/org/elasticsearch/cluster/coordination/MasterHistoryService.java
```diff
@@ -113,52 +114,51 @@ public List<DiscoveryNode> getRemoteMasterHistory() throws Exception {
      * @param node The node whose view of the master history we want to fetch
      */
     public void refreshRemoteMasterHistory(DiscoveryNode node) {
+        Version minSupportedVersion = Version.V_8_3_0;
+        if (node.getVersion().onOrAfter(minSupportedVersion)) { // This was introduced in 8.3.0
```
andreidan, line 118: Was this backported to 8.3?

PR #87306, opened 2022-06-01
Adding a transport action to get cluster formation info

masseyke:
This PR exposes ClusterFormationFailureHelper.ClusterFormationState in a transport action (ClusterFormationInfoAction), with the intention that it be used as part of the master stability health check. This was started as https://github.com/elastic/elasticsearch/pull/85782, but we realized that we needed more fine-grained information than just the string description of the reason the cluster formation failed.
2022-04-18 Coordination health
This action will be used in branch 1.2.2.4 of the attached diagram (near the bottom right corner). That is, when we detect that there has been no master node in the last 30 seconds (default), and when PeerFinder does not believe that there is an elected master node, and when the current node is master eligible, we will use this transport action to asynchronously send a request to all master-eligible nodes to gather their ClusterFormationStates.


masseyke:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/cluster/coordination/ClusterFormationFailureHelper.java
```diff
@@ -114,17 +120,120 @@ public String toString() {
...
+                in.readLong(),
+                new VotingConfiguration(in),
+                new VotingConfiguration(in),
+                in.readImmutableList(TransportAddress::new),
+                in.readImmutableList(DiscoveryNode::new),
+                in.readLong(),
+                in.readBoolean(),
+                readStatusInfo(in),
+                in.readList(
+                    in1 -> new JoinStatus(
```
andreidan, line 216: could we have a more meaningful name for the inner reader?

server/src/main/java/org/elasticsearch/cluster/coordination/ClusterFormationFailureHelper.java
```diff
@@ -114,17 +120,120 @@ public String toString() {
...
+                        new DiscoveryNode(in1),
+                        in1.readLong(),
+                        in1.readString(),
+                        new TimeValue(in1.readLong(), TimeUnit.valueOf(in1.readString()))
+                    )
+                )
+            );
+        }
+
+        private static StatusInfo readStatusInfo(StreamInput in) throws IOException {
```
andreidan, line 226: Would it be worth making `StatusInfo` `Writeable`?

server/src/main/java/org/elasticsearch/action/admin/cluster/coordination/ClusterFormationInfoAction.java
```diff
@@ -0,0 +1,141 @@
...
+        @Override
+        public int hashCode() {
+            return Objects.hash(clusterFormationState);
+        }
+    }
+
+    /**
+     * This transport action fetches the ClusterFormationState from a remote node.
+     */
+    public static class TransportAction extends HandledTransportAction<
```
andreidan, line 120: See @DaveCTurner 's comment on the previous PR

https://github.com/elastic/elasticsearch/pull/85782/files#r849933849

server/src/main/java/org/elasticsearch/action/admin/cluster/coordination/ClusterFormationInfoAction.java
```diff
@@ -0,0 +1,141 @@
...
+    public static final ClusterFormationInfoAction INSTANCE = new ClusterFormationInfoAction();
+    public static final String NAME = "cluster:internal/formation/info";
+
+    private ClusterFormationInfoAction() {
+        super(NAME, ClusterFormationInfoAction.Response::new);
+    }
+
+    public static class Request extends ActionRequest {
+
+        public Request() {}
```
andreidan, line 44: We'll have to request this information from all master eligible nodes so the request should be able to express that?

server/src/main/java/org/elasticsearch/cluster/coordination/Coordinator.java
```diff
@@ -284,7 +284,7 @@ public Coordinator(
         this.nodeHealthService = nodeHealthService;
     }
 
-    private ClusterFormationState getClusterFormationState() {
+    public ClusterFormationState getClusterFormationState() {
```
andreidan, line 287: @DaveCTurner does it make sense to send this over if the discovery cluster formation timeout has not lapsed yet?

server/src/main/java/org/elasticsearch/cluster/coordination/ClusterFormationFailureHelper.java
```diff
@@ -114,17 +120,120 @@ public String toString() {
...
+                    clusterState.version(),
+                    clusterState.getLastCommittedConfiguration(),
+                    clusterState.getLastAcceptedConfiguration()
+                ),
+                statusInfo,
+                inFlightJoinStatuses
+            );
+        }
+
+        private static boolean calculateHasDiscoveredQuorum(
```
andreidan, line 178: nit: would it make sense to obey the `isXXX`/`hasXXX` naming convention for methods returning boolean?

andreidan (changes requested):

Left a few comments

server/src/main/java/org/elasticsearch/action/admin/cluster/coordination/ClusterFormationInfoAction.java
DaveCTurner: This is good tho, it's just a simple `HandledTransportAction` here.

server/src/main/java/org/elasticsearch/cluster/coordination/ClusterFormationFailureHelper.java
masseyke: JoinStatus is now Writeable and readable from the wire.

server/src/main/java/org/elasticsearch/action/admin/cluster/coordination/ClusterFormationInfoAction.java
andreidan: Okido, sounds good

server/src/main/java/org/elasticsearch/action/admin/cluster/coordination/ClusterFormationInfoAction.java
andreidan: Ah of course - I just realised it does the needful 
```
        transportService.registerRequestHandler(actionName, executor, false, canTripCircuitBreaker, requestReader, new TransportHandler());
```

Sorry for the noise Keith

server/src/main/java/org/elasticsearch/monitor/StatusInfo.java
```diff
@@ -35,6 +37,20 @@ public Status getStatus() {
...
+        if (statusString == null) {
+            return null;
+        }
+        return Status.valueOf(statusString);
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        out.writeOptionalString(status == null ? null : status.name());
+        out.writeOptionalString(info);
```
andreidan, line 51: Are we changing the semantics for `StatusInfo` ? I don't think these fields are optional now. I think we should keep them non-null (ie. mandatory).

andreidan (approved):
LGTM, thanks for working on this Keith

server/src/main/java/org/elasticsearch/monitor/StatusInfo.java
masseyke: There was nothing preventing them from being null. I thought it was very unlikely that someone would pass in nulls (what would be the point?). But if they do, I don't want it to blow up during serialization. From what I can tell there's no good way to enforce null checks on a record. What do you think?

andreidan: I caught up with @dakrone on this and we were thinking of aligning the object to the requirements (ie. non-null fields) via `Objects.requireNonNull` calls in the constructor, alongside strict (non-optional) serialisation.

What do you think @masseyke ?

masseyke: I'm all for enforcing non-nullable fields. I'll change the serialization to assume no nulls, and we can enforce it on a subsequent PR.

PR #87307, opened 2022-06-01
Use Map in SegmentStats

rjernst:
SegmentStats has an ImmutableOpenMap tracking the files that exist
within an index segment. This commit converts the member to a Map. It
also slightly tweaks the contract of the class so that the map is no
longer immutable. The reason for this is that the class as a whole is
used as an accumulator when building stats for a segment. While doing
so, the other members of SegmentStats are modified. Yet with keeping
track of files, the map is needlessly rebuilt. Instead this commit
returns an immutable view for consumers, but keeps the map internally as
mutable for adding more stats. Note that the class is already not thread
safe, so immutability is not needed here for that purpose.

relates #86239


server/src/main/java/org/elasticsearch/index/engine/SegmentsStats.java
```diff
@@ -144,8 +132,8 @@ public ByteSizeValue getBitsetMemory() {
         return new ByteSizeValue(bitsetMemoryInBytes);
     }
 
-    public ImmutableOpenMap<String, FileStats> getFiles() {
-        return files;
+    public Map<String, FileStats> getFiles() {
```
original-brownbear, line 135: This method appears to be only used in tests. Lets add a comment to that effect so nobody starts using this outright and save ourselves the unmodifiable wrapping?

server/src/main/java/org/elasticsearch/index/engine/SegmentsStats.java
```diff
@@ -48,14 +51,7 @@ public SegmentsStats(StreamInput in) throws IOException {
...
         maxUnsafeAutoIdTimestamp = in.readLong();
-
-        final int size = in.readVInt();
-        final ImmutableOpenMap.Builder<String, FileStats> files = ImmutableOpenMap.builder(size);
-        for (int i = 0; i < size; i++) {
-            FileStats file = new FileStats(in);
-            files.put(file.getExt(), file);
-        }
-        this.files = files.build();
+        files = in.readList(FileStats::new).stream().collect(Collectors.toMap(FileStats::getExt, Function.identity()));
```
original-brownbear, line 54: This one could be a little hot in some cases I believe. Let's keep the previous pattern and pre-size with `org.elasticsearch.common.util.Maps#newMapWithExpectedSize` that does without so many allocations for the list/stream etc.?

server/src/main/java/org/elasticsearch/index/engine/SegmentsStats.java
```diff
@@ -78,18 +74,10 @@ public void addBitsetMemoryInBytes(long bitsetMemoryInBytes) {
...
-        for (Map.Entry<String, FileStats> entry : files.entrySet()) {
-            final String extension = entry.getKey();
-            if (map.containsKey(extension)) {
-                FileStats previous = map.get(extension);
-                map.put(extension, FileStats.merge(previous, entry.getValue()));
-            } else {
-                map.put(extension, entry.getValue());
-            }
+    public void addFiles(Map<String, FileStats> newFiles) {
+        for (var newEntry : newFiles.entrySet()) {
```
original-brownbear, line 78: nice :) though NIT, could just go all the way to
```java
newFiles.forEach((key, value) -> files.merge(key, value, FileStats::merge));
```

server/src/main/java/org/elasticsearch/index/engine/SegmentsStats.java
```diff
@@ -29,7 +32,7 @@ public class SegmentsStats implements Writeable, ToXContentFragment {
     private long versionMapMemoryInBytes;
     private long maxUnsafeAutoIdTimestamp = Long.MIN_VALUE;
     private long bitsetMemoryInBytes;
-    private ImmutableOpenMap<String, FileStats> files = ImmutableOpenMap.of();
+    private Map<String, FileStats> files = new HashMap<>();
```
original-brownbear, line 35: Related to the other constructor comment. Make this `final` but assign in the constructors so we can pre-size it when reading from `StreamInput`?

original-brownbear (commented):
Just a few random details.

rjernst:
@original-brownbear I believe I addressed all your comments.

server/src/main/java/org/elasticsearch/index/engine/SegmentsStats.java
```diff
@@ -50,12 +54,11 @@ public SegmentsStats(StreamInput in) throws IOException {
         maxUnsafeAutoIdTimestamp = in.readLong();
 
         final int size = in.readVInt();
-        final ImmutableOpenMap.Builder<String, FileStats> files = ImmutableOpenMap.builder(size);
+        files = Maps.newMapWithExpectedSize(size);
         for (int i = 0; i < size; i++) {
             FileStats file = new FileStats(in);
             files.put(file.getExt(), file);
         }
-        this.files = files.build();
```
idegtiarenko, line 58: Could be simplified with `this.files = in.readMapValues(FileStats::new, FileStats::getExt);`

original-brownbear (approved):
LGTM!

Issue #86472, opened 2022-05-05
Expose metadata via dedicated class for scripting index and update contexts


stu-elastic:
Currently, scripts access metadata such as by various keys on the context object.  Changing this to getters and setters on a metadata object will separate the document source from the metadata and allow better compile time checking for accidentally misspelled keys.

  | Index API | Update API | Update by query API | Reindex API
-- | -- | -- | -- | --
_index | W | R | R | W
_id | R | R | R | W
_routing | R | R | R | W
_version | R | R | R | W
_version_type | R |   |   |  
timestamp | R |   |   |  
now |   | R |   |  
op |   | W | W | W

PR #87309, opened 2022-06-01
Script: Add Metadata to ingest context.

stu-elastic:
Adds the `Metadata` class and `metadata()` method to the ingest context.

Metadata has getters and setters for index, id, routing, version and versionType.
It also has a getter for timestamp.

Refs: #86472


server/src/main/java/org/elasticsearch/script/field/MapBackedMetadata.java
```diff
@@ -0,0 +1,116 @@
...
+            if (number.doubleValue() != version) {
+                // did we round?
+                throw new IllegalArgumentException(
+                    "version may only be set to an int or a long but was [" + number + "] with type [" + obj.getClass().getName() + "]"
+                );
+            }
+            return version;
+        }
+        throw new IllegalArgumentException(
+            "version may only be set to an int or a long but was [" + obj + "] with type [" + obj.getClass().getName() + "]"
```
rjernst, line 65: Seeing this exception on a get is confusing. We should perhaps validate the ctx map when it is passed into the ctor?

server/src/main/java/org/elasticsearch/script/field/MapBackedMetadata.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public Object getRawVersion() {
+        return map.get(VERSION);
+    }
+
+    public MapBackedMetadata setVersion(Long version) {
+        return set(VERSION, version);
+    }
+
+    public void removeVersion() {
```
rjernst, line 77: When is it ok to remove version? It's odd having this and getRawVersion, which is different from the rest of the getters/setters.

server/src/main/java/org/elasticsearch/script/field/MapBackedMetadata.java
```diff
@@ -0,0 +1,116 @@
...
+        map.remove(VERSION);
+    }
+
+    public String getRouting() {
+        return getString(ROUTING);
+    }
+
+    public MapBackedMetadata setRouting(String routing) {
```
rjernst, line 85: Can we do without chaining? Shouldn't users be doing `=` with these so returning the metadata for chaining seems unnecessary?

rjernst (commented):
Looks pretty good, I have some comments.

server/src/main/java/org/elasticsearch/script/field/MapBackedMetadata.java
stu-elastic: The script still has access to ctx and may change it there.  Unfortunately, that means we cannot trust the contents of the map at any point.

server/src/main/java/org/elasticsearch/script/field/MapBackedMetadata.java
```diff
@@ -0,0 +1,116 @@
...
+    public Long getVersion() {
+        Object obj = getRawVersion();
+        if (obj == null) {
+            return null;
+        } else if (obj instanceof Number number) {
+            long version = number.longValue();
+            if (number.doubleValue() != version) {
+                // did we round?
+                throw new IllegalArgumentException(
+                    "version may only be set to an int or a long but was [" + number + "] with type [" + obj.getClass().getName() + "]"
```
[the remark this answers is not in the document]
stu-elastic: The script may put anything it wants in the backing map via ctx.

server/src/main/java/org/elasticsearch/script/field/MapBackedMetadata.java
stu-elastic: Yup, we'd have to override `merge`, `put`, `putAll`, `putIfAbsent`, `replace(k,v)`, `replace(k,v,v)` and `replaceAll`.

Would that be a backwards incompatible change?  It would be rejected after the script runs, but during script execution the map can hold anything as long as it's fixed up before script termination.

qhoxie: Would the wrapped map block/validate all changes, or just do this same type of validation for the [known metadata fields](https://github.com/elastic/elasticsearch/pull/87309/files/502aac29619e4c4021b69085d584deb1b9a30e7c#diff-6b5913f6e06049c3e3c48eb973883a79554093b08b481006d8af106a92a8680cR18-R22)?

rjernst: It would block writing to unknown keys. ie the only valid keys are the known metadata fields (the list you linked isn't all of them, it depends on the script context, but those I think are the ones for ingest, or most of them).

rjernst: Actually, on further inspection of IngestDocument, I think we can do this as I originally suggested @stu-elastic . The sourceAndMetadata map is what we need to wrap. On put, if the key is one of the known metadata fields, we validate the value (this would get validated later anyways when we try to index, this is just moving it more upfront). Unknown keys would just be delegated without validation because those are actually document keys.

stu-elastic: To help illustrate the BWC issue, the following valid script would become invalid if we validated the map.
```
ctx._version = 'foo';
...
ctx._version = (ctx._version == 'foo') ? 10: 20;
```

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.meta.txt
```diff
@@ -0,0 +1,36 @@
...
+    VersionType getVersionType()
+    void setVersionType(VersionType)
+    ZonedDateTime getTimestamp()
+}
+
+class org.elasticsearch.script.IngestScript {
+    Metadata meta()
+}
+
+class org.elasticsearch.index.VersionType {
```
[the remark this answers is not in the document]
stu-elastic: Yes, `VersionType.INTERNAL`.

We could possibly bring them in as constants on the script class, which should make that possible.  It would impact any variable named `internal`, however.

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.meta.txt
```diff
@@ -0,0 +1,36 @@
+#
+# Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License
+# 2.0 and the Server Side Public License, v 1; you may not use this file except
+# in compliance with, at your election, the Elastic License 2.0 or the Server
+# Side Public License, v 1.
+#
+
+# This file only whitelists the script class and the Metadata class for ingest scripts.
+# Due to name conflicts, this whitelist cannot be loaded with other scripts with conflicting Metadata types.
```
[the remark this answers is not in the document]
stu-elastic: The painless test context [loads](https://github.com/elastic/elasticsearch/pull/87309/files#diff-ba2678b0d3fb572e0a19426a2dd2c323425afc15b83369a7ff08ca449c803057L111) all script whitelists.  It's not an issue in this change, but when there are multiple Metadata classes, there is a name collision.

The solution of this PR is to avoid loading meta.txt whitelists in painless_test.  Another option is to explicitly name the script whitelists needed by painless test.

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.meta.txt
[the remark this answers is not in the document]
stu-elastic: We could make those available as getters on the script object but that would be a back compat issue with existing variables of the same name.  Scripts would get:

> variable [external] is already defined

We'd have to have a way to shadow to avoid that issue.

server/src/main/java/org/elasticsearch/script/field/MapBackedMetadata.java
tvernum: IIUC, we're only talking about validating the known metadata fields and the user-facing change would be that they're validated every time they're set rather than only being validated at the end of the script (or, when they're read from the context as part of document storage).

That feels OK to me. If we only validate on `getVersion` (etc) then there's a cost to the user that they cannot easily trace where the version was set. They know that something in their script set the version to a string, but that could potentially happen in more than 1 place.

So, yes, there's the possibility that it might break something for someone, but that's true of many bug fixes. This feels like a reasonable feature to validate on set rather than validate on first use. I agree it's borderline, and someone might have a compelling argument for why we should consider it "breaking" rather than "fixing", but I think I'm OK with it.

modules/lang-painless/src/main/java/org/elasticsearch/painless/PainlessPlugin.java
```diff
@@ -90,13 +90,13 @@ public final class PainlessPlugin extends Plugin implements ScriptPlugin, Extens
...
             List<Whitelist> contextWhitelists = new ArrayList<>();
-            if (PainlessPlugin.class.getResourceAsStream("org.elasticsearch.script." + context.name.replace('-', '_') + ".txt") != null) {
-                contextWhitelists.add(
-                    WhitelistLoader.loadFromResourceFiles(
-                        PainlessPlugin.class,
-                        "org.elasticsearch.script." + context.name.replace('-', '_') + ".txt"
-                    )
-                );
+            String contextWhitelistName = "org.elasticsearch.script." + context.name.replace('-', '_');
+            if (PainlessPlugin.class.getResourceAsStream(contextWhitelistName + ".txt") != null) {
```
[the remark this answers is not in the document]
stu-elastic: https://github.com/elastic/elasticsearch/pull/87539

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.meta.txt
```diff
@@ -0,0 +1,36 @@
...
+    void setVersion(long)
+    void removeVersion()
+    boolean hasVersion()
+    VersionType getVersionType()
+    void setVersionType(VersionType)
+    ZonedDateTime getTimestamp()
+}
+
+class org.elasticsearch.script.IngestScript {
+    Metadata meta()
```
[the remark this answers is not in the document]
stu-elastic: Done.

server/src/main/java/org/elasticsearch/script/IngestScript.java
```diff
@@ -33,18 +37,100 @@ public abstract class IngestScript {
...
+        this.metadata = metadata;
     }
 
     /** Return the parameters for this script. */
     public Map<String, Object> getParams() {
         return params;
     }
 
-    public abstract void execute(Map<String, Object> ctx);
+    public Map<String, Object> getCtx() {
```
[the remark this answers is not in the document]
stu-elastic: Added

server/src/main/java/org/elasticsearch/script/field/MapBackedMetadata.java
```diff
@@ -0,0 +1,116 @@
...
+
+/** Metadata storage backed by a Map for compatibility with the ctx Map used by ingest and update scripts. */
+public class MapBackedMetadata {
+    public static final String INDEX = "_index";
+    public static final String ID = "_id";
+    public static final String VERSION = "_version";
+    public static final String ROUTING = "_routing";
+    public static final String SOURCE = "_source";
+
+    protected Map<String, Object> map;
```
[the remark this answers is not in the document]
stu-elastic: Added.

server/src/main/java/org/elasticsearch/script/IngestScript.java
```diff
@@ -33,18 +37,100 @@ public abstract class IngestScript {
...
+        return metadata != null ? metadata.store.getMap() : null;
+    }
+
+    public Metadata meta() {
```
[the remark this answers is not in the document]
stu-elastic: Changed.

server/src/main/java/org/elasticsearch/script/field/MapBackedMetadata.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public MapBackedMetadata(Map<String, Object> map) {
+        this.map = map;
+    }
+
+    public MapBackedMetadata(int size) {
```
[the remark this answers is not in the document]
stu-elastic: Moved to utility.

server/src/main/java/org/elasticsearch/script/field/MapBackedMetadata.java
stu-elastic: This is used in a future change, removing until then.

stu-elastic:
After discussion with @rjernst, the plan is to
* Remove `org.elasticsearch.script.ingest.meta.txt` and put the metadata stuff in `org.elasticsearch.script.ingest.txt`
* Figure out how to handle `painless_test` loading all whitelists.  This will become an issue for when other metadata classes are whitelisted
* Remove all enums, take strings instead (`VersionType.EXTERNAL` becomes `'external'`)
* `MapBackedMetadata` will implement the `Map<String, Object>` interface and do validation on `put` for `ctx` backwards compatibility.  Only metadata keys will be validated.

server/src/main/java/org/elasticsearch/script/field/MapBackedMetadata.java
stu-elastic: [IngestService](https://github.com/elastic/elasticsearch/blob/a4fdb0ba669cf473a3238faaa28f780d612ff65b/server/src/main/java/org/elasticsearch/ingest/IngestService.java#L908) expects this to be non-null and throws a 500 if it is null.

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.meta.txt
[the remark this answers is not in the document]
stu-elastic: Changed to taking string.

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.meta.txt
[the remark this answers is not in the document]
stu-elastic: Removed

stu-elastic:
Added https://github.com/elastic/elasticsearch/pull/87673 for wrapping the `sourceAndMetadata` map.

stu-elastic:
@rjernst I've added https://github.com/elastic/elasticsearch/pull/88102 to enforce non-null `_version`.   This PR has been updated to use the data structure in https://github.com/elastic/elasticsearch/pull/87673.

It's ready for another pass when you have the time.

modules/ingest-common/src/main/java/org/elasticsearch/ingest/common/ScriptProcessor.java
```diff
@@ -120,16 +117,17 @@ public ScriptProcessor create(
...
                 // verify script is able to be compiled before successfully creating processor.
-                IngestScript ingestScript = null;
+                IngestScript.Factory ingestScriptFactory = null;
                 try {
-                    final IngestScript.Factory factory = scriptService.compile(script, IngestScript.CONTEXT);
-                    if (ScriptType.INLINE.equals(script.getType())) {
-                        ingestScript = factory.newInstance(script.getParams());
+                    ingestScriptFactory = scriptService.compile(script, IngestScript.CONTEXT);
+                    if (ScriptType.STORED.equals(script.getType())) {
+                        // do not cache stored scripts lest they change and invalidate the cached value
```
rjernst, line 124: Is this a change in current behavior? It might be worth pulling this out into a separate PR, only so that it can be noted (perhaps as a bugfix) in the release notes, whereas if it goes in within this enhancement/addition to the fields API it would be hidden.

server/src/main/java/org/elasticsearch/script/IngestScript.java
```diff
@@ -33,18 +33,31 @@ public abstract class IngestScript {
...
 
     /** Return the parameters for this script. */
     public Map<String, Object> getParams() {
         return params;
     }
 
-    public abstract void execute(Map<String, Object> ctx);
+    /** Provides backwards compatibility access to ctx */
+    public Map<String, Object> getCtx() {
+        return metadata;
```
rjernst, line 50: Can we pass ctx separately to ingest script, so that Metadata can remain strongly typed?

rjernst (commented):
This looks much better. A few more thoughts.

modules/ingest-common/src/main/java/org/elasticsearch/ingest/common/ScriptProcessor.java
stu-elastic: This is the same as current behavior, it failed a test which is how I found it.

server/src/main/java/org/elasticsearch/script/IngestScript.java
stu-elastic: Yup, done.

modules/ingest-common/src/main/java/org/elasticsearch/ingest/common/ScriptProcessor.java
stu-elastic: This is the current behavior, it's just the instance used to be cached.  This change is due to a failed unit test when I switched to the compiled script rather than the instance.

rjernst (approved):

PR #87294, opened 2022-06-01
Generate release notes for v8.3.0

craigtaverner:
Once CI is complete the documentation will be available at:
https://elasticsearch_87294.docs-preview.app.elstc.co/diff

PR #87318, opened 2022-06-02
Added migration/index.asciidoc generation support

craigtaverner:
When generating release notes for 8.3.0, it was discovered that they could not build due to missing imports in the migration/index.asciidoc file. This was easily added manually in the PR for that release at https://github.com/elastic/elasticsearch/pull/87294, but I felt it should have been handled by the gradle task `generateReleaseNotes`, so I have added what I think is the right stuff for that.

### DIFF
--- a/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/release/MigrateIndexGenerator.java
+++ b/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/release/MigrateIndexGenerator.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.gradle.internal.release;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static java.util.Comparator.reverseOrder;
+
+/**
+ * This class ensures that the migrate/index page has the appropriate anchors and include directives
+ * for the current repository version.
+ */
+public class MigrateIndexGenerator {
+
+    static void update(Set<MinorVersion> versions, File indexTemplate, File indexFile) throws IOException {
+        try (FileWriter indexFileWriter = new FileWriter(indexFile)) {
+            indexFileWriter.write(generateFile(versions, Files.readString(indexTemplate.toPath())));
+        }
+    }
+
+    @VisibleForTesting
+    static String generateFile(Set<MinorVersion> versionsSet, String template) throws IOException {
+        final Set<MinorVersion> versions = new TreeSet<>(reverseOrder());
+        versions.addAll(versionsSet);
+        final List<String> includeVersions = versions.stream().map(MinorVersion::underscore).collect(Collectors.toList());
+
+        final Map<String, Object> bindings = new HashMap<>();
+        bindings.put("versions", versions);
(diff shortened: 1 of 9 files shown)


build-tools-internal/src/main/resources/templates/migration-index.asciidoc
```diff
@@ -0,0 +1,35 @@
+[[breaking-changes]]
+= Migration guide
```
craigtaverner, line 2: Perhaps it would make more sense to have this template file contain nothing but the dynamic bits at the bottom, together with an include directive for the main content which should be in the docs section, and not in the template, for easier editing.

build-tools-internal/src/test/java/org/elasticsearch/gradle/internal/release/GenerateReleaseNotesTaskTest.java
```diff
@@ -180,6 +180,28 @@ public void getVersions_includesCurrentVersion() {
         );
     }
 
+    /**
+     * Check that when deriving a lit of major.minor versions from git tags, the current unreleased version is included.
```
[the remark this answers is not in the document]
craigtaverner: And included should be excluded.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/release/MigrateIndexGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static java.util.Comparator.reverseOrder;
+
+/**
+ * This class ensures that the migrate/index page has the appropriate anchors and include directives
+ * for the current repository version.
+ */
+public class MigrateIndexGenerator {
```
craigtaverner, line 26: Elsewhere it is `migration` not `migrate` so perhaps change name for consistency.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/release/MigrationIndexGenerator.java
```diff
@@ -0,0 +1,50 @@
...
+        try (FileWriter indexFileWriter = new FileWriter(indexFile)) {
+            indexFileWriter.write(generateFile(versions, Files.readString(indexTemplate.toPath())));
+        }
+    }
+
+    @VisibleForTesting
+    static String generateFile(Set<MinorVersion> versionsSet, String template) throws IOException {
+        final Set<MinorVersion> versions = new TreeSet<>(reverseOrder());
+        versions.addAll(versionsSet);
+        final List<String> includeVersions = versions.stream().map(MinorVersion::underscore).collect(Collectors.toList());
```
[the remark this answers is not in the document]
pugnascotia: Oh wait, I missed that `versions` goes into the bindings, ignore me.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/release/MinorVersion.java
```diff
@@ -0,0 +1,48 @@
...
+     * Converts a QualifiedVersion into a MinorVersion by deleting all but the major and minor components.
+     */
+    public static MinorVersion of(final QualifiedVersion v) {
+        Objects.requireNonNull(v);
+        return new MinorVersion(v.major(), v.minor());
+    }
+
+    @Override
+    public String toString() {
+        return String.format(Locale.ROOT, "%d.%d", major, minor);
```
pugnascotia, line 29: Do we need to call `format`? As opposed to:

```suggestion
        return major + "." + minor;
```

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/release/MinorVersion.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /** Generate version string with underscore instead of dot */
+    public String underscore() {
+        return String.format(Locale.ROOT, "%d_%d", major, minor);
```
pugnascotia, line 34: Do we need to call `format`? As opposed to:

```suggestion
        return major + "_" + minor;
```

pugnascotia (approved):

PR #87319, opened 2022-06-02
Upgrade folders after settings validation

henningandersen:
When upgrading from 7.x to 8.x, we remove the `nodes/0` subfolders from the
data path. This is an irreversible change and once done, downgrade is not
possible. Now do settings validation prior to moving folders to ensure that
a bad setting does not prevent a downgrade.


rjernst (approved):
LGTM

henningandersen:
Thanks Ryan!

PR #87322, opened 2022-06-02
Track ignored shards

idegtiarenko:
Please see comments inline

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceService.java
@@ -53,7 +53,7 @@ public class DesiredBalanceService {
      */
     boolean updateDesiredBalanceAndReroute(DesiredBalanceInput desiredBalanceInput, Predicate<DesiredBalanceInput> isFresh) {
 
-        logger.trace("starting to recompute desired balance");
+        logger.trace("starting to recompute desired balance for [{}]", desiredBalanceInput.index());
 
         final var routingAllocation = desiredBalanceInput.routingAllocation().mutableCloneForSimulation();
         final var routingNodes = routingAllocation.routingNodes();
@@ -244,7 +244,8 @@ public class DesiredBalanceService {
                 || ignored.unassignedInfo().getLastAllocationStatus() == UnassignedInfo.AllocationStatus.NO_ATTEMPT
                 : "Unexpected status: " + ignored.unassignedInfo().getLastAllocationStatus();
 
-            assignments.merge(ignored.shardId(), ShardAssignment.UNASSIGNED, ShardAssignment::merge);
+            var unassigned = ignored.unassignedInfo().getLastAllocationStatus() == UnassignedInfo.AllocationStatus.DECIDERS_NO;
+            assignments.merge(ignored.shardId(), unassigned ? ShardAssignment.UNASSIGNED : ShardAssignment.IGNORED, ShardAssignment::merge);
         }
 
         logger.trace(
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceService.java
```diff
@@ -244,7 +244,8 @@ boolean updateDesiredBalanceAndReroute(DesiredBalanceInput desiredBalanceInput,
                 || ignored.unassignedInfo().getLastAllocationStatus() == UnassignedInfo.AllocationStatus.NO_ATTEMPT
                 : "Unexpected status: " + ignored.unassignedInfo().getLastAllocationStatus();
 
-            assignments.merge(ignored.shardId(), ShardAssignment.UNASSIGNED, ShardAssignment::merge);
+            var unassigned = ignored.unassignedInfo().getLastAllocationStatus() == UnassignedInfo.AllocationStatus.DECIDERS_NO;
+            assignments.merge(ignored.shardId(), unassigned ? ShardAssignment.UNASSIGNED : ShardAssignment.IGNORED, ShardAssignment::merge);
```
idegtiarenko, line 248: It looks like when restoring shards from the snapshots they are ignored for some amount of time until their size could be estimated. We need to keep this information in desired balance to mark them as NO_ATTEMPT until this information is available.

idegtiarenko: This should fix multiple (if not all) snapshot related integration tests.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/ShardAssignment.java
```diff
@@ -15,15 +15,25 @@
...
+    }
+
+    private static Set<String> merge(Set<String> a, Set<String> b) {
+        assert (a.size() > 0 && b.size() > 0) == false : "Only one assignment is expected to have assigned node ids";
+        return a.size() > 0 ? a : b;
     }
 
-    public static ShardAssignment UNASSIGNED = new ShardAssignment(Set.of(), 1);
+    public static ShardAssignment UNASSIGNED = new ShardAssignment(Set.of(), 1, 0);
+    public static ShardAssignment IGNORED = new ShardAssignment(Set.of(), 1, 1);
```
idegtiarenko, line 38: Ignored shard should be marked as unassigned as well is it could not be placed anywhere

DaveCTurner (commented):
I think (worth checking/asserting) that there are really just three cases for ignoring shards - either we ignore all copies (we can't assign the primary yet so replicas are definitely out of scope) or we can handle the primary but ignore the replicas, or no copies are ignored. Does that sound right?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/ShardAssignment.java
```diff
@@ -15,15 +15,25 @@
...
+    }
 
     public static ShardAssignment of(List<ShardRouting> routings) {
-        return new ShardAssignment(routings.stream().map(ShardRouting::currentNodeId).collect(toUnmodifiableSet()), 0);
+        return new ShardAssignment(routings.stream().map(ShardRouting::currentNodeId).collect(toUnmodifiableSet()), 0, 0);
     }
 
     public static ShardAssignment merge(ShardAssignment a, ShardAssignment b) {
-        return new ShardAssignment(a.nodeIds.isEmpty() ? b.nodeIds : a.nodeIds, a.unassigned + b.unassigned);
+        return new ShardAssignment(merge(a.nodeIds, b.nodeIds), a.unassigned + b.unassigned, a.ignored + b.unassigned);
```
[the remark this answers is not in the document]
idegtiarenko: Accidental copy+paste issue

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceService.java
idegtiarenko: I am not sure I follow. `compute` seems to be more verbose as it always calls the `remappingFunction` and require null handling for previous value (when missing).  Could you please clarify why you would like to change this?

DaveCTurner (approved):
LGTM

PR #87324, opened 2022-06-02
Refactor ParametrizedMessage.getFormattedMessage into Strings.format

pgomulka:
ParametrizedMessage is part of log4j api and should not be used
in places where a simple String.format would be enough.
Many of usages like this are message formatting for exceptions being
thrown.

relates #86549


server/src/main/java/org/elasticsearch/cluster/metadata/SingleNodeShutdownMetadata.java
```diff
@@ -113,11 +114,11 @@ private SingleNodeShutdownMetadata(
         this.allocationDelay = allocationDelay;
         if (targetNodeName != null && type != Type.REPLACE) {
             throw new IllegalArgumentException(
-                new ParameterizedMessage(
-                    "target node name is only valid for REPLACE type shutdowns, " + "but was given type [{}] and target node name [{}]",
+                format(
+                    "target node name is only valid for REPLACE type shutdowns, " + "but was given type [%s] and target node name [%s]",
```
grcevski, line 118: We can get rid of the `+` here, it seems the message fits on one line.

x-pack/plugin/shutdown/src/main/java/org/elasticsearch/xpack/shutdown/PutShutdownNodeAction.java
```diff
@@ -156,11 +156,11 @@ public ActionRequestValidationException validate() {
 
             if (targetNodeName != null && type != SingleNodeShutdownMetadata.Type.REPLACE) {
                 arve.addValidationError(
-                    new ParameterizedMessage(
-                        "target node name is only valid for REPLACE type shutdowns, " + "but was given type [{}] and target node name [{}]",
+                    format(
+                        "target node name is only valid for REPLACE type shutdowns, " + "but was given type [%s] and target node name [%s]",
```
grcevski, line 160: Same here, we don't need to concatenate the strings with `+`

grcevski (approved):
LGTM! I left two minor comments.

pgomulka:
@elasticmachine update branch

PR #87334, opened 2022-06-02
Remove ImmutableOpenMap from most xpack tests

rjernst:
Many tests in xpack create ImmutableOpenMap for building mock cluster
states. However, since most of the API signatures for the cluster state
have changed to be Map, it is no longer necessary to build
ImmutableOpenMap. This commit converts most tests in xpack to build Maps
for their mock cluster states.

relates #86239


grcevski (approved):
LGTM!

Issue #87330, opened 2022-06-02
[CI] MlMappingsUpgradeIT testMappingsUpgrade failing


fcofdez:
**Build scan:**
https://gradle-enterprise.elastic.co/s/6nn6coec6m6s6/tests/:x-pack:qa:rolling-upgrade:v6.8.24%23oldClusterTest/org.elasticsearch.upgrades.MlMappingsUpgradeIT/testMappingsUpgrade

**Reproduction line:**
`./gradlew ':x-pack:qa:rolling-upgrade:v6.8.24#oldClusterTest' -Dtests.class="org.elasticsearch.upgrades.MlMappingsUpgradeIT" -Dtests.method="testMappingsUpgrade" -Dtests.seed=8619445F58C7B5CD -Dtests.bwc=true -Dtests.locale=fr-CA -Dtests.timezone=Asia/Ho_Chi_Minh -Druntime.java=8`

**Applicable branches:**
7.17

**Reproduces locally?:**
Didn't try

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.upgrades.MlMappingsUpgradeIT&tests.test=testMappingsUpgrade

**Failure excerpt:**
```
org.elasticsearch.client.ResponseException: method [POST], host [http://127.0.0.1:39383], URI [_ml/anomaly_detectors/ml-mappings-upgrade-job/_open], status line [HTTP/1.1 500 Internal Server Error]
{"error":{"root_cause":[{"type":"exception","reason":"Unexpected job state [failed] while waiting for job to be opened"}],"type":"exception","reason":"Unexpected job state [failed] while waiting for job to be opened"},"status":500}

  at __randomizedtesting.SeedInfo.seed([8619445F58C7B5CD:AB62A008D158CEA9]:0)
  at org.elasticsearch.client.RestClient.convertResponse(RestClient.java:347)
  at org.elasticsearch.client.RestClient.performRequest(RestClient.java:313)
  at org.elasticsearch.client.RestClient.performRequest(RestClient.java:288)
  at org.elasticsearch.upgrades.MlMappingsUpgradeIT.createAndOpenTestJob(MlMappingsUpgradeIT.java:95)
  at org.elasticsearch.upgrades.MlMappingsUpgradeIT.testMappingsUpgrade(MlMappingsUpgradeIT.java:59)
[...]

Issue #87329, opened 2022-06-02
[CI] UpgradeClusterClientYamlTestSuiteIT test {p0=old_cluster/30_ml_jobs_crud/ failing


fcofdez:
**Build scan:**
https://gradle-enterprise.elastic.co/s/6nn6coec6m6s6/tests/:x-pack:qa:rolling-upgrade:v6.8.24%23oldClusterTest/org.elasticsearch.upgrades.UpgradeClusterClientYamlTestSuiteIT/test%20%7Bp0=old_cluster%2F30_ml_jobs_crud%2FPut%20categorization%20job%20on%20the%20old%20cluster%20and%20post%20some%20data%7D

**Reproduction line:**
`./gradlew ':x-pack:qa:rolling-upgrade:v6.8.24#oldClusterTest' -Dtests.class="org.elasticsearch.upgrades.UpgradeClusterClientYamlTestSuiteIT" -Dtests.method="test {p0=old_cluster/30_ml_jobs_crud/Put categorization job on the old cluster and post some data}" -Dtests.seed=8619445F58C7B5CD -Dtests.bwc=true -Dtests.locale=sr-Latn-BA -Dtests.timezone=MET -Druntime.java=8`

**Applicable branches:**
7.17

**Reproduces locally?:**
Didn't try

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.upgrades.UpgradeClusterClientYamlTestSuiteIT&tests.test=test%20%7Bp0%3Dold_cluster/30_ml_jobs_crud/Put%20categorization%20job%20on%20the%20old%20cluster%20and%20post%20some%20data%7D

**Failure excerpt:**
```
[...]

Issue #87328, opened 2022-06-02
[CI] MlJobSnapshotUpgradeIT testSnapshotUpgrader failing


fcofdez:
**Build scan:**
https://gradle-enterprise.elastic.co/s/6nn6coec6m6s6/tests/:x-pack:qa:rolling-upgrade:v6.8.24%23oldClusterTest/org.elasticsearch.upgrades.MlJobSnapshotUpgradeIT/testSnapshotUpgrader

**Reproduction line:**
`./gradlew ':x-pack:qa:rolling-upgrade:v6.8.24#oldClusterTest' -Dtests.class="org.elasticsearch.upgrades.MlJobSnapshotUpgradeIT" -Dtests.method="testSnapshotUpgrader" -Dtests.seed=8619445F58C7B5CD -Dtests.bwc=true -Dtests.locale=sr-Latn-RS -Dtests.timezone=America/Yellowknife -Druntime.java=8`

**Applicable branches:**
7.17

**Reproduces locally?:**
Didn't try

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.upgrades.MlJobSnapshotUpgradeIT&tests.test=testSnapshotUpgrader

**Failure excerpt:**
```
org.elasticsearch.ElasticsearchStatusException: Elasticsearch exception [type=exception, reason=Unexpected job state [failed] while waiting for job to be opened]

  at __randomizedtesting.SeedInfo.seed([8619445F58C7B5CD:9A3A5D305FB21E43]:0)
  at org.elasticsearch.rest.BytesRestResponse.errorFromXContent(BytesRestResponse.java:178)
  at org.elasticsearch.client.RestHighLevelClient.parseEntity(RestHighLevelClient.java:2484)
  at org.elasticsearch.client.RestHighLevelClient.parseResponseException(RestHighLevelClient.java:2461)
  at org.elasticsearch.client.RestHighLevelClient.internalPerformRequest(RestHighLevelClient.java:2184)
  at org.elasticsearch.client.RestHighLevelClient.performRequest(RestHighLevelClient.java:2137)
  at org.elasticsearch.client.RestHighLevelClient.performRequestAndParseEntity(RestHighLevelClient.java:2105)
  at org.elasticsearch.client.MachineLearningClient.openJob(MachineLearningClient.java:378)
[...]

PR #87336, opened 2022-06-02
[ML] mute linux bwc from before 7 due to ubuntu 22 support

benwtrent:
Ubuntu 22 is not supported for native ML components. So muting BWC tests upgrading from 6.8.x.

closes https://github.com/elastic/elasticsearch/issues/87330
closes https://github.com/elastic/elasticsearch/issues/87328
closes https://github.com/elastic/elasticsearch/issues/87329


benwtrent:
A different path may be forward for handling these test failures. putting on pause.

Issue #87335, opened 2022-06-02
Only perform ensureNoSelfReferences check if pipeline contains a script processor


dakrone:
### Description

This is a followup to #85926.

As part of the ingest processing pipeline, we call `CollectionUtils.ensureNoSelfReferences(...)` to ensure that a script processor didn't accidentally introduce a circular object. However, this check is only necessary if we know there was the potential for a circular option, which can only happen with either a custom processor (a plugin where we have no control), or a script processor. We should be able to only call this sanity check if we know we need it, rather than for every document.

PR #87337, opened 2022-06-02
Skip ensureNoSelfReferences check in IngestService

dakrone:
A self reference check is performed by the XContentBuilder when the document map is serialized using the `indexRequest.source(...)` call. This means we can remove the call to `CollectionUtils.ensureNoSelfReferences` above and only do the check once instead of twice.

Relates to #85926 and (subsequent work in) #87335


jakelandis (approved):
LGTM

PR #87338, opened 2022-06-02
Disallow three-digit minor and revision versions

dakrone:
This came up because a yml test erroneously had 8.2.999 as a skip version, however, when the yml
test parser calls `Version.fromString("8.2.999")` the version `8.11.99` is what is actually
parsed out. This commit makes the minor and revision numbers have a limit of two digits, to match
our existing version conceptions. There is no limit on the number of major version digits.

Hopefully we don't need more than 100 minor versions within a single major (yet), or more than 100
patch releases.


rjernst (approved):
LGTM

PR #87347, opened 2022-06-03
Add logging for ID Token header on BadJoseException

ywangd:
When ID token validation fails with BadJoseException, the JWT Header
section can provide useful information for troubleshooting without
having to ask for the full ID token. This PR adds a debug message to
show the header.

Note the header may not be sufficient for debugging all failure
scenarios. But it should be helpful for more common configuration
errors.


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java
```diff
@@ -288,6 +289,7 @@ private void getUserClaims(
                         })
                     );
             } else {
+                LOGGER.debug("ID Token header: {}", idToken.getHeader());
```
justincr-elastic, line 292: Should we log idToken.getJWTClaimsSet() too?

justincr-elastic (changes requested):

justincr-elastic:
Thank you @ywangd. I was going to add the same debug log to OIDC realm. Two feedbacks.
1. Should we include the JWT claims set in the debug log? In other words, log the header and claims set? I did something similar in JWT realm.
2. Should we copy the same debug log into the catch block?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java
ywangd: I was trying to keep the scope limited. But since JWT realm does something similar already, I think we can do it here too.

ywangd:
> 1. Should we include the JWT claims set in the debug log? In other words, log the header and claims set? I did something similar in JWT realm.

Yes I think it makes sense to log both of them.

> 2. Should we copy the same debug log into the catch block?

I think it's worthwhile to add make the debug message generally available and not just in error conditions. So I moved it to the beginning of the `getUserClaims` method so the message is logged in all cases (when debug logging is enabled). I believe this is similar to how JWT realm handles it.

ywangd:
> > 1. Should we include the JWT claims set in the debug log? In other words, log the header and claims set? I did something similar in JWT realm.
> 
> Yes I think it makes sense to log both of them.

Sorry, on a second thought, I don't think it's necessary to log the claims. A few reasons:
1. The claims are already logged 
https://github.com/elastic/elasticsearch/blob/f4b720e1dfd071156eace73a89d254ab23dd102c/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java#L258-L261

It's at trace level and after validation is done. We can change it to debug level if we want, but it's not really necessary (see below). It won't be logged if validation fails and that's fine because the failure is either (1) with the Header which we are logging now or (2) with the claim parsing which means it is just not possible to log the claims.

2. Claims are effectively logged in other places, e.g. `NativeRoleMappingStore` [logs](https://github.com/elastic/elasticsearch/blob/f4b720e1dfd071156eace73a89d254ab23dd102c/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/support/mapper/NativeRoleMappingStore.java#L383) the `UserData` at debug level which contains the claims as its metadata.

So in summary, I lifed the logging message to the beginning of the method so it's general available. But it still just logs the Header.

justincr-elastic (approved):

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java
```diff
@@ -288,6 +289,7 @@ private void getUserClaims(
...
                 claimsListener.onFailure(new ElasticsearchSecurityException("Failed to parse or validate the ID Token", e));
             }
         } catch (com.nimbusds.oauth2.sdk.ParseException | ParseException | JOSEException e) {
```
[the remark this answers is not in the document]
ywangd: I added a debug logging for the idToken and nonce.

Issue #81238, opened 2021-12-02
When CCR disallow auto follow leader system indices should show error 


Leaf-Lin:
**Elasticsearch version** (`bin/elasticsearch --version`): 8.0.0-beta1

**Plugins installed**: []

**JVM version** (`java -version`):

**OS version** (`uname -a` if on a Unix-like system):

**Description of the problem including expected versus actual behavior**:
From https://github.com/elastic/elasticsearch/pull/72815 where it disallows CCR to auto follow leader system indices.  
This is fine, except that one can still use the API or Kibana UI to create auto follow pattern which contains system indices.

Internally, Elasticsearch prevents system indices getting replicated, but there's no error message associated with the API or UI makes user unable to figure out why system indices were not replicated.
 
**Steps to reproduce**:

 1. Create two deployments clusterA and clusterB both on 8.0.0.
 2. Setup CCR so that clusterA can access remote clusterB.
 3. On clsuterA, delete `.kibana_8.0.0_001`.
 4. Create an `auto_follow` pattern on clusterA specify leader system indices from clusterB:
```
PUT /_ccr/auto_follow/kibana
{
  "remote_cluster": "clusterB",
  "leader_index_patterns": [
    ".kibana_8.0.0_001"
  ],
  "leader_index_exclusion_patterns": [],
  "follow_index_pattern": "{{leader_index}}"
}
```

--> Expected behavior is that an error message should appear saying `.kibana_8.0.0_001` is a system index so that the operation will not succeed. 
--> Reality is that it returns the following as if this was accepted.
```
{
  "acknowledged" : true
}
```

Leaf-Lin:
I just encountered exactly the same issue, but for a different category of indices where CCR failed to report an error at the time when setting up auto_follow for searchable snapshot indices. 
### Prerequisite
1. First set up two clusters.  On the leader cluster config hot, cold and frozen tier. On the follower cluster, config a hot tier.
2. Setting up ILM policies on the leader cluster so that the index will transit from hot to cold to frozen.
```
### On the leader cluster
PUT _ilm/policy/timeseries_policy
{
  "policy": {
    "phases": {
      "hot": {                                
        "actions": {
          "rollover": {
            "max_age": "1m"
          }
        }
      },
      "cold": {
        "min_age": "5m",         
        "actions": {
          "searchable_snapshot" : {
            "snapshot_repository" : "found-snapshots"
          }                        
        }
      },
      "frozen": {
        "min_age": "10m",         
[...]

Issue #67668, opened 2021-01-18
ILM: document managing indices with ILM in the context of CCR


andreidan:
When indices in a CCR environment are managed by ILM the follower indices must be converted into regular indices. This is taken care of by the [unfollow action](https://www.elastic.co/guide/en/elasticsearch/reference/current/ilm-unfollow.html) and is injected automatically if rollover, shrink or the searchable_snapshot` action are configured.
However, the leader index must have the `index.lifecycle.indexing_complete` setting configured to `true` in order for the `unfollow` action to succeed. ILM takes care of managing the `indexing_complete` setting as part of the `rollover` action. Namely, after the index is rolled over the setting is configured to true (as the next generation index si the write index).
This subtlety is not explicitly documented and I think we should document it in the ILM docs. Also, the CCR docs should define a full guide on managing indices with ILM.

inqueue:
ILM policies and index templates need to be manually applied to follower clusters. This would be done anyway if configuring the follower for HA, though it would be helpful to explicitly mention it in the docs.

Leaf-Lin:
I've written a step-by-step tutorial, is it ok for @debadair to also look into converting this into a doc page?

# CCR with searchable snapshot

## Summary
CCR as it is designed today does not replicate searchable snapshot indices. Fortunately, the cold/frozen indices are expected to be read-only, so the following procedure outlines what is the solution to have data ageing through ILM policies and still be “replicated” across clusters.

The key points are:
- During the hot and warm phase, CCR will replicate the leader index as follower index from leader cluster to follower cluster.
- Before moving to the cold or frozen phase on the follower cluster, CCR will check leader index has completed indexing
Leader index must have `index.lifecycle.indexing_complete` set to `false` (can be activated by `rollover` action)
- On the follower cluster, ILM cold/frozen phase should contain an additional unfollow action.
- Once the follower cluster verified leader index has completed indexing, it will `unfollow` the leader, then it will convert the index into a searchable snapshot and move into the cold/frozen phase.
- Once the leader cluster verified follower is no longer following, it will convert the index into a searchable snapshot index while moving into the cold/frozen phase. 

The end results are:
- The leader cluster will have searchable snapshots stored on the leader repository.
- The follower cluster will have searchable snapshots on the follower repository.
- In the cold/frozen phase, data will no longer be replicated between leader and follower. (but these are expected to be read-only, so this should not be an issue)

[...]

Issue #87055, opened 2022-05-23
Should restored indices be automatically be replicated by auto follow patterns?


martijnvg:
Auto following automatically triggers index follow of indices in a remote cluster that match with configured patterns. 
When auto following finds indices in the remote cluster that need to be auto followed then it delegates the put follow api.
If the index following is triggered then it records the index uuid of these indices, so auto following doesn't auto 
follow these indices again.

If an index that has been auto followed and is deleted in the leader cluster and then restored and the original replicated index still exists in follower cluster then it can cause an error in the follower cluster:

```
[2022-05-11T16:54:32,137][WARN ][o.e.x.c.a.AutoFollowCoordinator] [xyz] failure occurred while auto following index [[xyz]] for auto follow pattern [xyz]
org.elasticsearch.snapshots.SnapshotRestoreException: [xyz:_latest_/_latest_] cannot restore index [xyz] because an open index with same name already exists in the cluster. Either close or delete the existing index or restore the index under a different name by providing a rename pattern and replacement name
	at org.elasticsearch.snapshots.RestoreService$RestoreSnapshotStateTask.validateExistingClosedIndex(RestoreService.java:1454) ~[elasticsearch-7.17.2.jar:7.17.2]
	at org.elasticsearch.snapshots.RestoreService$RestoreSnapshotStateTask.execute(RestoreService.java:1244) ~[elasticsearch-7.17.2.jar:7.17.2]
	at org.elasticsearch.cluster.ClusterStateUpdateTask.execute(ClusterStateUpdateTask.java:51) ~[elasticsearch-7.17.2.jar:7.17.2]
	at org.elasticsearch.cluster.service.MasterService.executeTasks(MasterService.java:836) ~[elasticsearch-7.17.2.jar:7.17.2]
	at org.elasticsearch.cluster.service.MasterService.calculateTaskOutputs(MasterService.java:403) ~[elasticsearch-7.17.2.jar:7.17.2]
	at org.elasticsearch.cluster.service.MasterService.runTasks(MasterService.java:243) [elasticsearch-7.17.2.jar:7.17.2]
	at org.elasticsearch.cluster.service.MasterService.access$100(MasterService.java:63) [elasticsearch-7.17.2.jar:7.17.2]
	at org.elasticsearch.cluster.service.MasterService$Batcher.run(MasterService.java:170) [elasticsearch-7.17.2.jar:7.17.2]
[...]

PR #87348, opened 2022-06-03
[DOCS] Add CCR limitation

### LINKED ISSUE #86121: [Doc] document Cross Cluster Replication (CCR) limitations
### Elasticsearch Version

7.x, 8.x

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

na

### Problem Description

Given the increased adoption of CCR, it seems it would make sense to provide a dedicated section or page to document any kind of limitations. For example, it is not clear whether system indices like `.security*` can or should be replicated or not.

One might think that CCR is an easy way to replicate Kibana saved objects, Fleet/APM settings, and similar and given that `leader_index_exclusion_patterns` seems to be empty by default, it is not clear whether it is discouraged to replicate such system indices or not.

For example https://github.com/elastic/enhancements/issues/11724

### Steps to Reproduce

-

### Logs (if relevant)

_No response_

Leaf-Lin:
Outlining CCR limitations in today's implementation.

closes https://github.com/elastic/elasticsearch/issues/86121

The searchable snapshot part should also link to https://github.com/elastic/elasticsearch/issues/67668 and https://github.com/elastic/elasticsearch/issues/81238

Preview link: https://elasticsearch_87348.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/xpack-ccr.html#ccr-limitations


lockewritesdocs:
@Leaf-Lin, I moved this content to a section within the overall CCR context, moved the links to other pages into the separate sections, and made some additional edits. Let me know if these changes work for you.

lockewritesdocs:

bahaaldine:
@Leaf-Lin should we also add that CCR doesn't work across clusters that are individually configured with private link?

lockewritesdocs:
Perhaps someone from @elastic/es-distributed can review these changes for accuracy.

lockewritesdocs:
@elasticmachine update branch

lockewritesdocs:
@Leaf-Lin, I consolidated this information considerably. I don't think that we need to get into the implementation details of where things are stored, and can instead list the data that isn't replicated with CCR and inform users that they'll need to manually replicate this information. Let me know if this change works for you, and if so, I'll merge.

docs/reference/ccr/index.asciidoc
```diff
@@ -307,6 +307,33 @@ and use {ccr}:
...
+* <<system-indices,System indices>>
+* {ml-docs}/machine-learning-intro.html[Machine learning jobs]
+* <<index-templates,index templates>>
+* <<index-lifecycle-management,{ilm-cap}>> and
+<<snapshot-lifecycle-management-api,{slm}>> polices
+* {ref}/mapping-roles.html[User permissions and role mappings]
+* <<snapshots-register-repository,Snapshot repository settings>>
+* <<modules-cluster,Cluster settings>>
+* <<searchable-snapshots,Searchable snapshot>>
+* <<restore-index-data-stream,Restored indices>>
```
henningandersen, line 325: Is this true, would auto-follow patterns not hit that?

henningandersen (commented):
Just one question, otherwise looking good.

docs/reference/ccr/index.asciidoc
Leaf-Lin: This was referring to the problem described in this issue: https://github.com/elastic/elasticsearch/issues/87055

lockewritesdocs: Based on the conversation in #87055, it sounds like restored indices aren't replicated if the leader index is deleted and then restored, but the original replicated index still exists in the remote cluster. The restored index in the leader cluster gets a new UUID, so the auto-follow pattern on the remote cluster sees it as new index, but it's actually just a replicated index. 

This sounds like an edge case that we could call attention to, but it seems like restored indices _would be_ replicated if the underlying leader index is still intact. Is that true @henningandersen?

henningandersen: I agree @lockewritesdocs, it will follow just fine if the restore results in an unoccupied name on the follower. This is similar to any sort of naming conflict on the follower though a delete and restore on the leader side is possibly a common cause of this.

I'd opt to remove this, since it is not true that restored indices are not followed.

lockewritesdocs: 👍 Sounds good @henningandersen -- I've removed "Restored indices" from the list of things that aren't followed.

henningandersen (approved):
LGTM.

PR #87351, opened 2022-06-03
Faster NumberFieldMapper.value

original-brownbear:
Calling `currentToken` isn't as cheap as a getter call since it hits a huge switch statement
and doesn't inline well. Even more so when a delegating x-content parser is called which has
a bunch of additional overhead -> saving the expensive call here to save a couple of % of runtime
for number field mappers.


cbuescher (approved):
LGTM

PR #87352, opened 2022-06-03
Only perform ensureNoSelfReferences check during ingest when needed

### LINKED ISSUE #87335: Only perform ensureNoSelfReferences check if pipeline contains a script processor

martijnvg:
The 'ensure no self reference' check during ingest only needs to be performed
when there is a chance that a map or list is added to `IngestDocument` that
references other entries in the `IngestDocument` and when there is a risk that
a circular reference is created between entries.

Today this is only possible in `ScriptProcessor`, since a custom script could
do this. So doing this check if no script processor is used is a not useful.
Doing this check just adds an additional tax to time spent on ingest, that
in many cases really isn't necessary.

A flag is added to `IngestDocument` that controls whether the
'ensure no self reference' check is performed at the end when all pipelines
have been executed. The `ScriptProcessor` has been modified to set this flag and
so this check will be performed if pipelines that execute have one or more script
processors.

Closes #87335


server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
```diff
@@ -843,6 +845,26 @@ List<String> getPipelineStack() {
...
+    /**
+     * Whether the ingest framework should perform a self referencing check after this ingest document
+     * has been processed by all pipelines. Doing this check adds an extra tax to ingest and should
+     * only be performed when really needed. Only if a processor is executed that could add self referencing
+     * maps or lists then this check must be performed. Most processors will not be able to do this, hence
+     * the default is <code>false</code>.
+     *
+     * @param doNoSelfReferencesCheck Whether a self referencing check should be performed
+     */
+    public void setDoNoSelfReferencesCheck(boolean doNoSelfReferencesCheck) {
```
dakrone, line 864: I think we don't need to conform to the name of the original check (which makes this quite a mouthful). Perhaps we could call this `doSelfReferenceCheck(boolean doCheck)` so it is a little easier to tell what it does from the name?

dakrone (approved):
LGTM, I left one comment about naming

PR #87354, opened 2022-06-03
Allow pipeline processor to ignore missing pipelines

### LINKED ISSUE #87323: Add ability to ignore a missing pipeline to the pipeline processor
### Description

I’m trying to create a pipeline processor that will avoid throwing errors if the pipeline it’s referencing doesn’t exist, but will still throw errors if the pipeline does exist but fails. It seems there are two potential approaches:

### 1. Add a `on_failure` block that can handle this case

My initial idea was to use an `on_failure` block with a single fail processor that has a condition to not fail if the failure was due to non-existent pipeline error. However, it seems I can’t figure how to access the `_ingest.on_failure_*` fields from within the `if` condition, but they are accessible from within the message option. I tried accessing this via `ctx['_ingest']` as well without any luck. It seems we need code changes to expose this to the `if` painless context. Here’s the pipeline I’m trying to create:

```
PUT _ingest/pipeline/test-pipeline
{
  "processors": [
    {
      "pipeline": {
        "name": "test-custom",
        "tag": "fleet-custom-caller",
        "on_failure": [
          {
            "fail": {
              "if": "_ingest.on_failure_processor_type != 'pipeline' || _ingest.on_failure_processor_tag != 'fleet-custom-caller' || !_ingest.on_failure_message.contains('non-existent pipeline [test-custom]')",
              "message": "test-custom pipeline failed with {{ _ingest.on_failure_message }}"
            }
          }
        ]
      }
    }
  ]
}
```

### 2. Add a dedicated `ignore_missing` option to the `pipeline` processor

As @felixbarny has discussed in https://github.com/elastic/package-spec/issues/129#issuecomment-1144758312, it's probably preferable to have a dedicated processor option for this in order to:
1. Avoid creating an exception at all and the overhead involved
2. Have a less fragile solution that is not dependent on an error message string

```
PUT _ingest/pipeline/test-pipeline
{
  "processors": [
    {
      "pipeline": {
        "name": "test-custom",
        "ignore_missing": true
      }
    }
  ]
}
```

martijnvg:
Add `ignore_missing_pipeline` option to `pipeline` processor.
This controls whether the `pipeline` processor should fail
with an error if no pipeline with a name specified in the
`name` option exists.

This enhancement is useful to setup a pipeline infrastructure
that lazily adds extension points for overwrites. So that
for specific cluster setups custom pre-processing can be added
at a later point in time.

Relates to #87323


joshdover (approved):
LGTM from a Fleet perspective. I tested this with different combinations of these pipelines dev tools commands and everything worked as expected:

```
PUT _ingest/pipeline/test-pipeline
{
  "processors": [
    {
      "append": {
        "field": "hello",
        "value": "from test-pipeline"
      }
    },
    {
      "pipeline": {
        "name": "test-custom",
        "ignore_missing_pipeline": true
      }
    }
  ]
}

PUT _ingest/pipeline/test-custom
{
  "processors": [
    {
      "append": {
        "field": "hello",
        "value": "from test-custom"
      }
    },
    {
      "fail": {
        "message": "boom!"
      }
    }
  ]
}

DELETE _ingest/pipeline/test-custom

POST _ingest/pipeline/test-pipeline/_simulate
{
  "docs": [
    { "_index": "x", "_source": { "foo": "bar" } }  
  ]
}
```

dakrone (approved):
LGTM, thanks for adding this Martijn!

andrewkroh (commented):
Great idea.

PR #87356, opened 2022-06-03
Lower overhead of RestApiVersion use in x-content parsing

original-brownbear:
Noticed loads of duplicate lambdas on the heap and code related to these predicates
etc. pop up during benchmarking things that are hot on x-content parsing.
These changes way simplify the rest-api code (though it could be made even simpler I think)
and remove it from profiling for the most part.


pgomulka (changes requested):
this PR has some great improvements (use enum set and predicate), but I am not sure how this helps with performance. @original-brownbear can you please share the performance results?

libs/core/src/main/java/org/elasticsearch/core/RestApiVersion.java
```diff
@@ -28,32 +28,30 @@ public enum RestApiVersion {
...
-
-    private static RestApiVersion fromMajorVersion(int majorVersion) {
-        return valueOf("V_" + majorVersion);
+    public boolean matches(Predicate<RestApiVersion> restApiVersionFunctions) {
+        return restApiVersionFunctions.test(this);
     }
 
     public static RestApiVersion minimumSupported() {
-        return current().previous();
+        return V_7;
```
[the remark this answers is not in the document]
joegallo: What do you think about 3fc1d001b8c790334b97cc2aded4b7d2a44705c9, @pgomulka? I was inclined to have a `PREVIOUS`/`previous()` pair, too.

original-brownbear:
> this PR has some great improvements (use enum set and predicate), but I am not sure how this helps with performance. 

The complicated code paths taken as a result of having to instantiate these predicates and then calling them doesn't seem to compile very well. In some cases this stuff starts taking up a couple of percent of the json parsing paths and breaks inlining here and there. Also, having this stuff in the profiling just makes interpreting results more complicated :)

libs/core/src/main/java/org/elasticsearch/core/RestApiVersion.java
```diff
@@ -23,37 +23,40 @@ public enum RestApiVersion {
...
-    public static RestApiVersion current() {
-        return CURRENT;
+        return V_7;
     }
 
-    public static Function<RestApiVersion, Boolean> equalTo(RestApiVersion restApiVersion) {
-        return r -> r.major == restApiVersion.major;
+    public static Predicate<RestApiVersion> equalTo(RestApiVersion restApiVersion) {
+        return switch (restApiVersion) {
+            case V_8 -> r -> r.major == V_8.major;
```
rjernst, line 50: If performance is a concern, these predicates could be created statically and then just referenced here?

rjernst (commented):
One minor suggestion

libs/core/src/main/java/org/elasticsearch/core/RestApiVersion.java
original-brownbear: That won't make a difference performance wise. We're not capturing anything in these lambdas so they just compile as static methods anyway I think?

libs/core/src/main/java/org/elasticsearch/core/RestApiVersion.java
```diff
@@ -28,32 +28,30 @@ public enum RestApiVersion {
...
+        return V_7;
     }
 
     public static RestApiVersion current() {
         return CURRENT;
     }
 
-    public static Function<RestApiVersion, Boolean> equalTo(RestApiVersion restApiVersion) {
-        return r -> r.major == restApiVersion.major;
+    public static Predicate<RestApiVersion> equalTo(RestApiVersion restApiVersion) {
```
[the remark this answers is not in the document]
original-brownbear: Yes, duplicate instances and they compiled to more complicated code and didn't inline so well apparently.

original-brownbear:
@elasticmachine update branch

libs/core/src/main/java/org/elasticsearch/core/RestApiVersion.java
rjernst: I'm not sure that is true. I thought a new lambda would be created here on every invocation, regardless of what is captured or not.

original-brownbear: I think the JLS doesn't explicitly enforce it, but practically we shouldn't have any allocation here. At least on OpenJDK17 we eventually always get the same instances returned because the `invokedynamic` is bootstrapped into a static method handle. Using a constant compiles the same I think.

libs/core/src/main/java/org/elasticsearch/core/RestApiVersion.java
```diff
@@ -23,37 +23,40 @@ public enum RestApiVersion {
...
+        return PREVIOUS;
     }
 
     public static RestApiVersion minimumSupported() {
-        return current().previous();
-    }
-
-    public static RestApiVersion current() {
-        return CURRENT;
+        return V_7;
```
pgomulka, line 45: can we use `previous` here too?

pgomulka (approved):
LGTM, left one comment

original-brownbear:
Thanks all!

PR #87362, opened 2022-06-03
Don't check contains query for multi-shapes in LatLonShapeDocValuesQueryTests

### LINKED ISSUE #85177: LatLonShapeDocValuesQueryTests.testIndexMultiShapes
### CI Link

n/a

### Repro line

./gradlew ':x-pack:plugin:spatial:test' --tests "org.elasticsearch.xpack.spatial.index.mapper.LatLonShapeDocValuesQueryTests.testIndexMultiShapes" -Dtests.seed=368A99A4FD18272B -Dtests.locale=lv -Dtests.timezone=Africa/Niamey -Druntime.java=17

### Does it reproduce?

Yes

### Applicable branches

7.17

### Failure history

_No response_

### Failure excerpt

```
java.lang.AssertionError: expected:<0> but was:<1>
        at __randomizedtesting.SeedInfo.seed([368A99A4FD18272B:EABCA164BFB771B4]:0)
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at org.elasticsearch.xpack.spatial.index.mapper.LatLonShapeDocValuesQueryTests.assertQueries(LatLonShapeDocValuesQueryTests.java:148)
        at org.elasticsearch.xpack.spatial.index.mapper.LatLonShapeDocValuesQueryTests.testIndexMultiShapes(LatLonShapeDocValuesQueryTests.java:141)
```

iverase:
Due to a rare bug in lucene (LUCENE-10514), this test might fail for CONTAINS queries. The fix has been fixed upstream but not back-ported to this lucene version. That's ok as the bug is rare and affects only multishapes where one shape contains another shape.

So we just skip CONTAINS queries for multishapes so the test doesn't fail.

fixes  #85177

### DIFF
--- a/x-pack/plugin/spatial/src/test/java/org/elasticsearch/xpack/spatial/index/mapper/LatLonShapeDocValuesQueryTests.java
+++ b/x-pack/plugin/spatial/src/test/java/org/elasticsearch/xpack/spatial/index/mapper/LatLonShapeDocValuesQueryTests.java
@@ -136,6 +136,10 @@ public class LatLonShapeDocValuesQueryTests extends ESTestCase {
         for (int i = 0; i < 25; i++) {
             LatLonGeometry[] geometries = randomLuceneQueryGeometries();
             for (ShapeField.QueryRelation relation : ShapeField.QueryRelation.values()) {
+                if (relation == ShapeField.QueryRelation.CONTAINS) {
+                    // We don't check this here as it might fail due to LUCENE-10514
+                    continue;
+                }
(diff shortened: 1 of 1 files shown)


craigtaverner (approved):
The fix works! I was surprised to see no muted tests unmuted though. So I guess this was a recent failure that had not yet been muted.

PR #87363, opened 2022-06-03
Warn when stored credential hashers not compliant with FIPS

n1v0lg:
Warn in FIPS mode, if hashers for API keys or service tokens are not
compliant with FIPS. Going for a warning log instead of validation
error for bwc.


x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/SecurityTests.java
```diff
@@ -619,7 +619,13 @@ public void testValidateForFipsNoErrorsOrLogs() throws IllegalAccessException {
                 )
             )
             .put(
-                XPackSettings.PASSWORD_HASHING_ALGORITHM.getKey(),
+                XPackSettings.SERVICE_TOKEN_HASHING_ALGORITHM.getKey(),
```
n1v0lg, line 622: The diff looks weird because there was a duplicate entry for `XPackSettings.PASSWORD_HASHING_ALGORITHM.getKey()` here before

ywangd (approved):
LGTM

PR #87366, opened 2022-06-03
[ML] Distribute model allocations respecting node allocated processors

dimitris-athanasiou:
We can improve a model's latency and throughput by using more
threads for each inference request. We can also improve its
throughput by processing multiple inference requests in parallel.
A user can set the model's `number_of_allocations` and `threads_per_allocation`
settings to increase performance.
However, if we use more threads than the node's allocated processors,
we end up with thread oversubscription and performance deteriorates throughout.

This commit changes the way model allocations are distributed across the
ML nodes of the cluster. Up to now, we were trying to allocate each model
on every node. Now, we make use of the `AssignmentPlanner` class introduced
in #86004 in order to compute an assignment plan that distributes model
allocations across the cluster while we maximize the number of allocations
we provide to each model without oversubscribing the nodes' CPU.


dimitris-athanasiou:
Raising this as draft for now as I want to test it on cloud.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -245,55 +256,107 @@ public void clusterStateProcessed(ClusterState oldState, ClusterState newState)
...
+                return;
             }
-        }
-        if (nodeToReason.isEmpty() == false) {
-            assignmentBuilder.setReason(
-                nodeToReason.entrySet()
-                    .stream()
-                    .map(entry -> String.format(Locale.ROOT, "Not allocating on node [%s]. Reason: %s", entry.getKey(), entry.getValue()))
-                    .collect(Collectors.joining("|"))
+            submitUnbatchedTask(reason, new ClusterStateUpdateTask() {
```
benwtrent, line 293: We should totally allow these assignments updates to be batched. We can at least keep track of the individual creation requests and handle them all at once in a batched manner. This would require creating these cluster update task objects, keeping track of the listeners, and putting them together, but we should do this extra work.

This will at least reduce the amount of work required by the "rebalanceAssignments" as we know we have all the assignments requested at once. Just not anomaly detection jobs.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -245,55 +256,107 @@ public void clusterStateProcessed(ClusterState oldState, ClusterState newState)
...
+                @Override
+                public ClusterState execute(ClusterState currentState) {
+
+                    if (areClusterStatesCompatibleForRebalance(clusterState, currentState)) {
+                        return update(currentState, rebalancedMetadata);
+                    }
+                    rebalanceAssignments(currentState, modelToAdd, reason, listener);
```
benwtrent, line 300: This is tricky. I am not sure here :/. Since it asynchronously kicks off another `rebalanceAssignments` it should be ok? 

I am just wondering how this behaves when there is a flurry of updates (maybe turning on some security integrations while creating multiple allocations? Does this lock up cluster state???)

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
```diff
@@ -238,16 +241,33 @@ void loadQueuedModels() {
     }
 
     public void stopDeploymentAndNotify(TrainedModelDeploymentTask task, String reason, ActionListener<AcknowledgedResponse> listener) {
+        Optional<RoutingInfo> routingInfo = getRoutingInfo(task.getModelId());
+        if (routingInfo.isEmpty()) {
+            return;
+        }
```
benwtrent, line 247: I am not sure about this. This just indicates that the cluster state doesn't have this route. It may be that the task is still loaded and model is around. Do we know for sure that exiting here won't leave a zombie process?

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
```diff
@@ -410,31 +519,41 @@ private void handleLoadSuccess(TrainedModelDeploymentTask task) {
...
             })
         );
     }
 
-    private void updateStoredState(
-        String modelId,
-        RoutingStateAndReason routingStateAndReason,
-        ActionListener<AcknowledgedResponse> listener
-    ) {
+    private Optional<RoutingInfo> getRoutingInfo(String modelId) {
```
benwtrent, line 530: I am not sure about interrogating the clusterService.state() for routing info like this. It very well may be an older state than the one handled in the `clusterChanged` as I think the `clusterChanged` stuff is handled by every node, and then the cluster service is updated? Or it may even be vice-versa. 

Regardless, the local node routing info shouldn't be grabbed like this. I think it should be handled within the `clusterChanged` information or stored within the task object directly for the model.

Relying on the clusterService.state() will cause some really weird race conditions that need to be handled.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
```diff
@@ -458,17 +577,37 @@ private void handleLoadFailure(TrainedModelDeploymentTask task, Exception ex) {
...
+                ActionListener.wrap(r -> stopTask.run(), e -> stopTask.run())
+            );
+        }
     }
 
     public void failAssignment(TrainedModelDeploymentTask task, String reason) {
+        Optional<RoutingInfo> routingInfo = getRoutingInfo(task.getModelId());
+        if (routingInfo.isEmpty()) {
+            return;
+        }
```
benwtrent, line 601: relying on clusterService.state here is dangerous. It could be out of sync meaning the state IS there, it just hasn't been fully handled yet. Consequently we get a weird zombie task.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentRebalancer.java
```diff
@@ -0,0 +1,283 @@
...
+    ) {
+        if (assignmentPlan.satisfiesAllocations(model)) {
+            return Optional.empty();
+        }
+
+        if (nodeLoads.isEmpty()) {
+            return Optional.of("No ML nodes exist in the cluster");
+        }
+
+        Map<String, String> nodeToReason = new TreeMap<>();
```
benwtrent, line 207: ++

benwtrent (commented):
Will need to do a second pass over the tests :)

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -168,22 +163,38 @@ public void createNewModelAssignment(
         StartTrainedModelDeploymentAction.TaskParams params,
         ActionListener<TrainedModelAssignment> listener
     ) {
-        submitUnbatchedTask("create model assignment", new ClusterStateUpdateTask() {
-            @Override
-            public ClusterState execute(ClusterState currentState) {
-                return createModelAssignment(currentState, params);
-            }
+        if (clusterService.state().nodes().getMinNodeVersion().before(DISTRIBUTED_MODEL_ALLOCATION_VERSION)) {
```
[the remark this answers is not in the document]
droberts195: I would keep it simple and just check the min node version in the whole cluster. Trying to be cleverer opens up nasty edge cases like an ML node on an older version that leaves and rejoins the cluster while other nodes are being upgraded.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
```diff
@@ -363,6 +408,63 @@ public void clusterChanged(ClusterChangedEvent event) {
...
+            TrainedModelDeploymentTask task = modelIdToTask.get(assignment.getModelId());
+            if (task == null) {
+                logger.debug(() -> format("[%s] task was removed whilst updating number of allocations", task.getModelId()));
+                continue;
+            }
+            RoutingInfo routingInfo = assignment.getNodeRoutingTable().get(nodeId);
+            deploymentManager.updateNumAllocations(
+                task,
+                assignment.getNodeRoutingTable().get(nodeId).getTargetAllocations(),
+                TimeValue.timeValueSeconds(60),
```
droberts195, line 434: This should be in a constant with a name that makes clear what it controls.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
```diff
@@ -363,6 +408,63 @@ public void clusterChanged(ClusterChangedEvent event) {
...
+            .toList();
+
+        for (TrainedModelAssignment assignment : modelsToUpdate) {
+            TrainedModelDeploymentTask task = modelIdToTask.get(assignment.getModelId());
+            if (task == null) {
+                logger.debug(() -> format("[%s] task was removed whilst updating number of allocations", task.getModelId()));
+                continue;
+            }
+            RoutingInfo routingInfo = assignment.getNodeRoutingTable().get(nodeId);
+            deploymentManager.updateNumAllocations(
```
droberts195, line 431: This method doesn't necessarily call its listener.  That's probably OK, as the situation is a bit like `task == null` above where a removal is racing with the change.  I think it would be worth adding a comment here to say that, just in case in the future somebody makes the assumption that the listener will get called.  Better still, make `updateNumAllocations` pass `ResourceNotFoundException` to its listener for this case and explicitly ignore that in the error branch of the listener here.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/MachineLearning.java
```diff
@@ -462,6 +463,7 @@ public class MachineLearning extends Plugin
     public static final String PRE_V7_BASE_PATH = "/_xpack/ml/";
     public static final String DATAFEED_THREAD_POOL_NAME = NAME + "_datafeed";
     public static final String JOB_COMMS_THREAD_POOL_NAME = NAME + "_job_comms";
+    public static final String PYTORCH_COMMS_THREAD_POOL_NAME = NAME + "_pytorch_comms";
```
droberts195, line 466: It might be worth calling this "native inference" instead of "PyTorch".  If we ever created another native inference process, e.g. for Tensorflow or ONNX runtime, presumably those would need to share the same allocation logic and have their resources balanced with the PyTorch processes.  But we'd be stuck with "PyTorch" showing up in threadpool stats forever.

x-pack/plugin/ml/src/test/java/org/elasticsearch/xpack/ml/job/NodeLoadTests.java
```diff
@@ -25,6 +25,6 @@ public void testIncrementCounts() {
             .build();
 
         assertThat(nodeLoad.getNumAssignedJobs(), equalTo(6));
-        assertThat(nodeLoad.remainingJobs(), equalTo(4));
+        assertThat(nodeLoad.remainingJobs(), equalTo(7));
```
droberts195, line 28: The method names are confusing now.  "Assigned jobs" is the same as before but "remaining jobs" is different, so "jobs" must have a different meaning in two methods of the same class.

Maybe:

* `getNumAssignedJobs` -> `getNumAssignedJobsAndModels`
* `incNumAssignedNativeInferenceJobs` -> `incNumAssignedNativeInferenceModels`

x-pack/plugin/ml/src/test/java/org/elasticsearch/xpack/ml/inference/deployment/DeploymentManagerTests.java
```diff
@@ -54,12 +54,12 @@ public void managerSetup() {
...
             ),
             new ScalingExecutorBuilder(
-                JOB_COMMS_THREAD_POOL_NAME,
+                PYTORCH_COMMS_THREAD_POOL_NAME,
                 1,
                 4,
                 TimeValue.timeValueMinutes(10),
                 false,
-                "xpack.ml.job_comms_thread_pool"
+                "xpack.ml.pytorch_comms_thread_pool"
```
droberts195, line 62: As above, consider _not_ using `pytorch` in a setting name that might eventually be used for more things.

droberts195 (commented):
Looks great in general.  Just a few nits and naming issues.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/MachineLearning.java
```diff
@@ -736,6 +742,19 @@ private void addMlNodeAttribute(Settings.Builder additionalSettings, String attr
...
+        try {
+            return new OsService(settings).info().getAllocatedProcessors();
+        } catch (IOException e) {
+            // OsService may throw an IOException whilst reading the OS name.
+            // This should be an extremely rare error.
+            // It is preferrable we learn about allocated processors from OsService
+            // but if that fails for some reason we can repeat the logic here.
+            logger.debug("failed to read allocated processors from OsService", e);
+            return Math.min(EsExecutors.allocatedProcessors(settings), Runtime.getRuntime().availableProcessors());
+        }
```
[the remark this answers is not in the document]
dimitris-athanasiou: Actually, looking into what `OsService` and `OsProbe` do, we don't need to call either of them. `EsExecutors.allocatedProcessors(settings)` is all they really do so we can call just that. I'll do that.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportInferTrainedModelDeploymentAction.java
```diff
@@ -94,15 +94,13 @@ protected void doExecute(
             listener.onFailure(ExceptionsHelper.conflictStatusException(message));
             return;
         }
-        String[] randomRunningNode = assignment.getStartedNodes();
-        if (randomRunningNode.length == 0) {
+        Optional<String> randomRunningNode = assignment.selectRandomStartedNodeWeighedOnAllocations();
+        if (randomRunningNode.isEmpty()) {
```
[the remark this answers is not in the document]
dimitris-athanasiou: Definitely!

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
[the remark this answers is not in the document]
dimitris-athanasiou: I agree there's not much to gain by complicating this.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/MachineLearning.java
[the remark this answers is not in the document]
droberts195: Yes, good idea.  I just noticed that's what we are already doing for data frame analytics in https://github.com/elastic/elasticsearch/blob/93bc2e382f2623107c4d052e764819eddfcc7bcf/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/MachineLearning.java#L980

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
dimitris-athanasiou: https://github.com/elastic/elasticsearch/issues/87551

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
dimitris-athanasiou: As this is asynchronous it shouldn't block. The method will return `currentState` and the cluster state update task will finish being a noop.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
dimitris-athanasiou: Good point. I was also not notifying the listener. I'll add a call to `deploymentManager.stopDeployment` and notify the listener. This should ensure the model is cleaned up in this edge case.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
dimitris-athanasiou: I can probably eliminate the need to get the current `RoutingInfo` by changing the update request to only take the bit we're changing. I'll give that a try by introducing a `RoutingInfoUpdate` object.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/MachineLearning.java
```diff
@@ -736,6 +742,19 @@ private void addMlNodeAttribute(Settings.Builder additionalSettings, String attr
         }
     }
 
+    private int getAllocatedProcessors() {
+        try {
+            return new OsService(settings).info().getAllocatedProcessors();
+        } catch (IOException e) {
+            // OsService may throw an IOException whilst reading the OS name.
```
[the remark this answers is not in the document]
dimitris-athanasiou: I'll replace this with simply `EsExecutors.allocatedProcessors(settings)`.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
dimitris-athanasiou: `DeploymentManager.updateNumAllocations` will indeed call the listener.onFailure if the process context is missing. It passes `listener::onFailure` in the error consumer of the `getProcessContext` method.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentRebalancer.java
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    private Optional<String> explainAssignment(
+        AssignmentPlan assignmentPlan,
+        DiscoveryNode node,
+        NodeLoad load,
+        AssignmentPlan.Model model
+    ) {
+        if (Strings.isNullOrEmpty(load.getError()) == false) {
+            logger.warn("[{}] failed to calculate current node load with error [{}]", model.id(), load.getNodeId());
```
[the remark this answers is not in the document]
dimitris-athanasiou: Yes, indeed. It used to do that previously. I think the rationale of logging a warning is that this scenario should not ever happen really. There is no reason for us to fail to detect node load unless something really horrible has happened.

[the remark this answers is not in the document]
dimitris-athanasiou: I will move the logging into `NodeLoadDetector` where it will better belongs.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentRebalancer.java
```diff
@@ -0,0 +1,283 @@
...
+            return Optional.of(load.getError());
+        }
+
+        int nodeProcessors = getNodeAllocatedProcessors(node).orElse(0);
+
+        // If any ML processes are running on a node we require some space to load the shared libraries.
+        // So if none are currently running then this per-node overhead must be added to the requirement.
+        boolean isPerNodeOverheadAccountedFor = load.getNumAssignedJobs() > 0
+            && assignmentPlan.getRemainingNodeCores(load.getNodeId()) < nodeProcessors;
```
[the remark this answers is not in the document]
dimitris-athanasiou: If we have not assigned any model on a node, then the remaining cores will be == to the node allocated processors. So, this second statement is meant to capture the case where the rebalancer assigned a model to a node where it was previously unassigned and therefore not visible in the node load. However, the second statement should be linked with `||` and not `&&`, so thanks for pointing it out!

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentRebalancer.java
```diff
@@ -0,0 +1,283 @@
...
+        long requiredMemory = model.memoryBytes() + (isPerNodeOverheadAccountedFor
+            ? 0
+            : MachineLearning.NATIVE_EXECUTABLE_CODE_OVERHEAD.getBytes());
+        long nodeFreeMemory = assignmentPlan.getRemainingNodeMemory(node.getId()) + (isPerNodeOverheadAccountedFor
```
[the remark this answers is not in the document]
dimitris-athanasiou: This is right to be added. The reason is that `getRemainingNodeMemory` doesn't include the overhead. Thus, we need to add it back in when we report the memory in the error message.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentRebalancer.java
```diff
@@ -0,0 +1,283 @@
...
+            ? 0
+            : MachineLearning.NATIVE_EXECUTABLE_CODE_OVERHEAD.getBytes());
+        if (model.memoryBytes() > assignmentPlan.getRemainingNodeMemory(node.getId())) {
```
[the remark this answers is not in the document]
dimitris-athanasiou: The comparison the assignment planner will do is that one. Thus, I'd rather keep that check. But you're right, the calculation of the displayed values can be moved inside the `if`. I'll do that.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
dimitris-athanasiou: I have now pushed a commit that allows partial updates to routing info and eliminates the need to retrieve a routing info based on some state.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
dimitris-athanasiou: I have now pushed a commit that allows partial updates to routing info and eliminates the need to retrieve a routing info based on some state.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/assignment/TrainedModelAssignment.java
```diff
@@ -153,6 +156,43 @@ public String[] getStartedNodes() {
...
+
+        int randomInt = Randomness.get().ints(1, 1, allocationSum + 1).iterator().nextInt();
+        int nodeIndex = Collections.binarySearch(cumulativeAllocations, randomInt);
+        if (nodeIndex < 0) {
+            nodeIndex = -nodeIndex - 1;
+        }
+        return Optional.of(nodeIds.get(nodeIndex));
+    }
+
+    public Optional<String> selectRandomStartedNodeVersionedBefore(
```
davidkyle, line 183: This is the fallback method used if `selectRandomStartedNodeWeighedOnAllocations` returns `empty` isn't it and the purpose is to pick a node for the inference request?

Why do the version filter when any node in the `STARTED` state could be used in the inference request

davidkyle (approved):
I left a few comments but in general LGTM

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/assignment/TrainedModelAssignment.java
```diff
@@ -153,6 +156,43 @@ public String[] getStartedNodes() {
...
+        List<String> nodeIds = new ArrayList<>(nodeRoutingTable.size());
+        List<Integer> cumulativeAllocations = new ArrayList<>(nodeRoutingTable.size());
+        int allocationSum = 0;
+        for (Map.Entry<String, RoutingInfo> routingEntry : nodeRoutingTable.entrySet()) {
+            if (RoutingState.STARTED.equals(routingEntry.getValue().getState())) {
+                nodeIds.add(routingEntry.getKey());
+                allocationSum += routingEntry.getValue().getCurrentAllocations();
+                cumulativeAllocations.add(allocationSum);
+            }
+        }
```
benwtrent, line 169: OK, I am wondering here, since we do the additional check regarding to previous versions, why can't we weigh old nodes with started deployments with an allocation value of 1?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/assignment/TrainedModelAssignment.java
```diff
@@ -153,6 +156,43 @@ public String[] getStartedNodes() {
...
+    }
+
+    public Optional<String> selectRandomStartedNodeVersionedBefore(
+        Function<String, Optional<Version>> nodeIdToVersionFunction,
+        Version version
+    ) {
+        List<String> nodeIds = nodeRoutingTable.entrySet()
+            .stream()
+            .filter(e -> e.getValue().getState() == RoutingState.STARTED)
+            .filter(e -> nodeIdToVersionFunction.apply(e.getKey()).orElse(version).before(version))
```
benwtrent, line 190: This is a non-obvious 'optimization' because we have already checked the "new" nodes and it gave us nothing. But this is assuming a specific behavior from the caller of this method. 

I think either:
 - we don't do this version check
 - We merge this logic into the other function
 - We weigh "old nodes" that are started as having 1 allocation.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -103,47 +109,43 @@ public void clusterChanged(ClusterChangedEvent event) {
...
-                    // and components are created before persistent task executors, we will try to allocate
-                    // trained models before we try to assign ML persistent tasks.
-                    return addRemoveAssignmentNodes(currentState);
-                }
-
-                @Override
-                public void onFailure(Exception e) {
-                    logger.warn("failed to allocate models", e);
-                }
+        if (clusterService.state().nodes().getMinNodeVersion().before(DISTRIBUTED_MODEL_ALLOCATION_VERSION)) {
```
benwtrent, line 112: We should use `event` here, not `clusterService`.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -103,47 +109,43 @@ public void clusterChanged(ClusterChangedEvent event) {
...
+            // it may get re-allocated to that node when another node is added/removed...
+            //
+            // As this produces a cluster state update task, we are certain that if the persistent
+            // task framework results in assigning some ML tasks on that same cluster state change
+            // we do not end up over-allocating a node. Both this service and the persistent task service
+            // will produce a cluster state update but the one that gets applied first wins. The other
+            // update will be rejected and we will retry to assign getting a correct update on available memory
+            // on each node.
+            rebalanceAssignments(
+                clusterService.state(),
```
benwtrent, line 131: why are we using `clusterService.state()` here and not the ones provided in the event?

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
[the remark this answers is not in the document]
benwtrent: Since we don't really support "downgrades", I think checking for min node version here like this against cluster service state is acceptable.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -168,22 +170,38 @@ public void createNewModelAssignment(
...
+                )
+            );
+            return;
+        }
 
-            @Override
-            public void onFailure(Exception e) {
-                listener.onFailure(e);
-            }
+        if (MlMetadata.getMlMetadata(clusterService.state()).isResetMode()) {
```
benwtrent, line 185: This is an ok check, but we should totally make the same check against the cluster state seen by the update task after the rebalancer is ran.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -245,55 +263,107 @@ public void clusterStateProcessed(ClusterState oldState, ClusterState newState)
...
-        }
-        if (nodeToReason.isEmpty() == false) {
-            assignmentBuilder.setReason(
-                nodeToReason.entrySet()
-                    .stream()
-                    .map(entry -> String.format(Locale.ROOT, "Not allocating on node [%s]. Reason: %s", entry.getKey(), entry.getValue()))
-                    .collect(Collectors.joining("|"))
+            submitUnbatchedTask(reason, new ClusterStateUpdateTask() {
+                @Override
+                public ClusterState execute(ClusterState currentState) {
```
benwtrent, line 302: Check for reset mode here and return the same state if reset mode is enabled

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
benwtrent: FYI, 

```
        state.set(newClusterState);

        callClusterStateListeners(clusterChangedEvent, stopWatch);
```
It seems to be that the `clusterService.state` is set and then all the listeners called. But I am not sure I would rely on that and use the `event` state wherever possible to prevent accidental race conditions.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
benwtrent: My main concern is that a bunch of different allocations start, a flurry of callbacks keep executing and we cripple the cluster.

benwtrent: Batching should alleviate this as all the tasks of the same type would be handled at the same time (and thus the number of times we do this async double check would be reduced dramatically).

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
```diff
@@ -88,6 +92,7 @@ public TrainedModelAssignmentNodeService(
         XPackLicenseState licenseState
     ) {
         this.trainedModelAssignmentService = trainedModelAssignmentService;
+        this.clusterService = clusterService;
```
benwtrent, line 95:
```suggestion
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
```diff
@@ -64,8 +66,10 @@ public class TrainedModelAssignmentNodeService implements ClusterStateListener {
     private static final String NODE_NO_LONGER_REFERENCED = "node no longer referenced in model routing table";
     private static final String ASSIGNMENT_NO_LONGER_EXISTS = "model assignment no longer exists";
     private static final TimeValue MODEL_LOADING_CHECK_INTERVAL = TimeValue.timeValueSeconds(1);
+    private static final TimeValue UPDATE_NUMBER_OF_ALLOCATIONS_TIMEOUT = TimeValue.timeValueSeconds(60);
     private static final Logger logger = LogManager.getLogger(TrainedModelAssignmentNodeService.class);
     private final TrainedModelAssignmentService trainedModelAssignmentService;
+    private final ClusterService clusterService;
```
benwtrent, line 72:
```suggestion
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentNodeService.java
```diff
@@ -120,6 +125,7 @@ public void beforeStop() {
         XPackLicenseState licenseState
     ) {
         this.trainedModelAssignmentService = trainedModelAssignmentService;
+        this.clusterService = clusterService;
```
benwtrent, line 128:
```suggestion
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentRebalancer.java
```diff
@@ -0,0 +1,286 @@
...
+        this.currentMetadata = Objects.requireNonNull(currentMetadata);
+        this.nodeLoads = Objects.requireNonNull(nodeLoads);
+        this.modelToAdd = Objects.requireNonNull(modelToAdd);
+    }
+
+    TrainedModelAssignmentMetadata.Builder rebalance() {
+        TrainedModelAssignmentMetadata.Builder builder = TrainedModelAssignmentMetadata.Builder.fromMetadata(currentMetadata);
+        if (modelToAdd.isPresent() && builder.hasModel(modelToAdd.get().getModelId())) {
+            throw new ResourceAlreadyExistsException("assignment for model with id [{}] already exists", modelToAdd.get().getModelId());
+        }
```
benwtrent, line 59: I was initially worried that we would't catch this error  anywhere, but I think we are ok double checking all the places in ClusterState, I just don't want this to be an unhandled exception within our threads that ends up causing issues :/

x-pack/plugin/ml/src/test/java/org/elasticsearch/xpack/ml/inference/deployment/DeploymentManagerTests.java
```diff
@@ -54,12 +54,12 @@ public void managerSetup() {
...
             ),
             new ScalingExecutorBuilder(
-                JOB_COMMS_THREAD_POOL_NAME,
+                NATIVE_INFERENCE_COMMS_THREAD_POOL_NAME,
                 1,
                 4,
                 TimeValue.timeValueMinutes(10),
                 false,
-                "xpack.ml.job_comms_thread_pool"
+                "xpack.ml.pytorch_comms_thread_pool"
```
benwtrent, line 62:
```suggestion
                "xpack.ml.native_inference_comms_thread_pool"
```

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/assignment/TrainedModelAssignment.java
dimitris-athanasiou: If a node is upgraded and it had an assignment, when the node is brought back we do not change its state from `STARTED` to `STARTING`, or remove its entry from the routing table. We do not want to run the balancer while the cluster hasn't been fully upgraded. Thus, the model won't be loaded until the full cluster has been upgraded. This is a scenario that only affects mixed clusters from before 8.4 and after 8.4+. We could try to change the state to `STARTING` but even doing that would mean there'll be a window of time when the state will be `STARTED` and we could end up routing a request to that node and subsequently get a failure. A bit tricky but this is one of the easiest fixes to having inference requests routed to nodes running the model in a mixed cluster situation.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentRebalancer.java
```diff
@@ -0,0 +1,286 @@
...
+
+            Map<AssignmentPlan.Node, Integer> assignments = assignmentPlan.assignments(model).orElseGet(Map::of);
+            for (Map.Entry<AssignmentPlan.Node, Integer> assignment : assignments.entrySet()) {
+                if (existingAssignment != null && existingAssignment.isRoutedToNode(assignment.getKey().id())) {
+                    RoutingInfo existingRoutingInfo = existingAssignment.getNodeRoutingTable().get(assignment.getKey().id());
+                    RoutingState state = existingRoutingInfo.getState();
+                    String reason = existingRoutingInfo.getReason();
+                    if (state == RoutingState.FAILED) {
+                        state = RoutingState.STARTING;
+                        reason = "";
```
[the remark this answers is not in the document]
dimitris-athanasiou: We already had a retrying mechanism for models that failed to load. If the process crashes, it seems to me we should definitely try to reload the model.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
dimitris-athanasiou: +++

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -109,9 +110,11 @@ public void clusterChanged(ClusterChangedEvent event) {
         if (event.state().blocks().hasGlobalBlock(GatewayService.STATE_NOT_RECOVERED_BLOCK)) {
             return;
         }
-        if (clusterService.state().nodes().getMinNodeVersion().before(DISTRIBUTED_MODEL_ALLOCATION_VERSION)) {
+        if (event.state().nodes().getMinNodeVersion().before(DISTRIBUTED_MODEL_ALLOCATION_VERSION)) {
             // we should not try to rebalance assignments while there may be nodes running on a version
             // prior to introducing distributed model allocation.
+            // But we should remove routing to removed or shutting down nodes.
+            removeRoutingToRemovedOrShuttingDownNodes(event);
```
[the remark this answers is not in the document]
dimitris-athanasiou: Correct. We will compute a new assignment plan which won't contain the crashed node.

dimitris-athanasiou:
@elasticmachine update branch

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -105,49 +111,125 @@ public void clusterChanged(ClusterChangedEvent event) {
...
-            submitUnbatchedTask("allocating models to nodes", new ClusterStateUpdateTask() {
+        if (event.localNodeMaster() == false) {
+            return;
+        }
+
+        if (event.state().nodes().getMinNodeVersion().before(DISTRIBUTED_MODEL_ALLOCATION_VERSION)) {
+            // we should not try to rebalance assignments while there may be nodes running on a version
+            // prior to introducing distributed model allocation.
+            // But we should remove routing to removed or shutting down nodes.
+            removeRoutingToRemovedOrShuttingDownNodes(event);
```
benwtrent, line 122: I know I am being paranoid here. But I want to be 100% sure. 

When a node is removed, routes are only rebalanced IFF one of the model's assigned to that node are no longer satisfied. 

I think this is always the case as we never over-allocate (so any loss is now no longer satisfied). Just wanted to capture my concern here :)

benwtrent (approved):

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
dimitris-athanasiou: Correct. The trick is in `TrainedModelAssignmentClusterService#shouldRebalanceModels`. We only return `true` if the removed node was hosting allocations. Indeed, that should be enough for us to rebalance as it would mean that at least that model is not satisfied as we do not over-allocate.

PR #87382, opened 2022-06-03
[Doc] User Profile - Improve docs for user profile APIs

ywangd:
This PR adds docs for the new hint parameter of the suggestProfiles API.
It also creates proper test setup so that most API call examples in user
profile docs are now runnable.

### DIFF
--- a/x-pack/docs/build.gradle
+++ b/x-pack/docs/build.gradle
@@ -742,4 +742,95 @@ tasks.named("buildRestTests").configure { buildRestTests ->
         service: fleet-server
         name: token42
 '''
+  setups['user_profiles'] = '''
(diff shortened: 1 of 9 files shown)


sethmlarson (approved):
The `hint` parameter in the `security.suggest_user_profiles` API LGTM, thanks for adding a test.

albertzaharovits (approved):
LGTM

ywangd:
Thanks both for review. @lockewritesdocs This needs to be 8.3 and it is also needed to block a few other PRs, so I'll merge it as is (after current CI passes). We can always come back to it later especially since it is still behind the feature flag. Thanks!

PR #87387, opened 2022-06-03
TEST yml REST test for  the profile has privileges API

albertzaharovits:
yml REST test for the profile has privilege API from https://github.com/elastic/elasticsearch/pull/85898.


rest-api-spec/src/main/resources/rest-api-spec/api/security.has_privileges_user_profile.json
```diff
@@ -0,0 +1,31 @@
+{
+  "security.has_privileges_user_profile":{
+    "documentation":{
+      "url":"https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-has-privileges-user-profile.html",
+      "description":"Determines whether the users with the specified profile IDs have all the requested privileges."
```
ywangd, line 5:
```suggestion
      "description":"Determines whether the users associated with the specified profile IDs have all the requested privileges."
```

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/user_profile/40_has_privileges.yml
```diff
@@ -0,0 +1,176 @@
...
+          privileges:
+            index:
+              - names:
+                  - index1*
+                  - index2*
+                privileges:
+                  - read
+  - length: { "has_privilege_uids": 2 }
+  - match: { "has_privilege_uids.0" : $profile_uid2 }
+  - match: { "has_privilege_uids.1" : $profile_uid1 }
```
ywangd, line 139: This is not stable since `has_privilege_uids` is a `Set`. You might have to use regex for it. It is not pretty but works. Here is an example:
https://github.com/elastic/elasticsearch/blob/23c41e45d9a20f43220d7a50d00dd373ab774394/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/10_basic.yml#L273

ywangd (approved):
LGTM

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/user_profile/40_has_privileges.yml
albertzaharovits: Thanks for the suggestion.

Upon further thought I think it would be reasonable to maintain the order from the request (the internal mget does preserve it). I vaguely remember considering it but I don't recall where we ended up with that. I think it would be easy to do.

Let me know what you think.

PR #87392, opened 2022-06-03
Fix a bug with flattened fields in terms aggregations

not-napoleon:
The bug occurs when terms incorrectly chooses the `LowCardinality` optimization with a flattened field.  This can only happen if `supportsGlobalOrdinalsMapping` is true, which it usually is not for flattened fields.  But, when we wrap the `ValuesSource` to add in the missing value, we weren't delegating that method, which led to this bug.

Draft because I still need to do some more testing, and probably `supportsGlobalOrdinalsMapping` should be abstract so this doesn't happen anywhere else.


nik9000:
> Draft because I still need to do some more testing, and probably `supportsGlobalOrdinalsMapping` should be abstract so this doesn't happen anywhere else.

:+1:

server/src/main/java/org/elasticsearch/search/aggregations/bucket/terms/TermsAggregatorFactory.java
```diff
@@ -47,6 +49,8 @@
 public class TermsAggregatorFactory extends ValuesSourceAggregatorFactory {
     static Boolean REMAP_GLOBAL_ORDS, COLLECT_SEGMENT_ORDS;
 
+    protected static final Logger logger = LogManager.getLogger(TermsAggregatorFactory.class);
```
nik9000, line 52: I'd go with `private` just for a bit of paranoia.

For what it's worth `LogManger.getLogger()` is very very fast. I still use the `getLogger(class)` mostly though.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/terms/TermsAggregatorFactory.java
```diff
@@ -130,6 +134,7 @@ public Aggregator build(
                 }
 
                 // TODO: [Zach] we might want refactor and remove ExecutionMode#create(), moving that logic outside the enum
+                logger.info("Creating bytes terms aggregator with execution mode [" + execution + "]");
```
nik9000, line 137: These'd probably be fine to keep with `debug` if you like`. Though you should do the `[{}]` substitution thing.

server/src/main/java/org/elasticsearch/search/aggregations/support/MissingValues.java
```diff
@@ -225,6 +225,11 @@ public LongUnaryOperator globalOrdinalsMapping(LeafReaderContext context) throws
                 );
             }
 
+            @Override
+            public boolean supportsGlobalOrdinalsMapping() {
+                return valuesSource.supportsGlobalOrdinalsMapping();
+            }
```
nik9000, line 231: :boom:

nik9000 (approved):

PR #87393, opened 2022-06-03
Use Map instead of ImmutableOpenMap in ClusterInfo

rjernst:
The internals of ClusterInfo are several maps. Currently these are
ImmutableOpenMaps. This commit converts them to Map.

relates #86239


rjernst:
Note that currently the implementation type is still ImmutableOpenMap, since these come from InternalClusterInfoService. This PR only broadens the type used in ClusterInfo.

original-brownbear (approved):

PR #87396, opened 2022-06-03
Force property expansion for security policy

rjernst:
When resolving the security policy files for server and components of
Elasticsearch, each jar file location is put into a special system
property value so that policy files may contain codeBase specific
grants. The mechanism for substituting system properties is part of the
JDK's policy parser. However, a security property exists,
policy.expandProperties, which controls whether properties will actually
be expanded. If a user ends up setting this, Elasticsearch will fail to
start.

This commit forces the value of the security property to ensure the
policy files can always be parsed correctly.

### DIFF
--- a/docs/changelog/85688.yaml
+++ b/docs/changelog/85688.yaml
@@ -1,0 +1,5 @@
+pr: 85688
+summary: Speed up `NumberFieldMapper`
+area: Mapping
+type: enhancement
+issues: []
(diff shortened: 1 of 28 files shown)


ChrisHegarty (commented):
The source code changes look good to me. Are the doc changes accidentally or a bad merge?

rjernst:
Not sure why those docs changes were showing up. I synced master and it seems to be ok now.

ChrisHegarty (approved):
LGTM

grcevski (approved):
LGTM!

PR #87397, opened 2022-06-04
Move spatial3d dependency to spatial

### LINKED ISSUE #87026: Rework lucene spatial3d dependency on server
The server module has a dependency on lucene spatial3d library but it is never used. The only dependency on that library happens in the spatial module which is accessed via the server module at the moment. There might be a better way to handle this dependency so opening this issue to track it.

rjernst:
Server depends on spatial3d, but it is only ever used by the spatial
xpack component. This commit moves the dependency there.

closes #87026


modules/legacy-geo/build.gradle
```diff
@@ -51,7 +51,26 @@ tasks.named("thirdPartyAudit").configure {
...
+    'org.apache.lucene.spatial3d.geom.GeoCircleFactory',
+    'org.apache.lucene.spatial3d.geom.GeoCompositeAreaShape',
+    'org.apache.lucene.spatial3d.geom.GeoPathFactory',
+    'org.apache.lucene.spatial3d.geom.GeoPoint',
+    'org.apache.lucene.spatial3d.geom.GeoPointShape',
+    'org.apache.lucene.spatial3d.geom.GeoPointShapeFactory',
+    'org.apache.lucene.spatial3d.geom.GeoPolygonFactory',
+    'org.apache.lucene.spatial3d.geom.GeoS2ShapeFactory',
+    'org.apache.lucene.spatial3d.geom.LatLonBounds',
+    'org.apache.lucene.spatial3d.geom.PlanetModel',
```
iverase, line 72: This classes come from the Geo3D implementation on lucene-extras that lucene-geo module is not using,

iverase (approved):
LGTM

PR #87406, opened 2022-06-05
Reference the entire example plugins dir in docs

rjernst:
The plugin author docs refer to a few example plugins, yet these are
just a few of the existing examples. While an overhaul is needed of
these docs, this commit slightly improves the situation by making it
clear that these are not the only examples, and linking to the entire
examples directory. Additionally, the wording is tweaked to remove the
superfluous use of "java plugin" because there is no longer a non java
plugin as there was in 1.x.


grcevski (approved):
LGTM!

PR #87408, opened 2022-06-06
[Doc] Add callout for security domain usage

ywangd:
Security domain is an advance feature and must be used with care. This
PR adds a section in the doc page to call it out.

[Doc preview](https://elasticsearch_87408.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/security-domain.html#security-domain-realm-roles)


x-pack/docs/en/security/authentication/security-domain.asciidoc
```diff
@@ -41,24 +41,52 @@ users with the same username, it's reasonable to treat these users as the
...
+If the owner user's realm is under a domain, the profile is shared across realms in that
+domain for the user. Other users that are no longer in the same domain as the owner user will get
+new profiles created the next time they call the Activate User Profile API.
+  - Other resources, such as API keys, will be owned by the user who originally created them.
+If the owner user's realm is under a domain, the resource is shared across realms in that
+domain for the user.
+
+NOTE: The "same username" rule is always respected. That is, resource sharing across realms
+is only possible between users with the same username under all circumstances.
+====
```
ywangd, line 88: This is the only real change. Other places are due to formatting issues.

x-pack/docs/en/security/authentication/security-domain.asciidoc
```diff
@@ -41,24 +41,52 @@ users with the same username, it's reasonable to treat these users as the
...
+[[security-domain-warning]]]]
+[IMPORTANT]
+====
+Security domain is an advanced feature and must be used with care. It can lead
+to unexpected behaviours if incorrectly configured or misused.
+Here are few things watch out when working with security domain:
+
+* Security domains must be configured consistently across all nodes. Since changing
+security domain requires node restart, this means they may not be fully functional
+during rolling restart.
```
tvernum, line 71: Can we be more specific here?
As it reads, it says "be careful" and "rolling restarts might not work well", but doesn't tell the admin how to avoid potential problems. Do we just expect them not to do any changes? Or do full cluster restarts?

It would be nice to tell them something that they can actually respond to (if that's possible).

x-pack/docs/en/security/authentication/security-domain.asciidoc
```diff
@@ -41,24 +41,52 @@ users with the same username, it's reasonable to treat these users as the
...
+* Security domains are better at grouping multiple realms, but not so much at splitting them afterwards.
```
tvernum, line 72: Maybe something like

```suggestion
* Security domains are designed to be effective for grouping multiple realms. Splitting existing
realms out of a security domain is more complex and can introduce unexpected behaviors.
```

x-pack/docs/en/security/authentication/security-domain.asciidoc
```diff
@@ -41,24 +41,52 @@ users with the same username, it's reasonable to treat these users as the
...
+This means assigning realms under a domain works as expected (assuming they are configured
+consistently across all nodes).
+But removing realms from a domain must be proceeded with caution. Resources
```
tvernum, line 75:
```suggestion
consistently across all nodes), but removing realms from a domain must be
performed with caution. Resources
```

x-pack/docs/en/security/authentication/security-domain.asciidoc
```diff
@@ -41,24 +41,52 @@ users with the same username, it's reasonable to treat these users as the
...
+created before domain splitting can be owned by different users depending on resource types:
+  - <<user-profile,User profile>> will be owned by the user who performs the last
+<<security-api-activate-user-profile,Activate User Profile>> request.
```
tvernum, line 78:
```suggestion
  - <<user-profile,User profiles>> will be owned by the user for whom it was last
<<security-api-activate-user-profile,activated>>.
```

x-pack/docs/en/security/authentication/security-domain.asciidoc
ywangd: I expanded a bit (dos and donts) on this.

ywangd:
Nested bullet points are tricky to get right and they do not look really great in a callout block. So I removed the top level bullet points and the list is now flat.

albertzaharovits (commented):
At a high-level, I think this is too verbose about the corner cases, the risk of which is that it all gets ignored. My suggestion is to be more succint about the warnings and disclaimers.

Personally, I would've opted to just mention that "it is a requirement that all the coordinating nodes (eg nodes that Kibanas/Fleets are connected to), have the same security domain configuration (i.e. the domains contain the same realms, by name and type), at all times (while serving requests). Failing the above, profile activation errors are expected".

The domain split part is OKish. It's factually correct.
I wish we could find a better story for profiles, the story for API Keys (that whichever realm created it, gets to keep it after the split) is reasonable.

I think this one is a case where we can argue about the form for ages, and it would benefit from the attention of someone from the docs writers team.

x-pack/docs/en/security/authentication/security-domain.asciidoc
```diff
@@ -41,24 +41,68 @@ users with the same username, it's reasonable to treat these users as the
...
+If the owner user's realm is under a domain, the profile is shared across realms in that
+domain for the user. Other users that are no longer in the same domain as the owner user will get
+new profiles created the next time they call the Activate User Profile API.
+- Other resources, such as API keys, will be owned by the user who originally created them.
+If the owner user's realm is under a domain, the resource is shared across realms in that
+domain for the user.
+
+NOTE: The "same username" rule is always respected. That is, resource sharing across realms
+is only possible between users with the same username under all circumstances.
+====
```
debadair, line 104: I think this is too much to try put in an admonition. I'd suggest incorporating most of it into the Configure a security domain section, and adding a section with the info about removing realms being problematic. See https://github.com/elastic/elasticsearch/commit/23710d55ba8fc5c3be1abe2d8f9abaa7bdf74f90.

debadair (changes requested):
I'd refactor this info rather than stuffing it all in a note. It's hard to suggest changes like that, so I pushed them to a branch in my fork.

x-pack/docs/en/security/authentication/security-domain.asciidoc
ywangd: Thanks Deb. The changes look good to me. They retain the same central idea but present it in a much better format. Do you mind directly pushing to this PR? Alternatively, I can incorporate them on my side. Please let me know,.

ywangd:
@debadair I pushed some [minor changes](https://github.com/elastic/elasticsearch/pull/87408/commits/52d8b642fcea185f9e38e94ed4dc6d2f2da91212):
* Delete some of the extra words which seem to be not entirely accurate and are just better off without them.
* Suggest "disabling" realms instead of "deleting" realms if necessary.

ywangd:
@elasticmachine update branch

ywangd:
@tvernum @debadair Could you please take another look? Thanks!

debadair (approved):
LGTM

tvernum (approved):

PR #87412, opened 2022-06-06
Expose segment details in PCSS debug log

DaveCTurner:
Elasticsearch 8.x won't start if the cluster state on disk contains 6.x
indices. We've seen a couple of cases where users shut down one or more
nodes, then delete their older indices, and then report that they cannot
upgrade the shut down nodes since those nodes (obviously) didn't hear
that the old indices were deleted. This is a pain to diagnose today.

With this commit we expose the details of the segments making up the
cluster state index in the `DEBUG` logs so we can see their timestamps
and other metadata which is needed to diagnose this kind of situation.


server/src/main/java/org/elasticsearch/gateway/PersistedClusterStateService.java
```diff
@@ -415,6 +420,39 @@ OnDiskState loadBestOnDiskState(boolean checkClean) throws IOException {
                     }
 
                     try (DirectoryReader directoryReader = DirectoryReader.open(directory)) {
+
+                        if (logger.isDebugEnabled()) {
+                            final var segmentsFileName = directoryReader.getIndexCommit().getSegmentsFileName();
```
tlrx, line 425: In the context of the recent issue I had to troubleshoot, logging the `IndexCommit`'s user data map would have been very helpful. Can we add those too?

DaveCTurner: Good point, added in 3af38cf292d.

tlrx (approved):
LGTM, thanks a lot David

PR #87414, opened 2022-06-06
Handle ordering in plain highlighter for multiple inputs

### LINKED ISSUE #87210: Plain Highlighter “order”:”score” is not working for fields with multiple values
### Environment:
Elasticsearch version:
```
{
    "number" : "7.15.0",
    "build_flavor" : "default",
    "build_type" : "zip",
    "build_hash" : "79d65f6e357953a5b3cbcc5e2c7c21073d89aa29",
    "build_date" : "2021-09-16T03:05:29.143308416Z",
    "build_snapshot" : false,
    "lucene_version" : "8.9.0",
    "minimum_wire_compatibility_version" : "6.8.0",
    "minimum_index_compatibility_version" : "6.0.0-beta1"
}
```
Plugins installed: []
JVM version :1.8
OS version :Windows 10

### Steps to reproduce:
```
PUT test_plain_highlighter 
{}

POST /test_plain_highlighter/_doc/1
{
  "description": [
    "Lorem Ipsum string Generator that helps to create dummy text for all layout needs.",
    "It has roots in a piece of classical Latin literature from 45 BC, making it search string over 2000 years old."
  ]
}

POST /test_plain_highlighter/_search
{
  "highlight": {
    "order": "score",
    "type": "plain",
    "fields": {
      "*": {}
    },
    "fragment_size": 50
  },
  "query": {
    "match": {
      "description": "search string"
    }
  },
  "_source": ""
}
```

### RECEIVED Highlighted Fragment in response:
```
[
   "Lorem Ipsum <em>string</em> Generator that helps to create",
   " literature from 45 BC, making it <em>search</em> <em>string</em> over 2000 years old."
]
```
### EXPECTED Highlighted Fragment:
```
[
   " literature from 45 BC, making it <em>search</em> <em>string</em> over 2000 years old.",
   "Lorem Ipsum <em>string</em> Generator that helps to create"
]
```

romseygeek:
The ordering logic in the plain highlighter only worked for single-valued
inputs.  If you had multiple inputs, then order-by-score would only do this
ordering for fragments within each input, and order-by-position would 
interleave fragments from the different inputs.  This commit fixes both
issues.

Fixes #87210

### DIFF
--- a/server/src/main/java/org/elasticsearch/search/fetch/subphase/highlight/PlainHighlighter.java
+++ b/server/src/main/java/org/elasticsearch/search/fetch/subphase/highlight/PlainHighlighter.java
@@ -170,12 +170,12 @@ public class PlainHighlighter implements Highlighter {
             }
         }
 
-        // For single text inputs, the fragments are already ordered by score.  If we have multiple
+        // For single text inputs, the fragments are already ordered by score. If we have multiple
         // inputs, or if we are ordering by fragment number, then we need to resort the output list
         if (textsToHighlight.size() > 1 || field.fieldOptions().scoreOrdered() == false) {
-            Comparator<OrderedTextFragment> comparator = field.fieldOptions().scoreOrdered() ?
-                Comparator.comparingDouble(OrderedTextFragment::score).reversed() :
-                Comparator.comparingInt(OrderedTextFragment::fragNum);
+            Comparator<OrderedTextFragment> comparator = field.fieldOptions().scoreOrdered()
+                ? Comparator.comparingDouble(OrderedTextFragment::score).reversed()
+                : Comparator.comparingInt(OrderedTextFragment::fragNum);
             fragsList.sort(comparator);
         }
 
--- a/server/src/main/java/org/elasticsearch/search/fetch/subphase/highlight/PlainHighlighter.java
+++ b/server/src/main/java/org/elasticsearch/search/fetch/subphase/highlight/PlainHighlighter.java
@@ -22,7 +22,6 @@ import org.apache.lucene.search.highlight.SimpleHTMLFormatter;
 import org.apache.lucene.search.highlight.SimpleSpanFragmenter;
 import org.apache.lucene.search.highlight.TextFragment;
 import org.apache.lucene.util.BytesRefHash;
-import org.apache.lucene.util.CollectionUtil;
 import org.elasticsearch.common.lucene.Lucene;
 import org.elasticsearch.common.text.Text;
 import org.elasticsearch.index.IndexSettings;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/elasticsearch/search/fetch/subphase/highlight/PlainHighlighter.java
```diff
@@ -157,10 +170,15 @@ public HighlightField highlight(FieldHighlightContext fieldContext) throws IOExc
...
-        // fragments are ordered by score by default since we add them in best
-        if (field.fieldOptions().scoreOrdered() == false) {
-            CollectionUtil.introSort(fragsList, (o1, o2) -> o1.getFragNum() - o2.getFragNum());
+        // For single text inputs, the fragments are already ordered by score. If we have multiple
+        // inputs, or if we are ordering by fragment number, then we need to resort the output list
+        if (textsToHighlight.size() > 1 || field.fieldOptions().scoreOrdered() == false) {
+            Comparator<OrderedTextFragment> comparator = field.fieldOptions().scoreOrdered()
+                ? Comparator.comparingDouble(OrderedTextFragment::score).reversed()
+                : Comparator.comparingInt(OrderedTextFragment::fragNum);
+            fragsList.sort(comparator);
```
cbuescher, line 179: Just out of curiosity, did you consider making the list sorted from the start or would this have major drawbacks? I didn't clearly check whether sorting is always done here, maybe not and thats the reason why we don't need a sorted list from the start...

server/src/main/java/org/elasticsearch/search/fetch/subphase/highlight/PlainHighlighter.java
```diff
@@ -157,10 +170,15 @@ public HighlightField highlight(FieldHighlightContext fieldContext) throws IOExc
             }
         }
 
-        // fragments are ordered by score by default since we add them in best
-        if (field.fieldOptions().scoreOrdered() == false) {
-            CollectionUtil.introSort(fragsList, (o1, o2) -> o1.getFragNum() - o2.getFragNum());
```
cbuescher, line 162: Also just curious about the switch from Lucene CollectionUtil to plain list sort here, any pros / cons about the former?

cbuescher (approved):
Left to questions for my own education, otherwise LGTM

server/src/main/java/org/elasticsearch/search/fetch/subphase/highlight/PlainHighlighter.java
romseygeek: The highlighter outputs fragments sorted by score, so if we only have a single input and we want score order then there's no need to do anything.  Otherwise we need to re-sort

server/src/main/java/org/elasticsearch/search/fetch/subphase/highlight/PlainHighlighter.java
romseygeek: Just that it means we get to use built-in collections; it's possible that lucene's introSort is slightly more efficient, but we're talking tiny lists and small sets of results so I thought clarity was more important.

PR #87415, opened 2022-06-06
Filter out nulls in writeUpdatedShardMetaDataAndComputeDeletes

DaveCTurner:
We use `null` to represent failure/corruption cases in the result, but
we don't account for `null` values specially so these paths would result
in NPEs. This commit filters out these `null` markers.

### DIFF
--- a/server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
+++ b/server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
@@ -1020,7 +1020,11 @@ public abstract class BlobStoreRepository extends AbstractLifecycleComponent imp
 
         // Listener that flattens out the delete results for each index
         final ActionListener<Collection<ShardSnapshotMetaDeleteResult>> deleteIndexMetadataListener = new GroupedActionListener<>(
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
```diff
@@ -1020,7 +1020,9 @@ private void writeUpdatedShardMetaDataAndComputeDeletes(
 
         // Listener that flattens out the delete results for each index
         final ActionListener<Collection<ShardSnapshotMetaDeleteResult>> deleteIndexMetadataListener = new GroupedActionListener<>(
-            onAllShardsCompleted.map(res -> res.stream().flatMap(Collection::stream).toList()),
+            onAllShardsCompleted.map(
+                res -> res.stream().filter(Objects::nonNull).flatMap(Collection::stream).filter(Objects::nonNull).toList()
```
original-brownbear, line 1024: I think we can't have any `null` values in `res` can we, the grouped-action listener filters those out? How'd you reproduce/see an NPE here?

DaveCTurner: Huh TIL. Sorry for the noise.

PR #87416, opened 2022-06-06
Docs for synthetic source

nik9000:
This adds some basic docs for synthetic source both to get us started
documenting it and to show how I'd like to get it documented - with a
central section in the docs for `_source` and "satellite" sections in
each of the supported field types that link back to the central section.

[Preview](https://elasticsearch_87416.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/mapping-source-field.html#synthetic-source)


nik9000:
I've pushed docs for the remaining fields.

docs/reference/mapping/fields/source-field.asciidoc
```diff
@@ -6,6 +6,11 @@ at index time. The `_source` field itself is not indexed (and thus is not
 searchable), but it is stored so that it can be returned when executing
 _fetch_ requests, like <<docs-get,get>> or <<search-search,search>>.
 
+ifeval::["{release-state}"=="unreleased"]
+include::synthetic-source.asciidoc[]
```
kilfoyle, line 10:
```suggestion
In some cases, particularly where reducing the volume of your stored data is a
priority, you may consider either using <<synthetic-source,synthetic `_source`>>
or <<disable-source-field,disabling the `_source` field>> altogether.

include::synthetic-source.asciidoc[]
```
People may not scroll all the way to the "Disabling the _source field" section so I think it'd helpful to have a short intro sentence here.

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,161 @@
+[[synthetic-source]]
+==== Synthetic `_source`
+
+Though very handy to have around, the source field takes up a fair amount of
+space on disk. Instead of storing it on disk precisely as you sent it, Elasticsearch
+can reconstruct it on the fly. Enable this by setting `synthetic: true` in `_source`:
```
kilfoyle, line 6:
```suggestion
Though very handy to have around, the source field takes up a significant amount
of space on disk. Instead of storing source documents on disk exactly as you
send them, Elasticsearch can reconstruct source content on the fly. Enable this
by setting `synthetic: true` in `_source`:
```

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,161 @@
...
+* <<numeric-synthetic-source,`float`>>
+* <<geo-point-synthetic-source,`geo_point`>>
+* <<numeric-synthetic-source,`half_float`>>
+* <<numeric-synthetic-source,`integer`>>
+* <<ip-synthetic-source,`ip`>>
+* <<keyword-synthetic-source,`keyword`>>
+* <<numeric-synthetic-source,`long`>>
+* <<numeric-synthetic-source,`scaled_float`>>
+* <<numeric-synthetic-source,`short`>>
+* <<text-synthetic-source,`text`>>
```
kilfoyle, line 38:
```suggestion
While this on the fly reconstruction is *generally* slower than saving the source
documents verbatim and loading them at query time, it saves a lot of storage
space. There are a couple of other restrictions to be aware of:

* When you retrieve sythetic `_source` content it undergoes minor
<<synthetic-source-modifications,modifications>> compared to the original JSON.
* Synthetic `_source` can be used with indices that contain only these field
types:

** <<boolean-synthetic-source,`boolean`>>
** <<numeric-synthetic-source,`byte`>>
** <<numeric-synthetic-source,`double`>>
** <<numeric-synthetic-source,`float`>>
** <<geo-point-synthetic-source,`geo_point`>>
** <<numeric-synthetic-source,`half_float`>>
** <<numeric-synthetic-source,`integer`>>
** <<ip-synthetic-source,`ip`>>
** <<keyword-synthetic-source,`keyword`>>
** <<numeric-synthetic-source,`long`>>
** <<numeric-synthetic-source,`scaled_float`>>
** <<numeric-synthetic-source,`short`>>
** <<text-synthetic-source,`text`>>
```
I suggest breaking out the restrictions as separate bullets and rephrasing slightly.

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,161 @@
...
+
+[[synthetic-source-modifications]]
+===== Synthetic source modifications
```
kilfoyle, line 41:
```suggestion
===== Synthetic source modifications

When synthetic `_source` is enabled, retrieved documents undergo some
modifications compared to the original JSON.
```

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,161 @@
...
+
+[[synthetic-source-modifications-alphabetical]]
+====== Sorts fields alphabetically
+Synthetic source will make sort all fields alphabetically so:
```
kilfoyle, line 45:
```suggestion
====== Alphabetical sorting
Synthetic `_source` fields are sorted alphabetically. For example:
```
I think the headings might read a bit better as nouns.

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,161 @@
...
+PUT idx/_doc/1
+{
+  "foo": 1,
+  "bar": 2,
+  "baz": 3
+}
+----
+// TEST[s/$/\nGET idx\/_doc\/1?filter_path=_source\n/]
+
+Will become:
```
kilfoyle, line 58:
```suggestion
Is retrieved as:
```

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,161 @@
...
+  "bar": 2,
+  "baz": 3,
+  "foo": 1
+}
+----
+// TEST[s/^/{"_source":/ s/\n$/}/]
+
+[[synthetic-source-modifications-leaf-arrays]]
+====== Moves arrays to leaf fields
+Synthetic source will move all arrays to leaves so:
```
kilfoyle, line 72:
```suggestion
====== Arrays moved to leaf fields
Synthetic `_source` arrays are moved to leaves. For example:
```

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,161 @@
...
+  }
+}
+----
+// TEST[s/^/{"_source":/ s/\n$/}/]
+
+[[synthetic-source-modifications-field-names]]
+====== Names fields as they are named in the mapping
+Synthetic source will name fields as they are mapped. <<dynamic,dynamic mapping>>
+defaults to interpreting fields with dots in their names as objects so dynamic
+mapping will make documents as "objecty" as possible so:
```
kilfoyle, line 106:
```suggestion
====== Fields named as they are mapped
Synthetic source names fields as they are named in the mapping. When used
with <<dynamic,dynamic mapping>>, fields with dots (`.`) in their names are, by
default, interpreted as multiple objects. For example:
```

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,161 @@
...
+  "foo": {
+    "bar": {
+      "baz": 1
+    }
+  }
+}
+----
+// TEST[s/^/{"_source":/ s/\n$/}/]
+
+You can keep dots in the names by settings <<subobjects>> to `false` so:
```
kilfoyle, line 131:
```suggestion
To change this behavior and store names without splitting them on dots, set
<<subobjects>> to `false`. For example:```

File: docs/reference/mapping/fields/synthetic-source.asciidoc [hunk 10]
```diff
@@ -0,0 +1,161 @@
...
+
+PUT idx/_doc/1
+{
+  "foo.bar.baz": 1
+}
+----
+// TEST[s/^/DELETE idx\n/]
+// TEST[s/$/\nGET idx\/_doc\/1?filter_path=_source\n/]
+
+Will stay:
```
kilfoyle, line 153:
```suggestion
Will remain as:
```

docs/reference/mapping/types/boolean.asciidoc
```diff
@@ -214,3 +214,39 @@ The following parameters are accepted by `boolean` fields:
...
     Metadata about the field.
+
+ifeval::["{release-state}"=="unreleased"]
+[[boolean-synthetic-source]]
+==== Synthetic source
+`boolean` fields support <<synthetic-source,synthetic `_source`>> in their
+default configuration. But they don't support synthetic `_source` if you add
+<<copy-to,`copy_to`>> or if you disable <<doc-values,`doc_values`>>.
+
+Synthetic source will always sort `boolean` fields so:
```
kilfoyle, line 225:
```suggestion
`boolean` fields support <<synthetic-source,synthetic `_source`>> in their
default configuration. Synthetic `_source` cannot be used together with
<<copy-to,`copy_to`>> or with <<doc-values,`doc_values`>> disabled.

Synthetic source always sorts `boolean` fields. For example:
```

docs/reference/mapping/types/geo-point.asciidoc
```diff
@@ -203,3 +203,47 @@ For performance reasons, it is better to access the lat/lon values directly:
...
+ifeval::["{release-state}"=="unreleased"]
+[[geo-point-synthetic-source]]
+==== Synthetic source
+`geo_point` fields support <<synthetic-source,synthetic `_source`>> in their
+default configuration. But they don't support synthetic `_source` if you add
+<<ignore-malformed,`ignore_malformed`>>, add <<copy-to,`copy_to`>>, or if you
+disable <<doc-values,`doc_values`>>.
+
+Synthetic source will always sort `geo_point` fields (by latitude then longitude)
+and reduce them to their stored precision. So:
```
kilfoyle, line 216:
```suggestion
`geo_point` fields support <<synthetic-source,synthetic `_source`>> in their
default configuration. Synthetic `_source` cannot be used together with 
<<ignore-malformed,`ignore_malformed`>>, <<copy-to,`copy_to`>>, or with
<<doc-values,`doc_values`>> disabled.

Synthetic source always sorts `geo_point` fields (first by latitude and then
longitude) and reduces them to their stored precision. For example:
```

docs/reference/mapping/types/ip.asciidoc
```diff
@@ -156,3 +156,46 @@ GET my-index-000001/_search
...
+
+ifeval::["{release-state}"=="unreleased"]
+[[ip-synthetic-source]]
+==== Synthetic source
+`ip` fields support <<synthetic-source,synthetic `_source`>> in their
+default configuration. But they don't support synthetic `_source` if you add
+<<ignore-malformed,`ignore_malformed`>>, <<copy-to,`copy_to`>>, or if you
+disable <<doc-values,`doc_values`>>.
+
+Synthetic source will always sort `ip` fields and remove duplicates so:
```
kilfoyle, line 168:
```suggestion
`ip` fields support <<synthetic-source,synthetic `_source`>> in their default
configuration. Synthetic `_source` cannot be used together with
<<ignore-malformed,`ignore_malformed`>>, <<copy-to,`copy_to`>>, or with
<<doc-values,`doc_values`>> disabled.

Synthetic source always sorts `ip` fields and removes duplicates. For example:
```

docs/reference/mapping/types/keyword.asciidoc
```diff
@@ -173,6 +173,45 @@ Dimension fields have the following constraints:
...
 
+ifeval::["{release-state}"=="unreleased"]
+[[keyword-synthetic-source]]
+==== Synthetic source
+`keyword` fields support <<synthetic-source,synthetic `_source`>> in their
+default configuration. But they don't support synthetic `_source` if you add
+<<ignore-above,`ignore_above`>>, a <<normalizer,`normalizer`>>,
+<<copy-to,`copy_to`>>, or if you disable <<doc-values,`doc_values`>>.
+
+Synthetic source will always sort `keyword` fields and remove duplicates so:
```
kilfoyle, line 184:
```suggestion
`keyword` fields support <<synthetic-source,synthetic `_source`>> in their
default configuration. Synthetic `_source` cannot be used together with
<<ignore-above,`ignore_above`>>, a <<normalizer,`normalizer`>>,
<<copy-to,`copy_to`>>, or with <<doc-values,`doc_values`>> disabled.

Synthetic source always sorts `keyword` fields and removes duplicates. For
example:
```

docs/reference/mapping/types/numeric.asciidoc
```diff
@@ -233,3 +233,71 @@ numeric field can't be both a time series dimension and a time series metric.
...
+ifeval::["{release-state}"=="unreleased"]
+[[numeric-synthetic-source]]
+==== Synthetic source
+All numeric fields except `unsigned_long` support
+<<synthetic-source,synthetic `_source`>> in their default configuration. But
+they don't support synthetic `_source` if you add
+<<ignore-malformed,`ignore_malformed`>>, add <<copy-to,`copy_to`>>, or if you
+disable <<doc-values,`doc_values`>>.
+
+Synthetic source will always sort numeric fields and remove duplicates so:
```
kilfoyle, line 246:
```suggestion
All numeric fields except `unsigned_long` support <<synthetic-source,synthetic
`_source`>> in their default configuration. Synthetic `_source` cannot be used
together with <<ignore-malformed,`ignore_malformed`>>, <<copy-to,`copy_to`>>, or
with <<doc-values,`doc_values`>> disabled.

Synthetic source always sorts numeric fields and removes duplicates. For example:
```

docs/reference/mapping/types/text.asciidoc
```diff
@@ -159,6 +159,57 @@ The following parameters are accepted by `text` fields:
...
 
+ifeval::["{release-state}"=="unreleased"]
+[[text-synthetic-source]]
+==== Synthetic source
+`text` fields support <<synthetic-source,synthetic `_source`>> if they have
+a `keyword` sub-field that supports synthetic `_source` and *doesn't* have
+<<copy-to,`copy_to`>>.
+
+Synthetic source will always sort `keyword` fields and remove duplicates so
+it will also do that for `text` fields using their sub-`keyword` field. So:
```
kilfoyle, line 170:
```suggestion
`text` fields support <<synthetic-source,synthetic `_source`>> if they have
a `keyword` sub-field that supports synthetic `_source` and *do not* have
<<copy-to,`copy_to`>>.

Synthetic source always sorts `keyword` fields and removes duplicates, so
for `text` fields it sorts based on the sub-`keyword` field. For example:
```

docs/reference/mapping/types/ip.asciidoc
```diff
@@ -156,3 +156,46 @@ GET my-index-000001/_search
...
+[source,console-result]
+----
+{
+  "ip": ["::afff:4567:890a", "10.10.12.123", "192.168.0.1", "2001:db8::1:0:0:1"]
+}
+----
+// TEST[s/^/{"_source":/ s/\n$/}/]
+
+NOTE: IPv4 addresses are sorted as though they were IPv6 addresses prefixed by
+      `::ffff:0:0:0/96` ala https://datatracker.ietf.org/doc/html/rfc6144[rfc6144].
```
kilfoyle, line 199:
```suggestion
NOTE: IPv4 addresses are sorted as though they were IPv6 addresses prefixed by
      `::ffff:0:0:0/96`, following
      https://datatracker.ietf.org/doc/html/rfc6144[rfc6144] guidelines.
```

kilfoyle:
Some very nice docs for a fantastic new feature! Thanks a lot for adding these @nik9000. I have a bunch of suggestions with minor rewordings, but overall I think these look great, especially the "satellite" approach.

docs/reference/mapping/types/ip.asciidoc
nik9000: I'm not sure I'd call this a "guideline". Maybe I should do, but I think of it more like a "regime" or "as specified by rfc6144.

nik9000:
Thanks @kilfoyle! I've accepted all but one of your suggestions and would like to debate a bit about the last one. You certainly write more formally than I do, but your suggestions make it more clear.

docs/reference/mapping/types/ip.asciidoc
kilfoyle: I think `as specified by` is perfect!

kilfoyle:
I'm glad the suggestions help! I definitely can be too formal sometimes. :-)

One thing I forgot to ask... near the bottom of the main `_source field` page there's the warning "Think before disabling the _source field". If those restrictions apply to synthetic source as well, we may want to move that warning up to the top of the page, or something like that.

nik9000:
> One thing I forgot to ask... near the bottom of the main `_source field` page there's the warning "Think before disabling the _source field". If those restrictions apply to synthetic source as well, we may want to move that warning up to the top of the page, or something like that.

For the most part none of those things are true for synthetic source. The debugging one is a little complicated and so is the corruption. But for the most part, no, synthetic source exists to solve those problems.

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+send them, Elasticsearch can reconstruct source content on the fly. Enable this
+by setting `synthetic: true` in `_source`:
+
+[source,console,id=enable-synthetic-source-example]
+----
+PUT idx
+{
+  "mappings": {
+    "_source": {
+      "synthetic": true
```
javanna, line 15: not a question for the docs specifically but users may wonder who takes the precedence between enabled and synthetic here if both are specified?

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
+}
+----
+// TESTSETUP
+
+While this on the fly reconstruction is *generally* slower than saving the source
```
javanna, line 22: it is slower at search time but doesn't it make indexing faster? Or is the indexing speed gain negligible?

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+** <<numeric-synthetic-source,`float`>>
+** <<geo-point-synthetic-source,`geo_point`>>
+** <<numeric-synthetic-source,`half_float`>>
+** <<numeric-synthetic-source,`integer`>>
+** <<ip-synthetic-source,`ip`>>
+** <<keyword-synthetic-source,`keyword`>>
+** <<numeric-synthetic-source,`long`>>
+** <<numeric-synthetic-source,`scaled_float`>>
+** <<numeric-synthetic-source,`short`>>
+** <<text-synthetic-source,`text`>>
```
javanna, line 43: isn't this half the truth given that text has to have a keyword sub-field?

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+
+Will become:
+
+[source,console-result]
+----
+{
+  "bar": 2,
+  "baz": 3,
+  "foo": 1
+}
```
javanna, line 74: reading this, as much as I understand why you mention it, keys ordering is never a guarantee in JSON land. I wonder if it's then needed to provide an example. Could we shorten it and say "You should never rely on keys ordering, but if you do beware you'll lose that with synthetic source as what you'll get back is not 100% what you sent"?

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,170 @@
...
+// TEST[s/$/\nGET idx\/_doc\/1?filter_path=_source\n/]
+
+Will remain as:
+
+[source,console-result]
+----
+{
+  "foo.bar.baz": 1
+}
+----
```
javanna, line 169: I wonder what is specific of synthetic source about how fields are mapped: isn't this all about dynamic mappings? Did you mean to say that synthetic _source recreate the objects structure or not depending on how fields are mapped?

docs/reference/mapping/types/boolean.asciidoc
```diff
@@ -214,3 +214,39 @@ The following parameters are accepted by `boolean` fields:
...
+  "bool": [true, false, true, false]
+}
+----
+// TEST[s/$/\nGET idx\/_doc\/1?filter_path=_source\n/]
+
+Will become:
+[source,console-result]
+----
+{
+  "bool": [false, false, true, true]
```
javanna, line 248: when I see this I am conflicted: first question that comes to mind is "why use an array of booleans?" but then is users to use this , I guess they may rely on ordering, and JSON arrays do preserve the order, so it looks like supporting retrieval from stored fields could be a good idea to preserve ordering for folks that do care about it?

docs/reference/mapping/fields/synthetic-source.asciidoc
nik9000: `enabled: false, synthetic: true` is rejected. Or, at least, it will be. I'm sad to make something like that, but it's kind of required for backwards compatibility.

docs/reference/mapping/fields/synthetic-source.asciidoc
nik9000: It's complicated. We still make `_recovery_source` so we still have to write the source, but we remove that `_recovery_source` pretty darn quick. I don't expect really much indexing performance change, but merging should be marginally faster.

docs/reference/mapping/fields/synthetic-source.asciidoc
nik9000: The link says that, yeah. Do you think I should add it in the list?

docs/reference/mapping/fields/synthetic-source.asciidoc
nik9000: I get that. I'm providing an example for everything so I like the.... balance? I don't know the right word. I could link to the spec and mention that ordering isn't supported by the spec.

docs/reference/mapping/fields/synthetic-source.asciidoc
nik9000: It recreates the structure precisely as the objects are mapped. At first I tried to explicitly create a field mapped as `foo.bar.baz` but the mapping infrastructure unraveled it so I went with this. I'll make an example, one moment.

docs/reference/mapping/fields/synthetic-source.asciidoc
javanna: why is that? folks that have disabled source in the past, could potentially switch to synthetic source if they have all fields with doc_values etc.?

javanna: or especially if we were to synthetize from stored fields too?

docs/reference/mapping/fields/synthetic-source.asciidoc
javanna: Got it, maybe add (with a keyword sub-field) in the list?

docs/reference/mapping/types/boolean.asciidoc
nik9000: Yeah. I think it's probably right to support loading from stored fields to preserve order. I made a prototype in #87480 and it wasn't as bad as I'd worried it would be. I'm not a fan of the stored field loading infrastructure we have, but it's all good.

But stored fields still do all the other things - they just keep order and dupes. That's *better* for some folks. And it's not too complex to support. So, yeah, I think we should.

And, yeah, and array of booleans is silly. It's more believable in combination with the "I pushed all of the values to leaves" behavior. But it's still not *good*.

docs/reference/mapping/fields/synthetic-source.asciidoc
javanna: I don't have a strong opinion but while the examples with arrays ought to be mentioned, taking this one out would make the docs page a little more compact without losing much?

docs/reference/mapping/fields/synthetic-source.asciidoc
nik9000: `synthetic` can't be modified either and old indices default to false. Mostly because I didn't want to have to think about what enabling it would mean. I hadn't really thought about folks with large old indices and source disabled. For what it's worth, there is a good chance it won't work either - it is the `picky` implementation of synthetic source after all.

docs/reference/mapping/fields/synthetic-source.asciidoc
nik9000: :+1:

nik9000: 5962c2716ea

docs/reference/mapping/fields/synthetic-source.asciidoc
javanna: Ok, I guess I was wondering why this needs to be specifically explained, isn't it the expected behaviour? :)

docs/reference/mapping/fields/synthetic-source.asciidoc
nik9000: We need a tie breaker. @romseygeek, break a tie!

docs/reference/mapping/fields/synthetic-source.asciidoc
javanna: I think you are explaining how dynamic mappings work in combination with subobjects fails, and that synthetic source respects that, but you could omit this here, because it would be surprising and would need to be mentioned if it worked otherwise.

nik9000: I suppose. But it was kind of surprising to me that it *did* work with subobjects. Maybe that's just stuck in my head.

javanna: Scratch that, maybe this is not so obvious :)
One thing to mention could be that you always recreate the object structure despite dots in fields names were provided in the first place. Why not always flatten by the way? After all, there are two ways to provide that document and that leads to the same mapping, you have 50% chances to pick the variant that was sent in the first place.

By the way we may start accepting docs with subobjects even when subobjects are disabled and treat them like dotted fields, but I think the current behaviour of "flattening" would still be good in that case.

docs/reference/mapping/fields/synthetic-source.asciidoc
javanna: No pressure :)

docs/reference/mapping/fields/synthetic-source.asciidoc
nik9000: I've removed the section

nik9000: Oh no! I removed the section!

> Why not always flatten by the way?

We asked some kibana friends and they said folks liked things shaped this way instead of flattened. The objects feel right to folks. Also, I figured if we were guessing anyway the "more objecty" approach was probably more likely to match what users sent. I figured flattened fields with dots in them is more rare.

javanna: fine with me. I think you removed only the section with subobjects false and you could make the remaining section about recreating objects shorter by saying that you follow the mappings structure hence prefer nested objects over dots in fields names but the opposite when subobjects are disabled.

nik9000: > I think you removed only the section with subobjects false and you could make the remaining section about recreating objects shorter by saying that you follow the mappings structure hence prefer nested objects over dots in fields names but the opposite when subobjects are disabled.

Two things:
1. Don't use the word `nested` unless you mean it. I get scared.
2. Do you want me to change the words above? Could you suggest a change? I think the "dynamic mappings work like BLAH" bit is useful because it explains where those objects come from. I don't mean to explain all of dynamic mappings, just enough to give folks a hint.

docs/reference/mapping/types/text.asciidoc
```diff
@@ -159,6 +159,63 @@ The following parameters are accepted by `text` fields:
...
+  ]
+}
+----
+// TEST[s/^/{"_source":/ s/\n$/}/]
+
+NOTE: Reordering text fields can have an effect on <<query-dsl-match-query-phrase,phrase>>
+      and <<span-queries,span>> queries. See the discussion about
+      <<position-increment-gap, here>> for more. You can avoid this by making sure the
+      `slop` parameter on the phrase queries is lower than the `position_increment_gap`.
+      This is the default.
```
kilfoyle, line 216:
```suggestion
NOTE: Reordering text fields can have an effect on <<query-dsl-match-query-phrase,phrase>>
      and <<span-queries,span>> queries. See the discussion about
      <<position-increment-gap,`position_increment_gap`>> for more detail. You
      can avoid this by making sure the `slop` parameter on the phrase queries
      is lower than the `position_increment_gap`. This is the default.
```
Just a small fix for that link.

kilfoyle (approved):
LGTM! 🚗 
Thanks a bunch for adding these Nik. They look great.

docs/reference/mapping/fields/synthetic-source.asciidoc
romseygeek: I'm with @javanna, we don't guarantee key ordering anywhere so I don't think an example is necessary here.

nik9000: Democracy wins. I'll remove the example. I'll reduce this to a note.

docs/reference/mapping/fields/source-field.asciidoc
```diff
@@ -6,6 +6,15 @@ at index time. The `_source` field itself is not indexed (and thus is not
 searchable), but it is stored so that it can be returned when executing
 _fetch_ requests, like <<docs-get,get>> or <<search-search,search>>.
 
+ifeval::["{release-state}"=="unreleased"]
+In some cases, particularly where reducing the volume of your stored data is a
+priority, you may consider either using <<synthetic-source,synthetic `_source`>>
+or <<disable-source-field,disabling the `_source` field>> altogether.
```
javanna, line 12: Is it worth quickly describing the difference between these two options, or people need to go to their corresponding pages to find out?

docs/reference/mapping/fields/synthetic-source.asciidoc
javanna: I see, maybe this could be something to consider and think about as a follow-up

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,119 @@
+[[synthetic-source]]
+==== Synthetic `_source`
+
+Though very handy to have around, the source field takes up a significant amount
+of space on disk. Instead of storing source documents on disk exactly as you
+send them, Elasticsearch can reconstruct source content on the fly. Enable this
```
javanna, line 6: Elasticsearch can reconstruct source content on the fly **upon retrieval**?

docs/reference/mapping/fields/synthetic-source.asciidoc
```diff
@@ -0,0 +1,119 @@
...
+  }
+}
+----
+// TEST[s/^/{"_source":/ s/\n$/}/]
+
+[[synthetic-source-modifications-field-names]]
+====== Fields named as they are mapped
+Synthetic source names fields as they are named in the mapping. When used
+with <<dynamic,dynamic mapping>>, fields with dots (`.`) in their names are, by
+default, interpreted as multiple objects. For example:
```
javanna, line 87: , while dots in field names are preserved within objects that have subobjects disabled (with a link to the docs for the subobjects paramater)

javanna (approved):
I left a couple small comments, LGTM otherwise. Thanks!

PR #87427, opened 2022-06-06
Use Map in DiscoveryNodes

rjernst:
DiscoveryNodes uses ImmutableOpenMap to internally keep track of various
sets of nodes with particular attributes, like master nodes and data
nodes. This commit converts these containers to Map.

relates #86239


DaveCTurner (approved):
LGTM

rjernst:
> Apparently except for the change in handling of null keys it seems.

Yeah I'm working through the tests, we have several tests using mocks which end up passing a null node id.

PR #87428, opened 2022-06-06
Avoid expensive search stats updates

### LINKED ISSUE #85586: Expensive search stats updates with many groups
The search section of the index stats API allows search stats to be grouped by custom tags. To tag what group a search belongs to, the search API supports a `stats` parameter.

In our implementation `ShardSearchStats`, the grouped stats are stored in an immutable map. Every time a new group is added, we duplicate the whole map. So if there are a large number of individual groups, the updates can be really expensive and can even affect search latency.

We should look into making these updates less costly, as well as maybe bounding the number of groups that we attempt to track.

jtibshirani:
Search stats can be grouped by a custom key. Currently the grouped stats are
stored in an immutable map, which is copied every time a new group is added.
This can be really expensive when there are a large number of unique groups.

This PR makes two changes:
* Switch to a concurrent map to avoid copying the map when groups are added.
* Bound the total number of groups at 1,000 to prevent the stats from taking
too much memory. Beyond this number, new groups will not be tracked.

Closes #85586.

### DIFF
--- a/docs/reference/search/search.asciidoc
+++ b/docs/reference/search/search.asciidoc
@@ -654,7 +654,8 @@ exclude fields from this subset using the `excludes` property.
 (Optional, array of strings)
 Stats groups to associate with the search. Each group maintains a statistics
 aggregation for its associated searches. You can retrieve these stats using the
-<<indices-stats,indices stats API>>.
+<<indices-stats,indices stats API>>. At maximum, only 1,000 groups will be
+tracked. Beyond this number, no new groups will be added to the stats.
 
 include::{es-repo-dir}/rest-api/common-parms.asciidoc[tag=terminate_after]
 +
--- a/docs/changelog/87428.yaml
+++ b/docs/changelog/87428.yaml
@@ -1,6 +1,6 @@
 pr: 87428
 summary: Avoid expensive search stats updates
-area: "Search, Stats"
+area: "Stats"
 type: bug
 issues:
  - 85586
--- a/docs/changelog/87428.yaml
+++ b/docs/changelog/87428.yaml
@@ -1,0 +1,6 @@
+pr: 87428
+summary: Avoid expensive search stats updates
+area: "Search, Stats"
+type: bug
+issues:
+ - 85586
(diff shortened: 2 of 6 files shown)


jtibshirani:

jtibshirani:
One note: this could be considered a breaking change in behavior. Before, you could track more than 1,000 search stats groups, whereas now we ignore groups past the 1,000 mark. I don't have a big concern about this, and consider it more of a "bug fix" since tracking a very large number of groups can cause high memory usage and slow searches.

docs/reference/search/search.asciidoc
```diff
@@ -654,7 +654,8 @@ exclude fields from this subset using the `excludes` property.
 (Optional, array of strings)
 Stats groups to associate with the search. Each group maintains a statistics
 aggregation for its associated searches. You can retrieve these stats using the
-<<indices-stats,indices stats API>>.
+<<indices-stats,indices stats API>>. At maximum, only 1,000 groups will be
+tracked. Beyond this number, no new groups will be added to the stats.
```
javanna, line 658: does this cause a silent failure? Should we consider failing the update settings call when the group number goes too high?

javanna: Also, curious about the 1000 number. Did you see real cases where a user had more than 1000?

javanna (commented):
left a couple of questions

docs/reference/search/search.asciidoc
jtibshirani: We ignore groups past the 1000 mark. There is no failure, they will just not be tracked. I could tweak the wording here to make that clearer. I also considered an LRU strategy to bound the size. But this is tricky -- the search stats logic increments then decrements certain stats, which is based on the assumption that stats groups are never evicted.

Also, the setting just exists to make testing easier ("This setting exists only for test purposes, and cannot be changed outside of tests").

As for 1000, I just chose a large number that I think users would not reasonably hit. I did see a case where a user had more than 1000 groups, when they were adding a new group per user session ID. This was creating far too many stats groups, and they were able to find another way to solve their use case. I think we should definitely discuss the 1000 number and whether we think it could affect users.

javanna: I am not too sure about the implications but I would consider failing hard rather than silently not tracking when groups are too many. I feel like it could be seen as a breaking change from some users, but I would also expect users to have noticed issues caused by too many groups, hence this can also be seen like a bugfix as too many groups hinder stability?

PR #87429, opened 2022-06-06
Security plugin close realms

### LINKED ISSUE #86286: OIDC back-channel http client threads hang Window service exit
### Elasticsearch Version

7.17.* 8.*

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

Windows

### Problem Description

When Elasticsearch runs as a Windows Service, the stop service action hangs indefinitely, when an OIDC realm is configured (doesn't have to be used).

-------------

Elasticsearch tries to gracefully terminate all the threads when the JVM is terminated.
That is true on all platforms supported by ES (though this is not a JVM requirement in general).

For a Windows Service deployment, given our configuration of procrun (the code that installs the ES JVM as a Windows Service) the JVM is terminated in a different way (not via `System.exit`).
The Windows Service stop procedure must necessarily terminate all the non-daemon threads, at which point the JVM will then terminate naturally. Any bug in omitting to terminate non-daemon threads is a serious problem for Windows Service deployments (for other deployment types it is a minor nuisance).

Unfortunately, the `OpenIdConnectRealm` (more specifically the `OpenIdConnectAuthenticator`) creates threads for its http client used for backchannel comms, which it does not close because the `Realm#close` method is never invoked.

### Steps to Reproduce

Configure any OIDC realm and install Elasticsearch as a Service on Windows.
The stop service action will hang indefinitely.

albertzaharovits:
The `Closeable#close` method on `Realm`s was never called upon node shutdown.
This is a problem when realms (eg OIDC) create non-daemon threads that expect
to be stopped when the `close` method is invoked. Specifically, it is a problem
on Windows where the graceful shutdown is implemented by terminatting all
non-daemon threads (see `Bootstrap#stop` and `Bootstrap#initializeNatives`).

Closes https://github.com/elastic/elasticsearch/issues/86286


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/Realms.java
```diff
@@ -354,6 +358,17 @@ public Map<String, Object> domainUsageStats() {
...
 
+    @Override
+    protected void doStart() {}
+
+    @Override
+    protected void doStop() {}
+
+    @Override
+    protected void doClose() throws IOException {
+        IOUtils.close(allConfiguredRealms.stream().filter(r -> r instanceof Releasable).map(r -> (Releasable) r).toList());
```
tvernum, line 369: Should this be?
```suggestion
        IOUtils.close(allConfiguredRealms.stream().filter(r -> r instanceof Closeable).map(r -> (Closeable) r).toList());
```

Every `Releasable` is also a `Closeable`, and `IOUtils` only needs `Closeable`.

If we only support `Releasable` then we risk someone implementing `Closeable` and being confused about why it doesn't work. So, in that case we probably should make it invalid to implement `Closeable` without implementing `Releaseable`.

Or, we can just allow `Closeable`.

tvernum:
I think this should be tagged `v8.2.3` as well (if we can get it approved & merged quickly)

ywangd (approved):
LGTM

slobodanadamovic (approved):

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/Realms.java
albertzaharovits: Yes, that's correct, it should close all `Closeable`s, thanks for the suggestion.

albertzaharovits:
>I think this should be tagged v8.2.3 as well (if we can get it approved & merged quickly)

albertzaharovits:
@elasticmachine update branch

albertzaharovits:

PR #87430, opened 2022-06-06
Add support for building elasticsearch source on macos before monterey (Closes #87426)

### LINKED ISSUE #87426: Build source in Apple M1 fails if not using MacOs Monterey
### Elasticsearch Version

8.4

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

Darwin kpl3x4j763.kalpatarugroup.com 20.3.0 Darwin Kernel Version 20.3.0: Thu Jan 21 00:06:51 PST 2021; root:xnu-7195.81.3~1/RELEASE_ARM64_T8101 arm64

### Problem Description

# Description
When running `./gradlew` in my Mac M1 with Big Sur I get an error from `sysctl` mentioning that `hw.perflevel0.physicalcpu` is an unknown property. When listing all properties using `sysctl -A` there's no `hw.perflevel*.*` properties at all.

Looking into this forum https://githubhot.com/repo/seanmonstar/num_cpus/issues/104 it seems the property is not displayed on versions before Monterey. I went ahead and changed this https://github.com/elastic/elasticsearch/blob/master/build-conventions/src/main/java/org/elasticsearch/gradle/internal/conventions/info/ParallelDetector.java#L67-L69 to always use the `"hw.physicalcpu"` property and was able to get a clean build.

I think we should either specify somewhere that MacOS Monterey is required for building Elasticsearch in M1 macs or change the code in `ParallelDetector.java` such that it uses `"hw.physicalcpu"` for versions older than Monterey. I personally prefer the second option and I'll be happy to take the PR for fixing it if you all agree.

### Steps to Reproduce

N/A

### Logs (if relevant)

_No response_

jcbages:
Allow building elasticsearch source code from macos versions older than Monterey by falling back to a previous `sysctl` variable when determining the parallelism of the build

Tested it in my Mac running Big Sur successfully

Closes https://github.com/elastic/elasticsearch/issues/87426

### DIFF
--- a/build-conventions/src/main/java/org/elasticsearch/gradle/internal/conventions/info/ParallelDetector.java
+++ b/build-conventions/src/main/java/org/elasticsearch/gradle/internal/conventions/info/ParallelDetector.java
@@ -8,6 +8,7 @@
 
 package org.elasticsearch.gradle.internal.conventions.info;
 
+import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
 import org.gradle.api.Project;
 import org.gradle.api.provider.ProviderFactory;
--- a/build-conventions/src/main/java/org/elasticsearch/gradle/internal/conventions/info/ParallelDetector.java
+++ b/build-conventions/src/main/java/org/elasticsearch/gradle/internal/conventions/info/ParallelDetector.java
@@ -8,6 +8,7 @@
 
 package org.elasticsearch.gradle.internal.conventions.info;
 
+import org.gradle.api.logging.Logging;
 import org.gradle.api.Project;
 import org.gradle.api.provider.ProviderFactory;
 
@@ -26,6 +27,8 @@ import java.util.stream.Collectors;
 public class ParallelDetector {
 
     private static Integer _defaultParallel = null;
+
+    private final static int MACOS_MONTEREY_MAJOR_VERSION = 12;
 
     public static int findDefaultParallel(Project project) {
         // Since it costs IO to compute this, and is done at configuration time we want to cache this if possible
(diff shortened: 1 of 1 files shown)


build-conventions/src/main/java/org/elasticsearch/gradle/internal/conventions/info/ParallelDetector.java
```diff
@@ -90,4 +94,15 @@ private static boolean isMac(ProviderFactory providers) {
         return providers.systemProperty("os.name").getOrElse("").startsWith("Mac");
     }
 
+    private static boolean isMontereyOrNewer(ProviderFactory providers) {
+        String rawVersion = providers.systemProperty("os.version").getOrElse("").trim();
+        if (rawVersion.isEmpty()) {
+            Logging.getLogger(ParallelDetector.class).warn("Failed to validate MacOs version.");
```
mark-vieira, line 100: Can we refactor the logger out into a static field? It's only used once here but it'll help us not duplicating this if we add more logging to the class down the road.

mark-vieira (changes requested):

mark-vieira:
@elasticmachine ok to test

docs/changelog/87430.yaml
```diff
@@ -0,0 +1,6 @@
+pr: 87430
```
mark-vieira, line 1: We can delete this changelog entry. This isn't a user-facing chnage so we typically don't include build changes in our release notes.

mark-vieira (changes requested):

build-conventions/src/main/java/org/elasticsearch/gradle/internal/conventions/info/ParallelDetector.java
jcbages: Sure! I just added a final static variable called LOGGER with that. Let me know if it looks good to you!

docs/changelog/87430.yaml
jcbages: Sounds good! I added it because of the checks below asking for it but I just erased it!

jcbages:
@mark-vieira just addressed your changes and the build was successful!

mark-vieira (approved):

mark-vieira:
Thanks for the pull request @jcbages!

PR #87431, opened 2022-06-06
Script: Add UpdateByQuery and Reindex contexts

stu-elastic:
Add two new scripting contexts, UpdateByQuery and Reindex, in addition
to the Update context.  These three contexts have different sets of
metadata and so their Metadata classes will be different.  This split
prepares for the addition of their own Metadata classes.

Refs: #86472

### DIFF
--- a/test/framework/src/main/java/org/elasticsearch/script/MockScriptEngine.java
+++ b/test/framework/src/main/java/org/elasticsearch/script/MockScriptEngine.java
@@ -177,6 +177,30 @@ public class MockScriptEngine implements ScriptEngine {
                 }
             };
             return context.factoryClazz.cast(factory);
+        } else if (context.instanceClazz.equals(ReindexScript.class)) {
+            ReindexScript.Factory factory = (parameters, ctx) -> new ReindexScript(parameters, ctx) {
+                @Override
+                public void execute() {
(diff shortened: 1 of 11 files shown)


test/framework/src/main/java/org/elasticsearch/script/MockScriptEngine.java
```diff
@@ -177,6 +177,30 @@ public void execute() {
...
+                public void execute() {
+                    final Map<String, Object> vars = new HashMap<>();
+                    vars.put("ctx", ctx);
+                    vars.put("params", parameters);
+                    vars.putAll(parameters);
+                    script.apply(vars);
+                }
+            };
+            return context.factoryClazz.cast(factory);
+        } else if (context.instanceClazz.equals(UpdateByQueryScript.class)) {
```
rjernst, line 192: We really need to figure out a better way to mock scripts, this has become unwieldy. :(

rjernst (approved):
LGTM

test/framework/src/main/java/org/elasticsearch/script/MockScriptEngine.java
stu-elastic: Yeah, it's pretty nasty.

PR #87432, opened 2022-06-06
Synthetic source: paranoid test for hidden copy_to

nik9000:
This adds a paranoid test to make sure that you can't hide `copy_to`
under a multi field. We don't want to support `copy_to` for synthetic
source at this point because we can't figure out which values are being
copied vs the values that come in the original source.


romseygeek (approved):
LGTM

PR #87440, opened 2022-06-07
Move fix common cluster issues to troubleshooting

Leaf-Lin:
I feel that the `fix-common-cluster-issues` page belongs to the troubleshooting guide instead of the how-to guide.
We probably should do something about the old URL to avoid people encountering 404 though. 

@debadair , can you please advise what can we do to redirect?


debadair:
@elasticmachine Retest this please

debadair (approved):
Agree that this can/should be included in the Troubleshooting section. (Ultimately want to get rid of the "How To" bucket entirely.) We don't need to worry about redirects, as changing where it's included won't affect the URL.

PR #87449, opened 2022-06-07
Regenerate release notes for BC3

craigtaverner:
Regenerated release notes for BC3, and two new PRs were added. See generated docs at https://elasticsearch_87449.docs-preview.app.elstc.co/diff

### DIFF
--- a/docs/reference/release-notes/8.3.0.asciidoc
+++ b/docs/reference/release-notes/8.3.0.asciidoc
@@ -205,6 +205,7 @@ Infra/Core::
 * Faster GET _cluster/settings API {es-pull}86405[#86405] (issue: {es-issue}82342[#82342])
 * Faster string writes by saving stream flushes {es-pull}86114[#86114]
 * Fleet: Add `start_time` and `minimum_execution_duration` attributes to actions {es-pull}86167[#86167]
+* Force property expansion for security policy {es-pull}87396[#87396]
 * Refactor array part into a `BytesRefArray` which can be serialized and … {es-pull}85826[#85826]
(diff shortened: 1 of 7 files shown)


lcawl (approved):
Thanks, docs LGTM

docs/reference/release-notes/8.3.0.asciidoc
```diff
@@ -319,6 +320,7 @@ SQL::
 * Update dependency - JLine - to v 3.21.0 {es-pull}83767[#83767] (issue: {es-issue}83575[#83575])
 
 Search::
+* Update to public lucene 9.2.0 release {es-pull}87162[#87162]
```
pugnascotia, line 323: @lcawl Should we remove the entries (and changelog files) for the Lucene snapshot upgrades? I would have thought that they're superseded by the proper 9.2.0 release?

craigtaverner: Perhaps we should simply delete the changelog files for 86277 and 86852?
We could do this before the next build candidate and get the effects from regenerating the release notes, or we could do it as part of this PR here?

debadair: Looking at the RNs for past releases, it looks like we removed the entries for the Snapshot upgrades, and just left the entry for the final upgrade. Removing the changelog files for the snapshot upgrades looks like the right thing to do IMO.

craigtaverner: Great. I've removed the two changelog entries now.

pugnascotia (approved):

PR #87450, opened 2022-06-07
[DOCS] Clarify index_prefix in prefix query docs

javanna:
The current docs mention that Elasticsearch indexes prefixes between 2 and 5 characters in a separate field. 2 and 5 are default values, and the size of the prefixes indexed depend on the configuration settings.


ywelsch (approved):
LGTM

PR #87456, opened 2022-06-07
Dry up string-prefix filtering on sorted maps

original-brownbear:
I have some incoming work that needs this pattern as well so drying up
the usage of it here.


server/src/main/java/org/elasticsearch/common/util/Maps.java
```diff
@@ -279,6 +280,25 @@ public static <K, V> LinkedHashMap<K, V> newLinkedHashMapWithExpectedSize(int ex
...
+     * @return map
+     */
+    public static <V> SortedMap<String, V> subMap(SortedMap<String, V> map, String prefix) {
+        // create the the next prefix right after the given prefix, and use it as exclusive upper bound for the sub-map to filter by prefix
+        // below
+        if (prefix.isEmpty()) {
+            return map;
+        }
+        char[] toPrefixCharArr = prefix.toCharArray();
+        toPrefixCharArr[toPrefixCharArr.length - 1]++;
```
grcevski, line 297: I know this code is as before, but there's a bug here (although very unlikely). If for some wild reason `toPrefixCharArr[toPrefixCharArr.length - 1]` was `\uFFFF`, the + 1 would wrap around and the subMap will return an incorrect result. An if/else with this condition and a prefix with an extra character of `\u0000` should handle this edge case.

I think we should fix this since the method is now part of the generic Maps class.

stu-elastic:
@original-brownbear I'm going to close this for book keeping purposes, please reopen if it's still relevant.

PR #87461, opened 2022-06-07
`403` on unauthorized grant API key action

### LINKED ISSUE #87438: Better error message for GrantApiKey API
Users with only `manage_own_api_key` privilege is not allowed to use the [GrantApiKey API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-grant-api-key.html#security-api-grant-api-key). The overall behaviour is correctly implemented but the error message could use some improvements. Currently it is something like the follows:

```
POST _security/api_key/grant
{
  "password": "password",
  "grant_type": "password",
  "username": "foo",
  "api_key": {
    "name": "k1"
  }
}

===
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "manage own api key privilege only supports API key requests (not org.elasticsearch.xpack.core.security.action.apikey.GrantApiKeyRequest)"
      }
    ],
    "type": "illegal_argument_exception",
    "reason": "manage own api key privilege only supports API key requests (not org.elasticsearch.xpack.core.security.action.apikey.GrantApiKeyRequest)"
  },
  "status": 400
}
```

This is different from other unauthorized error messages which should be along the line of 
> action [..] is unauthorized for user [...] with roles [...], this action is granted by ..

The status code should also be `403` instead of `400`.

n1v0lg:
Users with only the `manage_own_api_key` privilege are unauthorized to
grant API keys. The current error response in this scenario is a `400`,
with a confusing error message. This PR changes error handling to
return a `403` with a canonical "action ... is unauthorized for user
..." error message instead, since the underlying cause is an
authorization error rather than a bad request.

Closes #87438


x-pack/plugin/security/qa/security-trial/src/javaRestTest/java/org/elasticsearch/xpack/security/apikey/ApiKeyRestIT.java
```diff
@@ -176,4 +177,24 @@ public void testGrantApiKeyWithoutApiKeyNameWillFail() throws IOException {
         assertEquals(400, e.getResponse().getStatusLine().getStatusCode());
         assertThat(e.getMessage(), containsString("api key name is required"));
     }
+
+    public void testGrantApiKeyWithoutGrantApiKeyPrivilegeFails() throws IOException {
```
n1v0lg, line 181: Adding the test here instead of as a yaml REST test since there currently are none for granting API keys; happy to add some in a separate PR.

ywangd: A seperate PR is fine. It's also not entirely necessary to have both yaml REST test and Java REST test.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/privilege/ManageOwnApiKeyClusterPrivilege.java
```diff
@@ -92,6 +93,8 @@ protected boolean extendedCheck(String action, TransportRequest request, Authent
                 }
             } else if (request instanceof final QueryApiKeyRequest queryApiKeyRequest) {
                 return queryApiKeyRequest.isFilterForCurrentUser();
+            } else if (request instanceof GrantApiKeyRequest) {
+                return false;
             }
             throw new IllegalArgumentException(
                 "manage own api key privilege only supports API key requests (not " + request.getClass().getName() + ")"
```
ywangd, line 100: I think we should add an assertion before the throw because it is certainly a programming error for requests not related to API keys to reach here. It also helps to catch similar issue eariler in the future.

x-pack/plugin/security/qa/security-trial/src/javaRestTest/java/org/elasticsearch/xpack/security/apikey/ApiKeyRestIT.java
```diff
@@ -176,4 +177,24 @@ public void testGrantApiKeyWithoutApiKeyNameWillFail() throws IOException {
...
+        Request request = new Request("POST", "_security/api_key/grant");
+        request.setOptions(
+            RequestOptions.DEFAULT.toBuilder()
+                .addHeader("Authorization", UsernamePasswordToken.basicAuthHeaderValue(END_USER, END_USER_PASSWORD))
+        );
```
ywangd, line 186: This does not test the actual change: `END_USER` does not have the `manage_own_api_key` privilege which means the authorization denial does not go through the new changes in `ManageOwnApiKeyClusterPrivilege`.

ywangd (changes requested):

x-pack/plugin/security/qa/security-trial/src/javaRestTest/java/org/elasticsearch/xpack/security/apikey/ApiKeyRestIT.java
ywangd: > it might be getting called via findPrivilegesThatGrant, when constructing the exception message.

Good find! So the test actually does excercise the changes but in an unexpected way! I still think it's worth to explicity grant `manage_own_api_key` privilege to the `END_USER`. Then this test can kill both birds with one stone.

x-pack/plugin/security/qa/security-trial/src/javaRestTest/java/org/elasticsearch/xpack/security/apikey/ApiKeyRestIT.java
```diff
@@ -54,7 +55,7 @@ public void createUsers() throws IOException {
         createUser(SYSTEM_USER, SYSTEM_USER_PASSWORD, List.of("system_role"));
         createRole("system_role", Set.of("grant_api_key"));
         createUser(END_USER, END_USER_PASSWORD, List.of("user_role"));
-        createRole("user_role", Set.of("monitor"));
+        createRole("user_role", Set.of("monitor", "manage_own_api_key"));
```
ywangd, line 58: I'd prefer we have a different role and user for the new test since it is better for the existing grantApiKey tests to keep having a user without any API Key related privileges. It's not absolutely necessary since we know today it should not make a difference but it feels more future proof.

ywangd (approved):
LGTM

PR #87465, opened 2022-06-07
Fix troubleshooting doc tab to point to self-managed

andreidan:
This fixes the `Self Managed` tab to  load the 
self-managed instructions.


jbaiera (approved):
LGTM

PR #87478, opened 2022-06-07
Make bootstrap stacktrace printing less cumbersome

rjernst:
The special StartupException is a class that exists to allow truncating
long stacktraces from the terminal output on startup. This is
particularly important for guice errors, which can be enormous.
Currently BootstrapExceptions are wrapped with StartupException to
handle possibly truncating them when printing. However, since main no
longer throws Exception, all handling of stacktrace printing is within
exitWithUnknownException. This commit makes the truncated stacktrace
printing a utility method instead of a full blown RuntimeException
wrapper.


pugnascotia (approved):
LGTM

grcevski (commented):
LGTM!

PR #87480, opened 2022-06-07
Synthetic source: load text from stored fields

nik9000:
Adds support for loading `text` and `keyword` fields that have
`store: true`. We could likely load *any* stored fields, but I
wanted to blaze the trail using something fairly useful.


nik9000:
Force push incoming to resolve merge conflicts.

nik9000:
cloud deploy robot, please build me an image

nik9000:
run elasticsearch-ci/part-2

server/src/main/java/org/elasticsearch/index/mapper/SourceLoader.java
```diff
@@ -145,12 +178,12 @@ interface Leaf {
...
              * Position the loader at a document.
              */
-            boolean advanceToDoc(int docId) throws IOException;
+            boolean advanceToDoc(FieldsVisitor fieldsVisitor, int docId) throws IOException;
 
             /**
              * Write values for this document.
              */
-            void write(XContentBuilder b) throws IOException;
+            void write(FieldsVisitor fieldsVisitor, XContentBuilder b) throws IOException;
```
nik9000, line 186: I don't think I actually need `fieldsVisitor` here - I think `advanceToDoc` can grab it.

nik9000:
@romseygeek I'd love it if you could have a look at this and form an opinion about it sometime.

server/src/main/java/org/elasticsearch/index/get/ShardGetService.java
```diff
@@ -301,11 +305,20 @@ private GetResult innerGetFetch(
         );
     }
 
-    private static FieldsVisitor buildFieldsVisitors(String[] fields, FetchSourceContext fetchSourceContext) {
-        if (fields == null || fields.length == 0) {
+    private static FieldsVisitor buildFieldsVisitors(String[] fields, FetchSourceContext fetchSourceContext, SourceLoader loader) {
+        LogManager.getLogger(ShardGetService.class).error("ASDFDAFAF loading {}", loader.requiredStoredFields().toList());
+        if (fields != null && fields.length > 0) {
```
romseygeek, line 310: This will be much nicer with a future StoredFieldsLoader!

server/src/main/java/org/elasticsearch/index/mapper/SourceLoader.java
romseygeek: Yep. I wonder if we can avoid having it as a parameter in any of these methods and instead pass it StoredFieldSourceLoader implementations directly? Having a method param that is only used by a specific subset of implementations feels off to me.

server/src/main/java/org/elasticsearch/search/fetch/FetchPhase.java
```diff
@@ -241,7 +243,11 @@ public int compareTo(DocIdToIndex o) {
         }
     }
 
-    private static FieldsVisitor createStoredFieldsVisitor(SearchContext context, Map<String, Set<String>> storedToRequestedFields) {
+    private static FieldsVisitor createStoredFieldsVisitor(
+        SearchContext context,
+        Map<String, Set<String>> storedToRequestedFields,
+        SourceLoader sourceLoader
+    ) {
```
romseygeek, line 250: FieldsVisitor really needs to die as an API that we are forced to use anywhere...

romseygeek (commented):
I think this will work.  It will definitely be much nicer when I've done the StoredFieldsLoader work.

I think it will be interesting to see how this interacts with `match_only_text` as well.  It would be nice to be able to read the text to analyze there from a stored field rather than from the source; then we can support synthetic source in a whole bunch of our solutions just by setting `store:true` on their text fields.

server/src/main/java/org/elasticsearch/index/mapper/SourceLoader.java
nik9000: Like in the ctor?

nik9000: I could move it to the `leaf` method pretty easily. But it's kind of tricky because you have to advance the state in a specific way. And holding on to a reference to the thing for a while feels like it is more "at a distance". Like, we take a `docId` as a parameter, but we only use it if we're using doc values.

nik9000:
I have an idea! Somewhere to take this to fix the funny parameter. It
should make Alan happy and make this faster. It'll be easier to read as a
change to this PR. I'll see about building it in the morning.

wandergeek:

nik9000:
@romseygeek i think this is ready for another round when you are ready for it!

server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -1059,11 +1055,6 @@ protected SourceLoader.SyntheticFieldLoader syntheticFieldLoader(String simpleNa
         if (hasScript()) {
             return SourceLoader.SyntheticFieldLoader.NOTHING;
         }
-        if (hasDocValues == false) {
-            throw new IllegalArgumentException(
-                "field [" + name() + "] of type [" + typeName() + "] doesn't support synthetic source because it doesn't have doc values"
-            );
-        }
         if (fieldType().ignoreAbove() != Defaults.IGNORE_ABOVE) {
```
romseygeek, line 1058: Does `ignore_above` not work if `stored=true`?

server/src/main/java/org/elasticsearch/index/mapper/SortedNumericDocValuesSyntheticFieldLoader.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.Map;
+import java.util.stream.Stream;
+
+/**
+ * Load {@code _source} fields from {@link SortedNumericDocValues}.
+ */
+public abstract class SortedNumericDocValuesSyntheticFieldLoader implements SourceLoader.SyntheticFieldLoader {
+    private final String name;
+    private final String simpleName;
+    private CheckedConsumer<XContentBuilder, IOException> writer = b -> {};
```
romseygeek, line 28: This reads a bit weirdly to me, does it make more sense to leave `write` as abstract and just overload it in the two implementations?

server/src/main/java/org/elasticsearch/index/mapper/SortedSetDocValuesSyntheticFieldLoader.java
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * Load {@code _source} fields from {@link SortedSetDocValues}.
+ */
+public abstract class SortedSetDocValuesSyntheticFieldLoader implements SourceLoader.SyntheticFieldLoader {
+    private static final Logger logger = LogManager.getLogger(SortedSetDocValuesSyntheticFieldLoader.class);
+
+    private final String name;
+    private final String simpleName;
+    private CheckedConsumer<XContentBuilder, IOException> writer = b -> {};
```
romseygeek, line 34: Same here

server/src/main/java/org/elasticsearch/index/mapper/SourceLoader.java
```diff
@@ -103,54 +131,89 @@ public Leaf leaf(LeafReader reader, int[] docIdsInLeaf) throws IOException {
...
+     * of fields declared in the mapping but that only have values for
+     * dozens of them. It handles this in a few ways:
+     * <ul>
+     *     <li>{@link #docValuesLoader} must be called once per document
+     *         per field to load the doc values, but detects up front if
+     *         there are no doc values for that field. It's linear with
+     *         the number of fields, whether or not they have values,
+     *         but skips entirely missing fields.</li>
+     *     <li>{@link #storedFieldLoaders} are only called when we see
+     *         a stored field and are otherwise ignored. So it's fine
```
romseygeek, line 151: Can you clarify what you mean by 'when we see a stored field' - I guess you mean when we find that a field is present in the current document's stored fields?

server/src/main/java/org/elasticsearch/index/mapper/TextFieldMapper.java
```diff
@@ -1272,12 +1272,15 @@ public SourceLoader.SyntheticFieldLoader syntheticFieldLoader() {
...
+            return new StringStoredFieldFieldLoader(name(), simpleName());
+        }
         for (Mapper sub : this) {
             if (sub.typeName().equals(KeywordFieldMapper.CONTENT_TYPE)) {
                 KeywordFieldMapper kwd = (KeywordFieldMapper) sub;
-                if (kwd.fieldType().hasDocValues()
-                    && kwd.hasNormalizer() == false
-                    && kwd.fieldType().ignoreAbove() == KeywordFieldMapper.Defaults.IGNORE_ABOVE) {
+                if (kwd.hasNormalizer() == false
+                    && kwd.fieldType().ignoreAbove() == KeywordFieldMapper.Defaults.IGNORE_ABOVE
```
romseygeek, line 1282: Should this work with `ignore_above=true` and `stored=true` on the keyword subfield?

romseygeek (commented):
I like the API! I left a few questions.

server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java
nik9000: It doesn't store the field if it is above `ignore_above`.

server/src/main/java/org/elasticsearch/index/mapper/SortedNumericDocValuesSyntheticFieldLoader.java
nik9000: Those are per-segment writers. I'll see if I can make it less janky.

server/src/main/java/org/elasticsearch/index/mapper/SourceLoader.java
nik9000: yes. I'll push a comment for it.

server/src/main/java/org/elasticsearch/index/mapper/TextFieldMapper.java
nik9000: Same deal. We don't store the field if it is above `ignore_above`.

nik9000:
@romseygeek , I pushed a patch to remove the weird:
```
private CheckedConsumer<XContentBuilder, IOException> writer = b -> {};
```
thing. I *think* it's more like what we want when we want to support `ignore_above` as well. And I think it's more readable. Have a look!

nik9000:
run elasticsearch-ci/bwc

romseygeek (approved):
I think there are more cleanups to do around stored field loading, but this is a great start.  Thanks for all the iterations!

nik9000:
> I think there are more cleanups to do around stored field loading, but this is a great start. Thanks for all the iterations!

Woooh! Thanks for all the iterations too. I think we got something much nicer through them.

nik9000:
I'll work on adding some docs for this after I cover `ignore_above`. The words will merge conflict otherwise.

PR #87481, opened 2022-06-07
Fix serialization checks in force_synthetic_source

nik9000:
This fixes a missing check for unsupported version logic that'd cause
`force_synthetic_source` to be silently dropped when sent to nodes
before 8.4. It also fixes an incorrect version number in an error
message.

Relates to #87068


javanna (approved):

nik9000:
run elasticsearch-ci/part-2

PR #87485, opened 2022-06-07
Add note about license to "Restore an Entire Cluster" docs

TheRiffRafi:
One user reached out mentioning that it would be a good idea to remind users to re-upload the license after full cluster recovery from snapshot as one can easily miss this when trying to figure out why some features aren't working after the restore.


bytebilly:
Thanks @gwbrown for opening this issue.
Would it be better to load the license before restoring data? I'm not sure if this has drawbacks and it's possible in general, but I think that it may be safer in case of paid features use.
What do you think?

gwbrown:
@bytebilly To be clear, I didn't open this issue; this is a community PR that I decided to shepherd. To add some content to this, the license is kept in the cluster state, which is overwritten when a snapshot is restored with `include_global_state: true` (the default is `false`), which also overwrites all index templates, ILM policies, and everything else stored in cluster state.

The issue is that restoring the snapshotted cluster state may restore an old license (or at least so it appears from this contribution and reading the relevant code). I agree that it would be best to address this with a technical solution - it's weird that restoring a snapshot restores the license - but doing so will take time as we'll have to build new functionality to do that, so adding a note to the documentation here might not be a bad idea until we can do so. I do have some thoughts about the wording that I'll write up soon as a proper review.

tvernum:
> it's weird that restoring a snapshot restores the license

I can understand that it might be surprising in some cases, but if you are using snapshot + restore to migrate a cluster (e.g. to move to a different cloud provider) then having everything come along with you (including the license) is exactly the right behaviour. "weird" is going to be in the eye of the beholder.

AthenaEryma (approved):
Apologies for kind of losing track of this one - it would definitely be good to have this note in the docs, so I've cleaned up the wording a little and moved it to the top of the list of features to start up again after restoring and this should be merged shortly. Thanks!!

gwbrown:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.14|[](https://github.com/elastic/elasticsearch/pull/108435)|
|✅|8.13|[](https://github.com/elastic/elasticsearch/pull/108436)|
|✅|7.17|[](https://github.com/elastic/elasticsearch/pull/108437)|


### Questions ?

PR #87499, opened 2022-06-08
More detailed desired balance logging

idegtiarenko:
Please see comments in line


server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceReconciler.java
```diff
@@ -221,12 +222,18 @@ private void allocateUnassigned() {
...
                 }
 
                 if (logger.isTraceEnabled()) {
-                    logger.trace("No eligible node found to assign shard [{}] amongst [{}]", shard, assignment);
+                    logger.trace(
+                        "No eligible node found to assign shard [{}] using assignment [{}] with negative allocation decisions [{}]",
+                        shard,
+                        assignment,
+                        noDecisions
+                    );
```
idegtiarenko, line 236: This log entry is particularly helpful for situations when desired balance is computed, however `canAllocate` returns NO for each of the desired nodes.

idegtiarenko: It is even more helpful when setting `allocation.setDebugMode(RoutingAllocation.DebugMode.EXCLUDE_YES_DECISIONS);`. I wonder if we could do this for `logger.isTraceEnabled()` or if we should have a separate cluster setting for that?

DaveCTurner: I think this should only be a temporary state while a new desired balance is being computed. Are you seeing cases where shards cannot be assigned to their desired nodes even after convergence?

We could call `setDebugMode` if tracing is enabled on this branch if it helps development. Not sure we'd want that in production code tho.

idegtiarenko: Looks like this happens persistently for decisions produced by `ResizeAllocationDecider` such as `NO(source primary is allocated on another node)` or `NO(initial allocation of the shrunken index is only allowed on nodes [_id:"eh7a8csCQzOwCePaFlw9xA"] that hold a copy of every shard in the index)`

DaveCTurner (approved):
LGTM

PR #87504, opened 2022-06-08
Simplify RestResponse Implementations

original-brownbear:
This sets up chunked encoding rest responses. Make it so we only have
a single REST response implementation (+the security filtering override)
throughout the codebase. The test implementations were all redundant,
removing them made the abstract base class redundant so I collapsed
`BytesRestResponse` into it.
A follow-up to this will change the signature of the `content()` method
to an abstraction that allows for chunked encoding and simplify
the very complicated resource handling associated with REST responses in
`org.elasticsearch.http.DefaultRestChannel#sendResponse` based on the
simplification in here.


Tim-Brooks (approved):
LGTM

original-brownbear:
Thanks Tim!

PR #87505, opened 2022-06-08
Clamp auto-expand replicas to the closest value

### LINKED ISSUE #84788: Auto-expand replicas should clamp to closest value
### Elasticsearch Version

All

### Installed Plugins

N/A

### Java Version

_bundled_

### OS Version

N/A

### Problem Description

Today if an index has an `index.auto_expand_replicas` setting which cannot be realised in the cluster then the number of replicas is left unchanged by `AllocationService#adaptAutoExpandReplicas()`. Instead, I think we should change the number of replicas to the closest value within the permitted range.

For instance: if a cluster has a single data node which is marked as shutting down for restart then auto-expand replicas considers this cluster to have zero data nodes, and hence requires -1 replicas. If an index is created with `index.auto_expand_replicas: "0-1"` then it will not be adjusted to have zero replicas, even though this would be better than leaving it at the default value of 1 replica, and will ultimately be the correct setting.

### Steps to Reproduce

1. Create a cluster with one data node.
2. Put a `RESTART` shutdown marker on the data node.
3. Create an index with `index.auto_expand_replicas: "0-1"`
4. Observe that this results in an unassigned replica.

### Logs (if relevant)

_No response_

pxsalehi:
Ensure that the number of replicas chosen for an auto-expand-able index
is within the range of the available data nodes, i.e., excluding those
nodes that cannot be assigned a replica.

Closes #84788

### DIFF
--- a/server/src/internalClusterTest/java/org/elasticsearch/cluster/metadata/AutoExpandReplicasIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/cluster/metadata/AutoExpandReplicasIT.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.cluster.metadata;
+
+import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.test.ESIntegTestCase;
+
+import static org.hamcrest.Matchers.equalTo;
+
+@ESIntegTestCase.ClusterScope(scope = ESIntegTestCase.Scope.TEST, minNumDataNodes = 2, numClientNodes = 0)
+public class AutoExpandReplicasIT extends ESIntegTestCase {
+
+    /**
+     * Verifies that when no data node is available to host a replica, number of replicas are contracted to 0.
+     */
+    public void testAutoExpandToClosesValue() throws Exception {
+        final String indexName = "myindex";
+        final String autoExpandValue = randomFrom("0-1", "0-all");
+        createIndex(
+            indexName,
+            Settings.builder()
+                .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1)
+                .put(IndexMetadata.SETTING_AUTO_EXPAND_REPLICAS, autoExpandValue)
+                .build()
+        );
+
+        final int initialReplicas = autoExpandValue.equals("0-1") ? 1 : internalCluster().numDataNodes() - 1;
+
+        assertBusy(() -> {
+            assertThat(
+                client().admin()
+                    .indices()
+                    .prepareGetSettings(indexName)
+                    .setNames("index.number_of_replicas")
+                    .get()
+                    .getSetting(indexName, "index.number_of_replicas"),
+                equalTo(String.valueOf(initialReplicas))
+            );
+        });
+        ensureGreen(indexName);
+
+        updateIndexSettings(indexName, Settings.builder().put("index.routing.allocation.require._id", "non-existing-node"));
+
+        assertBusy(() -> {
+            assertThat(
+                client().admin()
+                    .indices()
+                    .prepareGetSettings(indexName)
+                    .setNames("index.number_of_replicas")
+                    .get()
+                    .getSetting(indexName, "index.number_of_replicas"),
+                equalTo("0")
+            );
+        });
+
+        ensureGreen(indexName);
+    }
+}
--- a/server/src/internalClusterTest/java/org/elasticsearch/cluster/metadata/AutoExpandReplicasIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/cluster/metadata/AutoExpandReplicasIT.java
@@ -43,7 +43,6 @@ public class AutoExpandReplicasIT extends ESIntegTestCase {
                 equalTo(String.valueOf(initialReplicas))
             );
         });
-        ensureGreen(indexName);
 
         updateIndexSettings(indexName, Settings.builder().put("index.routing.allocation.require._id", "non-existing-node"));
 
@@ -58,7 +57,6 @@ public class AutoExpandReplicasIT extends ESIntegTestCase {
                 equalTo("0")
             );
         });
-        ensureGreen(indexName);
 
         // Remove the setting
         updateIndexSettings(indexName, Settings.builder().put("index.routing.allocation.require._id", ""));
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/elasticsearch/cluster/metadata/AutoExpandReplicas.java
```diff
@@ -110,20 +109,21 @@ public OptionalInt getDesiredNumberOfReplicas(IndexMetadata indexMetadata, Routi
                 numMatchingDataNodes++;
             }
         }
+        return calculateDesiredNumberOfReplicas(numMatchingDataNodes);
+    }
 
+    // Package private only for testing purposes!
+    int calculateDesiredNumberOfReplicas(int numMatchingDataNodes) {
```
pxsalehi, line 116: I was wondering if it is a good idea to test a private method/logic. I think having a small test like `testCalculateDesiredNumberOfReplica` justifies exposing this otherwise private logic, since the test makes it clear what is expected when calculating replicas. 

I went with package-private since the other alternative of using protected doesn't make much sense considering this is a record class.

server/src/main/java/org/elasticsearch/cluster/metadata/AutoExpandReplicas.java
```diff
@@ -110,20 +109,21 @@ public OptionalInt getDesiredNumberOfReplicas(IndexMetadata indexMetadata, Routi
...
         final int min = minReplicas();
         final int max = getMaxReplicas(numMatchingDataNodes);
```
henningandersen, line 118: The `Math.min` done in `getMaxReplicas` seems redundant now, I think we should remove it, it is somewhat confusing now.

server/src/main/java/org/elasticsearch/cluster/metadata/AutoExpandReplicas.java
```diff
@@ -110,20 +109,21 @@ public OptionalInt getDesiredNumberOfReplicas(IndexMetadata indexMetadata, Routi
                 numMatchingDataNodes++;
             }
         }
+        return calculateDesiredNumberOfReplicas(numMatchingDataNodes);
+    }
 
+    // Package private only for testing purposes!
```
henningandersen, line 115:
```suggestion
    // package private for testing
```

server/src/main/java/org/elasticsearch/cluster/metadata/AutoExpandReplicas.java
henningandersen: Yes, that is fine, we do that often.

server/src/test/java/org/elasticsearch/cluster/metadata/AutoExpandReplicasTests.java
```diff
@@ -282,4 +282,16 @@ public void testOnlyAutoExpandAllocationFilteringAfterAllNodesUpgraded() {
             terminate(threadPool);
         }
     }
+
+    public void testCalculateDesiredNumberOfReplicas() {
+        String settingValue = randomFrom("0-all", "0-3");
```
henningandersen, line 287: Can we explore the space a bit more here, for instance by choosing both lower and upper bound randomly like:

```
int lowerBound = between(0,9);
int upperBound = between(lowerBound, 10);
String settingValue = lowerBound + "-" + upperBound;
```

Having a dedicated test (or random variation) with "all" also makes sense.

server/src/internalClusterTest/java/org/elasticsearch/cluster/metadata/AutoExpandReplicasIT.java
```diff
@@ -0,0 +1,64 @@
...
+
+import static org.hamcrest.Matchers.equalTo;
+
+@ESIntegTestCase.ClusterScope(scope = ESIntegTestCase.Scope.TEST, minNumDataNodes = 2, numClientNodes = 0)
+public class AutoExpandReplicasIT extends ESIntegTestCase {
+
+    /**
+     * Verifies that when no data node is available to host a replica, number of replicas are contracted to 0.
+     */
+    public void testAutoExpandToClosesValue() throws Exception {
```
henningandersen, line 22:
```suggestion
    public void testClampToMinMax() throws Exception {
```

henningandersen (commented):
I left a few comments, otherwise this looks good.

pxsalehi:
@elasticmachine update branch

server/src/internalClusterTest/java/org/elasticsearch/cluster/metadata/AutoExpandReplicasIT.java
```diff
@@ -0,0 +1,64 @@
...
+                client().admin()
+                    .indices()
+                    .prepareGetSettings(indexName)
+                    .setNames("index.number_of_replicas")
+                    .get()
+                    .getSetting(indexName, "index.number_of_replicas"),
+                equalTo(String.valueOf(initialReplicas))
+            );
+        });
+        ensureGreen(indexName);
```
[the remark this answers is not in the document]
pxsalehi: IMO, at least for the sake of making it clear what the test expects, we should leave it here. Removing it didn't seem to really speedup the test.

server/src/internalClusterTest/java/org/elasticsearch/cluster/metadata/AutoExpandReplicasIT.java
```diff
@@ -0,0 +1,64 @@
...
+                    .prepareGetSettings(indexName)
+                    .setNames("index.number_of_replicas")
+                    .get()
+                    .getSetting(indexName, "index.number_of_replicas"),
+                equalTo("0")
+            );
+        });
+
+        ensureGreen(indexName);
+    }
```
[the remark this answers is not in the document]
pxsalehi: Done in https://github.com/elastic/elasticsearch/pull/87505/commits/ec7d7f97bc039f46d8c508943783b44ea78edae2.

server/src/main/java/org/elasticsearch/cluster/metadata/AutoExpandReplicas.java
pxsalehi: That's a very good point! It also simplifies the code and the parsing test. Done in https://github.com/elastic/elasticsearch/pull/87505/commits/ec7d7f97bc039f46d8c508943783b44ea78edae2.

server/src/test/java/org/elasticsearch/cluster/metadata/AutoExpandReplicasTests.java
pxsalehi: Done in https://github.com/elastic/elasticsearch/pull/87505/commits/ec7d7f97bc039f46d8c508943783b44ea78edae2.

henningandersen (approved):
LGTM, provided tests pass

pxsalehi:
Thanks Henning! If index status is irrelevant here, then I agree with you. I removed them! :-)

PR #87509, opened 2022-06-08
Consolidate early startup into phase1

rjernst:
This commit introduces a new method in startup of Elasticsearch that
consolidates all the steps up to and including initialization of
logging. Some ordering is changed as some steps were not actually
necessary to run before logging, for example loading secure settings
since they are not actually used by logging.


grcevski (approved):
LGTM! It greatly streamlined things!

PR #87510, opened 2022-06-08
[ML] Fix WordPiece tokenization of unknown words with known subwords

davidkyle:
A word that is not in the dictionary that contains dictionary subwords was incorrectly being tokenized as partially recognised. For example `Elasticsearchffff` was tokenized as `Elastic`, `##search`, `[UNK]` when the entire word should be `[UNK]`

This resulted in a situation where the number of tokens is greater than the size of the token position map. The bug manifested as an assertion failure (if enable) or a NullPointerException due to the violation of the assumption that the token IDs and token map are the same size. The bug was introduced in 8.2, this change restores the behaviour to pre 8.2.

### DIFF
--- a/docs/changelog/87510.yaml
+++ b/docs/changelog/87510.yaml
@@ -1,0 +1,5 @@
+pr: 87510
+summary: Fix `WordPiece` tokenization of unknown words with known subwords
+area: Machine Learning
+type: bug
+issues: []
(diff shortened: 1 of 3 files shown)


droberts195 (approved):
LGTM

benwtrent (approved):
Indeed, we should replace the whole thing with [UNK]

PR #87515, opened 2022-06-08
Use a faster but less accurate log algorithm for computing Geotile Y coordinate

iverase:
This Pr introduces a new algorithm to ESSloppyMath to compute logarithm( base e). This algorithm is then used to compute the Y coordinate of a geotile instead of the default Math.log. This effectively doubles the computation speed as shown in the  following performance test:

```
|                                                        Metric |        Task |       Baseline |     Contender |       Diff |   Unit |   Diff % |
|--------------------------------------------------------------:|------------:|---------------:|--------------:|-----------:|-------:|---------:|
|                                                Min Throughput | geotilegrid |    0.595109    |   1.14372     |    0.54861 |  ops/s |  +92.19% |
|                                               Mean Throughput | geotilegrid |    0.623768    |   1.27022     |    0.64645 |  ops/s | +103.64% |
|                                             Median Throughput | geotilegrid |    0.627972    |   1.28572     |    0.65775 |  ops/s | +104.74% |
|                                                Max Throughput | geotilegrid |    0.636768    |   1.33775     |    0.70098 |  ops/s | +110.08% |
|                                       50th percentile latency | geotilegrid | 1512.38        | 689.169       | -823.207   |     ms |  -54.43% |
|                                       90th percentile latency | geotilegrid | 1525.84        | 696.181       | -829.66    |     ms |  -54.37% |
|                                      100th percentile latency | geotilegrid | 1661.64        | 767.075       | -894.562   |     ms |  -53.84% |
|                                  50th percentile service time | geotilegrid | 1512.38        | 689.169       | -823.207   |     ms |  -54.43% |
|                                  90th percentile service time | geotilegrid | 1525.84        | 696.181       | -829.66    |     ms |  -54.37% |
|                                 100th percentile service time | geotilegrid | 1661.64        | 767.075       | -894.562   |     ms |  -53.84% |

```
I checked and I haven't found any side effect for the loose precision.


libs/core/src/test/java/org/elasticsearch/common/util/ESSloppyMathTests.java
```diff
@@ -42,4 +45,14 @@ public void testSinh() {
             assertEquals(StrictMath.sinh(d), sinh(d), SINH_DELTA);
         }
     }
+
+    public void testLog() {
+        assertTrue(Double.isNaN(log(Double.NaN)));
+        assertEquals(Double.NaN, log(Double.NEGATIVE_INFINITY), LOG_DELTA);
+        assertEquals(Double.POSITIVE_INFINITY, log(Double.POSITIVE_INFINITY), LOG_DELTA);
+        for (int i = 0; i < 10000; i++) {
+            double d = randomDoubleBetween(-Double.MAX_VALUE, Double.MAX_VALUE, true);
```
craigtaverner, line 54: Since the equations above seem to have a bunch of special cases between 0 and 0.95, and 0.95 and 1.14, I think the current test is almost never going to touch those values.

I suggest either a few more explicit tests for values in those ranges, or add an additional test here with random between 0 and 2, perhaps?

libs/core/src/test/java/org/elasticsearch/common/util/ESSloppyMathTests.java
```diff
@@ -19,6 +20,8 @@ public class ESSloppyMathTests extends ESTestCase {
     static double ATAN_DELTA = 1E-15;
     // accuracy for sinh(x)
     static double SINH_DELTA = 1E-15; // for small x
+    // accuracy for log(x)
+    static double LOG_DELTA = 1E-12;
```
craigtaverner, line 24: Since this is a lower precision than elsewhere I wondered if it would have consequences. But I noticed that the y calculation uses log directly, so it seems the error on y should be exactly the error on log, and I presume 1E-12 is good enough for any real-world accuracy for latitude?

craigtaverner (changes requested):
Great performance improvement. I have two comments though:
* Could we have tests for values in the special cases (0.95 and 1.14)
* Do you have a reference for the original maths you got this from?

libs/core/src/test/java/org/elasticsearch/common/util/ESSloppyMathTests.java
craigtaverner: OK. I added some more testing in https://github.com/elastic/elasticsearch/pull/87515/commits/23d7005e6445341d53e776bc7246290cd80fe655

craigtaverner (approved):
I added more testing in https://github.com/elastic/elasticsearch/pull/87515/commits/23d7005e6445341d53e776bc7246290cd80fe655

libs/core/src/test/java/org/elasticsearch/common/util/ESSloppyMathTests.java
iverase: Yes, I think the accuracy is more than enough for us

PR #87520, opened 2022-06-08
Remove cluster block preflight check from health api

### LINKED ISSUE #87464: Relax the prerequisite to evalute the indicators
The master-is-stable indicator will retrieve useful information
regarding the discovery. This will be useful even if the local
cluster state is not recovered.

We currently do not evaluate the instance-has-master indicator
if the cluster state is not recovered, but we should remove this
constraint once master-is-stable is merged

jbaiera:
Removes the logic at the start of the health service that returns UNKNOWN status in case the cluster state not recovered block is in place. Instead, we will defer directly to the master-is-stable indicator and all following preflight indicators.

fixes #87464

### DIFF
--- a/docs/changelog/87520.yaml
+++ b/docs/changelog/87520.yaml
@@ -1,0 +1,6 @@
+pr: 87520
+summary: Remove cluster block preflight check from health api
+area: Health
+type: enhancement
+issues:
+ - 87464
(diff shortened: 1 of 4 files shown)


masseyke (approved):
LGTM (assuming CI is happy)

PR #87525, opened 2022-06-08
[ML] Fix parsing of pytorch thread settings

dimitris-athanasiou:
During #86277 an error was introduced in parsing of pytorch
thread settings. This commit fixes the issue.


dimitris-athanasiou:
Labelling `non-issue` as the bug that is getting fixed has not been released yet.

benwtrent (approved):

davidkyle (approved):
🤭

PR #87535, opened 2022-06-08
Remove ImmutableOpenMap from autoscaling

rjernst:
Autoscaling uses ImmutableOpenMap to keep track of the memory used on
each node. A Map should work just fine here. Note that the wrapper
Collections.unmodifiableMap is used because of the way modifications are
made to the map. Map.copyOf could be used, but it's unclear if the extra
indirection is worse than the extra copying since this should not be
changing or being accessed in any hot loops.

relates #86239


henningandersen (approved):
LGTM.

PR #87536, opened 2022-06-08
Add force_synthetic_source to GET

nik9000:
This adds the option to force synthetic source to the GET API. See
 #87068 for more discussion on why you'd want to do that - the short
version is to get an upper bound on the performance cost of using
synthetic source in GET.

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/get/TransportGetAction.java
+++ b/server/src/main/java/org/elasticsearch/action/get/TransportGetAction.java
@@ -116,7 +116,8 @@ public class TransportGetAction extends TransportSingleShardAction<GetRequest, G
                 request.realtime(),
                 request.version(),
                 request.versionType(),
-                request.fetchSourceContext()
+                request.fetchSourceContext(),
+                request.isForceSyntheticSource()
             );
         return new GetResponse(result);
     }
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/elasticsearch/action/get/TransportGetAction.java
```diff
@@ -116,7 +116,8 @@ protected GetResponse shardOperation(GetRequest request, ShardId shardId) throws
                 request.realtime(),
                 request.version(),
                 request.versionType(),
-                request.fetchSourceContext()
+                request.fetchSourceContext(),
+                request.isForceSyntheticSource()
```
romseygeek, line 120: This was making me think that maybe we should add `forceSynthetic` as a member on FetchSourceContext, but it looks as though that is used in loads of other places so it would end up being a much bigger change.

server/src/main/java/org/elasticsearch/action/get/TransportShardMultiGetAction.java
```diff
@@ -117,7 +117,15 @@ protected MultiGetShardResponse shardOperation(MultiGetShardRequest request, Sha
...
                 GetResult getResult = indexShard.getService()
-                    .get(item.id(), item.storedFields(), request.realtime(), item.version(), item.versionType(), item.fetchSourceContext());
+                    .get(
+                        item.id(),
+                        item.storedFields(),
+                        request.realtime(),
+                        item.version(),
+                        item.versionType(),
+                        item.fetchSourceContext(),
+                        false
```
romseygeek, line 127: This will be enabled in a follow-up, right?

romseygeek (approved):
LGTM!

server/src/main/java/org/elasticsearch/action/get/TransportGetAction.java
nik9000: Yeah. I thought about it and didn't like.

PR #87538, opened 2022-06-08
Improve kNN with filtering docs

jtibshirani:
This change tries to make it easier to find kNN with filtering in the docs:
* Mention filtering support in the kNN API description
* In kNN tutorial, link to the kNN search API page more prominently


mayya-sharipova (approved):
Thanks Julie, additions LGTM!

PR #87539, opened 2022-06-08
Script: Load Whitelists as Resource

stu-elastic:
Changes PainlessPlugin.class.getResourceAsStream to
PainlessPlugin.class.getResource to avoid leaking unclosed
input streams


rjernst (approved):
LGTM

PR #87541, opened 2022-06-08
Remove impossible checked exception from jarhell

rjernst:
The jarhell check declares a URISyntaxException. However, this should
not be possible as the paths and URLs come from the jdk conversion. This
commit makes a URISyntaxException when converting form URL to URI an
assertion error, similar to MalformedURLException when creating a URL.


grcevski (approved):
LGTM!

PR #87542, opened 2022-06-09
Remove impossible exception from security configuration

rjernst:
Bootstraping security declares a checked NoSuchAlgorithmException.
However this seems to be a historical implementation detail that is no
longer present, as none of the methods called from Security.configure
declare this as a thrown exception. This commit removes the exception
from the method's declaration.


stu-elastic (approved):
Allowing `NoSuchAlgorithmException` to bubble up to main makes sense as it's JVM dependent and there's nothing useful to do lower down in the stack anyway.

PR #87545, opened 2022-06-09
EQL: Add circuit breaker to sampling

luigidellaquila:


x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sample/SampleIterator.java
```diff
@@ -174,8 +191,7 @@ private void finalStep(ActionListener<Payload> listener) {
                 if (docGroupsCounter == maxCriteria) {
                     List<SearchHit> match = matchSample(sample, maxCriteria);
                     if (match != null) {
-                        finalSamples.add(match);
-                        samples.add(new Sample(sampleKeys.get(responseIndex / maxCriteria), match));
+                        addSample(new Sample(sampleKeys.get(responseIndex / maxCriteria), match));
```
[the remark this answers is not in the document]
astefan: [This](https://github.com/elastic/elasticsearch/pull/87545/files#diff-bbe495fb163402a367ec596f1f7d8e13f8559d4379b58df170a1d3e340ac7194R136) is the only place where a Page is added to the stack. The page size depends on how many join keys are returned and its max size is configurable (defaults to 1000). I'm wondering if this shouldn't be a better place for calculating the memory usage. We add a new page elements counter and every 1000 (?) new elements in pages, we calculate the memory usage.

astefan (commented):
I'm wondering if another code line is a better place for calculating the memory usage. Added a comment related to that.

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sample/SampleIterator.java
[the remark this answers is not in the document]
luigidellaquila: Yes, it makes sense, I'm adding a memory check there as well. 

As a side effect, the page size becomes a constraint for circuit breaker precision (ie. the bigger the pages, the less precise the circuit breaker becomes), limiting our ability of fine tuning. 
To mitigate this problem, probably it's worth having the memory check both on stack push and on sample add (the samples collection is monotonically growing and potentially unlimited, so it's a hotspot for memory usage)

costin (commented):
I agree that computing memory for every sample is expensive however I'm not sure why there are two sampling sizes. 
I would pick only one strategy and see how well it works, for example pick one of:
a. check the memory at the end of every listener call; essentially after the results have been parsed
b. check the memory every X samples added to the stack

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sample/SampleIterator.java
```diff
@@ -54,12 +58,34 @@ public class SampleIterator implements Executable {
...
+     */
+    protected static final int CB_STACK_SIZE_PRECISION = 1000;
+    private static final String CB_COMPLETED_LABEL = "sample_completed";
+    private static final String CB_INFLIGHT_LABEL = "sample_inflight";
+    private final CircuitBreaker circuitBreaker;
+    private long samplesRamBytesUsed = 0;
+    private long stackRamBytesUsed = 0;
+    private long totalRamBytesUsed = 0;
+    private long totalItemsAddedToStack = 0;
+    private long lastStackSizeWhenCalculatedMemory = 0;
```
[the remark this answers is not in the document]
luigidellaquila: It's not really samples count (it's intermediate data, not samples), but I see your point.
Probably we can go with `totalPageSize` and `lastTotalPageSize`, WDYT?

luigidellaquila:
Thanks @costin, I think your comment (better having only one strategy to about memory calculation) together with what @astefan suggested (use the stack to decide when to execute the memory check) leads us to a simplification: I'm removing the check on the samples for now, and leaving only the one on the stack insertions.
Let's see how it performs on real scenarios; if we see it's not enough, we can always review the strategy.

x-pack/plugin/eql/src/test/java/org/elasticsearch/xpack/eql/execution/sample/CircuitBreakerTests.java
```diff
@@ -0,0 +1,335 @@
...
+        expectThrows(CircuitBreakingException.class, () -> iterator.pushToStack(new SampleIterator.Page(1)));
+    }
+
+    public void testMemoryCleared() {
+        testMemoryCleared(randomBoolean());
+    }
+
+    private void testMemoryCleared(boolean fail) {
+        List<BreakerSettings> eqlBreakerSettings = Collections.singletonList(
+            new BreakerSettings(
```
bpintea, line 109: Wondering if it'd be worth making EqlPlugin#getCircuitBreaker() static and feed it Settings.EMPTY to get this value.

x-pack/plugin/eql/src/test/java/org/elasticsearch/xpack/eql/execution/sample/CircuitBreakerTests.java
```diff
@@ -0,0 +1,335 @@
...
+        try (
+            CircuitBreakerService service = new HierarchyCircuitBreakerService(
+                Settings.EMPTY,
+                eqlBreakerSettings,
+                new ClusterSettings(Settings.EMPTY, ClusterSettings.BUILT_IN_CLUSTER_SETTINGS)
+            );
+            ESMockClient esClient = new ESMockClient(service.getBreaker(CIRCUIT_BREAKER_NAME));
+        ) {
+            CircuitBreaker eqlCircuitBreaker = service.getBreaker(CIRCUIT_BREAKER_NAME);
+            EqlConfiguration eqlConfiguration = new EqlConfiguration(
```
bpintea, line 126: Also here, wondering if it'd be worth adding a EqlTestUtils method to configure a configuration, since most of this is contained in the `TEST_CFG`.

x-pack/plugin/eql/src/test/java/org/elasticsearch/xpack/eql/execution/sample/CircuitBreakerTests.java
```diff
@@ -0,0 +1,335 @@
...
+        }
+    }
+
+    private Sample mockSample() {
+        List<SearchHit> searchHits = new ArrayList<>();
+        searchHits.add(new SearchHit(1, String.valueOf(1), null, null));
+        searchHits.add(new SearchHit(2, String.valueOf(2), null, null));
+        Sample sample = new Sample(new SequenceKey(randomAlphaOfLength(10)), searchHits);
+        return sample;
+    }
```
bpintea, line 211: "Local variable 'sample' is redundant"

x-pack/plugin/eql/src/test/java/org/elasticsearch/xpack/eql/execution/sample/CircuitBreakerTests.java
```diff
@@ -0,0 +1,335 @@
...
+            try {
+                return new InternalComposite(new StreamInput() {
+                    @Override
+                    public byte readByte() throws IOException {
+                        return 0;
+                    }
+
+                    @Override
+                    public void readBytes(byte[] b, int offset, int len) throws IOException {
+
```
bpintea, line 286: Nit: could this empty methods be collapsed on one line?

bpintea (approved):
LGTM.

astefan (approved):
LGTM
Left two minor comments.

x-pack/plugin/eql/src/test/java/org/elasticsearch/xpack/eql/execution/sample/CircuitBreakerTests.java
luigidellaquila: Good point, having a helper method for that in the test suite would save some code.
Unfortunately we cannot make `EqlPlugin#getCircuitBreaker()` static though, since it's part of `CircuitBreakerPlugin` interface contract...

x-pack/plugin/eql/src/test/java/org/elasticsearch/xpack/eql/execution/sample/CircuitBreakerTests.java
luigidellaquila: I think that's what `EqlTestUtils#randomConfiguration()` already does, actually. Double-checking to see if there are significant differences, but I guess we can just use that.

x-pack/plugin/eql/src/test/java/org/elasticsearch/xpack/eql/execution/sample/CircuitBreakerTests.java
luigidellaquila: Unfortunately our style check rules complain about that

x-pack/plugin/eql/src/test/java/org/elasticsearch/xpack/eql/execution/sample/CircuitBreakerTests.java
```diff
@@ -0,0 +1,335 @@
...
+            @Override
+            public void query(QueryRequest r, ActionListener<SearchResponse> l) {}
+
+            @Override
+            public void fetchHits(Iterable<List<HitReference>> refs, ActionListener<List<List<SearchHit>>> listener) {}
+        }, mockCriteria(), randomIntBetween(10, 500), CIRCUIT_BREAKER);
+
+        CIRCUIT_BREAKER.startBreaking();
+        iterator.pushToStack(new SampleIterator.Page(CB_STACK_SIZE_PRECISION / 2));
+        iterator.pushToStack(new SampleIterator.Page(CB_STACK_SIZE_PRECISION - (CB_STACK_SIZE_PRECISION / 2) - 1));
```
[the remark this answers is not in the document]
bpintea: I guess that's due to rounding (should the constant be changed to an odd number). But yes, I also thought it could be simplified (smth like a push with CB_STACK_SIZE_PRECISION, then +1).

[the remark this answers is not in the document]
luigidellaquila: Not really needed indeed, I'm simplifying it

costin (approved):
LGTM

Luegg (approved):

PR #87549, opened 2022-06-09
Fix listener execution

idegtiarenko:
This pr introduces `PendingListenersQueue` that fixes a case when a listener might not have been executed in time.

Fixes: ` IndexRecoveryIT. testCancelRecoveryWithAutoExpandReplicas (stuck after creating index [0-all] index in a cluster with a single master and no data nodes)`

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
@@ -100,6 +100,8 @@ public class DesiredBalanceShardsAllocator implements ShardsAllocator, ClusterSt
                         rerouteServiceSupplier.get().reroute("desired balance changed", Priority.NORMAL, ActionListener.noop());
                     } else {
                         logger.trace("Executing listeners up to [{}] as desired balance did not require reroute", lastConvergedIndex);
+                        // TODO desired balance this still does not guarantee the correct behaviour in case there is
+                        // extra unrelated allocation between one that triggered this computation and one produced by above reroute.
                         queue.complete(lastConvergedIndex);
                     }
                 }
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
```diff
@@ -109,9 +98,9 @@ protected void processInput(DesiredBalanceInput desiredBalanceInput) {
                 if (isFresh) {
                     if (shouldReroute) {
                         rerouteServiceSupplier.get().reroute("desired balance changed", Priority.NORMAL, ActionListener.noop());
-                    } else if (allocating == false) {
+                    } else {
                         logger.trace("Executing listeners up to [{}] as desired balance did not require reroute", lastConvergedIndex);
-                        executeListeners(lastConvergedIndex, desiredBalanceInput.routingAllocation().nodes().isLocalNodeElectedMaster());
+                        queue.complete(lastConvergedIndex);
```
DaveCTurner, line 103: This doesn't seem quite right to me - AFAICT we could still be waiting on a previously-submitted reroute here. In more detail:

- Computation 42 converges with changes (`isFresh == true && shouldReroute == true`) and submits a reroute task.
- Computation 43 converges with no changes (`isFresh == true && shouldReroute == false`) and completes all listeners ≤43, including listener 42 even though its reroute hasn't completed.

idegtiarenko: Queue#complete will complete the listeners immediately only if the `paused = false`. 

Queue is paused in the beginning of allocation cycle (triggered by the reroute) and resumed back after reconciler is complete. resume will execute listeners (if any) that are marked as complete during the time the queue was paused.

I believe this should cover above scenario.

idegtiarenko: Oh, do you mean computation 43 is not caused by a reroute but a separate allocation? Yeah, I do not think this is handled by old nor new implementation yet

idegtiarenko: I think that it might pick up a previous version of desired balance and reconciliate against it. Then this means listener is potentially executed too early.

DaveCTurner (approved):
LGTM

PR #87550, opened 2022-06-09
Remove filter-by-filter specialization for match_all and match_none.

jpountz:
With better rewriting of `bool` queries and the introduction of `Weight#count`,
we no longer need to specialize filter-by-filter for `match_all` and
`match_none` queries and can instead rely on their implementations of
`Weight#count` to return maxDoc / 0.


server/src/main/java/org/elasticsearch/search/aggregations/bucket/filter/QueryToFilterAdapter.java
```diff
@@ -134,23 +127,26 @@ QueryToFilterAdapter<?> union(Query extraQuery) throws IOException {
...
+        Query rewrittenUnion = searcher().rewrite(new ConstantScoreQuery(builder.build()));
+        if (rewrittenUnion instanceof ConstantScoreQuery) {
+          rewrittenUnion = ((ConstantScoreQuery) rewrittenUnion).getQuery();
+        }
+        // This union is inefficient if Lucene cannot merge clauses of the boolean query through a rewrite.
+        final boolean inefficientUnion = rewrittenUnion instanceof BooleanQuery;
+        return new QueryToFilterAdapter(searcher(), key(), rewrittenUnion) {
             public boolean isInefficientUnion() {
-                return true;
+                return inefficientUnion;
```
nik9000, line 149: Ah. I see. You are letting the rewrite logic remove the match_all and match_none. That makes sense.

nik9000 (approved):
Yeah. That makes a lot of sense. Awesome. I'm glad to see these mostly being removed. They never made me particularly happy.

PR #87552, opened 2022-06-09
Speed up ClusterStateHealth constructor

original-brownbear:
This does not need to build the routing nodes, nor do we need to iterate all indices
twice. Not as massive a win because we tend to then build the routing nodes elsewhere 
but not iterating all indices twice saves visibly and we remove a needless use of the
routing nodes.

relates #77466


idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

Issue #85946, opened 2022-04-16
Keyword tokenizer + keep_types filter causing empty string to be returned as only token


veqryn:
### Elasticsearch Version

8.1.2

### Java Version

_bundled_

### OS Version

latest linux docker version

### Problem Description

Using a custom analyzer with the tokenizer set to "keyword", and putting a "keep_types" filter in it, result in no tokens.

This should return a single token of "1234567890", but instead it returns just an empty string: ""

If I replace the tokenizer with 'standard', I get the correct token back: "1234567890"

Changing the types to ALPHANUM and/or changing the text to include letters, words, or spaces, all still result in an empty string returned.

It is important to me to use the keyword tokenizer instead of the standard tokenizer because I want only 1 token (the entire string) being passed to all the filters.  Using the standard tokenizer would result in a string like "1234 AB 56789" being turned into 3 tokens before it gets passed to the filters, which I don't want.

### Steps to Reproduce

`POST _analyze`
```json
{
  "tokenizer": "keyword",
  "filter": [
    {
      "type": "keep_types",
      "types": [
        "<NUM>"
      ]
    }
  ],
  "text": "1234567890"
}
```

Returns an empty token instead of "1234567890"

veqryn:
My attempt to find a workaround reveals that simple_pattern also causes the same issue:
`POST _analyze`
```json
{
  "tokenizer": {
    "type": "simple_pattern",
    "pattern": ".+"
  },
  "filter": [
    "decimal_digit",
    {
      "type": "keep_types",
      "types": [
        "<NUM>"
      ]
    }
  ],
  "text": "1234567890"
}
```

And also the character split and pattern split also fail.
Probably the issue with keep_types doing something wrong and keeping nothing.

romseygeek:
Neither KeywordTokenizer nor PatternTokenizer set the type attribute, so a type filter won't work here indeed.  You could instead use a `predicate_token_filter` with a regex to check that the input token contains only numbers.

veqryn:
Ok, so you are saying that this is working as intended?
Maybe "keep_types" needs better documentation then, as my understanding of it was that the 'type' of any particular byte was inherent to the byte itself and didn't need anything else done first.

As far as a workaround goes, I am currently using this filter instead:
```json
        "only_numbers": {
          "pattern": "[^0-9]",
          "type": "pattern_replace",
          "replacement": ""
        }
```

amitmbm:
[...]

PR #87553, opened 2022-06-09
#85946 Added list of tokenizers don't support keep types token filter

amitmbm:
fix for #85946

### DIFF
--- a/docs/reference/analysis/tokenfilters/keep-types-tokenfilter.asciidoc
+++ b/docs/reference/analysis/tokenfilters/keep-types-tokenfilter.asciidoc
@@ -25,7 +25,7 @@ Certain token filters can also add token types. For example, the
 type.
 
 Some tokenizers don't support this token filter, for example keyword, simple_pattern, and
-simple_pattern_split tokenizers.
+simple_pattern_split tokenizers, as they don't support setting the token type attribute.
 ====
 
 This filter uses Lucene's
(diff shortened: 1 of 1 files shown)


amitmbm:
@javanna @romseygeek please review and let me know if you have any feedbacks.
Thanks in advance.

docs/reference/analysis/tokenfilters/keep-types-tokenfilter.asciidoc
```diff
@@ -20,9 +20,12 @@ produce a variety of token types, including `<ALPHANUM>`, `<HANGUL>`, and
 <<analysis-lowercase-tokenizer,`lowercase`>> tokenizer, only produce the `word`
 token type.
 
-Certain token filters can also add token types. For example, the 
+Certain token filters can also add token types. For example, the
 <<analysis-synonym-tokenfilter,`synonym`>> filter can add the `<SYNONYM>` token
 type.
+
+Some tokenizers don't support this token filter, for example keyword, simple_pattern, and
+simple_pattern_split tokenizers.
```
javanna, line 28: I would add something like "as they don't support setting the token type attribute."

amitmbm: Thanks @javanna so entire sentence would be like 
```
Some tokenizers don't support this token filter, for example keyword, simple_pattern, and
simple_pattern_split tokenizers, as they don't support setting the token type attribute.
```
let me know if this sounds OK to you and I would make the required change.

javanna (approved):
thanks @amitmbm

Issue #86639, opened 2022-05-10
Report stats related to new sizing guidance


DaveCTurner:
In #86223 we adjusted our guidance about node capacity to be more in line with actual resource usage and constraints given recent changes (#77466). The metrics mentioned in the guidance are a little difficult to understand since they depend on opaque mechanisms like mapping deduplication. We should report the relevant stats directly to avoid any misunderstandings.

In particular, in nodes stats we should report the total ~number~ estimated overhead of field mappers on each node, and in cluster stats we should report the total number of fields in the cluster (after deduplication).

Edited (2022-06-09) to change "total number of field mappers" to "total estimated overhead of field mappers", because I expect we'll want to refine our "1kB-per-field" guidance in the future, which we can do if we build this guidance into ES too. Total number of field mappers is also likely still useful.

- [x] Estimated overhead of mappers in node stats https://github.com/elastic/elasticsearch/pull/89807
- [x] Field counts & mapping sizes in cluster stats https://github.com/elastic/elasticsearch/pull/87556
- [x] Update sizing guidance docs to refer to new stats https://github.com/elastic/elasticsearch/pull/90274

kingherc:
I briefly discussed with @original-brownbear about this feature, and gave me the following tips on getting started:
- We can introduce a new `mappers` field under `NodeIndicesStats` with two values, one for the total field count (note that there's no deduplication for data nodes) and one for the estimated overhead (just multiply the field count by 1KiB).
- In `org.elasticsearch.indices.IndicesService#statsByShard` function, for each index, use the index Service to get the mapper Service to finally get the count of the fieldMappers. There's a small doubt whether this count contains nested ones, but I will verify.
- Maybe add a test for NodeStats similar to ClusterStatsIT.

DaveCTurner:
> note that there's no deduplication for data nodes

Although there's no deduplication of mappers _today_ I expect this will be subject to improvement in a future version (see https://github.com/elastic/elasticsearch/issues/86440). It would be good if the solution on which we land doesn't assume "no deduplication" too fundamentally.

> the total number of fields we want to expose should only be part of the node level ... and not part of the indices ... nor shards level

Definitely yes to node-level and no to shard-level, but I think it would make sense to report it at the index level too. That way users can see which indices are contributing most to this memory usage which will help them address it.

kingherc:
[...]

PR #87556, opened 2022-06-09
Report overall mapping size in cluster stats

DaveCTurner:
Adds measures of the total size of all mappings and the total number of
fields in the cluster (both before and after deduplication).

Relates #86639
Relates #77466


docs/reference/cluster/stats.asciidoc
```diff
@@ -1363,6 +1379,10 @@ The API returns the following response:
          "file_sizes": {}
       },
       "mappings": {
+        "total_field_count": 0,
+        "total_deduplicated_field_count": 0,
+        "total_deduplicated_mapping_size": "0b",
+        "total_deduplicated_mapping_size_in_bytes": "0b",
```
sethmlarson, line 1385: Should this be `0` instead of `"0b"`?

server/src/main/java/org/elasticsearch/action/admin/cluster/stats/MappingStats.java
```diff
@@ -170,16 +199,51 @@ private static int countOccurrences(String script, Pattern pattern) {
...
+    public long getTotalDeduplicatedFieldCount() {
+        return totalDeduplicatedFieldCount;
+    }
+
+    /**
+     * @return the total size of all mappings (including those for system indices) accounting for deduplication and compression.
+     */
+    public long getTotalMappingSizeBytes() {
+        return totalMappingSizeBytes;
+    }
```
idegtiarenko, line 245: Should above getters return `Long` to avoid NPE?

idegtiarenko (approved):

docs/reference/cluster/stats.asciidoc
DaveCTurner: ++ thanks, well spotted

server/src/main/java/org/elasticsearch/action/admin/cluster/stats/MappingStats.java
DaveCTurner: ++ or `OptionalLong`, even better

PR #87557, opened 2022-06-09
Simplify + speedup Metadata build some more

original-brownbear:
No need to iterate the indices map another time just to add up
the shard counts.
Also, removes the redundant use of the indices key-set together with
the indices map as well as a redundant loop over the aliases in the
name collision check.

relates #77466


martijnvg (approved):
👍 Nice simplication

PR #87451, opened 2022-06-07
Fail shard if STARTED after master failover

DaveCTurner:
If `cluster.no_master_block: all` then we remove all shards locally
whenever there's no master, but there might still be a shard-started
message in flight. When the new master is elected we start to recover
our shards again and the stale shard-started message could arrive and
move this shard to `STARTED` in the cluster state too soon. This is
pretty rare so we fix it by just failing the shard and starting the
recovery again.

Closes #87367

PR #87558, opened 2022-06-09
[7.17] Fail shard if STARTED after master failover (#87451)

DaveCTurner:
 - Fail shard if STARTED after master failover (#87451)


DaveCTurner:
@elasticmachine update branch

PR #87564, opened 2022-06-09
[WIP] speed up retrieving the last metric of a group of tsids

iverase:
This is just initial work to speed up the following query against a time series index:

```
GET tsdb/_search?request_cache=false
{
  "size": 0,
  "aggs": {
    "by_tsid": {
      "time_series": {},
      "aggs": {
        "metric": {
          "top_metrics": {
             "metrics": {"field": "my-metric"},
             "sort": {"@timestamp": "desc"}
          }
        }
      }
    }
  }
}
```

This query retrieves the latest metric called `my-metric` for each tsid on the index.  Because we are visiting the data for each tsid in order, we only need to visit the first document for each tsid while we could skip the other ones. Still in the current implementation we will be visiting each document on the index.

In order to skip the documents, we need to be able to:

a) Tell the time series engine that we only need to visit a number of documents on the tsid and the other ones can be skipped. 

In order to do that I changed the signature of `boolean isInSortOrderExecutionRequired()` to `int numberOfDocumentsInSortOrderExecution()`. The method returns -1 for non time series aggregation and a positive number for time series aggregation. 

In order to control this number I added a size parameter to the aggregation so currently, we need to manually tell the aggregation how many documents we want to visit. If the size is not provided, it defaults to Integer.MAX_VALUE. This is not ideal but it helped me moving forward with the prototype.

The important thing is that now the `TimeSeriesIndexSearcher` is initialised with the number of documents it needs to visit per tsid.

b) Efficiently move to the first document of a tsid. 

I added a scape path on the doc value iteration when we have visited enough documents that will call the new method `LeafWalker. nextTsid()`.  This method is still not optimal because we don't have enough information to move our iterator efficiently to the next tsid although this implementation is still faster of what we are currently doing.

This is just a prototype and non intended to be merged, any feedback welcome!


jpountz:
In terms of implementation, I wonder if we should reuse the `LeafBucketCollector#competitiveIterator()` API and propagate through the tree of aggregations when appropriate. `TimeSeriesIndexSearcher` could intersect the Query `Scorer` with this iterator to only visit one TSID per segment. My reasoning is that there are other aggregations beyond those that need in-order execution that could benefit from this optimization, e.g. a `cardinality` aggregation running on a field that is a dimension in a TSDB index, or a `cardinality` aggregation running on a field that happens to be the primary index sort in a non-TSDB index.

server/src/main/java/org/elasticsearch/search/aggregations/timeseries/TimeSeriesAggregator.java
```diff
@@ -98,6 +102,14 @@ protected LeafBucketCollector getLeafCollector(LeafReaderContext context, LeafBu
         throws IOException {
         return new LeafBucketCollectorBase(sub, null) {
 
+            @Override
+            public DocIdSetIterator competitiveIterator() throws IOException {
+                if (needCounts) {
+                    return null;
+                }
+                return sub.competitiveIterator();
```
jpountz, line 110: I suspect that there is an assumption here that the competitive iterator of the sub aggregation may only be a competitive iterator that visits every TSID. But what if not? Should we OR the sub iterator with an iterator that is guaranteed to visit every TSID? Maybe we can later work on detecting whether we're OR-ing twice the same iterator and dedup them.

x-pack/plugin/analytics/src/main/java/org/elasticsearch/xpack/analytics/topmetrics/TopMetricsAggregator.java
```diff
@@ -121,14 +128,41 @@ public ScoreMode scoreMode() {
...
+            CompetitiveIterator competitiveIterator;
+
+            @Override
+            public DocIdSetIterator competitiveIterator() throws IOException {
+                if (aggCtx != null && aggCtx.getTsid() != null) {
+                    // TODO: we need to check that the sorting is that in @timestamp in descending order
+                    SortedDocValues tsids = DocValues.getSorted(ctx.reader(), TimeSeriesIdFieldMapper.NAME);
+                    competitiveIterator = new CompetitiveIterator(tsids);
+                }
+                intArray = bigArrays().newIntArray(1);
```
jpountz, line 152: I don't feel great about having side effects in `competitiveIterator` or needing to change the `collect` logic. What about only optimizing the case when `size` is 1, which I would expect to be by far the main case we're interested in optimizing?

jpountz (commented):
This looks nice!

x-pack/plugin/analytics/src/main/java/org/elasticsearch/xpack/analytics/topmetrics/TopMetricsAggregator.java
iverase: ok, we can revisit this when there is a clear use case.

server/src/main/java/org/elasticsearch/search/aggregations/timeseries/TimeSeriesAggregator.java
iverase: I extracted two iterators:

- DocsPerOrdIterator: Iterator that visits each ordinal a provided number of times
- DisjunctionDocIdSetIterator: iterator that is the disjunction of the provided iterators.

The timeseries competitive iterator is the disjunction between the subiterators and an iterator that visit each tsid at least once.

x-pack/plugin/analytics/src/main/java/org/elasticsearch/xpack/analytics/topmetrics/TopMetricsAggregator.java
iverase: I changed the implementation so we are using now the introduced DocsPerOrdIterator which does not introduce side effects in competitiveIterator. Still we need to modify the collect logic as we need to tell the competitive iterator that a doc has been consumed. We don't know which document might intersect with a possible query so we need this information from collect.

PR #87567, opened 2022-06-09
Fix configuration cache compatibility issues in gradle plugins

breskeby:
This fixes references to project that makes the plugin incompatible with Gradle
configuration cache

Related to #57918

### DIFF
--- a/build-tools-internal/src/main/groovy/org/elasticsearch/gradle/internal/doc/DocsTestPlugin.groovy
+++ b/build-tools-internal/src/main/groovy/org/elasticsearch/gradle/internal/doc/DocsTestPlugin.groovy
@@ -15,13 +15,26 @@ import org.elasticsearch.gradle.internal.test.rest.CopyRestTestsTask
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.file.Directory
+import org.gradle.api.file.ProjectLayout
+import org.gradle.api.internal.file.FileOperations
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.TaskProvider
+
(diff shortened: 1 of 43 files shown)


pugnascotia (approved):

breskeby:
@pugnascotia I'll ping you again when done here. Sorry should have been a draft. I want to fix more related issues before merging

breskeby:
@pugnascotia 
I think this is ready now. There's more to fix around task listeners and LoggedExec, but that requires more rework that I don't want to mix with other stuff and keep it separated

pugnascotia (approved):
Is the `xpackProject` utility redundant now, or is it going to become redundant? I'm curious about why we're moving away from it?

breskeby:
@pugnascotia the problem with `xpackProject` was that:

1. using xpackProject in certain situations can break configure configuration cache compatibility as it uses a mutual project object under the hood that is discouraged to use in some use cases (e.g. at execution time) 

2. It always breaks compatibility with `--configure-on-demand`  

3. using `xpackProject` uses the `project` of the `:x-pack` project. referencing other `project` objects from other subproject should avoided where possible to decouple (sub project configurations). There's a good explanation of why we want to decouple our project configurations as much as possible here: https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:decoupled_projects

4. it adds little value over default out of the box gradle api (just use `project(':x-pack:someProject')` instead of `xpackProject('someProject')` Also in some occasions its even shorter. e.g. when this is used as  `xpackProject('someProject').path` instead of just passing `:x-pack:someProject`


I'll try to put a bit more context in the PR description in the future to make the motivation behind these kind of changes more clear upfront

PR #87570, opened 2022-06-09
[TRANSFORM] Add authorization info to transform config listings

droberts195:
Transforms remember permissions from their time of creation
or most recent update. It can be quite hard to determine what
these are when listing transform configs, which can lead to
uncertainty about whether some minor update changed the
permissions.

This change adds information about the permissions to the
config listings.

If the last update was from a user then the permissions used
are the roles they had at the time, and a list of these roles
is added to the new `authorization` field of the transform config
listing.

If the last update was using an API key then the permissions
of that API key are used for subsequent transform operations,
and the new `authorization` field of the transform config listing
shows the API key name and ID.

If a service account did the last update then the service
account name is added to the new `authorization` field of the transform
config listing.


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/xcontent/XContentUtils.java
```diff
@@ -65,4 +72,43 @@ public static String[] readStringArray(XContentParser parser, boolean allowNull)
...
+     */
+    public static void addStoredHeaderInfo(final XContentBuilder builder, final Map<String, String> headers) throws IOException {
+        if (headers == null) {
+            return;
+        }
+        String authKey = headers.get(AuthenticationField.AUTHENTICATION_KEY);
+        if (authKey == null) {
+            return;
+        }
+        builder.startObject(RoleDescriptor.Fields.RUN_AS.getPreferredName());
```
ywangd, line 96: I think it's better to find a different name for this field. `run_as` in RoleDescriptor as a specific meaning for user impersonation, which includes an authenticating user and an effective user. It is also a previlege that user can configure when creating a role. For the use case here, maybe something more generic like `authorization` could be better?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/xcontent/XContentUtils.java
```diff
@@ -65,4 +71,43 @@ public static String[] readStringArray(XContentParser parser, boolean allowNull)
...
+     * - If the permissions are based on a user's roles at the time the config was created then the list of these
+     *   roles is added.
+     * - If the permissions come from an API key then the ID and name of the API key are added.
+     * - If the permissions come from a service account then the name of the service account is added.
+     * @param builder The {@link XContentBuilder} that the extra fields will be added to.
+     * @param headers Security headers that were stored to determine which permissions a background service
+     *                will run as. If <code>null</code> or no authentication key entry is present then no
+     *                fields are added.
+     */
+    public static void addStoredHeaderInfo(final XContentBuilder builder, final Map<String, String> headers) throws IOException {
```
dimitris-athanasiou, line 87: Could we rename this method to something like `addAuthorizationInfo`?

dimitris-athanasiou (commented):
Just a question about renaming the method. Also, please make sure to update the PR + the changelog message to reflect the added field is `authorization` and not `run_as`.

dimitris-athanasiou (approved):
LGTM

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/xcontent/XContentUtils.java
```diff
@@ -65,4 +71,43 @@ public static String[] readStringArray(XContentParser parser, boolean allowNull)
...
+        }
+        String authKey = headers.get(AuthenticationField.AUTHENTICATION_KEY);
+        if (authKey == null) {
+            return;
+        }
+        builder.startObject("authorization");
+        Subject authenticationSubject = AuthenticationContextSerializer.decode(authKey).getEffectiveSubject();
+        switch (authenticationSubject.getType()) {
+            case USER -> builder.array(User.Fields.ROLES.getPreferredName(), authenticationSubject.getUser().roles());
+            case API_KEY -> {
```
przemekwitek, line 99: Don't we need `break;` statements after each `case`?

droberts195: No, `break` statements aren’t used with the modern switch syntax that uses `->` for each case. Only the expression immediately after the `->` is executed for each case (which can be a code block in braces).

przemekwitek (commented):
LGTM

PR #87576, opened 2022-06-09
[8.3] Add error message when searchable snapshot cache wasn't cleared (#87503)

arteam:
 - Add error message when searchable snapshot cache wasn't cleared (#87503)


arteam:

PR #87578, opened 2022-06-09
Tests for synthetic _source from translog

nik9000:
This adds tests to make sure that we use all of the normal synthetic
source machinery, even when loading from the translog. So all GETs on
synthetic source indices will require an in memory index. That'll be an
extra cost on indices that are updated very very frequently.


romseygeek (approved):
LGTM

PR #87579, opened 2022-06-09
Use Map in SnapshotShardSizeInfo

rjernst:
SnapshotShardSizeInfo keeps a mapping of shard to size of the shard.
This commit converts that from an ImmutableOpenMap to Map. Note that
this does not yet change the implementation to HashMap (except in
tests). ImmutableOpenMap is still used in InternalSnapshotInfoService
for now.

relates #86239


original-brownbear (approved):
LGTM

PR #87581, opened 2022-06-10
Consolidate bootstrap shutdown hook

rjernst:
The shutdown hook registered in bootstrap is run when the JVM is
shutting down. An almost identical method, stop(), exists for a special
Windows case that can skip shutdown hooks. This commit consolidates the
two into a new shutdown() method used for both. Additionally, the flag
for determining whether the shutdown hook should be added in bootstrap
setup was always true, so it is removed.


grcevski (approved):
LGTM! I think we should add the test_windows PR label to ensure Windows didn't regress for some reason.

PR #87582, opened 2022-06-10
#87311 update doc for highlight API in case of _source is disabled

### LINKED ISSUE #87311: [Docs] Tune for disk usage section for _source missed critical information on highlight feature
### Elasticsearch Version

8.2

### Installed Plugins

NA

### Java Version

_bundled_

### OS Version

NA

### Problem Description

[disable _source section of the tune for indexing speed](https://www.elastic.co/guide/en/elasticsearch/reference/current/tune-for-disk-usage.html#disable-source ) didn't mention that highlight won't work if field is not stored (default is false), which would be very dangerous for applications building the search functionalities, with `update` and `reindex` we should also mention, that highlight would also not work, if you don't have `store:true` explicitly on the fields highlight is required.

### Steps to Reproduce

NA

### Logs (if relevant)

NA

amitmbm:
fixed #87311

### DIFF
--- a/docs/reference/how-to/disk-usage.asciidoc
+++ b/docs/reference/how-to/disk-usage.asciidoc
@@ -77,7 +77,7 @@ Keep in mind that large shard sizes come with drawbacks, such as long full recov
 [[disable-source]]
 === Disable `_source`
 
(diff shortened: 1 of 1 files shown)


nik9000 (approved):

nik9000:
Looks good. I've merged and poked the backport robots.

PR #87597, opened 2022-06-13
Remove submitUnbatchedStateUpdateTask(...) usage from UpdateTimeSeriesRangeService

martijnvg:
This service periodically updates the cluster state in background (updating `index.time_series.end_time` setting of the latest backing index of all data streams). It was using the unbatched forbidden method. This PR changes the service to use the `clusterService.submitStateUpdateTask(...)` variant that uses task executor. Although it is unlikely that more than one concurrent update is submitted from `UpdateTimeSeriesRangeService`, it is good to move away from forbidden method usages. The executor is simple, in that it just executes `updateTimeSeriesTemporalRange(...)` once if in the event that multiple `UpdateTimeSeriesTask` are submitted.


nik9000 (commented):
It looks like if we had more of these in the queue they'd just merge together and updates to the "current" time.

nik9000 (approved):

martijnvg:
> It looks like if we had more of these in the queue they'd just merge together and updates to the "current" time.

Right, so that `updateTimeSeriesTemporalRange(...)` is invoked only once.

martijnvg:
@elasticmachine update branch

PR #87600, opened 2022-06-13
Speed up iteration over ImmutableOpenMap on hot paths

original-brownbear:
The adjusted spots can see very large maps of O(index_count) getting iterated.
Moving them to the efficient `forEach` that iterates the action in a tight loop
over an array instead of using the indirection of the iterator loop gives
a measurable speedup in many-shards benchmark bootstrapping.
Also, removed some dead code from the immutable open map.

relates #77466


idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

PR #87603, opened 2022-06-13
Refactor various ParameterizedMessage usages into String.format

pgomulka:
ParameterizedMessage is part of log4j api and should not be used
in places where a lambda and String.format would be enough.

This commit is the last batch of refactorings to get rid of ParameterizedMessage
 in ES codebase and consists of various, not related usages.


x-pack/plugin/identity-provider/src/main/java/org/elasticsearch/xpack/idp/saml/authn/SamlAuthnRequestValidator.java
```diff
@@ -156,8 +155,8 @@ private void validateAuthnRequest(
             if (Strings.hasText(parsedQueryString.signature)) {
                 if (Strings.hasText(parsedQueryString.sigAlg) == false) {
                     logAndRespond(
-                        new ParameterizedMessage(
-                            "Query string [{}] contains a Signature but SigAlg parameter is missing",
+                        org.elasticsearch.core.Strings.format(
```
rjernst, line 158: We should look at either consolidating or renaming the two "Strings" classes in server/core, so that fqcn usages are not necessary.

rjernst (approved):
LGTM

x-pack/plugin/identity-provider/src/main/java/org/elasticsearch/xpack/idp/saml/authn/SamlAuthnRequestValidator.java
pgomulka: that would be great. There is more than two.. :)  I created an issue to track this. https://github.com/elastic/elasticsearch/issues/87610

PR #87609, opened 2022-06-13
Handle JWTRealm Rotating JWKS and align docs

### LINKED ISSUE #87608: JWTRealm Auto Rotating JWKS
### Elasticsearch Version

8.2.2

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

Elastic Cloud

### Problem Description

According to the docs the JWT Realm will re-fetch the key if the signature is not valid using the current loaded JWKSet.
But this does not happen, as seen here: https://github.com/elastic/elasticsearch/blob/76be7bcdc060a1031a05aa21b1cfc8627b5e7093/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/jwt/JwtRealm.java#L427

**Docs Reference:**
https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html#:~:text=elasticsearch%20automatically%20caches%20the%20retrieved%20jwk%20set%20to%20avoid%20unnecessary%20http%20requests%2C%20but%20will%20attempt%20to%20refresh%20the%20jwk%20upon%20signature%20verification%20failure%2C%20as%20this%20might%20indicate%20that%20the%20jwt%20provider%20has%20rotated%20the%20signing%20keys.

### Steps to Reproduce

1. Create a Google Cloud Service Account
2. When using this configuration (with Google's daily auto-rotating JWKSet)
```yaml
xpack:
    security:
        authc:
            realms:
                jwt:
                    jwt1:
                        order: 2
                        client_authentication.type: none
                        allowed_issuer: "https://accounts.google.com"
                        allowed_audiences: ["xxx"]
                        allowed_signature_algorithms: ["RS256"]
                        pkc_jwkset_path: "https://www.googleapis.com/oauth2/v3/certs"
                        claims.principal: sub
```
3. Create a role mapping with realm set to `jwt1` and the username to the ID of the service account in Google Cloud
4. After about a day this authentication will not work anymore

### Logs (if relevant)

No logs just 401's

nielsbox:
Fixes: https://github.com/elastic/elasticsearch/issues/87608


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/jwt/JwtRealm.java
```diff
@@ -423,17 +427,38 @@ public void authenticate(final AuthenticationToken authenticationToken, final Ac
...
+                        jwksAndAlgs.algs,
+                        jwksAndAlgs.jwks
+                    );
+                } catch (JwtValidateUtil.InvalidSignatureException e) {
+                    /**
+                     * Elasticsearch automatically caches the retrieved JWK set to avoid unnecessary HTTP requests, 
+                     * but will attempt to refresh the JWK upon signature verification failure, 
+                     * as this might indicate that the JWT Provider has rotated the signing keys.
+                     */
+                    this.refreshJwksAlgs();
```
justincr-elastic, line 447: Based on `isJwtAlgHmac`, only one JWKSet needs to be refreshed, not both PKC and HMAC.
1. If isJwtAlgHmac=false, only the PKC JWKSet needs refresh.
a. Get content from URL or file.
2. If isJwtAlgHmac=true, only the HMAC JWKSet needs refresh.
a. Get content from elasticsearch-keystore. Note, only HMAC JWKSet or HMAC OIDC-compatible JWK can be set at any time.

There are some corner cases where revalidation is not required. Validation uses CPU, and we generally want to minimize the impact on resources that might be in use for search queries or data ingest.

3. If there was an error during reload, keep old settings (ex: network connectivity issue connecting to PKC HTTPS URL, etc).
4. Compare old vs new JWKSet content. If exact match, there is no need to attempt revalidation.
5. If content changed, then set the corresponding this.jwkAlgsPkc or this.jwkAlgsHmac, and run this.verifyAnyAvailableJwkAndAlgPair.
6. If this.verifyAnyAvailableJwkAndAlgPair throws an error, skip validation. We detected the realm config is now in an unusable state. Intervention may be needed to fix the config (ex: PKC JWKSet changes JWK attributes from RS256=>RS512, ES256=>ES512, or HS256=>HS512, but the realm only allows RS256,ES256,HS256 signatures).
7. If this.verifyAnyAvailableJwkAndAlgPair is OK, then revalidation can be attempted.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/jwt/JwtValidateUtil.java
```diff
@@ -310,7 +310,7 @@ public static void validateSignature(final SignedJWT jwt, final List<JWK> jwks)
                 return; // VERIFY SUCCEEDED
             }
         }
-        throw new Exception("Verify failed using " + jwksStrength.size() + " of " + jwks.size() + " provided JWKs.");
+        throw new InvalidSignatureException("Verify failed using " + jwksStrength.size() + " of " + jwks.size() + " provided JWKs.");
```
justincr-elastic, line 313: I think you would need to check if jwksStrength is empty before the for-loop, and throw the original catch-all Exception. Otherwise the code will fall through an empty for-loop and report a misleading InvalidSignatureException.

There are 4 filters in the code before the for-loop.
1. JWT mandatory alg header vs mandatory realm allowed signature algorithms
2. JWT optional kid header vs optional JWK kid attributes
3. JWT mandatory alg header vs optional JWK alg attributes
4. JWT mandatory alg strength vs JWK strength (HMAC and EC only)

One example is the realm config may allow "HS512,ES512", but the submitted JWT was signed by HS256, HS384, ES256, or ES384.

Another example is realm config has a PKC JWKSet with one RSA key, and that one RSA JWK has attributes like kid=123 or alg=RS512, but the submitted JWT has a header like header kid=456 or alg=RS256.

justincr-elastic (changes requested):
Hello. Thank you for your code submission. I provided some feedback on your PR.

In additional to these changes, new unit tests would need to be created to verify the changes. It probably requires refactoring the JwtIssuer test helper class, and other test classes, to trigger a JWKSet reload in the in-memory HTTPS Server or on-disk JWKSet(s)/JWKs, so some tests can attempt validation before and after a reload.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/jwt/JwtRealm.java
nielsbox: Agreed seems my original solution was a bit "quick and dirty". I have implemented your suggestions. One thing I was wondering though `Intervention may be needed to fix the config` do we completely throw exceptions or keep accepting JWTs with old signed keys? I assume nothing will happen with the cached JWT keys even if the new key is raising exceptions?

justincr-elastic: I think we need to accept the new JWKs, and invalidate the cache to stop old JWTs from working.

That node is now in a degraded state. We might be stopping JWT authentication from continuing to work on that node. However, all of the other things will continue working using all other authentication methods, such as searching, indexing, shard replication, cross-cluster follower indexing, etc.

The sooner that JWT authentication problem is noticed, the sooner it can be investigated and fixed.

Consider what happens if a JWT config issue is left for too long. Nodes will start getting bounced as part of normal cloud providers recycling VMs. Those restarted nodes will fail to start. Now all functionality of that node is lost. That is a much more urgent situation. The operator has to do reactive fixing.

If we accept the new JWKs as-is, and invalidate the cache, there is time to report the problem and do proactive fixing of the JWT realm configuration.

nielsbox: I agree with that logic, good to hear though that only the JWT Realm stops working. Was wondering that because I don't know the full inner workings of Elastic Search (yet).

It is indeed the responsibility of the user to fix the configuration or the JWTIssuer Source. So the most secure option for ES is to make this realm in an unusable state.

I have implemented said logic.

nielsbox:
@justincr-elastic I have adressed all PR Review Comments and run the (new) tests including the full `gradle check`(Skipping all tests that my Apple Silicon Mac couldn't handle cause of the AArch64 architecture).

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/jwt/JwtRealm.java
```diff
@@ -165,6 +175,40 @@ boolean isEmpty() {
         }
     }
 
+    private Boolean rotateJwksAlgs(final Boolean isJwtAlgHmac) {
+        try {
+            if (isJwtAlgHmac) {
+                final JwtRealm.JwksAlgs newJwksAlgsHmac = this.parseJwksAlgsHmac();
+                if (this.jwksAlgsHmac != null && this.jwksAlgsHmac.equals(newJwksAlgsHmac) == false) {
```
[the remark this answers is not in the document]
justincr-elastic: Your options for returning two objects from those two methods are probably:
1. Create a record (ex: JwtRealm.StringAndJwksAlgs).
2. Use Tuple class (ex: Tuple<String, JwtRealm.StringAndJwksAlgs>.

Record is preferred. See `ExpiringUser` in JwtRealm class. It is a one-liner, so easy to add and cheap to use.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/jwt/JwtRealm.java
```diff
@@ -165,6 +175,40 @@ boolean isEmpty() {
         }
     }
 
+    private Boolean rotateJwksAlgs(final Boolean isJwtAlgHmac) {
```
justincr-elastic, line 178:
```suggestion
    private boolean rotateJwksAlgs(final boolean isJwtAlgHmac) {
```

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/jwt/JwtRealm.java
```diff
@@ -165,6 +175,40 @@ boolean isEmpty() {
...
+        try {
+            if (isJwtAlgHmac) {
+                final JwtRealm.JwksAlgs newJwksAlgsHmac = this.parseJwksAlgsHmac();
+                if (this.jwksAlgsHmac != null && this.jwksAlgsHmac.equals(newJwksAlgsHmac) == false) {
+                    LOGGER.info("Rotated JWKSet HMAC detected: JWKs [{}]. Algorithms [{}].",
```
justincr-elastic, line 183:
```suggestion
                    LOGGER.warn("Rotated JWKSet HMAC detected: JWKs [{}]. Algorithms [{}].",
```

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/jwt/JwtValidateUtil.java
```diff
@@ -310,7 +310,12 @@ public static void validateSignature(final SignedJWT jwt, final List<JWK> jwks)
                 return; // VERIFY SUCCEEDED
             }
         }
-        throw new Exception("Verify failed using " + jwksStrength.size() + " of " + jwks.size() + " provided JWKs.");
+
+        if (jwksStrength.isEmpty()) {
```
justincr-elastic, line 314: Move before the for-loop.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/jwt/JwtRealmAuthenticateTests.java
```diff
@@ -170,8 +172,152 @@ public void testPkcJwkSetUrlNotFound() throws Exception {
...
+     *
+     * @throws Exception Unexpected test failure
+     */
+    public void testJwkSetRotationHttps() throws Exception {
+        final JwtRealmsService jwtRealmsService = this.generateJwtRealmsService(this.createJwtRealmsSettingsBuilder());
+        final String principalClaimName = randomFrom(jwtRealmsService.getPrincipalClaimNames());
+
+        final List<Realm> allRealms = new ArrayList<>(); // authc and authz realms
+        final boolean createHttpsServer = true; // force issuer to create HTTPS server for its PKC JWKSet
+        final JwtIssuer jwtIssuer = this.createJwtIssuer(0, principalClaimName, 12, 1, 1, 1, createHttpsServer);
```
justincr-elastic, line 186:
```suggestion
        final int algsCount = randomIntBetween(1,12);
        final JwtIssuer jwtIssuer = this.createJwtIssuer(0, principalClaimName, algsCount, 1, 1, 1, createHttpsServer);
```

I think it is a good idea to randomize the number of algorithms, so this test covers different use cases for reloading when using HMAC-only, PKC-only, or both together.

justincr-elastic (changes requested):
I added a lot of comments. I did not get through all of the test code.

Are the new tests for success path when rotation goes OK? I skimmed the tests for now, but I don't think I noticed any negative path tests. For example, test that the realm can do in to bad state.

nielsbox:
> I added a lot of comments. I did not get through all of the test code.
> 
> Are the new tests for success path when rotation goes OK? I skimmed the tests for now, but I don't think I noticed any negative path tests. For example, test that the realm can do in to bad state.

Indeed currently the tests only check for a successful rotation: First check if normal auth works, then rotate, then check if auth works using new jwkset, then retest using the first jwt token and expect that fails.

I will add extra test cases when: fetching the new pair fails (example a gateway error). Which means we should not revalidate and throw a failed validation. The second test case should be when we are able to fetch the new jwkset but the algorithm is wrong. We then should expect everything to fail in the jwtrealm.

justincr-elastic:
I thought of another use case to consider while going through your latest changes. If the realm gets into a bad state, does it need to stay in a bad state indefinitely until manual intervention?

Consider what would happen if a rotated JWKSet was misconfigured, and then fixed. The JWT realm would go into a bad state, but then it should be able to recover.

If the external JWKSet configuration is fixed, it is highly desirable that manual intervention should be required. The JWT reaLM should recover on its own are start working again.

nielsbox:
> I thought of another use case to consider while going through your latest changes. If the realm gets into a bad state, does it need to stay in a bad state indefinitely until manual intervention?
> 
> Consider what would happen if a rotated JWKSet was misconfigured, and then fixed. The JWT realm would go into a bad state, but then it should be able to recover.
> 
> If the external JWKSet configuration is fixed, it is highly desirable that manual intervention should be required. The JWT reaLM should recover on its own are start working again.

Agreed, I will test that aswel! Tomorrow I should have some time to address the comments.

nielsbox:
@justincr-elastic I addressed all new comments. As a heads up, I did have to change the tactic of when to rotate and when not to, to allow for an autorecoverable Realm when a JWKSet's Algorithms but also signing keys change. Why also the Algorithms you ask? I also addressed that in the comments, but I will fully explain my reasoning here. 

Theres a chance that the path to the JWKSet adds or removes a new algorithm or partly keeps previous algs and adds new algorithms. If we did not want to recover from that we would just filter that JWKSets algorithms using the allowedJKWSetAlgs in the settings, but once that is done there is no going back. For example to illustrate:
If we have these initially:`"RS256", "RS384", "PS256"`, if forexample the JWKSet issuer only wanted to support RS512 because that is more secure, even though the path has changed by the JWKIssuer using the previous logic we would not be able to recover or rotate from this. This is because we'd only rotate when the signature isn't valid, but if the algorithms didn't match we'd immediately send an unsuccessfull response. Now everything that relates to the JWKSet, if a JWT is invalid, will trigger a reload of the JWKSet source.

As seen in the tests it will make sure that if and when a rotation failure happens the realm will be quickly able to recover.
Although the PR is now significantly bigger, I welcome and appreciate your efforts in getting this to `master`!

justincr-elastic:
There are a lot of corner cases for sure. I will need to think through a lot of them when reviewing this PR.

I would like to keep the scope of this PR as narrow as possible (i.e. triggered reload if falling through that JWK verify loop). That will keep the PR size and time down.

nielsbox:
> There are a lot of corner cases for sure. I will need to think through a lot of them when reviewing this PR.
> 
> I would like to keep the scope of this PR as narrow as possible (i.e. triggered reload if falling through that JWK verify loop). That will keep the PR size and time down.

Indeed, I did the same when writing this enhancement, seems quite complex but the tests really helped focus on the core problems. 

I also wish to keep the scope smaller, for example removing the auto recoverability would remove some of those changes. Let me know what we shouldn't handle in this PR and I will move those changes to a separate PR.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/jwt/JwtIssuer.java
```diff
@@ -94,6 +94,37 @@ record AlgJwkPair(String alg, JWK jwk) {}
         }
     }
 
+    public void rotate(
```
justincr-elastic, line 97: This method duplicates most of the code in the constructor. After a certain line:
- Missing update to `this.algorithms`
- HTTPS server contents is updated

It should be possible for a single method to handle a call from constructor or rotate. That would reduce duplication, and avoid a copy-and-paste issues like missing update to `this.algorithms`.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/jwt/JwtRealmTestCase.java
```diff
@@ -620,6 +708,7 @@ protected void doMultipleAuthcAuthzAndVerifySuccess(
                 final Exception authcFailed = new Exception("Authentication test failed.");
                 realmFailureExceptions.forEach(authcFailed::addSuppressed); // realm exceptions
                 authcFailed.addSuppressed(t); // final throwable (ex: assertThat)
+                LOGGER.error(t.getMessage());
```
justincr-elastic, line 711:
```suggestion
```

justincr-elastic: `t` is added to authcFailed as a suppressed exception, and authcFails is logged at error level. Why log it again?

justincr-elastic (changes requested):

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/jwt/JwtRealm.java
```diff
@@ -62,6 +62,8 @@ public class JwtRealm extends Realm implements CachingRealm, Releasable {
 
     record ExpiringUser(User user, Date exp) {}
 
+    record StringAndJwksAlgs(String jwkStringContent, JwtRealm.JwksAlgs algs) {}
```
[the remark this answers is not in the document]
justincr-elastic: This should probably be a global change to make the variable names consistent. I am thinking `ContentAndJwksAlgs` or `JwksContentAndJwksAlgs` would be more descriptive.

justincr-elastic:
This PR is getting quite big. It is taking longer to do each review pass, and I haven't had a chance to dive into the tests or think through the corner cases. The exception for zero algs during validation is the kind thing where it would be easy to introduce problems.

At some point, this PR may get too big and it might make sense for me to merge your changes into my fork where I could continue working on the PR. Changes would still be visible, so you would be able to see progress and make comments. However, you seem motivated to iterate on this PR, so we can keep going and reevalutate as we go.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/jwt/JwtRealmTestCase.java
nielsbox: Was a debug leftover, deleted it :P

nielsbox:
> This PR is getting quite big. It is taking longer to do each review pass, and I haven't had a chance to dive into the tests or think through the corner cases. The exception for zero algs during validation is the kind thing where it would be easy to introduce problems.
> 
> At some point, this PR may get too big and it might make sense for me to merge your changes into my fork where I could continue working on the PR. Changes would still be visible, so you would be able to see progress and make comments. However, you seem motivated to iterate on this PR, so we can keep going and reevalutate as we go.

I agree, but I don't see this PR changing a lot anymore, all functionality is here, the tests for the functionality that we want seem to run fine. Maybe we could review the tests first so we can touch on the cases that we wish to support. Since some test cases depend on the comments in your review (Like rotating a wrong JWKSet (after successfull retrieval). If we decide not to support that test case we can remove the corresponding code.

If you see fit that it would be much easier to work on it yourself, that is fine too but I do like the idea of contributing and getting this to master as the author :P. Either way this doesn't matter as long as this functionality gets to a new Elastic Search version. Since we do need this at the company I work at.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/jwt/JwtValidateUtil.java
```diff
@@ -124,7 +124,10 @@ public static void validateAudiences(final SignedJWT jwt, List<String> allowedAu
     public static void validateSignatureAlgorithm(final SignedJWT jwt, final List<String> allowedAlgorithms) throws Exception {
         final JWSAlgorithm algorithm = jwt.getHeader().getAlgorithm();
         if ((algorithm == null) || (allowedAlgorithms.contains(algorithm.getName()) == false)) {
-            throw new Exception("Rejected algorithm [" + algorithm + "]. Allowed [" + String.join(",", allowedAlgorithms) + "]");
+            // JWT algorithm does not match one of the algorithms in the JWKSet. Could be explained by a rotated JKWSet,
+            // we should check if we have the latest JWKSet.
+            // That is, only, when the given algorithm is also in the allowed Algorithms in the settings of the JWTRealm.
+            throw new JWKSetValidationError("Rejected algorithm [" + algorithm + "]. Allowed [" + String.join(",", allowedAlgorithms) + "]");
```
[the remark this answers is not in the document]
nielsbox: Meaning this part of the code will raise `Exception` like before.

justincr-elastic:
There is a contributor tag in the PR merge when people collaborate on a PR, so you would get recognition.

nielsbox:
> There is a contributor tag in the PR merge when people collaborate on a PR, so you would get recognition.

Yeah that is not really a problem for me but thats good to hear, if you wish to take over you have my permission. I will try to review and comment whenever possible.

justincr-elastic:
@elasticmachine update branch

justincr-elastic:
My concern is digging into the tests. If I run them locally and I find issues, I would be reporting them back here and waiting for fixes. Doing that async is slow. There could be many iterations of that. It is more efficient to find, fix, and re-run locally to work through that phase of a PR. A workaround for that would be if both of us could push changes into a shared branch.

justincr-elastic:
I did a local pull. I found a few issues.

I think it will be easier and quicker if I pick it up from here. I will link from the new PR to this one when it is ready. You will be listed as a contributor, and the new PR will link to your PR.

nielsbox:
> I did a local pull. I found a few issues.
> 
> I think it will be easier and quicker if I pick it up from here. I will link from the new PR to this one when it is ready. You will be listed as a contributor, and the new PR will link to your PR.

Alright no problem, looking forward to the changes you will make and what issues you found.

justincr-elastic:
@elasticmachine update branch

nielsbox:
> PR #88023 has been submitted for review if you want to check it out. Thank you.

I have been following your progress in the background. Looks great! Will take a closer look soon.

nielsbox:
As this PR which was superseded by https://github.com/elastic/elasticsearch/pull/88023, is now no longer needed I will close this :)

PR #87617, opened 2022-06-13
Remove total and active shard count from RoutingNodes

original-brownbear:
These are not used any longer.


idegtiarenko (approved):

original-brownbear:
Thanks Ievgen

PR #87627, opened 2022-06-14
More compact serialization of ShardStats

original-brownbear:
The two paths are almost always the same, we can serialize
them only once to reduce memory consumption for shard stats
request handling on the coordinating node.

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/admin/indices/stats/ShardStats.java
+++ b/server/src/main/java/org/elasticsearch/action/admin/indices/stats/ShardStats.java
@@ -8,6 +8,7 @@
 
 package org.elasticsearch.action.admin.indices.stats;
 
+import org.elasticsearch.Version;
 import org.elasticsearch.cluster.routing.ShardRouting;
 import org.elasticsearch.common.io.stream.StreamInput;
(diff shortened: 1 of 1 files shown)


joegallo:
The code looks fine enough, but my initial reaction to this is that it seems like the likely improvement (small) isn't enough to warrant the additional complexity (also small). I'd say I'm convincible, though -- can you give a bit more flavor on the improvement side of things to shift the balance in favor of putting this through?

original-brownbear:
@joegallo :) 

> can you give a bit more flavor on the improvement side of things to shift the balance in favor of putting this through

Let me try. I basically just found this as a result of looking at an in-trouble cluster heap dump and saw endless duplication of these strings. In the end the math is simple: 
For the common case we can assume ~100b for the data path. So 10k shards will cost you 1M for these strings (1M shards -> 100M saved ;) ). That doesn't sound like a whole lot, but for extremely large clusters in particular it's a nice win IMO.
We could also go another step here and just deduplicate these strings globally (since it's always the same string on a given node pretty much) which would become twice as cheap if we didn't have to double read the duplicates also.

joegallo (approved):

original-brownbear:
Thanks Joe!

PR #87629, opened 2022-06-14
Compare shards allocators

idegtiarenko:
This is not intended to be merged yet (or at least until we find a permanent place for it).

This test allows to simulate re-balancing in a big cluster with multiple nodes and many shards.
It logs unnecessary (same shard is moved more then once) relocations if any in the end of execution.


idegtiarenko:
Outdated

PR #87631, opened 2022-06-14
Add ParameterizedMessage to forbidden api

pgomulka:
ParameterizedMessage will not be part of the new ES logging API and therefore should not be used.
java.util.Supplier<String> and String.format should be used instead.

this commit adds ParameterizedMessage to forbidden api 


grcevski (approved):
LGTM!

Issue #86765, opened 2022-05-12
Should (internal) client actions fork before completing their listeners by default?


DaveCTurner:
Actions executed via the transport client would fork to the `LISTENER` threadpool to complete their listener to avoid problems with running too much work on a transport thread. In contrast, actions executed internally via `NodeClient` don't fork by default. We removed this forking behaviour entirely in https://github.com/elastic/elasticsearch/issues/53049 as part of the removal of the transport client.

Since then we've seen at least a couple of issues where internal actions end up inadvertently doing nontrivial work on a transport thread:

- https://github.com/elastic/elasticsearch/issues/82255 (see linked support case from there)
- https://github.com/elastic/elasticsearch/issues/86764

I think this is trappy, there's not even anywhere to specify which thread to use to complete the listener so it's understandable if a developer might think it's going to be completed somewhere safe, perhaps on the same threadpool as was used to call the action in the first place.

I think we should at least require an `executor` parameter when executing an action via the client so that callers must make a conscious decision about whether to fork for the response listener or not. Alternatively we could just make them all fork by default (to where? reinstate the `LISTENER` threadpool?) and perhaps allow a non-forking API for performance-critical cases.

henningandersen:
The es-distributed team discussed this and arrived at requiring an `executor` parameter, filled in with `SAME` in the initial commit (possibly picking another one for known cases like the two cases above).

DaveCTurner:
This is basically the other half of #97916.

PR #87637, opened 2022-06-14
Remove TaskListener interface

DaveCTurner:
`TaskListener` is almost always a trivial wrapper around an
`ActionListener`. There's a couple of places where that's not true, it
reports the task ID in a log message, but these logging effects can be
achieved in other ways which don't require polluting various widely-used
APIs so much.

This commit replaces all usages of `TaskListener` with `ActionListener`,
reworking the logging effects, and removing the now-unnecessary overrides.

Relates #86765 in that this commit simplifies some of the APIs that
#86765 will need to adjust.


pxsalehi (approved):
I've recently learned about this part of the code base, and checking this PR was a good way of digging a bit deeper into that. I have asked one question, but as far as I can tell, LGTM!

PR #87644, opened 2022-06-14
Check earlier for nested type with disabled subobjects while parsing doc

javanna:
With #87218 we have expanded testing and made sure that we check that the nested type is not used within objects that have subobjects disabled.
This commit moves the check needed while applying dynamic mappings to an earlier point, where it is possible to distinguish between nested and object, so we can expose a different error in the two scenarios.

Also, this commit rewords the error message thrown so that it is more user-friendly.


server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
```diff
@@ -483,6 +483,26 @@ private static void parseObjectDynamic(DocumentParserContext context, ObjectMapp
                 dynamicObjectMapper = DynamicFieldsBuilder.createDynamicObjectMapper(context, currentFieldName);
                 context.addDynamicMapper(dynamicObjectMapper);
             }
+            if (mapper.subobjects() == false) {
+                if (dynamicObjectMapper instanceof NestedObjectMapper) {
+                    throw new IllegalArgumentException(
```
romseygeek, line 488: If these now get wrapped in a MapperParsingException, maybe it makes more sense to just throw MapperParsingException directly here?

server/src/test/java/org/elasticsearch/index/mapper/DynamicTemplatesTests.java
```diff
@@ -1175,6 +1175,76 @@ public void testSubobjectsFalseArrayOfObjects() throws IOException {
...
+
+        MapperParsingException err = expectThrows(MapperParsingException.class, () -> mapper.parse(source("""
+            {
+              "metrics.object" : [
+                {}
+              ]
+            }
+            """)));
+        assertEquals(
+            "Object [metrics] has subobjects set to false hence it does not support nested object [object]",
```
romseygeek, line 1209: I'd reword this to `Tried to add nested object [object] to mapper [metrics] which does not support subobjects`

romseygeek (approved):
LGTM. I left a couple of suggestions but no need for another review

server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
ywelsch: +1

ywelsch (approved):
I'm wondering if this PR counts as a true enhancement (which goes into release notes) or whether it should just be a "non-issue" (how I would label these).

server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
javanna: oh well, that does make sense....

server/src/test/java/org/elasticsearch/index/mapper/DynamicTemplatesTests.java
javanna: ++ will do the same for the error that's thrown when parsing object mapper.

Issue #86664, opened 2022-05-11
[CI] ReadinessClusterIT testReadinessDuringRestartsNormalOrder failing


DaveCTurner:
**Build scan:**
https://gradle-enterprise.elastic.co/s/25sfuxxbfq63s/tests/:server:internalClusterTest/org.elasticsearch.readiness.ReadinessClusterIT/testReadinessDuringRestartsNormalOrder

**Reproduction line:**
`./gradlew ':server:internalClusterTest' --tests "org.elasticsearch.readiness.ReadinessClusterIT.testReadinessDuringRestartsNormalOrder" -Dtests.seed=48E77C708ACFAACB -Dtests.locale=lt -Dtests.timezone=Pacific/Auckland -Druntime.java=17`

**Applicable branches:**
master

**Reproduces locally?:**
No

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.readiness.ReadinessClusterIT&tests.test=testReadinessDuringRestartsNormalOrder

**Failure excerpt:**
```
junit.framework.AssertionFailedError: Expected exception IOException but no exception was thrown

  at org.apache.lucene.tests.util.LuceneTestCase.expectThrows(LuceneTestCase.java:2895)
  at org.apache.lucene.tests.util.LuceneTestCase.expectThrows(LuceneTestCase.java:2881)
  at org.elasticsearch.test.readiness.ReadinessClientProbe.lambda$tcpReadinessProbeFalse$2(ReadinessClientProbe.java:69)
  at java.security.AccessController.doPrivileged(AccessController.java:318)
  at org.elasticsearch.test.readiness.ReadinessClientProbe.tcpReadinessProbeFalse(ReadinessClientProbe.java:68)
  at org.elasticsearch.test.readiness.ReadinessClientProbe.tcpReadinessProbeFalse(ReadinessClientProbe.java:60)
  at org.elasticsearch.readiness.ReadinessClusterIT$2.onNodeStopped(ReadinessClusterIT.java:136)
  at org.elasticsearch.test.InternalTestCluster$NodeAndClient.closeForRestart(InternalTestCluster.java:1001)
  at org.elasticsearch.test.InternalTestCluster.restartNode(InternalTestCluster.java:1873)
  at org.elasticsearch.test.InternalTestCluster.restartNode(InternalTestCluster.java:1840)
[...]

PR #87652, opened 2022-06-14
Add more debugging for readiness test timeout

grcevski:
This failure here https://github.com/elastic/elasticsearch/issues/86664 happened again. It could essentially mean two things, the initial 10 second timeout I chose is not enough or that I have another concurrency bug somewhere. 

With this PR I'm extending the timeout to 30 seconds and I've also put an assert to see what the await status is when we fail. In case the timeout hasn't expired and we still see this error, that would mean it's an unknown concurrency issue of some kind.

Relates #86664


rjernst:
> It could essentially mean two things, the initial 10 second timeout I chose is not enough or that I have another concurrency bug somewhere.

The latter seems more likely. 10 seconds should be enough for basically any test? Can we add anymore debugging to help diagnose the next time we see this failure (for example, some additional test logging so we can see how long and when certain parts of this test are running)?

grcevski:
@rjernst, I added extra instrumentation and reverted back the timeout to 10 seconds. Given it failed again last night, I think with this extra info I'll be able to at least pinpoint where the issue is.

rjernst (approved):
LGTM

Issue #86889, opened 2022-05-18
AssertionError (lost uuid) in mixed cluster test


DaveCTurner:
This test run failed due to a tripped assertion in one of the 7.0.1 nodes (#86887) but then one of the 7.17.4 nodes also fell over for apparently-unrelated reasons:

```
[2022-05-18T12:15:22,564][ERROR][o.e.b.ElasticsearchUncaughtExceptionHandler] [v7.0.1-0] fatal error in thread [elasticsearch[v7.0.1-0][snapshot][T#3]], exiting
java.lang.AssertionError: lost uuid JCGi3zjyQgytAyClMFs3Gg
        at org.elasticsearch.repositories.RepositoryData.snapshotsToXContent(RepositoryData.java:693) ~[elasticsearch-7.17.4-SNAPSHOT.jar:7.17.4-SNAPSHOT]
        at org.elasticsearch.repositories.RepositoryData.snapshotsToXContent(RepositoryData.java:648) ~[elasticsearch-7.17.4-SNAPSHOT.jar:7.17.4-SNAPSHOT]
        at org.elasticsearch.repositories.blobstore.BlobStoreRepository.lambda$writeIndexGen$65(BlobStoreRepository.java:2323) ~[elasticsearch-7.17.4-SNAPSHOT.jar:7.17.4-SNAPSHOT]
        at org.elasticsearch.common.blobstore.fs.FsBlobContainer.writeToPath(FsBlobContainer.java:274) ~[elasticsearch-7.17.4-SNAPSHOT.jar:7.17.4-SNAPSHOT]
        at org.elasticsearch.common.blobstore.fs.FsBlobContainer.writeBlob(FsBlobContainer.java:253) ~[elasticsearch-7.17.4-SNAPSHOT.jar:7.17.4-SNAPSHOT]
        at org.elasticsearch.repositories.blobstore.BlobStoreRepository.writeAtomic(BlobStoreRepository.java:2636) ~[elasticsearch-7.17.4-SNAPSHOT.jar:7.17.4-SNAPSHOT]
        at org.elasticsearch.repositories.blobstore.BlobStoreRepository.lambda$writeIndexGen$66(BlobStoreRepository.java:2321) ~[elasticsearch-7.17.4-SNAPSHOT.jar:7.17.4-SNAPSHOT]
[...]

PR #87654, opened 2022-06-14
Throw exception on illegal RepositoryData updates

original-brownbear:
In some edge cases we could attempt a broken write here today,
lets throw in addition to the assertion to not corrupt the repository
until a solution for the underlying problem is found.

relates #86889


DaveCTurner (approved):
LGTM

PR #87658, opened 2022-06-14
Remove ImmutableOpenmap from disk threshold tests

rjernst:
These tests pass ImmutableOpenMaps to ClusterInfo, but that now takes
Map. This commit uses HashMap in the tests instead.

relates #86239


dakrone (approved):
LGTM

PR #87660, opened 2022-06-14
Improve ownership handling of bytesref and bytesrefarray

hendrikmuhs:
When de-serializing a `BytesRefArray` and creating a `BytesRefHash` from it, the ownership wasn't properly defined. This PR improves ownership handling and passing ownership. Also fixes an unnecessary buffer cleaning on de-serialization.

@grcevski I need you one more time, hoping this is the final PR. ;-)


server/src/main/java/org/elasticsearch/common/util/BytesRefArray.java
```diff
@@ -106,6 +121,18 @@ public void close() {
         Releasables.close(bytes, startOffsets);
     }
 
+    public static BytesRefArray takeOwnershipOf(BytesRefArray other) {
+        BytesRefArray b = new BytesRefArray(other.startOffsets, other.bytes, other.size, other.bigArrays);
+
+        // don't leave a broken array behind, although it isn't used any longer
+        // on append both arrays get re-allocated
+        other.startOffsets = LONG_ARRAY_DUMMY;
```
grcevski, line 129: I'm concerned that this will cause the dummy array to expand at some point by accident and we'll effectively share the static between two refs. I was looking at the grow code and there's a path where we do resizeInPlace, I think the pointer doesn't change.

I think we need to ensure this works through tests, essentially we need to check if the inner bytes and startOffset isn't same as the dummy after being taken ownership and grown. We should also ensure that two byterefarrays that are taken ownership and grown don't share the same underlying pointers anymore.

Essentially, it would be really nice if we somehow made the dummy arrays immutable and threw an error if they were attempted to be modified.

hendrikmuhs: ok, it looks like `BigArrays` ignores `size` and over-allocates. It would be better if it would respect the size of `1`. Its not possible to allocate a 1 byte array directly. 

I am thinking about changing it back to the original version and add the suggested `assert`'s: https://github.com/elastic/elasticsearch/pull/83055/files#diff-3e6c436f29ffc4ef48787dbbc1f7cf46035c7d7d43c6d67fd872407f3c6cbd15R119

After passing ownership the array should not be used any longer. In the end I am trying to fix a case which is considered bad anyway.

> Essentially, it would be really nice if we somehow made the dummy arrays immutable and threw an error if they were attempted to be modified.

The above and that require a lot of in-depth changes.

Would you be fine with the `assert` solution? In detail this means setting the member to `null` and adding `assert`'s, see the suggestion in the link above.

grcevski: The assert sounds better. Messing up with shared arrays can lead to some very hard to track problems, I'd prefer the NullPointerException at runtime too, it crashes but at least we can track the problem.

server/src/main/java/org/elasticsearch/common/util/BytesRefArray.java
```diff
@@ -56,14 +56,14 @@ public BytesRefArray(StreamInput in, BigArrays bigArrays) throws IOException {
             // startOffsets
             size = in.readVLong();
             long sizeOfStartOffsets = size + 1;
-            startOffsets = bigArrays.newLongArray(sizeOfStartOffsets, true);
+            startOffsets = bigArrays.newLongArray(sizeOfStartOffsets, false);
```
grcevski, line 59: I have more of a question here, would not clearing of the arrays on resize cause any change of behaviour? I don't know enough about the existing uses of this API, perhaps it's all good, I just want to confirm that we've looked at all scenarios that use this code.

server/src/main/java/org/elasticsearch/common/util/BytesRefHash.java
```diff
@@ -29,7 +29,7 @@ public final class BytesRefHash extends AbstractHash implements Accountable {
         // spare BytesRef
         + RamUsageEstimator.shallowSizeOfInstance(BytesRef.class);
 
-    private final BytesRefArray bytesRefs;
+    private BytesRefArray bytesRefs;
```
grcevski, line 32: I can't quite see why we removed the final from this field definition here? Also related question, can we make hashes final too?

grcevski (commented):
This looks good, I just left two questions. Thanks

hendrikmuhs:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/common/util/BytesRefArray.java
hendrikmuhs: clearing is not necessary, this is mistake I fix on the side, see line 39 and 41, it uses `false` for the normal instantiation.

grcevski (approved):
LGTM!

PR #87665, opened 2022-06-14
Ingest: IngestDocument requires non-null version

stu-elastic:
Changes the type of the version paramter in IngestDocument from
Long to long and moves it to the third argument, so all required
values occur before nullable arguments.

The IngestService expects a non-null version for a document and will
throw an NullPointerException if one is not provided.

Related: #87309


rjernst (approved):
LGTM

martijnvg (approved):
👍

PR #87667, opened 2022-06-14
Fixup highlighting with synthetic source

nik9000:
Synthetic source has a habit of reordering text fields. This frustrates
highlighting because it *often* wants to use index structures to find
the offsets to values in the field. This disables the FVH highlighter
for multi-valued text fields when synthetic source is enabled and runs
the unified highlighter in "analyze" mode when synthetic source is
enabled. That's *enough* to stop them from spitting out wrong answers.

We might be leaving some performance on the table when the unified
highlighter works on a single valued text field that is indexed with
offsets or term vectors. We don't really expect that to be common at all
though because *generally* folks will enable synthetic source to save
space and adding offsets or term vectors is quite space inefficient. If
it comes up, we might be able to improve here.


rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.highlight/50_synthetic_source.yml
```diff
@@ -0,0 +1,314 @@
...
+            match_phrase:
+              foo.vectors: "the quick brown fox jumped over the lazy dog"
+          highlight:
+            fields:
+              foo.vectors:
+                type: fvh
+  - match: { hits.hits.0.highlight.foo\.vectors.0: <em>the quick brown fox jumped over the lazy dog</em> }
+
+---
+text multi fvh:
```
romseygeek, line 302: Can you add a test that this also occurs when you ask for fragments in order, so that we exercise both FragmentsBuilder implementations?

romseygeek (approved):
LGTM, thanks Nik.  I think we need one more test to ensure that we cover all paths in FVH but other than that this is good to merge.

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.highlight/50_synthetic_source.yml
nik9000: > Can you add a test that this also occurs when you ask for fragments in order, so that we exercise both FragmentsBuilder implementations?

:+1:

PR #87482, opened 2022-06-07
Adding additional capability to the master_is_stable health indicator service

masseyke:
This PR builds on #86524 by supporting two additional conditions, both of which happen when there has been
no elected master for more than 30 seconds (from the queried node's point of view), and both of which return
a RED status:

1. There are no master-eligible nodes found in the cluster
2. The node being queried sees a master-eligible node that has been elected the master, but cannot join it

This PR does *NOT* cover two additional conditions when there has been no elected master for more than
30 seconds that will be handled in subsequent PRs:

1. There are master-eligible nodes but none is elected master and the queried node is not master-eligible
2. There are master-eligible nodes but none is elected master and the queried node is master-eligible

Here is an example response in the case that there are no master-eligible nodes (case 1 above):
```
{
    "status": "red",
    "cluster_name": "TEST-TEST_WORKER_VM=[--not-gradle--]-CLUSTER_SEED=[-7055790250029418587]-HASH=[CCFA36BEC69DB]-cluster",
    "components": {
        "cluster_coordination": {
            "status": "red",
            "indicators": {
                "master_is_stable": {
                    "status": "red",
                    "summary": "No master eligible nodes found in the cluster",
                    "help_url": "https://ela.st/fix-master",
                    "details": {
                        "recent_masters": [
                            {
                                "node_id": "FhnTO8lgQJeTnL2y_TqjOQ",
                                "name": "node_t0"
                            }
                        ],
                        "cluster_formation": {
                            "description": "master not discovered yet: have discovered [{node_t1}{ZK-WxrjtSQ2FVSS5DOJo_A}{GJ_d8eTwR5yHfJQ3iVmNeQ}{node_t1}{127.0.0.1}{127.0.0.1:49785}{d}]; discovery will continue using [127.0.0.1:49782] from hosts providers and [{node_t0}{FhnTO8lgQJeTnL2y_TqjOQ}{HKoK0SzXSDqqolptGB2TKg}{node_t0}{127.0.0.1}{127.0.0.1:49782}{m}] from last-known cluster state; node term 1, last-accepted version 4 in term 1"
                        }
                    },
                    "impacts": [
                        {
                            "severity": 1,
                            "description": "The cluster cannot create, delete, or rebalance indices, and cannot insert or update documents.",
                            "impact_areas": [
                                "ingest"
                            ]
                        },
                        {
                            "severity": 1,
                            "description": "Scheduled tasks such as Watcher, ILM, and SLM will not work. The _cat APIs will not work.",
[...]

Issue #85624, opened 2022-04-01
Cluster coordination indicator - report if the master is stable and an impact/troubleshoot guide otherwise


andreidan:
Each node will need to keep track of the last few master node changes (this should be fine in memory) and its local node information (a node might _see_ the master node flapping, but the master node itself is fine). e.g. if the master node has changed >3 times in the last 30 minutes, then it's not stable. otherwise, nothing to report.

The coordinating node might have to contact a master eligible node which in turn might have to contact other master eligible nodes, but this is the worst-case scenario and definitely does not involve fanning out to all nodes.

- [x] Store a view of the last 30 minutes of master history on each node, and add the ability to query any node for its view of master history #85941
- [x] Determine if the node has master (if one is present in the cluster state or otherwise if it has seen one in the last 30 seconds) #86524

In case the node has master according to above definition:
- [x]  Check if a seen master is stable - master did not go `null` repeatedly in the last 30 *minutes*. Report GREEN #86524
- [x] Check if master is UNSTABLE - did it change more than 3 times in the last 30 *minutes* ? Report YELLOW (add history info on who was master and when) #86524
   - [x] RCA for unstable master - try and contact a previous master about the reason it stepped down (take disconnect/timeout of the network call into account). For this we'll need to expose the "master history log" we built at the above points over the wire. #86524
- [x] Check if *we* (the node coordinating this check) are unstable - has the witnessed master gone null/not-null more than 3 times in the last 30 minutes but the identity hasn't changed? Report the unstable master case above (report YELLOW) with the same RCA #86524

In case the node does not have a master node: 
- [x] Check if we know of any master-eligible nodes - in case we don't know of any report RED due to a discovery problem (include witnessed masters history). #87482
In case we know of some use the `PeerFinder` to check:
  - [x] Check if one of them is master - in which case report RED as we are having issues joining a cluster #87482
  - [x] None of them is master and we aren't master eligible 
       - [x] RCA reach out to a master eligible node and run the same checks (take disconnect/timeout of the network call into account)
- [x] If *we* are master eligible nodes collect the information form all known master eligible nodes (the information about term/version/voting config should be available in `ClusterFormationFailureHelper`, however we'd need that exposed over the wire #87306 ) -  take disconnect/timeout of the network call into account

- [x] Document the settings we create for the master stability and cluster diagnostics service(s) #87901

- [x] Create `master-is-stable` indicator
  - [x] Outline the details we'll be able to provide on drill down
[...]

PR #87672, opened 2022-06-14
Move the master stability logic into its own service separate from the HealthIndicatorService

masseyke:
For various reasons, we want the logic that decides whether we have had a stable master node separated from the logic that generates the HealthIndicatorResult (which is specific to the health API). This commit separates the logic of the master stability code from the HealthIndicatorService code. Most pressingly, we need to be able to call the stable master logic remotely from within StableMasterHistoryIndicatorService (when there has been no master recently, there is at least one master eligible node, none of those nodes are elected master, and the current node is not master-eligible). This commit splits StableMasterHistoryIndicatorService into StableMasterHistoryIndicatorService and StableMasterService. StableMasterService has a calculate method much like StableMasterHistoryIndicatorService's, but it returns a custom result with only stable-master-specific information. The job of StableMasterHistoryIndicatorService is just to transform this result into a HealthIndicatorResult.

The goal of this change is just to split the logic between StableMasterHistoryIndicatorService and StableMasterService. No externally-visible functionality is being added. A follow-up PR will make StableMasterService available through a transport action (which will include making the output of StableMasterService.calculate() serializable). And a follow-up PR to that will make use of that transport action (in the 1.2.2.3 branch of the diagram at https://github.com/elastic/elasticsearch/pull/87482#issuecomment-1155285356).
Relates #85624


server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterService.java
```diff
@@ -0,0 +1,348 @@
...
+ * this will report GREEN.
+ * If we have had a master within the last 30 seconds, but that master has changed more than 3 times in the last 30 minutes (and that is
+ * confirmed by checking with the last-known master), then this will report YELLOW.
+ * If we have not had a master within the last 30 seconds, then this will will report RED with one exception. That exception is when:
+ * (1) no node is elected master, (2) this node is not master eligible, (3) some node is master eligible, (4) we ask a master-eligible node
+ * to run this service, and (5) it comes back with a result that is not RED.
+ * Since this service needs to be able to run when there is no master at all, it does not depend on the dedicated health node (which
+ * requires the existence of a master).
+ */
+public class StableMasterService implements ClusterStateListener {
```
andreidan, line 42: As this would now be a standalone service, should we name this something a bit more descriptive (this service will look into discovery and cluster formation - and maybe expose some of those analyses via different public methods at some point, even though I do appreciate now it only has one public method)

Would `ClusterAnalysisService`/ `CoordinationAnalysisService` / `CoordinationDiagnosticsService` be a bit more descriptive? 
With a method `analyseMasterStability`/ `diagnoseMasterStability` instead of `calculate` ? 
In the same context, would `StableMasterDetails` be better named something along the lines of `CoordinationAnalysisDetails`?

These are all off-the-cuff suggestions (naming is, well, hard :) )

cc @DaveCTurner

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterService.java
```diff
@@ -0,0 +1,348 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.elasticsearch.cluster.ClusterChangedEvent;
+import org.elasticsearch.cluster.ClusterStateListener;
+import org.elasticsearch.cluster.node.DiscoveryNode;
+import org.elasticsearch.cluster.service.ClusterService;
+import org.elasticsearch.common.settings.Setting;
+import org.elasticsearch.core.Nullable;
+import org.elasticsearch.core.TimeValue;
+import org.elasticsearch.health.HealthStatus;
```
andreidan, line 20: This creates a circular dependency - the `cluster.coordination` package depends on the `health`package and `health` depends on `cluster.coordination`.

I think we should either have a status entity in `cluster.coordination` and interpret it when this service is consumed, or don't have a status at all but a different way to express the result of this service - a diagnostic /analysis of sorts

server/src/main/java/org/elasticsearch/common/settings/ClusterSettings.java
```diff
@@ -514,9 +514,9 @@ public void apply(Settings value, Settings current, Settings previous) {
         IndexingPressure.MAX_INDEXING_BYTES,
         ShardLimitValidator.SETTING_CLUSTER_MAX_SHARDS_PER_NODE_FROZEN,
         DataTier.ENFORCE_DEFAULT_TIER_PREFERENCE_SETTING,
-        StableMasterHealthIndicatorService.IDENTITY_CHANGES_THRESHOLD_SETTING,
-        StableMasterHealthIndicatorService.NO_MASTER_TRANSITIONS_THRESHOLD_SETTING,
-        StableMasterHealthIndicatorService.NODE_HAS_MASTER_LOOKUP_TIMEFRAME_SETTING,
+        StableMasterService.IDENTITY_CHANGES_THRESHOLD_SETTING,
+        StableMasterService.NO_MASTER_TRANSITIONS_THRESHOLD_SETTING,
+        StableMasterService.NODE_HAS_MASTER_LOOKUP_TIMEFRAME_SETTING,
```
andreidan, line 519: Should we document these settings?

andreidan (commented):

I think this looks good, just left a few minor questions and suggestions about having the terminology more targeted towards a standalone service (that could be used outside of the health API context) and modularity

server/src/main/java/org/elasticsearch/common/settings/ClusterSettings.java
masseyke: I think we ought to. I've been trying to figure out where would be a good place to document these without cluttering more important documentation. Have any ideas? Maybe a settings page for the whole health API?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterService.java
masseyke: Oh good catch. I'll fix that.

server/src/main/java/org/elasticsearch/common/settings/ClusterSettings.java
andreidan: Shall we have a page under https://www.elastic.co/guide/en/elasticsearch/reference/current/settings.html that's called something around what this service will end up being called? 
ie. cluster diagnostics settings? 

These settings will need some big warnings like [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-discovery-settings.html#_expert_settings)

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterService.java
masseyke: Sounds as good as anything. I'll go with that unless someone comes up with something better.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -0,0 +1,354 @@
...
+        return new StableMasterResult(CoordinationDiagnosticsStatus.GREEN, summary, details);
+    }
+
+    /**
+     * Returns the health result for the case when we have NOT seen a master recently (at some point in the last 30 seconds).
+     * @param localMasterHistory The master history as seen from the local machine
+     * @param explain Whether to calculate and include the details in the result
+     * @return The StableMasterResult for the given localMasterHistory
+     */
+    private StableMasterResult diagnoseOnHaveNotSeenMasterRecently(MasterHistory localMasterHistory, boolean explain) {
```
andreidan, line 275: nit: should we name this `Diagnostic` or something along those lines?

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -0,0 +1,354 @@
...
+    public record StableMasterResult(CoordinationDiagnosticsStatus status, String summary, StableMasterDetails details) {}
+
+    public enum CoordinationDiagnosticsStatus {
+        GREEN,
+        UNKNOWN,
+        YELLOW,
+        RED;
+    }
+
+    public record StableMasterDetails(
```
andreidan, line 328: nit: should we name this `DiagnosticDetails`?

server/src/main/java/org/elasticsearch/health/HealthStatus.java
```diff
@@ -52,4 +53,15 @@ public static HealthStatus merge(Stream<HealthStatus> statuses) {
     public String xContentValue() {
         return name().toLowerCase(Locale.ROOT);
     }
+
+    public static HealthStatus fromStableMasterStatus(
```
andreidan, line 57: nit: should we name this `fromDiagnosticStatus` ?

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -0,0 +1,354 @@
...
+
+    /**
+     * Returns the details for the calculateOnMasterHasFlappedNull method. This method populates the StableMasterDetails
+     * with the currentMaster, and optionally the remoteExceptionMessage and remoteExceptionStackTrace.
+     * @param explain If false, nothing is calculated and StableMasterDetails.EMPTY is returned
+     * @param localMasterHistory The localMasterHistory
+     * @param remoteHistoryException An exception that was found when retrieving the remote master history. Can be null
+     * @return The StableMasterDetails
+     */
+    private StableMasterDetails getStableMasterDetailsOnMasterHasFlappedNull(
```
andreidan, line 247: nit: would "getDetailsOnMasterHasFlappedNull` be a bit more concise?

andreidan (approved):
LGTM, thanks for working on this Keith 🚀

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: Oops, yeah. I thought I'd caught all of these.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -0,0 +1,354 @@
...
+                 * box's view of the master history is
+                 */
+                if (master != null && clusterService.localNode().equals(master) == false) {
+                    masterHistoryService.refreshRemoteMasterHistory(master);
+                }
+            }
+        }
+    }
+
+    public record StableMasterResult(CoordinationDiagnosticsStatus status, String summary, StableMasterDetails details) {}
```
jbaiera, line 319: Should this also be `CoordinationDiagonosticResult`?

jbaiera (approved):
LGTM, pending the last few name changes left

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: Ohh for some reason intellij refactoring is silently failing on these! So I only thought I'd changed them!

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: Yeah, intellij refactoring failed silently. Search and replace worked though!

PR #87673, opened 2022-06-14
Ingest: Add validation and strong typing to sourceAndMetdata map

stu-elastic:
Adds `IngestSourceAndMetdata` to replace the sourceAndMetadata map. 

This validates metadata values when they are added to the map for use in
scripts and other process as well as provides typed getters and for use
inside of server.

This change lays the foundation for strongly typed Metadata access in scripting.


modules/ingest-common/src/test/java/org/elasticsearch/ingest/common/AppendProcessorTests.java
```diff
@@ -121,39 +120,6 @@ public void testConvertScalarToList() throws Exception {
         }
     }
 
-    public void testAppendMetadataExceptVersion() throws Exception {
```
rjernst, line 124: why is this test removed?

modules/ingest-common/src/test/java/org/elasticsearch/ingest/common/DotExpanderProcessorTests.java
```diff
@@ -46,7 +46,7 @@ public void testEscapeFields() throws Exception {
         document = new IngestDocument(source, Collections.emptyMap());
         processor = new DotExpanderProcessor("_tag", null, null, "foo.bar");
         processor.execute(document);
-        assertThat(document.getSourceAndMetadata().size(), equalTo(1));
+        assertThat(document.getSourceAndMetadata().size(), equalTo(2));
```
rjernst, line 49: Why did the map size change?

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
```diff
@@ -886,36 +893,6 @@ public String toString() {
         return "IngestDocument{" + " sourceAndMetadata=" + sourceAndMetadata + ", ingestMetadata=" + ingestMetadata + '}';
     }
 
-    public enum Metadata {
```
rjernst, line 889: Can we keep this here? It creates a lot of churn, and I don't think there is any real reason it needs to be inside the implementation specific map.

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
```diff
@@ -0,0 +1,260 @@
...
+ *
+ * The Metadata values in {@link Metadata} are validated when put in the map.
+ * _index, _id and _routing must be a String or null
+ * _version_type must be a lower case VersionType or null
+ * _version must be representable as a long without loss of precision and may not be null
+ *
+ * The map is expected to be used by processors where-as the typed getter and setters should
+ * be used by server code where possible.
+ */
+public class IngestSourceAndMetadata extends ValidatingMap {
```
rjernst, line 38: We should be able to make this an impl detail, the type doesn't need to be exposed, so it can be package private?

server/src/main/java/org/elasticsearch/script/field/ValidatingMap.java
```diff
@@ -0,0 +1,259 @@
...
+ * ValidatingMap is an implementation of {@code Map<String, Object>} that has a validating functions for a subset of keys.
+ *
+ * The validators are BiFunctions that should either return the identity or throw an {@link IllegalArgumentException}.  The
+ * arguments to the validator are the key and the proposed value or {@code null} for key removal.
+ *
+ * Validation occurs everywhere updates can happen, either directly via the {@link #put(String, Object)}, {@link #replace(Object, Object)},
+ * {@link #merge(Object, Object, BiFunction)} family of methods, via {@link Map.Entry#setValue(Object)},
+ * via the linked Set from {@link #entrySet()} or {@link Collection#remove(Object)} from the linked collection via {@link #values()}
+ */
+public class ValidatingMap extends AbstractMap<String, Object> {
```
rjernst, line 31: This can be package private? Or perhaps this doesn't need to be a separate class, we just need it for ingest right now, let's fold this into the real impl class and split if there is a need in the future?

server/src/main/java/org/elasticsearch/script/field/ValidatingMap.java
```diff
@@ -0,0 +1,259 @@
...
+        public Map.Entry<String, Object> next() {
+            return cur = new Entry(iterator.next());
+        }
+
+        /**
+         * Remove current entry from the backing Map.  Checks the Entry's key's validator, if one exists, before removal.
+         * @throws IllegalArgumentException if the validator does not allow the Entry to be removed
+         */
+        @Override
+        public void remove() {
```
rjernst, line 222: Wouldn't the iterator remove eventually delegate to a remove on the map, which will run the validator? Or why is a validator run at all? It seems like a hacky way to disallow removal.

rjernst (commented):
Not a full review, but a few quick thoughts

modules/ingest-common/src/test/java/org/elasticsearch/ingest/common/DotExpanderProcessorTests.java
stu-elastic: _version is added in the test constructor so it doesn't have to be added in every test that uses IngestDocument.

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
stu-elastic: Will do.

modules/ingest-common/src/test/java/org/elasticsearch/ingest/common/DotExpanderProcessorTests.java
rjernst: This seems trappy. Just above here the IngestDocument is created. If we need helpers to not have to specify version or other metadata fields, let's add helper methods, but not make it magical. If I were debugging this code, I would be very confused why I see 2 here.

server/src/main/java/org/elasticsearch/script/field/ValidatingMap.java
stu-elastic: There is this map and the wrapped map.  This iterator is needed because `AbstractSet` requires subclasses to implement `iterator()`.  If we just take the underlying maps entry set, the validators will not be called.

rjernst: Ok, but why are the validators run on removal?

stu-elastic: You cannot remove `_version`.  If scripts remove `_version` then either we need to return a sentinel value from `getVersion()` (`Long.MIN_VALUE`  or similar) or we have to throw an `IllegalArgumentException`

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
```diff
@@ -0,0 +1,260 @@
...
+     * Get Metadata value using the field name
+     */
+    public Object get(Metadata field) {
+        return get(field.fieldName);
+    }
+
+    /**
+     * One time operation that removes all metadata values and their validators and returns the underlying map, containing only source keys
+     */
+    public Map<String, Object> extractSource() {
```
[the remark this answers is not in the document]
stu-elastic: The existing pattern of use in `IngestDocument` is to remove the `Metadata` and then pass resulting map as the source.  The `IngestDocument` is [thrown away](https://github.com/elastic/elasticsearch/blob/d42211c4314aaeec0211f3cad40feea90c4ab1c5/server/src/main/java/org/elasticsearch/ingest/IngestService.java#L900..L920) so there's no need to allocate another, potentially large, map to hold the source to avoid modification.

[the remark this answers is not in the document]
stu-elastic: > Alternatively, what if source was kept separate, and delegated to.

We can do that, it would slightly complicate `EntrySet` but that's not a problem

server/src/main/java/org/elasticsearch/script/field/ValidatingMap.java
stu-elastic: Removed as a separate class.

server/src/main/java/org/elasticsearch/script/field/ValidatingMap.java
stu-elastic: Validators are still run on removal because a validator may raise on removal, however this change does not have any such validator.

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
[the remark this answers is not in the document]
stu-elastic: `source` and `metadata` maps are separated.

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
stu-elastic: Done.

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
```diff
@@ -707,26 +719,31 @@ private Map<String, Object> createTemplateModel() {
...
-        }
-        return metadataMap;
+    public IngestSourceAndMetadata getIngestSourceAndMetadata() {
+        return sourceAndMetadata;
+    }
+
+    /**
+     * Get all Metadata values in a Map
+     */
+    public Map<String, Object> getMetadata() {
```
rjernst, line 738: Can users get call getSourceAndMetadata() and call the appropriate method instead of adding access here?

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
```diff
@@ -738,12 +755,17 @@ public Map<String, Object> getIngestMetadata() {
...
     /**
-     * Returns the document including its metadata fields, unless {@link #extractMetadata()} has been called, in which case the
-     * metadata fields will not be present anymore.
-     * Modify the document instead using {@link #setFieldValue(String, Object)} and {@link #removeField(String)}
+     * Fetch the timestamp from the ingestMetadata, if it exists
+     * @return the timestamp for the document or null
      */
-    public Map<String, Object> getSourceAndMetadata() {
-        return this.sourceAndMetadata;
+    public static ZonedDateTime getTimestamp(Map<String, Object> ingestMetadata) {
```
rjernst, line 761: This seems like an odd place for a utility method. Is there another place it could live, perhaps on metadata itself?

test/framework/src/main/java/org/elasticsearch/ingest/TestIngestDocument.java
```diff
@@ -34,10 +39,30 @@ public static IngestDocument ofSourceAndMetadata(Map<String, Object> sourceAndMe
         return new IngestDocument(sourceAndMetadata, new HashMap<>());
     }
 
+    /**
+     * Create an IngestDocument for testing but takes a pre-constructed {@link IngestSourceAndMetadata}
```
rjernst, line 43: this comment doesn't seem to match the signature

server/src/test/java/org/elasticsearch/ingest/IngestSourceAndMetadataTests.java
```diff
@@ -0,0 +1,355 @@
...
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasEntry;
+import static org.hamcrest.Matchers.notNullValue;
+
+public class IngestSourceAndMetadataTests extends ESTestCase {
+
+    IngestSourceAndMetadata map;
+
+    @Override
+    public void setUp() throws Exception {
```
rjernst, line 29: this seems unnecesary, maybe leftover from debugging?

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
```diff
@@ -0,0 +1,613 @@
...
+ * _if_primary_term must be a long or null
+ *
+ * The map is expected to be used by processors, server code should the typed getter and setters where possible.
+ */
+class IngestSourceAndMetadata extends AbstractMap<String, Object> {
+    protected final ZonedDateTime timestamp;
+
+    /**
+     * map of key to validating function. Should throw {@link IllegalArgumentException} on invalid value, otherwise return identity
+      */
```
rjernst, line 48: nit: spacing on javadoc off :)

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
```diff
@@ -0,0 +1,613 @@
...
+ * _if_seq_no must be a long or null
+ * _if_primary_term must be a long or null
+ *
+ * The map is expected to be used by processors, server code should the typed getter and setters where possible.
+ */
+class IngestSourceAndMetadata extends AbstractMap<String, Object> {
+    protected final ZonedDateTime timestamp;
+
+    /**
+     * map of key to validating function. Should throw {@link IllegalArgumentException} on invalid value, otherwise return identity
```
rjernst, line 47: Why make it return anything? If validators throw IAE, then they could just be Consumers?

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
```diff
@@ -0,0 +1,613 @@
...
+
+    /**
+     * Create a IngestSourceAndMetadata using the underlying map and set of validators.  The validators are applied to the map to ensure
+     * the incoming map matches the invariants enforced by the validators.
+     * @param sourceAndMetadata the wrapped map.  Should not be externally modified after creation.  This map is modified to remove
+     *                          metadata values and will become source
+     * @param timestamp the timestamp of ingestion
+     * @throws IllegalArgumentException if a validator fails for a given key
+     */
+    public static IngestSourceAndMetadata ofMixedSourceAndMetadata(Map<String, Object> sourceAndMetadata, ZonedDateTime timestamp) {
```
rjernst, line 126: Similar to IngestDocument, there seem to be a lot of different ways to construct an IngestSourceAndMetadata. Can these be consolidated?

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
```diff
@@ -0,0 +1,613 @@
...
+        Number version = getNumber(IngestDocument.Metadata.VERSION.getFieldName());
+        assert version != null : IngestDocument.Metadata.VERSION.getFieldName() + " validation allowed null version";
+        return version.longValue();
+    }
+
+    public void setVersion(long version) {
+        put(IngestDocument.Metadata.VERSION.getFieldName(), version);
+    }
+
+    // timestamp isn't backed by the map
```
rjernst, line 234: why isn't timestamp part of the metadata map?

rjernst (commented):
This is looking good. A few more comments

modules/ingest-common/src/test/java/org/elasticsearch/ingest/common/AppendProcessorTests.java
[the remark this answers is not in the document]
stu-elastic: Yes, the only metadata that doesn't fail is `_type` and I think that's because it's missing validation.

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
stu-elastic: `org.elasticsearch.action.ingest.WriteableIngestDocument` can't see the package private `org.elasticsearch.ingest.IngestSourceAndMetadata`

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
stu-elastic: It's a new field (https://github.com/elastic/elasticsearch/issues/80038) so I didn't think it was necessary to backport it into the ctx map and create a potential bwc issue.

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
```diff
@@ -97,6 +102,13 @@ public static IngestDocument of(Map<String, Object> sourceAndMetadata, Map<Strin
         return new IngestDocument(sourceAndMetadata, ingestMetadata);
     }
 
+    /**
+     * Build an IngestDocument from values read via deserialization
+     */
+    public static IngestDocument of(Map<String, Object> source, Map<String, Object> metadata, Map<String, Object> ingestMetadata) {
```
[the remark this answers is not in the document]
stu-elastic: I will take another pass.

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
```diff
@@ -707,26 +719,31 @@ private Map<String, Object> createTemplateModel() {
...
-     * Metadata fields that used to be accessible as ordinary top level fields will be removed as part of this call.
+     * Get source and metadata map
      */
-    public Map<Metadata, Object> extractMetadata() {
-        Map<Metadata, Object> metadataMap = new EnumMap<>(Metadata.class);
-        for (Metadata metadata : Metadata.values()) {
-            metadataMap.put(metadata, sourceAndMetadata.remove(metadata.getFieldName()));
-        }
-        return metadataMap;
+    public Map<String, Object> getSourceAndMetadata() {
```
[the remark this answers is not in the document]
stu-elastic: We lose the fact that it's an `IngestSourceAndMetadata` if we make the signature a `Map<String, Object>` (which loses access to the typed getters) and if we return an `IngestSourceAndMetadata` then many of the tests for the ingest processors fail to compile because `IngestSourceAndMetadata` is package private, eg

````
modules/ingest-geoip/src/test/java/org/elasticsearch/ingest/geoip/GeoIpProcessorFactoryTests.java:497:
 error: IngestSourceAndMetadata.get(Object) is defined in an inaccessible class or interface                                          
            Map<?, ?> geoData = (Map<?, ?>) ingestDocument.getIngestSourceAndMetadata().get("geoip");  
```

Similar in `org.elasticsearch.action.ingest.AsyncIngestProcessorIT`.

test/framework/src/main/java/org/elasticsearch/ingest/TestIngestDocument.java
stu-elastic: Fixed.

server/src/test/java/org/elasticsearch/ingest/IngestSourceAndMetadataTests.java
stu-elastic: Removed.

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
stu-elastic: Fixed.

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
stu-elastic: Changed to BiConsumers

stu-elastic:

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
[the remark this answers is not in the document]
stu-elastic: Down to the same number as before.

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
stu-elastic: Reduced to two.

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
stu-elastic: Moved to IngestSourceAndMetadata.

stu-elastic:

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
```diff
@@ -706,43 +718,39 @@ private Map<String, Object> createTemplateModel() {
...
-     * Does the same thing as {@link #extractMetadata} but does not mutate the map.
+     * Get source and metadata map as {@link IngestSourceAndMetadata}
      */
-    public Map<Metadata, Object> getMetadata() {
-        Map<Metadata, Object> metadataMap = new EnumMap<>(Metadata.class);
-        for (Metadata metadata : Metadata.values()) {
-            metadataMap.put(metadata, sourceAndMetadata.get(metadata.getFieldName()));
-        }
-        return metadataMap;
+    public IngestSourceAndMetadata getIngestSourceAndMetadata() {
```
rjernst, line 730: since IngestSourceAndMetadata is package private, this method can be as well?

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
```diff
@@ -58,42 +58,54 @@ public final class IngestDocument {
...
-        this.ingestMetadata = ingestMetadata;
+    public IngestDocument(Map<String, Object> sourceAndMetadata, Map<String, Object> ingestMetadata) {
+        Tuple<Map<String, Object>, Map<String, Object>> sm = IngestSourceAndMetadata.splitSourceAndMetadata(sourceAndMetadata);
+        this.sourceAndMetadata = new IngestSourceAndMetadata(
+            sm.v1(),
+            sm.v2(),
+            IngestSourceAndMetadata.getTimestamp(ingestMetadata),
+            IngestSourceAndMetadata.VALIDATORS
+        );
+        this.ingestMetadata = ingestMetadata != null ? ingestMetadata : new HashMap<>();
```
rjernst, line 100: Doesn't this need to be the metadata from sourceAndMetadata? If the assumption is it should already exist in the passed in sourceAndMetadata, then shouldn't ingestMetadata always be non null? The naming here is a bit confusing because "sourceAndMetadata" implies it contains the metadata, but then ingestMetadata is *also* passed in. When this was package private it's scoped more narrowly, but making this ctor public opens this up to trappy usage.

server/src/main/java/org/elasticsearch/ingest/IngestSourceAndMetadata.java
```diff
@@ -0,0 +1,588 @@
...
+    public ZonedDateTime getTimestamp() {
+        return timestamp;
+    }
+
+    // These are not available to scripts
```
rjernst, line 222: If they are not available to scripts, why are they in the metadata map? It doesn't appear to be added when the IngestSourceAndMetadata is created, so when and how does it get in there?

test/framework/src/main/java/org/elasticsearch/ingest/TestIngestDocument.java
```diff
@@ -8,30 +8,32 @@
...
-     * provided elasticsearch metadata, source and ingest metadata.
-     *
-     * This is needed because the ingest metadata will be initialized with the current timestamp at init time, which makes equality
-     * comparisons impossible in tests.
+     * Create an IngestDocument for testing that pass an empty mutable map for ingestMetaata
      */
-    public static IngestDocument ofSourceAndIngest(Map<String, Object> sourceAndMetadata, Map<String, Object> ingestMetadata) {
-        return new IngestDocument(sourceAndMetadata, ingestMetadata);
+    public static IngestDocument ofSourceAndMetadata(Map<String, Object> sourceAndMetadata) {
+        return new IngestDocument(sourceAndMetadata, new HashMap<>());
```
rjernst, line 28: Doesn't this need to be the metadata from the sourceAndMetadata?

rjernst (approved):
Looks fine. I left a couple more thoughts.

test/framework/src/main/java/org/elasticsearch/ingest/TestIngestDocument.java
stu-elastic: No, there's a separate IngestMetadata for internal processor use.

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
stu-elastic: Removed `HashMap` instantiation on `null`.

PR #87674, opened 2022-06-14
Mark `apm_user` for removal in "a future version"

sorenlouv:
Meta issue: https://github.com/elastic/kibana/issues/116760
The `apm_user` role was [marked as deprecated](https://github.com/elastic/elasticsearch/pull/68749) in 7.13 and was supposed to be removed in 8.0 but it didn't happen. Now we are aiming to remove the role in 9.0 and are updating the deprecation message.

All mentions of `apm_user` role have been removed from docs and in-product mentions in https://github.com/elastic/kibana/pull/132790.


ywangd (dismissed):
I think we should use `the next major release` instead of `9.0` since the major release strategy is not entirely clear. Also most places in the code use "the next major release/version", e.g.:
https://github.com/elastic/elasticsearch/blob/5b6838e6ec4cb205b87cb5b6e58167f9449c36e4/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/IndicesPermission.java#L471

bytebilly:
> I think we should use `the next major release` instead of `9.0` since the major release strategy is not entirely clear.

We could be even more generic with "a future major release". It was not so uncommon that we decided to keep a deprecated feature across majors, and in that case the message may be confusing.

sqren:
I changed the deprecation to "This role will be removed in a future major release. Please use editor and viewer roles instead"

ywangd:
@elasticmachine update branch

ywangd:
@pugnascotia CI fails for this PR because it does not like the area label of deprecation. Existing ones are

https://github.com/elastic/elasticsearch/blob/fc09896ec63393caee7beace584eba49664bb8d0/build-tools-internal/src/main/resources/changelog-schema.json#L177-L191

But none of them seems to fit the deprecation here which is for a builtin role. What would be your recommendation? Thanks!

sqren:
I dropped this but this is still something I'd like to merge. 

@pugnascotia Afaict I added "Authorization" in https://github.com/elastic/elasticsearch/pull/87674/commits/72b1163e5325fd42d3ea5595a75c320552b65067. Still, I'm seeing a bunch of failures. 

What's the suggested way forward?

sqren:
Also, I don't see the button to rebase with main - is there a Github comment to do that?

pugnascotia:
@elasticmachine update branch

pugnascotia:
@sqren I updated the branch but there are formatting problems. To fix them, update your local branch checkout and run:

    ./gradlew spotlessApply

Then commit and push. I don't believe we have automation for this at present.

pugnascotia (approved):

PR #87675, opened 2022-06-15
Support removing ignore filters for audit logging

### LINKED ISSUE #68588: Audit trail filtering rules cannot be deleted in the usual way other cluster settings are removed.
A user reported an issue where they wanted to delete an existing [audit filter rule ](https://www.elastic.co/guide/en/elasticsearch/reference/7.9/auditing-settings.html#audit-event-ignore-policies) by changing the cluster setting to null (the usual way cluster settings are effectively removed).
Unfortunately the net effect was to create a rule that filtered all audit log entries.

I reproduced this by adding this test to [LoggingAuditTrailFilterTests](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/audit/logfile/LoggingAuditTrailFilterTests.java)

    public void testNullPolicyDoesNotMatchEvent() throws Exception {
        final Logger logger = CapturingLogger.newCapturingLogger(Level.INFO, null);
        final ThreadContext threadContext = new ThreadContext(Settings.EMPTY);
        final Settings.Builder settingsBuilder = Settings.builder().put(settings);
        settingsBuilder.putNull("xpack.security.audit.logfile.events.ignore_filters.userPolicy.users");
        final LoggingAuditTrail auditTrail = new LoggingAuditTrail(settingsBuilder.build(), clusterService, logger, threadContext);
        final User unfilteredUser = new User("Fred");
        // Null setting should not match
        assertFalse("Shouldn't match users wiih a null rule",
                auditTrail.eventFilterPolicyRegistry.ignorePredicate()
                        .test(new AuditEventMetaInfo(Optional.of(unfilteredUser), Optional.empty(), Optional.empty(), Optional.empty())));
    }

ywangd:
Ignore filters of audit logging can be configured with the cluster
settings APIs. While adding new filters work correclty, removing
filters does not work until node restart due to a bug (#68588). This PR
fixes the bug by correctly remove the ignore filter when all rules of a
filtering policy is set to null.

Resolves: #68588


ywangd:
@elasticmachine update branch

albertzaharovits (approved):
LGTM

I did use the settings API to run a couple of policy update and removes, as I find the `addAffixGroupUpdateConsumer` logic, and the settings updater infra, in general, very hard to follow.

PR #87678, opened 2022-06-15
[ML] Skip ML BWC tests against versions before 7.17.5 and 8.2.2 on RHEL 9

droberts195:
RHEL 9 suffers the same problems as Ubuntu 22.04 that were fixed by
https://github.com/elastic/ml-cpp/pull/2272.

The glibc version on RHEL 9 is 2.34 (versus 2.35 on Ubuntu 22.04),
so we need to adjust the skip logic.


dimitris-athanasiou (approved):
LGTM

PR #87681, opened 2022-06-15
Make GetIndexAction cancellable

DaveCTurner:
The get-indices API does some nontrivial work on the master and at high
index counts the response may be very large and could take a long time
to compute. Some clients will time out and retry if it takes too long.
Today this API is not properly cancellable which leads to a good deal of
wasted work in this situation, and the potentially-enormous response is
serialized on a transport worker thread.

With this commit we make the API cancellable and move the serialization
to a `MANAGEMENT` thread.

Relates https://github.com/elastic/elasticsearch/issues/77466

### DIFF
--- a/docs/changelog/87681.yaml
+++ b/docs/changelog/87681.yaml
@@ -1,0 +1,5 @@
+pr: 87681
+summary: Make `GetIndexAction` cancellable
+area: Indices APIs
+type: bug
+issues: []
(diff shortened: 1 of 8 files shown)


DaveCTurner:
Backporting this to 7.17 is debatable, although I think it's reasonable to do so.

martijnvg (approved):
LGTM, thanks David.
I wonder what other APIs we may need to make this change for?
(currently: cluster state, get snapshot, get mapping, segment stats APIs?)

PR #87685, opened 2022-06-15
Cleanup indirection in batch close utils

original-brownbear:
Some of these show up in profiling of heavy transport load.
It's not a big thing but it's always nice to save a few cycles
on transport threads and save a few LoC.

### DIFF
--- a/libs/core/src/main/java/org/elasticsearch/core/IOUtils.java
+++ b/libs/core/src/main/java/org/elasticsearch/core/IOUtils.java
@@ -31,8 +31,6 @@ import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
 import java.nio.file.attribute.BasicFileAttributes;
-import java.util.Arrays;
-import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


libs/core/src/main/java/org/elasticsearch/core/IOUtils.java
```diff
@@ -95,50 +113,38 @@ public static void close(final Exception e, final Closeable... objects) throws I
...
-                throw (IOException) firstException;
-            } else {
-                // since we only assigned an IOException or a RuntimeException to ex above, in this case ex must be a RuntimeException
-                throw (RuntimeException) firstException;
-            }
+            throwRuntimeOrIOException(firstException);
         }
     }
 
+    private static Exception addOrSuppress(Exception firstException, Exception e) {
```
rjernst, line 130: The naming here is confusing. Perhaps call it `maybeSuppress`?

rjernst (approved):
Looks fine. One nit on naming

libs/core/src/main/java/org/elasticsearch/core/IOUtils.java
original-brownbear: I chose this because we have the exact same method in the server package's exception helpers and was going to dry this up in a follow-up. But yea it is a confusing name I agree, we can change it on the utility method in server in that follow up :)

Issue #86059, opened 2022-04-21
Circuit breaker with negative `used` value


DaveCTurner:
### Elasticsearch Version

7.17.2, 8.1.1 and 8.1.2

### Installed Plugins

Various

### Java Version

_bundled_

### OS Version

Cloud

### Problem Description

I've observed a few new-ish clusters in the wild that are closing transport connections because of an exception constructing a circuit-breaker message due to the circuit breaker having a negative `used` value. It seems a little extreme to fail like this, but then again this situation would trip an assertion if assertions were enabled.

### Steps to Reproduce

Unknown at time of writing 😞

Seems pretty rare: looking at the past week's data I only see this in a handful of clusters, running 7.17.2, 8.1.1 and 8.1.2.

Not even sure how to label this. I suspect something networky but there are other places that interact with circuit breakers.

### Logs (if relevant)

```
[instance-0000000xxx] exception caught on transport layer [Netty4TcpChannel{localAddress=/xx.xx.xx.xx:xxxxx, remoteAddress=/xx.xx.xx.xx:xxxxx, profile=default}], closing connection
java.lang.IllegalArgumentException: Values less than -1 bytes are not supported: -370832b
	at org.elasticsearch.common.unit.ByteSizeValue.<init>(ByteSizeValue.java:85) ~[elasticsearch-8.1.2.jar:8.1.2]
	at org.elasticsearch.common.unit.ByteSizeValue.<init>(ByteSizeValue.java:80) ~[elasticsearch-8.1.2.jar:8.1.2]
	at org.elasticsearch.indices.breaker.HierarchyCircuitBreakerService.lambda$checkParentLimit$8(HierarchyCircuitBreakerService.java:431) ~[elasticsearch-8.1.2.jar:8.1.2]
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197) ~[?:?]
	at java.util.Iterator.forEachRemaining(Iterator.java:133) ~[?:?]
	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1845) ~[?:?]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509) ~[?:?]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499) ~[?:?]
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921) ~[?:?]
[...]

PR #87687, opened 2022-06-15
Make CBE message creation more robust

DaveCTurner:
Child circuit breakers rely on proper matching of acquire/release pairs.
This can be tricky to get right. If we get it wrong and accidentally
double-release a CB then it may end up with a negative `used` value.
This is definitely a bad situation in which to find ourselves, but today
in production it's made a whole lot worse because it causes exceptions
on every attempt to report a `CircuitBreakerStats` or to construct a
parent `CircuitBreakingException`.

This commit makes the message construction and stats serialization a
little more robust so that it's clearer what is going on in production.

Relates #86059.


grcevski (approved):
LGTM! Does it make sense to log a warning of some kind when we encounter a negative value with a stack trace? I'm hoping here that with enough stack traces we could at least pinpoint the general area where we have this double decrement, or perhaps that won't help us much.

DaveCTurner:
Sure, I added such logging in 5ff3a42. This should at least tell us which CB is affected in a way that's easy to search across all clusters' logs.

DaveCTurner:
Reverted due to test failures.

PR #87689, opened 2022-06-15
Add help guide to the contact support user action

andreidan:
This adds a help URL for the "contact support" user action
and rewords a bit the message to "get help".

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
@@ -56,6 +56,7 @@ public class StableMasterHealthIndicatorService implements HealthIndicatorServic
 
     public static final String NAME = "master_is_stable";
     private static final String HELP_URL = "https://ela.st/fix-master";
+    public static final String GET_HELP_GUIDE = "https://ela.st/getting-help";
 
     private final ClusterService clusterService;
     private final MasterHistoryService masterHistoryService;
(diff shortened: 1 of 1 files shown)


Leaf-Lin (approved):
LGTM! One tiny suggestion, can we also add the [master sizing guidance link](https://www.elastic.co/guide/en/elasticsearch/reference/master/size-your-shards.html#shard-count-recommendation) so that some users will still be able to self-help with stabilising their master node?

server/src/main/java/org/elasticsearch/cluster/coordination/StableMasterHealthIndicatorService.java
```diff
@@ -269,9 +270,8 @@ private List<UserAction> getContactSupportUserActions(boolean explain) {
         if (explain) {
             UserAction.Definition contactSupport = new UserAction.Definition(
                 "contact_support",
-                "The Elasticsearch cluster does not have a stable master node. Please contact Elastic Support "
-                    + "(https://support.elastic.co) to discuss available options.",
-                null
+                "The Elasticsearch cluster does not have a stable master node. Get help at " + GET_HELP_GUIDE,
```
Leaf-Lin, line 273: I wonder if we can point users to this page as well? 
```suggestion
                "The Elasticsearch cluster does not have a stable master node. You may need to scale up the master node based on sizing guidance. Need more help? " + GET_HELP_GUIDE,
```
Can you also link the scale up the master node to this page? https://www.elastic.co/guide/en/elasticsearch/reference/master/size-your-shards.html#shard-count-recommendation

andreidan:
> so that some users will still be able to self-help with stabilising their master node?

@Leaf-Lin When the master_is_stable indicator will report yellow/red it'll likely be the case of either a platform bug or a cluster that barely holds itself together. I'm not sure a scale-up would work in cloud in either of these cases? 
I'm also a bit reluctant in recommending a user action (ie. scaling up) without having performed a root cause analysis - which maybe we should introduce in a different indicator?

@DaveCTurner what do you think?

Leaf-Lin:
It's true that the master size problem has improved a lot in 8.x. Let's ignore my suggestion and go ahead with the merge. 

Thank you!

PR #87691, opened 2022-06-15
[Tests] grant API key yml rest tests

n1v0lg:
This PR adds YAML REST tests for granting API keys. We cover this
functionality in Java REST tests, however YAML tests are nice from a
specification standpoint.


x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/10_grant.yml
```diff
@@ -0,0 +1,216 @@
+---
```
n1v0lg, line 1: I wasn't able to find any documentation around test file name prefixes: when do we use `10_` vs `11_` vs `20_` etc?

ywangd: There is likely a convention. But it was not clear to me either. My preferences/observations were:
1. Not reusing the same number
2. Choose the next number that is 10 apart, e.g. 10, 20, 30. So it leaves some space in case we need add 11, 21.
3. Numbering rougly corresponds when feature was added, e.g. 10 for the earliest.

If we use above "rules", I'd say `12` seems to fit.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/10_grant.yml
```diff
@@ -0,0 +1,216 @@
...
+  - match: { api_key.name: "my-api-key" }
+
+  - do:
+      security.clear_api_key_cache:
+        ids: "${api_key_id}"
+
+  - match: { _nodes.failed: 0 }
+
+---
+"Test grant api key forbidden":
```
ywangd, line 188: For this test, I'd suggest we have a 3rd user which has `manage_own_api_key` privilege and assert it cannot grant API keys (not even for itself)

ywangd (approved):
LGTM

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/10_grant.yml
tvernum: Are y'all too young to have programmed in BASIC?
https://en.wikipedia.org/wiki/Line_number#BASIC

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/10_grant.yml
n1v0lg: Yup, I can add that. I wanted to keep these relatively high level since we cover other corner cases in the Java REST tests already, but happy to make the addition 👍

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/10_grant.yml
n1v0lg: Lol does visual basic count?

PR #87692, opened 2022-06-15
Troubleshooting docs for ACTION_RESTORE_FROM_SNAPSHOT

gmarouli:
Troubleshooting guide to restore indices and data streams that have missing data from a snapshot.

This will be associated with the user action `ACTION_RESTORE_FROM_SNAPSHOT`.

Preview link: https://elasticsearch_87692.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/restore-from-snapshot.html


andreidan:
Preview link https://elasticsearch_87692.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/restore-from-snapshot.html

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,212 @@
+// tag::cloud[]
+In order to restore the indices and data streams that are missing data:
+
+**Use {kib}**
+
+//tag::kibana-api-ex[]
+. Log in to the {ess-console}[{ecloud} console].
++
```
dakrone, line 8: These pluses are all showing on the page, but I don't think they were intended. Should we removed them?

dakrone (commented):
I think we should have a pretty big caveat/warning/note that this potentially loses data (since it deletes indices and data streams that are red), and that the user should be aware of that before proceeding.

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
gmarouli: Definitely not intended. I will fix it.

gmarouli:
@dakrone I agree. I added this part:

> Given that you have found one or more snapshots to restore your indices you can proceed.

Now I am thinking we might need something stronger. I can also mention it at the beginning that this troubleshooting guide requires you to find a workable backup of the index. What do you think?

docs/reference/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,13 @@
+[[restore-from-snapshot]]
+== Restore from snapshot
+
+Elasticsearch is using snapshots to store a copy of your data outside a cluster. You can restore a snapshot to recover
+indices and data streams for which there are no copies of the shards in the cluster. This can happen after a deletion or
```
andreidan, line 5:
```suggestion
indices and data streams for which there are no copies of the shards in the cluster. This can happen if the data (indices or data streams) was deleted or
```

docs/reference/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,13 @@
...
+hardware failure.
```
andreidan, line 6:
```suggestion
if the cluster membership changed and the current nodes in the system do not contain a copy of the data anymore.
```

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,211 @@
...
+[role="screenshot"]
+image::images/troubleshooting/data/kibana-index-management-select-red-indices.png[{kib} indices,align="center"]
++
+[role="screenshot"]
+image::images/troubleshooting/data/kibana-index-management-select-red-data-steams.png[{kib} data streams,align="center"]
++
+. Use the left side menu to go to **Snapshot and Restore**, there you can verify that you have a snapshot to restore
+the target indices and data streams from. By clicking on every snapshot you can see the indices and data streams it
+contains. Not all resources need to be restored from the same snapshot. **Do not proceed** there is no workable
```
andreidan, line 34:
```suggestion
contains. Not all resources need to be restored from the same snapshot. **Do not proceed** if there is no workable
```

docs/reference/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,13 @@
...
+
+IMPORTANT: Restoring the missing data requires you to have a usable backup of the affected indices and data streams. Please do not
```
andreidan, line 8: What does usable mean? Should we clarify that or remove "usable" ?

andreidan, line 8:
```suggestion
IMPORTANT: Restoring the missing data requires you to have a backup of the affected indices and data streams. Please do not
```

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,211 @@
...
+<1> The name of the snapshot.
++
+<2> The repository of the snapshot.
++
+<3> The indices backed up in the snapshot.
++
+<4> If the snapshot was successful.
+
+. The snapshot `snapshot-20200617` contains the two indices we want to restore.
+You might have multiple snapshots from which you could restore the target indices.The explanation in our case indicates the index allocation configurations are not correct.
```
andreidan, line 144: What does "The explanation in our case indicates the index allocation configurations are not correct"  mean?
Would this refer to the explain API? What index configurations would this refer to? 

Should we remove this or maybe change it to be more precise?

andreidan, line 144:
```suggestion
You might have multiple snapshots from which you could restore the target indices.
```

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,211 @@
...
+Choose the latest one. Now that we found a snapshot, we will close the target indices via the
```
andreidan, line 145:
```suggestion
Choose the latest snapshot. Now that we found a snapshot, we will close the target indices via the <<indices-close, close indices API>>.
```

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,211 @@
...
+To review your allocation settings, use the <<indices-close, close indices API>>.
```
andreidan, line 146: A different API should be used here? 
And maybe moved below the example of using the close indices API?

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,211 @@
...
+  "include_aliases": true                                                       <2>
+}
+----
+// TEST[skip:illustration purposes only]
++
+<1> The indices to restore.
++
+<2> We also want to restore the aliases.
+
+. Finally we can verify that the indices are not `red` any more via the <<cat-indices,cat indices API>>.
```
andreidan, line 189:
```suggestion
. Finally we can verify that the indices health is now `green` via the <<cat-indices,cat indices API>>.
```

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
andreidan: I find this a bit confusing - what does "workable backup" mean here? An up-to-date enough backup or the index existing at all in the snapshot? 

Should we remove it or be more precise?

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,211 @@
...
+the target indices and data streams from. By clicking on every snapshot you can see the indices and data streams it
+contains. Not all resources need to be restored from the same snapshot. **Do not proceed** there is no workable
+backup of the affected resources in a snapshot, this will result in further data loss.
++
+[role="screenshot"]
+image::images/troubleshooting/data/kibana-restore-snapshot-content.png[{kib} snapshots,align="center"]
++
+. Go back to **Index Management** via the left side menu. In **Indices** enable **Include hidden indices** toggle and
+delete all the indices that have noted, by selecting them and then using the **Manage index** button
+choose delete index. In the **Data Streams** tab repeat the same by selecting the `red` data streams and then
```
andreidan, line 42: Would closing the index be enough? It seems that we're recommending closing them in the on-prem version

andreidan (commented):
Thanks for working on this Mary 

Left a few comments

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
gmarouli: Well it would be enough but the user experience is not as good. 

Let me elaborate, when a user restores a snapshot via the UI they cannot select a backing index, they can only select a data stream. This would mean that we have to explain to the user, that they need to go to the index, select all the backing indices of their data stream, close them, and then restore the data stream. I thought since we are trying to avoid confusing the user that we might want to stay away from backing indices etc. 

What do you think?

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
gmarouli: I think we should make it more precise, it is good to warn the user about permanent implications.

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
gmarouli: Actually, this: `To review your allocation settings, use` shouldn't be there

gmarouli: Fixed

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
andreidan: Ah I see, thanks for the details

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,212 @@
...
+The response will look like this:
++
+[source,console-result]
+----
+index                                status health
+.ds-my-data-stream-2022.06.17-000001 close   red
+kibana_sample_data_flights           close   red
+----
++
+. The indices are closed, now we can restore the, from snapshots without causing
```
cjcenizal, line 173: Possible typo: "now we can restore the," -> "now we can restore them"

cjcenizal (approved):
Found a couple typos but other than that, these docs LGTM. Thanks for creating them!

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,212 @@
...
++
+[role="screenshot"]
+image::images/troubleshooting/data/kibana-restore-snapshot-select-resources.png[{kib} select resources to restore,align="center"]
++
+. Click **Next** twice to continue without changing anything else and finally click **Restore Snapshot**.
++
+[role="screenshot"]
+image::images/troubleshooting/data/kibana-restore-snapshot-confirm.png[{kib} confirm snapshot restore,width=30%]
++
+. When the process will be complete you will see in the **Restore Status** tab the result of your restore.
```
Leaf-Lin, line 58:
```suggestion
. When the process is complete, you will see in the **Restore Status** tab the result of your restore.
```
The sentence reads better with the current tense and a comma.

Leaf-Lin (approved):
LGTM. 

Since we are having two widgets with `Elasticsearch Service` and `Self-managed`, the former is completely fine but could also be a self-managed cluster with Kibana enabled, while the latter feels more like the API guide.

Is it worth to reconsider the name, maybe `Kibana guide` vs `API guide` or something along these lines?


Also, in the API guide, we instruct users to `close` old indices while in the UI guide we instruct users to `delete` old indices. Is it worth making them consistent? I prefer to close than delete as it's less destructive.

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,212 @@
...
+For more guidance on creating and restoring snapshots see
+<<snapshot-restore, this guide>>.
+
+//end::kibana-api-ex[]
+// end::cloud[]
+
+// tag::self-managed[]
+In order to restore the indices that are missing shards:
+
+. View the affected indices shards using the <<cat-indices,cat indices API>>.
```
Leaf-Lin, line 72:
```suggestion
. View the affected indices using the <<cat-indices,cat indices API>>.
```
Do you mean to include the word `shards` here?

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,212 @@
...
+The `red` health of the indices above indicates that these indices are missing primary shards,
+meaning they are missing data.
++
+. In order to restore the data we need to find a snapshot that contains these two indices.To find
+such a snapshot use the
+<<get-snapshot-api,get snapshot API>>.
++
+[source,console]
+----
+GET _snapshot/snap-repo/*?verbose=false
```
Leaf-Lin, line 98:
```suggestion
GET _snapshot/my_repository/*?verbose=false
```
Keep it consistent with the restore-from-snapshot.html doc, also without the word `my`, it's hard for users to distinguish variables from ES API commands.

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,212 @@
...
+The response will look like this:
++
+[source,console-result]
+----
+{
+  "snapshots" : [
+    {
+      "snapshot" : "snapshot-20200617",                                     <1>
+      "uuid" : "dZyPs1HyTwS-cnKdH08EPg",
+      "repository" : "snap-repo",                                           <2>
```
Leaf-Lin, line 111:
```suggestion
      "repository" : "my_repository",                                           <2>
```

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,212 @@
...
+      "indices" : [                                                         <3>
+        ".apm-agent-configuration",
+        ".apm-custom-link",
+        ".ds-.logs-deprecation.elasticsearch-default-2022.06.17-000001",
```
Leaf-Lin, line 115:
```suggestion
```
Remove this line so that the console in the doc will not run over its limited width making users scroll.

andreidan:
@Leaf-Lin 
Regarding `close` vs `delete` https://github.com/elastic/elasticsearch/pull/87692#discussion_r901476968

Also, regarding 
> Is it worth to reconsider the name, maybe Kibana guide vs API guide or something along these lines?

Cloud is moving away from mentioning Kibana. IMO we should keep the current terminology.

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
gmarouli: This sentence was not intended, some weird mistake on my end, I removed it.

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,212 @@
...
+index                                status health
+.ds-my-data-stream-2022.06.17-000001 open   red
+kibana_sample_data_flights           open   red
+----
+// TEST[skip:illustration purposes only]
++
+The `red` health of the indices above indicates that these indices are missing primary shards,
+meaning they are missing data.
++
+. In order to restore the data we need to find a snapshot that contains these two indices.To find
```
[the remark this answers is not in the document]
gmarouli: Thank you @cjcenizal for reporting them!

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
gmarouli: Fixed

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
gmarouli: Good point, I will fix all occurrences.

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,212 @@
...
+.ds-my-data-stream-2022.06.17-000001 close   red
+kibana_sample_data_flights           close   red
+----
++
+. The indices are closed, now we can restore the, from snapshots without causing
+any complications using the <<restore-snapshot-api, restore snapshot API>>:
++
+[source,console]
+----
+POST _snapshot/snap-repo/snapshot-20200617/_restore
```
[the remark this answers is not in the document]
gmarouli: I see you found them all :)

docs/reference/troubleshooting/data/restore-from-snapshot.asciidoc
gmarouli: Replaced the word `usable` in a way that is consistent with the rest of the document.

andreidan:
> @andreidan why do you think we should keep the existing naming?

I think we should keep the current naming.

> There is not difference between self-managed and cloud

I don't agree with this dichotomy. I think most guides have differences and even more so, the differences are ESS vs self-managed (the term `cloud` is too broad IMO) eg. we talk about adding capacity instead of nodes, enabling tiers instead of tagging nodes with tier specific node roles

andreidan:
> I was talking about this case only.

Oh, not sure if that would help or be more confusing amongst all the other guides that have this separation? I'll let others chime in on this UX aspect (my view is that consistency would benefit our users)

gmarouli:
@Leaf-Lin should we leave it as is for consistency then? It's hard for me to get in the shoes of an average user in situations like this.

Leaf-Lin:
IMO, I felt that for this particular page, we should show the tabs `UI` and `API` because that's what they are.

For consistency, we can add a paragraph at the top of the documentation along the lines of:
> Users from either `Elasticsearch Services` or `Self-managed` platform are able to use either `UI` or `API` steps to restore snapshot. Choose your preferred method to start. 

And at the UI page, we can add something like, if you are self-managed user, you can go straight to Kibana and start from step 3.

gmarouli:
@Leaf-Lin I incorporated your suggestions to the latest version of this doc.

gmarouli:
@elasticmachine update branch

gmarouli:
Changed UI to Kibana after an offline discussion and to be consistent with other docs.

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,213 @@
...
+<4> If the snapshot was successful.
+
+. The snapshot `snapshot-20200617` contains the two indices we want to restore.
+You might have multiple snapshots from which you could restore the target indices. Choose the latest snapshot.
++
+. Now that we found a snapshot, we will close the target indices via the use the <<indices-close, close indices API>>.
++
+[source,console]
+----
+POST kibana_sample_data_flights,.ds-my-data-stream-2022.06.17-000001/_close
```
andreidan, line 152: For the last generation backing index of a data stream, in case the data stream has more than one generation, this call will fail. 
If the last generation of a data stream needs to be restored I think we should recommend rolling over the data stream and closing the (now) N-1 generation index.

andreidan (commented):
Thanks for iterating on this @gmarouli 

Left a comment. 

I still have concerns over recommeding deleting data in the Kibana version as opposed to closing resources. 
I'd be in favour of using Dev Tools in Kibana and recommending closing the data stream backing indices (like we do in the API version) as opposed to deleting (a lot of) data.

@dakrone How would you feel ^^ ? (the data stream use case is particularly concerning - deleting ALL generations just so we can restore one generation)

dakrone:
I also think it'd be better to close any existing indices instead of deleting them.

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
gmarouli: Great catch @andreidan , I will adjust this.

gmarouli:
@dakrone & @andreidan +1 from me too. I will bring it up at our next Health API catch up so we can make a decision and finalize this.

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
gmarouli: Hm, I tried it and it didn't fail, this is what I tried:

1. Created a data stream
2. Indexed 3 documents
3. Manually rolled over
4. Index one more document, confirmed that it got indexed in the last generation
5. Closed the last generation
6. Tried to index another document and I got an `index_closed_exception`.

This is the scenario that you expected to fail, right?

That said, I do think it's a nice idea to rollover because that will ensure smooth indexing even while restoring a snapshot. My only concern is that I do not want to confuse the reader too much with data stream terminology (especially if we provide this as the only troubleshooting scenario for cloud and self managed).

What do you think if we say that if the index name has the pattern `.ds-my-data-stream-xxxxx` the user can assume it's a data stream and then always rollover first. This way they do not have to check if it's the latest one. Rolling over prematurely is not the best but I think since this is an "emergency" scenario it would pay off to cut that corner. 

WDYT @andreidan & @dakrone ?

gmarouli: Done: https://github.com/elastic/elasticsearch/issues/88837

gmarouli:
@elasticmachine update branch

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,456 @@
...
++
+The response above shows that `kibana_sample_data_flights` is not part of a data stream because it doesn't have a
+field called `data_stream` in the settings.
++
+On the contrary, `.ds-my-data-stream-2022.06.17-000001` is part of the data stream called `my-data-stream`. When you
+find an index like this, which belongs to a data stream, you need to check if data are still being indexed.
+You can see that by checking the `settings`, if you can find this property: `"index.lifecycle.indexing_complete" : "true"`,
+it means that indexing is completed in this index and you can continue to the next step.
++
+If it is not there you need to rollover the data stream so you can restore the missing data without blocking
```
andreidan, line 381:
```suggestion
If `index.lifecycle.indexing_complete` is not there or is configured to `false` you need to rollover the data stream so you can restore the missing data without blocking
```

docs/reference/tab-widgets/troubleshooting/data/restore-from-snapshot.asciidoc
```diff
@@ -0,0 +1,456 @@
...
```
andreidan, line 161:
```suggestion
```suggestion
If `index.lifecycle.indexing_complete` is not there or is configured to `false` you need to rollover the data stream so you can restore the missing data without blocking
```
```

andreidan (approved):
LGTM 🙌 Thanks Mary !

abdonpijpelink:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.3|[](https://github.com/elastic/elasticsearch/pull/89243)|

### Questions ?

PR #84631, opened 2022-03-03
Integrate ES with APM

pugnascotia:
This PR merges in the proof-of-concept work for adding tracing to Elasticsearch. It uses the OpenTelemetry library, with the expectation that the traces will be sent to Elastic's APM integration.

Much of the APM-specific code is in the `apm-integration` plugin. Most of the rest of the changes are concerned with expression some activity in ES more clearly in terms of tasks, since this work maps tasks to spans.

Issue #84369, opened 2022-02-24
Instrument Elasticsearch with APM


pugnascotia:
**NOTE**: this issue will evolve as we scope out this work.

## Description

"Why is Elasticsearch slow?" is a common question from users. We have tools to investigate certain aspects of this question already, for instance the search slowlog (good if the shard-level searches are slow) and the hot threads API (good if the slowness is an ongoing thing) but there are many gaps too. For instance, how would we discover that a Kibana dashboard triggers unreasonably many searches if each of those searches completes fairly quickly? How would we discover that requests are spending unexpectedly long in queues? How do we see if the slow steps all involve a particular node? What if that node is on a remote cluster? It's hard to take a structured approach to performance questions with the tools we have today.

Distributed tracing is a great way to answer questions of this nature. Elastic has a distributed tracing product, [APM](https://www.elastic.co/apm/), which sits on top of Elasticsearch, but today Elasticsearch itself is opaque to APM: we cannot trace the execution of a request through Elasticsearch. Let's fix that.

This work will build on an existing exploratory project that instrumented a number of "tasks" in Elasticsearch. More types of tasks will be instrumented, as well as requests / responses at the REST level.

## Tasks

   * Technical integration
      * [x] Update the proof-of-concept branch of APM in ES
      * [x] Engage with the APM team to check what is the currently recommended way of integrating a Java application with APM
      * [x] Benchmark the overhead of running with APM disabled, to determine if we can merge safely
      * [x] Ensure that distributed tracing works with other Elastic stack components
      * [x] Add testing that uses the APM agent
      * [x] Make sure using APM agent gives the same results
         * [x] https://github.com/elastic/apm-server/issues/8067
         * [x] https://github.com/elastic/apm-agent-java/pull/2657
      * Code changes
         * [x] #87917
         * [x] #87921 
         * [x] #88443
         * [ ] Add docs for tracing ⚠️ Need to check whether we want to document APM at this time ⚠️ 
         * [ ] Developer education / awareness
         * [x] Raise Cloud issue for wiring up APM
   * Documentation
      * [x] Document how to extend tracing in Elasticsearch e.g. coding conventions, what attributes to apply, how to model spans.
   * Span / transaction modelling
      * [x] Check whether the way we are capturing spans in Elasticsearch works with what Kibana etc are doing. The end-to-end picture needs be coherent.
   * Configuration:
      * [x] Provide richer options for filtering captured spans
[...]

PR #87696, opened 2022-06-15
Integrate ES with APM

pugnascotia:
Supersedes #84631. Part of #84369.

This PR takes the proof-of-concept work for adding tracing to Elasticsearch and carries it to an initial completion. It adds an `apm-integration` X-Pack plugin, which uses the OpenTelemetry API to create spans. Also if APM is enabled, Elasticsearch is started using the APM java agent as a Java agent. This is responsible for ships spans and other metrics to an APM server.

There are also a number of changes that are concerned with expressing some activity in ES more clearly in terms of tasks, since this work maps tasks to spans.

### Notes to reviewers

Start with the file `TRACING.md`, it provides useful context for this work. If you want to see it in action, I can point you at an existing APM server, or give you a `docker-compose.yml` which will let you run your own stack with distributed tracing.

I have run several benchmarks with APM disabled, and could see nothing consistently slower. I still need to run some benchmarks with APM enabled, though my hope is that this isn't a blocker to merging, and that we can iterate post-merge if there are hot areas.

I would particularly like feedback about the config mechanism.

Todo:

   * [ ] Run benchmarks with APM enabled
   * [ ] Docs


build-tools/src/main/java/org/elasticsearch/gradle/testclusters/ElasticsearchNode.java
```diff
@@ -1356,19 +1359,12 @@ private void createConfiguration() {
...
-                configFiles = stream.collect(Collectors.toList());
-            }
-            logToProcessStdout("Copying additional config files from distro " + configFiles);
-            for (Path file : configFiles) {
-                Path dest = configFile.getParent().resolve(file.getFileName());
-                if (Files.exists(dest) == false) {
-                    Files.copy(file, dest);
-                }
-            }
+            final Path distConfigDir = getDistroDir().resolve("config");
```
pugnascotia, line 1362: These changes were ultimately unnecessary, but I've kept them because the make it possible to retain the config file hierarchy when starting a test cluster.

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/BootstrapJvmOptions.java
```diff
@@ -42,7 +42,7 @@ public static List<String> bootstrapJvmOptions(Path plugins) throws IOException
     private static List<PluginInfo> getPluginInfo(Path plugins) throws IOException {
         final List<PluginInfo> pluginInfo = new ArrayList<>();
 
-        final List<Path> pluginDirs = Files.list(plugins).collect(Collectors.toList());
+        final List<Path> pluginDirs = listFiles(plugins);
```
pugnascotia, line 45: These changes just ensure that the `Stream` from `Files.list(Path)` is closed.

rjernst (commented):
This PR is quite large. Can it be split out into a few PRs? For example, working through the Tracing interface and initialization before anything actually uses it, then adding the task/etc uses of tracer.

Some overall thoughts after a skim of the code:
* The tracer looks very similar to logging. I wonder if we could configure it statically like logging, so that we could have eg TraceManager.getTracer(...). This would seem closer to the openTelemetry object in the open telemetry api, ie the thing to get a tracer from.
* The tracer in open telemetry just has a single method to get a span. Why not replicate this API?

henningandersen (commented):
++ on splitting into multiple PRs, otherwise getting the right folks to review the individual parts will be difficult. I'd think all the libs and the Tracer interface itself could go in one PR and then each usage more or less in its own PR.

pugnascotia:
I'll work on splitting this up. It grew over time.

> The tracer looks very similar to logging. I wonder if we could configure it statically like logging, so that we could have eg TraceManager.getTracer(...). This would seem closer to the openTelemetry object in the open telemetry api, ie the thing to get a tracer from.

@rjernst at the moment we can't configure it statically, because the tracer implementation lives in a module. We could move this into `server`, if you'd be OK with that? Thinking about it, this is now more practical with the modularization work, since we can avoid exporting anything that we want to keep private.

> The tracer in open telemetry just has a single method to get a span. Why not replicate this API?

I'm not sure I follow, can you elaborate. A better comparator would be the `SpanBuilder` in the OTel API, since having started building a span, you need to make various calls on the builder before finally getting a `Span` out of it.

rjernst:
> the tracer implementation lives in a module. We could move this into server, if you'd be OK with that?

I could see a small part of this being in server.

> > The tracer in open telemetry just has a single method to get a span. Why not replicate this API?

> I'm not sure I follow, can you elaborate. A better comparator would be the SpanBuilder in the OTel API

I was looking at the [Tracer class](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/Tracer.java) in the opentelemetry java sdk. That class has a single method to return a SpanBuilder. It looks stuff from their SpanBuilder is much of what is on the Tracer in this PR. Why the collapse? ie why is there no SpanBuilder here?

Something else that I could not understand from my brief reading of this PR is what thread safety or multi threaded expectations exist in in this new api (mostly on the Tracer class, which appears to be a single instance for the node).

TRACING.md
```diff
@@ -0,0 +1,141 @@
...
+set in the config file cannot be overridden via system properties.
+
+Now, in order to send tracing data to the APM server, ES needs to configured with
+either a `secret_key` or an `api_key`. We could configure these in the agent via
+system properties, but then their values would be available to any Java code
+that can read system properties.
+
+Instead, when Elasticsearch bootstraps itself, it compiles all APM settings
+together, including any `secret_key` or `api_key` values from the ES keystore,
+and writes out a temporary APM config file containin all static configuration
```
breskeby, line 49:
```suggestion
and writes out a temporary APM config file containing all static configuration
```

pugnascotia:
This PR has been broken up into smaller PRs.

PR #87697, opened 2022-06-15
[ML] Require that threads_per_allocation is a power of 2

dimitris-athanasiou:
As the number of cores in CPUs is typically a power of 2,
this commit adds a validation that trained model deployments
start with `threads_per_allocation` set to be a power of 2.
When we look for how we distribute the allocations across the
cluster, this prevents situations where we have a lot of wasted
CPU cores.

### DIFF
--- a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/action/StartTrainedModelDeploymentAction.java
+++ b/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/action/StartTrainedModelDeploymentAction.java
@@ -66,6 +66,9 @@ public class StartTrainedModelDeploymentAction extends ActionType<CreateTrainedM
             AllocationStatus.State.STARTED,
             AllocationStatus.State.STARTING,
             AllocationStatus.State.FULLY_ALLOCATED };
+
+        private static final int MAX_THREADS_PER_ALLOCATION = 32;
+
         public static final ParseField MODEL_ID = new ParseField("model_id");
         public static final ParseField TIMEOUT = new ParseField("timeout");
         public static final ParseField WAIT_FOR = new ParseField("wait_for");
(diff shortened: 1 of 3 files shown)


x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/ml/action/StartTrainedModelDeploymentRequestTests.java
```diff
@@ -80,6 +84,21 @@ public void testValidate_GivenThreadsPerAllocationIsNegative() {
...
+            Request request = createRandom();
+            request.setThreadsPerAllocation(n);
+
+            ActionRequestValidationException e = request.validate();
+
+            assertThat(e, is(not(nullValue())));
+            assertThat(e.getMessage(), containsString("[threads_per_allocation] must be a power of 2"));
+        }
+    }
+
```
benwtrent, line 101: Could you add a test that shows that it does NOT throw if it is a power of two?

benwtrent (approved):

droberts195 (commented):
I also wonder if we should limit the max to 32 for the time being.  Currently 32 is the highest number of vCPUs you can get on the types of ML instances used in ESS.  32 is likely good enough for self managed scenarios for the time being too.  As hardware gets more powerful in the future we can double this limit in future versions.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/action/StartTrainedModelDeploymentAction.java
```diff
@@ -209,12 +209,23 @@ public ActionRequestValidationException validate() {
...
             }
             return validationException.validationErrors().isEmpty() ? null : validationException;
         }
 
+        private static boolean isPowerOf2(int value) {
+            if (value == 0) {
+                return false;
+            }
+            double log2OfValue = Math.log(value) / Math.log(2);
+            return Math.floor(log2OfValue) == Math.ceil(log2OfValue);
```
droberts195, line 226:
```suggestion
            return Integer.bitCount(value) == 1;
```

Avoids the need for conversion to floating point and expensive logarithm calculations.

dimitris-athanasiou: Clever!

dimitris-athanasiou:
@droberts195 I like the idea of also capping this at `32`. Will add it to this PR.

dimitris-athanasiou:
@droberts195 and @benwtrent I have updated the PR, could you take another look please?

benwtrent (approved):

droberts195 (approved):
LGTM

PR #87701, opened 2022-06-15
Fix TSDS example setup

kilfoyle:
This should hopefully fix the break caused by #86905


kilfoyle:
Closing. We reverted the problem PR instead.

PR #87704, opened 2022-06-15
Add ability to select execution mode for cardinality aggregation

not-napoleon:
Turns out, our tests don't cover the segment ordinals execution path for Cardinality.  This PR adds an execution hint, like Terms uses, and uses that hint to force some tests onto the segment ordinals path.


not-napoleon:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/search/aggregations/bucket/terms/TermsAggregatorFactory.java
```diff
@@ -111,7 +111,7 @@ public Aggregator build(
                     execution = ExecutionMode.fromString(executionHint);
                 }
                 // In some cases, using ordinals is just not supported: override it
-                if (valuesSource instanceof ValuesSource.Bytes.WithOrdinals == false) {
+                if (valuesSource.hasOrdinals() == false) {
```
not-napoleon, line 114: This is unrelated, but I saw a chance to replace an `instanceof` with a capability check, and I took it.

nik9000 (commented):
Seems right to me. Probably worth adding something to `CardinalityTests` to cover the new case. Maybe renaming that file so it's easier to find too.....

not-napoleon:
@elasticmachine update branch

not-napoleon:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/search/aggregations/metrics/CardinalityAggregator.java
```diff
@@ -61,6 +62,7 @@ public CardinalityAggregator(
         String name,
         ValuesSourceConfig valuesSourceConfig,
         int precision,
+        CardinalityAggregatorFactory.ExecutionMode executionMode,
```
nik9000, line 65: Should this be called `exeuctionHint`? Because `null` means "you pick?"

nik9000: Ah. I see you expect a refactor soon. Well, may as well wait on the rename until then anyway.

server/src/test/java/org/elasticsearch/search/aggregations/metrics/CardinalityTests.java
```diff
@@ -20,6 +22,9 @@ protected final CardinalityAggregationBuilder createTestAggregatorBuilder() {
         if (randomBoolean()) {
             factory.missing("MISSING");
         }
+        if (true) {
```
nik9000, line 25: Probably should be `if (randomBoolean)`

server/src/main/java/org/elasticsearch/search/aggregations/metrics/CardinalityAggregator.java
not-napoleon: I was considering an explicit "heuristic" mode instead of relying on the null.  And I'm not sure when I'm going to get to that refactor.  WDYT?

server/src/test/java/org/elasticsearch/search/aggregations/metrics/CardinalityTests.java
not-napoleon: oops, yeah.  I forced it for testing, and forgot to unforce it.

server/src/main/java/org/elasticsearch/search/aggregations/metrics/CardinalityAggregator.java
nik9000: I'd prefer a `HEURISTIC` mode over null, yeah.

docs/reference/aggregations/metrics/cardinality-aggregation.asciidoc
```diff
@@ -237,3 +237,24 @@ POST /sales/_search?size=0
 --------------------------------------------------
 // TEST[setup:sales]
 <1> Documents without a value in the `tag` field will fall into the same bucket as documents that have the value `N/A`.
+==== Execution Hint
+
+There are different mechanisms by which cardinality aggregations can be executed:
+
+ - by using field values directly (`direct`)
+ - by using global ordinals of the field and resolving those values after
+   finishing a shard(`global_ordinals`)
```
nik9000, line 246: missing space.

nik9000 (approved):

docs/reference/aggregations/metrics/cardinality-aggregation.asciidoc
not-napoleon:
```suggestion
   finishing a shard (`global_ordinals`)
```

PR #87709, opened 2022-06-15
[ML] performance improvement for precompiled normalization

benwtrent:
This is a performance improvement for the precompiled normalizer. It no longer requires the graphemes to be sub-strings and relies on code-point counts between grapheme boundaries for normalization.


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/PrecompiledCharMapNormalizer.java
```diff
@@ -144,7 +144,7 @@ Optional<BytesRef> normalizePart(byte[] strBytes, int offset, int len) {
             secondIndex++;
         }
         if (secondIndex == firstIndex) {
-            return Optional.empty();
+            return Optional.of(new BytesRef(BytesRef.EMPTY_BYTES));
```
benwtrent, line 147: This was a bug, we should return empty bytes here as when the two indices are equal to eachother, it indicates that this part of the string should be removed (think control characters...)

benwtrent:
@elasticmachine update branch

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/PrecompiledCharMapNormalizer.java
```diff
@@ -155,40 +155,43 @@ String normalize(String str) {
...
+        int bytePos = 0;
         for (int end = b.next(); end != BreakIterator.DONE; start = end, end = b.next()) {
-            // TODO: It would be awesome if we could translate these starts and ends to byte positions, if we could performance would be
-            // dramatically improved
-            String unicodeStr = str.substring(start, end);
-            byte[] unicode = unicodeStr.getBytes(StandardCharsets.UTF_8);
+            int byteLen = 0;
+            int numCp = str.codePointCount(start, end);
+            for (int i = start; i < numCp + start; i++) {
+                byteLen += numUtf8Bytes(strCp[i]);
```
droberts195, line 171: I think there's a problem here.  It's assuming that `start` equals the number of code points that have been processed previously.  But actually `start` stores the number of UTF-16 characters that have been processed previously.  They might be the same, but they might not.

I think in this line `start` needs to be `str.codePointCount(0, start)`.  But it's probably more efficient to add an extra variable that keeps a running count of that, similar to what `bytePos` does for the bytes.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/PrecompiledCharMapNormalizer.java
```diff
@@ -155,40 +155,43 @@ String normalize(String str) {
...
                 if (subStr.isPresent()) {
                     strBuilder.append(subStr.get());
                 } else {
-                    int numBytes = UnicodeUtil.UTF16toUTF8(unicodeCharArray, charIndex, 1, reusableCharByteBuffer);
+                    int numBytes = UnicodeUtil.UTF16toUTF8(strChars, i, 1, reusableCharByteBuffer);
                     strBuilder.append(reusableCharByteBuffer, 0, numBytes);
                 }
-                charByteIndex += numUtf8Bytes(c);
-                ++charIndex;
+                charByteIndex += numUtf8Bytes(strChars[i]);
```
droberts195, line 192: This is the second time `numUtf8Bytes(strChars[i])` is calculated in this loop, so it would make the code easier for the compiler to optimize if the result was stored in a local variable as the first statement inside the loop.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/PrecompiledCharMapNormalizer.java
```diff
@@ -155,40 +155,47 @@ String normalize(String str) {
...
-            byte[] unicode = unicodeStr.getBytes(StandardCharsets.UTF_8);
+        strBuilder.grow(strBytes.length);
+        int bytePos = 0;
+        for (int end = b.next(); end != BreakIterator.DONE; startIter = end, end = b.next()) {
+            int byteLen = 0;
+            int numCp = str.codePointCount(startIter, end);
+            for (int i = codePointPos; i < numCp + codePointPos; i++) {
+                byteLen += numUtf8Bytes(strCp[i]);
+            }
+            codePointPos += numCp;
```
benwtrent, line 175: @droberts195 we always start at `0`, but here we only iterate codepoints based on the previously counted code points. So it should satisfy your concerns.

droberts195 (approved):
LGTM

PR #87711, opened 2022-06-15
[Transforms] fix bug when unsetting retention policy

### LINKED ISSUE #87688: [Transform] Error when removing retention policy
**Found in version**
- Observed in 8.2.3, but might exist longer

**Steps to reproduce**
On a multi node cloud cluster (might also work on a local multi node cluster):
- Create a transform with a retention policy
- Update the transform and remove the retention policy

**Expected result**
- The retention policy is removed

**Actual result**
- This sometimes results in an error:
```
An error occurred calling the API endpoint to update transforms.
Bad Request: [illegal_argument_exception: [illegal_argument_exception] Reason: 
Unknown NamedWriteable [org.elasticsearch.xpack.core.transform.transforms.RetentionPolicyConfig][null_retention_policy]]:
Unknown NamedWriteable [org.elasticsearch.xpack.core.transform.transforms.RetentionPolicyConfig][null_retention_policy]
```

**Additional information**
- Quote from slack discussion with @benwtrent:
  > It wouldn’t happen if you hit a node that can execute the update immediately without serializing to a different node

benwtrent:
It is currently not possible to unset the retention policy in a typical multi-node cluster. 

This commit fixes that bug. If unsetting the retention policy failed, it would fail looking like the following error:

```
An error occurred calling the API endpoint to update transforms.
Bad Request: [illegal_argument_exception: [illegal_argument_exception] Reason: 
Unknown NamedWriteable [org.elasticsearch.xpack.core.transform.transforms.RetentionPolicyConfig][null_retention_policy]]:
Unknown NamedWriteable [org.elasticsearch.xpack.core.transform.transforms.RetentionPolicyConfig][null_retention_policy]
```

closes: https://github.com/elastic/elasticsearch/issues/87688

### DIFF
--- a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/XPackClientPlugin.java
+++ b/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/XPackClientPlugin.java
@@ -212,6 +212,7 @@ import org.elasticsearch.xpack.core.transform.action.PreviewTransformAction;
 import org.elasticsearch.xpack.core.transform.action.PutTransformAction;
 import org.elasticsearch.xpack.core.transform.action.StartTransformAction;
 import org.elasticsearch.xpack.core.transform.action.StopTransformAction;
+import org.elasticsearch.xpack.core.transform.transforms.NullRetentionPolicyConfig;
 import org.elasticsearch.xpack.core.transform.transforms.RetentionPolicyConfig;
 import org.elasticsearch.xpack.core.transform.transforms.SyncConfig;
 import org.elasticsearch.xpack.core.transform.transforms.TimeRetentionPolicyConfig;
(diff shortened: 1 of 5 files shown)


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/XPackClientPlugin.java
```diff
@@ -534,6 +535,11 @@ public List<NamedWriteableRegistry.Entry> getNamedWriteables() {
                     TransformField.TIME.getPreferredName(),
                     TimeRetentionPolicyConfig::new
                 ),
+                new NamedWriteableRegistry.Entry(
```
davidkyle, line 538: `TransformNamedXContentProvider` is also missing the `NullRetentionPolicyConfig` named writable

benwtrent: @davidkyle it isn't required. This thing is never written to XContent. But I can add it for completeness.

davidkyle: Please do I think it is worth having to avoid a similar problem in future

davidkyle (approved):
LGTM

PR #87712, opened 2022-06-15
Add indicator service details to health API documentation.

jbaiera:
This PR adds listings of all the current details that can be returned from the implemented Health Indicator Services. Response details are unique to each indicator and describe the state of the system that the indicator is basing its health decisions on.

### Preview
https://elasticsearch_87712.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/health-api.html

### DIFF
--- a/docs/reference/health/health.asciidoc
+++ b/docs/reference/health/health.asciidoc
@@ -285,7 +285,7 @@ details have contents and a structure that is unique to each indicator.
 
 `recent_masters`::
     (Optional, array) A list of nodes that have been elected or replaced as master in a recent
-    time window (based on `master_history.max_age`). This field is present if the master
+    time window. This field is present if the master
     is changing rapidly enough to cause problems, and also present as additional information
     when the indicator is `green`.
 +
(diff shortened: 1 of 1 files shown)


docs/reference/health/health.asciidoc
```diff
@@ -260,6 +260,123 @@ for health status set `explain` to `false` to disable the more expensive analysi
...
+`node_id`::
+    (string) The node id of the currently elected master, or null if no master is elected.
+
+`name`::
+    (string) The node name of the currently elected master, or null if no master is elected.
+====
+
+`recent_masters`::
+    (Optional, array) A list of nodes that have been elected or replaced as master in a recent
+    time window (based on `master_history.max_age`). This field is present if the master
```
andreidan, line 288:
```suggestion
    time window. This field is present if the master
```

I don't think we documented `master_history.max_age` yet. I think we should do that separately? And then mention it back here

andreidan (approved):
LGTM, much ❤️ for adding this Jimmy

docs/reference/health/health.asciidoc
```diff
@@ -260,6 +260,123 @@ for health status set `explain` to `false` to disable the more expensive analysi
...
+==== Indicator Details
+
+Each health indicator in the health api returns a set of details that further explains the state of the system. The
+details have contents and a structure that is unique to each indicator.
+
+[[health-api-response-details-master-is-stable]]
+===== master_is_stable
+
+`current_master`::
+    (object) Information about the currently elected master.
```
masseyke, line 274: This is also optional

PR #87723, opened 2022-06-16
Optimize log cluster health performance.

howardhuanghua:
Cluster health would be checked in each allocation task, since construct cluster state health would iterate all the shards, `logClusterHealthStateChange` is costly if cluster has huge number of shards.

```
70.4% (352ms out of 500ms) cpu usage by thread 'elasticsearch[1592969635000032535][transport_worker][T[#32](https://git.woa.com/ces/elasticsearch/issues/32)]'
4/10 snapshots sharing following 62 elements
java.util.Collections$UnmodifiableCollection$1.(Collections.java:1041)
java.util.Collections$UnmodifiableCollection.iterator(Collections.java:1040)
java.util.Collections$UnmodifiableCollection$1.(Collections.java:1041)
java.util.Collections$UnmodifiableCollection.iterator(Collections.java:1040)
org.elasticsearch.cluster.routing.IndexShardRoutingTable.iterator(IndexShardRoutingTable.java:155)
org.elasticsearch.cluster.health.ClusterShardHealth.(ClusterShardHealth.java:90)
org.elasticsearch.cluster.health.ClusterIndexHealth.(ClusterIndexHealth.java:121)
org.elasticsearch.cluster.health.ClusterStateHealth.(ClusterStateHealth.java:78)
```

With this commit, we could skip redundant shards iteration. Quickly check cluster status by:
1. No unassigned and inactive shards, return `GREEN` directly.
2. No unassigned and inactive primary shards, return `YELLOW` directly.
3. Iterate index, if we found one RED one then return `RED` directly.


server/src/main/java/org/elasticsearch/cluster/health/ClusterStateHealth.java
```diff
@@ -144,6 +151,44 @@ public ClusterStateHealth(
...
+            return ClusterHealthStatus.GREEN;
+        }
+        if (routingNodes.hasUnassignedPrimaries() == false && routingNodes.hasInactivePrimaries() == false) {
+            // no unassigned, inactive primaries, but has unassigned or inactive replicas
+            return ClusterHealthStatus.YELLOW;
+        }
+
+        // cluster at least YELLOW
+        computeStatus = ClusterHealthStatus.YELLOW;
+        for (IndexRoutingTable indexRoutingTable : routingTable) {
```
original-brownbear, line 171: Practically, we might end up in this loop quite a bit won't we?  I wonder if it's worth it computing `RoutingNodes` for one of the states (the other already has them) here if that's the case.

Maybe we could do a smaller change here and do this PR without the short-cuts through `RoutingNodes` and take the win for not having to create the `ClusterIndexHealth` etc. instances and map of them here for now and do without the routing nodes for a straight-forward speedup in all cases?
So concretely, always do this loop but start with green health and do the same loop logically that we do in `org.elasticsearch.cluster.health.ClusterStateHealth#ClusterStateHealth(org.elasticsearch.cluster.ClusterState, java.lang.String[])`.

To me it it seems like if we wanted to use `RoutingNodes` here (again, I removed its used a short while ago) we'd just want to compute the index health when building the routing nodes directly and make this method free.
WDYT?

original-brownbear (commented):
Thanks @howardhuanghua this is a really nice idea + speedup.
See my comments inline: I'm a little hesitant about using `RoutingNodes` here and suggested a smaller change for a first step. Let me know what you think.

server/src/main/java/org/elasticsearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -217,6 +221,10 @@ public boolean allPrimaryShardsActive() {
         return primaryShardsActive() == shards.length;
     }
 
+    public boolean allShardsActive() {
+        return totalShardCount == allActiveShards.size();
```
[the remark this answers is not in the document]
howardhuanghua: Yes, we could cache boolean directly.

server/src/main/java/org/elasticsearch/cluster/health/ClusterStateHealth.java
```diff
@@ -144,6 +151,44 @@ public ClusterStateHealth(
         this.indices = indices;
     }
 
+    public static ClusterHealthStatus getHealthStatus(
+        final RoutingTable routingTable,
+        final RoutingNodes routingNodes,
+        final Logger logger
+    ) {
+        ClusterHealthStatus computeStatus;
```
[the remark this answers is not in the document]
howardhuanghua: I added cluster block check before looping the index shard routing table.

server/src/main/java/org/elasticsearch/cluster/health/ClusterStateHealth.java
howardhuanghua: I thought `RoutingNodes` has been constructed before. I am not sure I got all the parts of you suggestion, I have updated the PR to show the code, please help to check again, thank you.

server/src/main/java/org/elasticsearch/cluster/health/ClusterStateHealth.java
```diff
@@ -144,6 +149,40 @@ public ClusterStateHealth(
         this.indices = indices;
     }
 
+    public static ClusterHealthStatus getHealthStatus(
```
original-brownbear, line 152: We could move this to `AllocationService` couldn't we? Then we won't have to pass the logger around.

original-brownbear (commented):
Thanks @howardhuanghua looks really nice now! Just one question/request left.

server/src/main/java/org/elasticsearch/cluster/health/ClusterStateHealth.java
original-brownbear: Thanks! Sorry if I was unclear here, turned out exactly as I had hoped though :)
We don't always have the `RoutingNodes` here for the new cluster state. We probably will have to build them later anyway, but it's a lot easier for me to reason about the performance of this if we use the routing table :)

original-brownbear:
Jenkins test this

original-brownbear:
Jenkins test this

server/src/main/java/org/elasticsearch/cluster/routing/allocation/AllocationService.java
```diff
@@ -513,6 +513,37 @@ private static void logClusterHealthStateChange(
...
 
+    public static ClusterHealthStatus getHealthStatus(final ClusterState clusterState) {
+        if (clusterState.blocks().hasGlobalBlockWithStatus(RestStatus.SERVICE_UNAVAILABLE)) {
+            return ClusterHealthStatus.RED;
+        }
+
+        ClusterHealthStatus computeStatus = ClusterHealthStatus.GREEN;
+        for (String index : clusterState.metadata().getConcreteAllIndices()) {
+            IndexRoutingTable indexRoutingTable = clusterState.routingTable().index(index);
+            if (indexRoutingTable.allShardsActive()) {
```
original-brownbear, line 524: We're failing a test here if this turns out to be `null` (which could be the case and is not a bug) currently. We have to short-circuit on `indexRoutingTable == null` here as well, that should fix things.

original-brownbear (changes requested):
Tests found a small bug in this, see comment inline.

original-brownbear:
Jenkins test this

original-brownbear (approved):
LGTM thanks @howardhuanghua this is a nice speedup in our many-shards benchmark run.

howardhuanghua:
Thank you @original-brownbear.

PR #87732, opened 2022-06-16
Fix possible NPE in startup exception handling

ChrisHegarty:
Trivially, move startup exception cause message output into the null-protected block (since the cause can be null). This is a lithely oversight of a recent refactoring.


rjernst (approved):
Thanks Chris. Since this problem is very recent, it has not appeared in a release. It can be marked as “>non-issue”, the “>bug” label removed, and the release notes removed.

PR #87735, opened 2022-06-16
Use desired nodes during data tier allocation decisions 

fcofdez:
Today when the DataTierAllocationDecider checks if a shard can be allocated
in a node or not it uses the current cluster topology. This works fine in most cases
when the data flows from hotter to colder data nodes as it was designed. But
in certain situations users might want to remove a tier (i.e. the cold tier) that
allows having shards in hotter tiers, but with the current implementation once
a shard is allocated into the highest preference tier and there are nodes of that
tier, the shard won't move. This commit introduces a change that allows to use
the desired nodes (when available) to compute the preferred tier based on the
cluster topology provided by the desired nodes (only taking into account desired
members that are already members of the cluster).


x-pack/plugin/core/src/internalClusterTest/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierAllocationDeciderIT.java
```diff
@@ -67,6 +95,195 @@ public void testDefaultIndexAllocateToContent() {
...
+            .get()
+            .getSetting(index, INDEX_NUMBER_OF_REPLICAS_SETTING.getKey());
+
+        assertThat(Integer.parseInt(replicas), is(equalTo(numberOfColdNodes - 1)));
+
+        final List<DesiredNode> desiredNodesWithoutColdTier = new ArrayList<>(warmDesiredNodes);
+        desiredNodesWithoutColdTier.add(masterDesiredNode);
+
+        updateDesiredNodes(desiredNodesWithoutColdTier);
+
```
fcofdez, line 274: We have a small race here, since we execute the reroute asynchronously after updating the desired nodes, it's possible to break the invariants in:
```
    private void reroute(RoutingAllocation allocation) {
        assert hasDeadNodes(allocation) == false : "dead nodes should be explicitly cleaned up. See disassociateDeadNodes";
        assert AutoExpandReplicas.getAutoExpandReplicaChanges(allocation.metadata(), () -> allocation).isEmpty()
            : "auto-expand replicas out of sync with number of nodes in the cluster";
```
i.e: if we execute a reroute after updating the desired nodes, the test fail.

I haven't checked if we have the same issues just after adding a new node, but I think it might be possible?

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierAllocationDeciderTests.java
```diff
@@ -87,113 +90,94 @@ public class DataTierAllocationDeciderTests extends ESAllocationTestCase {
...
-            d = DataTierAllocationDecider.INSTANCE.canAllocate(shard, node, allocation);
-            assertThat(node.toString(), d.type(), equalTo(Decision.Type.NO));
-            assertThat(
-                node.toString(),
-                d.getExplanation(),
-                containsString(
-                    "index has a preference for tiers [data_warm,data_cold] " + "and node does not meet the required [data_cold] tier"
-                )
+        {
+            final var desiredNodes = createDesiredNodesWithActualizedNodes(WARM_DESIRED_NODE);
```
henningandersen, line 138: Perhaps add a short comment that we remove the cold tier here?

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierAllocationDeciderTests.java
```diff
@@ -87,113 +90,94 @@ public class DataTierAllocationDeciderTests extends ESAllocationTestCase {
...
-        assertThat(
-            node.toString(),
-            d.getExplanation(),
-            containsString("index has a preference for tiers [data_warm,data_cold] and node has tier [data_cold]")
-        );
+        {
+            // There's a warm node in the desired nodes, but it hasn't joined the cluster yet,
+            // in that case we consider that there aren't any nodes with the preferred tier in the cluster
+            final var clusterState = clusterStateWithIndexAndNodes(
+                "data_warm,data_cold",
```
henningandersen, line 166: Perhaps randomly pick between `"data_cold,data_warm", "data_warm,data_hot"`?

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierAllocationDeciderTests.java
```diff
@@ -222,43 +206,236 @@ public void testTierNodesPresent() {
...
+        assertTrue(DataTierAllocationDecider.tierNodesPresent("data_warm", nodes));
+        assertTrue(DataTierAllocationDecider.tierNodesPresent("data_cold", nodes));
+        assertTrue(DataTierAllocationDecider.tierNodesPresent("data_content", nodes));
+    }
+
     public void testPreferredTierAvailable() {
-        DiscoveryNodes nodes = DiscoveryNodes.builder().build();
+        {
+            final var nodes = DiscoveryNodes.builder().build();
+            final DesiredNodes desiredNodes = null;
```
henningandersen, line 238: Not sure I follow why we have this variable? Perhaps you intended to randomize it to sometimes have desired nodes?

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierAllocationDeciderTests.java
```diff
@@ -222,43 +206,236 @@ public void testTierNodesPresent() {
...
+                ),
+                equalTo(Optional.of("data_hot"))
+            );
+        }
+
+        {
+            // When there are desired nodes that haven't joined the cluster yet, those are not considered
+            final var nodes = DiscoveryNodes.builder().add(WARM_NODE).add(CONTENT_NODE).build();
+            // i.e. HOT_DESIRED_NODE might be part of the DesiredNodes, but it is not part of the cluster yet
+            final var desiredNodes = createDesiredNodesWithActualizedNodes(WARM_DESIRED_NODE, CONTENT_DESIRED_NODE);
```
henningandersen, line 340: Did you intend to add `HOT_DESIRED_NODE` as pending here?

henningandersen (commented):
This looks good. Left a few comments to tackle, but otherwise no concerns.

x-pack/plugin/core/src/internalClusterTest/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierAllocationDeciderIT.java
fcofdez: > I am not entirely sure if you mean that the two invariants mentioned can break or whether you are arguing why the subsequent assertion is in an assertBusy, can you clarify?

I meant that the invariant `assert AutoExpandReplicas.getAutoExpandReplicaChanges(allocation.metadata(), () -> allocation).isEmpty()` can break if there's a reroute executed before the reroute that we trigger asynchronously after updating the desired nodes.

server/src/main/java/org/elasticsearch/action/admin/cluster/desirednodes/TransportUpdateDesiredNodesAction.java
```diff
@@ -194,7 +198,12 @@ public ClusterState execute(ClusterState currentState, List<TaskContext<UpdateDe
 
             desiredNodes = DesiredNodes.updateDesiredNodesStatusIfNeeded(currentState.nodes(), desiredNodes);
 
-            return desiredNodes == initialDesiredNodes ? currentState : replaceDesiredNodes(currentState, desiredNodes);
+            if (desiredNodes == initialDesiredNodes) {
+                return currentState;
+            } else {
+                final var clusterStateWithUpdatedDesiredNodes = replaceDesiredNodes(currentState, desiredNodes);
+                return allocationService.adaptAutoExpandReplicas(clusterStateWithUpdatedDesiredNodes);
```
henningandersen, line 205: nit: I would prefer:

```suggestion
                final ClusterState withUpdatedDesiredNodes = replaceDesiredNodes(currentState, desiredNodes);
                return allocationService.adaptAutoExpandReplicas(withUpdatedDesiredNodes);
```

henningandersen (approved):
LGTM, though there is a related test failure to resolve. Perhaps also run it through CI a few times (locally or on the PR) to shake out more problems.

fcofdez:
@elasticmachine update branch

fcofdez:

fcofdez:

fcofdez:
@elasticmachine update branch

fcofdez:
Thanks for all the feedback Henning!

PR #87736, opened 2022-06-16
Log 'output of node' separators to stdout

DaveCTurner:
Today when using Gradle-managed test clusters we separate the log output
from each node with a separator that looks like this:

    === Log output of node `node{:qa:mixed-cluster:v8.2.0-0}` ===

However, we write this to `stderr` and yet write the rest of the log
output to `stdout`. This commit moves the separators to the same log
level as the rest of this output so that they can all be redirected to
the same place.


pugnascotia (approved):
LGTM

PR #87740, opened 2022-06-16
Modularize the legacy-geo component

ChrisHegarty:
This change modularizes the legacy-geo component, by adding a module-info.java

Additionally, a number of transitive dependencies, namely s2-geometry-library-java and lucene-spatial3d, have been explicitly added, since they are required to be present in the module graph.

relates #78744


ChrisHegarty:

build-tools-internal/version.properties
```diff
@@ -23,6 +23,8 @@ netty             = 4.1.76.Final
 
 commons_lang3                   = 3.9
 
+s2_geometry_library_java        = 1.0.1
```
rjernst, line 26: I think this can just be local to the legacy-geo module. Versions listed in this file are normally for when they need to be shared across components.

rjernst (approved):
LGTM

Issue #87729, opened 2022-06-16
`TokenService` clean up


n1v0lg:
### Description

The `TokenService` could use some spring cleaning 🧹 🧹 🧹 

It has evolved quite organically; a sizeable portion of the code is either (1) not actually used or (2) there for backwards compatibility with older ES versions (pre-`7.17`).

This issue tracks the various cuts we can make.

- [ ] Remove dead code (several areas, but here is an [example](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/TokenService.java#L451))
- [ ] Remove key rotation code (e.g. [this](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/TokenService.java#L2351-L2357)). This code was never actually used. 

The real gain comes if we are willing to drop BWC support for tokens created in ES versions older than `7.17`. This issue is bigger in scope and we still need to align on whether we want to proceed. It's tracked separately under #87726.

Issue #87726, opened 2022-06-16
`TokenService` - remove code for BWC support for pre-`7.17` ES versions  


n1v0lg:
### Description

Throughout the evolution of the `TokenService`, we've ensured that access tokens created in older versions remain valid in newer versions of ES. Effectively, upgrades should remain "unnoticable" to end-users and not affect their Kibana sessions; access tokens created before a version upgrade (or during, if it was a rolling upgrade) should remain valid _after_ the upgrade.

As a result, there is a lot of code that handles BWC for legacy tokens. A guesstimate is over 20% of the entire class. This makes the service harder to understand, maintain, and evolve.

This issue proposes to remove BWC support code for tokens from ES versions older than `7.17`. The rationale is:
 
1. Tokens are ephemeral and short-lived; they expire after at most one hour. As such, all the support burden is to effectively ensure a BWC window of one hour. After an upgrade is complete, new tokens are created according the format and rules of the new ES version. 
1. To upgrade to `8.x` from pre-`7.17`, customers _must_ first upgrade to `7.17`. `7.17` tokens remain fully compatible with `8.x`.

The benefits of maintaining BWC in this case seems unwarranted given the overall maintenance burden.

If we remove BWC support, customers would only be impacted if they upgraded from `pre-7.17` -> `7.17` -> `8.x` in under an hour. The impact itself would be modest: end-users with "unsupported", still-active legacy tokens would have to repeat login. (**Note:** I still need to fully validate this claim and ensure that there aren't cases where login sessions get affected in more permanent, damaging ways. It's very unlikely).

To further reduce impact we could explicitly advise customers to wait for one hour before running the upgrade from `7.17` -> `8.x`; this would ensure that all legacy tokens are expired.

Some code we would be able to cut (there may be more):

- [ ] Handling tokens stored in the `.security` index (we introduced a [dedicated index for tokens in 7.2](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/TokenService.java#L215)) 
- [ ] Handling tokens stored as UUIDs (also [7.2](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/TokenService.java#L215))
[...]

PR #87744, opened 2022-06-16
Remove key rotation code from `TokenService`

n1v0lg:
The key rotation code in the `TokenService` was never used outside of
tests. I've made a conservative pass with this PR that removes key
rotation related code but maintains BWC. 

I haven't touched things like the `keyCache` field and the code related
to the actual use of the encryption key. AFAICT that code is still
relevant from a BWC perspective, where upgrades from old versions are
concerned. Making BWC cuts is tracked separately (#87726).

Relates #87729


tvernum (approved):
LGTM

n1v0lg:
@elasticmachine run elasticsearch-ci/part-1-fips plz

PR #87746, opened 2022-06-16
Modularize the ingest-geoip component

ChrisHegarty:
This change modularizes the ingest-geoip component, by adding a module-info.java

Additionally, the Jackson dependencies have been changed to `runtimeOnly`, since they are all transitive (from com.maxmind.geoip2).

relates #78744


modules/ingest-geoip/src/main/java/module-info.java
```diff
@@ -0,0 +1,22 @@
...
+
+module org.elasticsearch.ingest.geoip {
+    requires org.elasticsearch.base;
+    requires org.elasticsearch.server;
+    requires org.elasticsearch.xcontent;
+    requires org.apache.logging.log4j;
+    requires org.apache.lucene.core;
+    requires com.fasterxml.jackson.annotation;
+    requires com.fasterxml.jackson.core;
+    requires com.fasterxml.jackson.databind;
```
rjernst, line 17: Do *we* actually use databind in geoip, or is it transitive? If so, shouldn't we just pick this up from the jackson core module-info, so we don't need to require here?

modules/ingest-geoip/build.gradle
```diff
@@ -26,7 +26,7 @@ dependencies {
   // geoip2 dependencies:
   implementation("com.fasterxml.jackson.core:jackson-annotations:2.13.1")
   implementation("com.fasterxml.jackson.core:jackson-databind:2.13.1")
-  runtimeOnly("com.fasterxml.jackson.core:jackson-core:2.13.1")
+  implementation("com.fasterxml.jackson.core:jackson-core:2.13.1")
```
rjernst, line 29: This is unfortunate since it means jackson will now show up in eg autocomplete again when working within ingest-geoip classes :/

rjernst (approved):
Looks fine, one minor question

modules/ingest-geoip/build.gradle
ChrisHegarty: Yeah, this is not correct. Let me revisit this.

modules/ingest-geoip/src/main/java/module-info.java
ChrisHegarty: You are correct, the requires on Jackson is not required (the dependency is transitive)

modules/ingest-geoip/build.gradle
```diff
@@ -39,6 +39,14 @@ restResources {
...
 }
 
+compileJava {
+  options.compilerArgs << '-Xlint:-classfile'
+}
+
+compileTestJava {
+  options.compilerArgs << '-Xlint:-classfile'
+}
+
```
ChrisHegarty, line 49: For now, suppress the lint warnings to avoid issues like this:

```
/Users/chegar/.gradle/caches/modules-2/files-
2.1/com.maxmind.geoip2/geoip2/3.0.0/523b0d63f3dbeacb7dfceb7431cb17fa56cf79fa/geoip2-
3.0.0.jar(/com/maxmind/geoip2/model/CityResponse.class): 
warning: Cannot find annotation method 'value()' in type 'JsonProperty': class file for 
com.fasterxml.jackson.annotation.JsonProperty not found
```

There is likely a javac issue here. Investigating.

ChrisHegarty: @breskeby Is there a better way to suppress the warning here, rather than three separate blocks?

breskeby:
@ChrisHegarty You can do

tasks.withType(JavaCompile).configureEach{
…
}

breskeby:
No worries :-)

PR #87750, opened 2022-06-16
Desired balance calculator

idegtiarenko:
It is possible that `org.elasticsearch.cluster.routing.allocation.allocator.DesiredBalanceShardsAllocator#allocate` are executed very fast (because the current balance is empty or very small). On the other side even single iteration inside `desiredBalanceComputer.compute` is taking significant time (may be there were multiple many shard index just created).

This way there could be multiple allocates happening before we even detect that current desired balance computation is outdated. This will result in first one (even while being outdated) computing correct desired balance with (isFresh=false, shouldReroute=true) and the last one becoming (isFresh=true, shouldReroute=false).

To fix this situation we should update reroute check to compare computed desired balance with one that was actually applied last.

This was reproduce-able with ~6% probability in: org.elasticsearch.snapshots.BlobStoreIncrementalityIT#testIncrementalBehaviorOnPrimaryFailover (and is fixed in this pr)

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalance.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalance.java
@@ -28,7 +28,5 @@ public record DesiredBalance(long lastConvergedIndex, Map<ShardId, ShardAssignme
         return Objects.equals(a.assignments, b.assignments) == false;
     }
 
-    public static DesiredBalance initial() {
-        return new DesiredBalance(-1, Map.of());
-    }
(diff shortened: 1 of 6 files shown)


idegtiarenko:
This change resulted in new integration test failures: 
* RareClusterStateIT. testDeleteCreateInOneBulk
* SimpleIndexStateIT. testFastCloseAfterCreateContinuesCreateAfterOpen

I am looking into them

idegtiarenko:
It also looks like there are no stuck/timing out tests on this branch (so probably :crossed_fingers:  there are no more leaking listeners)

DaveCTurner (approved):
LGTM

PR #87751, opened 2022-06-16
Update lang-expression transitive dependencies

ChrisHegarty:
This change updates the ASM dependencies of the lang-expression component.

ASM is a transitive dependency, not used directly by the code in lang-expression but transitively by lucene-expressions. We're currently at Lucene-9.2.0, which depends upon `org.ow2.asm:*=7.2`.

P.S. I stumbled upon this when attempting to add a module-info to lang-expressions, since the older ASM does not have the same module name as that of more recent ASM versions.


ChrisHegarty:
A question here, that @breskeby can probably best answer. For such transitive dependencies, is `api`, or `implementation`, or something else, more appropriate?

breskeby:
@ChrisHegarty Actually reading your description more carefully and not just the code change what I described above is only true for compile dependencies you wanna bring in. Since `lang-expression` is not using asm at all at compile time, `runtimeOnly` is the right choice in this case

jdconrad (approved):
@ChrisHegarty Thanks for updating this! Looks good to me with the `runtimeOnly` change suggested by @breskeby .

ChrisHegarty:
@breskeby Thanks for the detailed comments, they are very helpful.  I've updated the transitive dependencies to be `runtimeOnly`.

breskeby (commented):
👍

PR #86315, opened 2022-04-30
Reject openConnection attempt while closing

DaveCTurner:
Today while the `ClusterConnectionManager` is closing it will reject
attempts to open _managed_ connections (i.e. using `connectToNode`), but
it still permits ad-hoc connections (i.e. using `openConnection`). This
commit extends the existing refcounting mechanism to cover both cases,
preventing all concurrent connection attempts while shutting down.

Closes #86249
Relates #77539

PR #87756, opened 2022-06-16
Add more detail to assertion in TransportService

DaveCTurner:
#86315 was supposed to fix this but it looks like it still tripped in https://gradle-enterprise.elastic.co/s/gkje7i7ctfs6g. This commit adds a bit more detail to the message (action name, connection state) that might be helpful.


fcofdez (approved):
LGTM

PR #87763, opened 2022-06-16
Add immutable 'operator' metadata classes for cluster state

grcevski:
This PR relates to #86224 (Operator/File cluster settings) where we need to store various information in the Metadata about the most recent processing of the operator cluster state update.

Since the original PR is too large on its own, I'm separating the individual components of the PR for easier review.


server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
```diff
@@ -938,6 +943,24 @@ public Map<String, Custom> customs() {
...
+    public Map<String, OperatorMetadata> operatorMetadata() {
+        return this.operatorMetadata;
+    }
+
+    /**
+     * Returns the {@link OperatorMetadata} for a given namespace.
+     * @param namespace the namespace to lookup operator metadata for
+     * @return {@link OperatorMetadata} or null if not found
+     */
+    public OperatorMetadata operatorMetadata(String namespace) {
```
rjernst, line 960: Why have this method and also the one above returning the entire map? If the one above is needed, then it could also be used in place of this method?

server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
```diff
@@ -1146,6 +1181,7 @@ public Metadata apply(Metadata part) {
             builder.indices(indices.apply(part.indices));
             builder.templates(templates.apply(part.templates));
             builder.customs(customs.apply(part.customs));
+            builder.put(operatorMetadata.apply(part.operatorMetadata));
```
rjernst, line 1184: Does the jdk map diff return an immutable/unmodifiable map? We could either wrap or copy this here to make it immutable, or even better open a separate issue to make the jdk map diff's apply return an immutable map (I've had the need for this in recent hppc PRs)

server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
```diff
@@ -1635,6 +1683,26 @@ public Builder customs(Map<String, Custom> customs) {
             return this;
         }
 
+        /**
+         * Adds a map of namespace to {@link OperatorMetadata} into the metadata builder
+         * @param operatorMetadata a map of namespace to {@link OperatorMetadata}
+         * @return {@link Builder}
+         */
+        public Builder put(Map<String, OperatorMetadata> operatorMetadata) {
```
rjernst, line 1691: Is this used anywhere?

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorErrorMetadata.java
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+ * A metadata class to hold error information about errors encountered
+ * while applying a cluster state update for a given namespace.
+ * <p>
+ * This information is held by the {@link OperatorMetadata} class.
+ */
+public record OperatorErrorMetadata(
+    Long version,
+    org.elasticsearch.cluster.metadata.OperatorErrorMetadata.ErrorKind errorKind,
```
rjernst, line 32: nit: static import ErrorKind?

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorErrorMetadata.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    /**
+     * Reads an {@link OperatorErrorMetadata} from a {@link StreamInput}
+     * @param in the {@link StreamInput} to read from
+     * @return {@link OperatorErrorMetadata}
+     * @throws IOException
+     */
+    public static OperatorErrorMetadata readFrom(StreamInput in) throws IOException {
+        Builder builder = new Builder().version(in.readLong())
```
rjernst, line 58: Why create a builder, these can be passed to the canonical ctor?

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorErrorMetadata.java
```diff
@@ -0,0 +1,232 @@
...
+            return new OperatorErrorMetadata(version, errorKind, Collections.unmodifiableList(errors));
+        }
+
+        /**
+         * Serializes an {@link OperatorErrorMetadata} to xContent
+         *
+         * @param metadata {@link OperatorErrorMetadata}
+         * @param builder {@link XContentBuilder}
+         */
+        public static void toXContent(OperatorErrorMetadata metadata, XContentBuilder builder) throws IOException {
```
rjernst, line 152: This could be a member method of the record?

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorErrorMetadata.java
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        /**
+         * Reads an {@link OperatorErrorMetadata} from xContent
+         *
+         * @param parser {@link XContentParser}
+         * @return {@link OperatorErrorMetadata}
+         * @throws IOException
+         */
+        public static OperatorErrorMetadata fromXContent(XContentParser parser) throws IOException {
```
rjernst, line 167: Shouldn't this be on the record, not the builder?

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorHandlerMetadata.java
```diff
@@ -0,0 +1,144 @@
...
+            return new OperatorHandlerMetadata(name, Collections.unmodifiableSet(keys));
+        }
+
+        /**
+         * Serializes an {@link OperatorHandlerMetadata} to xContent
+         *
+         * @param metadata {@link OperatorHandlerMetadata}
+         * @param builder {@link XContentBuilder}
+         */
+        public static void toXContent(OperatorHandlerMetadata metadata, XContentBuilder builder) throws IOException {
```
rjernst, line 110: Why is toXContent needed on both the builder (as a static that takes the record?) and on the record itself?

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorMetadata.java
```diff
@@ -0,0 +1,311 @@
...
+         * @param namespace the operator metadata namespace. See {@link Metadata} operatorMetadata.
+         * @return {@link OperatorMetadata}
+         * @throws IOException
+         */
+        public static OperatorMetadata fromXContent(XContentParser parser, String namespace) throws IOException {
+            OperatorMetadata.Builder builder = new OperatorMetadata.Builder(namespace);
+
+            String currentFieldName = skipNamespace(parser);
+            XContentParser.Token token;
+            while ((token = parser.nextToken()) != XContentParser.Token.END_OBJECT) {
```
rjernst, line 277: Could a ConstructingObjectParser be used instead of this loop (same goes for other fromXContent in this PR)? If using that, would a builder even be necessary?

rjernst (changes requested):
Thanks for splitting this out. Overall the use of records is nice here, and I appreciate all the javadocs! I left some questions. The downside of the way the PR is split is I can't tell *how* any of these classes are intended to be used, so it is difficult to comment on particular methods and whether they could be simplified. I'm not sure how to remedy that...

grcevski:
> Thanks for splitting this out. Overall the use of records is nice here, and I appreciate all the javadocs! I left some questions. The downside of the way the PR is split is I can't tell _how_ any of these classes are intended to be used, so it is difficult to comment on particular methods and whether they could be simplified. I'm not sure how to remedy that...

Yeah, I figured it will be super hard without context. I'll start adding comments on various PR lines with context from the original PR in how it's used. I hope that will help.

server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
grcevski: Ah yeah, I used to call this method putOperatorState, but it was too long and it didn't match the naming of other methods we have in Metadata, we simply call `.put` with different types. Here's the usage here:

https://github.com/elastic/elasticsearch/pull/86224/files#diff-8ca6818d706ec1720e045ea8a1f4f8e4b5dda47c5610aed2bf8e7d55121d8391R129

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorErrorMetadata.java
grcevski: Sure, I'll move them in, I was following patterns used elsewhere in the metadata.

grcevski: Actually for this one is a lot more convenient to be a builder method, we are iterating over the parser elements and we fill in the pieces of data we parse inside the builder. I had made the toXContent part of the builder too since they share the static strings, but I can remove that wrapping.

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorHandlerMetadata.java
grcevski: I'll move it to the record only, similar to other places, I'm sharing the string constants for to and from xcontent.

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorMetadata.java
grcevski: I'll look into that, it would be great if we can.

server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
```diff
@@ -942,6 +947,15 @@ public Map<String, Custom> customs() {
         return this.customs;
     }
 
+    /**
```
grcevski, line 950: Here's an example in the related PR of how this will be used:

https://github.com/elastic/elasticsearch/pull/86224/files#diff-d017ebae07494c580da12a539a666fe1d87537f167e71761cf60451b8913d554R127

The related PR is out of date, it calls the API operatorState and it has the extra helper method to perform the additional .get.

Instead of:

```
OperatorMetadata existingMetadata = state.metadata().operatorState(namespace);
```

we'll be calling:

```
OperatorMetadata existingMetadata = state.metadata().operatorMetadata.get(namespace);
```

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorErrorMetadata.java
```diff
@@ -0,0 +1,151 @@
...
+import static org.elasticsearch.xcontent.ConstructingObjectParser.constructorArg;
+
+/**
+ * A metadata class to hold error information about errors encountered
+ * while applying a cluster state update for a given namespace.
+ * <p>
+ * This information is held by the {@link OperatorMetadata} class.
+ * </p>
+ */
+public record OperatorErrorMetadata(Long version, ErrorKind errorKind, List<String> errors)
```
grcevski, line 33: The error metadata is set when a file fails to get processed for some reason. In this case we launch a cluster state update task that will update this metadata. Example here:

https://github.com/elastic/elasticsearch/pull/86224/files#diff-a7da9bd5e6ce66769b5113745baa42c88b7c21b916a8ea290434048fc5af2c5fR29

The builder code is now removed, so the OperatorErrorMetadata will be directly created.

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorHandlerMetadata.java
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Metadata class to hold the keys set in operator mode for each operator handler.
+ *
+ * <p>
+ * Since we hold operator metadata state for multiple namespaces, the same handler can appear in
+ * multiple namespaces. See {@link OperatorMetadata} and {@link Metadata}.
+ * </p>
+ */
+public record OperatorHandlerMetadata(String name, Set<String> keys)
```
grcevski, line 36: The OperatorHandlerMetadata is stored when we successfully process a new operator cluster state. Here's an example usage from the task that does the final update of the cluster state:

https://github.com/elastic/elasticsearch/pull/86224/files#diff-8ca6818d706ec1720e045ea8a1f4f8e4b5dda47c5610aed2bf8e7d55121d8391R40

specifically here:

https://github.com/elastic/elasticsearch/pull/86224/files#diff-8ca6818d706ec1720e045ea8a1f4f8e4b5dda47c5610aed2bf8e7d55121d8391R128

The only difference is that I renamed the API so instead of:

```
Metadata.Builder metadataBuilder = Metadata.builder(state.metadata()).putOperatorState(operatorMetadataBuilder.build());
```

we'll do:

```
Metadata.Builder metadataBuilder = Metadata.builder(state.metadata()).put(operatorMetadataBuilder.build());
```

grcevski:
I think I applied all suggestions and I linked with the relevant code in the WIP PR. I hope it's sufficiently described for a review.

grcevski:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorHandlerMetadata.java
```diff
@@ -0,0 +1,99 @@
...
+     * @throws IOException
+     */
+    public static OperatorHandlerMetadata readFrom(StreamInput in) throws IOException {
+        return new OperatorHandlerMetadata(in.readString(), in.readSet(StreamInput::readString));
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+        builder.startObject(name());
+        builder.stringListField(KEYS.getPreferredName(), new TreeSet<>(keys())); // ordered set here to ensure output consistency
```
rjernst, line 62: Building a TreeSet would seem more complicated and costly (lots of object creations) than simply sorting the keys?

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorMetadata.java
```diff
@@ -0,0 +1,292 @@
...
+    public static Builder builder(String namespace, OperatorMetadata metadata) {
+        return new Builder(namespace, metadata);
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+        builder.startObject(namespace());
+        builder.field(VERSION.getPreferredName(), version);
+        builder.startObject(HANDLERS.getPreferredName());
+        var sortedKeys = new TreeSet<>(handlers.keySet());
```
rjernst, line 164: Instead of using a TreeSet here to do the sorting, we can use the stream api?

```
handlers.entrySet().stream().sorted(Map.Entry.comparingByKey()).forEach(e -> {
    e.getValue().toXContent(builder, params);
}
```

rjernst (approved):
LGTM. Just a couple suggestions for avoiding TreeSets

server/src/main/java/org/elasticsearch/cluster/metadata/OperatorMetadata.java
```diff
@@ -0,0 +1,292 @@
...
+    private static final ParseField VERSION = new ParseField("version");
+    private static final ParseField HANDLERS = new ParseField("handlers");
+    private static final ParseField ERRORS_METADATA = new ParseField("errors");
+
+    /**
+     * OperatorMetadata contains information about settings set in operator mode.
+     *
+     * <p>
+     * These settings cannot be updated by the end user and are set outside of the
+     * REST layer, e.g. through file based settings or by plugin/modules.
```
rjernst, line 58: I forgot to mention this naming issue. Since all of these classes are about more than operator settings, I think we need better terminology that applies to file based settings and programmatic fixed settings. I'm not sure what that naming should be, but as it is the "operator" terminology will be confusing when we start using this in plugins and modules.

rjernst: Perhaps a name referencing cluster state. Some ideas:

```
unmodifiable cluster state
immutable cluster state
....
```

So then this class might be `ImmutableClusterStateMetadata`. It's not the best, but it demonstrates the idea.

grcevski: Good idea, I've been struggling with terminology on this. I like the immutable prefix, it's really what it is. Operator is just the mode we set them in initially.

grcevski:
@elasticmachine update branch

Issue #80504, opened 2021-11-08
Mappings should fall back to _source when doc values are disabled


jpountz:
People working on scripting have been looking into making scripts fall back automatically to `_source` when doc values are not enabled, in order to make scripts easier to use. @jdconrad and I discussed it recently, and it would be better to do it at the mapping level. This way, whether doc values or `_source` is used for scripts would be completely transparent to script engines, and we could also support sorting or aggregations on fields that have doc values disabled. We need to have the ability to read from `_source` directly for runtime fields anyway, so hopefully we could reuse code?

One question is whether this could be a performance trap at times. Maybe we should look into ways to tell users that their queries could run faster with better mappings?

jdconrad:
This is an important component of the scripting (Painless) fields API that we would like to get prioritized. See (https://github.com/elastic/elasticsearch/issues/79105).

jpountz:
cc @giladgal

jdconrad:
I discussed this further with @jpountz today, and he suggested that we leverage the existing fielddataBuilder in MappedFieldType to do the fallback in the case when doc values are not enabled. He also suggested that fallback use the existing runtime fields infrastructure to generate the appropriate doc values from source. 

I believe this is a good path forward but does require some additional plumbing work to get the source paths for a specific field into scripting and other places that would want to leverage source fallback.

@jpountz also suggested for text fields as noted in (https://github.com/elastic/elasticsearch/issues/81246) we could possibly follow up with another fielddataBuilder method that would default to source for specific field types or allow further customization of the values returned even if doc values were to exist. The scripting fields API could then use this second method instead of the first one.

javanna:
We have re-discussed this with the team and decided to decouple the scripting needs (load doc_values or _source transparently, depending on what's available) from the idea of falling back to _source when doc_values are not available for sorting and aggregations.

[...]

PR #87765, opened 2022-06-16
Add source fallback for keyword fields

jdconrad:
This change adds source fallback specifically for the keyword field type as part of its mapped field type as an example of how this could be done for all the fields types required for scripting (and possibly other areas such as aggs and sorting?)

I've tried to implement the design as discussed with @jpountz @jtibshirani and @nik9000 as closely as possible with the minimally invasive change. The design is discussed further in this comment (https://github.com/elastic/elasticsearch/issues/80504#issuecomment-1155757482).

My questions for y'all are the following:
1. Are there any design changes that you think would be an improvement? Or is the design as it stands now good?
2. I've only added one test section as part of Painless. What other tests would you expect to get decent coverage? Outside of scripting, I'm not very familiar with what else would need to be tested directly.
3. Should we add a source fallback enabled boolean to fielddataBuilder? I think it's important at least initially that source fallback be optional since it has some heavy performance implications, and as I discussed with @jpountz, since the old-style scripting syntax for doc (`doc['field'].value`) also depends on this code path, without the boolean there is a possible inconsistency in what values a user may get from a script.

### DIFF
--- a/x-pack/plugin/core/src/main/resources/monitoring-logstash-mb.json
+++ b/x-pack/plugin/core/src/main/resources/monitoring-logstash-mb.json
@@ -228,6 +228,9 @@
                           "type": "long"
                         }
                       }
+                    },
+                    "timestamp": {
+                      "type": "date"
                     }
                   }
                 },
--- a/x-pack/plugin/monitoring/src/main/java/org/elasticsearch/xpack/monitoring/MonitoringTemplateRegistry.java
+++ b/x-pack/plugin/monitoring/src/main/java/org/elasticsearch/xpack/monitoring/MonitoringTemplateRegistry.java
@@ -78,7 +78,7 @@ public class MonitoringTemplateRegistry extends IndexTemplateRegistry {
      * writes monitoring data in ECS format as of 8.0. These templates define the ECS schema as well as alias fields for the old monitoring
      * mappings that point to the corresponding ECS fields.
      */
-    public static final int STACK_MONITORING_REGISTRY_VERSION = Version.V_8_0_0.id + 2;
+    public static final int STACK_MONITORING_REGISTRY_VERSION = Version.V_8_0_0.id + 3;
     private static final String STACK_MONITORING_REGISTRY_VERSION_VARIABLE = "xpack.stack.monitoring.template.release.version";
     private static final String STACK_TEMPLATE_VERSION = "8";
     private static final String STACK_TEMPLATE_VERSION_VARIABLE = "xpack.stack.monitoring.template.version";


jpountz:
I didn't review the change thouroughly, only the `KeywordFieldMapper` change, and it has the `if (hasDocValues()) { /* use doc values */ } otherwise { /* use source */ }` logic that I was expected, in my opinion this is the right way to implement source fallback.

> Should we add a source fallback enabled boolean to fielddataBuilder? I think it's important at least initially that source fallback be optional since it has some heavy performance implications,

If you asked me several years ago, I would have agreed on the trapiness of the performance implication. I'm less convinced now, we make it so easy to follow the slow path using runtime fields, which you can even provide as part of the search request. Maybe we should just embrace it. Given that fields have doc values by default, if it's slow, it means that users opted in for slowness so we shouldn't worry?

> since the old-style scripting syntax for doc (doc['field'].value) also depends on this code path, without the boolean there is a possible inconsistency in what values a user may get from a script.

What inconsistencies do you have in mind? The fielddata instance returned by runtime fields, which you are reusing, is supposed to be consistent with the instance that is returned when fields have doc values, so this change should be transparent to scripting APIs?

jdconrad:
@jpountz Thank you for the feedback! I realize we discussed this already, and my main concern still lies in the expected results. Now users will get results from source when they may have anticipated an exception, but as you say it's probably just an improvement outside of performance implications.

jdconrad:
Removing the WIP label as this is the pattern I would like to follow moving forward.

server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -681,12 +682,19 @@ NamedAnalyzer normalizer() {
...
-                CoreValuesSourceType.KEYWORD,
-                (dv, n) -> new KeywordDocValuesField(FieldData.toString(dv), n)
-            );
+            if (hasDocValues()) {
+                return new SortedSetOrdinalsIndexFieldData.Builder(
+                    name(),
+                    CoreValuesSourceType.KEYWORD,
+                    (dv, n) -> new KeywordDocValuesField(FieldData.toString(dv), n)
+                );
+            } else {
```
nik9000, line 691: I tend to prefer removing the `else`if you `return` from the if. No big deal either way.

server/src/main/java/org/elasticsearch/script/StringFieldScript.java
```diff
@@ -43,6 +43,23 @@ public boolean isResultDeterministic() {
         }
     };
 
+    public static final StringFieldScript.Factory PARSE_FROM_SOURCE_PATHS = new Factory() {
```
nik9000, line 46: It's probably worth javadoc to explain how this is different from `PARSE_FROM_SOURCE`.

server/src/main/java/org/elasticsearch/script/AbstractFieldScript.java
```diff
@@ -111,6 +114,14 @@ protected final void emitFromSource() {
         }
     }
 
+    protected final void emitFromSourcePaths() {
```
nik9000, line 117: It's probably worth adding javadoc to explain how this is different from `emitFromSource`.

server/src/main/java/org/elasticsearch/script/StringFieldScript.java
jdconrad: The runtime fields change is definitely the part I was least confident in the solution for. Should source paths be used for all source-only runtime fields? Would that make more sense than having two of these? Does it ever make sense to specify a source-only field as part of the runtime mappings where the source would be part of the parent field?

server/src/main/java/org/elasticsearch/script/AbstractFieldScript.java
jdconrad: Agreed, at a minimum! Questions from the previous comment are relevant here too.

server/src/main/java/org/elasticsearch/script/AbstractFieldScript.java
```diff
@@ -75,6 +77,7 @@ public AbstractFieldScript(String fieldName, Map<String, Object> params, SearchL
 
         this.fieldName = fieldName;
         Map<String, Object> docAsMap = docAsMap();
+        this.sourcePaths = searchLookup.sourcePaths(fieldName);
```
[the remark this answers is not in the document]
jdconrad: I'm happy to make this change. I do think this will be extended to all available runtime fields, though. Maybe it would make sense to provide an additional constructor or a boolean value to have this be null?

jdconrad:
@nik9000 Thank you for the feedback! Left some questions in the comments. Also, do you know if there are any negative effects that could come from synthetic source or not having source at all? Would it just output no values?

server/src/main/java/org/elasticsearch/script/StringFieldScript.java
nik9000: > Should source paths be used for all source-only runtime fields?

Truly runtime fields *can't* have sub-fields. That concept only exists in the mapping and runtime fields just have a single path.

nik9000: That isn't a simple thing, for what it's worth. We could have built sub-fields into runtime fields. We just didn't because we didn't know why we would have. We didn't think, at the time, that they'd *do* anything.

server/src/main/java/org/elasticsearch/script/AbstractFieldScript.java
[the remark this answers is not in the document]
nik9000: I have an instinct that if you implemented it with copy and paste for two field types, like, `keyword` and `ip` or something, and then tried to remove the duplications *something* would pop out. I don't know what. I just think it might be worth trying before adding this for all runtime fields.

jdconrad:
@javanna @jtibshirani Was wondering if either of you had any thoughts, suggestions, or concerns? I'd like to collect all the feedback before I make any further changes. Thanks!

javanna:
I posted my quick thoughts [yesterday](https://github.com/elastic/elasticsearch/issues/80504#issuecomment-1160384729). I think this change deserves more discussion, but it may very well be that this is my feeling as I was not involved when the decision was made to go down this route.

jdconrad:
@javanna Thanks for the feedback! Apologies as I missed your comment in the other issue.

jdconrad:
Github closed this on me when I was trying to fix the changelog. Will open a new issue with the updated design.

PR #87767, opened 2022-06-16
Add OperatorHandler interface

grcevski:
This PR relates to #86224 (Operator/File cluster settings). We are introducing a new action handler interface called OperatorHandler, which will be used to implement various operator mode actions to update the cluster state. Since handlers can belong to modules and plugins, we introduce a SPI to declare the handlers a plugin/module provides.

For reference, please see OperatorLifecycleAction.java in #86224.

Since the original PR is too large on its own, I'm separating the individual components of the PR for easier review.


grcevski:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -295,6 +296,31 @@ static void loadExtensions(Collection<LoadedPlugin> plugins) {
...
+    /**
+     * SPI convenience method that uses the {@link ServiceLoader} JDK class to load various SPI providers
+     * from plugins/modules. For example:
+     *
+     * var pluginHandlers = pluginsService.loadServiceProviders(OperatorHandlerProvider.class);
+     *
+     * @param service A templated service class to look for providers in plugins
+     * @return an immutable {@link List} of discovered providers in the plugins/modules
+     */
+    public <T> List<? extends T> loadServiceProviders(Class<T> service) {
```
rjernst, line 313: Can this be split out into a separate PR? It seems like this is most of this PR, but the title would suggest otherwise.

grcevski: Yes, sounds good, I'll take this out from this PR and contribute it back once the operator handler interface is in.

server/src/main/java/org/elasticsearch/operator/OperatorHandler.java
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Convenience method that creates a {@link XContentParser} from a content map so that it can be passed to
+     * existing REST based code for input parsing.
+     *
+     * @param config XContentParserConfiguration for this mapper
+     * @param source the operator content as a map
+     * @return
+     */
+    default XContentParser mapToXContentParser(XContentParserConfiguration config, Map<String, ?> source) {
```
rjernst, line 110: This doesn't seem to have anything to do with operator config, could it be in a more general place for xcontent utilities?

server/src/main/java/org/elasticsearch/operator/OperatorHandler.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Convenience method to convert the incoming passed in input to the transform method into a map.
+     *
+     * @param input the input passed into the operator handler after parsing the content
+     * @return
+     */
+    @SuppressWarnings("unchecked")
+    default Map<String, ?> asMap(Object input) {
```
rjernst, line 95: This doesn't seem like something specific to operator config, could it be in a more appropriate location for utility methods?

server/src/main/java/org/elasticsearch/operator/OperatorHandler.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * All implementations of OperatorHandler should call the request validate method, by calling this default
+     * implementation. To aid in any special validation logic that may need to be implemented by the operator handler
+     * we provide this convenience method.
+     *
+     * @param request the master node request that we base this operator handler on
+     */
+    default void validate(MasterNodeRequest<?> request) {
```
rjernst, line 81: Does this differ across any operator handler implementations? If not, could it just be a local utility method in the operator code?

rjernst (approved):
This looks good at a high level. The three methods, name, transform and dependencies make sense. One general comment on the javadocs are that they should generally have a clear one sentence explanation about what the method does. Currently these methods all have large blocks of text that are a good explanation, but they are too detailed for the quick summary that should be the first, separate, line.

PR #87770, opened 2022-06-16
Ingest: move IngestDocument test ctor usage to static builder

stu-elastic:
`IngestDocument` has a test constructor that takes a `sourceAndMetadata` and `ingestMetadata` map.

However, that test constructor was also used from wire deserialization in `WriteableIngestDocument` 
and during `XContent` parsing.

This change moves test usage into test static builders and the wire usage into another static builder to
make the callers intent clear.

`IngestDocument`s constructed for test usage will not need metadata
validation performed on construction when that is added in a future commit.


martijnvg (approved):
Thanks @stu-elastic, LGTM.
The static methods make tests easier to understand.

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
```diff
@@ -83,15 +83,20 @@ public IngestDocument(IngestDocument other) {
...
-    public IngestDocument(Map<String, Object> sourceAndMetadata, Map<String, Object> ingestMetadata) {
+    IngestDocument(Map<String, Object> sourceAndMetadata, Map<String, Object> ingestMetadata) {
         this.sourceAndMetadata = sourceAndMetadata;
         this.ingestMetadata = ingestMetadata;
     }
 
+    /**
+     * Build an IngestDocument from values read via deserialization
+     */
+    public static IngestDocument fromWire(Map<String, Object> sourceAndMetadata, Map<String, Object> ingestMetadata) {
```
rjernst, line 96: I think the more typical naming would be `of`, so `IngestDocument.of(...)`. I would at least rename this method, and probably the test methods as well, eg `TestIngestDocument.ofSourceAndIngest`.

rjernst (approved):
LGTM, one minor thing on naming

PR #87773, opened 2022-06-17
Add setting for keep-alive duration for oidc back-channel

### LINKED ISSUE #75515: Custom keep-alive strategy for outgoing OIDC connections
We have been seeing regular connection timeouts in certain cases when our stack is being used as an OIDC RP, with specific OP OIDC implementations/vendors. The problem manifests as authentication timeouts for end users. 

The summary of what we have figured out happens is as follows: 

- Our implementation (Apache HttpClient `CloseableAsyncClient` ) doesn't set a HTTP keep-alive duration. This is presumably because we are using the `org.apache.http.impl.client.DefaultConnectionKeepAliveStrategy` which doesn't have a default value and the OP does not return a Keep-Alive HTTP response header in its responses to our requests to the OIDC endpoints.
- Something ( a host, a LB, the target server itself ) between us and the vendor/implementation drops the connection at some point but doesn't notify us (sending a packet with `FIN` or `RST` flag, for instance )
- Our implementation is not aware that the connection is dropped and since it assumes keep-alive duration is indefinitely, it attempts to reuse the same socket.
- Since the new connection attempt is not preambled by a `TCP` packet with `SYN` flag, the packet is dropped somewhere between us and the target server as malicious/wrong.  
- Our implementation doesn't get a response and times out on the `HTTP` layer. Presumably, if we set the `HTTP` connection timeout to a very large value, we will hit the underlying `TCP` connection timeout.

We could expose a setting for the OIDC realm that sets the default minimum keep-alive (by implementing an ConnectionKeepAliveStrategy ) even/only if the server doesn't set a preference in the HTTP response. 
This wouldn't fix the cause of the described issue but it could allow administrators to fine-tune the configuration to alleviate its effects.

ywangd:

the timeout issue.

The new setting has a 3min default which means idle connections are
closed every 3 min if server response does not specify a shorter keep-alive.

Resolves: #75515


ywangd:
@tvernum This might be worth to backport to earlier versions (including 7.17?) if it proves to work. For now, I just tagged `v8.4.0`.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java
```diff
@@ -707,7 +710,15 @@ private CloseableHttpAsyncClient createHttpClient() {
...
                     .setConnectionManager(connectionManager)
-                    .setDefaultRequestConfig(requestConfig);
+                    .setDefaultRequestConfig(requestConfig)
+                    .setKeepAliveStrategy((response, context) -> {
+                        final TimeValue timeValue = realmConfig.getSetting(HTTP_CONNECTION_POOL_TTL);
+                        if (TimeValue.MINUS_ONE.equals(timeValue)) {
+                            return DefaultConnectionKeepAliveStrategy.INSTANCE.getKeepAliveDuration(response, context);
+                        } else {
+                            return timeValue.millis();
+                        }
```
ywangd, line 720: Given the know situation, I think it is safer to always honour the setting value first because the server may not be aware there are proxies in between.

docs/changelog/87773.yaml
```diff
@@ -0,0 +1,5 @@
+pr: 87773
+summary: Add setting for keep-alive duration for oidc back-channel
```
tvernum, line 2: Since we're now enabling this by default, I think a better summary would be
```suggestion
summary: Automatically close idle connections in OIDC back-channel
```

docs/reference/settings/security-settings.asciidoc
```diff
@@ -1858,6 +1858,21 @@ connections allowed per endpoint.
...
+Controls the behavior of the http client used for back-channel communication to
+the OpenID Connect Provider endpoints. Specifies the time-to-live of connections
+in the connection pool. If a connection is not re-used within this timeout,
+it is closed.
+By default, the time-to-live is 3 minutes meaning that connections are closed
+after being idle for 3 minutes.
+
+Configure this setting to `-1` to let the server dictates this value using the `Keep-Alive` HTTP
+response header. If the header is not set by the server, the time-to-live is infinite meaning
+that connections never expire.
```
tvernum, line 1873: I think we should reword this, but I'll come back with a suggestion later.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/oidc/OpenIdConnectRealmSettings.java
```diff
@@ -212,6 +213,12 @@ private OpenIdConnectRealmSettings() {}
         "http.max_endpoint_connections",
         key -> Setting.intSetting(key, 200, Setting.Property.NodeScope)
     );
+
+    public static final Setting.AffixSetting<TimeValue> HTTP_CONNECTION_POOL_TTL = Setting.affixKeySetting(
+        RealmSettings.realmSettingPrefix(TYPE),
+        "http.connection_pool_ttl",
```
tvernum, line 219: I wonder whether this should be
```suggestion
        "http.connection_pool.ttl",
```

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java
```diff
@@ -705,9 +708,17 @@ private CloseableHttpAsyncClient createHttpClient() {
...
                 HttpAsyncClientBuilder httpAsyncClientBuilder = HttpAsyncClients.custom()
                     .setConnectionManager(connectionManager)
-                    .setDefaultRequestConfig(requestConfig);
+                    .setDefaultRequestConfig(requestConfig)
+                    .setKeepAliveStrategy((response, context) -> {
+                        if (TimeValue.MINUS_ONE.equals(connectionTtl)) {
+                            return DefaultConnectionKeepAliveStrategy.INSTANCE.getKeepAliveDuration(response, context);
+                        } else {
+                            return connectionTtl.millis();
+                        }
```
tvernum, line 720: Should we completely ignore the server's keep alive if a TTL is configured?
I wonder if we want:
```suggestion
                        var serverKeepAlive = DefaultConnectionKeepAliveStrategy.INSTANCE.getKeepAliveDuration(response, context);
                        if (TimeValue.MINUS_ONE.equals(connectionTtl)) {
                            return serverKeepAlive
                        } else if (serverKeepAlive == -1) {
                            return connectionTtl.millis();
                        } else {
                            return Math.min(serverKeepAlive, connectionTtl.millis());
                        }
```

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/oidc/OpenIdConnectRealmSettings.java
tvernum: Or even
```suggestion
        "http.connection_pool.idle_timeout",
```

ywangd: This name is taken after the similar [watcher setting](https://www.elastic.co/guide/en/elasticsearch/reference/current/notification-settings.html#general-notification-settings). It's probably better to keep consistent?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java
ywangd: Taking the minimum of the two makes sense. Thanks

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java
```diff
@@ -705,9 +707,23 @@ private CloseableHttpAsyncClient createHttpClient() {
...
+                        final long actualKeepAlive;
+                        if (serverKeepAlive == -1) {
+                            actualKeepAlive = userConfiguredKeepAlive;
+                        } else if (userConfiguredKeepAlive == -1) {
+                            actualKeepAlive = serverKeepAlive;
+                        } else {
+                            actualKeepAlive = Math.min(serverKeepAlive, userConfiguredKeepAlive);
+                        }
+                        LOGGER.debug("effective HTTP connection keep-alive: [{}]ms", actualKeepAlive);
+                        return actualKeepAlive;
```
ywangd, line 725: Tweaked this part a bit more and added a debug logging.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/oidc/OpenIdConnectRealmSettings.java
tvernum: Fair enough

docs/reference/settings/security-settings.asciidoc
ywangd: I reword'd this paragraph and also made it consistent to the new behaviour of handling keep-alive response header.

docs/reference/settings/security-settings.asciidoc
```diff
@@ -1858,6 +1858,21 @@ connections allowed per endpoint.
...
+`http.connection_pool_ttl` {ess-icon}::
+(<<static-cluster-setting,Static>>)
+Controls the behavior of the http client used for back-channel communication to
+the OpenID Connect Provider endpoints. Specifies the time-to-live of connections
+in the connection pool (default to 3 minutes). A connection is closed if it is
+idle for more than the specified timeout.
+
+The server can also set the `Keep-Alive` HTTP response header. The effective
+time-to-live value is the smaller value between this setting and the `Keep-Alive`
+reponse header. Configure this setting to `-1` to let the server dictates the value.
```
tvernum, line 1871:
```suggestion
reponse header. Configure this setting to `-1` to let the server dictate the value.
```

docs/reference/settings/security-settings.asciidoc
```diff
@@ -1858,6 +1858,21 @@ connections allowed per endpoint.
...
+If the header is not set by the server and the setting has value of `-1`,
+the time-to-live is infinite meaning that connections never expire.
```
tvernum, line 1873:
```suggestion
the time-to-live is infinite and connections never expire.
```

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java
```diff
@@ -705,9 +707,23 @@ private CloseableHttpAsyncClient createHttpClient() {
...
+                    });
```
tvernum, line 726: I think we need to come up with a test strategy for this. 
We could do it with a mock HTTP Server and ensure that the connection is pulled down, but that seems like a lot of work.
Is it possible to check the value configured on the client itself?

tvernum (commented):
Everything that's here LGTM, but we do need a plan for automated testing.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java
ywangd: I came up a way for testing the changes for both the configured TTL and the actual behaviour of "not reusing the same connection". Please see https://github.com/elastic/elasticsearch/pull/87773/commits/058872f1fd3f78f383f1517ac4917ed1a007330f

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticatorTests.java
```diff
@@ -1006,6 +1020,85 @@ public void testLogIdTokenAndNonce() throws URISyntaxException, BadJOSEException
...
+                if (false == firstClientPort.compareAndSet(null, currentPort)) {
+                    assertThat(currentPort, not(equalTo(firstClientPort.get())));
+                }
+                final byte[] bytes = randomByteArrayOfLength(2);
+                exchange.sendResponseHeaders(200, bytes.length);
+                exchange.getResponseBody().write(bytes);
+            } finally {
+                exchange.close();
+            }
+        });
```
ywangd, line 1043: For testing the behaviour that two requests with interval greater than the TTL (1s) should use different connections.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticatorTests.java
```diff
@@ -1006,6 +1020,85 @@ public void testLogIdTokenAndNonce() throws URISyntaxException, BadJOSEException
...
+        Loggers.setLevel(logger, Level.DEBUG);
+        try {
+            appender.addExpectation(
+                new MockLogAppender.PatternSeenEventExpectation(
+                    "log",
+                    logger.getName(),
+                    Level.DEBUG,
+                    ".*Connection .* can be kept alive for 1.0 seconds"
+                )
+            );
```
ywangd, line 1069: Testing for the configured TTL in action.

tvernum (approved):

PR #87774, opened 2022-06-17
Move logging config assertion to server cli

rjernst:
The log4j configuration file is shipped with all ES distributions. We
also check recursively for files possibly added by plugins. If no files
are found, we give a helpful startup error message. However, since the
log4j2 configuration file shipped with ES should always exist, we can
check upfront in the cli before even initializing logging.

This commit moves the validation of an existing log4j2 properties file
to the server cli.


ChrisHegarty (approved):
LGTM

PR #87775, opened 2022-06-17
Remove legacy bootstrap plugins

rjernst:
Bootstrap plugins were an internal mechanism added to allow a
filesystemprovider for cloud with the quota-aware-fs plugin. Since that
was removed, bootstrap plugins no longer serve a purpose. They were
never officially documented because they were for internal use only.
This commit removes the bootstrap plugins infrastructure.


rjernst:
Note that I have explicitly *not* marked this as breaking, because we were very explicit in never documenting this internal capability of plugins.

pugnascotia (commented):
The changes look good but there's a number of CI failures to address.

rjernst:
@pugnascotia The remaining CI failure is expected right now. The example plugin build is broken due to unrelated reasons.

pugnascotia (approved):

pugnascotia (approved):
Still LGTM

rjernst:
@elasticmachine update branch

PR #82975, opened 2022-01-24
Desired nodes API

fcofdez:
This commit adds the Desired Nodes API, allowing orchestrators
that manage Elasticsearch clusters to let the system know about the 
current/planned topology that the cluster will run on. 
This allows the system to take better decisions based on the entire 
cluster topology, including nodes that will be added/removed in the 
near future.

This commit adds the basic endpoints to manage the desired nodes
state:
- GET /_internal/desired_nodes
- PUT /_internal/desired_nodes/<history_id>/<version>
- DELETE /_internal/desired_nodes

PR #87778, opened 2022-06-17
Make Desired Nodes API operator-only

### LINKED ISSUE #87777: Make desired node operator only
The new Desired Nodes API was introduced it #82975. We need to make it [operator-only](https://www.elastic.co/guide/en/elasticsearch/reference/8.3/operator-only-functionality.html#operator-only-functionality).

arteam:
Resolves #87777.


arteam:
@elasticmachine update branch

fcofdez (commented):
Thanks Artem! I was wondering if we should write a test to ensure that the authorization policy is enforced.

arteam:
Good point! Let me see if can write an intergration test for this use case.

fcofdez (approved):
LGTM, thanks Artem!

arteam:
Thanks Francisco!

fcofdez:
@arteam I've just noticed that we didn't update the documentation with this change, would you mind opening a PR updating it? thanks!

arteam:
Sure, let me take care of this.

arteam:
Hey @fcofdez. It looks likes you already updated the docs part that lists the operator only APIs: https://github.com/elastic/elasticsearch/pull/82975/files#diff-7348c90a6eb98211295dafd41c39e24fd03f97518dce7c53cd1d8bec814934f1R25

Do you want to see some other mentions of that in the docs?

fcofdez:
You're right 🤦, sorry for the noise and thanks for looking into this!

PR #87786, opened 2022-06-17
Remove dead code in rest and http packages

original-brownbear:
Removing some more unused and needlessly verbose things here,
this code is complicated enough as it is already.

### DIFF
--- a/server/src/main/java/org/elasticsearch/http/HttpHandlingSettings.java
+++ b/server/src/main/java/org/elasticsearch/http/HttpHandlingSettings.java
@@ -10,7 +10,6 @@ package org.elasticsearch.http;
 
 import org.elasticsearch.common.settings.Settings;
 
-import static org.elasticsearch.http.HttpTransportSettings.SETTING_CORS_ENABLED;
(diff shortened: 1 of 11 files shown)


idegtiarenko (approved):

server/src/main/java/org/elasticsearch/http/HttpHandlingSettings.java
```diff
@@ -71,54 +39,7 @@ public static HttpHandlingSettings fromSettings(Settings settings) {
             SETTING_HTTP_RESET_COOKIES.get(settings),
             SETTING_HTTP_COMPRESSION.get(settings),
             SETTING_HTTP_COMPRESSION_LEVEL.get(settings),
-            SETTING_HTTP_DETAILED_ERRORS_ENABLED.get(settings),
-            SETTING_PIPELINING_MAX_EVENTS.get(settings),
```
pugnascotia, line 75: Why are these removed?

original-brownbear: These aren't used from this object, we load them into fields on the actual http transport and use them there.

pugnascotia (approved):

PR #87787, opened 2022-06-17
Bound ack time

idegtiarenko:
Currently, our default wait time for ack is unbound. This could result
in an entire test suite timeout rather than individual test cases failures.

We should also consider merging this to master.

### DIFF
--- a/test/framework/src/main/java/org/elasticsearch/test/hamcrest/ElasticsearchAssertions.java
+++ b/test/framework/src/main/java/org/elasticsearch/test/hamcrest/ElasticsearchAssertions.java
@@ -17,7 +17,6 @@ import org.elasticsearch.action.ActionRequestBuilder;
 import org.elasticsearch.action.admin.cluster.health.ClusterHealthRequestBuilder;
 import org.elasticsearch.action.admin.cluster.health.ClusterHealthResponse;
 import org.elasticsearch.action.admin.indices.create.CreateIndexResponse;
-import org.elasticsearch.action.admin.indices.delete.DeleteIndexRequestBuilder;
 import org.elasticsearch.action.admin.indices.template.get.GetIndexTemplatesResponse;
 import org.elasticsearch.action.bulk.BulkResponse;
 import org.elasticsearch.action.get.GetResponse;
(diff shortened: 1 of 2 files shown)


test/framework/src/main/java/org/elasticsearch/test/hamcrest/ElasticsearchAssertions.java
```diff
@@ -100,10 +103,6 @@ public static void assertAcked(AcknowledgedResponse response) {
         assertThat(response.getClass().getSimpleName() + " failed - not acked", response.isAcknowledged(), equalTo(true));
     }
 
-    public static void assertAcked(DeleteIndexRequestBuilder builder) {
-        assertAcked(builder.get());
-    }
```
idegtiarenko, line 105: This overloaded variant is covered by ones above and is actually not needed

DaveCTurner:
Seems reasonable, or at least it's in line with our preference for failing tests more quickly than the suite timeout. I'm not sure 30s is a good value tho, there might be some tests that need a longer wait which this would cause to fail. Could we make this change in `master` first so it gets more CI exposure?

x-pack/plugin/autoscaling/src/internalClusterTest/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageIT.java
```diff
@@ -178,7 +178,7 @@ public void testScaleFromEmptyLegacy() {
                     .put(INDEX_STORE_STATS_REFRESH_INTERVAL_SETTING.getKey(), "0ms")
                     .put(IndexMetadata.INDEX_ROUTING_INCLUDE_GROUP_SETTING.getKey() + "data_tier", "hot")
                     .build()
-            )
+            ).setWaitForActiveShards(ActiveShardCount.NONE)
```
idegtiarenko, line 181: This was waiting 30 seconds for a shards that could not be allocated. Following call updates settings for the correct allocation anyways...

DaveCTurner (approved):
LGTM

PR #87789, opened 2022-06-17
Modularize the lang-mustache component

ChrisHegarty:
This change modularizes the lang-mustache component, by adding a module-info.java

relates #78744

### DIFF
--- a/modules/lang-mustache/src/main/java/module-info.java
+++ b/modules/lang-mustache/src/main/java/module-info.java
@@ -13,5 +13,5 @@ module org.elasticsearch.mustache {
     requires org.apache.logging.log4j;
     requires com.github.mustachejava;
 
-    exports org.elasticsearch.script.mustache;
+    exports org.elasticsearch.script.mustache; // TODO: remove once the HLRC is removed
 }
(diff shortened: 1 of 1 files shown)


modules/lang-mustache/src/main/java/module-info.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+module org.elasticsearch.mustache {
+    requires org.elasticsearch.base;
+    requires org.elasticsearch.server;
+    requires org.elasticsearch.xcontent;
+    requires org.apache.logging.log4j;
+    requires com.github.mustachejava;
+
+    exports org.elasticsearch.script.mustache;
```
rjernst, line 16: Where is the export needed? Is this only needed by server for reflective construction?

rjernst (approved):
Looks good, one minor question.

modules/lang-mustache/src/main/java/module-info.java
ChrisHegarty: The HLRC uses types from `org.elasticsearch.script.mustache`, see `org.elasticsearch.client.RequestConverters`.

rjernst: Ok, perhaps add a TODO to remove this once HLRC is removed (it's dead code, just used by tests now)

PR #87794, opened 2022-06-17
Modularize the kibana, mapper-extras, parent-join, and systemd, components

ChrisHegarty:
This change modularizes the kibana, mapper-extras, parent-join, and systemd, components, by adding a module-info.java

These components have been group together since their changes are quite straightforward.

relates #78744


rjernst (approved):
LGTM

PR #87796, opened 2022-06-17
Remove HLRC impl of PinnedQueryBuilder

romseygeek:
We have an extra implementation of PinnedQueryBuilder that was added to 
expose it to the high-level rest client.  This is no longer needed, so we can
safely remove it.


ywelsch (approved):
LGTM

PR #87799, opened 2022-06-17
[DOCS]: replace `instance_has_master` with `master_is_stable`

andreidan:


docs/reference/health/health.asciidoc
```diff
@@ -81,9 +81,12 @@ also optionally provide URLs to troubleshooting documentation.
     components are:
 +
 --
-  `instance_has_master`::
+  `master_is_stable`::
       (Component `cluster_coordination`) Reports health issues regarding
-      connections to the master node.
+      the stability of the node that's seen as the master by the node handling
+      the health request. In case of enough subsequent observed master changes
```
masseyke, line 87: I don't think the word `subsequent` is needed here, and it makes the sentence confusing -- subsequent to what?

jbaiera (approved):
LGTM, one small nit for readability

docs/reference/health/health.asciidoc
andreidan: Would `consecutive` be better? (there's a time element to it)
If we remove `subsequent` it becomes 
```
In case of enough observed master changes
```
which could be interpreted as changes during the lifetime of a deployment (?)

docs/reference/health/health.asciidoc
```diff
@@ -81,9 +81,12 @@ also optionally provide URLs to troubleshooting documentation.
     components are:
 +
 --
-  `instance_has_master`::
+  `master_is_stable`::
       (Component `cluster_coordination`) Reports health issues regarding
-      connections to the master node.
+      the stability of the node that is seen as the master by the node handling
+      the health request. In case of enough subsequent observed master changes
```
andreidan, line 87:
```suggestion
      the health request. In case of enough observed master changes in a short period of time
```

PR #87800, opened 2022-06-17
Upgrade GCS Plugin to 1.118.1

original-brownbear:
Staying up to date.


henningandersen (approved):
LGTM, provided CI agrees.

original-brownbear:
Yea moving to version 2 is going to be trickier and we'll have to evaluate what it means for performance+stability first before we set out on that project.

PR #87000, opened 2022-05-20
Temporarily skip JarHellTests::testNonJDKModuleURLs on JDK 19

ChrisHegarty:
Temporarily skip JarHellTests::testNonJDKModuleURLs on JDK 19, until this issue is resolved https://bugs.openjdk.java.net/browse/JDK-8287097

closes #86833

PR #87803, opened 2022-06-17
Re-enable several tests since JDK-8287097 has been fixed

ChrisHegarty:
With JDK-8287097 [1] now fixed in the latest JDK 19 build, b26, we can revert the test changes that were introduced to temporarily skip affected tests.

[1] https://bugs.openjdk.org/browse/JDK-8287097

relates #87000 #87019


rjernst (approved):
LGTM

PR #85503, opened 2022-03-30
[ML] rename trained model allocations to assignments

benwtrent:
This renames the internal concept of a trained model allocation into an assignment.

Now models are assigned to a node and routes created for inference. Not "allocated". 

This is an internal rename only. The user facing concepts of trained models and deployments are untouched.

PR #87806, opened 2022-06-17
[ML] Rename trained model metadata after full cluster upgrade

dimitris-athanasiou:
In #85503 we renamed trained model allocations to assignments.
However, it is important to wait for the full cluster to be
upgraded before we rewrite the cluster state so that there are
no undesired side-effects in a mixed cluster.

### DIFF
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
@@ -12,6 +12,7 @@ import org.apache.logging.log4j.Logger;
 import org.elasticsearch.ElasticsearchStatusException;
 import org.elasticsearch.ResourceAlreadyExistsException;
 import org.elasticsearch.ResourceNotFoundException;
+import org.elasticsearch.Version;
 import org.elasticsearch.action.ActionListener;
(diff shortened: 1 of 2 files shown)


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -245,18 +248,24 @@ public void clusterStateProcessed(ClusterState oldState, ClusterState newState)
...
     }
 
+    private static ClusterState forceUpdate(ClusterState currentState, TrainedModelAssignmentMetadata.Builder modelAssignments) {
+        Metadata.Builder metadata = Metadata.builder(currentState.metadata());
+        if (currentState.getNodes().getMinNodeVersion().onOrAfter(RENAME_ALLOCATION_TO_ASSIGNMENT_VERSION)) {
+            metadata.putCustom(TrainedModelAssignmentMetadata.NAME, modelAssignments.build())
+                .removeCustom(TrainedModelAssignmentMetadata.DEPRECATED_NAME);
+        } else {
+            metadata.putCustom(TrainedModelAssignmentMetadata.DEPRECATED_NAME, modelAssignments.build())
+                .removeCustom(TrainedModelAssignmentMetadata.NAME);
```
droberts195, line 264: Is this removal necessary?  Surely it's only needed if we've already released software that prematurely upgrades?

droberts195 (approved):
LGTM

Issue #52482, opened 2020-02-18
Introduce first query


mayya-sharipova:
This is related to #42811 – deprecation of the function_score query.
`funciton_score` query had a unique functionality –  `first`, that is not present in the current query DSL.

A proposal  is to implement this functionality through a dedicated query type.  
This query evaluates the child queries in order, and for a document returns the score of the first matching query.  
```js
{
  "query": {
    "first": {
      "queries": [
        {
          "match": {"message": "elasticsearch"}
        },
        {
         "match": {"author": "kibana"}
        }
      ]
    }
  }
}
```
Other name proposals:
- `first_in_order`
- `first_matching`

mayya-sharipova:
Relates to https://github.com/elastic/elasticsearch/issues/49761 but addresses in for more general use cases where filters can't be expressed through script.

Issue #51967, opened 2020-02-05
Compound query


mayya-sharipova:
This is related to #42811 -- deprecating function_score query

A proposal is to have a new type of  compound query that has an ability to combine scores from different queries:

```js
{
  "query": {
    "compound": {
      "queries": [
        {
          "match": {"message": "elasticsearch"}
        },
        {
         "match": {"author": "shay banon"}
        }
      ],
      "combine_mode": "sum"
    }
  }
}
```

combine_mode has the same values and definitions as in [Function Score Query's score_mode](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html)

combine_mode | definition
-- | --
sum | scores are summed (default)
multiply | scores are multiplied 
avg | scores are averaged
first | the first function that has a matching filter is applied
max | maximum score is used
min | minimum score is used

----
The intention for this query is to allow custom scoring logic, so internal queries are primarily expected to be `script_score` queries. For example, the intention fo the following query is to find "best fit" candidate with necessary skills: 
```js
{
  "query": {
    "compound": {
      "queries": [
        {
          "script_score": {
            "query": {
              "match": {
                "skills": "Java"
              }
            },
            "script": { "source": "10.0"}
          }
        },
        {
          "script_score": {
            "query": {
              "match": {
                "skills": "Scala"
              }
            },
            "script": {"source": "5.0"}
          }
        },
        {
          "script_score": {
            "query": {
              "match": {
                "skills": "Python"
              }
            },
            "script": {"source": "1.0"}
          }
        }
      ],
      "combine_mode": "first"
    }
  }
}
```

mayya-sharipova:
Closing this issue, as it is replaced by #52482

jayaddison:
Hi @mayya-sharipova - closing of the linked issue reminded me of this, and I noticed that there was an implementation in-progress at one point (#52387).  Are there any plans to introduce something like a `compound` query in ES in future?  (it's ok either way, I'm just curious)

mayya-sharipova:
@jayaddison Thank you for your interest. We did not pursue this work till the end, as we had some other priorities to work on. We will try to re-discuss this topic and get back to it.

jayaddison:
Thanks @mayya-sharipova, that makes sense.

[...]

Issue #42811, opened 2019-06-03
Deprecate Function Score Query in favour of Script Score Query


mayya-sharipova:
We would like to deprecate Function Score Query in ~~7.x~~ 8.0 and completely remove it starting from ~~8.0~~ 9.0 (increasing versions as we are far deep in 7.x already).

The new [script score query](https://www.elastic.co/guide/en/elasticsearch/reference/master/query-dsl-script-score-query.html) will replace it.
[Here](https://www.elastic.co/guide/en/elasticsearch/reference/master/query-dsl-script-score-query.html#_transition_from_function_score_query) we describe how a function score query can be replaced with a script score query.

We would like to know from the community about cases when something was possible to do with Function Score Query, but the new Script Score Query can't address it.

mayya-sharipova:
We have identified 2 `function_score` query functionalities that are missing  in the `script_score` query 

1.  `score_mode` –`first`, applying the score from the function with the 1st matching filter (wonder how prevalent is this use-case)
2. `_explain`  explanation of score calculation doesn't work in script_score query.

mayya-sharipova:
For score_mode `first`, we had a [discussion](https://github.com/elastic/elasticsearch/issues/27588#issuecomment-444887726) before.

A proposal is to have a new type of  compound query that has an ability to combine scores from different queries:

```js
{
  "query": {
    "compound": {
      "queries": [
        "match": {
          "message": "elasticsearch"
        },
        "match": {
          "author": "kibana"
        }
      ],
      "score_mode": "first"
    }
  }
}
```

score_mode has the same values and definitions as in [Function Score Query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html)

score_mode | definition
-- | --
sum | scores are summed (default)
multiply | scores are multiplied 
avg | scores are averaged
first | the first function that has a matching filter is applied
max | maximum score is used
min | minimum score is used

mayya-sharipova:
Or as an alternative we can have a script compound query where scores can be freely combined using script:

```js
{
  "query": {
    "script_compound": {
      "queries": [
        "match": {
          "message": "elasticsearch"
        },
        "match": {
          "author": "shay banon"
        }
      ],
      "script": {
        "source": "_scores[0] + _scores[1]"
      }
    }
  }
}
```

But for the `first` mode, we would still need to implement a type of `in-order` query, as it is difficult to implement this logic through a script. A possible API for `in_order` query:

```js
{
[...]

Issue #27588, opened 2017-11-29
Replacing the function_score with discrete queries


polyfractal:
## Overview
The `function_score` is a powerful query, but can be somewhat unwieldy and difficult to use.  It is a monolithic query that has many parameters and options which makes it difficult for new users to learn.  It can be difficult to tweak, and now that we've moved to BM25, the defaults don't work well (namely, multiplying scores).

We'd like to deprecate `function_score` and replace it's functionality with a number of discrete queries.  The guiding principle is that these replacement queries will be small, simple and single-purpose.  Individual queries should be simpler to implement and maintain, easier for users to learn and should allow the same functionality as `function_score` by mixing/matching as required.  

#### Arbitrary numeric functions 
This query will allow working with arbitrary numerics and allow applying various mathematical functions.  For example, a document may have a `popularity` field that you wish to roll into the score somehow.

Functions should include those of `field_value_factor`, like logarithm, sqrt, reciprocal, etc.  We'd also like to include a sigmoid and rational function.  The query should also be able to return just the value itself.  I think the random number functionality of `function_score` can be rolled into this too.

We could potentially implement this query entirely through scripting, assuming we can provide custom functions through the script context for more complicated operations like sigmoid.

#### Distance: Numeric and Time
This query would provide a set of decay functions that allow you to score the "distance" from a field value to some point.  When dealing with dates this "distance" would essentially represent recency, while with numerics it'd just be a geometric distance.

We'll likely want numeric/time to be grouped together since the operations are essentially the same.

#### Distance: Geo
This would be similar to numeric/time distance, except operating on geo points and physical distance.  The thinking is that geo would be separate, as we may have plans in the future for geo querying to become more robust in general.  Geo points are also sufficiently different that the syntax will likely need to be a bit different. 

Otherwise the functionality is similar, providing a set of decay functions.

## Potential queries

Some other potential queries that we kicked around, unsure if they are needed or quite how the functionality would work.  Some of these have a direct comparison in function_score, some are only tangentially related.

#### Min/Max query
Function score allows taking the min or max score from the set of functions.  We could potentially add a combination query that executes the children, then only passes on the min/max score from the children.

#### Bandpass query / cutoff score query
Function score allows setting max thresholds on scores generated by the functions.  It may be useful to have a query that allows setting min or max or both, and the scores that come out of the query would be limited to those values.  Similar to a constant score in that it wraps a set of queries, but instead of setting a single score it just limits the scores that are generated to the range.

[...]

PR #87807, opened 2022-06-17
Undeprecate function_score query

### LINKED ISSUE #71934: [DOCS] Is function_score being deprecated?
The `script_score` docs say that function score is being deprecated, but that section was last changed two years ago.
See "Transition from function_score" section https://github.com/elastic/elasticsearch/blob/master/docs/reference/query-dsl/script-score-query.asciidoc

The `function_score` docs say nothing about deprecation/superseding by `script_score`.

This was/is confusing to me. Maybe deprecation is no longer planned or another good reason, but I think the docs for those queries should be in sync.

mayya-sharipova:
We had a plan to deprecate function_score query with
script_score query, but ran into a roadblock of missing
functionality to combine scores from different
functions (particularly "first" script_score).
Wee have several proposal to address this missing
functionality:
 [scripted_boolean](https://github.com/elastic/elasticsearch/issues/27588#issuecomment-444887726)
 [compound_query](https://github.com/elastic/elasticsearch/issues/51967)
 [first_query](https://github.com/elastic/elasticsearch/issues/52482)

But for now, we decided not to deprecate function_score query,
and hence we need to remove any mention that we are deprecating it.

Relates to #42811
Closes #71934

### DIFF
--- a/docs/reference/query-dsl/script-score-query.asciidoc
+++ b/docs/reference/query-dsl/script-score-query.asciidoc
@@ -241,10 +241,11 @@ can efficiently skip non-competitive hits:
 
 [[script-score-function-score-transition]]
 ===== Transition from the function score query
-We are deprecating the <<query-dsl-function-score-query, `function_score`>>
-query. We recommend using the `script_score` query instead.
+We recommend using the `script_score` query instead of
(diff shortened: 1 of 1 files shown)


javanna (approved):
LGTM thanks!

PR #87809, opened 2022-06-17
Simplify bootstrap error logging

rjernst:
During startup of Elasticsearch we go to great lengths to present errors
in a meaningful way to users. Over time, though, the error handling has
been amended to address various issues, and that has resulted in a
complicated system of try/catches handling various cases. One
particularly kludgy piece is removing the console logger in special
cases to avoid printing exceptions to the console. Additionally, the
console removal wasn't actually effective because later in exception
handling the exception would be both logged anyways, and then also sent
to stderr, meaning that we could see the same exception several times.

This commit reworks how exceptions are logged during bootstrap. To
address the concern of printing full exceptions to the console, a new
log4j exception filter is added to the console appender which will only
print out the exception message and some additional explanatory info. To
address logging multiple times, the try/catch within init is removed so
that excpetions can propagate to the try/catch in main, which now
handles all exceptions. Additionally, phase 1 (before logging) handles
it's own failure cases since there is definitely no logging at that
time. This simplifies the other failure cases latere so that they do not
need to check if logging has been initialized through sysprops.

### DIFF
--- a/server/src/main/java/org/elasticsearch/bootstrap/Elasticsearch.java
+++ b/server/src/main/java/org/elasticsearch/bootstrap/Elasticsearch.java
@@ -42,8 +42,9 @@ class Elasticsearch {
 
         PrintStream out = getStdout();
         PrintStream err = getStderr();
+        final ServerArgs serverArgs = initPhase1(err);
+
         try {
-            final ServerArgs serverArgs = initPhase1();
             initPidFile(serverArgs.pidFile());
             Bootstrap.init(
                 serverArgs.daemonize() == false,
@@ -64,8 +65,8 @@ class Elasticsearch {
             exitWithUserException(err, ExitCodes.CONFIG, e);
         } catch (UserException e) {
             exitWithUserException(err, e.exitCode, e);
-        } catch (Exception e) {
-            exitWithUnknownException(err, e);
+        } catch (Throwable t) {
+            exitWithUnknownException(err, t);
         }
     }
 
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/elasticsearch/common/logging/ConsoleThrowablePatternConverter.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @param config  The current Configuration.
+     * @return instance of class.
+     */
+    public static ConsoleThrowablePatternConverter newInstance(final Configuration config, final String[] options) {
+        return new ConsoleThrowablePatternConverter(options, config);
+    }
+
+    @Override
+    public void format(final LogEvent event, final StringBuilder toAppendTo) {
```
pgomulka, line 41: the logic here seems to be very simple, but do you think it is worth a test?

server/src/main/java/org/elasticsearch/bootstrap/Elasticsearch.java
```diff
@@ -75,18 +76,18 @@ private static void exitWithUserException(PrintStream err, int exitCode, Excepti
...
-            // this is a horrible hack to see if logging has been initialized
-            // we need to find a better way!
-            Logger logger = LogManager.getLogger(Elasticsearch.class);
-            logger.error("fatal exception while booting Elasticsearch", e);
-        }
-        // format exceptions to the console in a special way to avoid 2MB stacktraces from guice, etc.
-        StartupException.printStackTrace(e, err);
+    private static void exitWithUnknownException(PrintStream err, Throwable e) {
+        Logger logger = LogManager.getLogger(Elasticsearch.class);
+        logger.error("fatal exception while booting Elasticsearch", e);
```
pgomulka, line 81: so given we use %consoleException for console appender this will result with `Elasticsearch failed to startup normal...` in console.
However we will print full stacktrace in server logs now? Do we want this?

pgomulka: I am worried that this is harder to control. THe exception formatting in json logs is controlled by ecs-logging-lib and I don't see a way to limit the size of stactrace..
https://github.com/elastic/ecs-logging-java/blob/723a5572eb9f9f462605df2cf7a601571c4e6dee/ecs-logging-core/src/main/java/co/elastic/logging/EcsJsonSerializer.java#L239

pgomulka (changes requested):
I left comments, I think this all looks good.
However I wonder if we do not miss a feature in ecs-logging-java to allow for custom throwable formatting in order to limit the size of the stacktrace. I don't see how to do this now.
let me know what you think

pgomulka (approved):
LGTM

PR #87810, opened 2022-06-17
SQL: Make automatic type conversion consistent for comparisons

### LINKED ISSUE #77055: SQL: Comparison beetween datetime field and string behaves inconsistently when used in filter vs. projection, grouping and ordering
All kinds of comparisons between a datetime field and a string behave as expected in `WHERE` clauses but produce `null` in `SELECT`, `GROUP BY` and `ORDER BY` clauses.

A simple example is:

```
SELECT birth_date = '1955-10-04T00:00:00.000Z' a FROM test_emp WHERE a

       a
---------------
null
```

The query correctly produces one record but the `a` field should contain `true`.

Another example with `GROUP BY` and `<`:

```
SELECT COUNT(*) FROM test_emp GROUP BY birth_date < '1955-10-04T00:00:00.000Z'

   count(*)
---------------
100
```

Expected result contains two groups.

### LINKED ISSUE #87500: SQL: make automatic type conversion consistent
Currently, most of the SQL implementation relies on the usage of the Search API, that in many circumstances does automatic type conversion.
Eg. both the following queries

```
SELECT * FROM foo WHERE id = "1";
SELECT * FROM foo WHERE id = 1;
```

will return results where `id` property value is `1` (numeric).

This behavior should be preserved in all the SQL operations that are not pushed to Search, like during query optimization/folding, local execution and on operations translated to Painless scripts.

Particular attention has to be put on date math expressions (eg. "now-10") that have to be evaluated and then converted.

luigidellaquila:
### Problem to solve

Since SQL heavily relies on Search API, it also inherits its behavior regarding type conversion.
Eg. the following queries are equivalent, when `id` is a numeric field:

```
SELECT * FROM someIndex WHERE id = 1 ;
SELECT * FROM someIndex WHERE id = '1'; 
```

In the second query, the string `'1'` is automatically converted to the numeric `1`.
A slightly different version of this query, though, will return an unexpected result:

```
SELECT id, id = '1' as x FROM someIndex WHERE id = '1'; 

|  id  |   x    |
+------+--------+
|  1   | null   |
```

The same filter that is `true` as a condition, is `null` as a projection.
This is due to the fact that the projection is executed locally (ie. not pushed to Search).

### Scope of the PR

This PR extends automatic type conversion to all the **local comparison operations** (equalities, inequalities, BETWEEN, IN), **when the conversion is possible at execution planning** time (ie. when the expression to convert is **constant**, can be folded and converted during the query optimization phase).

As an additional point, String to Date conversion is particularly complicated, due to [date math](https://www.elastic.co/guide/en/elasticsearch/reference/current/sql-functions-datetime.html#_comparison). 
This PR also addresses this problem, enhancing the date parsing and preserving the logic of comparison with date math expressions (eg. `x = '2020-01-01||/y'` is true for any `x` in the year 2020).

This is enough to cover all the common use cases and to make the usability experience more natural (ie. uniform and consistent behavior regarding data conversion in comparisons)

#### Runtime conversion

This PR only covers data conversion for constant (ie. foldable) expressions. 
Some conversions can only be performed at execution time, eg. the following:

```
SELECT aDateField = aStringField FROM index
```

needs to convert each value of `aStringField` after it's returned by the search phase (assuming that type conversion is the desired behavior in this case).

#### Mathematic operations

Data conversion can also be attempted in mathematic operation like `1 + '1'`. This PR does not address this case

----

Fixes #87500
Fixes #77055


x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/optimizer/Optimizer.java
```diff
@@ -1266,6 +1284,241 @@ private boolean foldable(Expression e) {
...
+                                something IN (<daterange>, <somethingElse1>, <somethingElse2>)
+                                becomes
+                                something BETWEEN <min(daterange)> AND <max(daterange)> OR someting IN (<somethingElse1>, <somethingElse2>)
+                                */
+                                additionalConditions.add(
+                                    new Range(
+                                        in.source(),
+                                        in.value(),
+                                        tryCast(exp, range.v1(), type),
+                                        true,
```
bpintea, line 1345: Should the lower bound be included, if changed (i.e. rounded down)?

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/optimizer/Optimizer.java
```diff
@@ -1266,6 +1284,241 @@ private boolean foldable(Expression e) {
...
+            }
+
+            if (newRight.size() > 0 && additionalConditions.size() > 0) {
+                return new Or(
+                    in.source(),
+                    new In(in.source(), in.value(), newRight, in.zoneId()),
+                    toOrTree(in.source(), additionalConditions)
+                );
+            } else if (newRight.size() > 0) {
+                return new In(in.source(), in.value(), newRight, in.zoneId());
```
bpintea, line 1377: Could the IN have been modified here? I.e. shouldn't the original `in` be returned?

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/optimizer/Optimizer.java
```diff
@@ -1266,6 +1284,241 @@ private boolean foldable(Expression e) {
...
+            } else if (additionalConditions.size() > 0) {
+                return toOrTree(in.source(), additionalConditions);
```
bpintea, line 1379: If we can rid the Range, maybe we could have an IN here instead.

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/optimizer/Optimizer.java
```diff
@@ -1266,6 +1284,241 @@ private boolean foldable(Expression e) {
...
+            }
+
+            return in;
```
bpintea, line 1382: If we get here, would this mean there's an empty IN?

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/optimizer/Optimizer.java
```diff
@@ -1266,6 +1284,241 @@ private boolean foldable(Expression e) {
...
+        }
+
+        private Expression toOrTree(Source source, List<Expression> items) {
```
bpintea, line 1385: Could be made static.

bpintea (commented):
Made a partial pass, left a few questions.

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/optimizer/Optimizer.java
```diff
@@ -1266,6 +1284,241 @@ private boolean foldable(Expression e) {
...
+                                foldedValue,
+                                foldedValue,
+                                true,
+                                true,
+                                ZoneOffset.UTC,
+                                DateUtils.DATE_MATH_PARSER,
+                                DateFieldMapper.Resolution.MILLISECONDS,
+                                () -> System.currentTimeMillis()
+                            );
+                            if (range.v1() < range.v2()) {
```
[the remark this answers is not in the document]
luigidellaquila: They differ in all the date math expressions. Eg:
- for `2022-01-01`, range.v1() = `2022-01-01T00:00:00.000Z` and  range.v2() = `2022-01-01T23:59:59.999Z`
- for `2022-08-15||/y`, range.v1() = `2022-01-01T00:00:00.000Z` and  range.v2() = `2022-12-31T23:59:59.999Z`

I didn't add any explicit tests on these values (it's part of Server logic), but most of the spec tests on dates take these values into consideration.

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/optimizer/Optimizer.java
luigidellaquila: Yes, it would simplify things a lot, but unfortunately we cannot (see my previous comment)

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/optimizer/Optimizer.java
luigidellaquila: I'm preserving the semantics of BETWEEN operator here, ie. upper and lower bound are included.
In this case it's just a bit more convoluted, since I have to take into consideration that the date expression can be a date math, so including it all could mean including the range upper or lower bound, if the expression is the first or the second value of the BETWEEN

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/optimizer/Optimizer.java
luigidellaquila: A few lines before there is an explicit check:

```
            if (converted == false) {
                return in;
            }
```

If the method reached this point, it means there were some conversions (probably not date math conversions, otherwise there would be some `additionalConditions`), so the `newRight` list of values should be different from the original one.

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/optimizer/Optimizer.java
luigidellaquila: Good point, and yes, the only reason why the code could reach this point is an empty IN, but since I added an initial condition:

```
            if (in.list().isEmpty()) {
                return in;
            }
```

this should never happen.

I think I can just remove the initial check and add a comment here (having an empty IN is uncommon, but it could happen I guess).

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/optimizer/Optimizer.java
```diff
@@ -1266,6 +1284,241 @@ private boolean foldable(Expression e) {
...
+                        try {
+                            // try date math first
+                            Tuple<Long, Long> range = DateFieldMapper.DateFieldType.toTimestampRange(
+                                foldedValue,
+                                foldedValue,
+                                true,
+                                true,
+                                ZoneOffset.UTC,
+                                DateUtils.DATE_MATH_PARSER,
+                                DateFieldMapper.Resolution.MILLISECONDS,
```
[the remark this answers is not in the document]
luigidellaquila: Not sure, I'll investigate a bit

x-pack/plugin/sql/src/test/resources/org/elasticsearch/xpack/sql/planner/querytranslator_tests.txt
```diff
@@ -40,9 +40,26 @@ CONTAINS "term":{"int":{"value":5
 
 TermEqualityForDate
 SELECT some.string FROM test WHERE date = 5;
-"term":{"date":{"value":5
+"range":{"date":{"gte":"1970-01-01T00:00:00.005Z","lte":"1970-01-01T00:00:00.005Z"
 ;
 
+TermEqualityForDateWithConversion
```
bpintea, line 46: Would be nice to add some more variance in the tests (`now-1y` etc.)

bpintea (commented):
Left another round of comments.

luigidellaquila:

luigidellaquila:
@elasticmachine update branch

luigidellaquila:

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/optimizer/Optimizer.java
```diff
@@ -1266,6 +1284,241 @@ private boolean foldable(Expression e) {
...
+                                    predicate.source(),
+                                    predicate.left(),
+                                    tryCast(predicate.right(), range.v1(), DataTypes.DATETIME),
+                                    true,
+                                    tryCast(predicate.right(), range.v2(), DataTypes.DATETIME),
+                                    true,
+                                    null
+                                );
+                                return predicate instanceof NotEquals ? new Not(predicate.source(), newRange) : newRange;
+                            } else if (predicate instanceof GreaterThan || predicate instanceof LessThanOrEqual) {
```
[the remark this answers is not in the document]
luigidellaquila: Actually a query like this works just fine

```
select * from person where birthdate >= 'now-10y'
```

and is translated as 

```
"query": {
        "range": {
            "birthdate": {
                "gte": "now-10y",
                "time_zone": "Z",
                "boost": 1.0
            }
        }
    }
```

I tried to reproduce the exact same behavior with this fix, but maybe I'm missing something. Can you please elaborate further?

x-pack/plugin/sql/src/test/resources/org/elasticsearch/xpack/sql/planner/querytranslator_tests.txt
luigidellaquila: I couldn't find a way to test "now" expressions in this context, since the expected result would change at every execution and I think there is no way to make it dynamic.
Any suggestion is welcome of course.

astefan (commented):
I've looked over this PR, not necessarily for how correct it is, but more for what kind of use cases it tries to address, as I think this is a very important aspect. The description of the PR mentions what scenarios it fixes, but it tries to list those that it doesn't and I think there are some things missing from there:
- a `version` type of field is omitted from these implicit conversions on purpose in code
- if numbers and dates implicit conversions are possible, there is a limitation in ES SQL that prevents us to support **all** types of dates conversions. If the user defines a field with a custom date format  there are high chances the implicit conversion won't work since in the PR the format is hardcoded. ES SQL uses `field_caps` to discover the fields and this API doesn't expose the custom date format.
- not all boolean values are implicitly converted
```
SELECT b='1',b='0',b=2,b=1,b=0,b=true,b='true' from test1 where b is not null

     b='1'     |     b='0'     |      b=2      |      b=1      |      b=0      |    b=true     |   b='true'    
---------------+---------------+---------------+---------------+---------------+---------------+---------------
null           |null           |true           |true           |false          |true           |true           
null           |null           |false          |false          |true           |false          |false          
```

Maybe there are other situations, not sure.

In conclusion, I like the idea of improving a bit the user experience in some cases, but these scenarios don't seem to be that many to be worth it. Especially when the PR changes some fundamentals in `between`, `in` and `range` queries.

PR #87811, opened 2022-06-17
Fix ignored shards detection

idegtiarenko:
When created index could not be assigned has a replica it has following shard assignment:
`ShardAssignment[nodeIds=[], unassigned=2, ignored=1]`

Above record means that 2 shards (primary and replica) are unassigned, but only 1 is ignored (replica).
It turns out that replicas are ignored when corresponding primaries are not assigned. This should be reflected in the check.

This issue is reproduced by: `SimpleIndexStateIT. testFastCloseAfterCreateContinuesCreateAfterOpen`.


server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/ShardAssignment.java
```diff
@@ -15,13 +15,18 @@
...
 import static java.util.stream.Collectors.toUnmodifiableSet;
 
-public record ShardAssignment(Set<String> nodeIds, int unassigned, int ignored) {
+public record ShardAssignment(Set<String> nodeIds, int total, int unassigned, int ignored) {
 
-    public boolean isIgnored() {
-        return ignored > 0 && nodeIds.isEmpty();
+    public ShardAssignment {
+        assert total > 0 : "Shard assignment should not be empty";
+        assert nodeIds.size() + unassigned == total : "Shard assignment should account for all shards";
```
DaveCTurner, line 22: Can we also assert that `ignored` is one of `0`, `total-1` or `total` depending (respectively) on whether we're ignoring nothing, just replicas, or all copies?

DaveCTurner (approved):
LGTM

PR #87813, opened 2022-06-17
Remove default-value parameters from non-core queries

romseygeek:
This slims down the representations of all remaining queries not in the core
* percolate
* function_score
* script_score
* pinned

Relates to #76515


javanna (approved):
LGTM

Issue #87625, opened 2022-06-14
Integrate ANN into `_search` endpoint


jtibshirani:
Currently approximate nearest neighbor search is powered by its own `_knn_search` endpoint. We introduced ANN in a separate endpoint because there were big open questions around how it should integrate with the search API (for example around score combination, aggregations, pagination, etc.) We now have a better understanding on many of these topics and plan to integrate ANN into `_search`.

### Proposed API

ANN will be powered by a new `knn` section in the search request:

```
POST products/_search
{
  "query": {
    "multi_match": {
      "query": "black flower dress",
      "fields": ["title", "description"],
      "boost": 0.9
    }
  },
  "knn": {
    "field": "title_vector",
    "query_vector": [0.3, 0.1, ...],
    "k": 5,
    "num_candidates": 50,
    "boost": 0.1
  },
  "size": 20
}
```

This search finds the global top k=5 vector results, finds all term-based matches, combines them, and finally takes the top 20 top-scoring results. One way to think about it: it's just a normal search, but the `knn` section adds `k` additional matching documents based on vector similarity, like taking a boolean disjunction between the query and the ANN search. You can page through search results as usual, in case you've retrieved more interesting vector and term-based matches than fit on one page.

**Scoring.** In this initial version we'll compute the score of each hit as a weighted sum of the vector and term-based scores. This matches the usual way scores are combined in Elasticsearch, by summing query scores with optional boosts. In the future we can consider other ways of combining scores like reciprocal rank fusion.

**Retrieval.** The top `k` vector results represent the global nearest neighbors across all shards. This contrasts with a shard-centric approach, where we would retrieve the `k` nearest neighbors per shard. In a shard-centric approach, there would actually be `k * num_shards` matches, which you could see while paging through results, etc. We think the global behavior will be more predictable and easier to understand, especially when searching across multiple indices, where each index could have a different number of shards.

The new `knn` section would **still be marked experimental/ technical preview**. Making ANN generally available (GA) is a separate step.

### Implementation

A naive implementation would create a bool query on each shard, taking a disjunction between the term-based `query` and a "knn" query that retrieves `k` nearest documents. But this would give the shard-centric behavior described above, where there are `k * num_shards` total vector matches, instead of a global top `k`.

Our current plan is to perform ANN during an initial search phase that runs before the query phase. Specifically:
[...]

PR #87815, opened 2022-06-17
Move kNN search and dense vectors to core

jtibshirani:
This PR proposes moving kNN search and dense vector support out of an xpack
plugin and into server.

In #87625 we plan to integrate ANN search into the main `_search` endpoint as a
new top-level component called `knn`. So kNN will be a dedicated part of the
search request, and we'll have kNN logic within the search phases. The classes
and logic will live in server, matching the other search components like
suggesters, field collapsing, etc.

Why can't we leave dense vector in its own plugin, and just add kNN search
logic in server?
* This makes testing more difficult/ confusing, since kNN search can only be
exercised effectively using dense vector fields. All tests for kNN search would
need to live in the dense vector plugin, although kNN is part of core search.
* It feels cleanest for all logic related to kNN search to live in the same
place.


server/src/main/java/org/elasticsearch/index/mapper/MappingLookup.java
```diff
@@ -257,8 +258,8 @@ public PostingsFormat getPostingsFormat(String field) {
      */
     public KnnVectorsFormat getKnnVectorsFormatForField(String field) {
         Mapper fieldMapper = fieldMappers.get(field);
-        if (fieldMapper instanceof PerFieldKnnVectorsFormatFieldMapper) {
-            return ((PerFieldKnnVectorsFormatFieldMapper) fieldMapper).getKnnVectorsFormatForField();
+        if (fieldMapper instanceof DenseVectorFieldMapper) {
```
javanna, line 261: drive-by comment, that has little to do with this change specifically, as this logic was already here. Do we need an additional public method in MappingLookup for this? Would it make sense to streamline this into the Mapper base class to be overridden by DenseVectorFieldMapper?

jtibshirani: Yes, I'm not sure why I thought we needed this! I can remove that `MappingLookup` method now. I am not sure if we want a new top-level method `Mapper#getKnnVectorsFormatForField`, as opposed just doing an `instanceof` check for `DenseVectorFieldMapper`? My inspiration for this change was `CompletionFieldMapper`, which has a special package-private method to define `PostingsFormat`.

mayya-sharipova (approved):
@jtibshirani Thanks, this arrangement LGTM.

server/src/main/java/org/elasticsearch/index/mapper/MappingLookup.java
javanna: thanks for the background Alan, makes sense to me as well.

modules/lang-painless/src/main/generated/whitelist-json/painless-score.json
```diff
@@ -1 +1 @@
...
...tUtils$DecayNumericLinear","name":"decayNumericLinear","return":"double","read_only":4,"parameters":["double","double","double","double","double"]},{"declaring":"org.elasticsearch.script.VectorScoreScriptUtils$DotProduct","name":"dotProduct","return":"double","read_only":3,"parameters":["org.elasticsearch.script.ScoreScript","java.util.List","java.lang.String"]},{"declaring":"org.elasticsearch.script.VectorScoreScriptUtils$L1Norm","name":"l1norm","return":"double","read_only":3,"parameters":["org.elasticsearch.script.ScoreScript","java.util.List","java.lang.String"]},{"declaring":"org.elasticsearch.script.VectorScoreScriptUtils$L2Norm","name":"l2norm","return":"double","read_only":3,"parameters":["org.elasticsearch.script.ScoreScript","java.util.List","java.lang.String"]},{"declaring":"org.elasticsearch.script.ScoreScriptUtils$RandomScoreDoc","name":"randomScore","return":"double","read_only":2,"parameters":["org.elasticsearch.script.ScoreScript","int"]},{"declaring":"org.elasticsearch.script.ScoreScriptUtils$RandomScoreField","name":"randomScore","return":"double","read_only":3,"parameters":["org.elasticsearch.script.ScoreScript","int","java.lang.String"]}],"instance_bindings":[]}
```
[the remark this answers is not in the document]
jtibshirani: Some of the full class names changed they were renamed, like `org.elasticsearch.script.VectorScoreScriptUtils$CosineSimilarity`.

PR #87821, opened 2022-06-18
Make DiskThresholdDecider a little faster

original-brownbear:
This is the most expensive allocator left in many-shards benchmarks.
It gets a little faster with this change iterating over the cheaper
array and inlining a little nicer by having specialized methods
for the two cached states.
Also removed needless varags on the generic states method which is almost exclusively used in tests now (only a single prod use case left).

relates #77466


server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
```diff
@@ -213,29 +213,34 @@ public int numberOfShardsWithState(ShardRoutingState... states) {
...
 
+    private static final ShardRouting[] EMPTY_SHARD_ROUTING_ARRAY = new ShardRouting[0];
+
+    public ShardRouting[] initializing() {
+        return initializingShards.toArray(EMPTY_SHARD_ROUTING_ARRAY);
+    }
+
+    public ShardRouting[] relocating() {
+        return relocatingShards.toArray(EMPTY_SHARD_ROUTING_ARRAY);
+    }
```
idegtiarenko, line 242: Is it better to use `.toArray(ShardRouting[]::new)` in those two cases to avoid a constant?

idegtiarenko (approved):

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
original-brownbear: I always thought the empty array version was still the fastest. Probably doesn't matter too much even here, but with the empty case so likely here I figured that was even more true since we always use the same constant array then.

original-brownbear:
Thanks Ievgen!

PR #87843, opened 2022-06-20
Fix optimization in BalancedShardsAllocator

original-brownbear:
The optimization for a `null` does not apply today as we never pass `null` to this method, it should special case plain `YES` instead.

relates #77466


idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

Issue #31549, opened 2018-06-25
Consider adding support for OFFSET


astefan:
At the moment, ES SQL supports `LIMIT` only. Would be helpful to also support `OFFSET` to allow users to use pagination.

flydocs-jignesh:
Any hope?

astefan:
@flydocs-jignesh As an update, OFFSET is not on our near future feature list.

miaoj0312:
How long?

hank583746309:
I want know how long . can support 'offset' too. dear~

caihaibin1991:
me too.... dear ....

thomasj-git:
Hope to support offset as soon as possible!!!!

caihaibin1991:
I use /_xpack/sql/translate to turn the sql to be es sql, and then append from to strip the offset ....

leanderli:
It's really useful.Hope to support it ...

wdgrantham:
Just hopping on here.  I've been migrating a very large MySQL database over to elastic search to improve search speed.  A key part of my application is allowing my users to pull lots of data from my database, and I've historically used offsets to ensure that they don't get duplicate records.  

Now that I know that elastic search sql doesn't support offsets, I'm having to write some funky logic to do it manually.  For example, where I used to use a simple offset command like this: 

`SELECT fn, ln, addr from "Db"."Table" WHERE (fn like 'And%') LIMIT 5000 OFFSET 5000`

I'm now having to write something like this:

``SELECT id, fn, ln, addr from "Db"."Table" WHERE (fn like 'Jam%') and (id > ${last_id_user_downloaded} ) ORDER BY id LIMIT 5000``

Then I have to strip the `id` field from the results of all the rows before I create the download.

Would really appreciate if you all could support `OFFSET`

chenrenfei:
Why not consider pagination by sql？
and provide an example？

Mokaful:
I'm very very very eager for pagination.when can we get it?

leanderli:
> I'm very very very eager for pagination.when can we get it?

Opendistro-for-elasticsearch SQL support it and powerful.

sephiroka:
Does new version's ES-SQL provide solutions to help paging?

cloder:
Is there any news about this?

astefan:
OFFSET is not on our map at the moment.
[...]

PR #87846, opened 2022-06-20
Add support for 'size' in EQL Sample queries

luigidellaquila:
EQL queries allow to pass a "size" parameter to limit the number of returned results:

```
{
  "size": 3,
  "query": "<an EQL query>"
}
```

This PR adds support for this parameter on Sample queries.


x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sample/SampleIterator.java
```diff
@@ -174,8 +178,16 @@ private void finalStep(ActionListener<Payload> listener) {
                 if (docGroupsCounter == maxCriteria) {
                     List<SearchHit> match = matchSample(sample, maxCriteria);
                     if (match != null) {
-                        finalSamples.add(match);
-                        samples.add(new Sample(sampleKeys.get(responseIndex / maxCriteria), match));
+                        if (samplesDiscarded < limit.offset()) {
```
astefan, line 181: WHy do you use both offset and limit?

astefan (commented):
Left some comments

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/plan/physical/SampleExec.java
```diff
@@ -24,20 +25,22 @@
 public class SampleExec extends PhysicalPlan {
 
     private final List<List<Attribute>> keys;
+    private final Limit limit;
```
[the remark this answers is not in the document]
luigidellaquila: That's what I did at first, but then I realized that a Limit could have a few advantages:
- it's consistent with the other EQL command implementations (eg. SequenceMatcher uses a Limit as well)
- it anticipates the need for pagination, that is pretty natural in most scenarios (see also [here](https://github.com/elastic/elasticsearch/issues/31549))
- it doesn't really add any significant complication or overhead

Anyway, it's not strictly necessary for the scope of this PR, so let's discuss it; if there is no agreement, I can easily reduce it to a number.

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sample/SampleIterator.java
luigidellaquila: As long as we have Limit (see previous comment), IMHO it makes sense to have a complete support for it.
The logic is trivial and the added value could be significant: with a small additional effort on syntax, we'll have at least a basic support for pagination.

astefan: Then why not having Limit as a node in the logical plan tree (vs a property of Sample plan)? (like SQL and EQL do it)

luigidellaquila: Good point, checking

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/parser/LogicalPlanBuilder.java
```diff
@@ -102,7 +102,13 @@ public Object visitStatement(StatementContext ctx) {
             if (ctx.pipe().size() > 0) {
                 throw new ParsingException(source(ctx.pipe().get(0)), "Samples do not support pipes yet");
             }
-            return plan;
+            LimitWithOffset limitPlan = new LimitWithOffset(
```
Luegg, line 105: nit pick

```suggestion
            return new LimitWithOffset(
```

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/parser/LogicalPlanBuilder.java
```diff
@@ -102,7 +102,13 @@ public Object visitStatement(StatementContext ctx) {
             if (ctx.pipe().size() > 0) {
                 throw new ParsingException(source(ctx.pipe().get(0)), "Samples do not support pipes yet");
```
Luegg, line 103: Shouldn't support for pipes now also be possible since limit and offset are both supported?

astefan: No, because the limit/size comes from a request parameter, not from the query itself (through the query parser).

Luegg: but it works with `sequence`, doesn't it? why would `sample` be different in this regard?

astefan: With `sequence` you have `| head 5 | tail 3 | head 1`. How do you do this with `sample`s?

luigidellaquila: I see HEAD and TAIL as very natural operators for sequences, that have well defined semantics defined by their order (eg. `TAIL 10`: give me last ten - more recent - events). 
Samples, on the other hand, do not have a natural order: "last ten samples" is not a clear concept by itself.
In this context, LIMIT/OFFSET is much more similar to the SQL equivalent, where you just need a way to retrieve the results in batches or to show them in a UI in pages.
I agree with @Luegg on the fact that we could support pipes here eventually, but we probably need some slightly different syntax, eg `| LIMIT 10 OFFSET 20`; or reuse HEAD with `| OFFSET 20 | HEAD 10`, though it seems less natural in this context.

astefan: @Luegg, as @luigidellaquila well put this, re-using `HEAD` and `TAIL` doesn't reflect the nature of `sample`s which do not have a chronological ordering concept, like `sequence`s do. The only type of ordering that make sense is the one on the join keys, but at that point the simple concepts of `HEAD` and `TAIL` are not enough to express the multitude of ordering options available (order by first join key asc, order by the second join key desc, order by the third join key asc and so on; or order by all keys asc, or all keys desc, but at that point one would ask the question why cannot I order by each key individually etc). At this point in the evolution of `sample`s, I believe having a circuit breaker and a simple way of limiting (without computing all the samples) the number of results returned, is enough. With time, and more input from users, we can evaluate our options in providing something more advanced and more granular in terms of limiting the number of results and, at the same time, sorting the results.

luigidellaquila: @Luegg I agree with you, tail is definitely not straight forward at this stage.
Also for a proper implementation of pipes on samples (and limit/offset in particular), I think we still need to define a few small details that do not really fit in the scope of this PR. 
For this specific one, I'd suggest to stick to the original goal, but keeping a door open for further improvements.

luigidellaquila:

x-pack/plugin/eql/qa/common/src/main/resources/test_sample.toml
```diff
@@ -250,3 +250,61 @@ expected_event_ids  = [18,11]
 join_keys = ["doom","win10"]
 
 
+[[queries]]
+name = "size0"
+size = 0
```
astefan, line 255: Other eql queries use `size` as well, but none of them require the toml tests to have `size` in them. I'm wondering if there isn't any other way to test `size` given that so far there was no need for `size` in IT toml tests.

astefan (commented):
There were two approaches in this PR regarding `size`:
- a Sample having a `size` attribute, which I don't believe it's something that the sample node must have
- a Limit logical plan with a Sample logical plan child node

In the latter case (the one in the PR at this point), transforming the logical plan into a physical one doesn't follow the usual steps. For example, the `Mapper` is the one that generates physical plans based on the logical ones and, according to what's in there already a `LimitWithOffset` should map to a `LimitWithOffsetExec`. In your case, `SampleExec` physical plan includes a `Limit`, but it should in fact belong to a physical plan having as parent a `LimitWithOffsetExec`.

x-pack/plugin/eql/qa/common/src/main/resources/test_sample.toml
luigidellaquila: > For example, the Mapper is the one that generates physical plans based on the logical ones and, according to what's in there already a LimitWithOffset should map to a LimitWithOffsetExec. In your case, SampleExec physical plan includes a Limit, but it should in fact belong to a physical plan having as parent a LimitWithOffsetExec.

I'm not sure I got this comment: `LimitWithOffset+Sample` logical plan is transformed to `LimitWithOffsetExec+SampleExec`, then `QueryFolder.FoldLimit` rule applies the limit (from LimitWithOffsetExec) to SampleExec, that is then used to build SampleIterator.
This is the exact same thing that happens for Sequence execution planning.

About "size" in the tests, we have unit tests (mostly for the execution planning, plus some specific tests for the algorithms) in Java, but they do not fully execute any query (mocking the full execution is extremely convoluted due to the usage of internal - Server - components in the execution, that cannot be instantiated in our tests). 
The only real execution tests we have are TOML tests, this is why I added a size attribute there. It allows us to test the size on a full execution path.
In addition, I think also Sequence tests can take advantage of it.

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sample/SampleIterator.java
bpintea: Pagination won't work without state. 
Without reading the comments first, I also found this code confusing and had to check if we provide a non-null offset anywhere. I guess this could be clarified with a comment, but not sure where such a comment should be placed (if we ever add an offset, a comment placed here will nearly certainly not be updated/removed).
IMO, since the logic is trivial, it could be added when actually needed. But I guess it could also be kept.

bpintea (approved):
lgtm

astefan (commented):
Left few more, minor comments.

astefan (approved):
LGTM

x-pack/plugin/eql/qa/common/src/main/java/org/elasticsearch/test/eql/BaseEqlSpecTestCase.java
```diff
@@ -46,6 +46,7 @@ public abstract class BaseEqlSpecTestCase extends RemoteClusterAwareEqlRestTestC
      * For now, every value will be converted to a String.
      */
     private final String[] joinKeys;
+    private final Integer size;
```
costin, line 49: I have only one nitpick around the usage of Integer (object) instead of a primitive.
The sign doesn't seem to be used (head and tail take care of where the limit is applied) hence why not use the negative value as indicator that the size is out of band and thus unspecified?
As Integer, size has 3 states - valid if >=0, unspecified if null and invalid if <0.
What I'm saying is using only 2 through a primitive (int) - valid if >=0, unspecified if <0.

costin (approved):
LGTM with a small nitpick.

What happens if both size and limit are specified?

x-pack/plugin/eql/qa/common/src/main/java/org/elasticsearch/test/eql/BaseEqlSpecTestCase.java
luigidellaquila: > What happens if both size and limit are specified?

Do you mean something like `... | tail n` with n different from size?
For samples, at this stage, pipes are not supported yet, so it cannot happen.
For sequences, both conditions are applied, so the smaller one prevails; this seems a reasonable behavior, so I'd say we could have the same for samples in the future

PR #87849, opened 2022-06-20
[TSDB] Expose `aggregate_metric_double` fields as their own field type instead of `double`

csoulios:
Fields of type `aggregate_metric_double` are currently exposed in the [field caps API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-field-caps.html) as fields of type `double`

This PR changes this behavior so that `aggregate_metric_double` fields are exposed with their own field type.

## Context

In the past, we had taken the decision to expose fields of type `aggregate_metric_double` as regular `double` fields in the field caps API. This decision intended to hide the complexity of this field from Kibana and only expose it as a simple numeric field. After recent discussions with the Kibana team, we decided that Elasticsearch should be transparent as to when a field is an `aggregate_metric_double` field and Kibana will deal with managing the field type.

## Consequences
**This is a breaking change**


x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/aggregate-metrics/100_field_caps.yml
```diff
@@ -0,0 +1,58 @@
+setup:
+  - skip:
+      version: " - 7.15.99"
```
nik9000, line 3: Should this be be 8.4?

csoulios: yes, it was left from copy-paste. changed it now.

romseygeek (approved):
LGTM, thanks @csoulios

csoulios:
@elasticmachine update branch

PR #87706, opened 2022-06-15
Add Lintian overrides to ignore Intel MKL not linked to libc

droberts195:
We don't build these libraries ourselves and the license forbids
us from modifying them in any way, so we won't be able to make
this rule pass on them. All we can do is override it.

Fixes #87632

PR #87850, opened 2022-06-20
[8.2] Add Lintian overrides to ignore Intel MKL not linked to libc

droberts195:
We don't build these libraries ourselves and the license forbids
us from modifying them in any way, so we won't be able to make
this rule pass on them. All we can do is override it.

Backport of #87706


droberts195:
@elasticmachine update branch

PR #87851, opened 2022-06-20
[Docs] Fix runtime grok script example

### LINKED ISSUE #87848: The explanation of the code does not match the results
At the link  https://www.elastic.co/guide/en/elasticsearch/reference/8.2/runtime-examples.html, there is a such a piece of code:

`PUT my-index-000001/_mappings
{
  "runtime": {
    "http.client_ip": {
      "type": "ip",
      "script": """
        String clientip=grok('%{COMMONAPACHELOG}').extract(doc["message"].value)?.clientip;
        if (clientip != null) emit(clientip); 
      """
    }
  }
}`

The explanation in the text:

` If the pattern doesn’t match (clientip != null), the script just returns the field value without crashing.`

My understanding is that clientip != null, it indicates a match instead of a mismatch.

Thanks

kilfoyle:
This corrects a small mistake in the "Define a runtime field with a grok pattern" example. The example snippet `clientip != null` should go with the text "If the pattern matches" rather than "If the pattern doesn't match".

Closes: #87848

### DIFF
--- a/docs/reference/scripting/common-script-uses.asciidoc
+++ b/docs/reference/scripting/common-script-uses.asciidoc
@@ -76,9 +76,9 @@ field as a runtime field in the mapping. The following runtime script defines a
 grok pattern that extracts structured fields out of the `message` field. 
 
 The script matches on the `%{COMMONAPACHELOG}` log pattern, which understands
-the structure of Apache logs. If the pattern matches, the script emits the
-value matching the IP address. If the pattern doesn't match
-(`clientip != null`), the script just returns the field value without crashing.
+the structure of Apache logs. If the pattern matches (`clientip != null`), the
+script emits the value matching the IP address. If the pattern doesn't match,
+the script just returns the field value without crashing.
 
 [source,console]
 ----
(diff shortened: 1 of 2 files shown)


docs/reference/scripting/common-script-uses.asciidoc
```diff
@@ -76,9 +76,9 @@ field as a runtime field in the mapping. The following runtime script defines a
 grok pattern that extracts structured fields out of the `message` field. 
 
 The script matches on the `%{COMMONAPACHELOG}` log pattern, which understands
-the structure of Apache logs. If the pattern matches, the script emits the
-value matching the IP address. If the pattern doesn't match
-(`clientip != null`), the script just returns the field value without crashing.
+the structure of Apache logs. If the pattern matches (`clientip != null`), the
+script emits the value of the matching the IP address. If the pattern doesn't match,
```
lockewritesdocs, line 80:
```suggestion
script emits the value of the matching IP address. If the pattern doesn't match,
```

lockewritesdocs (approved):
Small typo fix, but LGTM otherwise 🦖 Thanks @kilfoyle!

Issue #87721, opened 2022-06-16
Add user actions for the ilm,slm, and repository_integrity indicators


andreidan:

andreidan:
This has been completed as of 8.4

PR #87854, opened 2022-06-20
Add start slm user action

andreidan:
This creates a user action for the slm health indicator that will help
the user to start SLM.

Relates to https://github.com/elastic/elasticsearch/issues/87721


x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/slm/SlmHealthIndicatorService.java
```diff
@@ -38,6 +39,10 @@ public class SlmHealthIndicatorService implements HealthIndicatorService {
     public static final String NAME = "slm";
 
     public static final String HELP_URL = "https://ela.st/fix-slm";
+    public static final UserAction SLM_NOT_RUNNING = new UserAction(
+        new UserAction.Definition("slm-not-running", "Snapshot lifecycle management is not running", HELP_URL),
```
jbaiera, line 43: Similar to the ILM PR, lets say that they should enable SLM here instead of just repeating the indicator status.

jbaiera (approved):
LGTM, one suggestion

PR #87857, opened 2022-06-20
More tests for filter-by-filter fast paths

nik9000:
This adds some more tests for the filter-by-filter execution mechanism
for `terms` and `date_histogram` around the "fast path" that it can take
when we can use the lucene `Weight#count` method. Specifically, this
fixes some flakiness in a test when some segments can't take the fast
path. And it adds a new test where all segments can take the fast path
even though there is a filter - but the filter matches all documents.
Finally, it fixes another flaky test that assumed there was always a
single segment.

Closes #87722


not-napoleon (approved):
LGTM

PR #87860, opened 2022-06-20
Remove superfluous iterator usage from RoutingNode

arteam:
In internal cases where we filter shards by their state, we can iterate directly over the shard map without
creating an additional unmodifiable iterator.


original-brownbear (commented):
Not sure it's worth adding more symbols here, these methods are never hot to my knowledge?

PR #87866, opened 2022-06-20
Simplify adding dynamic sub-fields to their dynamic parent object

javanna:
DocumentParserContext holds all the dynamic mappers created while parsing a document, as well as dynamic object mappers on a separate map so they can be looked up by name quickly when parsing their sub-fields and looking for their parent field. For every sub-field, so far we looked up the corresponding object mapper and then call .newBuilder on it. Calling newBuilder can be done already when the object mapper is added to the map, which simplifies the logic down the line as instead of creating a new builde for each sub-field, we re-use the same builder and add sub-fields to it. This has no effect on the resulting dynamic update, but it simplifies the code and it should have a positive impact on performance.

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
@@ -700,13 +700,15 @@ public final class DocumentParser {
             parentMapper = context.mappingLookup().objectMappers().get(parentName);
             if (parentMapper == null) {
                 // If parentMapper is null, it means the parent of the current mapper is being dynamically created right now
-                parentMapper = context.getDynamicObjectMapper(parentName);
-                if (parentMapper == null) {
+                ObjectMapper.Builder dynamicObjectMapperBuilder = context.getDynamicObjectMapperBuilder(parentName);
+                if (dynamicObjectMapperBuilder == null) {
                     // it can still happen that the path is ambiguous and we are not able to locate the parent
                     break;
                 }
+                dynamic = dynamicObjectMapperBuilder.dynamic;
+            } else {
+                dynamic = parentMapper.dynamic();
             }
-            dynamic = parentMapper.dynamic();
         }
         if (dynamic == null) {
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
```diff
@@ -690,26 +690,26 @@ private static void parseCopyFields(DocumentParserContext context, List<String>
...
-            if (parentMapper == null) {
+            parentName = parentMapper.name().substring(0, lastDotNdx);
+            ObjectMapper parent = context.mappingLookup().objectMappers().get(parentName);
+            if (parent == null) {
                 // If parentMapper is null, it means the parent of the current mapper is being dynamically created right now
-                parentMapper = context.getDynamicObjectMapper(parentName);
-                if (parentMapper == null) {
+                ObjectMapper.Builder dynamicObjectMapperBuilder = context.getDynamicObjectMapperBuilder(parentName);
+                if (dynamicObjectMapperBuilder == null) {
                     // it can still happen that the path is ambiguous and we are not able to locate the parent
```
romseygeek, line 706: I wonder if this is still true? I guess it doesn't matter so much.

romseygeek (approved):
LGTM.

One question, does this mean that we can get rid of the object merging when we build the final dynamic update now? Because we should only have a single builder per object, right?

server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
javanna: we are very much aligned :) I want to to do something about this in a follow-up. I think probably have an assertion that this never happens. If it does happen I would like to know why...

javanna:
> One question, does this mean that we can get rid of the object merging when we build the final dynamic update now? Because we should only have a single builder per object, right?

Not yet, but that would be the plan more or less... I haven't figured out exactly how to get there but I am on it.

javanna:
Heads, up this was reverted due to a considerable performance regression when building the root object mapper as part of creating the dynamic mapping update. I will dig to understand what caused that and eventually reopen.

PR #87868, opened 2022-06-21
Add setting for tcp_keepalive for oidc back-channel

ywangd:
This PR adds a new setting to enable tcp keepalive probes for the
connections used by the oidc back-channel communication. It defaults to
true as tcp keepalive is generally useful for ES.

Relates: #87773


ywangd:
NOTE: This is currently a draft PR because:
1. I cannot find a way to test tcp keepalive
2. Depending on the OS's configuration, the keepalive probe may not be enough keep the connection up. #87773 is still the right candidate to address #75515. But maybe this change can be useful for some other situations.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/oidc/OpenIdConnectRealmSettings.java
```diff
@@ -214,6 +214,12 @@ private OpenIdConnectRealmSettings() {}
         key -> Setting.intSetting(key, 200, Setting.Property.NodeScope)
     );
 
+    public static final Setting.AffixSetting<Boolean> HTTP_TCP_KEEP_ALIVE = Setting.affixKeySetting(
+        RealmSettings.realmSettingPrefix(TYPE),
+        "http.tcp.keep_alive",
+        key -> Setting.boolSetting(key, false, Setting.Property.NodeScope)
```
ywangd, line 220: Maybe it should default to `true`.

ywangd: I updated it to true

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticator.java
```diff
@@ -691,7 +693,9 @@ private CloseableHttpAsyncClient createHttpClient() {
         try {
             SpecialPermission.check();
             return AccessController.doPrivileged((PrivilegedExceptionAction<CloseableHttpAsyncClient>) () -> {
-                ConnectingIOReactor ioReactor = new DefaultConnectingIOReactor();
+                ConnectingIOReactor ioReactor = new DefaultConnectingIOReactor(
+                    IOReactorConfig.custom().setSoKeepAlive(realmConfig.getSetting(HTTP_TCP_KEEP_ALIVE)).build()
+                );
```
ywangd, line 698: There is no easy way to test this setting. Other than invasive reflection, one possibility is to subclass `DefaultConnectingIOReactor` and override the `prepareSocket` method where we can assert the socket configuration. It feels heavy weighted for testing though. We may need the subclass in production if we decide to support for configuring keepalive_time, interval etc in future. So I think we can save the test till we actually need that in production.

DaveCTurner (approved):
It would be good to at least confirm manually that this does enable keepalives on the connection (e.g. using `ss -toe`). I expect it does so this LGTM.

If we were to do something like #84653 then we could check this timer directly in a test (at least on Linux).

ywangd:
> It would be good to at least confirm manually that this does enable keepalives on the connection (e.g. using `ss -toe`).

Thanks for the tip. I manually verified it and it worked:

Before 

> ESTAB         0              0                         ***:60142                 ***:443            uid:1000 ino:253847 sk:2002 <->

After

> ESTAB       0            0                     *:53448              *:443         timer:(keepalive,119min,0) uid:1000 ino:261040 sk:4001 <->

docs/reference/settings/security-settings.asciidoc
```diff
@@ -1858,6 +1858,13 @@ connections allowed per endpoint.
 Defaults to `200`.
 // end::oidc-http-max-endpoint-connections-tag[]
 
+// tag::oidc-http-tcp-keepalive-tag[]
+`http.tcp.keep_alive` {ess-icon}::
+(<<static-cluster-setting,Static>>)
+Wehther to enable TCP keepalives on HTTP connections used for back-channel communication
```
tvernum, line 1864:
```suggestion
Whether to enable TCP keepalives on HTTP connections used for back-channel communication
```

tvernum (approved):
LGTM, pending fixed typo in docs.

PR #87869, opened 2022-06-21
Fix pending listener completion in queue

idegtiarenko:
Please see comments inline

### DIFF
--- a/server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/PendingListenersQueueTests.java
+++ b/server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/PendingListenersQueueTests.java
@@ -71,6 +71,45 @@ public class PendingListenersQueueTests extends ESTestCase {
         }
     }
 
+    /**
+     * This test simulates following scenario:
+     *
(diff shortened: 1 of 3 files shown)


server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/PendingListenersQueueTests.java
```diff
@@ -71,6 +71,45 @@ public void testShouldExecuteOnlyCompleted() throws InterruptedException {
...
+     * Master: submits new balance calculation (index 2)
+     * Master: allocates using already computed balance (index 1)
+     *
+     * Calculator: starts computing new balance (index 2)
+     * Calculator: new balance is computed (isFresh=true, hasChanges=false), completes the queue for the new balance (index 2)
+     *
+     * Master: completes the queue for applied balance (index 1)
+     * Master: resumes the queue
+     *
+     * Expects: Listener (index 2) should be completed
```
idegtiarenko, line 88: In this case index was rolled back to 1 and corresponding listener was stuck until something else advances

DaveCTurner (approved):
LGTM, nice find

PR #87871, opened 2022-06-21
[DOCS] Adds note to time_of_week function about how values are calculated

szabosteve:
## Overview

This PR adds a note to the time functions page in the anomaly detection book. The note briefly describes how the `actual` and `typical` values of the `time_of_week` function are calculated.

Related to https://github.com/elastic/kibana/issues/133745.

### Preview

[Time_of_week](https://elasticsearch_87871.docs-preview.app.elstc.co/guide/en/machine-learning/master/ml-time-functions.html#ml-time-of-week)

### DIFF
--- a/docs/reference/ml/anomaly-detection/functions/ml-time-functions.asciidoc
+++ b/docs/reference/ml/anomaly-detection/functions/ml-time-functions.asciidoc
@@ -2,8 +2,8 @@
 = Time functions
 
 The time functions detect events that happen at unusual times, either of the day
-or of the week. These functions can be used to find unusual patterns of behavior,
-typically associated with suspicious user activity.
+or of the week. These functions can be used to find unusual patterns of 
+behavior, typically associated with suspicious user activity.
 
 The {ml-features} include the following time functions:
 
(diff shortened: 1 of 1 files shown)


docs/reference/ml/anomaly-detection/functions/ml-time-functions.asciidoc
```diff
@@ -77,6 +77,12 @@ its past behavior.
 The `time_of_week` function detects when events occur that are outside normal
 usage patterns. For example, it detects login events on the weekend.
 
+IMPORTANT: The `time_of_week` function is modeled in UTC and the duration of the 
+  week is in seconds. It means that the `typical` and `actual` values are 
+  seconds after a whole number of weeks after 1/1/1970 in UTC which is a 
+  Thursday. For example, a value of `475` is 475 seconds after midnight on 
+  Thursday in UTC.
```
tveasey, line 84: The first sentence doesn't read quite right to me. How about:
```suggestion
IMPORTANT: The `time_of_week` function models time in epoch seconds modulo the duration
  of a week in seconds. It means that the `typical` and `actual` values are 
  seconds after a whole number of weeks since 1/1/1970 in UTC which is a 
  Thursday. For example, a value of `475` is 475 seconds after midnight on 
  Thursday in UTC. 
```

tveasey (commented):
I proposed a slight tweak. Modulo is the technical term for the operation we do. I think since you explain it subsequently it is fine to use this. I think it is also important to say epoch seconds since UTC is just a time zone and epoch captures the fact that it is seconds since 1970 in the UTC time zone.

tveasey (approved):
LGTM

PR #87875, opened 2022-06-21
Add high level java docs to DesiredNodes

fcofdez:


server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNodes.java
```diff
@@ -39,6 +40,71 @@
...
+ *  Each {@link DesiredNode} part of {@link DesiredNodes} has a {@link DesiredNodeWithStatus.Status}
+ *  depending on whether or not the node has been part of the cluster at some point.
+ * </p>
+ *
+ *  The two possible statuses {@link DesiredNodeWithStatus.Status} are:
+ *  <ul>
+ *      <li>{@code PENDING}: The {@link DesiredNode} is not part of the cluster yet</li>
+ *      <li>{@code ACTUALIZED}: The {@link DesiredNode} ir or has been part of the cluster.
+ *          Notice that it is possible that a node has {@code ACTUALIZED} status but it is not part of {@link DiscoveryNodes},
+ *          this is a conscious decision as it is expected that nodes can leave the cluster momentarily due to network issues,
```
henningandersen, line 92: Can we include restarts and node failures here?

henningandersen (approved):
LGTM.

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNodes.java
```diff
@@ -39,6 +40,71 @@
...
+ *  allowing it to make better decisions about allocation, autoscaling, auto-expand replicas, etc.
+ * </p>
+ *
+ * <p>
+ *  Additionally, settings validation is done during desired nodes updates avoiding boot-looping
+ *  when an invalid setting is provided before the node is started.
+ * </p>
+ *
+ * <p>
+ *  To modify the desired nodes it is needed to provide the entire collection of nodes that will
```
dakrone, line 61: minor grammar nit

```suggestion
 *  To modify the desired nodes it is necessary to provide the entire collection of nodes that will
```

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNodes.java
```diff
@@ -39,6 +40,71 @@
...
+ *  be part of the proposed cluster topology.
+ * </p>
+ *
+ * <p>
+ *  Desired nodes are expected to be part of a lineage defined by the provided {@code historyId}.
+ *  The {@code historyId} is provided by the orchestrator taking care of managing the cluster.
+ *  In order to identify the different proposed desired nodes within the same history, it is
+ *  also expected that the orchestrator provides a monotonically increasing {@code version}
+ *  when it communicates about future topology changes.
+ *  The cluster would reject desired nodes updated with a {@code version} less than or equal
```
dakrone, line 71: The word "would" is a little grammatically confusing here

```suggestion
 *  The cluster rejects desired nodes updated with a {@code version} less than or equal
```

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNodes.java
```diff
@@ -39,6 +40,71 @@
...
+ *  than the current {@code version} for the same {@code historyId}.
+ * </p>
+ *
+ * <p>
+ *  The {@code historyId} is expected to remain stable during the cluster lifecycle, but it is
+ *  possible that the orchestrator loses its own state and needs to be restored to a
+ *  previous point in time with an older desired nodes {@code version}. In those cases it is
+ *  expected to use new {@code historyId} that would allow starting from a different version.
```
dakrone, line 79: Does `historyId` need to be monotonically increasing, or will any different history id work?

dakrone (approved):
LGTM also

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNodes.java
fcofdez: `historyId` is a string, so any different history id works in this case

fcofdez:
Thanks both!

PR #87876, opened 2022-06-21
Make CBE message creation more robust

DaveCTurner:
Child circuit breakers rely on proper matching of acquire/release pairs.
This can be tricky to get right. If we get it wrong and accidentally
double-release a CB then it may end up with a negative `used` value.
This is definitely a bad situation in which to find ourselves, but today
in production it's made a whole lot worse because it causes exceptions
on every attempt to report a `CircuitBreakerStats` or to construct a
parent `CircuitBreakingException`.

This commit makes the message construction and stats serialization a
little more robust so that it's clearer what is going on in production.

Relates #86059.
Backport of #87687.

### DIFF
--- a/docs/changelog/87687.yaml
+++ b/docs/changelog/87687.yaml
@@ -1,0 +1,5 @@
+pr: 87687
+summary: Make CBE message creation more robust
+area: Infra/Circuit Breakers
+type: bug
+issues: []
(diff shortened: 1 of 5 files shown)


DaveCTurner:
Reverted due to test failures.

PR #78327, opened 2021-09-27
Build indicesAccessControl for selective indices

ywangd:
During index authorization, the IndicesAccessControl is built for all
original indices. The list of original indices can be large for a large
size cluster and requests with liberal wildcards. Therefore computing
IndicesAccessControl for every one of them can be expensive.

The observation is that many shard/index level request only really just
target a single concrete index and hence just need IndicesAccessControl
for that single index. One twist is that the single index may be
requested with its alias or authorized as part of a data stream. In this
case, the relevant aliases and data stream also need to be considered.
Still this list of concrete index and its aliases and data stream can be
much smaller than the original indices.

This PR adds an effective allowlist for requests that we know how to
extract the single concrete target index and only compute
IndicesAccessControl for the index and relevant aliases, data stream.

PR #87877, opened 2022-06-21
Remove aliases list from IndexAbstraction

original-brownbear:
We are not using this list outside of a couple of tests but setting it up
was a non-trivial contributor to the performance of building the indices lookup.

relates #77466


martijnvg (approved):
Thanks for removing this! I added this for #78327,
but that then never got merged. And never reverted
my change.

original-brownbear:
Thanks Martijn!

PR #87884, opened 2022-06-21
[ML] Add authorization info to ML config listings

droberts195:
ML datafeeds and data frame analytics jobs remember permissions
from their time of creation or most recent update. It can be
quite hard to determine what these are when listing their
configs, which can lead to uncertainty about whether some minor
update changed the permissions.

This change adds information about the permissions to the
config listings.

If the last update was from a user then the permissions used
are the roles they had at the time, and a list of these roles
is added to the new `authorization` field of the datafeed or
data frame analytics job config listing.

If the last update was using an API key then the permissions
of that API key are used for subsequent searches and the new
`authorization` field of the datafeed or data frame analytics
job config listing shows the API key name and ID.

If a service account did the last update then the service
account name is added to the new `authorization` field of the
datafeed or data frame analytics job config listing.

### DIFF
--- a/docs/changelog/87884.yaml
+++ b/docs/changelog/87884.yaml
@@ -1,0 +1,5 @@
+pr: 87884
+summary: Add authorization info to ML config listings
+area: Machine Learning
+type: enhancement
+issues: []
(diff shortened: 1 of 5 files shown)


przemekwitek (approved):
LGTM

droberts195:

Issue #87610, opened 2022-06-13
Consolidate Strings helper classes


pgomulka:
### Description

we have multiple Strings helper classes at the moment. It would be great to merge them together or categorise them and name accordingly. Some might be worth moving into internal packages. 

- [ ] libs/core  org.elasticsearch.core.Strings
- [ ] server org.elasticsearch.common.Strings
- [ ] server  org.elasticsearch.common.inject.internal.Strings
- [ ] x-pack/plugin/watcher org.elasticsearch.xpack.watcher.support.Strings
- [ ] x-pack/plugin/old-lucene-version org.elasticsearch.xpack.lucene.bwc.codecs.lucene54.LegacyStringHelper
- [ ] x-pack/plugin/eql ->  org.elasticsearch.xpack.eql.util.StringUtils
- [ ] x-pack/plugin/eql -> org.elasticsearch.xpack.eql.expression.function.scalar.string.StringUtils
- [ ] x-pack/plugin/sql/sql-proto -> org.elasticsearch.xpack.sql.proto.StringUtils
- [ ] x-pack/plugin/sql/sql-client -> org.elasticsearch.xpack.sql.client.StringUtils
- [ ] x-pack/plugin/sql -> org.elasticsearch.xpack.sql.expression.function.scalar.string.StringFunctionutils
- [ ] x-pack/plugin/ql -> org.elasticsearch.xpack.ql.util.StringUtils
- [ ] x-pack/plugin/ql  -> org.elasticsearch.xpack.ql.type.StringUtils

icemakegirl:
I made a pull request, can you look at it?

PR #87892, opened 2022-06-21
Removing redundant method in server org.elasticsearch.common.inject.internal.Strings

PDTCCLF:
This is my attempt to remove redundancy in server org.elasticsearch.common.inject.internal.Strings. Given the fact server org.elasticsearch.common.Strings already has a method to capitalize characters, an import should do the same as creating a new method in the previous code. It is a pull request to solve some parts of issue #87610.

### DIFF
--- a/docs/changelog/87892.yaml
+++ b/docs/changelog/87892.yaml
@@ -1,0 +1,6 @@
+pr: 87892
+summary: Removing redundant method in server org.elasticsearch.common.inject.internal.Strings
+area: Core/Infra
+type: refactoring
+issues:
+  - 87892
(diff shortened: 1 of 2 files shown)


rjernst:
Note that everything under o.e.common.inject should generally be left alone. This is a vendored version of Guice from a decade ago: making changes to it is something we try to avoid.

PDTCCLF:
> Hi @PDTCCLF, it appears that the Strings.capitalize method from the inject.internal package is in use at the moment in `TypeConverterBindingProcessor`. This use would first need to be refactored and then it appears that the whole class can be removed.

I understand. Thanks for pointing me in the right direction.

PDTCCLF:
> Note that everything under o.e.common.inject should generally be left alone. This is a vendored version of Guice from a decade ago: making changes to it is something we try to avoid.

Ok. I guess it is better to close this PR, then?

rjernst:
> I guess it is better to close this PR, then?

Yes, I think so. Thanks for understanding!

PR #87896, opened 2022-06-21
Use Map instead of ImmutableOpenMap in RoutingTable

rjernst:
The RoutingTable class usea a map for keeping track of index to shards.
This comit changes that class to use a HashMap internally. The jdk's
Map.copyOf is used when building the map since the read side is more
frequent then modification.

relates #86239


original-brownbear (changes requested):
Sorry missed this one, we can't do this here. The routing table is potentially huge and this would make applying the diffs and mutating the routing table way more expensive.

rjernst:
Closing this for now, will revisit in the future

PR #87898, opened 2022-06-21
Remove redundant jackson dependencies from discovery-azure

arteam:
The APIs that we use in `azure-svc-mgmt-compute` use the Apache HTTP client and the built-in Java XML parser, so it doesn't require Jersey JAXB bindings for databinding JSON/XML data to Java objects via old Jackson dependencies.


DaveCTurner (approved):
LGTM

henningandersen (approved):
LGTM 2

arteam:
Thank you both!

PR #87899, opened 2022-06-21
Use HashMap inside cluster info service

rjernst:
The InternalClusterInfoService internally uses ImmutableOpenMap
for keeping track of available space on each node. This commit converts
those usages to use HashMap. Note that an unmodifiableMap wrapper is
used because updates to this (from each node) are likely to happen often
as disk is used.

relates #86239


dakrone (approved):
LGTM

PR #87900, opened 2022-06-21
Merging both String Utility classes on EQL

LeeTgk:
As stated at the issue #87610.
It would be great to merge StringUtils together or categorise them and name accordingly.

This PR tries to adress 2 tasks of the whole Issue by merging ``org.elasticsearch.xpack.eql.expression.function.scalar.string.StringUtils`` into ``org.elasticsearch.xpack.eql.util.StringUtils``, since they are both part of EQL. 
I know it's rather simple but i think it's progress towards the whole issue.

The reason for the direction of the merge was because ``eql.util`` is more general than ``expression.function.scalar.string.StringUtils``.
I didn't found any conflicts with it's uses, or tests after the merge was done, but please, correct me if i'm wrong.


LeeTgk:
May i have some feedback on this ? :worried:

costin:
Sorry this took so long to respond.

Thank you for the contribution however please be mindful of this section of the [contributing guide](https://github.com/elastic/elasticsearch/blob/main/CONTRIBUTING.md#reviewing-and-accepting-your-contribution) as reviewing PRs carries a cost:

>We normally immediately reject isolated PRs that only perform simple refactorings or otherwise "tidy up" certain aspects of the code. We think the benefits of this kind of change are very small, and in our experience it is not worth investing the substantial effort needed to review them. 
>This especially includes changes suggested by tools.

PR #87902, opened 2022-06-21
Throw an error when parent object is not found while parsing field

javanna:
When parsing fields from incoming documents, we check the dynamic settings of their parent objects, effectively going up the tree structure until we found an object that set a specific dynamic value. The object mapper of the parent could either already be part of the mappings, or part of the dynamic mappers to be mapped as a result of indexing the current document. In this last case, we have a conditional to cover for a situation where the parent is not found, which should never hold.

This commit replaces the leniency around the previous conditional , which simply went out of the loop, with throwing an illegal state exception so that issues in this area are surfaced.


romseygeek (approved):
LGTM

PR #87904, opened 2022-06-22
[DOCS] Add authorization info to get anomaly detection jobs API

lcawl:
Relates to https://github.com/elastic/elasticsearch/issues/87905

This PR adds the `authorization` property to the output for the get anomaly detection jobs API.

### Preview

https://elasticsearch_87904.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/ml-get-job.html

### DIFF
--- a/docs/reference/ml/anomaly-detection/apis/get-job.asciidoc
+++ b/docs/reference/ml/anomaly-detection/apis/get-job.asciidoc
@@ -89,7 +89,9 @@ property is informational; you cannot change its value.
 ====
 `authorization`:::
 (Optional, object)
-The security privileges that the {dfeed} uses to run its queries.
+The security privileges that the {dfeed} uses to run its queries. If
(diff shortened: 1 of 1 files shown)


docs/reference/ml/anomaly-detection/apis/get-job.asciidoc
```diff
@@ -86,6 +87,36 @@ property is informational; you cannot change its value.
 .Properties of `datafeed_config`
 [%collapsible%open]
 ====
+`authorization`:::
+(Optional, object)
+The security privileges that the {dfeed} uses to run its queries.
```
droberts195, line 92: Maybe add something like:

> If security was disabled at the time of the most recent update to the {dfeed}
> then this field will not be present.

droberts195 (approved):
LGTM

PR #87921, opened 2022-06-22
Introduce tracing interfaces

pugnascotia:
Part of #84369. Split out from #87696. Introduce tracing interfaces in advance of adding APM support to Elasticsearch. The only implementation at this point is a no-op class.


pugnascotia:
@rjernst I think this is ready for review. We can continue the discussion from the PR #87696 here.

pugnascotia:
@original-brownbear I'd appreciate your thoughts on this PR, it's a precursor to #88443, which I asked you about yesterday.

server/src/main/java/org/elasticsearch/http/DefaultRestChannel.java
```diff
@@ -82,66 +80,87 @@ protected BytesStream newBytesOutput() {
...
     @Override
     public void sendResponse(RestResponse restResponse) {
-        // We're sending a response so we know we won't be needing the request content again and release it
-        httpRequest.release();
-
-        final ArrayList<Releasable> toClose = new ArrayList<>(3);
-        if (HttpUtils.shouldCloseConnection(httpRequest)) {
-            toClose.add(() -> CloseableChannel.closeChannel(httpChannel));
-        }
+        try (var ignored = tracer.withScope(this)) {
```
original-brownbear, line 83: What is this actually going to be used for? Could you add some javadoc to `withScope` maybe just so reviewers can understand what the lifecyle needs of this thing are?
If we can help it I would definitely want to avoid adding this kind of complexity to the already complex resource management in this method.

original-brownbear (commented):
I gave this a quick read and added a few questions that would help with giving better feedback here.

server/src/main/java/org/elasticsearch/http/DefaultRestChannel.java
pugnascotia: I took this wrapping out. I added Javadoc to the `withScope()` method that explains when you might want to use it, but it's not import here.

server/src/main/java/org/elasticsearch/rest/RestChannel.java
```diff
@@ -10,15 +10,19 @@
...
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
 
 /**
  * A channel used to construct bytes / builder based outputs, and send responses.
  */
-public interface RestChannel {
+public interface RestChannel extends Traceable {
```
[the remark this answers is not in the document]
pugnascotia: I've tried to simplify this by moving the default implementations to `AbstractRestChannel`, which is a better home for them anyway. Does that help?

server/src/main/java/org/elasticsearch/rest/RestChannel.java
```diff
@@ -39,4 +43,46 @@ XContentBuilder newBuilder(@Nullable XContentType xContentType, @Nullable XConte
...
+        attributes.put("http.method", req.method().name());
+        attributes.put("http.url", req.uri());
+        switch (req.getHttpRequest().protocolVersion()) {
+            case HTTP_1_0 -> attributes.put("http.flavour", "1.0");
+            case HTTP_1_1 -> attributes.put("http.flavour", "1.1");
+        }
+        return attributes;
+    }
+
+    void setTracePath(String path);
```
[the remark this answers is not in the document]
pugnascotia: When tracing REST requests, APM typically names spans and transactions using a concatenation of the HTTP method and request path. However, it's better for the name to be low cardinality, so if you're using path variables, you want to use paths that still have the `/{var}` placeholders. This trace path stuff is all about getting access to the path without variables.

server/src/main/java/org/elasticsearch/tracing/Tracer.java
```diff
@@ -0,0 +1,80 @@
...
+     * Called when the {@link Traceable} activity ends.
+     * @param traceable the thing to stop tracing
+     */
+    void onTraceStopped(Traceable traceable);
+
+    void onTraceEvent(Traceable traceable, String eventName);
+
+    void onTraceException(Traceable traceable, Throwable throwable);
+
+    void setAttribute(Traceable traceable, String key, boolean value);
```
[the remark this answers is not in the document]
pugnascotia: I've added more class-level Javadoc to this class, which I hope covers this.

server/src/main/java/org/elasticsearch/tracing/Traceable.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.elasticsearch.tracing;
+
+import java.util.Map;
+
+/**
+ * Something which maps onto a <i>span</i> in a distributed trace.
+ */
+public interface Traceable {
```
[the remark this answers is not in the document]
pugnascotia: The abstraction was introduced in the original SpaceTime project. TBH I never thought about changing it.

We have this abstraction because we won't stop with tracing tasks. For example, in https://github.com/pugnascotia/elasticsearch/commit/06d6105ed10bc51fd8b507c34f2969e976e5c445 (and bear in mind that is just a quickly hacked-together example, hence it does some questionable things), we're tracing `FetchSubPhaseProcessor` instances, and there's quite a few implementations of that interface. It's not difficult to imagine that each span might want to add different attributes to the span, and it would be better for that logic to live on the individual implementations that rather in the `FetchPhase` class.

It's also important the the logic for the span ID is the same when opened and closing spans, so having it in once place via the interface ensures this.

Having said all that, I would be particularly interested for us to add tracing to e.g. repository API requests (S3), so I would be very interested to know whether you think this API could work for that, or if we need to make some adjustments. Would you be able to do that?

server/src/main/java/org/elasticsearch/rest/RestChannel.java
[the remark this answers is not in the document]
original-brownbear: Not better by much I'm afraid. My main concern is simply with the fact that the `RestChannel` as an abstraction is already kind of broken and on the list of things to refactor away. 

With this change we get a lot less flexible in fixing the abstractions here. If you just inline the logic from here to where we call the tracer on the other hand, the change has zero negative impact on the flexibility we have in refactoring the REST layer.
Looking at this from a different perspective: There really isn't a clear definition of the lifecycle of a `RestChannel` today so either this doesn't need to live on the rest channel and is just needless complexity or it introduces new subtle requirements on its lifecycle. Just inlining the code and making it completely independent of the channel instances avoids trouble either way.

server/src/main/java/org/elasticsearch/tracing/Traceable.java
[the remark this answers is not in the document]
original-brownbear: > It's also important the the logic for the span ID is the same when opened and closing spans, so having it in once place via the interface ensures this.

But it's not really "one place". Now we're forcing new implicit life-cycle requirements on things like the rest-channel (which to me seems the worst one from this perspective) and making our whole codebase less flexible because any change to functionality that implements this interface now has to account for the tracing logic.

If you build the three objects here in-line where they're actually used you avoid complicating the codebase like that and we retain all flexibility. For the specific case of the fetch phase we could still add an abstraction to cover different implementations maybe but for all the other spots forcing this interface on top of random existing classes will increase the cost of future changes needlessly as far as I can tell.
If you think it's easiest to piggy back onto existing classes at this point, a better solution to this may be to just have `RestChannel` or `Task` return a `Traceable` via a getter without introducing a common abstraction across these very different classes. Then if we want to change or remove any of these we can just pull the `Traceable` out of them and pass it around some other way pretty easily? (also doing this `Traceable` could just become a `record` and doesn't even need to be an interface maybe because we don't have any state associated with an instance anyway as its getters only?)

original-brownbear (commented):
Thanks for the responses Rory,  I think I understand the ideas here better now. I tried to come up with an alternative suggestion here that retains the current level of flexibility in the code. Let me know what you think :)

pugnascotia:
@original-brownbear I experimented with dropping `Traceable` entirely in f0f0d773585, please take a look and see what you think.

pugnascotia:
Argh, I fumbled the commit, please see 00d4b50eda1 instead.

original-brownbear (commented):
Thanks Rory this is pretty much exactly what I had in mind! The approach now LGTM :)

server/src/main/java/org/elasticsearch/http/DefaultRestChannel.java
```diff
@@ -93,6 +97,11 @@ public void sendResponse(RestResponse restResponse) {
         boolean success = false;
         String opaque = null;
         String contentLength = null;
+        final Runnable onFinish = () -> {
+            Releasables.close(toClose);
+            tracer.onTraceStopped(traceId);
```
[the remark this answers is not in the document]
pugnascotia: I have no idea, it's so obvious now you've pointed it out 🤷‍♂️

server/src/main/java/org/elasticsearch/rest/RestController.java
```diff
@@ -403,25 +409,78 @@ private static boolean isContentTypeDisallowed(RestRequest request) {
...
+            // Invalid methods throw an exception
+        }
+        String name;
+        if (method != null) {
+            name = method + " " + restPath;
+        } else {
+            name = restPath;
+        }
+
+        Map<String, Object> attributes = new HashMap<>();
```
original-brownbear, line 461: Maybe pre-size this correctly e.g. using `Maps.` utilities? This isn't on the super hot path but sill it's on a network thread :)

server/src/test/java/org/elasticsearch/rest/RestControllerTests.java
```diff
@@ -710,9 +746,82 @@ public Exception getInboundException() {
         assertThat(channel.getRestResponse().getHeaders().get("Allow"), hasItem(equalTo(GET.toString())));
     }
 
+    /**
+     * Check that when dispatching a request, if an IllegalArgumentException is thrown, then a trace span is started
+     * and the exception is captured in the span.
+     */
+    public void testDispatchUnsupportedHttpMethodTracesException() {
+        final RestRequest request = RestRequest.request(parserConfig(), new HttpRequest() {
```
original-brownbear, line 754: Can't we override `FakeHttpRequest` here? Seems we only want the `method()` override right?

server/src/test/java/org/elasticsearch/tasks/TaskManagerTests.java
```diff
@@ -260,6 +272,95 @@ public void testTaskAccounting() {
...
+                    return null;
+                }
+
+                @Override
+                public TaskId getParentTask() {
+                    return TaskId.EMPTY_TASK_ID;
+                }
+            },
+            null,
+            new ActionListener<>() {
```
original-brownbear, line 349: If you want this shorter, you can use `ActionTestUtils.assertNoFailure(r -> {})`

server/src/main/java/org/elasticsearch/tasks/TaskManager.java
```diff
@@ -137,10 +146,21 @@ public Task register(String type, String action, TaskAwareRequest request) {
...
             Task previousTask = tasks.put(task.getId(), task);
             assert previousTask == null;
+            startTrace(threadContext, task);
         }
         return task;
     }
 
+    private void startTrace(ThreadContext threadContext, Task task) {
+        TaskId parentTask = task.getParentTaskId();
+        Map<String, Object> attributes = new HashMap<>();
```
original-brownbear, line 156: We could pre-size the map correctly here or just use `Map.of(` (with a slight bit of repetition for the task id key but whatever) to save some memory and make this compile nicer since its on the hot path.

server/src/main/java/org/elasticsearch/http/DefaultRestChannel.java
```diff
@@ -85,14 +89,18 @@ public void sendResponse(RestResponse restResponse) {
         // We're sending a response so we know we won't be needing the request content again and release it
         httpRequest.release();
 
+        final String traceId = "rest-" + this.request.getRequestId();
+
         final ArrayList<Releasable> toClose = new ArrayList<>(3);
```
original-brownbear, line 94: can make this a `4` now :)

server/src/main/java/org/elasticsearch/rest/RestController.java
```diff
@@ -403,25 +409,78 @@ private static boolean isContentTypeDisallowed(RestRequest request) {
...
             }
         }
         return false;
     }
 
+    private void startTrace(ThreadContext threadContext, RestChannel channel) {
+        startTrace(threadContext, channel, null);
+    }
+
+    private void startTrace(ThreadContext threadContext, RestChannel channel, String restPath) {
```
original-brownbear, line 443: We can pass the `RestRequest` here, we don't need the channel I think (and should have the request available in a variable at most callsites already anyway.
Mainly asking for this since I noticed we already have some strange redundancy in caller methods where we pass both the channel and the request (which we also have on the channel) which is confusing.

server/src/main/java/org/elasticsearch/tracing/Tracer.java
```diff
@@ -0,0 +1,170 @@
...
+ * You can open a span using
+ * {@link #onTraceStarted(ThreadContext, String, String, Map)}, and stop it using
+ * {@link #onTraceStopped(String)}, at which point the tracing system will queue
+ * the data to be sent somewhere for processing and storage.
+ * <p>
+ * You can add additional data to a span using the {@code setAttribute(...)}
+ * methods, e.g.  {@link #setAttribute(String, String, String)}. This allows you
+ * to attach data that is not available when the span is opened.
+ */
+public interface Tracer {
```
original-brownbear, line 31: I think we should add a comment here that the methods must have performance characteristics similar to logging since they're all used on hot transport threads.
If we implement this interface to send out network requests on any method we need to make sure that it's all non-blocking and with appropriate queuing to avoid slowing down transport threads.

original-brownbear (commented):
Some comments on code details and one on docs, looks good otherwise!

original-brownbear (approved):
LGTM if CI is happy, though Ryan should probably have a look as well to make sure he's good with the overall approach also.
Thanks for the extra iterations here!

server/src/main/java/org/elasticsearch/plugins/Plugin.java
```diff
@@ -205,4 +206,9 @@ public void close() throws IOException {
     public Collection<IndexSettingProvider> getAdditionalIndexSettingProviders(IndexSettingProvider.Parameters parameters) {
         return Collections.emptyList();
     }
+
+    /**
+     * Called with a Tracer so that each plugin has a chance to work with it.
+     */
+    public void onTracer(Tracer tracer) {}
```
rjernst, line 213: "onXXX" doesn't really fit with other plugin methods (the one exception is onIndexModule, but that is a holdover from the older style push based model, which index specific plugin extensions still use). The purpose of this seems to be to give plugins this global Tracer object, so that they can stash it to use later.

If the tracer object needs to be passed around the entire system, then it doesn't seem any better than just having actual global state for this purpose. Otherwise every time we realize something else needs access to the tracer, we have to plumb it through, which it seems like could eventually be the entire system? If I've misread the purpose of this object, then it would seem better as an argument to createComponents, where other "common" systems are passed to plugins so they can be utilized in plugin service construction.

server/src/main/java/org/elasticsearch/node/Node.java
```diff
@@ -1068,6 +1078,16 @@ protected Node(
         }
     }
 
+    private Tracer getTracer(Collection<Object> pluginComponents) {
+        final List<Tracer> tracers = pluginComponents.stream().map(c -> c instanceof Tracer t ? t : null).filter(Objects::nonNull).toList();
```
rjernst, line 1082: Thanks to java modules, we have a newer way of achieving this type of extensibility that we only intend a single implementation. See RestInterceptorActionPlugin. Basically you would have a specific interface for creating the tracer, and that package would only be exported to the module implementing it. Since this implementation is coming in a followup PR, it could be in a new unqualified export in this PR, then make it qualified in the followup.

server/src/main/java/org/elasticsearch/tracing/Tracer.java
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * Called when a span starts.
+     * @param threadContext the current context. Required for tracing parent/child span activity.
+     * @param id a unique identifier for the activity, and will not be sent to the tracing system. Add the ID
+     *           to the attributes if it is important
+     * @param name the name of the span. Sent to the tracing system
+     * @param attributes arbitrary key/value data for the span. Sent to the tracing system
+     */
+    void onTraceStarted(ThreadContext threadContext, String id, String name, Map<String, Object> attributes);
```
rjernst, line 47: nit: Could all these trace methods be named without the "on" prefix? That is typically for listeners or visitors, but here the caller is making an active choice to start a tracing event right? Could these be "startTrace" and "stopTrace"?

server/src/main/java/org/elasticsearch/tracing/Tracer.java
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * Some tracing implementations support the concept of "events" within a span, marking a point in time during the span
+     * when something interesting happened. If the tracing implementation doesn't support events, then nothing will be recorded.
+     * This should only be called when a trace already been started on the {@code traceable}.
+     * @param id an identifier for the span
+     * @param eventName the event that happened. This should be something meaningful to people reviewing the data, for example
+     *                  "send response", "finished processing", "validated request", etc.
+     */
+    void onTraceEvent(String id, String eventName);
```
rjernst, line 63: Related to the comment above on naming, this could just be "traceEvent" and the next one "traceException"?

rjernst (commented):
Thanks for splitting this out from the larger PR. I left a few thoughts.

server/src/main/java/org/elasticsearch/http/AbstractHttpServerTransport.java
```diff
@@ -407,7 +411,7 @@ private void handleIncomingRequest(final HttpRequest httpRequest, final HttpChan
             restRequest = innerRestRequest;
         }
 
-        final HttpTracer trace = tracer.maybeTraceRequest(restRequest, exception);
+        final HttpTracer maybeHttpTracer = httpTracer.maybeTraceRequest(restRequest, exception);
```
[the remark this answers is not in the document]
pugnascotia: I agree, it is confusing, and for a while I have `HttpTracer` implementing `Tracer`, but I ended up taking it as part of simplifying some stuff with Armin. The fundamental problem is that `HttpTracer` is actually just a logger, but we reference it in our docs so renaming the class isn't 100% trivial. However I've renamed many of the uses / instances of `HttpTracer` so make it clearer that it's about logging.

server/src/main/java/org/elasticsearch/tracing/Tracer.java
pugnascotia: I renamed the methods.

pugnascotia:
@rjernst I'm not sure if this is exactly what you intended, but: I introduced a `TracerPlugin` interface, which is a marker for plugins that can create a Tracer. This allows me to create a tracer and pass it into `createComponents()`, instead of waiting until `createComponents()` is called on all the plugins and then pulling it out of the components.

The `TracerPlugin` class could later have its visibility restricted. Does that work?

rjernst:
> The TracerPlugin class could later have its visibility restricted. Does that work?

Yes, the TracerPlugin + createComponents changes are what I was thinking there.

server/src/main/java/org/elasticsearch/tracing/Tracer.java
rjernst: Thanks for the renames. However, it didn't quite address my concern. The "traceStarted" and "traceStopped" still don't sound right. These methods are actions, yet those method names imply some notificications. I think "startTrace" and "stopTrace" make more sense? The other methods are good as "traceException" and "traceEvent", since those are also action, using the trace verb.

server/src/main/java/org/elasticsearch/http/AbstractHttpServerTransport.java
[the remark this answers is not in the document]
rjernst: Thanks for the rename of the member, it does help in the reading. However it is still confusing since the class is still called HttpTracer.

> we reference it in our docs so renaming the class isn't 100% trivial.

I think this could be handled through keeping the HttpTracer class, but with no impl, so that whatever the class is renamed to can still have `LogManager.getLogger(HttpTracer.class)`. We do need a way deprecate such logging; it is an existing problem that our class and package names leak into logger configuration since it restricts us from easily changing, or we accidentally make a configuration breaking change without realizing it.

The rename to HttpLogger I suggest above isn't crucial for this PR, but I do think it should be done as a followup.

rjernst (approved):
LGTM assuming you agree with "startTrace" and "stopTrace"

pugnascotia:
> LGTM assuming you agree with "startTrace" and "stopTrace"

Oh man, I meant to rename them that way before. 🤦‍♂️

PR #87922, opened 2022-06-22
fix RareClusterStateIT

idegtiarenko:


server/src/internalClusterTest/java/org/elasticsearch/cluster/coordination/RareClusterStateIT.java
```diff
@@ -140,7 +141,12 @@ private <Req extends ActionRequest, Res extends ActionResponse> ActionFuture<Res
             }
         });
         ActionFuture<Res> future = req.execute();
-        assertBusy(() -> assertTrue(masterCoordinator.cancelCommittedPublication()));
+        // cancel all commit publications produced by the request
+        internalCluster().getCurrentMasterNodeInstance(ThreadPool.class).generic().execute(() -> {
+            while (!future.isDone()) {
```
DaveCTurner, line 146: I'm not sure that's enough - IIRC deleting an index may trigger another reroute but it doesn't wait for that reroute to complete.

DaveCTurner (commented):
Looks good, but should we make this change in `master` first?

idegtiarenko:
>Looks good, but should we make this change in master first?

Okay, let me change the base for this pr

idegtiarenko:
Done, this is now targeting master

DaveCTurner (approved):
Nice, LGTM

PR #87923, opened 2022-06-22
Fix rest example plugin

rjernst:
This is a followup to
https://github.com/elastic/elasticsearch/pull/87504, to fix the example
plugin that used BytesRestResponse.


mark-vieira (approved):
LGTM. Thanks, Ryan.

Issue #87870, opened 2022-06-21
Support updates of API key attributes (single operation route) 


n1v0lg:
Today, [API key](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html) attributes, including metadata and roles, are immutable, i.e., cannot be changed after an API key has been created.

There are use cases for API keys in which both the metadata and, importantly, the roles associated with an API key change. The current approach to solve this is rotation: invalidating the old key, creating a new one with the desired attributes, and replacing the use of the old key. This can be both cumbersome and performance intensive.

This issue proposes to implement a new route `PUT /_security/api_key/{id}` through which the following attributes of an API key can be updated:

* `role_descriptors` - The list of role descriptors specified for the key. This is one of the two parts that determines an API key’s privileges.
* `metadata_flattened` - The searchable metadata associated to an API key

Several other attributes will be automatically updated for each call, including:

* `limited_by_role_descriptors` - The snapshot of owner user’s privilege. This is the other part that determines an API key’s privileges.
* `creator` - The snapshot of owner user’s information, including `principal`, `full_name`, `email`, `metadata`, `realm` and `domain`.
* `version` - This field represents the minimal node version when the API key is first created. It is not currently used. Updating an API key should also bring its format to the latest version, e.g. adding metadata which is not available before v7.13. Therefore we should also update this field to the current minimum node version.

Only the creator of an API key is authorized to update it, provided the `manage_own_api_key` privilege. 

This breaks down into the following tasks:

- [x] Service level implementation (`ApiKeyService`), including caching and automatic format updates for version changes 
- [x] REST and transport layer, including permissions

PR #87924, opened 2022-06-22
Support updates of API key attributes [service layer]

n1v0lg:
Service level implementation to add support for updating attributes of
existing API keys. This allows end-users to modify privileges and
metadata associated with API keys dynamically, without requiring
rolling out new API keys every time there is a change.

Updatable attributes are `role_descriptors` and `metadata`. Several
other attributes are updated automatically, on every update call,
including `limited_by_role_descriptors`, `creator`, and `version`. API
key attributes are replaced, _not_ merged.

On every update, the API key doc cache is cleared for the updated API
key.

This PR implements the necessary service layer changes in
`ApiKeyService`. I will integrate this with the REST and transport
layers in a subsequent PR.

Relates: #87870

Note: labeling `>non-issue` since I would rather include a `>feature`
tag and changelog entry on the REST & transport layer PR.


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -332,6 +339,146 @@ private void createApiKeyAndIndexIt(
...
+        Authentication authentication,
+        UpdateApiKeyRequest request,
+        Set<RoleDescriptor> userRoles,
+        Version version,
+        ApiKeyDocWithSeqNoAndPrimaryTerm apiKeyDocWithSeqNoAndPrimaryTerm
+    ) throws IOException {
+        final var apiKeyDoc = apiKeyDocWithSeqNoAndPrimaryTerm.apiKey();
+        final var newRoleDescriptors = request.getRoleDescriptors();
+        // TODO feels backwards that we deserialize just to serialize again
+        final var keyRoles = newRoleDescriptors != null
```
n1v0lg, line 451: This part I'm not sure about:

It seems odd to take the already serialized roles from the doc, deserialize them, then serialize them in `newDocument`. Same for metadata.

Ideally, I would just take the `ByteReference` instances and directly include them in the source but I haven't found an appropriate way to do this yet.

Wondering if you had suggestions, or whether you think the current approach is fine.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -985,6 +1125,23 @@ public static QueryBuilder filterForRealmNames(String[] realmNames) {
         }
     }
 
+    private void findApiKeyDocsForSubject(
+        Subject subject,
+        String[] apiKeyIds,
+        ActionListener<Collection<ApiKeyDocWithSeqNoAndPrimaryTerm>> listener
+    ) {
+        findApiKeysForUserRealmApiKeyIdAndNameCombination(
+            new String[] { subject.getRealm().getName() },
```
n1v0lg, line 1134: Ah this is wrong. Fixing.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
ywangd: You can use something like the follows:
```
builder.rawField("role_descriptors", apiKeyDoc.roleDescriptorsBytes, XContentType.JSON);
```

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -332,6 +353,135 @@ private void createApiKeyAndIndexIt(
...
+            throw new ValidationException().addValidationError("cannot update legacy api key [" + apiKeyId + "] without name");
+        }
+    }
+
+    private boolean isActive(ApiKeyDoc apiKeyDoc) {
+        return apiKeyDoc.invalidated == false
+            && (apiKeyDoc.expirationTime == -1 || Instant.ofEpochMilli(apiKeyDoc.expirationTime).isAfter(clock.instant()));
+    }
+
+    private void translateResponseAndClearCache(String apiKeyId, BulkResponse bulkResponse, ActionListener<UpdateApiKeyResponse> listener) {
```
n1v0lg, line 408: I'm not happy with the name but blanking on anything better.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -332,6 +353,135 @@ private void createApiKeyAndIndexIt(
...
+
+        if (Version.fromId(apiKeyDoc.version).before(Version.V_8_2_0)) {
+            throw new ValidationException().addValidationError(
+                "cannot update legacy api key [" + apiKeyId + "] with version [" + Version.fromId(apiKeyDoc.version) + "]"
+            );
+        }
+        if (isActive(apiKeyDoc) == false) {
+            throw new ValidationException().addValidationError("cannot update inactive api key [" + apiKeyId + "]");
+        }
+        if (Strings.isNullOrEmpty(apiKeyDoc.name)) {
```
n1v0lg, line 398: I don't think we should support auto-upgrading API keys without names -- we have an implicit assumption that an 8.x API key has a name. Auto-generating a name or using the ID as the name seems like too much of a BWC overreach. Let me know if you feel strongly this.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/apikey/UpdateApiKeyResponse.java
```diff
@@ -0,0 +1,66 @@
...
+import org.elasticsearch.common.io.stream.StreamOutput;
+import org.elasticsearch.common.io.stream.Writeable;
+import org.elasticsearch.xcontent.ToXContentObject;
+import org.elasticsearch.xcontent.XContentBuilder;
+
+import java.io.IOException;
+import java.util.Objects;
+
+public final class UpdateApiKeyResponse extends ActionResponse implements ToXContentObject, Writeable {
+    private final String id;
```
n1v0lg, line 21: We didn't include the id in our design doc but it makes the code slightly cleaner and seems like a reasonable thing to have on the response.

ywangd (commented):
Here are some comments on the main code. I still need time to read the test code. 

Also a general style preference issue, you can consider this as a Nit: I am all for for splitting large methods into multiple, smaller, well-named methods. But if a method isn't too large, e.g. it can roughly fit into a single page when reading from the browser or IDE, and/or the splitted method is not reused, I find it easier to read without splitting. Otherwise, it makes the reader jumping back and forth for small things and the overall reading experience is less fluent. The code is also cluttered with repetition of parameters and corresponding arguments (often occupying multiple lines). For example, I find it unnecessary to have a separate `executeBulkIndexRequest`. I'd appreciate if it is inlined.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/apikey/UpdateApiKeyResponse.java
n1v0lg: Good point re. the inconsistency with the GetApi -- the part where this gets a bit cleaner is with handling the cache. There, we clear based on API key ID and the response object. If the response object already has the ID on it, we don't have to pass a separate argument. I don't feel strongly about this though. Will remove id from the response.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -985,6 +1184,23 @@ public static QueryBuilder filterForRealmNames(String[] realmNames) {
         }
     }
 
+    private void findVersionedApiKeyDocsForSubject(
+        Authentication authentication,
+        String[] apiKeyIds,
+        ActionListener<Collection<VersionedApiKeyDoc>> listener
+    ) {
+        findApiKeysForUserRealmApiKeyIdAndNameCombination(
+            getOwnersRealmNames(authentication),
```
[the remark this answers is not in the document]
n1v0lg: > We will likely have the "not-an-api-key" check at REST or transport level.

Exactly, that's what I had envisioned. Good with also catching it at the service layer. What about including the check early on in `updateApiKey`, something like:

```java
public void updateApiKey(
        UpdateApiKeyRequest request,
        Set<RoleDescriptor> userRoles,
        ActionListener<UpdateApiKeyResponse> listener
    ) {
        ensureEnabled();

        if (authentication == null) {
            listener.onFailure(new IllegalArgumentException("authentication must be provided"));
            return;
        }
        if (authentication.isApiKey()) {
           listener.onFailure(...)
        }
        ...

```

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -332,6 +353,135 @@ private void createApiKeyAndIndexIt(
...
+        VersionedApiKeyDoc currentVersionedDoc,
+        Authentication authentication,
+        UpdateApiKeyRequest request,
+        Set<RoleDescriptor> userRoles,
+        Version version
+    ) throws IOException {
+        return client.prepareIndex(SECURITY_MAIN_ALIAS)
+            .setId(request.getId())
+            .setSource(
+                updatedDocument(
```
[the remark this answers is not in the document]
n1v0lg: I went with `updatedDocument` to mirror `newDocument` -- `update` implies that we are actually updating the existing doc, which we are not. Maybe `buildUpdatedDocument`?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -346,13 +496,77 @@ static XContentBuilder newDocument(
...
+        } else {
+            assert currentApiKeyDoc.roleDescriptorsBytes != null;
+            builder.rawField("role_descriptors", currentApiKeyDoc.roleDescriptorsBytes.streamInput(), XContentType.JSON);
+        }
+
+        addLimitedByRoleDescriptors(builder, userRoles);
+
+        builder.field("name", currentApiKeyDoc.name).field("version", version.id);
+        if (metadata != null) {
+            builder.field("metadata_flattened", metadata);
```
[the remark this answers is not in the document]
n1v0lg: Good call! Will add an assertion.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -332,6 +353,135 @@ private void createApiKeyAndIndexIt(
...
+                buildBulkIndexRequestForUpdate(versionedDocs, authentication, request, userRoles),
+                ActionListener.wrap(bulkResponse -> translateResponseAndClearCache(apiKeyId, bulkResponse, listener), listener::onFailure)
+            );
+        }, listener::onFailure));
+    }
+
+    // package-private for testing
+    void validateCurrentApiKeyDocForUpdate(String apiKeyId, Authentication authentication, ApiKeyDoc apiKeyDoc) {
+        assert authentication.getEffectiveSubject().getUser().principal().equals(apiKeyDoc.creator.getOrDefault("principal", null));
+        assert authentication.getEffectiveSubject().getRealm().getName().equals(apiKeyDoc.creator.getOrDefault("realm", null));
```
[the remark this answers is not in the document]
n1v0lg: Good catch! 

> This actually reminded that API keys will have the same "domain splitting" issue as user profiles and we should update this [doc page](https://www.elastic.co/guide/en/elasticsearch/reference/master/security-domain.html#security-domain-remove-realm) before this feature goes GA.

Will track via a GH task.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -332,6 +353,135 @@ private void createApiKeyAndIndexIt(
...
+            listener.onFailure(new IllegalArgumentException("authentication must be provided"));
+            return;
+        }
+
+        findVersionedApiKeyDocsForSubject(authentication, new String[] { request.getId() }, ActionListener.wrap((versionedDocs) -> {
+            final var apiKeyId = request.getId();
+
+            if (versionedDocs.isEmpty()) {
+                throw apiKeyNotFound(apiKeyId);
+            }
```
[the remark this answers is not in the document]
n1v0lg: Agreed that the error message is too generic and will be confusing.

I think a 404 can still be appropriate here though since we are looking for a resource by its ID with extra filter parameters of owner name and realm name and are not finding it.

I'm good with "API key not found or not owned by the requesting user". Alternatively, wondering about `API key owned by requesting user not found` since it more explicitly describes the resource we were unable to find?

I'd rather avoid the extra `GetApiKey` call -- it seems overkill just to construct a message.

n1v0lg:
> But if a method isn't too large, e.g. it can roughly fit into a single page when reading from the browser or IDE, and/or the splitted method is not reused, I find it easier to read without splitting. 

Good point, there is a balance to find here for sure. My reason for generally splitting is that using methods like `executeBulkIndexRequest` allows me to represent code at different levels of abstraction:

```
findVersionedApiKeyDocsForSubject(authentication, new String[] { request.getId() }, ActionListener.wrap((versionedDocs) -> {
    final var apiKeyId = request.getId();

    if (versionedDocs.isEmpty()) {
        throw apiKeyNotFound(apiKeyId);
    }

    validateCurrentApiKeyDocForUpdate(apiKeyId, authentication, single(versionedDocs).doc());

    executeBulkIndexRequest(
        buildBulkIndexRequestForUpdate(versionedDocs, authentication, request, userRoles),
        ActionListener.wrap(bulkResponse -> translateResponseAndClearCache(apiKeyId, bulkResponse, listener), listener::onFailure)
    );
}, listener::onFailure));
```

essentially communicates to the reader _without_ underlying implementation details that we are fetching a key, validating it, then creating a request to update it, then executing the request. It makes no mention of how we are executing it and if the reader is only interested in the high level flow, they can stop reading here. In-lining: 

```
securityIndex.prepareIndexIfNeededThenExecute(
            listener::onFailure,
            () -> executeAsyncWithOrigin(client.threadPool().getThreadContext(), SECURITY_ORIGIN, bulkRequest, listener, client::bulk)
        );
```

Leaks extra context that's much lower level that the reader at this stage might not care about.

It terms of re-use: I think the create API key code can actually re-use a big chunk of the new update code, including `executeBulkIndexRequest`. Was thinking about adding a follow up task to try to unify more of that code. 

I'm happy to in-line more, esp. since it's consistent with the rest of the ES codebase, just wanted to give my rationale for not in-lining.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -332,6 +352,115 @@ private void createApiKeyAndIndexIt(
...
+        final Set<RoleDescriptor> userRoles,
+        final ActionListener<UpdateApiKeyResponse> listener
+    ) {
+        ensureEnabled();
+
+        if (authentication == null) {
+            listener.onFailure(new IllegalArgumentException("authentication must be provided"));
+            return;
+        } else if (authentication.isApiKey()) {
+            listener.onFailure(new IllegalArgumentException("cannot use an api key to update api keys"));
```
n1v0lg, line 367: Still trying to come up with a reasonable message here. This one sounds clunky.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -332,6 +352,115 @@ private void createApiKeyAndIndexIt(
...
+        findVersionedApiKeyDocsForSubject(authentication, new String[] { request.getId() }, ActionListener.wrap((versionedDocs) -> {
+            final var apiKeyId = request.getId();
+
+            if (versionedDocs.isEmpty()) {
+                throw new ResourceNotFoundException("no api key owned by requesting user found for requested id [" + apiKeyId + "]");
+            }
+
+            validateCurrentApiKeyDocForUpdate(apiKeyId, authentication, single(versionedDocs).doc());
+
+            executeBulkIndexRequest(
```
n1v0lg, line 380: Following your suggestion, I've in-lined several other smaller methods. In terms of readability I have a slight preference to keep `executeBulkIndexRequest` though -- it keeps `updateApiKey` high level and essentially free of implementation details which I think is useful for someone just wanting a high level idea of what the method does. 

Let me know if you feel strongly!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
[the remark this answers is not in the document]
n1v0lg: Added as a task to the meta issue

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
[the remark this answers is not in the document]
ywangd: 👍

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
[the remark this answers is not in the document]
ywangd: I am OK with just expanding the error message. I also like your 2nd version. Unrelated: between "API key" and "api key", what do you prefer? I personally prefer the former but we haven't been consistent.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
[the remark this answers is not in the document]
ywangd: Nevermind. I forgot it meant to be an adjective rather than a verb.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
ywangd: Thanks for your consideration. The comment was close to a "Nit". So I don't feel strongly about it.

Like you said, there is balance to strike. I am _not_ advocating for long-winded methods. Depending on the reader, exposing some of details may or may not be useful. As a code reviewer, I found them useful since I need to understand every single detail. That's probably true most of the time for people in our immediate team. But not so much for anyone else. 

It also has something to do with ES's general callback style coding. Most split'd methods make your read flow to jump from the innermost of a method to a different place. This coupled with the burdens of ensuring correct threadContext, error handling and code repitition decrease the value of one-off smaller methods.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -1405,6 +1426,326 @@ public void testSecurityIndexStateChangeWillInvalidateApiKeyCaches() throws Exce
...
+        }
+    }
+
+    private Map<String, Object> getApiKeyDocument(String apiKeyId) {
+        final GetResponse getResponse = client().execute(GetAction.INSTANCE, new GetRequest(SECURITY_MAIN_ALIAS, apiKeyId)).actionGet();
+        return getResponse.getSource();
+    }
+
+    private ServiceWithNodeName getServiceWithNodeName() {
+        final var nodeName = internalCluster().getNodeNames()[0];
```
ywangd, line 1742: Nit: we can random from all node names.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -1405,6 +1426,326 @@ public void testSecurityIndexStateChangeWillInvalidateApiKeyCaches() throws Exce
...
+        return switch (caseNo) {
+            case 0 -> List.of(new RoleDescriptor(randomAlphaOfLength(10), new String[] { "all" }, null, null));
+            case 1 -> List.of(
+                new RoleDescriptor(randomAlphaOfLength(10), new String[] { "all" }, null, null),
+                RoleDescriptorTests.randomRoleDescriptor()
+            );
+            case 2 -> null;
+            default -> throw new IllegalStateException("unexpected case no");
+        };
+    }
```
ywangd, line 1511: I'd appreciate if we could move all private methods after public ones.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -1405,6 +1426,326 @@ public void testSecurityIndexStateChangeWillInvalidateApiKeyCaches() throws Exce
...
+
+        // Test not found exception on other user's API key
+        final var otherUsersApiKey = createApiKey("user_with_manage_api_key_role", null);
+        testUpdateApiKeyNotFound(
+            serviceWithNodeName,
+            fileRealmAuth(serviceWithNodeName.nodeName(), ES_TEST_ROOT_USER, ES_TEST_ROOT_ROLE),
+            new UpdateApiKeyRequest(otherUsersApiKey.v1().getId(), request.getRoleDescriptors(), request.getMetadata())
+        );
+
+        // Test not found exception on API key of user from other realm
```
ywangd, line 1550: Nit:
```suggestion
        // Test not found exception on API key of user with the same username but from a different realm
```

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -1405,6 +1426,326 @@ public void testSecurityIndexStateChangeWillInvalidateApiKeyCaches() throws Exce
...
+        final var apiKeysNotAllowedEx = expectThrows(ExecutionException.class, updateListener::get);
+
+        assertThat(apiKeysNotAllowedEx.getCause(), instanceOf(IllegalArgumentException.class));
+        assertThat(
+            apiKeysNotAllowedEx.getMessage(),
+            containsString("authentication via an api key is not supported for updating api keys")
+        );
+    }
+
+    private void testUpdateApiKeyNotFound(
```
ywangd, line 1613:
```suggestion
    private void doTestUpdateApiKeyNotFound(
```

for easier filtering in IDE. Also would be great to move after public methods.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -1405,6 +1426,326 @@ public void testSecurityIndexStateChangeWillInvalidateApiKeyCaches() throws Exce
...
+                nodeName,
+                includeDomain
+                    ? new RealmDomain(
+                        ESTestCase.randomAlphaOfLengthBetween(3, 8),
+                        Set.of(new RealmConfig.RealmIdentifier(realmType, realmName))
+                    )
+                    : null
+            )
+        );
+    }
```
ywangd, line 1704: Besides the realm authentication for the user itself, the update should also be possible for:
(1) A token of the user 
(2) Another user run as this user
(3) This username but from a different realm of the same domain

So I think this can use a bit more randomness. Using something like
```
AuthenticationTestHelper.builder().user(...).realmRef(...).build()
```
can address (1) and (2). For (3), we need to hand roll a random domain since the test helper does not provide one that is helpful here (or maybe we can add one test helper method to do that).

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -1405,6 +1426,326 @@ public void testSecurityIndexStateChangeWillInvalidateApiKeyCaches() throws Exce
...
+                new RoleDescriptor(randomAlphaOfLength(10), new String[] { "all" }, null, null),
+                RoleDescriptorTests.randomRoleDescriptor()
+            );
+            case 2 -> null;
+            default -> throw new IllegalStateException("unexpected case no");
+        };
+    }
+
+    public void testUpdateApiKeyNotFoundScenarios() throws ExecutionException, InterruptedException {
+        final var createdApiKey = createApiKey(ES_TEST_ROOT_USER, null);
```
ywangd, line 1514: Nit: Personally, this is one example where `var` does not help. The name `apiKey` is so heavily overloaded. By just looking at the variable name and the method name, I have no idea what the type is (even though I may have written the method, or part of it, initially -_-!). Explicit types would help here.

ywangd (commented):
This is almost ready. I don't have major points. Other than below comments, I also felt we are a bit light on debug loggings. For example, when the version number gets updated, I think it's worth for a logging message.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/action/apikey/UpdateApiKeyRequestTests.java
```diff
@@ -0,0 +1,57 @@
...
+        try (BytesStreamOutput out = new BytesStreamOutput()) {
+            request.writeTo(out);
+            try (StreamInput in = out.bytes().streamInput()) {
+                final var serialized = new UpdateApiKeyRequest(in);
+                assertEquals(id, serialized.getId());
+                if (roleDescriptorsPresent) {
+                    assertEquals(descriptorList, serialized.getRoleDescriptors());
+                } else {
+                    assertNull(serialized.getRoleDescriptors());
+                }
```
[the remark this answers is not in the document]
n1v0lg: Yup, I made several passes over this; `assertEquals` should indeed work. Thanks for the simplification!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -332,6 +352,115 @@ private void createApiKeyAndIndexIt(
...
+                            apiKeyDoc.doc(),
+                            authentication,
+                            userRoles,
+                            request.getRoleDescriptors(),
+                            version,
+                            request.getMetadata()
+                        )
+                    )
+                    .setIfSeqNo(apiKeyDoc.seqNo())
+                    .setIfPrimaryTerm(apiKeyDoc.primaryTerm())
```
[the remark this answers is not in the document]
n1v0lg: Good catch!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -346,13 +475,82 @@ static XContentBuilder newDocument(
...
+    }
+
+    static XContentBuilder buildUpdatedDocument(
+        ApiKeyDoc currentApiKeyDoc,
+        Authentication authentication,
+        Set<RoleDescriptor> userRoles,
+        List<RoleDescriptor> keyRoles,
+        Version version,
+        Map<String, Object> metadata
+    ) throws IOException {
```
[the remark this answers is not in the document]
n1v0lg: Agreed. I'll collect this and other refactoring opportunities in a follow up issue.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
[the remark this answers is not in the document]
n1v0lg: Going with `buildUpdatedDocument` because I think anyone ever looking at `updatedDoc` will assume it's a typo. No need to surprise people 👍

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -993,6 +1209,28 @@ private void findApiKeysForUserRealmApiKeyIdAndNameCombination(
...
+
+    private <T> void findApiKeysForUserRealmApiKeyIdAndNameCombination(
+        String[] realmNames,
+        String userName,
+        String apiKeyName,
+        String[] apiKeyIds,
+        boolean filterOutInvalidatedKeys,
+        boolean filterOutExpiredKeys,
+        ActionListener<Collection<T>> listener,
+        Function<SearchHit, T> hitParser
```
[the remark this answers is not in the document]
n1v0lg: Whoops sorry, I changed this for one method but overlooked this ones. BTW `ScrollHelper.fetchAllByEntity` has the same issue but I'd rather not touch that in this PR: it leads to a lot of changes in completely unrelated classes.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -993,6 +1194,28 @@ private void findApiKeysForUserRealmApiKeyIdAndNameCombination(
...
+            realmNames,
+            userName,
+            apiKeyName,
+            apiKeyIds,
+            filterOutInvalidatedKeys,
+            filterOutExpiredKeys,
+            listener,
+            ApiKeyService::convertSearchHitToApiKeyInfo
+        );
+    }
```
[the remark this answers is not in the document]
n1v0lg: Good call. In-lined.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/ApiKeyServiceTests.java
```diff
@@ -1639,6 +1642,118 @@ public void testApiKeyDocDeserialization() throws IOException {
...
+        final List<RoleDescriptor> newKeyRoles;
+        if (nullKeyRoles) {
+            newKeyRoles = null;
+        } else {
+            newKeyRoles = List.of(RoleDescriptorTests.randomRoleDescriptor());
+        }
+
+        final var metadata = ApiKeyTests.randomMetadata();
+        final var version = Version.CURRENT;
+        final var authentication = AuthenticationTestHelper.builder().user(new User("user", "role")).build(false);
```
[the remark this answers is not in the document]
n1v0lg: The username does matter here because otherwise we trip the username assertion in the validation. But I'm assuming that a variation of `randomValueOtherThanMany` should still work.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
n1v0lg: Since my plan is to update these tests to use the client and transport actions directly I think some of this might actually become redundant/would need to be tweaked since we will need "real" users. Would it be fine to address 1 and 2 here, because it's straight-forward but keep 3 for the next PR?

ywangd: Sounds good to me.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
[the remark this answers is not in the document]
n1v0lg: I'm good with the former! I picked lowercase at random as the first "style" I've seen but API should be capped.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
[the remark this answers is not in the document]
n1v0lg: Oh hm that's apparently not allowed:

`java.lang.IllegalArgumentException: opType must be 'create' or 'index', found: [UPDATE]`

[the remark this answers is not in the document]
n1v0lg: I'm going with `INDEX` instead to be explicit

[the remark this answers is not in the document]
ywangd: ah right. sorry for the mis-information.  The usage of `if_primary_term` and `if_seq_no` already guarantees it being an update.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
n1v0lg: Absolutely see the benefit of having more things in one place. For now, I want to leave the current style and revisit when we get to implementing bulk updates since the overlap and method re-use there will further inform and possibly change the overall structure of the code.

Quite open to more in-lining at that stage. Thanks for the discussion around this!

n1v0lg:
Build failure is unrelated and tracked [here](https://github.com/elastic/elasticsearch/issues/88118)

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -332,6 +353,60 @@ private void createApiKeyAndIndexIt(
...
+    void validateCurrentApiKeyDocForUpdate(String apiKeyId, Authentication authentication, ApiKeyDoc apiKeyDoc) {
+        assert authentication.getEffectiveSubject().getUser().principal().equals(apiKeyDoc.creator.get("principal"));
+
+        if (apiKeyDoc.invalidated) {
+            throw new IllegalArgumentException("cannot update invalidated API key [" + apiKeyId + "]");
+        }
+
+        final var now = clock.instant();
+        boolean expired = apiKeyDoc.expirationTime != -1
+            && (Instant.ofEpochMilli(apiKeyDoc.expirationTime).equals(now) || Instant.ofEpochMilli(apiKeyDoc.expirationTime).isBefore(now));
```
ywangd, line 400: Can this be 
```suggestion
            && now.isAfter(Instant.ofEpochMilli(apiKeyDoc.expirationTime));
```
?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -346,56 +421,72 @@ static XContentBuilder newDocument(
...
+        builder.startObject()
+            .field("doc_type", "api_key")
+            .field("creation_time", currentApiKeyDoc.creationTime)
+            .field("expiration_time", currentApiKeyDoc.expirationTime == -1 ? null : currentApiKeyDoc.expirationTime)
+            .field("api_key_invalidated", false);
+
+        addApiKeyHash(builder, currentApiKeyDoc.hash.toCharArray());
+
+        if (keyRoles != null) {
+            logger.trace("Building API key doc with updated role descriptors [{}]", keyRoles);
```
ywangd, line 459: Parameterized message can help here since serialising list of role descriptors can take measurable time:
```suggestion
            logger.trace(new ParameterizedMessage("Building API key doc with updated role descriptors [{}]", keyRoles));
```

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -346,56 +421,72 @@ static XContentBuilder newDocument(
...
+        addLimitedByRoleDescriptors(builder, userRoles);
+
+        builder.field("name", currentApiKeyDoc.name).field("version", version.id);
+
+        assert currentApiKeyDoc.metadataFlattened == null
+            || MetadataUtils.containsReservedMetadata(
+                XContentHelper.convertToMap(currentApiKeyDoc.metadataFlattened, false, XContentType.JSON).v2()
+            ) == false : "API key doc to be updated contains reserved metadata";
+        if (metadata != null) {
+            logger.trace("Building API key doc with updated metadata [{}]", metadata);
```
ywangd, line 475: Same here as well
```suggestion
            logger.trace(new ParameterizedMessage("Building API key doc with updated metadata [{}]", metadata));
```

ywangd (approved):
LGTM 

Thanks for the iterations!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
n1v0lg: Yup. I went for the other option to keep the "before" object to the left, but this is cleaner because it gets rid of the `||`. Thanks!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -332,6 +352,115 @@ private void createApiKeyAndIndexIt(
         }));
     }
 
+    public void updateApiKey(
+        final Authentication authentication,
+        final UpdateApiKeyRequest request,
+        final Set<RoleDescriptor> userRoles,
```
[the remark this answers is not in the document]
ywangd: Btw, I am _not_ suggestion having the refactoring as part of the "Updatable API keys" work stream. By "get back to it in future", I meant some point after we deliver the planned feature. So no need to bundle it in the PR for REST layer as it might add unnecessary complexities.

PR #87925, opened 2022-06-22
[ML] Use normalized memory in bin packing node prioritization

dimitris-athanasiou:
This commit fixes a bug in the `LinearProgrammingPlanSolver`
used for distributing model allocations. When we are doing the
initial bin-packing, we should be using normalized memory
as a factor. Instead, we were using actual memory which would
dominate the heuristic and result in always prioritize the
largest nodes in terms of memory and completely disregard existing
allocations.

The linear solver would still provide a good solution most of
the times but in some edge cases we'd end up with a bad solution.
This commit adds a test with an edge case like this.

In addition, we improve the `AssignmentPlanner` by trying out
both approaches (1. preserve one allocation per assignment,
2. preserve all allocations) when the first try does not give
a result that satisfies all models.


przemekwitek (approved):
LGTM

PR #87930, opened 2022-06-22
Speed up synthetic keyword, ip, and text fields

nik9000:
This speeds up synthesizing _source for keywords, text, and ip fields by
converting from ordinal to utf-8 one time. When loading from disk blocks
that are hot in the page cache the speed up is 7-10%:

```
|  50th percentile service time | default_1k | 39.4338 | 35.6935 | ms |  -9.49% |
|  90th percentile service time | default_1k | 41.3508 | 38.3602 | ms |  -7.23% |
|  99th percentile service time | default_1k | 50.6734 | 45.4686 | ms | -10.27% |
| 100th percentile service time | default_1k | 61.3771 | 56.4886 | ms |  -7.96% |
```

This should be much more substantial when loading from scattered disk
blocks because it takes care to load the ordinals in increasing order.

### DIFF
--- a/server/src/main/java/org/elasticsearch/search/fetch/FetchPhase.java
+++ b/server/src/main/java/org/elasticsearch/search/fetch/FetchPhase.java
@@ -120,25 +120,29 @@ public class FetchPhase {
         List<FetchSubPhaseProcessor> processors = getProcessors(context.shardTarget(), fetchContext, profiler);
         NestedDocuments nestedDocuments = context.getSearchExecutionContext().getNestedDocuments();
 
-        int currentReaderIndex = -1;
-        LeafReaderContext currentReaderContext = null;
+        List<LeafReaderContext> leaves = context.searcher().getIndexReader().leaves();
         LeafNestedDocuments leafNestedDocuments = null;
         CheckedBiConsumer<Integer, FieldsVisitor, IOException> fieldReader = null;
         boolean hasSequentialDocs = hasSequentialDocs(docs);
         SourceLoader.Leaf leafSourceLoader = null;
+        int leafIndex = -1;
+        LeafReaderContext leafReaderContext = null;
+        int endReaderIdx = -1;
         for (int index = 0; index < context.docIdsToLoadSize(); index++) {
             if (context.isCancelled()) {
                 throw new TaskCancelledException("cancelled");
             }
             int docId = docs[index].docId;
             try {
-                int readerIndex = ReaderUtil.subIndex(docId, context.searcher().getIndexReader().leaves());
-                if (currentReaderIndex != readerIndex) {
+                if (index >= endReaderIdx) {
                     profiler.startNextReader();
                     try {
-                        currentReaderContext = context.searcher().getIndexReader().leaves().get(readerIndex);
-                        currentReaderIndex = readerIndex;
-                        if (currentReaderContext.reader()instanceof SequentialStoredFieldsLeafReader lf
+                        leafIndex++;
+                        leafIndex = ReaderUtil.subIndex(docId, leaves.subList(leafIndex, leaves.size())) + leafIndex;
+                        leafReaderContext = context.searcher().getIndexReader().leaves().get(leafIndex);
+                        endReaderIdx = endReaderIdx(context, leafReaderContext, index, docs);
+                        int[] docIdsInLeaf = docIdsInLeaf(index, endReaderIdx, docs, leafReaderContext.docBase);
+                        if (leafReaderContext.reader()instanceof SequentialStoredFieldsLeafReader lf
                             && hasSequentialDocs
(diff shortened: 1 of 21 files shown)


server/src/main/java/org/elasticsearch/search/fetch/FetchPhase.java
```diff
@@ -120,25 +120,29 @@ private SearchHits buildSearchHits(SearchContext context, Profiler profiler) {
...
                     try {
-                        currentReaderContext = context.searcher().getIndexReader().leaves().get(readerIndex);
-                        currentReaderIndex = readerIndex;
-                        if (currentReaderContext.reader()instanceof SequentialStoredFieldsLeafReader lf
+                        leafIndex++;
+                        leafIndex = ReaderUtil.subIndex(docId, leaves.subList(leafIndex, leaves.size())) + leafIndex;
+                        leafReaderContext = context.searcher().getIndexReader().leaves().get(leafIndex);
+                        endReaderIdx = endReaderIdx(context, leafReaderContext, index, docs);
+                        int[] docIdsInLeaf = docIdsInLeaf(index, endReaderIdx, docs, leafReaderContext.docBase);
+                        if (leafReaderContext.reader()instanceof SequentialStoredFieldsLeafReader lf
```
romseygeek, line 145: nit: spacing before `instanceof`

server/src/main/java/org/elasticsearch/search/fetch/FetchPhase.java
```diff
@@ -120,25 +120,29 @@ private SearchHits buildSearchHits(SearchContext context, Profiler profiler) {
...
-        int currentReaderIndex = -1;
-        LeafReaderContext currentReaderContext = null;
+        List<LeafReaderContext> leaves = context.searcher().getIndexReader().leaves();
         LeafNestedDocuments leafNestedDocuments = null;
         CheckedBiConsumer<Integer, FieldsVisitor, IOException> fieldReader = null;
         boolean hasSequentialDocs = hasSequentialDocs(docs);
         SourceLoader.Leaf leafSourceLoader = null;
+        int leafIndex = -1;
+        LeafReaderContext leafReaderContext = null;
+        int endReaderIdx = -1;
```
romseygeek, line 130: Yeah this still isn't pretty but FetchPhase as a whole is ugly already so I don't think this is making it actively worse. One day I'll wrap this up into a PerSegmentFetchPhase or something.

server/src/test/java/org/elasticsearch/index/mapper/DateFieldMapperTests.java
```diff
@@ -583,12 +583,12 @@ protected SyntheticSourceSupport syntheticSourceSupport() {
                 : DateFieldMapper.DEFAULT_DATE_TIME_NANOS_FORMATTER;
 
             @Override
-            public SyntheticSourceExample example() {
+            public SyntheticSourceExample example(int maxValues) {
```
romseygeek, line 586: Is this directly related to the main change or is it part of a separate refactor?

romseygeek (approved):
LGTM! The extra messiness is pretty well contained in FetchPhase and that's something we can look at separately.

server/src/main/java/org/elasticsearch/search/fetch/FetchPhase.java
nik9000: You can't! There's a bug in our formatter that needs, like, a fix upstream. The delivery folks are managing it. But, for now, you *can't* have a space there!

server/src/main/java/org/elasticsearch/search/fetch/FetchPhase.java
nik9000: I really like the `LeafBlahBlah` model instead of the `setNextLeaf` model. Not that I'm an expert, but I think the code is generally more readable.

server/src/test/java/org/elasticsearch/index/mapper/DateFieldMapperTests.java
nik9000: Related. I can cut it to it's own change. But it's related.

The new behavior only kicks in if you fetch more than one doc. And, in general, now that there is more state on the SourceLoader it's important to test more than one doc. So I wrote that "more than one doc" version of the test. To test the new behavior and out of paranoia. And *that* needs this. It needs it so that the "more than one doc" can consistently force singleton values. If we didn't have this then it'd be *super* rare to get singletons. But with this the "more than one doc" test can do `int maxValue = randomBoolean() ? 1 : 5;` up front. And gets singletons half the time.

romseygeek: +1, sounds sensible to me, I'm happy to keep it as part of the same change

PR #87932, opened 2022-06-22
Upgrade to new Lucene snapshot

dnhatn:
This PR uses Lucene-9.3 snapshot in Elasticsearch 8.4. Noticeable changes in this Lucene snapshot:

- Merge-on-refresh
- No more pathological merging
- SortedSetDocValues#count for value_count aggs

I have to disable merges on refresh in this upgrade because some tests rely on the stable storage after refresh/flush. Should we introduce an index setting for merges-on-fresh so users can disable or adjust the wait-for-merges on refresh/flush interval? Other option is to add a test-only setting so we can disable or use random value in tests. @jpountz WDYT?


server/src/main/java/org/elasticsearch/index/engine/InternalEngine.java
```diff
@@ -2408,6 +2408,8 @@ private IndexWriterConfig getIndexWriterConfig() {
             mergePolicy = new ShuffleForcedMergePolicy(mergePolicy);
         }
         iwc.setMergePolicy(mergePolicy);
+        // TODO: Introduce an index setting for setMaxFullFlushMergeWaitMillis
+        iwc.setMaxFullFlushMergeWaitMillis(-1);
```
dnhatn, line 2412: We disable merges on refresh here.

jpountz:
> I have to disable merges on refresh in this upgrade because some tests rely on the stable storage after refresh/flush. Should we introduce an index setting for merges-on-fresh so users can disable or adjust the wait-for-merges on refresh/flush interval? Other option is to add a test-only setting so we can disable or use random value in tests. @jpountz WDYT?

@ywelsch brought up the point that merge-on-refresh might also interfere in undesirable ways with internal refreshes. Let's disable merge-on-refresh always for now, and discuss how to enable it (including a setting if necessary) in a separate PR?

dnhatn:
> Let's disable merge-on-refresh always for now, and discuss how to enable it (including a setting if necessary) in a separate PR?

Yes, it's already disabled in this PR.

jpountz (approved):

PR #87933, opened 2022-06-22
Prevent migration of system indices that match templates

grcevski:
Legacy and composable templates may apply to system indices. We don't want this behavior during feature migration. The destination index should get its settings, mappings, and aliases from its system index descriptor. A user who gets this warning should modify their templates so that they do not match system indices, at least during the migration.

This PR is an 8.x version of #87262, where we add an additional flag to all `SystemIndexDescriptors` that allows certain SystemIndexDescriptors to use index templates. At the moment this is only applicable to Kibana (see https://github.com/elastic/kibana/issues/134897).

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/admin/cluster/migration/TransportGetFeatureUpgradeStatusAction.java
+++ b/server/src/main/java/org/elasticsearch/action/admin/cluster/migration/TransportGetFeatureUpgradeStatusAction.java
@@ -38,6 +38,7 @@ import static org.elasticsearch.action.admin.cluster.migration.GetFeatureUpgrade
 import static org.elasticsearch.action.admin.cluster.migration.GetFeatureUpgradeStatusResponse.UpgradeStatus.IN_PROGRESS;
 import static org.elasticsearch.action.admin.cluster.migration.GetFeatureUpgradeStatusResponse.UpgradeStatus.MIGRATION_NEEDED;
 import static org.elasticsearch.action.admin.cluster.migration.GetFeatureUpgradeStatusResponse.UpgradeStatus.NO_MIGRATION_NEEDED;
+import static org.elasticsearch.indices.SystemIndices.UPGRADED_INDEX_SUFFIX;
 import static org.elasticsearch.upgrades.SystemIndexMigrationTaskParams.SYSTEM_INDEX_UPGRADE_TASK_NAME;
 
 /**
@@ -150,6 +151,7 @@ public class TransportGetFeatureUpgradeStatusAction extends TransportMasterNodeA
         ).map(FeatureMigrationResults::getFeatureStatuses).map(results -> results.get(feature.getName())).orElse(null);
 
         final String failedFeatureName = featureStatus == null ? null : featureStatus.getFailedIndexName();
+        final String failedFeatureUpgradedName = failedFeatureName == null ? null : failedFeatureName + UPGRADED_INDEX_SUFFIX;
         final Exception exception = featureStatus == null ? null : featureStatus.getException();
 
         return feature.getIndexDescriptors()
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/elasticsearch/action/admin/cluster/migration/TransportGetFeatureUpgradeStatusAction.java
```diff
@@ -161,7 +163,8 @@ static List<GetFeatureUpgradeStatusResponse.IndexInfo> getIndexInfos(ClusterStat
                 indexMetadata -> new GetFeatureUpgradeStatusResponse.IndexInfo(
                     indexMetadata.getIndex().getName(),
                     indexMetadata.getCreationVersion(),
-                    indexMetadata.getIndex().getName().equals(failedFeatureName) ? exception : null
+                    (indexMetadata.getIndex().getName().equals(failedFeatureName)
+                        || indexMetadata.getIndex().getName().equals(failedFeatureUpgradedName)) ? exception : null
```
grcevski, line 167: This is interesting, in 8 we weren't properly determining the error status of the upgrade because the index reported has the upgraded suffix.

grcevski:
@rjernst I added the flag as we discussed and I tagged @gwbrown to review as he was reviewing the initial 7.17 version of the PR. I'll backport the flag back to 7.17.x and the migration logic/tests once we merge this here.

server/src/main/java/org/elasticsearch/action/admin/cluster/migration/TransportGetFeatureUpgradeStatusAction.java
gwbrown: Whoops, that one's my bad! Good catch.

AthenaEryma (approved):
Left some minor notes, otherwise LGTM!

rjernst (approved):
LGTM. My only question is whether we could be a little more strict. Does Kibana use v1 or v2 templates? Could we completely block the other one (ie not allow that unused template type completely)?

grcevski:
> LGTM. My only question is whether we could be a little more strict. Does Kibana use v1 or v2 templates? Could we completely block the other one (ie not allow that unused template type completely)?

Good idea, let me see what they use. They mentioned they are using the new type, but I'd have to confirm.

grcevski:
OK, I don't think I can restrict it further. At some point they had a legacy index template, which they explicitly remove now and replace it with the new version. However a customer upgrading from an old version can potentially have the legacy one at the time of Elasticsearch upgrade because they haven't upgraded Kibana yet. From their code:

```
    const sessionIndexTemplateName = `${this.options.kibanaIndexName}_security_session_index_template_${SESSION_INDEX_TEMPLATE_VERSION}`;
    return (this.indexInitialization = new Promise<void>(async (resolve, reject) => {
      try {
        // Check if legacy index template exists, and remove it if it does.
        let legacyIndexTemplateExists = false;
        try {
          legacyIndexTemplateExists = await this.options.elasticsearchClient.indices.existsTemplate(
            {
              name: sessionIndexTemplateName,
            }
          );
        } catch (err) {
          this.options.logger.error(
            `Failed to check if session legacy index template exists: ${err.message}`
          );
          return reject(err);
        }

        if (legacyIndexTemplateExists) {
          try {
            await this.options.elasticsearchClient.indices.deleteTemplate({
              name: sessionIndexTemplateName,
            });
            this.options.logger.debug('Successfully deleted session legacy index template.');
          } catch (err) {
            this.options.logger.error(
              `Failed to delete session legacy index template: ${err.message}`
            );
            return reject(err);
          }
        }

        // Check if required index template exists.
        let indexTemplateExists = false;
        try {
          indexTemplateExists = await this.options.elasticsearchClient.indices.existsIndexTemplate({
            name: sessionIndexTemplateName,
          });
        } catch (err) {
          this.options.logger.error(
            `Failed to check if session index template exists: ${err.message}`
          );
          return reject(err);
        }
```

PR #87942, opened 2022-06-23
Improve console exception messages

rjernst:
This commit improves upon the recent console exception logging isolation
to emit more than just the exception message. The exception class is
also printed, as well as up to the first 5 elements of the stack. THe
stack trace is a hint, and can often be enough, yet the full stack trace
(or a truncated version if a guice exception) will still exist in the
logs. Addtionally, unit tests are added.


pgomulka (approved):
LGTM

PR #87963, opened 2022-06-23
Add user action for the instance_has_master indicator

andreidan:


server/src/main/java/org/elasticsearch/cluster/coordination/InstanceHasMasterHealthIndicatorService.java
```diff
@@ -29,16 +29,26 @@ public class InstanceHasMasterHealthIndicatorService implements HealthIndicatorS
...
-    private static final String NO_MASTER_INGEST_IMPACT = "The cluster cannot create, delete, or rebalance indices, and cannot insert or "
+    static final String INSTANCE_HAS_MASTER_GREEN_SUMMARY = "Health coordinating instance has an elected master node.";
+    static final String INSTANCE_HAS_MASTER_RED_SUMMARY = "Health coordinating instance does not have an elected master node.";
+
+    public static final String HELP_URL = "https://ela.st/fix-master";
+    public static final String GET_HELP_GUIDE = "https://ela.st/getting-help";
+    public static final UserAction NO_MASTER = new UserAction(
+        new UserAction.Definition(
+            "no-master-on-instance",
+            "The instance handling the health request doesn’t recognise a master node. Please try executing the API against a different "
```
jbaiera, line 40: This looks very similar to the preflight messages for an Unknown status. Should this be closer to the current stable master indicator user action of "get help from support or the docs?"

andreidan: I agree it looks similar. I'm not entirely sure we should explicitly redirect to support (although, the get help link will take our users there) as I don't particularly see this as a problem necessarily (ie. a node temporarily without a master might just be a candidate so we can't **really** say it's a problem).
This indicator shouldn't be used for much IMO and should act as a stepping stone to `master_is_stable` more than anything else. 

IMO the users should just try again on other nodes. @dakrone what do you think?

dakrone: The wording of "recognize" is a little confusing here because it could be ambiguous about "I have a master, but I can't validate the identity of that master". 

How long can we go with the messages? Can we say something like "`The instance handling the health request does not have an elected master node. Please try executing the API against a different node or try again later. If the problem persists please contact support`"?

andreidan: Sounds good to me

andreidan:
@elasticmachine update branch

jbaiera (approved):
LGTM 🚢

pugnascotia:
@andreidan does this work require a changelog YAML file? I think the automation skipped creating one because this PR is also labelled `>docs`, which is unforunate.

pugnascotia:
I added something manually to #88332.

PR #87964, opened 2022-06-23
EQL: support multiple test outcomes for non-deterministic TOML tests

luigidellaquila:
Some EQL test cases (Sample test cases in particular) are not completely deterministic; their outcome can change based on how the data are indexed on multiple shards.

This PR allows to define multiple possible outcomes (`expected_event_ids`) for a toml test case.

Both the following are now valid:

- single array of IDs: only one possible outcome is possible for this test
```
expected_event_ids  = [12, 13, 14]
```

- array of arrays of IDs: any of the ID arrays is a valid outcome for the test 
```
expected_event_ids  = [
  [12, 13, 14],
  [12, 13, 15]
]
```


x-pack/plugin/eql/qa/common/src/main/resources/test_sample.toml
```diff
@@ -3,11 +3,14 @@ name = "anyRulesWithOneDifferentNameKey"
 query = '''
```
astefan, line 3: At the top of this file, at least, please add a comment describing this syntax (array of arrays in case of multiple possible solutions).

x-pack/plugin/eql/qa/common/src/test/java/org/elasticsearch/test/eql/EqlSpecLoaderTests.java
```diff
@@ -0,0 +1,50 @@
...
+ * 2.0.
+ */
+
+package org.elasticsearch.test.eql;
+
+import org.elasticsearch.test.ESTestCase;
+
+import java.util.List;
+
+public class EqlSpecLoaderTests extends ESTestCase {
```
astefan, line 14: I think this test is not really needed. The description to be added to test_sample.toml and the tests in samples should be enough.

x-pack/plugin/eql/qa/common/src/main/java/org/elasticsearch/test/eql/EqlSpec.java
```diff
@@ -10,15 +10,17 @@
...
+import java.util.stream.Collectors;
 
 public class EqlSpec {
     private String name;
     private String description;
     private String note;
     private String[] tags;
     private String query;
-    private long[] expectedEventIds;
+    private List<long[]> expectedEventIds;
```
astefan, line 23: Please, add a comment to this line as well, mentioning a list of possible valid results to check.

astefan (commented):
This looks good.
I've left some comments. Nothing major.

x-pack/plugin/eql/qa/common/src/main/java/org/elasticsearch/test/eql/EqlSpecLoader.java
```diff
@@ -87,19 +87,27 @@ private static List<EqlSpec> readFromStream(InputStream is, Set<String> uniqueTe
                 String[] tags = new String[arr.size()];
                 int i = 0;
                 for (Object obj : arr) {
-                    tags[i] = (String) obj;
+                    tags[i++] = (String) obj;
```
bpintea, line 90: :ok_hand:

x-pack/plugin/eql/qa/common/src/main/java/org/elasticsearch/test/eql/BaseEqlSpecTestCase.java
```diff
@@ -168,20 +169,42 @@ public String toString() {
...
-                "unexpected result for spec[{}] [{}] -> {} vs {}",
-                name,
-                query,
-                Arrays.toString(expected),
-                Arrays.toString(actual)
-            ),
-            expected,
-            actual
-        );
+        if (eventIds.size() == 1) {
```
bpintea, line 173: Is this case singled out just to make use of `assertArrayEquals()`? Otherwise using the code in the other branch should be enough, imo. The message could be customized based on list size.

bpintea (approved):
LGTM

x-pack/plugin/eql/qa/common/src/main/java/org/elasticsearch/test/eql/BaseEqlSpecTestCase.java
luigidellaquila: It's just because assertArrayEquals() returns a more precise error message (ie. it also points to the exact array element that differs), that is useful in that specific case, but does not make much sense when you have multiple possible results

PR #87974, opened 2022-06-23
Issue #87610

icemakegirl:
I tried to perform the refactorings regarding this issue
server -  org.elasticsearch.common.inject.internal.Strings e  
 x-pack/plugin/eql - org.elasticsearch.xpack.eql.expression.function.scalar.string.StringUtils
issue #87610


server/src/main/java/org/elasticsearch/common/inject/internal/FirstStringUpper.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package org.elasticsearch.common.inject.internal;
+
+/**
+ * String utilities.
+ *
+ * @author crazybob@google.com (Bob Lee)
+ */
+public class FirstStringUpper {
```
thecoop, line 24: Looks like this class has been copied from `org.elasticsearch.common.inject.internal.Strings`, rather than moved?

server/src/main/java/org/elasticsearch/common/inject/internal/FirstStringUpper.java
```diff
@@ -0,0 +1,55 @@
...
+     *    capitalize("Foo bar").equals("Foo bar");
+     *    capitalize("").equals("");
+     * </pre>
+     *
+     * @param str the string whose first character is to be uppercased
+     * @return a string equivalent to {@code s} with its first character
+     *         converted to FirstStringUpper
+     * @throws NullPointerException if {@code s} is null
+     */
+    public static String capitalize(String str) {
```
thecoop, line 47: This method is in a single class - is there an existing method in another class that has similar functionality, that could be used instead, so allowing this class (and the `internal.Strings` class) to be removed entirely?

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/expression/function/scalar/string/ExtractSubstring.java
```diff
@@ -14,9 +14,9 @@
 import static org.elasticsearch.common.Strings.hasLength;
 import static org.elasticsearch.xpack.ql.util.StringUtils.EMPTY;
 
-final class StringUtils {
+final class ExtractSubstring {
```
thecoop, line 17: Similar to the code above - is there an existing method with similar functionality that could be used instead of this method?

thecoop (changes requested):
Many thanks for your interest in Elasticsearch! I have added a few comments to the PR, with suggestions on where it could be improved.

Could you also ensure your PR is updated with more recent code from `main`? There have been several other changes in this area that might help with further improvements here.

PR #87975, opened 2022-06-23
[Health API] Propagate the disk health thresholds via the cluster state

gmarouli:
*Problem statement*
In order to enable each node to be able to check their disk usage and report their health status we need certain thresholds. The thresholds may vary from node to node. We want to use the same thresholds and especially the ones defined by the master node.

*Proposal*
When a node is a elected as master, it will add a custom metadata to the cluster state that will describe these thresholds. For example:
```
    "health_metadata": {
      "disk_thresholds": {
        "low_watermark": "85%",
        "high_watermark": "90%",
        "flood_stage_watermark": "95%",
        "frozen_flood_stage_watermark": "95%",
        "frozen_flood_stage_max_headroom": "20gb",
        "yellow_threshold": "90%",
        "red_threshold": "95%"
      }
    }
```
Part of #84811


gmarouli:

gmarouli:
Fix for test failures:
[FollowIndexSecurityIT » testCleanShardFollowTaskAfterDeleteFollower](https://gradle-enterprise.elastic.co/s/zeq5gybrx7hgq/tests/:x-pack:plugin:ccr:qa:security:follow-cluster/org.elasticsearch.xpack.ccr.FollowIndexSecurityIT/testCleanShardFollowTaskAfterDeleteFollower)
[FollowIndexSecurityIT » testFollowIndex](https://gradle-enterprise.elastic.co/s/zeq5gybrx7hgq/tests/:x-pack:plugin:ccr:qa:security:follow-cluster/org.elasticsearch.xpack.ccr.FollowIndexSecurityIT/testFollowIndex) 

is coming in https://github.com/elastic/elasticsearch/pull/88074

gmarouli:
> Fix for test failures: [FollowIndexSecurityIT » testCleanShardFollowTaskAfterDeleteFollower](https://gradle-enterprise.elastic.co/s/zeq5gybrx7hgq/tests/:x-pack:plugin:ccr:qa:security:follow-cluster/org.elasticsearch.xpack.ccr.FollowIndexSecurityIT/testCleanShardFollowTaskAfterDeleteFollower) [FollowIndexSecurityIT » testFollowIndex](https://gradle-enterprise.elastic.co/s/zeq5gybrx7hgq/tests/:x-pack:plugin:ccr:qa:security:follow-cluster/org.elasticsearch.xpack.ccr.FollowIndexSecurityIT/testFollowIndex)
> 
> is coming in #88074

I had overwritten the feature flag by accident. Change reverted.

server/src/main/java/org/elasticsearch/common/util/DiskThresholdSettingParser.java
```diff
@@ -0,0 +1,185 @@
...
+    // Checks that a value is definitely not a percentage by testing if it ends on `b` which implies that it is probably a byte size value
+    // instead. This is used to make setting validation skip attempting to parse a value as a percentage/ration for the settings in this
+    // class that accept either a byte size value. The main motivation of this method is to make tests faster. Some tests call this method
+    // frequently when starting up internal cluster nodes and using exception throwing and catching when trying to parse as a ratio as a
+    // means of identifying that a string is not a ratio is quite slow.
+    public static boolean definitelyNotPercentage(String value) {
+        return value.endsWith("b");
+    }
+
+    public static void doValidate(String newValue, List<ThresholdSetting> thresholdsInOrder) {
```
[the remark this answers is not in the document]
andreidan: I'm also a bit concerned that we use a `List<XXX>` that we expect to be in order, but we don't enforce such a thing. Should we add an assertion? 
It's also not exactly clear where the `newValue` belongs to? Is it validated against the settings in the list? Is it already in the list of settings? (I'd say it shouldn't be as we're just validating if it _can_ be )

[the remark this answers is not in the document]
andreidan: I'm guessing this is a replacement of the `doValidate(low, high, flood)` and I'm wondering if it's better to keep it that way? (maybe have a `doValidate(low, high)` too)

server/src/main/java/org/elasticsearch/health/metadata/HealthDiskThresholdSettings.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+
+/**
+ * A container to keep the settings for health disk thresholds up to date with cluster setting changes.
+ * These thresholds are used to determine if a node that does not hold data has enough disk space.
+ */
+public class HealthDiskThresholdSettings {
+
+    public static final Setting<String> CLUSTER_HEALTH_DISK_YELLOW_THRESHOLD_SETTING = new Setting<>(
+        "cluster.health.disk.threshold.yellow",
```
andreidan, line 30: nit: Should we keep the low/high terminology our users are used to already? ie. "cluster.health.reporting.disk.threshold.low"

andreidan: I think we should also document these settings

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadata.java
```diff
@@ -0,0 +1,299 @@
...
+import java.io.IOException;
+import java.util.EnumSet;
+import java.util.Objects;
+
+/**
+ * A cluster state entry that contains a list of all the thresholds used to determine if a node is healthy.
+ */
+public final class HealthMetadata extends AbstractNamedDiffable<Metadata.Custom> implements Metadata.Custom {
+
+    public static final String TYPE = "health_metadata";
```
andreidan, line 39: nit: Would "health" be enough?

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
```diff
@@ -0,0 +1,181 @@
...
+            @Override
+            public ClusterState execute(ClusterState currentState, List<TaskContext<UpdateHealthMetadataTask>> taskContexts)
+                throws Exception {
+                final HealthMetadata originalHealthMetadata = HealthMetadata.getHealthCustomMetadata(currentState);
+                HealthMetadata mostRecentHealthMetadata = originalHealthMetadata;
+                if (taskContexts.isEmpty() == false) {
+                    mostRecentHealthMetadata = taskContexts.get(taskContexts.size() - 1).getTask().healthMetadata();
+                    if (Objects.equals(originalHealthMetadata, mostRecentHealthMetadata)) {
+                        mostRecentHealthMetadata = originalHealthMetadata;
+                    }
```
andreidan, line 168: This is a bit confusing and if I understand correctly it's a micro optimisation? Do we aim to skip some task executions? 

Would it be simpler to iterate over all tasks and execute them like we do everywhere else? ie. https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/ingest/IngestService.java#L115

andreidan (commented):
Thanks for working on this Mary. Left a few comments

server/src/main/java/org/elasticsearch/health/metadata/HealthDiskThresholdSettings.java
```diff
@@ -0,0 +1,146 @@
...
+        clusterSettings.addSettingsUpdateConsumer(CLUSTER_HEALTH_DISK_YELLOW_THRESHOLD_SETTING, this::setYellowThreshold);
+        clusterSettings.addSettingsUpdateConsumer(CLUSTER_HEALTH_DISK_RED_THRESHOLD_SETTING, this::setRedThreshold);
+    }
+
+    private void setYellowThreshold(String yellowThreshold) {
+        this.yellowThreshold = RelativeByteSizeValue.parseRelativeByteSizeValue(
+            yellowThreshold,
+            CLUSTER_HEALTH_DISK_YELLOW_THRESHOLD_SETTING.getKey()
+        );
+        listeners.forEach(Listener::onChange);
```
dakrone, line 62: Do we want listener exceptions to cause further listeners to be skipped? We probably want to wrap the invocation in a `try/catch`

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
```diff
@@ -0,0 +1,181 @@
...
+                }
+                for (final var taskContext : taskContexts) {
+                    taskContext.success(() -> {});
+                }
```
dakrone, line 172: If you move this above the equality check you can make the check just return `currentState`, so something like:

```java
taskContexts.forEach(tc -> tc.success(() -> {}));
if (Objects.equals(originalHealthMetadata, mostRecentHealthMetadata) {
    return currentState;
}
return ClusterState.build(currentState)
    .metadata(Metadata.builder(currentState.metadata()).putCustom(HealthMetadata.TYPE, mostRecentHealthMetadata))
    .build();
```

dakrone (commented):
It's a little difficult for me to follow this PR, with `HealthMetadata`, that is node-specific, but I see that some of our settings are node-specific and some are not. For example, the `HealthDiskThresholdSettings` are dynamic and those going to be cluster-wide (at least if they are changed), but a lot of our disk threshold settings are node-specific and storing only a single copy looks like we store a single copy in the cluster state metadata (if I read the PR correctly).

I was thinking we would store something more like `Map<String, HealthMetadata>` where we could store the node-specific thresholds in the state?

server/src/main/java/org/elasticsearch/common/settings/ClusterSettings.java
```diff
@@ -183,345 +186,355 @@ public void apply(Settings value, Settings current, Settings previous) {
...
+            ReadinessService.PORT
+        );
+        Set<Setting<?>> behindFeatureFlag = HealthNode.isEnabled()
+            ? Set.of(
+                HealthNodeTaskExecutor.ENABLED_SETTING,
+                HealthDiskThresholdSettings.CLUSTER_HEALTH_DISK_YELLOW_THRESHOLD_SETTING,
+                HealthDiskThresholdSettings.CLUSTER_HEALTH_DISK_RED_THRESHOLD_SETTING
+            )
+            : Set.of();
+        BUILT_IN_CLUSTER_SETTINGS = Sets.union(buildInClusterSettings, behindFeatureFlag);
```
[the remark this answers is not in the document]
gmarouli: Thanks for pointing it out, I completely missed that. This is much better

server/src/main/java/org/elasticsearch/common/util/DiskThresholdSettingParser.java
```diff
@@ -0,0 +1,185 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.common.util;
```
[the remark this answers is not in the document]
gmarouli: I completely agree, I thought that package had actual settings, but I see now that it has more than that! I will move it

server/src/main/java/org/elasticsearch/common/util/DiskThresholdSettingParser.java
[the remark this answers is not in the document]
gmarouli: Okay, I will preserve the current structure, I understand the concerns.

server/src/main/java/org/elasticsearch/health/metadata/HealthDiskThresholdSettings.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.Map;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+/**
+ * A container to keep the settings for health disk thresholds up to date with cluster setting changes.
+ * These thresholds are used to determine if a node that does not hold data has enough disk space.
+ */
+public class HealthDiskThresholdSettings {
+
+    public static final Setting<String> CLUSTER_HEALTH_DISK_YELLOW_THRESHOLD_SETTING = new Setting<>(
```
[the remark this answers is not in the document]
gmarouli: I was thinking about this too. From class name perspective it makes perfect sense, but based on the package structure it implies that these disk thresholds are relevant to allocation, and the new thresholds we are introducing are not really relevant to that, right? 

Maybe I could move them to settings and group them all together. What do you think?

server/src/main/java/org/elasticsearch/health/metadata/HealthDiskThresholdSettings.java
gmarouli: +1. I just didn't want to document until they are not behind a feature flag. I was planning to add them in a separate PR.

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadata.java
gmarouli: hm.... yeah :)

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
```diff
@@ -0,0 +1,181 @@
...
+
+        @Override
+        public void clusterStateProcessed(ClusterState oldState, ClusterState newState) {
+            assert false : "never called";
+        }
+
+        @Override
+        public void onFailure(@Nullable Exception e) {
+            logger.error("unexpected failure during health metadata update", e);
+            retry.run();
```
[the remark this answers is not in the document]
gmarouli: You are right, I checked on other cases and we simply log a failure. I was wondering if there could be cases where this fails and the metadata get out of sync for a long time since they do not get updated often. But if this scenario is rare, I would prefer too to remove the retry mechanism because it makes the code much more complex.

server/src/main/java/org/elasticsearch/common/util/DiskThresholdSettingParser.java
```diff
@@ -0,0 +1,185 @@
...
+import org.elasticsearch.ElasticsearchParseException;
+import org.elasticsearch.common.unit.ByteSizeValue;
+import org.elasticsearch.common.unit.RatioValue;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.stream.Collectors;
+
+/**
+ * A parser that parses and validates disk thresholds. The disk thresholds accept two different type of inputs either bytes or percentages.
```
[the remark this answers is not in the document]
gmarouli: Fixed

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
gmarouli: Kind of yes, since we just pick the latest version and we update the cluster state I thought it makes sense to just pick the last one. On the other hand since we are already iterating the task contexts to signal the success then it is not the optimization I thought it was. I will change it.

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
gmarouli: I remodeled it after https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/ingest/IngestService.java#L115.

server/src/main/java/org/elasticsearch/health/metadata/HealthDiskThresholdSettings.java
gmarouli: Yes, I wouldn't want one listener to fail them all.

gmarouli:
This PR is split into two parts to help with the review:
- https://github.com/elastic/elasticsearch/pull/88175
- coming soon

server/src/main/java/org/elasticsearch/common/util/DiskThresholdSettingParser.java
```diff
@@ -0,0 +1,185 @@
...
+            }
+            throw ex;
+        }
+    }
+
+    /**
+     * Checks if a threshold string is a valid percentage or byte size value,
+     * @return the threshold value given
+     */
+    public static String validThresholdSetting(String threshold, String settingName) {
```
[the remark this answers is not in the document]
gmarouli: Since extracting the validation will not happen for now, this is not applicable. I will double check this when we introduce the new thresholds.

Issue #87931, opened 2022-06-22
Make GetTrainedModelsStatsAction cancellable


DaveCTurner:
I encountered a Cloud cluster with an overworked master due (partly) to processing multiple calls to `GET /_ml/anomaly_detectors/_all/_stats` originating from an external Metricbeat monitoring process. Metricbeat imposes a 10s timeout after which it closes the HTTP connection and tries again. However, `GetTrainedModelsStatsAction` does not notice if the client connection closes (i.e. the REST handler does not use `RestCancellableNodeClient` and the resulting transport task is not a `CancellableTask`) so it carries on wastefully processing the request even after the client timeout.

Relates https://github.com/elastic/elasticsearch/issues/55550

droberts195:
Based on the Slack thread that went with this I think it's actually `GetJobsStatsAction` that was the problem, not `GetTrainedModelsStatsAction`.  But we should probably add cancellation capability to _all_ our stats actions that do more than one step internally to get the stats.

droberts195:
OK, I saw `/_ml/anomaly_detectors/_all/_stats` and that corresponds to `GetJobsStatsAction`.

But it looks like it really is `GetTrainedModelsStatsAction`.

@benwtrent as well as making the action cancellable it looks like we need a more efficient way to list which ingest pipelines a particular trained model is used in. It looks like it's instantiating _all_ the processors at the moment, then looking for inference processors, taking the ID and throwing everything else away.  But instantiating a Grok processor seems to be very expensive.  So maybe we need a specialist method at a lower level that can search for inference processors in an ingest pipeline without instantiating every other processor.

benwtrent:
Arg, when I first wrote this code, I was extracting from opaque maps. Then I figured "Ya know, we could build the ingest pipelines and just extract from there and not iterate maps...". Welp, Time to go back and find that old code... 🤦

PR #87978, opened 2022-06-23
[ML] improve trained model stats API performance

benwtrent:
Previous, get trained model stats API would build every pipeline defined in cluster state. 

This is problematic when MANY pipelines are defined. Especially if those pipelines take some time to parse (consider `GROK`). 

This improvement is part of fixing: https://github.com/elastic/elasticsearch/issues/87931


benwtrent:
Reviewer, issue https://github.com/elastic/elasticsearch/issues/87931 is focused on making the API cancellable. 

I will do that in a separate PR to reduce the review space and code churn :).

benwtrent:
@elasticmachine update branch

droberts195:
I know we have a report saying that this got slow after upgrading from 7.17 to 8.x, but I cannot see why this couldn't have been just as slow in 7.17 assuming the same ingest pipelines existed.

Is there a good explanation for why the 7.17 get trained model stats code would _not_ have suffered from the problem of slow Grok processor instantiation?  If there isn't then I think this fix should be backported to 7.17.6 (and 8.3.1 too, to get it out to 8.x users sooner).

benwtrent:
@droberts195 the only thing I can think of is that there may be more solutions/apps in Kibana and Elasticsearch making more "system managed" ingest pipelines in 8 vs 7. So, the increase in the number of pipelines pushed this over the edge.

Doing a quick search of any "grok" specific changes, indicates nothing that would make stuff incredibly slower.

It is true, any user with a large number of pipelines (especially complicated GROK ones), may run into this problem. I am good with backporting it to 7.17

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/ingest/InferenceProcessor.java
```diff
@@ -213,86 +206,11 @@ public Factory(Client client, ClusterService clusterService, Settings settings)
...
-                        }
-                    }
-                    // We cannot throw any exception here. It might break other pipelines.
-                } catch (Exception ex) {
-                    logger.debug(() -> "failed gathering processors for pipeline [" + configuration.getId() + "]", ex);
-                }
+            try {
+                currentInferenceProcessors = InferenceProcessorInfoExtractor.countInferenceProcessors(state);
+            } catch (Exception ex) {
+                logger.debug("failed gathering processors for pipelines", ex);
```
droberts195, line 212: If might be best to keep the old comment: `// We cannot throw any exception here. It might break other pipelines.`

x-pack/plugin/ml/src/test/java/org/elasticsearch/xpack/ml/utils/InferenceProcessorInfoExtractorTests.java
```diff
@@ -0,0 +1,212 @@
...
+        for (String id : modelId) {
+            for (int i = 0; i < numPipelineReferences; i++) {
+                configurations.put(
+                    "pipeline_with_model_" + id + i,
+                    randomBoolean() ? newConfigurationWithInferenceProcessor(id) : newConfigurationWithForeachProcessorProcessor(id)
+                );
+            }
+
+        }
+        for (int i = 0; i < randomInt(5); i++) {
```
droberts195, line 153: It's confusing to regenerate the upper bound on each loop iteration, and means there's very little chance of ever getting to 4.  It would be better to generate the upper bound once, before the loop.

droberts195 (approved):
LGTM apart from a couple of minor nits

x-pack/plugin/ml/src/test/java/org/elasticsearch/xpack/ml/utils/InferenceProcessorInfoExtractorTests.java
benwtrent: Ah, definitely.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/ingest/InferenceProcessor.java
benwtrent, line 212:
```suggestion
            } catch (Exception ex) {
                // We cannot throw any exception here. It might break other pipelines.
                logger.debug("failed gathering processors for pipelines", ex);
```

benwtrent:
@elasticmachine update branch

benwtrent:
@elasticmachine update branch

benwtrent:
@elasticmachine update branch

PR #87980, opened 2022-06-23
Speedup setting default values a little

original-brownbear:
Many shards benchmarks have recently seen a bit of an increase in the time spent on
settings from new settings getting added. This makes some defaults quite a bit
cheaper by saving re-serialization of the same default value when its not a function
of the settings instance which helps with some cluster state application cases for large index counts.

relates #77466


grcevski (approved):
LGTM! This looks good to me.

original-brownbear:
Thanks Nikola!

PR #87981, opened 2022-06-23
[TSDB] Simplify the rollup action

csoulios:
This PR changes the `_rollup` action so that it will only create the rollup index. 
Tasks such index swap in the data stream and source index deletion will not be handled by the rollup action. 
Instead, those actions will be taken care of by the rollup ILM.

Those changes make rollup behavior more consistent with similar operations such shrink or snapshot.


nik9000 (commented):
Looks right to me.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -182,10 +175,8 @@ protected void masterOperation(
         // 3. Run rollup indexer
         // 4. Make rollup index read-only and set replicas
         // 5. Refresh rollup index
-        // 6. Add rollup index to data stream and publish rollup metadata
-        // 7. Mark rollup index as "completed successfully"
-        // 8. Delete the source index
-        // 9. Force-merge the rollup index to a single segment
+        // 6. Mark rollup index as "completed successfully"
+        // 7. Force-merge the rollup index to a single segment
```
dakrone, line 179: Out of curiosity, in the future (not in this PR), should we have this action performed by ILM as well? We currently do it also for the `searchable_snapshot` action.

dakrone (approved):
LGTM also, I assume we're going to do the index setting copying change in a different PR also?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: My intention was that the produced rollup index would be fully merged at the end of the process. If we implement it here, this can happen even when rollup indices are manually created through invoking the API endpoint. 

Are there any specific advantages to ILM merging the index?

Issue #56494, opened 2020-05-10
SQL: Consolidate Date/Time related Utils


matriv:
Currently we have various date/time utility constants and methods in ql eql sql and sql/jdbc projects.
It would be great to consolidate those utils into one class if possible.
Probably we would need a new small module qa-util/common that is shared across all modules
and can be included in the JDBC jar.

The same could be done for the test util classes.

wchaparro:
Superceded by ES|QL

PR #87985, opened 2022-06-23
Consolidating x-pack DateUtils classes

LeeTgk:
As stated at the issue https://github.com/elastic/elasticsearch/issues/56494
It would be great to consolidate those Date/Time utils into one class if possible.

This PR tries to adress the whole Issue by creating a new class ``org.elasticsearch.xpack.qautils.DateUtils``, that is used to consolidate most uses of the other classes across the xpack module.

This PR includes:
Changing
 the use cases of ``ql.util.DateUtils`` and ``ql.type.DateUtils`` to use the newly created class instead.

Changing most of the uses of ``sql.util.DateUtils``, because just moving all of them would create a situation of interdependency, between ``qa-util, ql, and sql``.

Still thinking on how to do it properly and totally accepting feedback on how to do it.
On top of that, i'm gonna move the Test classes to the new module in the future as suggested in the Issue.
Also not sure if i declared the ``build.gradle`` of the new modules correctly for the occasion. So please tell me how to improve if needed.


LeeTgk:
While seeing if it's possible to move all test methods that use the new ``qautils.DateUtils`` created by me, to a test class into the ``qautils`` module, it came to me that most of those uses are rather specific and just uses the DateUtils class to test methods of another classes, as can be seen at the main repo (linking the line of my own repo didn't work sorry): https://github.com/elastic/elasticsearch/blob/97df136b81f685bdd7e0e79bafc97a91b82a3ca2/x-pack/plugin/sql/src/test/java/org/elasticsearch/xpack/sql/type/SqlDataTypeConverterTests.java#L132

https://github.com/elastic/elasticsearch/blob/97df136b81f685bdd7e0e79bafc97a91b82a3ca2/x-pack/plugin/ql/src/test/java/org/elasticsearch/xpack/ql/type/DataTypeConversionTests.java#L60

That said i'm not sure if it would be interesting to move them creating unecessary dependencies, but please correct me if i'm wrong.
Still gonna do one more pass on the project to see if any other DateUtils functions can go into the consolidated one without creating interdependencies, otherwise i would consider it done , and will put it as a normal PR, until futher feedback is given.

luigidellaquila (changes requested):
Hi @LeeTgk 

Thank you very much for your contribution.
I see you started it based on what is described in https://github.com/elastic/elasticsearch/issues/56494, that is a long standing issue. 
I think that issue does not fully describe the general problem though: SQL and EQL are only part of the story, there is a lot of date manipulation logic also in the Server module (that SQL and EQL use a lot) and in other places, that should be rationalized and made available in a single, consistent way. 
This would bring real advantage to the development of all the query languages, in terms of function availability and maintainability, but also in terms of general consistency. 
Just to pick one single example (but there could be many more), it would be nice to have date math logic exposed from this single utility package and used by all the query languages consistently.
I'm afraid this is not a mere refactoring exercise though, it could potentially involve some architectural choices (eg. what's in x-pack and what is not)

Thanks

Luigi

PR #87989, opened 2022-06-23
Move the ingest attachment processor to the default distribution

rjernst:
The ingest attachment processor is currently available as a plugin. This
commit moves the processor to the default distribution so it is always
available.


docs/plugins/redirects.asciidoc
```diff
@@ -145,3 +145,17 @@ include::redirects.asciidoc[tag=gcs-repo-migration]
 === Google Cloud Storage repository plugin
 
 include::redirects.asciidoc[tag=gcs-repo-migration]
+
+[role="exclude",id="ingest-attachment"]
+=== Ingest Attachment plugin
+
+// tag::ingest-attachment-migration[]
+The Ingest Attachment plugin is now included in {es}.
+See the {ref}/ingest-attachment.html[Ingest Attachment] processor.
```
lcawl, line 154: > See the {ref}/ingest-attachment.html[Ingest Attachment] processor.

In my local tests, this URL fails. I think it's because it should be {ref}/attachment.html per the changes in this PR.

dakrone (approved):
LGTM assuming the docs issues are cleared up, thanks Ryan

rjernst:
@lcawl I think I addressed the issues you found, can you take another look?

lcawl:
@elasticmachine run elasticsearch-ci/docs rebuild

lcawl (approved):
I added some minor doc edits, tested locally, and both the plugins and elasticsearch guides LGTM

rjernst:
The reason for the FIPS incompatibility was a transitive dependency on bouncycastle. However, that dependency is now remove (see https://github.com/elastic/elasticsearch/pull/88031).

PR #87991, opened 2022-06-23
Cache notice generation tasks

rjernst:
This commit makes the NoticeTask cacheable. Although it is not that
expensive (a few seconds), it's worth it overall.


mark-vieira (approved):
LGTM. Thanks, Ryan.

PR #69096, opened 2021-02-17
Always add file and native realms unless explicitly disabled

ywangd:
This PR makes sure that the file and native realms are always added to the realm chain unless explicitly disabled. 

Currently, they are only impliciltly added when:
1. No other realms are configured
2. No configured realms can be used with current license (so an expired license can fallback to these basic realms)

A side effect (intended?) is that file and native realm cannot be truely disabled at all time because the above two rules always apply regardless whether the realms are disabled or not.

This PR makes the behaviour more explicit. _If the file or native realm is explicitly disabled, it will be disabled at all time._ If they are not explicitly disabled, they will always be added to the realm chain. Two scenarios are possible:
1. File or native realm is explicitly configured. In this case, their order value must be provided and honoured
2. File or native realm is not configured. In this case, they are implicitly added to the beginning of the realm chain (file then native).

PS: I labelled this as `v8.0.0` only because it is a breaking change.

TODO:
- [x] doc update

Resolves: #50892

PR #87999, opened 2022-06-24
Fix unique realm name check to cover default realms

ywangd:
All enabled realms must have unique names. This PR tightened the logic
for ensuring realm name uniqueness. Previously the unique name check
does not cover realms that are automatically enabled.

Relates: #69096


x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/RealmsTests.java
```diff
@@ -1097,6 +1089,147 @@ public void testWarningsForReservedPrefixedRealmNames() throws Exception {
...
+            )
+        );
+        assertThat(
+            e1.getMessage(),
+            containsString(
+                "Found realm configured with name clashing with the [" + defaultRealm.v2() + "] realm: [" + realmSettingKey + "]"
+            )
+        );
+
+        // It is OK if the explicitly configured realm is disabled
```
ywangd, line 1126: This and a few other test scenarios below can be seen as a leniency. Ideally these names should not be used at all. We may want to do something about it, e.g. log warnings. But it is a separate issue from what we are trying to fix here.

ywangd:
@elasticmachine update branch

tvernum (approved):
LGTM

n1v0lg (approved):
LGTM! Thanks for the fix.

ywangd:
@elasticmachine run elasticsearch-ci/part-2-fips

Issue #81788, opened 2021-12-15
Support ANN with filtering


jtibshirani:
It's useful to be able to combine kNN with a filter ("find the nearest documents that also satisfy this search query"). We could extend the recently added `_knn_search` endpoint to accept search filters too:
```
POST index/_knn_search
{
 "knn": {
  "field": "image_vector",
  "query_vector": [0.3f, 0.1f, ...],
  "k": 10,
  "num_candidates": 100
 },
 "filter": {
  "term": {
   "file_type": "jpeg"
  }
 }
}
```
The filter section accepts a search filter, defined through the Query DSL. The response will contain the nearest `k` documents that also match the filter. We could extend the `filter` section to accept a list of filters if this proves helpful.

jtibshirani:
We've begun to work on adding support in Lucene: https://issues.apache.org/jira/browse/LUCENE-10382

coreation:
Nice, looking forward to be using this feature in ElasticSearch!

yanchaoguo:
> We've begun to work on adding support in Lucene: https://issues.apache.org/jira/browse/LUCENE-10382
Which version is expected to be available
Keep an eye on

Issue #73971, opened 2021-06-09
Async search support for cross cluster search minimize round trip optimization 


ppf2:
[Async search](https://www.elastic.co/guide/en/elasticsearch/reference/current/async-search.html) does not currently support minimize roundtrip optimization when used with cross cluster search.  This is a documented limitation.  

With Kibana now using async searches by default, this could cause performance issues in some environments that use cross cluster search with slower/inconsistent network latency.  One workaround is to enable the deprecated advanced setting `courier:batchSearches` in Kibana which will switch Kibana back to using the old way of querying Elasticsearch via `msearch` (but this option is slated for removal in 8.0).

javanna:
@quux00 can this be closed?

quux00:
I was planning on waiting to hear back from the Kibana team on whether they will make it their default or whether we need to do that on the backend before closing this. But I'll go ahead and close it and we can reopen if need be.

quux00:
Kibana has changed to use ccs_minimize_roundtrips=true now. PR: https://github.com/elastic/kibana/pull/159848

PR #88002, opened 2022-06-24
Add 'knn' section to search endpoint

jtibshirani:
This PR adds a new `knn` section to the search request:
```
POST products/_search
{
  "query": {
    "multi_match": {
      "query": "black flower dress",
      "fields": ["title", "description"],
    },
    "boost": 0.9
  },
  "knn": {
    "field": "title_vector",
    "query_vector": [0.3f, 0.1f, ...],
    "k": 5,
    "num_candidates": 50,
    "boost": 0.1
  },
  "size": 20
}
```

As explained in #87625, the search returns a global top k documents across all
shards:
* In the DFS phase, collect the top k vector matches per shard, combine them
and keep the global top k. 
* Then move to the query phase. Convert the top k vector matches into a new
`KnnScoreDocQueryBuilder` that matches only those top documents. For the final
query, take a boolean disjunction between this `KnnScoreDocQueryBuilder` query and
the search request `query`.

Commits:
* Rename KnnSearchRequestBuilder -> KnnSearchRequestParser to avoid confusion with new KnnSearchBuilder
* Add `KnnScoreDocQueryBuilder`
* Implement kNN search using DFS phase

Addresses #87625.


jtibshirani:
NOTE: this PR targets a feature branch `knn-search`. There are still several TODOs it doesn't address:
* Support for filters
* Optimize case where there is only one shard, or when there is only kNN with no other query or aggregations
* Finalize aggregations behavior

server/src/test/java/org/elasticsearch/action/search/TransportSearchActionTests.java
```diff
@@ -1059,6 +1060,21 @@ public void testShouldMinimizeRoundtrips() throws Exception {
...
+        {
+            SearchRequest searchRequest = new SearchRequest();
+            SearchSourceBuilder source = new SearchSourceBuilder();
+            source.knnSearch(new KnnSearchBuilder("field", new float[] { 1, 2, 3 }, 10, 50));
+            searchRequest.source(source);
+
+            // Emulate TransportSearchAction logic: first adjust search type, then check minimize roundtrips
+            TransportSearchAction.adjustSearchType(searchRequest);
+
+            // Minimize roundtrips should always be false because kNN uses DFS_QUERY_THEN_FETCH
```
javanna, line 1072: too bad that we can't minimize roundtrips! maybe that's a good reminder to rethink whether we can do something about this, I will try to think about this.

server/src/main/java/org/elasticsearch/action/search/TransportSearchAction.java
```diff
@@ -478,6 +479,19 @@ private void executeRequest(
         Rewriteable.rewriteAndFetch(original, searchService.getRewriteContext(timeProvider::absoluteStartMillis), rewriteListener);
     }
 
+    static void adjustSearchType(SearchRequest searchRequest) {
+        // if there's a kNN search, always use DFS_QUERY_THEN_FETCH
+        if (searchRequest.hasKnnSearch()) {
+            searchRequest.searchType(DFS_QUERY_THEN_FETCH);
```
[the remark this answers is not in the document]
jtibshirani: Thanks @javanna, I meant to add a comment about this. Yes in a follow-up PR I plan to disallow setting the search type explicitly when `knn` is provided too.

server/src/test/java/org/elasticsearch/action/search/TransportSearchActionTests.java
jtibshirani: In a follow-up, I hope to optimize the case where there is only kNN, with no query or aggregations. Then we can run kNN during the main query phase (using QUERY_THEN_FETCH), and it will be possible to minimize roundtrips.

sethmlarson (commented):
API looks good, one thought:

server/src/main/java/org/elasticsearch/search/vectors/KnnSearchBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+        List<Float> vector = (List<Float>) args[1];
+        float[] vectorArray = new float[vector.size()];
+        for (int i = 0; i < vector.size(); i++) {
+            vectorArray[i] = vector.get(i);
+        }
+        return new KnnSearchBuilder((String) args[0], vectorArray, (int) args[2], (int) args[3]);
+    });
+
+    static {
+        PARSER.declareString(constructorArg(), FIELD_FIELD);
```
mayya-sharipova, line 50: Did we also want to have a `filter` section under `knn` section? Or the plan is to introduce it in the follow up?

server/src/main/java/org/elasticsearch/search/vectors/KnnSearchBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+            );
+        }
+        if (numCands > NUM_CANDS_LIMIT) {
+            throw new IllegalArgumentException("[" + NUM_CANDS_FIELD.getPreferredName() + "] cannot exceed [" + NUM_CANDS_LIMIT + "]");
+        }
+        return new KnnVectorQueryBuilder(field, queryVector, numCands).boost(boost);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
mayya-sharipova, line 122: Should `equal` and `hashCode` also include `boost` field?

server/src/main/java/org/elasticsearch/search/vectors/KnnSearchBuilder.java
jtibshirani: This will be included in a follow-up. You can see the TODO list at the bottom here: https://github.com/elastic/elasticsearch/issues/87625.

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.vectors/40_knn_search.yml
```diff
@@ -106,9 +158,12 @@ setup:
 
 ---
 "Test nonexistent field":
+  - skip:
+      version: ' - 8.3.99'
+      reason: 'kNN added to search endpoint in 8.4'
   - do:
       catch: bad_request
-      knn_search:
+      search:
```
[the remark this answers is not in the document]
jtibshirani: 👍 good catch

server/src/main/java/org/elasticsearch/search/vectors/KnnSearchBuilder.java
jtibshirani: Oops, will fix this.

server/src/main/java/org/elasticsearch/action/search/TransportSearchAction.java
```diff
@@ -478,6 +479,19 @@ private void executeRequest(
         Rewriteable.rewriteAndFetch(original, searchService.getRewriteContext(timeProvider::absoluteStartMillis), rewriteListener);
     }
 
+    static void adjustSearchType(SearchRequest searchRequest) {
+        // if there's a kNN search, always use DFS_QUERY_THEN_FETCH
```
mayya-sharipova, line 483: This means that if `knn` section present, dfs stage will always be applied to `query` part  even if we don't need it. Do you think we should invoke dfs stage for query only when explicitly requested?

jtibshirani: Right, in the docs I planned to present this as "if `knn` is provided, then the `dfs_query_then_fetch` type is always used". I thought this was easiest to understand and kept the implementation simple:
* Knowing that `dfs_query_then_fetch` is used helps explain why things like `ccs_minimize_roundtrips` will not work. kNN has all the same limitations as if `dfs_query_then_fetch` was set.
* It feels weird to run the DFS search phase but not compute DFS?

In my understanding, the main overhead of DFS is from the extra roundtrip, not gathering the stats, so it won't be too expensive. Interested to know your opinion!

server/src/test/java/org/elasticsearch/action/search/TransportSearchActionTests.java
javanna: I see, agreed that it's a bit hacky because it requires adjustSearchType to be called. I would rather add a very simple unit test around adjustSearchType then. this specific test method already tests the scenario with dfs enabled, so we should just verify that the search type is adjusted when knn is used.

javanna: Also, Nhat is running benchmarks to asses how bad it is that we don't minimize roundtrips in some cases, in the context of https://github.com/elastic/elasticsearch/issues/73971 . I would imagine that the results of the benchmark will also inform whether we need to put more effort into minimizing roundtrips when knn is used.

jtibshirani: 👍 I'll rework this test. Looking forward to seeing Nhat's results.

build-tools-internal/src/integTest/groovy/org/elasticsearch/gradle/fixtures/AbstractGradleInternalPluginFuncTest.groovy
```diff
@@ -8,11 +8,11 @@
 
 package org.elasticsearch.gradle.fixtures
 
-import org.elasticsearch.gradle.internal.conventions.precommit.PrecommitPlugin
+import org.gradle.api.Plugin
```
mayya-sharipova, line 11: Are these intentional changes?

jtibshirani: Sorry about that, I accidentally merged in `master` to get a test fix, without updating the feature branch first. I'll fix this now.

server/src/main/java/org/elasticsearch/action/search/TransportSearchAction.java
mayya-sharipova: > "if knn is provided, then the dfs_query_then_fetch type is always used"

This sounds good to me. 

> It feels weird to run the DFS search phase but not compute DFS?

What do you mean by "not compute DFS" ? It looks to me if `query` and `knn` search are present, we always compute and use DFS stats information? right?

server/src/main/java/org/elasticsearch/action/search/DfsQueryPhase.java
```diff
@@ -115,4 +127,18 @@ public void onFailure(Exception exception) {
             );
         }
     }
+
+    private void addKnnResultsToQuery(SearchSourceBuilder source) {
+        if (source == null || source.knnSearch() == null) {
+            return;
+        }
+
```
mayya-sharipova, line 135: May be we can create an appropriate QueryBuilder once for all shards and then pass to each ShardSearchRequest?

server/src/main/java/org/elasticsearch/search/vectors/ScoreDocQueryBuilder.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * A query that matches the provided docs with their scores. This query is used
+ * when executing a kNN search during the search query phase, to include the documents
+ * that matched the initial kNN query during the DFS phase.
+ */
+public class ScoreDocQueryBuilder extends AbstractQueryBuilder<ScoreDocQueryBuilder> {
+    public static final String NAME = "score_doc";
```
mayya-sharipova, line 35: Do you think we should give more descriptive name (e.g. by adding knn) to this query in case it is used for debugging (e.g. slow log).?

server/src/main/java/org/elasticsearch/search/vectors/KnnSearchBuilder.java
sethmlarson: @jtibshirani How is the `filter` section for kNN different from the `query` property of search? If I understood the previous _knn_search implementation correctly they were similar?

server/src/main/java/org/elasticsearch/search/vectors/ScoreDocQueryBuilder.java
jtibshirani: 👍 maybe we could do `knn_score_doc`. I think it's unlikely this query will be useful in a different context.

server/src/main/java/org/elasticsearch/action/search/TransportSearchAction.java
mayya-sharipova: Thanks for clarifications, Julie!

server/src/test/java/org/elasticsearch/action/search/TransportSearchActionTests.java
```diff
@@ -1059,6 +1064,48 @@ public void testShouldMinimizeRoundtrips() throws Exception {
...
+            SearchSourceBuilder source = new SearchSourceBuilder();
+            source.suggest(new SuggestBuilder().addSuggestion("field", new TermSuggestionBuilder("value")));
+            searchRequest.source(source);
+
+            TransportSearchAction.adjustSearchType(searchRequest, randomBoolean());
+            assertFalse(searchRequest.requestCache());
+            assertEquals(SearchType.QUERY_THEN_FETCH, searchRequest.searchType());
+        }
+        {
+            // Single-shard searches should always use QUERY_THEN_FETCH
```
mayya-sharipova, line 1103: may be also add here: "in the absence of knn searches"

server/src/main/java/org/elasticsearch/action/search/DfsQueryPhase.java
```diff
@@ -115,4 +127,18 @@ public void onFailure(Exception exception) {
...
+        ScoreDocQueryBuilder knnQueryBuilder = new ScoreDocQueryBuilder(knnResults.scoreDocs());
+        if (source.query() == null) {
```
mayya-sharipova, line 137: In the follow up we can clarify in the documentation that if `query` is not set, it means we are running only knn search.  Because a usual search without a query set implies `match_all` query.

mayya-sharipova (approved):
@jtibshirani Thanks Julie! Fantastic work !!!

server/src/main/java/org/elasticsearch/search/vectors/KnnSearchBuilder.java
jtibshirani: @sethmlarson the `query` and `knn` results will be combined together like a disjunction. This lets you combine term-based matches and vector-based matches for the same query text. The `filter` is a (non-scoring) query that limits the results that kNN search considers.

In the `_knn_search` API we put the `filter` at the top-level: https://github.com/elastic/elasticsearch/issues/81788. Here, we instead planning to nest it under the `knn` section like this:
```
"knn": {
    "field": "title_vector",
    "query_vector": [0.3f, 0.1f, ...],
    "k": 5,
    "num_candidates": 100,
    "filter": {
      "range": {
        "price": {
          "lte": 60
        }
      }
    }
  }
```

This will make it clearer that it differs from `query` and that it's specifically a prefilter on the kNN search.

jtibshirani:

jtibshirani:
Thanks @mayya-sharipova for the helpful review!

@jpountz do you have time to review just to check the overall strategy looks okay? As a reminder, the PR targets a feature branch knn-search. There are still several TODOs (tracked in https://github.com/elastic/elasticsearch/issues/87625) like support for filters.

server/src/main/java/org/elasticsearch/action/search/TransportSearchAction.java
```diff
@@ -478,6 +478,25 @@ private void executeRequest(
...
+            searchRequest.searchType(QUERY_THEN_FETCH);
+        }
+
+        // if there's only suggest, disable request cache and always use QUERY_THEN_FETCH
+        if (searchRequest.isSuggestOnly()) {
+            searchRequest.requestCache(false);
+            searchRequest.searchType(QUERY_THEN_FETCH);
+        }
+
+        // if there's a kNN search, always use DFS_QUERY_THEN_FETCH
```
jtibshirani, line 494: In a follow-up I'm thinking of handling this in `SearchSourceBuilder#rewrite` instead.

server/src/main/java/org/elasticsearch/action/search/FetchSearchPhase.java
```diff
@@ -102,7 +102,8 @@ private void innerRun() throws Exception {
         final boolean isScrollSearch = context.getRequest().scroll() != null;
         final List<SearchPhaseResult> phaseResults = queryResults.asList();
         final SearchPhaseController.ReducedQueryPhase reducedQueryPhase = resultConsumer.reduce();
-        final boolean queryAndFetchOptimization = queryResults.length() == 1;
+        final boolean queryAndFetchOptimization = queryResults.length() == 1
+            && context.getRequest().searchType() == SearchType.QUERY_THEN_FETCH;
```
jpountz, line 106: Why would we not run the fetch phase and the query phase in the same roundtrip when the search type has a DFS phase?

server/src/main/java/org/elasticsearch/search/vectors/ScoreDocQuery.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.Arrays;
+import java.util.Objects;
+
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
+
+/**
+ * A query that matches the provided docs with their scores.
+ *
+ * Note: this query was adapted from Lucene's DocAndScoreQuery from the
+ * class {@link org.apache.lucene.search.KnnVectorQuery}.
```
jpountz, line 33: maybe mention that it was copied because it's pkg-private in Lucene, not because of limitations of the Lucene impl?

server/src/main/java/org/elasticsearch/common/lucene/Lucene.java
```diff
@@ -384,6 +384,10 @@ public static ScoreDoc readScoreDoc(StreamInput in) throws IOException {
         return new ScoreDoc(in.readVInt(), in.readFloat());
     }
 
+    public static ScoreDoc readScoreDocWithShardIndex(StreamInput in) throws IOException {
+        return new ScoreDoc(in.readVInt(), in.readFloat(), in.readInt());
```
jpountz, line 388: Nit: the shard index should be greater than or equal to -1, so we could read it with `in.readVInt() - 1` to hopefully always encode it on a single byte.

jpountz (commented):
I only skimmed through the PR. One thing that looks a bit odd to me if I read the code correctly is that we compute global nearest vectors on the coordinating node and then send them all to all shards, where many of them might get ignored as part of `QueryBuilder#toQuery`. Would it be possible to only send the relevant vectors to each shard?

jpountz:
The closest thing that I know of is how searches sorted by descending `@timestamp` fan out to shards that hold recent data first and then to shards that hold older data with a `search_after` key that hopefully sometimes helps skip these shards entirely in the common case when the most recent documents were all in the recent shards.

I agree that it shouldn't be a big deal in practice. One could argue that it is `k` times the number of shards, but the size of these queries is still likely smaller than the responses that shards will send back.

jtibshirani:
@jpountz thanks for your comments! It sounds like things look okay overall, no major flags. In the interest of time, I'm going to continue with these PRs (merging into a feature branch) and tag both you and @mayya-sharipova for a final review.

server/src/main/java/org/elasticsearch/common/lucene/Lucene.java
jtibshirani: 👍

server/src/main/java/org/elasticsearch/search/vectors/ScoreDocQuery.java
jtibshirani: 👍

server/src/main/java/org/elasticsearch/action/search/FetchSearchPhase.java
jtibshirani: Previously, when there was a single shard we would always set the search type to `QUERY_THEN_FETCH`: https://github.com/elastic/elasticsearch/blob/8fb440dbf5bceb7df07bf80203834cc3eed82862/server/src/main/java/org/elasticsearch/action/search/TransportSearchAction.java#L988-L992

This makes sense, since you don't need DFS with one shard. Now, we might still execute `DFS_QUERY_THEN_FETCH` with a single shard, when kNN is used. From failing tests, I saw that we don't apply the "query and fetch" optimization in the query phase when DFS is enabled. (Surprisingly, we use different codepath to execute the query phase when DFS is enabled vs. not. There are two different actions `QUERY_ID_ACTION_NAME` and `QUERY_ACTION_NAME`).

Here was my thinking for next steps:
* In a follow-up, I plan to use QUERY_THEN_FETCH for kNN with a single shard. This will remove the special handling, since we will always know QUERY_THEN_FETCH is used when there's one shard.
* For now I'll change this check to `queryResults.length() == 1 && context.getRequest().hasKnnSearch();` and add a comment.

jtibshirani:
I pushed some updates:
* Add comment to clarify why the 'query and fetch' optimization doesn't work when kNN is enabled. It'd be best to support this -- I'll address this later when we tackle the TODO item "Optimize the single shard case".
* Only send the relevant score docs to each shard. This can be more efficient, especially when some shards don't return any kNN results (maybe because the filter doesn't match?) It also made things cleaner since I could avoid the new methods `Lucene.writeScoreDocWithShardIndex`. We do something similar in `SearchQueryThenFetchAsyncAction#rewriteShardSearchRequest`, so I used the same method name.

server/src/main/java/org/elasticsearch/action/search/FetchSearchPhase.java
jpountz: Thanks for explaining, the proposed follow-up sounds good to me.

server/src/main/java/org/elasticsearch/action/search/DfsQueryPhase.java
```diff
@@ -115,4 +128,30 @@ public void onFailure(Exception exception) {
...
+        }
+        scoreDocs.sort(Comparator.comparingInt(scoreDoc -> scoreDoc.doc));
+        KnnScoreDocQueryBuilder knnQuery = new KnnScoreDocQueryBuilder(scoreDocs.toArray(new ScoreDoc[0]));
+
+        SearchSourceBuilder newSource = source.shallowCopy().knnSearch(null);
+        if (source.query() == null) {
+            newSource.query(knnQuery);
+        } else {
+            newSource.query(new BoolQueryBuilder().should(knnQuery).should(source.query()));
+        }
```
jpountz, line 152: Should we specialize the case when no global vectors belong to one shard either here or through implementing `rewrite` on `KnnDocsQueryBuilder`?

jtibshirani: We do implement `KnnScoreDocQuery#rewrite`, we could implement `KnnScoreDocQueryBuilder#rewrite` for completeness too. This query is never used in `can_match` so not sure it makes a difference right now 🤔

jpountz: Agreed, I think we disable caching too with `DFS_QUERY_THEN_FETCH`, so it shouldn't matter.

sethmlarson (commented):
Looks good from an API perspective.

I'm guessing that the `filter` property of `knn` is being implemented in a future PR since there are no YAML tests cases with the property? If so that's fine.

server/src/main/java/org/elasticsearch/action/search/DfsQueryPhase.java
```diff
@@ -131,17 +129,29 @@ public void onFailure(Exception exception) {
...
 
-        KnnScoreDocQueryBuilder knnQueryBuilder = new KnnScoreDocQueryBuilder(knnResults.scoreDocs());
+        List<ScoreDoc> scoreDocs = new ArrayList<>();
+        for (ScoreDoc scoreDoc : knnResults.scoreDocs()) {
+            if (scoreDoc.shardIndex == request.shardRequestIndex()) {
+                scoreDocs.add(scoreDoc);
+            }
+        }
+        scoreDocs.sort(Comparator.comparingInt(scoreDoc -> scoreDoc.doc));
+        KnnScoreDocQueryBuilder knnQuery = new KnnScoreDocQueryBuilder(scoreDocs.toArray(new ScoreDoc[0]));
```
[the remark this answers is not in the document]
jtibshirani: We discussed this here: https://github.com/elastic/elasticsearch/pull/88002#discussion_r910127764. We make sure to rewrite `KnnVectorQueryBuilder` to `MatchNoneQueryBuilder` when there are no score docs, which has the same effect as avoiding adding the query in the first place (but I found to be simpler).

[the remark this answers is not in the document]
mayya-sharipova: Got, thanks for the clarification, rewriting to MatchNoneQueryBuilder makes sense to me!

mayya-sharipova (approved):
@jtibshirani Thanks, new changes LGTM@

PR #88007, opened 2022-06-24
[DOCS] Warn about impact of large readahead on search

dliappis:
When using LVM or software raid on Linux the kernel, or specific
distribution rules, may use higher ergonomic defaults for the
readahead of resulting block device(s). This can adversely affect
search performance due to high page cache thrashing, in search
heavy scenarios when mmap is involved.

Add a clarification section in the docs raising awareness about this
value and preferring the lower default.

### DIFF
--- a/docs/reference/how-to/search-speed.asciidoc
+++ b/docs/reference/how-to/search-speed.asciidoc
@@ -29,7 +29,7 @@ You can check the current value in `KiB` using
 Consult the documentation of your distribution on how to alter this value
 (for example with a `udev` rule to persist across reboots, or via
 https://man7.org/linux/man-pages/man8/blockdev.8.html[blockdev --setra]
-as a transient setting). We recommend a value of `128KiB` for `readahead`.
+as a transient setting). We recommend a value of `128KiB` for readahead.
 
 WARNING: `blockdev` expects values in 512 byte sectors whereas `lsblk` reports
 values in `KiB`. As an example, to temporarily set readahead to `128KiB`
--- a/docs/reference/how-to/search-speed.asciidoc
+++ b/docs/reference/how-to/search-speed.asciidoc
@@ -10,6 +10,27 @@ goes to the filesystem cache so that Elasticsearch can keep hot regions of the
 index in physical memory.
 
 [discrete]
+=== Avoid page cache thrashing by using modest read ahead values on Linux
+
+Search can cause a lot of randomized read IO. When using the default
+Elasticsearch filestore (fs/<<hybridfs,hybridfs>>), Lucene files using memory
+mapping may contribute to unnecessary read I/O when the underlying block device
+has a high readahead value.
+
+Most Linux distributions use a sensible readahead value of `128KiB` for a
+single plain device, however, when using software raid, LVM or dm-crypt the
+resulting block device (backing Elasticsearch <<path-settings,path.data>>)
+may end up having a very large readahead value (in the range of several MiB).
+This usually results in severe page (filesystem) cache thrashing adversely
+affecting search (or <<docs,update>>) performance.
+
+You can check the current value using `lsblk -o NAME,RA,MOUNTPOINT,TYPE,SIZE`.
+Consult the documentation of your distribution on how to alter this value
+(for example with a `udev` rule to persist across reboots, or via
+https://man7.org/linux/man-pages/man8/blockdev.8.html[blockdev setra]
+as a transient setting). We recommend a value of `128KiB` for `readahead`.
+
+[discrete]
 === Use faster hardware
 
 If your searches are I/O-bound, consider increasing the size of the filesystem
--- a/docs/reference/how-to/search-speed.asciidoc
+++ b/docs/reference/how-to/search-speed.asciidoc
@@ -28,7 +28,7 @@ You can check the current value in `KiB` using
 `lsblk -o NAME,RA,MOUNTPOINT,TYPE,SIZE`.
 Consult the documentation of your distribution on how to alter this value
 (for example with a `udev` rule to persist across reboots, or via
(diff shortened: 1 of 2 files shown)


docs/reference/how-to/search-speed.asciidoc
```diff
@@ -9,6 +9,27 @@ fast. In general, you should make sure that at least half the available memory
...
+[discrete]
+=== Avoid page cache thrashing by using modest read ahead values on Linux
+
+Search can cause a lot of randomized read IO. When using the default
+Elasticsearch filestore (fs/<<hybridfs,hybridfs>>), Lucene files using memory
+mapping may contribute to unnecessary read I/O when the underlying block device
+has a high readahead value.
+
+Most Linux distributions use a sensible readahead value of `128KiB` for a
+single plain device, however, when using software raid, LVM or dm-crypt the
```
dliappis, line 21: I don't think that `dm-crypt` by itself does that, but given that any combination of those maybe used, I am being defensive here and raising awareness that it is important to check the current readahead setting of the block device backing Elasticsearch's `path.data`

pquentin (approved):
Thanks! LGTM.

jpountz (approved):

docs/reference/how-to/search-speed.asciidoc
```diff
@@ -9,6 +9,27 @@ fast. In general, you should make sure that at least half the available memory
 goes to the filesystem cache so that Elasticsearch can keep hot regions of the
 index in physical memory.
 
+[discrete]
+=== Avoid page cache thrashing by using modest read ahead values on Linux
+
+Search can cause a lot of randomized read IO. When using the default
```
danielmitterdorfer, line 15: Actually readahead is used regardless of the file store (even when its `niofs`) as all implementations leverage the page cache? So maybe we should simplify this paragraph and only say:

```
Search can cause a lot of randomized read I/O. When the underlying block device has a high readahead value, there may be a lot of unnecessary read I/O done.
```

docs/reference/how-to/search-speed.asciidoc
```diff
@@ -9,6 +9,27 @@ fast. In general, you should make sure that at least half the available memory
...
+Elasticsearch filestore (fs/<<hybridfs,hybridfs>>), Lucene files using memory
+mapping may contribute to unnecessary read I/O when the underlying block device
+has a high readahead value.
+
+Most Linux distributions use a sensible readahead value of `128KiB` for a
+single plain device, however, when using software raid, LVM or dm-crypt the
+resulting block device (backing Elasticsearch <<path-settings,path.data>>)
+may end up having a very large readahead value (in the range of several MiB).
+This usually results in severe page (filesystem) cache thrashing adversely
```
danielmitterdorfer, line 24: I think the term "page cache" is established enough by now so we can remove `(filesystem)`?

docs/reference/how-to/search-speed.asciidoc
```diff
@@ -9,6 +9,27 @@ fast. In general, you should make sure that at least half the available memory
...
+affecting search (or <<docs,update>>) performance.
+
+You can check the current value using `lsblk -o NAME,RA,MOUNTPOINT,TYPE,SIZE`.
```
danielmitterdorfer, line 27: Maybe add a hint that `RA` denotes the readahead value in KiB?

danielmitterdorfer (commented):
Thanks, that's a great addition! I left a few suggestions.

docs/reference/how-to/search-speed.asciidoc
dliappis: That's right! However, my understanding is (e.g. here: https://biriukov.dev/docs/page-cache/0-linux-page-cache-for-sre/) that the impact is much more pronounced when using memory mapping.

docs/reference/how-to/search-speed.asciidoc
dliappis: Good point; unfortunately the term `filesystem cache` is used rather consistently in this doc instead of page cache (e.g. [here](https://github.com/elastic/elasticsearch/blame/227018225f2f70c148e4394c8baddd4176585356/docs/reference/how-to/search-speed.asciidoc#L380)). Therefore I wanted to use the more familiar to systems folks "page cache" term, but still have some connection with the rest of the doc.

docs/reference/how-to/search-speed.asciidoc
inqueue: ++ The impact is pronounced due to the nature of mmap. In randomized read I/O use cases, high readahead with memory-mapped files may likely result in loading the "wrong" pages into the page cache leading to more major page faults. Do we want to touch on this?

docs/reference/how-to/search-speed.asciidoc
dliappis: Addressed in e7e8609d2d72475f979022fe702930ae6937f52f

docs/reference/how-to/search-speed.asciidoc
```diff
@@ -9,6 +9,27 @@ fast. In general, you should make sure that at least half the available memory
 goes to the filesystem cache so that Elasticsearch can keep hot regions of the
 index in physical memory.
 
+[discrete]
+=== Avoid page cache thrashing by using modest read ahead values on Linux
```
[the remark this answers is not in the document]
dliappis: Addressed in e7e8609d2d72475f979022fe702930ae6937f52f

inqueue (approved):
LGTM, thank you!

docs/reference/how-to/search-speed.asciidoc
danielmitterdorfer: I see, that works for me then but I think we should get rid of the old name eventually in our docs.

danielmitterdorfer (approved):
Looks great with the new changes. Thanks!

PR #88009, opened 2022-06-24
[ML] make trained model rest APIs cancellable

### LINKED ISSUE #87931: Make GetTrainedModelsStatsAction cancellable

benwtrent:
This change makes all the trained model APIs cancellable, and addresses the handful of APIs that rely on our abstract resource structure.

closes: https://github.com/elastic/elasticsearch/issues/87931
relates: #88010

### DIFF
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/persistence/TrainedModelProvider.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/persistence/TrainedModelProvider.java
@@ -40,6 +40,7 @@ import org.elasticsearch.common.bytes.CompositeBytesReference;
 import org.elasticsearch.common.regex.Regex;
 import org.elasticsearch.common.util.set.Sets;
 import org.elasticsearch.common.xcontent.LoggingDeprecationHandler;
+import org.elasticsearch.core.Nullable;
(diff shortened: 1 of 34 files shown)


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/persistence/TrainedModelProvider.java
```diff
@@ -1174,30 +1208,6 @@ private Set<String> matchedResourceIds(String[] tokens) {
         return Collections.unmodifiableSet(matchedModels);
     }
 
-    private static <T> T handleSearchItem(
```
dimitris-athanasiou, line 1177: nice!

dimitris-athanasiou (approved):
LGTM

PR #88012, opened 2022-06-24
[ML] Rebalance should not notify listener before update is applied

dimitris-athanasiou:
When we are rebalancing trained model assignments, we eventually
update cluster state with the new metadata. We only want to
notify the listener after we have applied the cluster state update.
This commit fixes a bug where we could notify before the update
was actually applied resulting in NPE when writing the response
of the `CreateTrainedModelAssignmentAction`.

This is a fix following the work done in #87366.

### DIFF
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
@@ -358,6 +358,7 @@ public class TrainedModelAssignmentClusterService implements ClusterStateListene
     }
 
     private static ClusterState forceUpdate(ClusterState currentState, TrainedModelAssignmentMetadata.Builder modelAssignments) {
+        logger.debug(() -> format("updated assignments: %s", modelAssignments.build()));
         Metadata.Builder metadata = Metadata.builder(currentState.metadata());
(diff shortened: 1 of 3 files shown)


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -406,7 +411,9 @@ public void onFailure(Exception e) {
 
                 @Override
                 public void clusterStateProcessed(ClusterState oldState, ClusterState newState) {
-                    listener.onResponse(TrainedModelAssignmentMetadata.fromState(newState));
+                    if (isUpdated) {
+                        listener.onResponse(TrainedModelAssignmentMetadata.fromState(newState));
+                    }
```
benwtrent, line 416: Ah, yeah, this would happen because of the `return currentState` in execute kicking off another rebalancing.

benwtrent (approved):

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
dimitris-athanasiou: Yeah, you were right there was something off there!

Issue #87959, opened 2022-06-23
[CI] MLModelDeploymentsUpgradeIT testTrainedModelDeployment failing


andreidan:
**Build scan:**
https://gradle-enterprise.elastic.co/s/rrbldaxcklpvk/tests/:x-pack:qa:rolling-upgrade:v8.2.4%23twoThirdsUpgradedTest/org.elasticsearch.upgrades.MLModelDeploymentsUpgradeIT/testTrainedModelDeployment

**Reproduction line:**
`./gradlew ':x-pack:qa:rolling-upgrade:v8.2.4#twoThirdsUpgradedTest' -Dtests.class="org.elasticsearch.upgrades.MLModelDeploymentsUpgradeIT" -Dtests.method="testTrainedModelDeployment" -Dtests.seed=B48E616E1A60493F -Dtests.bwc=true -Dtests.locale=no-NO -Dtests.timezone=America/Kralendijk -Druntime.java=17`

**Applicable branches:**
master

**Reproduces locally?:**
Didn't try

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.upgrades.MLModelDeploymentsUpgradeIT&tests.test=testTrainedModelDeployment

**Failure excerpt:**
```
org.elasticsearch.client.ResponseException: method [POST], host [http://[::1]:35509], URI [/_ml/trained_models/upgrade-deployment-test/deployment/_infer], status line [HTTP/1.1 404 Not Found]
{"error":{"root_cause":[{"type":"status_exception","reason":"[upgrade-deployment-test] unable to find deployment task for inference please stop and start the deployment or try again momentarily"}],"type":"status_exception","reason":"[upgrade-deployment-test] unable to find deployment task for inference please stop and start the deployment or try again momentarily"},"status":404}

  at __randomizedtesting.SeedInfo.seed([B48E616E1A60493F:4F5627313BA0F3A8]:0)
  at org.elasticsearch.client.RestClient.convertResponse(RestClient.java:347)
  at org.elasticsearch.client.RestClient.performRequest(RestClient.java:313)
  at org.elasticsearch.client.RestClient.performRequest(RestClient.java:288)
  at org.elasticsearch.upgrades.MLModelDeploymentsUpgradeIT.infer(MLModelDeploymentsUpgradeIT.java:249)
[...]

PR #88015, opened 2022-06-24
Retry after all S3 get failures that made progress

### LINKED ISSUE #87243: Increase tenacity of S3RetryingInputStream
The `S3RetryingInputStream` hides cases where S3 closes a connection partway through downloading a blob. By default it retries 3 times before failing the download. However, the number of failures tends to increase on larger blobs and often 3 failures is not enough to complete a multi-GB download if S3 is suffering from a cluster of failures as sometimes happens. Typically we make 10s-to-100s of MBs of progress between each failure even in this state.

I think we should increase the tenacity of `S3RetryingInputStream` when downloading larger blobs. For instance, we could not count a partial download towards the retry limit if it makes significant progress before failing.

DaveCTurner:
S3 sometimes enters a state where blob downloads repeatedly fail but
with nontrivial progress between failures. Often each attempt yields 10s
or 100s of MBs of data. Today we abort a download after three (by
default) such failures, but this may not be enough to completely
retrieve a large blob during one of these flaky patches.

With this commit we start to avoid counting download attempts that
retrieved at least 1% of the configured `buffer_size` (typically 1MB)
towards the maximum number of retries.

Closes #87243


modules/repository-s3/src/main/java/org/elasticsearch/repositories/s3/S3RetryingInputStream.java
```diff
@@ -160,31 +162,31 @@ private void ensureOpen() {
...
-                blobKey,
-                start + currentOffset,
-                attempt,
-                maxAttempts
-            ),
-            e
+        final Supplier<String> messageSupplier = () -> format(
+            """
+                failed reading [%s/%s] at offset [%s]; this was attempt [%s] to read this blob which yielded [%s] bytes; in total \
+                [%s] of the attempts to read this blob have made meaningful progress; the maximum number of attempts is [%s]""",
```
kingherc, line 175: This log line will be a bit weird because it will say something like:

> this was attempt 103; in total 100 of the attempts made meaningful progress; the maximum number of attempts is 3.

Feels a bit weird that 103 >> 3. Maybe to make it more clear you could say at the end "the maximum number of attempts **(without meaningful progress)** is 3".

modules/repository-s3/src/test/java/org/elasticsearch/repositories/s3/S3BlobContainerRetriesTests.java
```diff
@@ -439,6 +443,82 @@ public void testWriteLargeBlobStreaming() throws Exception {
...
+        final int meaningfulProgressBytes = Math.max(1, bufferSizeBytes / 100);
+
+        final byte[] bytes = randomBlobContent();
+
+        @SuppressForbidden(reason = "use a http server")
+        class FlakyReadHandler implements HttpHandler {
+            private int failuresWithoutProgress;
+
+            @Override
+            public void handle(HttpExchange exchange) throws IOException {
```
kingherc, line 462: So here basically you first exhaust the "meaningless" failures, and then have a series of failures with meaningful progress that should not finally throw an exception and the blob should be successfully read (either in its entirety or partly).

kingherc (approved):
LGTM. Left one optional comment.

Of course please do not count my sole review for now -- it's my first review. Thanks for adding me!

modules/repository-s3/src/main/java/org/elasticsearch/repositories/s3/S3RetryingInputStream.java
DaveCTurner: :+1: good point

modules/repository-s3/src/test/java/org/elasticsearch/repositories/s3/S3BlobContainerRetriesTests.java
DaveCTurner: Yep, pretty much, although note that `sendIncompleteContent` will sometimes send a meaningful amount of data too.

DaveCTurner:
@elasticmachine please run elasticsearch-ci/bwc - failure tracked at https://github.com/elastic/elasticsearch/issues/87959

tlrx (approved):
LGTM

PR #88016, opened 2022-06-24
[ML] Stopping a model deployment should trigger rebalance

dimitris-athanasiou:
When a model deployment is stopped we remove the allocations
it was using. This opens up capacity for granting more allocations
to existing deployments that may not be fully satisfied.

With this commit we attempt to rebalance model assignments
after a model deployment has been stopped.


davidkyle (approved):
LGTM

PR #88017, opened 2022-06-24
Consolidate remaining startup in phase 3

rjernst:
This commit moves the remaining pieces of startup into a "phase 3". This
consists of everything after security manager is initialized. Phase 3
is where the node is constructed and the system becomes multi threaded.


grcevski (approved):
LGTM!

PR #88022, opened 2022-06-24
Initialize ES logging api for tests

rjernst:
This commit adds initialization of the new ES logging api into
ESTestCase, so that tests can begin using the new logging api.


pgomulka (approved):
LGTM

PR #88025, opened 2022-06-24
Synthetic source numbers in columns

nik9000:
This speeds up synthetic source for numbers and dates by loading them
column by column. On cached disk blocks, on average it's only 1ms per
for 1k documents, but it seems to help a fair bit in the worst case
and I expect it'll help much more on non-cached disk blocks.

```
|   50th percentile service time | default_1k | 32.9131 | 31.6141 | ms |  -3.95% |
|   90th percentile service time | default_1k | 34.937  | 34.8247 | ms |  -0.32% |
|   99th percentile service time | default_1k | 42.2246 | 40.0853 | ms |  -5.07% |
| 99.9th percentile service time | default_1k | 54.0964 | 41.993  | ms | -22.37% |
|  100th percentile service time | default_1k | 55.2969 | 53.4642 | ms |  -3.31% |
```


server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
```diff
@@ -1632,12 +1632,83 @@ protected NumericSyntheticFieldLoader(String name, String simpleName) {
...
+                    writeValue(b, dv.nextValue());
+                }
+                b.endArray();
+            }
+        }
+
+        /**
+         * Load all values for all docs up front. This should be much more
+         * disk and cpu-friendly than {@link ImmediateLeaf} because it resolves
+         * the values all at once, keeping the disk .
```
romseygeek, line 1692: keeping the disk ... in suspense?

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
```diff
@@ -1646,30 +1717,26 @@ public boolean empty() {
 
                 @Override
                 public boolean advanceToDoc(int docId) throws IOException {
-                    return hasValue = leaf.advanceExact(docId);
+                    idx = Arrays.binarySearch(docIdsInLeaf, docId);
```
romseygeek, line 1720: Do we need to do a binary search here, given that we know we're visiting the docs from `docIdsInLeaf` in-order?

romseygeek (approved):
LGTM, one nit and one question left.

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
nik9000: Are you saying everytime folks call this we just advance `idx` and check? I think that's fine, yes.

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
nik9000: https://www.youtube.com/watch?v=wlwnbcxBuzI

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
romseygeek: Yes, exactly.  Won't make a big difference for small sets of documents, but with `size=1000` and lots of fields then I think avoiding the binary search is a win.

PR #88029, opened 2022-06-24
REST tests for percentiles_bucket agg

nik9000:
Adds REST tests for the `percentiles_bucket` pipeline bucket
aggregation. This gives us forwards and backwards compatibility tests
for these aggs as well as mixed version cluster tests for these aggs.

Relates to #26220


not-napoleon (approved):
LGTM

PR #88035, opened 2022-06-24
Sort ingest pipeline stats by time spent executing

dakrone:
This commit changes the `/_nodes/stats` API to return the pipelines in descending order by the time
spent executing. If a pipeline ties another pipeline, they are sorted by ingest count.

This ensures that the most used and most expensive pipelines show up first in the JSON response (to
help tracking down which pipelines may be most prevalent).

For example:

```
{
  "nodes" : {
    "yiv09_ETRKSv4q8kz0d59A" : {
      ...
      "ingest" : {
        "total" : {...},
        "pipelines" : {
          "set-name" : {
            "count" : 4, // <-- tie-breaker
            "time_in_millis" : 12, // <-- sorted first by this
            ...
          },
          "set-face" : {
            "count" : 2,
            "time_in_millis" : 8,  // <-- less than 12
            ...
          }
```


dakrone:

server/src/main/java/org/elasticsearch/ingest/IngestStats.java
```diff
@@ -38,8 +38,18 @@ public class IngestStats implements Writeable, ToXContentFragment {
      */
     public IngestStats(Stats totalStats, List<PipelineStat> pipelineStats, Map<String, List<ProcessorStat>> processorStats) {
         this.totalStats = totalStats;
-        this.pipelineStats = pipelineStats;
+        this.pipelineStats = pipelineStats.stream().sorted((p1, p2) -> {
+            final IngestStats.Stats p2Stats = p2.stats;
+            final IngestStats.Stats p1Stats = p1.stats;
+            final int ingestCountCompare = Long.compare(p2Stats.ingestCount, p1Stats.ingestCount);
+            if (ingestCountCompare == 0) {
+                return Long.compare(p2Stats.ingestTimeInMillis, p1Stats.ingestTimeInMillis);
```
martijnvg, line 46: I wonder whether `ingestTimeInMillis` should be the primary sort property?
Because I think now pipelines that are used often but are relatively fast show up before pipelines that are used less often but are slower. I think the latter is a more important signal?

dakrone: Yeah I was on the fence about which one to be the primary and which one to be the fallback. I could go either way.

dakrone: Maybe @jakelandis would have an opinion here too, since he's looked more into ingest performance than I have

martijnvg: I prefer `ingestTimeInMillis` to be the primary sort property in this case, because I think pipelines with higher time spent are more interesting to look at first irregardless of how often these pipelines were used.

dakrone: Okay, that works for me too, I'll swap the order here.

martijnvg (approved):
LGTM

dakrone:
@elasticmachine update branch

Issue #87620, opened 2022-06-13
Setup elasticsearch dependency monitoring with Snyk


breskeby:
We want to monitor the elasticsearch dependencies on a regular basis using Snyk to be able to recognise vulnerabilities faster and being able to proactively react on security issues within the production dependencies. 

- [x] Setup Gradle plugin for monitoring dependencies in snyk ( [#88036](https://github.com/elastic/elasticsearch/pull/88036) )
- [x] Allow configuring snyk target reference and lifecycle properties ( [88220](https://github.com/elastic/elasticsearch/pull/88220) )
- [x] Setup CI job for automated update of snyk dependency monitoring of versions under development ( [88522](https://github.com/elastic/elasticsearch/pull/88522) )
- [x] Setup automation to update of snyk dependency monitoring of released versions

mark-vieira:
> My assumption (possibly mistaken) has been that published SBOMs would include only what is included in the released product and would not include test code.

This would be my understanding as well. I think publishing test dependencies would only cause confusion for folks. From my perspective test dependencies are completely internal and have nothing to do with the delivered software. I'd argue we should explicitly omit them for that reason, completely independent of the level of effort of including this information.

breskeby:
Understood. So for now we will only monitor dependencies for now that are shipped with our distro and can be a potential security thread for our customers.

breskeby:
We've integrated the snyk monitoring data publishing of released versions as part of the elasticsearch finalize release step

PR #88036, opened 2022-06-24
Setup elasticsearch dependency monitoring with Snyk for production code

breskeby:
This setups the generation and upload logic of gradle dependency graphs to snyk as raised in #87620

We directly implemented a rest api based snyk plugin as:

- the existing snyk gradle plugin delegates to the snyk command line tool
- the command line tool uses custom gradle logic by injecting a init file that is 
  a) using deprecated build logic which we definitely want to avoid
  b) uses gradle api we avoid like eager task creation.
- shipping this as a internal gradle plugin gives us the most flexibility

As we only want to monitor production code for now we apply this plugin as part of the `elasticsearch.build` plugin, that usage has been for now the de-facto indicator if a project is considered a "production" project that ends up in our distribution or public maven repositories. This isnt yet ideal and we will revisit the distinction between production and non production code / projects in a separate effort. 

The snyk upload can be triggered from command line via:

```
>./gradlew uploadSnykDependencyGraph -PsnykToken=XXX 
```  
where XXX is the api token that can be configured at https://app.snyk.io/account

The outcome of this can be found at https://app.snyk.io/org/elasticsearch/projects and https://app.snyk.io/org/elasticsearch/reports/ 
 
Setting up build jobs for regular updating the dependency monitoring in snyk will be covered in a separate PR.

As part of this effort we added the `elasticsearch.build` plugin to more projects that actually end up in the distribution. To unblock us on this we for now disabled a few check tasks that started failing by applying `elasticsearch.build`. We'll address this in separate PR at its out of scope of this story to fix.


build-conventions/src/main/java/org/elasticsearch/gradle/internal/conventions/info/GitInfo.java
```diff
@@ -168,7 +168,6 @@ private static String readFirstLine(final Path path) throws IOException {
 
     /** Find the reponame. */
     public String urlFromOrigin() {
-        String oritin = getOrigin();
```
breskeby, line 171: fixed typo on the way

build-tools-internal/src/integTest/groovy/org/elasticsearch/gradle/fixtures/http/HttpServerRule.groovy
```diff
@@ -0,0 +1,92 @@
...
+class HttpServerRule extends ExternalResource {
+
+    private HttpServer server
+    private List<String> contexts = new ArrayList<>()
+    private int port
+
+    @Override
+    protected void before() throws Throwable {
+        // TODO revisit port range allocation
+        def socket = new ServerSocket(0)
```
breskeby, line 28: We do not have a good concept for sharing port ranges within build-tools-internal tests as we only apply the Port allocator plugin on the elasticsearch root project for now. This is not optimal, but I wonder if this is good enough here for now or will cause problems on CI until we proper address the problem. thoughts?

build-tools-internal/src/integTest/groovy/org/elasticsearch/gradle/internal/snyk/SnykDependencyMonitoringGradlePluginFuncTest.groovy
```diff
@@ -0,0 +1,200 @@
...
+
+class SnykDependencyMonitoringGradlePluginFuncTest extends AbstractGradleInternalPluginFuncTest {
+
+    @Rule
+    public HttpServerRule httpServer = new HttpServerRule()
+
+    Class<? extends Plugin> pluginClassUnderTest = SnykDependencyMonitoringGradlePlugin.class
+
+    def setup() {
+        configurationCacheCompatible = false // configuration is not cc compliant
```
breskeby, line 27: For creating the dependency graph we need to work on the `Configuration` object that is not compliant with the gradle configuration-cache at the moment. This should not affect our developers as this task should never be triggered in day to day life from dev environments

build-tools-internal/src/integTest/groovy/org/elasticsearch/gradle/internal/snyk/SnykDependencyMonitoringGradlePluginFuncTest.groovy
```diff
@@ -0,0 +1,200 @@
...
+        buildFile << """
+            apply plugin:'java'
+            version = "1.0-SNAPSHOT"
+
+            repositories {
+                mavenCentral() 
+            }
+            
+            dependencies {
+                implementation 'org.apache.lucene:lucene-monitor:9.2.0'
```
breskeby, line 41: For now, I don't want to go down the route of creating a maven test fixture to reproduce dependency graphs from scratch. Therefore we test against real dependencies from real repositories.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/snyk/UploadSnykDependenciesGraph.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @TaskAction
+    void upload() {
+        String endpoint = calculateEffectiveEndpoint();
+        CloseableHttpResponse response;
+        try (CloseableHttpClient client = HttpClients.createDefault()) {
+            String content = FileUtils.readFileToString(inputFile.getAsFile().get());
+            HttpPut putRequest = new HttpPut(endpoint);
+            putRequest.addHeader("Authorization", "token " + token.get());
+            putRequest.addHeader("Content-Type", "application/json");
```
breskeby, line 59: For now we send the plain json in text format to the endpoint. Snyk supports gzipped data here but I didn't see a major performance issue here at all that would make the effort to support gz worth it for now.

build-tools/src/testFixtures/groovy/org/elasticsearch/gradle/fixtures/AbstractGradleFuncTest.groovy
```diff
@@ -33,14 +34,16 @@ abstract class AbstractGradleFuncTest extends Specification {
     File propertiesFile
     File projectDir
 
+    public boolean configurationCacheCompatible = false
```
breskeby, line 37: This adds support for testing our plugins with configuration-cache. at one point we can hopefully make the switch to `true` by default.

client/benchmark/build.gradle
```diff
@@ -17,8 +17,7 @@ tasks.named("assemble").configure {  enabled = true }
 archivesBaseName = 'client-benchmarks'
 mainClassName = 'org.elasticsearch.client.benchmark.BenchmarkMain'
 
-// never try to invoke tests on the benchmark project - there aren't any
-tasks.named("test").configure {enabled = false }
```
breskeby, line 21: I removed those explicit set tests to not execute. Gradle will shortcut that anyhow as no source is detected. It also makes the outcome of that task more clear IMO as it will tell `NO_SOURCE` instead of SKIPPED which can have different reasons.

client/client-benchmark-noop-api-plugin/build.gradle
```diff
@@ -20,5 +20,4 @@ esplugin {
 tasks.named("assemble").configure { enabled = false }
 tasks.named("dependencyLicenses").configure { enabled = false }
 tasks.named("dependenciesInfo").configure { enabled = false }
-// no unit tests
-tasks.named("test").configure { enabled = false }
+tasks.named("uploadSnykDependencyGraph").configure { enabled = false }
```
breskeby, line 23: this plugin applies `elasticsearch.build` but I don't consider it production code

client/benchmark/build.gradle
```diff
@@ -17,8 +17,7 @@ tasks.named("assemble").configure {  enabled = true }
...
+tasks.named("uploadSnykDependencyGraph").configure {enabled = false }
```
breskeby, line 20: this plugin applies `elasticsearch.build` but I don't consider it production code

distribution/tools/windows-service-cli/build.gradle
```diff
@@ -7,3 +7,7 @@ dependencies {
 
   testImplementation project(":test:framework")
 }
+
```
breskeby, line 10: plenty of forbidden api failures. TODO: raise ticket

plugins/repository-hdfs/hadoop-client-api/build.gradle
```diff
@@ -8,3 +8,9 @@ dependencies {
 tasks.named('shadowJar').configure {
   exclude 'org/apache/hadoop/util/ShutdownHookManager$*.class'
 }
+
+['jarHell', 'thirdPartyAudit', 'forbiddenApisMain', 'splitPackagesAudit'].each {
```
breskeby, line 12: All kinds of problems here that have been not discovered before. 
TODO: raise ticket

test/framework/build.gradle
```diff
@@ -47,7 +47,6 @@ tasks.named('forbiddenApisMain').configure {
 // TODO: should we have licenses for our test deps?
 tasks.named("dependencyLicenses").configure { enabled = false }
 tasks.named("dependenciesInfo").configure { enabled = false }
-tasks.named("dependenciesGraph").configure { enabled = false }
```
breskeby, line 50: I wonder if this should be part of the snyk check or not. It seems we upload this somewhere so I would consider it kind of production code

test/x-content/build.gradle
```diff
@@ -30,7 +30,6 @@ tasks.named('forbiddenApisMain').configure {
 // TODO: should we have licenses for our test deps?
 tasks.named("dependencyLicenses").configure { enabled = false }
 tasks.named("dependenciesInfo").configure { enabled = false }
-tasks.named("dependenciesGraph").configure { enabled = false }
```
breskeby, line 33: same as above

build-tools-internal/build.gradle
```diff
@@ -291,6 +291,7 @@ dependencies {
     exclude module: "groovy"
   }
   testImplementation buildLibs.spock.junit4
+  testImplementation buildLibs.json.assert
```
breskeby, line 294: allows asserting generated json without hassle.

build-tools-internal/src/integTest/groovy/org/elasticsearch/gradle/fixtures/http/HttpServerRule.groovy
breskeby: Now looking into this I might just port this to our existing WireMock infrastructure 🤦

breskeby:
@jkakavas 

about 1) The way this works is split in 2 tasks. If you just run `"generateSnykDependencyGraph"` you can generate the file  containing actually the json as we push it to snyk.
 about 2) Indeed we have plans to move to transitive dependency handling at one point. The logic in this PR on how the graph is created takes this already into account. The hierarchy is not flat as it was with the former `DependenciesGraphPlugin`. So we will see which how a dependency was brought in in the first place.

client/benchmark/build.gradle
mark-vieira, line 20: Maybe this project simply shouldn't have the `elasticsearch.build` plugin applied and could just use `elasticsearch.java` instead?

mark-vieira (approved):
Couple minor comments but otherwise this looks great.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/snyk/GenerateSnykDependencyGraph.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    private Map<String, Object> generateGradleGraphPayload() {
+        Set<ResolvedDependency> firstLevelModuleDependencies = configuration.get()
+            .getResolvedConfiguration()
+            .getFirstLevelModuleDependencies();
+        SnykDependencyGraphBuilder builder = new SnykDependencyGraphBuilder(gradleVersion.get());
+        String effectiveProjectPath = projectPath.get();
+        builder.walkGraph(
+            (effectiveProjectPath.equals(":") ? projectName.get() : effectiveProjectPath),
```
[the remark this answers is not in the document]
breskeby: I mimicked the existing snyk command line behaviour here. You could argue that we should change this to use the same coordinates but tbh I think that that doesn't add much value and I think having the same project naming (e.g. `:server` makes navigating through the project list at https://app.snyk.io/org/elasticsearch/projects easier IMO and less confusing.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/snyk/GenerateSnykDependencyGraph.java
```diff
@@ -0,0 +1,144 @@
...
+        return Map.of(
+            "remoteUrl",
+            "http://github.com/elastic/elasticsearch.git",
+            "branch",
+            GitInfo.gitInfo(buildLayout.getRootDirectory()).getRevision()
+        );
+    }
+
+    private Map<String, Object> generateMetaData() {
+        return Map.of(
```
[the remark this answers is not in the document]
breskeby: 👍

client/benchmark/build.gradle
breskeby: that's what I thought

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/snyk/UploadSnykDependenciesGraph.java
```diff
@@ -0,0 +1,104 @@
...
+            putRequest.setEntity(new StringEntity(content));
```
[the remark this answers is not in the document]
breskeby: yes definitely. nice improvement.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/snyk/SnykDependencyMonitoringGradlePlugin.java
```diff
@@ -0,0 +1,62 @@
...
+            // the elasticsearch snyk project id
+            t.getProjectId().set("f27934bf-9ad1-4d91-901c-cb77168a34db");
+        });
+
+        project.getPlugins().withType(JavaPlugin.class, javaPlugin -> generateTaskProvider.configure(generateSnykDependencyGraph -> {
+            SourceSet main = project.getExtensions()
+                .getByType(JavaPluginExtension.class)
+                .getSourceSets()
+                .getByName(SourceSet.MAIN_SOURCE_SET_NAME);
+            Configuration runtimeConfiguration = project.getConfigurations().getByName(main.getRuntimeClasspathConfigurationName());
```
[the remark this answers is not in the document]
breskeby: you mean projectId? we can make it so. makes the api cleaner I guess

PR #88041, opened 2022-06-25
Consolidate bootstrap state

rjernst:
Now that initialization is broken into phases, the state between these
phases can be better organized. This commit reintroduces the Bootstrap
class, but now uses it as the holder for state between the phases.


rjernst:
> My guess is that some non-daemon thread isn't ending

That is my guess too. I added a dump of the ES log and it's clear that ES is trying to shut down. I'll keep iterating on the debugging.

rjernst:
@grcevski I think I fixed the issue, can you take another look?

grcevski (approved):
LGTM! Wow, it was printing the ready marker as int vs char.

PR #88043, opened 2022-06-26
Presize RoutingNode

original-brownbear:
Presize maps in `RoutingNode` with the best-guess that every node
holds it's proportionate share of double the number of indices which
seems like the best possible guess (1 shards + 1 replica per index,
all nodes homogeneous).
This saves a lot of re-hashing when data nodes hold lots of shards
and speeds up bootstrapping 50k indices in the many-shards benchmark
by more than 2 percent.

relates #77466


original-brownbear:
Jenkins run elasticsearch-ci/bwc

idegtiarenko (approved):

PR #88045, opened 2022-06-26
Faster DatastreamMetadata Diff application

original-brownbear:
This was doing a needless copy which showed up in many shards benchmarking.

relates #77466


martijnvg (approved):
LGTM

original-brownbear:
Thanks Martijn!

PR #88050, opened 2022-06-27
Make AsyncPersistedState.resetVotingConfiguration fast

original-brownbear:
Using the builder just to update that one field in the metadata means
running through the name collision checks on the metadata which are very
expensive. In the many-shards benchmark run it takes ~2% of total CPU
time just for this operation.
Even though this is running async on a thread off the critical path I think
it's worthwhile fixing this to free up some CPU for other tasks.

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
+++ b/server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
@@ -363,6 +363,34 @@ public class Metadata extends AbstractCollection<IndexMetadata> implements Diffa
         );
     }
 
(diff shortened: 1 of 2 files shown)


DaveCTurner (approved):
Seems reasonable :+1:

original-brownbear:
Jenkins run elasticsearch-ci/bwc  (random Jenkins issue)

original-brownbear:
Jenkins run elasticsearch-ci/part-1 (real failure but unrelated will investigate separately)

original-brownbear:
Thanks David!

PR #88052, opened 2022-06-27
[TRANSFORM] Add YAML test assertions for roles in transform configs

droberts195:
In #87570 authorization info was added to transform listings.
Authorization information was also implicitly added to transform
update responses. However, YAML tests were not added to either.
This PR adds some assertions to the YAML tests for getting transforms
and transform updates to prove that the responses really do include
authorization information.


przemekwitek (approved):
LGTM

PR #88060, opened 2022-06-27
Replace ilm/slm with their full names

andreidan:
This also fixes some doc typos "reponse" -> "response".


docs/reference/tab-widgets/troubleshooting/data/start-ilm.asciidoc
```diff
@@ -61,14 +61,14 @@ The reponse will look like this:
 // end::cloud[]
 
 // tag::self-managed[]
-<<ilm-start, Start ILM>>:
+<<ilm-start, Start Index Lifecycle Management>>:
```
gmarouli, line 64: I am curious why are we specifically typing here `Index Lifecycle Management` and not `{ilm}` like we do in the `start-slm` doc?

andreidan:
@elasticmachine update branch

docs/reference/tab-widgets/troubleshooting/data/start-ilm.asciidoc
gmarouli: I see, that's why in the slm it is `<<ilm-start, Start>> {slm}`. I would suggest to do the same here so we are consistently using the macros.

gmarouli (approved):
LGTM, just a small suggestion about using a macro.

docs/reference/tab-widgets/troubleshooting/data/start-ilm.asciidoc
andreidan, line 64:
```suggestion
<<ilm-start, Start>> {ilm}:
```

PR #88066, opened 2022-06-27
Handle throttling in deciders

idegtiarenko:
This change whitelists `DECIDERS_THROTTLED` status and awaits next reroute to converge the computation


DaveCTurner (approved):
LGTM

PR #88068, opened 2022-06-27
Remove unnecessary string concatenation

idegtiarenko:
We have lots of unnecessary string concatenation in our code base.
This pr removes some of it.


libs/x-content/src/main/java/org/elasticsearch/xcontent/ObjectParser.java
```diff
@@ -773,6 +773,6 @@ public EnumSet<XContentParser.Token> supportedTokens() {
 
     @Override
     public String toString() {
-        return "ObjectParser{" + "name='" + name + '\'' + ", fields=" + fieldParserMap + '}';
+        return "ObjectParser{name='" + name + "', fields=" + fieldParserMap + '}';
```
idegtiarenko, line 776: `'\'' + ` should also be part of the following constant string on the same line

PR #88070, opened 2022-06-27
[ML] muting MLModelDeploymentsUpgradeIT #87959

benwtrent:
relates to: https://github.com/elastic/elasticsearch/issues/87959


dimitris-athanasiou:
Could you please hold off muting the test? I'm missing logs for `org.elasticsearch.xpack.ml.inference.allocation` as old nodes will have the package names before the rename.

dimitris-athanasiou:
Ah, no, my bad. I already have logging for the higher level `inference` so I should be fine. Hm. Ok, proceed with the mute.

PR #88072, opened 2022-06-27
Fix migrate to data stream IT

idegtiarenko:
The failure was caused by missing listener call on the branch that does not require index creation


DaveCTurner (approved):
LGTM

PR #88073, opened 2022-06-27
Skip building redundant set in `org.elasticsearch.indices.cluster.IndicesClusterStateService#removeIndices`

original-brownbear:
No need to build a fresh set of all the indices, we already have a map of all of them
in the routing nodes that we can check. This saves quite a bit of work on applying CS
to data nodes with many indices.

relates #77466


idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

PR #88082, opened 2022-06-27
Add generic interface for loading service providers from plugins

grcevski:
Since plugins/modules are loaded in their own class loader, loading service providers (SPI implementations) from various service interfaces in plugins is not a simple call to `ServiceLoader.load`. This PR adds a helper method on PluginsService to load providers from plugins/modules given a service class.

Example usage of this helper can be found here (in the related PR #86224):

https://github.com/elastic/elasticsearch/pull/86224/files#diff-b56bed42f5b0025886eb243ecb48678ac048bfba0ef047545dbb6504e357edf2R901

Relates to https://github.com/elastic/elasticsearch/pull/86224


server/src/main/resources/org/elasticsearch/bootstrap/security.policy
```diff
@@ -170,4 +170,7 @@ grant {
 
   // system memory on Linux systems affected by JDK bug (#66629)
   permission java.io.FilePermission "/proc/meminfo", "read";
+
+  // needed for PluginsServiceTests
+  permission java.lang.RuntimePermission "createClassLoader";
```
grcevski, line 175: I've made a mistake here, wrong policy file. I'll submit a new commit.

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -293,6 +294,34 @@ static void loadExtensions(Collection<LoadedPlugin> plugins) {
...
+
+    /**
+     * SPI convenience method that uses the {@link ServiceLoader} JDK class to load various SPI providers
+     * from plugins/modules.
+     *
+     * <p>
+     * For example:
+     *
+     * var pluginHandlers = pluginsService.loadServiceProviders(OperatorHandlerProvider.class);
+     * </p>
```
rjernst, line 310: nit: I *think* the closing p tag is normally omitted in javadoc (or must be?)

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -293,6 +294,34 @@ static void loadExtensions(Collection<LoadedPlugin> plugins) {
...
+    }
+
+    /**
+     * SPI convenience method that uses the {@link ServiceLoader} JDK class to load various SPI providers
+     * from plugins/modules.
+     *
+     * <p>
+     * For example:
+     *
+     * var pluginHandlers = pluginsService.loadServiceProviders(OperatorHandlerProvider.class);
```
rjernst, line 309: This should be in a `<pre>` block?

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -293,6 +294,34 @@ static void loadExtensions(Collection<LoadedPlugin> plugins) {
...
+     * </p>
+     * @param service A templated service class to look for providers in plugins
+     * @return an immutable {@link List} of discovered providers in the plugins/modules
+     */
+    public <T> List<? extends T> loadServiceProviders(Class<T> service) {
+        List<T> result = new ArrayList<>();
+        getClass().getModule().addUses(service);
```
rjernst, line 316: Why is this necessary? If a type is known in server, then it can be declared explicitly in server's module-info?

server/src/test/java/org/elasticsearch/plugins/PluginsServiceTests.java
```diff
@@ -613,6 +621,76 @@ public void testThrowingConstructor() {
...
+            mock(PluginDescriptor.class),
+            fakePlugin1,
+            fakeClassLoader1,
+            ModuleLayer.boot()
+        );
+        ((MockPluginsService) service).setLoadedPlugins(List.of(plugin1, plugin2));
+
+        service = spy(service);
+
+        doReturn(List.of(provider1).iterator()).when(service).providersIterator(eq(OperatorHandlerProvider.class), eq(fakeClassLoader));
```
rjernst, line 671: Since we are dealing with different classloaders, we should probably test more real-world situations. Can the test use real classloaders instead of mocking them? We have utilities for creating dummy jars with module infos, classes, etc. It seems doable to actually test the service loading.

rjernst (commented):
I have a few questions

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
grcevski: I thought of it as a convenience, but we can always explicitly declare the use in module-info. I'll remove it.

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
grcevski: I thought it was allowed, I see we use it in few places, but it's stripped now and I'll remove it from other PR.

rjernst (commented):
Sorry I had these comments stuck pending 🤦

grcevski:
Thanks for the review @rjernst, I removed all mocking, real SPI loading now.

server/src/main/java/org/elasticsearch/plugins/PluginsService.java
```diff
@@ -293,6 +294,33 @@ static void loadExtensions(Collection<LoadedPlugin> plugins) {
         }
     }
 
+    // package private for testing
+    <T> Iterator<T> providersIterator(Class<T> service, ClassLoader classLoader) {
```
rjernst, line 298: I think this can be folded into loadServiceProviders now that the tests don't mock it?

server/src/test/java/org/elasticsearch/plugins/PluginsServiceTests.java
```diff
@@ -613,6 +622,66 @@ public void testThrowingConstructor() {
         assertThat(e.getCause().getCause(), hasToString(containsString("test constructor failure")));
     }
 
+    private ClassLoader buildTestProviderPlugin(String name) throws Exception {
+        String pluginClass = Strings.format("""
+            package r;
+            import org.elasticsearch.plugins.spi.TestService;
+            import org.elasticsearch.plugins.ActionPlugin;
+            import org.elasticsearch.plugins.Plugin;
+            public final class FooPlugin extends Plugin implements ActionPlugin, TestService {
```
rjernst, line 631: I would separate this out. Due to the current plugin design, a Plugin subclass must be present, but this class should not be required to be the one that implements all service providers. In fact, I think it is important we test them decoupled, since that is the future we are working towards.

rjernst (changes requested):
The new test looks much better! I'm glad the recently added utilities help make testing this easier. Just one last nit on the testing.

server/src/test/java/org/elasticsearch/plugins/PluginsServiceTests.java
grcevski: Sure thing, I was just lazy :). I'll create a separate TestService provider class.

grcevski: Done now. I've split the provider class into separate file.

rjernst (approved):
Thanks for the iterations. LGTM!

Issue #86607, opened 2022-05-10
Publish maven artifact with CircleProcessor class


porunov:
### Description

It would be really great if [CircleProcessor.java](https://github.com/elastic/elasticsearch/blob/0699c9351f1439e246d408fd6538deafde4087b6/x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/ingest/CircleProcessor.java) could be published in some maven artifact.
We are thinking to convert Geoshape Circle into Geoshape Polygon on application side instead of ElasticSearch Server side. The benefits of it is that it would make possible for us to choose `error_distance` dynamically depending on different conditions.
With direct code access to [CircleProcessor.java](https://github.com/elastic/elasticsearch/blob/0699c9351f1439e246d408fd6538deafde4087b6/x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/ingest/CircleProcessor.java) and [SpatialUtils.java](https://github.com/elastic/elasticsearch/blob/0699c9351f1439e246d408fd6538deafde4087b6/x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/SpatialUtils.java) it would be possible to make application side custom conversion logic.
ElasticSearch version: >= 8

iverase:
We discussed in the team and we think this is a fair ask and we could move the CircleProcessor logic into the geo library which is provided as an artifact.

porunov:
> We discussed in the team and we think this is a fair ask and we could move the CircleProcessor logic into the geo library which is provided as an artifact. 

Thank you! It would be really helpful.

porunov:
@iverase I was trying to resolve this issue but have some problems with introducing lucene dependency to `geo` library.
Basically, `CircleProcessor` is using `SpatialUtils` class.
`SpatialUtils` class depends on `org.apache.lucene.util.SloppyMath` which is provided via `lucene-core`.
I tried to include `api "org.apache.lucene:lucene-core:${versions.lucene}"` into `libs/geo/build.gradle` but it didn't help. For some reason this library isn't included into the classpath after build is finished. Looks like some other building logic excluding `lucene-core` for some reason.
I created PR to who the problem here: #88088 
If you have any suggestions on how to resolve this issue, that would be super helpful.

iverase:
This is done in 8.6.0/

PR #88088, opened 2022-06-27
Move SpatialUtils to geo library

porunov:
Moving parts of the code related to calculating Polygon from Circle.

Fixes to #86607

### DIFF
--- a/x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/ingest/CircleProcessor.java
+++ b/x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/ingest/CircleProcessor.java
@@ -12,6 +12,7 @@ import org.elasticsearch.common.xcontent.XContentHelper;
 import org.elasticsearch.geometry.Circle;
 import org.elasticsearch.geometry.Geometry;
 import org.elasticsearch.geometry.ShapeType;
+import org.elasticsearch.geometry.utils.CircleUtils;
 import org.elasticsearch.geometry.utils.StandardValidator;
 import org.elasticsearch.ingest.AbstractProcessor;
 import org.elasticsearch.ingest.ConfigurationUtils;
@@ -24,7 +25,6 @@ import org.elasticsearch.xcontent.XContentBuilder;
 import org.elasticsearch.xcontent.XContentFactory;
 import org.elasticsearch.xcontent.XContentType;
 import org.elasticsearch.xcontent.support.MapXContentParser;
-import org.elasticsearch.xpack.spatial.SpatialUtils;
 
 import java.util.Arrays;
 import java.util.EnumSet;
@@ -36,8 +36,6 @@ import java.util.Map;
  */
 public final class CircleProcessor extends AbstractProcessor {
     public static final String TYPE = "circle";
-    static final int MINIMUM_NUMBER_OF_SIDES = 4;
-    static final int MAXIMUM_NUMBER_OF_SIDES = 1000;
 
     private final String field;
     private final String targetField;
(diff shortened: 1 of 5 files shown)


libs/geo/src/main/java/org/elasticsearch/geometry/utils/SpatialUtils.java
```diff
@@ -4,19 +4,21 @@
...
 import org.apache.lucene.util.SloppyMath;
 import org.elasticsearch.geometry.Circle;
 import org.elasticsearch.geometry.LinearRing;
 import org.elasticsearch.geometry.Polygon;
-import org.elasticsearch.index.mapper.GeoShapeIndexer;
 
 /**
  * Utility class for storing different helpful re-usable spatial functions
  */
 public class SpatialUtils {
```
iverase, line 17: Can we change the name of the class now that it is in this library. Maybe CircleUtils?

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/ingest/CircleProcessor.java
```diff
@@ -94,7 +92,7 @@ public IngestDocument execute(IngestDocument ingestDocument) {
             Geometry geometry = geometryFormat.fromXContent(StandardValidator.instance(true), true, true, parser);
             if (ShapeType.CIRCLE.equals(geometry.type())) {
                 Circle circle = (Circle) geometry;
-                int numSides = numSides(circle.getRadiusMeters());
+                int numSides = SpatialUtils.circleToPolygonNumSides(circle.getRadiusMeters(), errorDistance);
```
iverase, line 95: See above, maybe use strict formula so we don't need to bring the dependency?

iverase (changes requested):

iverase:
@elasticsearchmachine  ok to test

iverase:
I think you need to put the appropriate license header on those files, in this case would be the SSPL ones.

porunov:
> I think you need to put the appropriate license header on those files, in this case would be the SSPL ones. In generate you can run ./gradlew precommit locally to see the errors and fix them until precommit is happy.

Thank you. This command is helpful. There were also some formatting problems which are now fixed.

porunov:
Looks like after formatting fix the test was broken Oo. Looking at it

iverase:
The packaging-tests-windows-sample is not your fault

x-pack/plugin/spatial/src/test/java/org/elasticsearch/xpack/spatial/ingest/CircleProcessorTests.java
```diff
@@ -60,22 +58,6 @@
...
-        double radiusDistanceMeters = randomDoubleBetween(0.01, 6371000, true);
-        CircleShapeFieldType shapeType = randomFrom(SHAPE, GEO_SHAPE);
-        CircleProcessor processor = new CircleProcessor("tag", null, "field", "field", false, radiusDistanceMeters, shapeType);
-
-        // radius is same as error distance
-        assertThat(processor.numSides(radiusDistanceMeters), equalTo(4));
-        // radius is much smaller than error distance
-        assertThat(processor.numSides(0), equalTo(4));
-        // radius is much larger than error distance
-        assertThat(processor.numSides(Math.pow(radiusDistanceMeters, 100)), equalTo(1000));
```
porunov, line 73: The previous test is flaky. If `radiusDistanceMeters` is `1.11` or less then this assert won't pass because when we power the number in 100 it won't be big enough. Moreover, if the number is less then 1 then power will work in another direction which will make `radiusDistanceMeters` smaller then the error distance used.
I'm will fix it in the test

porunov: @iverase fixed the test. It's now called `testCircleToPolygonNumSides`. In addition to testing random values between 0.01 and 6371000 it's now testing fixed small and big values as well to avoid flakiness.

porunov:
> The packaging-tests-windows-sample is not your fault

All CI checks passed except the one you mentioned. Do you know if it is considered to be a blocker for the PR to be merged? If yes, should I rebase the PR's branch to include some commits with the fix? I see that CI in main branch passes but in the PR it doesn't for some reason.

iverase:

iverase (approved):
LGTM

iverase:
Thanks @porunov for the contribution!

porunov:
Thank you @iverase for helping with the PR!

PR #88097, opened 2022-06-27
Implement few operator handlers

grcevski:
Implement operator handlers for cluster state and ILM.

These operator handlers are the first of the operator handlers we'll be implementing, one specific to cluster state settings and another dealing with cluster state entities (ILM).


x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/action/TransportPutLifecycleAction.java
```diff
@@ -121,21 +125,44 @@ public static class UpdateLifecyclePolicyTask extends AckedClusterStateUpdateTas
...
 
         public UpdateLifecyclePolicyTask(
             Request request,
             ActionListener<AcknowledgedResponse> listener,
             XPackLicenseState licenseState,
             Map<String, String> filteredHeaders,
             NamedXContentRegistry xContentRegistry,
-            Client client
+            Client client,
+            boolean verboseLogging
```
martijnvg, line 137: Maybe remove the `verboseLogging` param here? 
and just set the `verboseLogging` to `true`?
Then the other constructor can't reuse this constructor, but
then for regular usage of this class, verbose logging is guaranteed to be enabled.

server/src/main/java/org/elasticsearch/action/support/master/TransportMasterNodeAction.java
```diff
@@ -142,6 +145,33 @@ private ClusterBlockException checkBlockIfStateRecovered(Request request, Cluste
...
+     * This method is used by the operator handler logic (see {@link org.elasticsearch.operator.OperatorHandler})
+     * to verify if the keys don't conflict with an existing key set in operator mode.
+     *
+     * @param request the TransportMasterNode request
+     * @return set of String keys intended to be modified/set/deleted by this request
+     */
+    protected Set<String> modifiedKeys(Request request) {
+        return Collections.emptySet();
+    }
+
```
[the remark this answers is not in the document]
martijnvg: Regarding the `validateForOperatorState(...)` validation method in the referenced link, 
I do think the logic looks good, but I do think this also needs to be validated as part
of a cluster state update. Right now the validation there only occurs on the action/transport
level and there is no guarantee that what is validated at that point in time is also 
true when the update if really performed. I think it is a good pre-check, that is likely
to catch almost all cases when resources are updates that are being managed operator state.

For example in order to guarantee that no operator state managed ilm policies are modified 
via API. The UpdateLifecyclePolicyTask/DeleteLifecyclePolicyTask should do this validation
check as well as part of their respective `execute(...)` method.

martijnvg (commented):
The ILM parts look good, I left some minor comments around it.
I also commented on [your comment](https://github.com/elastic/elasticsearch/pull/88097#discussion_r907789427), which is a bit more involved,
but I think falls a bit outside of this PR, but I think is important.

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/immutablestate/ILMImmutableStateHandlerProvider.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+public class ILMImmutableStateHandlerProvider implements ImmutableClusterStateHandlerProvider {
+    private static final Set<ImmutableClusterStateHandler<?>> handlers = ConcurrentHashMap.newKeySet();
+
+    @Override
+    public Collection<ImmutableClusterStateHandler<?>> handlers() {
+        return handlers;
+    }
+
+    public static void handler(ImmutableClusterStateHandler<?> handler) {
```
martijnvg, line 28: I think does need to be invoked and should provide a `ImmutableLifecycleAction` instance, right?

grcevski: Ah nice catch, I should've brought in this change in this PR. https://github.com/elastic/elasticsearch/pull/86224/files#diff-245140b1f56e32ce2f87c455562395af21e0b6730984b7077dd236b7086e5164R275

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/immutablestate/action/ImmutableLifecycleAction.java
```diff
@@ -0,0 +1,113 @@
...
+import static org.elasticsearch.common.xcontent.XContentHelper.mapToXContentParser;
+
+/**
+ * This {@link org.elasticsearch.immutablestate.ImmutableClusterStateHandler} is responsible for immutable state
+ * CRUD operations on ILM policies in, e.g. file based settings.
+ * <p>
+ * Internally it uses {@link TransportPutLifecycleAction} and
+ * {@link TransportDeleteLifecycleAction} to add, update and delete ILM policies.
+ */
+public class ImmutableLifecycleAction implements ImmutableClusterStateHandler<LifecyclePolicy> {
```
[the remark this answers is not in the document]
grcevski: Good idea!

server/src/main/java/org/elasticsearch/immutablestate/action/ImmutableClusterSettingsAction.java
```diff
@@ -0,0 +1,90 @@
...
+ * It is used by the ImmutableClusterStateController to update the persistent cluster settings.
+ * Since transient cluster settings are deprecated, this action doesn't support updating transient cluster settings.
+ */
+public class ImmutableClusterSettingsAction implements ImmutableClusterStateHandler<ClusterUpdateSettingsRequest> {
+
+    public static final String NAME = "cluster_settings";
+
+    private final ClusterSettings clusterSettings;
+
+    public ImmutableClusterSettingsAction(ClusterSettings clusterSettings) {
```
martijnvg, line 39: Should this also get wired up like was done here: https://github.com/elastic/elasticsearch/pull/86224/files#diff-b56bed42f5b0025886eb243ecb48678ac048bfba0ef047545dbb6504e357edf2R901 ?

grcevski: Yes, 100%. I left that part to happen in my follow-up draft PR I have now based on this one here: https://github.com/elastic/elasticsearch/pull/88224/files#diff-b56bed42f5b0025886eb243ecb48678ac048bfba0ef047545dbb6504e357edf2R905

After I wire them up, I call the initialize on the controller, which I haven't brought in yet from the initial WIP PR. Once we merge this in, that's the next one.

ChrisHegarty (approved):
LGTM.

martijnvg (approved):
ILM side LGTM 👍

grcevski:
Thanks @ChrisHegarty and @martijnvg !

PR #88102, opened 2022-06-28
Ingest: Enforce _version metadata not null in sourceAndMetadata map

stu-elastic:
The `_version` metadata field should always exist in the sourceAndMetadata
map, this change enforces that invariant but allows tests to avoid it
if necessary.

Refs: #87309


stu-elastic:
@elasticsearch run elasticsearch-ci/part-1

test/framework/src/main/java/org/elasticsearch/ingest/TestIngestDocument.java
```diff
@@ -20,14 +21,51 @@
...
+        Map<String, BiConsumer<String, Object>> validators = replaceValidator(VERSION, IngestSourceAndMetadata::longValidator);
+        Tuple<Map<String, Object>, Map<String, Object>> sm = IngestSourceAndMetadata.splitSourceAndMetadata(sourceAndMetadata);
+        IngestSourceAndMetadata withNullableVersion = new IngestSourceAndMetadata(sm.v1(), sm.v2(), null, validators);
+        return new IngestDocument(withNullableVersion, ingestMetadata);
+    }
+
+    /**
+     * Create an {@link IngestDocument} with {@link #DEFAULT_VERSION} as the _version metadata, if _version is not already present.
+     */
+    public static IngestDocument ofSourceAndMetadataWithDefaultVersion(Map<String, Object> sourceAndMetadata) {
```
rjernst, line 51: If this is the common case, can we shorten the method name to the original name? These are extremely long...

test/framework/src/main/java/org/elasticsearch/ingest/TestIngestDocument.java
```diff
@@ -39,8 +77,10 @@ public static IngestDocument ofMetadataWithValidator(Map<String, Object> metadat
     /**
      * Create an empty ingest document for testing
      */
-    public static IngestDocument emptyIngestDocument() {
-        return new IngestDocument(new HashMap<>(), new HashMap<>());
+    public static IngestDocument emptyIngestDocumentWithDefaultVersion() {
```
rjernst, line 80: I think it would be fine to leave the original name and just add to the javadoc that the default version is added.

rjernst (commented):
A couple naming suggestions

test/framework/src/main/java/org/elasticsearch/ingest/TestIngestDocument.java
stu-elastic: `ofSourceAndMetadataWithDefaultVersion` -> `withDefaultVersion`
`ofSourceAndMetadataWithNullableVersion` -> `withNullableVersion`
`ofSourceAndIngestWithNullableVersion` -> `ofIngestWithNullableVersion`
`emptyIngestDocumentWithDefaultVersion` -> `emptyIngestDocument`

test/framework/src/main/java/org/elasticsearch/ingest/TestIngestDocument.java
stu-elastic: Kept original, added javadoc comment.

rjernst (approved):
Much better names, thanks! LGTM

stu-elastic:
The discussions of the BWC issues happened in this thread: https://github.com/elastic/elasticsearch/pull/87309#discussion_r892487643, lead to this change.

PR #88113, opened 2022-06-28
Don't extend AbstractIndexComponent in AbstractTokenFilter

original-brownbear:
No need for this extension, we don't make use of the settings or deprecation logger
in production any more. Also, this slows down CS operations that require a
temporary index service which builds quite a bit slower when the loggers
need to be set up via reflective calls.

I left the existing tests around deprecations in place though they seem somewhat pointless now, should we just remove them?


romseygeek:
> I left the existing tests around deprecations in place though they seem somewhat pointless now, should we just remove them?

+1 to remove these as well

original-brownbear:
Alright, thanks @romseygeek dropped the tests that became redundant now.

romseygeek (approved):
LGTM, thanks @original-brownbear

original-brownbear:
np + thanks Alan!

PR #88114, opened 2022-06-28
[8.2][ML] Log node selection during inference routing (#88084)

dimitris-athanasiou:
This adds logging that prints the routing table and the selected
node while we are selecting a node to route an inference request to.

Backport of #88084


dimitris-athanasiou:
run elasticsearch-ci/rest-compatibility

Issue #88063, opened 2022-06-27
[CI] TransformContinuousIT testContinuousEvents failing


benwtrent:
**Build scan:**
https://gradle-enterprise.elastic.co/s/d2xk4wogo6gdg/tests/:x-pack:plugin:transform:qa:multi-node-tests:javaRestTest/org.elasticsearch.xpack.transform.integration.continuous.TransformContinuousIT/testContinuousEvents

**Reproduction line:**
`./gradlew ':x-pack:plugin:transform:qa:multi-node-tests:javaRestTest' --tests "org.elasticsearch.xpack.transform.integration.continuous.TransformContinuousIT.testContinuousEvents" -Dtests.seed=9C41A496F649E927 -Dtests.locale=lv-LV -Dtests.timezone=SystemV/YST9YDT -Druntime.java=17`

**Applicable branches:**
master

**Reproduces locally?:**
Didn't try

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.xpack.transform.integration.continuous.TransformContinuousIT&tests.test=testContinuousEvents

**Failure excerpt:**
```
java.lang.AssertionError: Error in test case [continuous-terms-on-date-pivot-test].If you want to mute the test, please mute [org.elasticsearch.xpack.transform.integration.continuous.TermsOnDateGroupByIT] only, but _not_ [org.elasticsearch.xpack.transform.integration.continuous.TransformContinuousIT] as a whole.

  at __randomizedtesting.SeedInfo.seed([9C41A496F649E927:64EDD250B2BEC433]:0)
  at org.elasticsearch.xpack.transform.integration.continuous.TransformContinuousIT.testContinuousEvents(TransformContinuousIT.java:277)
  at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(NativeMethodAccessorImpl.java:-2)
  at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
  at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.lang.reflect.Method.invoke(Method.java:568)
  at com.carrotsearch.randomizedtesting.RandomizedRunner.invoke(RandomizedRunner.java:1758)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$8.evaluate(RandomizedRunner.java:946)
[...]

Issue #85016, opened 2022-03-16
[Transform] Decouple scheduling from retrying


przemekwitek:
### Description

Currently, the retrying mechanism in transform is tightly tied to the scheduling, i.e. the retry happens every configured `frequency`.
We could decouple scheduling from retrying by introducing a wrapper around `SchedulerEngine` that would both schedule transform task on `frequency` as well as on a retry event.
This requires some refactoring.

przemekwitek:
I'm closing this issue as the main objective mentioned in the title ("Decouple scheduling from retrying") is achieved.
The follow-up work will be handled in separate issues.

PR #88116, opened 2022-06-28
[Transform] Increase transform scheduler frequency so that it runs every 1s

przemekwitek:
Relates https://github.com/elastic/elasticsearch/issues/85016
Relates https://github.com/elastic/elasticsearch/issues/88063


dimitris-athanasiou (approved):
LGTM

PR #88125, opened 2022-06-28
Don't extend AbstractIndexComponent in AbstractCharFilterFactory

original-brownbear:
Same as #88113 but for AbstractCharFilterFactory.


plugins/analysis-icu/src/test/java/org/elasticsearch/plugin/analysis/icu/IcuAnalyzerTests.java
```diff
@@ -25,12 +23,11 @@ public class IcuAnalyzerTests extends BaseTokenStreamTestCase {
     public void testMixedAlphabetTokenization() throws IOException {
 
         Settings settings = Settings.builder().put(IndexMetadata.SETTING_VERSION_CREATED, Version.CURRENT).build();
```
romseygeek, line 25: I think you can just use `Settings.EMPTY` for all of these now.

romseygeek (approved):
LGTM

original-brownbear:
Thanks Alan!

PR #88127, opened 2022-06-28
Stream input and output support for optional collections

n1v0lg:
This PR adds support for reading and writing optional collections via
`StreamInput` and `StreamOutput`.

A use case for this came up via https://github.com/elastic/elasticsearch/pull/87924#discussion_r907948009.


server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
```diff
@@ -24,6 +24,7 @@
 import org.elasticsearch.common.bytes.ReleasableBytesReference;
 import org.elasticsearch.common.collect.ImmutableOpenMap;
 import org.elasticsearch.common.geo.GeoPoint;
+import org.elasticsearch.common.io.stream.Writeable.Writer;
```
rjernst, line 27: nit: is this import strictly necessary? the existing qualified Writeable.Writer is more symmetric with Readable.Writer

server/src/test/java/org/elasticsearch/common/io/stream/AbstractStreamTests.java
```diff
@@ -256,12 +256,26 @@ public int hashCode() {
...
+            in -> in.readOptionalList(FooBar::new)
+        );
     }
 
     public void testStringCollection() throws IOException {
         runWriteReadCollectionTest(() -> randomUnicodeOfLength(16), StreamOutput::writeStringCollection, StreamInput::readStringList);
     }
 
+    public void testOptionalStringCollection() throws IOException {
+        runWriteReadCollectionTest(
```
rjernst, line 272: Instead of relying on randomization, can we have explicit tests for the relevant cases? The data for the collection can be randomized, but we should explicitly test null and not null.

rjernst (commented):
A couple thoughts

server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
n1v0lg: good catch, will revert 👍

ywangd (approved):
LGTM pending green CI

I'd also wait for @rjernst's approval. Thanks!

rjernst (approved):
LGTM

PR #62522, opened 2020-09-17
Hide c.a.s.s.i.UseArnRegionResolver noise

DaveCTurner:
A recent AWS SDK upgrade has introduced a new source of spurious `WARN`
logs when the security manager prevents access to the user's home
directory and therefore to `$HOME/.aws/config`. This is the behaviour we
want, and it's harmless and handled by the SDK as if the config doesn't
exist, so this log message is unnecessary noise.  This commit suppresses
this noisy logging by default.

Relates #20313, #56346, #53962
Closes #62493

PR #56346, opened 2020-05-07
Hide c.a.a.p.i.BasicProfileConfigFileLoader noise

DaveCTurner:
A recent AWS SDK upgrade has introduced a new source of spurious `WARN` logs
when the security manager prevents access to the user's home directory and
therefore to their shared client configuration. This is actually the behaviour
we want, and it's harmless and handled by the SDK as if the profile config
doesn't exist, so this log message is unnecessary noise. This commit suppresses
this noisy logging by default.

Relates #20313
Closes #56333

PR #20313, opened 2016-09-02
Configure AWS SDK logging configuration

jasontedor:
Because of security permissions that we do not grant to the AWS SDK (for
use in discovery-ec2 and repository-s3 plugins), certain calls in the
AWS SDK will lead to security exceptions that are logged at the warning
level. These warnings are noise and we should suppress them. This commit
adds plugin log configurations for discovery-ec2 and repository-s3 to
ship with default Log4j 2 configurations that suppress these log
warnings.

Closes #20294

PR #88133, opened 2022-06-28
Configure AWS SDK logging at runtime to hide ~/.aws/config WARN messages

tlrx:
We used to hide some noisy WARN log messages emitted by
the AWS SDK by setting them to ERROR level in log4j2.properties.
Since the repository plugins have been converted to modules, the
log4j2.properties files are not always updated during if users
have modified the log4j2.properties.

This pull request configures the logging level of some specific
AWS classes in order to hide the noisy messages even in the case
the log4j2.properties files have been modified.

Relates #56346, #62522, #20313


DaveCTurner (commented):
:+1: certainly works.

Could we have something in the relevant test suites to assert that we don't emit any warning logs from `com.amazonaws.*`? As well as testing this specific change, that'd make sure we proactively address any future bogus warnings too.

modules/repository-s3/src/main/java/org/elasticsearch/repositories/s3/S3RepositoryPlugin.java
```diff
@@ -54,9 +63,26 @@ public class S3RepositoryPlugin extends Plugin implements RepositoryPlugin, Relo
...
+                // some AWS classes emit noisy WARN logs so we hide them by setting logger to ERROR level
+                final Map<Logger, Level> awsConfigLoggers = List.of(
+                    AwsSdkMetrics.class,
+                    BasicProfileConfigFileLoader.class,
+                    SdkMBeanRegistrySupport.class,
+                    UseArnRegionResolver.class
+                ).stream().map(LogManager::getLogger).collect(Collectors.toUnmodifiableMap(logger -> logger, Logger::getLevel));
+                try {
+                    awsConfigLoggers.forEach((logger, level) -> {
+                        if (level.isLessSpecificThan(Level.ERROR)) {
```
DaveCTurner, line 76: Hmm actually I'm not sure about this bit. This means we can't override this behaviour at runtime even if we wanted to see these messages. Is that the intention? I guess we have to override `INFO` and `WARN` since these are the possible defaults, but maybe if the level is `DEBUG` or `TRACE` we shouldn't override it?

modules/repository-s3/src/main/java/org/elasticsearch/repositories/s3/S3RepositoryPlugin.java
```diff
@@ -54,9 +63,26 @@ public class S3RepositoryPlugin extends Plugin implements RepositoryPlugin, Relo
...
-                // TODO: fix that
-                Class.forName("com.amazonaws.ClientConfiguration");
+
+                // some AWS classes emit noisy WARN logs so we hide them by setting logger to ERROR level
+                final Map<Logger, Level> awsConfigLoggers = List.of(
+                    AwsSdkMetrics.class,
+                    BasicProfileConfigFileLoader.class,
+                    SdkMBeanRegistrySupport.class,
+                    UseArnRegionResolver.class
+                ).stream().map(LogManager::getLogger).collect(Collectors.toUnmodifiableMap(logger -> logger, Logger::getLevel));
```
DaveCTurner, line 73: Nit: `List.of(...).stream()` is just `Stream.of(...)`.

modules/repository-s3/src/main/java/org/elasticsearch/repositories/s3/S3RepositoryPlugin.java
tlrx: Looking at my change and your comment, I think I should address this differently. What I did here hides the noisy warning messages during `ClientConfiguration` class initialization only. This is still possible that `BasicProfileConfigFileLoader` et al emit more warning messages after that.

I pushed f193c2fabb64e4b94e0d2ee6e3e34fac7b346c74 to change a bit this logic: it now detects if the AWS loggers are explicitly configured in `log4j2.properties` and in case they aren't, their levels are set to `ERROR`. This way we don't override any existing file configuration (if a user really wants to see debug or warn messages) during the startup and until the cluster settings are applied (in which case a different logging level might override the ERROR level we configured). I also added tests.

Let me know what you think.

DaveCTurner (commented):
I think this is a reasonable approach too but it does mean that users have to set the level on these specific loggers, and not their ancestors. For instance, we explicitly set `com.amazonaws` to level `WARN` in the config file, but changing that isn't going to change these specific loggers.

I think that's ok for the sake of implementation simplicity but feel it deserves a mention in the docs to spell out exactly which classes are affected by this.

arteam (approved):
LGTM! I think this should prevent noisy AWS logging for users with a custom log4j configu

arteam:
@elasticmachine update branch

PR #88136, opened 2022-06-28
8.4 release notes ported from 8.3.0 release

craigtaverner:
Forward ported the release notes PRs from 8.3.0 release to the master branch and then regenerated release notes for current 8.4.0 status to check. See generated docs at https://elasticsearch_88136.docs-preview.app.elstc.co/diff


debadair:

debadair:
The conflicts need to be resolved and migrate_8_4.asciidoc is replicating the deprecation from 8.3, which is causing a build failure due to the duplicate ID.

debadair (approved):
LGTM

pugnascotia:
> @pugnascotia can you take a look here? It's possible our [instructions for forward porting release notes](https://github.com/elastic/elasticsearch-team/blob/master/onboarding-and-how-we-work/release/elasticsearch-release-manager.md#forward-port-the-release-documentation) might need some improvements. I don't see why we would be _adding_ changelog yaml files to `master` and @craigtaverner reported that running `pruneChangelogs` didn't delete the entries he expected.

The docs don't are 100% explicit, but the expectation is that only the asciidoc changes are forward-ported. I can't think of. a reason why we'd every forward-port changelog YAML files. I raised https://github.com/elastic/elasticsearch-team/pull/894.

As for the pruning, that'll be hard to diagnose without being able to replicate it. The logic is that we list all release tags before the release we just did for the current and prior major series, and for each tag find all changelog files at that tag. Then we find the intersection between those files and the files currently on disk. There would be a problem if the latest release tag hadn't appeared yet in the repo, and it's quite possible that it hasn't when the teams are told to execute the version bump. But why you'd end up with a partial prune, I don't know.

PR #88142, opened 2022-06-28
[ML] make more GET and heavier processing APIs cancellable

benwtrent:
This commit makes the following APIs cancellable:
 - get datafeeds
 - get jobs
 - evaluate data frames
 - explain data frames
 - get job model snapshot
 - preview datafeed
 - preview data frame
 - get categories (this can be expensive due to grok parsing)


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/action/ExplainDataFrameAnalyticsAction.java
```diff
@@ -30,6 +45,118 @@ private ExplainDataFrameAnalyticsAction() {
         super(NAME, ExplainDataFrameAnalyticsAction.Response::new);
     }
 
+    public static class Request extends AcknowledgedRequest<Request> implements ToXContentObject {
```
benwtrent, line 48: I added this request object (just a copy of `PutDataFrameAnalytics`) as the PUT object was being reused here, but I didn't want to make that request's task cancellable (as PUT isn't cancellable).

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportPreviewDataFrameAnalyticsAction.java
```diff
@@ -101,13 +102,14 @@ protected void doExecute(Task task, Request request, ActionListener<Response> li
     }
 
     void preview(Task task, DataFrameAnalyticsConfig config, ActionListener<Response> listener) {
+        final TaskId parentTaskId = new TaskId(clusterService.getNodeName(), task.getId());
```
benwtrent, line 105: I am not sure if `task.getParentTaskId()` was strictly correct. `task` is indeed the transport action task (created via our request object) that is being executed here. It SHOULD have a parent task that is the original HTTP request (I think?) but it seems that the prevailing pattern is to rely on this `task` object instead of its parent.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportPreviewDatafeedAction.java
```diff
@@ -131,7 +135,7 @@ private void previewDatafeed(
             // requesting the preview doesn't have permission to search the relevant indices.
             DatafeedConfig previewDatafeedConfig = previewDatafeedBuilder.build();
             DataExtractorFactory.create(
-                new ParentTaskAssigningClient(client, clusterService.localNode(), task),
+                new ParentTaskAssigningClient(client, parentTaskId),
```
benwtrent, line 138: This was a nice idea, but since the task being created by the request was never cancellable (and neither was the http request), the cascade effect of cancellations was missed :/

benwtrent:
@elasticmachine update branch

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/action/ExplainDataFrameAnalyticsAction.java
benwtrent: Also, I added a BWC test for `explain` in a mixed cluster to ensure this doesn't break anything. Ran it many many times and its all good since the explain request and the put request can be serialized to each other without issue.

benwtrent:
@elasticmachine update branch

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportGetModelSnapshotsAction.java
```diff
@@ -79,6 +93,7 @@ private void getModelSnapshots(GetModelSnapshotsAction.Request request, ActionLi
             request.getSort(),
             request.getDescOrder(),
             request.getSnapshotId(),
+            parenTaskId,
```
droberts195, line 96:
```suggestion
            parentTaskId,
```

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportGetModelSnapshotsAction.java
```diff
@@ -63,13 +69,21 @@ protected void doExecute(
...
+            request.getJobId(),
+            parentTaskId,
+            ActionListener.wrap(ok -> getModelSnapshots(request, parentTaskId, listener), listener::onFailure)
+        );
     }
 
-    private void getModelSnapshots(GetModelSnapshotsAction.Request request, ActionListener<GetModelSnapshotsAction.Response> listener) {
+    private void getModelSnapshots(
+        GetModelSnapshotsAction.Request request,
+        TaskId parenTaskId,
```
droberts195, line 84:
```suggestion
        TaskId parentTaskId,
```

droberts195 (approved):
LGTM

benwtrent:
@elasticmachine update branch

PR #88145, opened 2022-06-29
Remove ingest attachment limitation for fips

rjernst:
The ingest attachment plugin was unsupported in FIPS mode. This was
because of bouncycastle being included by some tika processors. The
bouncycastle dependency of those processors has now been removed, so
this commit removes the limitation from the FIPS docs.

relates #88031


tvernum (approved):
LGTM

PR #88148, opened 2022-06-29
Run gradle integration tests with configuration cache enabled by default

breskeby:
This changes AbstractGradleFuncTest to run with configuration cache enabled by default. That gives us better test coverage on how our build logic works with configuration cache enabled and makes it explicit which parts are not yet working. 
We fixed some minor configuration cache related issues in our code


breskeby:
This changes AbstractGradleFuncTest to run with configuration cache enabled by default. That gives us better test coverage on how our build logic works with configuration cache enabled and makes it explicit which parts are not yet working.
We fixed some minor configuration cache related issues in our code.

PR #81237, opened 2021-12-02
More avoidance for loadAuthorizedIndices

ywangd:
This PR replaces the concrete list of authorizeIndices with a lazily load class
so that loading is not triggered if not necessary. 

For example, a search targeting a concret name, `GET index/_search` no longer
triggers the loading. But names with wildcard will still trigger the loading.

However if we organize the indices with data streams and alias, it is possible
to achieve similar wildcard effect while still avoid the loading. That is,
searches like `GET alias/_search` or `GET data_stream/_search` will *not*
trigger the loading even they target multiple indices behind the single name.

PR #88149, opened 2022-06-29
Avoid loadAuthorizedIndices for requests with concrete names (#81237)

ywangd:
This PR replaces the concrete list of authorizeIndices with a lazily load class
so that loading is not triggered if not necessary.

For example, a search targeting a concret name, GET index/_search no longer
triggers the loading. But names with wildcard will still trigger the loading.

However if we organize the indices with data streams and alias, it is possible
to achieve similar wildcard effect while still avoid the loading. That is,
searches like GET alias/_search or GET data_stream/_search will not
trigger the loading even they target multiple indices behind the single name.

Backport: #81237


ywangd:
@elasticmachine update branch

PR #88154, opened 2022-06-29
Cleanup needless anonymous classes where lamdbas suffice

original-brownbear:
Cleaning up a couple more spots of these, saving some verbosity
and in some cases also some allocations.


original-brownbear:
@elasticmachine update branch

grcevski (approved):
LGTM! This looks good to me, the only issue I see is that one method with poor test coverage to prove this works through the CI being green. I think it would be very easy to add a test in NetworkUtilsTests or skip this change here, especially since it's in a deprecated method.

original-brownbear:
Thanks Nikola!

> skip this change here

jup went with this :) no need for noise in deprecated  code IMO :)

PR #88155, opened 2022-06-29
Use the provided SAS token without SDK sanitation that can produce invalid signatures

### LINKED ISSUE #88140: Certain Azure SAS tokens are sanitized incorrectly by the SDK
After introducing https://github.com/elastic/elasticsearch/pull/86135 the underlying Azure SDK sanitizes incorrectly certain SAS tokens, preventing the correct authentication against the Azure storage service.

fcofdez:
When the SAS token is prefixed with ? the SDK parses incorrectly
all the query params and removes the first query param, this
produces an invalid signature preventing the repository to
authenticate against the storage service. This commit reverts
to the previous behaviour where the SAS token was appended to
the URL as it is provided by the user.

Closes #88140


fcofdez:
The CI failures seems to be infrastructure related, this is ready for review.

modules/repository-azure/src/test/java/org/elasticsearch/repositories/azure/AzureSasTokenTests.java
```diff
@@ -0,0 +1,81 @@
...
+        if (randomBoolean()) {
+            sasToken = "?sv=2018-11-09&spr=https&st=2021-09-21T13%3A00%3A07Z&se=2071-09-21T13%3A00%3A07Z"
+                + "&sr=c&sp=racwdl&sig=4%2Fak3ibKW%2FXILJI%2B8mInVaLiDw8n3Es%2FQbTSiG3LOt0%3D";
+        } else {
+            sasToken = "sv=2018-11-09&spr=https&st=2021-09-21T13%3A00%3A07Z&se=2071-09-21T13%3A00%3A07Z&"
+                + "sr=c&sp=racwdl&sig=4%2Fak3ibKW%2FXILJI%2B8mInVaLiDw8n3Es%2FQbTSiG3LOt0%3D";
+        }
+        secureSettings.setString(SAS_TOKEN_SETTING.getConcreteSettingForNamespace(clientName).getKey(), sasToken);
+
+        httpServer.createContext("/account/container/sas_test", exchange -> {
```
arteam, line 47: I believe `exchange` needs to be closed after we finished writing the response.

modules/repository-azure/src/test/java/org/elasticsearch/repositories/azure/AzureSasTokenTests.java
```diff
@@ -0,0 +1,81 @@
...
+            final var queryParams = exchange.getRequestURI().getRawQuery();
+            if (sasToken.startsWith("?")) {
+                assertThat(queryParams, is(equalTo(sasToken.substring(1))));
+            } else {
+                assertThat(queryParams, is(equalTo(sasToken)));
+            }
+
+            Streams.readFully(exchange.getRequestBody());
```
arteam, line 55: Do we need to read the body if only care about the request method?

modules/repository-azure/src/test/java/org/elasticsearch/repositories/azure/AzureSasTokenTests.java
```diff
@@ -0,0 +1,81 @@
...
+public class AzureSasTokenTests extends AbstractAzureServerTestCase {
+    public void testSasTokenIsUsedAsProvidedInSettings() throws Exception {
+        final int maxRetries = randomIntBetween(1, 5);
+        final byte[] bytes = randomBlobContent();
+
+        final String clientName = randomAlphaOfLength(5).toLowerCase(Locale.ROOT);
+        final MockSecureSettings secureSettings = new MockSecureSettings();
+        secureSettings.setString(ACCOUNT_SETTING.getConcreteSettingForNamespace(clientName).getKey(), "account");
+        final String sasToken;
+        if (randomBoolean()) {
```
arteam, line 38: Consider avoiding duplication with
```java
String prefix = randomBoolean() ? "?" : "";
        String sasToken = prefix
            + "sv=2018-11-09&spr=https&st=2021-09-21T13%3A00%3A07Z&se=2071-09-21T13%3A00%3A07Z&"
            + "sr=c&sp=racwdl&sig=4%2Fak3ibKW%2FXILJI%2B8mInVaLiDw8n3Es%2FQbTSiG3LOt0%3D";
```

modules/repository-azure/src/main/java/org/elasticsearch/repositories/azure/AzureClientProvider.java
```diff
@@ -180,18 +183,30 @@ AzureBlobServiceClient createClient(
         final HttpClient httpClient = new NettyAsyncHttpClientBuilder(nettyHttpClient).disableBufferCopy(true).proxy(proxyOptions).build();
 
         final String connectionString = settings.getConnectString();
+        final StorageConnectionString storageConnectionString = StorageConnectionString.create(connectionString, clientLogger);
+        final StorageEndpoint endpoint = storageConnectionString.getBlobEndpoint();
```
arteam, line 187: I believe there an issue with that because the SDK doesn't export these classes in `module-info`, but I'm not sure whether we enforce it.

arteam (commented):
LGTM! I've left a couple of comments

henningandersen (approved):
LGTM (once CI passes, perhaps provoke new runs immediately if caused by infra).

modules/repository-azure/src/test/java/org/elasticsearch/repositories/azure/AzureSasTokenTests.java
fcofdez: Yes, to ensure that we read the entire request.

modules/repository-azure/src/main/java/org/elasticsearch/repositories/azure/AzureClientProvider.java
fcofdez: We were using the `StorageConnectionString` class already. I agree that we should move to our own connection string parsing, but we can defer that.

arteam (approved):
LGTM! Thanks Francisco!

modules/repository-azure/src/main/java/org/elasticsearch/repositories/azure/AzureClientProvider.java
```diff
@@ -33,7 +33,10 @@
 import com.azure.storage.blob.BlobServiceAsyncClient;
 import com.azure.storage.blob.BlobServiceClient;
 import com.azure.storage.blob.BlobServiceClientBuilder;
+import com.azure.storage.common.StorageSharedKeyCredential;
+import com.azure.storage.common.implementation.connectionstring.StorageAuthenticationSettings;
 import com.azure.storage.common.implementation.connectionstring.StorageConnectionString;
+import com.azure.storage.common.implementation.connectionstring.StorageEndpoint;
```
kasobol-msft, line 39: FYI.
The fix for your problem will go out with next storage release.
See https://github.com/Azure/azure-sdk-for-java/issues/29998
https://github.com/Azure/azure-sdk-for-java/pull/30003

When the fix is out I suggest to not use types from `*.implementation.*` - we don't export them in our modules - which might be a problem on latest java runtime versions.

If there's anything else we can do to help here, please let us know.

PR #88110, opened 2022-06-28
Avoid WindowsFS in LongGCDisruption

DaveCTurner:
Closes #87914

PR #88160, opened 2022-06-29
[8.3] Avoid WindowsFS in LongGCDisruption (#88110)

DaveCTurner:
 - Avoid WindowsFS in LongGCDisruption (#88110)


DaveCTurner:
@elasticmachine update branch

PR #88165, opened 2022-06-29
repair downloading JDK before running test cases

dormanze:
If the JDK version on the environment is consistent with the JDK version required for running, you do not need to download the JDK.


breskeby:
@dormanze What use case do you try to solve here. What problem do you face with the code. From my point of view that change doesn't make sense, sorry.

dormanze:
> @dormanze What use case do you try to solve here. What problem do you face with the code. From my point of view that change doesn't make sense, sorry.
Thank you for your reply , I run "gradle :server:test --tests 'org.elasticsearch.*'" and my jdk version is "openjdk 18.0.1",but also need to download JDK18 when initializing the running environment.I don't think I need to download the JDK if my local JDK version is consistent with the version needed to run the test cases

breskeby:
If you are using openjdk 18.0.1 and have that also set as your JAVA_HOME this should work. the flag you are trying to change here is for indicating if explicitly another runtime has been configured for the build by usually passing `RUNTIME_JAVA_HOME` environment variable or a `runtime.java` system property. Have you set one of those? if so try removing it. It seems its pointing to an unexpected JDK for you from what I understand. If those properties are not set the build should pick the java runtime the gradle process is using for forked test processes. 
I'm closing this for now as it works as expected for us.

If you need help diagnosing your local problem in more detail, please provide a step by step reproduction description of your scenario including defined env variables etc.

PR #88173, opened 2022-06-29
Simplify QueryCache Construction

original-brownbear:
We don't need to keep references to index settings in all of these,
nor do we need per-index loggers or deprecation loggers here.
Making the construction here a little cheaper helps with the
performance of some cluster state operations that create temporary
index-services.
Also, removing references to `IndexSettings` and other per-index state
is helpful in identifying spots to reduce ther per-index overhead on
data-nodes down the road.

relates #77466


romseygeek (approved):
LGTM

original-brownbear:
@elasticmachine update branch

original-brownbear:
@elasticmachine update branch

original-brownbear:
Thanks Alan!

Issue #81406, opened 2021-12-07
Revisit default disk watermark on different data tiers


Leaf-Lin:
Today, different data tiers all have the same `cluster.routing.allocation.disk.watermark.low|high|flood_stage` settings.  These settings by default are relative values means that for a larger disk, the default value can introduce some level of wastage.

Do we want to revisit the default value for disk watermark especially for different data tiers, there might be a different level of space requirement?

dakrone:
One thing that would make this difficult is that these are dynamic settings, and we currently don't have the Settings functionality to be able to specify setting a particular value on a subset of nodes. This would mean that we could have separate defaults, but if the user ever dynamically changed one of the settings, it would change for all tiers.

I'm adding the core/infra label for this, since they'd need to help us with the Settings infrastructure if we wanted to support something like this fully.

Leaf-Lin:
There was a discussion that this would be an operator-only setting on the cloud once the default settings are sensible.

dakrone:
> There was a discussion that this would be an operator-only setting on the cloud once the default settings are sensible.

That helps a bit. I think it'd still be good to think about where we want to go with tier-specific settings in the future, both for the purpose of on-prem users, but also for other settings (that may be non-operator-only) that we want to set to different values (for example, we already do this with a recovery setting: https://github.com/elastic/elasticsearch/pull/68480)

DaveCTurner:
I wonder if we really need more control here, as opposed to changing how Elasticsearch computes its disk watermarks to better utilise larger nodes. Today ES (roughly) aims to keep 10% of disk free on each node, but if we say changed this to target `min(10%,100GB)` of free space then larger nodes would be better-utilised automatically without needing to adjust any settings.

henningandersen:
I agree with David here, doing similarly to what we did for frozen tier on all tiers would in my mind effectively solve this problem.

Making it tier specific is not enough/ideal during a rolling up-size anyway.

dakrone:
Supporting a rudimentary syntax of *either*: `0.75` or `10%` or `100gb` or `min(10%,100gb)` or `max(10%,100gb)` (and then changing the default for our needs) seems like a reasonable thing to do for this to me also.

kingherc:
From conversing with @henningandersen , we could use 20GB headroom for flood_stage, 100GB for high, and 150GB for low watermarks.

kingherc:
[...]

PR #88175, opened 2022-06-29
Add disk thresholds in the cluster state

gmarouli:
**Problem statement**
For a data node, we use the watermarks to determine if a node's disk usage is healthy. The watermarks can be configured in different ways and it's possible that each node has a different watermark configuration. This is not desirable, we want to use the same thresholds for all nodes and specifically the ones that the master is using.

**Proposal**
When a node is a elected as master, it will add a custom metadata to the cluster state that will describe these thresholds. For example:
```
    "health": {
      "disk": {
        "low_watermark": "85%",
        "high_watermark": "90%",
        "flood_stage_watermark": "95%",
        "frozen_flood_stage_watermark": "95%",
        "frozen_flood_stage_max_headroom": "20gb"
      }
    }
```

In this PR, we introduce the health metadata and we wire the existing disk thresholds to update the health metadata in the cluster state.

Part of https://github.com/elastic/elasticsearch/issues/84811


gmarouli:
@dakrone referring to https://github.com/elastic/elasticsearch/pull/87975#pullrequestreview-1022382259

I think there is a misunderstanding. The health metadata are not supposed to be node specific, we intend them to be the same for all the nodes of the cluster and determined by the master node. 

Effectively, that's what is happening right when ti comes to allocation too. All nodes have disk threshold settings (potentially different ones), but the elected master node will use their own in the allocation decider. In a similar way, we want the master node to propagate the thresholds so every node can check their disk usage and report back using the same thresholds.

Is this more clear now?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
```diff
@@ -8,26 +8,33 @@
...
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 /**
  * A container to keep settings for disk thresholds up to date with cluster setting changes.
  */
 public class DiskThresholdSettings {
```
gmarouli, line 35: Keep in mind that I am considering in the next part of this PR to move this class to `org.elasticsearch.common.settings`. 

Then we can have one place with all the disk threshold settings, their validation and how to parse them. I haven't done it yet because it's going to touch a lot of files and it might get confusing. I thought it would be easier to review this, with that in mind, and do the refactoring in a single commit after we verify that the PR so far is ok.

If you prefer a different approach let me know.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
```diff
@@ -278,38 +288,43 @@ private void setEnabled(boolean enabled) {
...
         // Watermark is expressed in terms of used data, but we need "free" data watermark
         this.lowWatermarkRaw = lowWatermark;
-        this.freeDiskThresholdLow = 100.0 - thresholdPercentageFromWatermark(lowWatermark);
-        this.freeBytesThresholdLow = thresholdBytesFromWatermark(
+        this.freeDiskThresholdLow = 100.0 - DiskThresholdSettingParser.parseThresholdPercentage(lowWatermark);
+        this.freeBytesThresholdLow = DiskThresholdSettingParser.parseThresholdBytes(
             lowWatermark,
             CLUSTER_ROUTING_ALLOCATION_LOW_DISK_WATERMARK_SETTING.getKey()
         );
+        notifyListeners();
```
andreidan, line 296: I don't think we need another observer mechanism in place as these methods are already listeners to settings update https://github.com/elastic/elasticsearch/pull/88175/files#diff-697233b32fbd24830c548746773a3013fc03a1cacc6ca877c8d75f0b9f68f751L121

Could the HealthMetadataService be a listener on the settings and consume the "settings changed" notifications that way?

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
```diff
@@ -0,0 +1,152 @@
...
+            @Override
+            public ClusterState execute(ClusterState currentState, List<TaskContext<UpdateHealthMetadataTask>> taskContexts)
+                throws Exception {
+                final HealthMetadata initialHealthMetadata = HealthMetadata.getHealthCustomMetadata(currentState);
+                HealthMetadata currentHealthMetadata = initialHealthMetadata;
+                for (TaskContext<UpdateHealthMetadataTask> taskContext : taskContexts) {
+                    currentHealthMetadata = taskContext.getTask().healthMetadata();
+                    taskContext.success(() -> {});
+                }
+                final var finalHealthMetadata = currentHealthMetadata;
```
andreidan, line 145: is the `finalHealthMetadata` needed still?

Would checking `return currentHealthMetadata == initialHealthMetadata` below be enough?

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    private void submitHealthMetadata(String source) {
+        HealthMetadata localHealthMedata = createHealthMetadata();
+        var task = new UpdateHealthMetadataTask(localHealthMedata);
+        var config = ClusterStateTaskConfig.build(Priority.NORMAL);
+        clusterService.submitStateUpdateTask(source, task, config, taskExecutor);
+    }
+
+    record UpdateHealthMetadataTask(HealthMetadata healthMetadata) implements ClusterStateTaskListener {
```
andreidan, line 122: I'm a bit confused by this record.

It's a task but doesn't have an execute method ? Should the `createHealthMetadata` be a sort of ... `upsertHealthMedata` and be executed inside the task's execute method?

server/src/internalClusterTest/java/org/elasticsearch/health/HealthMetadataServiceIT.java
```diff
@@ -0,0 +1,129 @@
...
+                    ? randomIntBetween(60, 80) + "%"
+                    : new ByteSizeValue(randomIntBetween(10, 100)).toString();
+                String nodeName = startNode(internalCluster, customWatermark);
+                watermarkByNode.put(nodeName, customWatermark);
+            }
+            ensureStableCluster(numberOfNodes);
+
+            String electedMaster = internalCluster.getMasterName();
+            {
+                HealthMetadata.DiskThresholds.DiskThreshold lowWatermark = HealthMetadata.getHealthCustomMetadata(
```
andreidan, line 53: naming game :) 

This is a bit difficult to parse (at least to me) - `HealthMetadata.DiskThresholds.DiskThreshold`

Since this is in the health metadata could (should?) we come up with a new concept? Alert limits? Health limits? Disk health limits?

andreidan (commented):
Thanks for working on this Mary.

I've left a few suggestions and questions

server/src/internalClusterTest/java/org/elasticsearch/health/HealthMetadataServiceIT.java
gmarouli: I see, I have to admit that threshold sounds better than limits but I will play around with naming see if we can come up with something better.

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
gmarouli: I see. What about this:
In the health metadata class we have classes that monitor the metadata settings. The task takes this monitors and in the execute method creates the current version of the HealthMetadata. 

I implemented this idea, in my latest commit. What do you think?

server/src/internalClusterTest/java/org/elasticsearch/health/HealthMetadataServiceIT.java
gmarouli: What about `HealthMetadata` contains `DiskMetadata` (for now it's mainly thresholds, but you never know what else we might need) and then we have the `DiskThreshold` I could also go for `DiskHealthLimits` but I am not sure if it is easier to read or not. 

About `HealthLimits` or `AlertLimits` I think it's too generic and we will have issues when we introduce other resource metadata.

Another idea I had, and I am leaning towards this tbh, is to use the `RelativeByteSizeValue` instead of `DiskThreshold`. What do you think?

I am open to other ideas :)

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
gmarouli: It's needed because of the lambda, the `currentHealthMetadata` is not final.

server/src/internalClusterTest/java/org/elasticsearch/health/HealthMetadataServiceIT.java
andreidan: How about `HealthMetadata.Disk.Threshold` ? (the FQN heps removing the extra "metadata" and "disk" names)

> Another idea I had, and I am leaning towards this tbh, is to use the RelativeByteSizeValue instead of DiskThreshold. What do you think?

Are you thinking of naming here? Or the actual `RelativeByteSizeValue` entity instead of the DiskThreshold entity?

gmarouli: I was thinking to use the actual `RelativeByteSizeValue` java entity since we are serializing it via string conversion. 

> How about HealthMetadata.Disk.Threshold

I do like this! I wanted to avoid generic names because it might confuse people when searching in the IDE. But since it's an inner class and it looks quite nice... I am reconsidering :) Shall we go for this?

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
andreidan: Thanks for iterating on this.

Do we need this extra concept of "monitoring" though? Do we need the caching it provides for a particular use case?

Could the `HealthMetadataService` be the settings listener and trigger the necessary cluster state updates when they change? (apologies if I'm missing something)

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
andreidan: Ah you're right, we can't make it effectively final due to the loop :sadpanda: 
Sorry for the noise

server/src/internalClusterTest/java/org/elasticsearch/health/HealthMetadataServiceIT.java
andreidan: I'm a fan :) If you like it too I say we go for this (@dakrone might have a different opinion but we can incorporate those too if they come through :) )

gmarouli: I want for `HealthMetadata.Disk.Threshold` :)

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
gmarouli: I do not think you are missing something. I was trying to optimize the case when we initialize the health metadata for the first time or when a master node gets elected. 

We could also parse the settings on the spot and have no cache, I was trying to avoid parsing them multiple times but on the other hand if this doesn't happen often maybe it's less important.

Do you think it would be cleaner to only track the settings?

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
```diff
@@ -0,0 +1,275 @@
...
+        }
+    }
+
+    private void updateHealthMetadataIfNecessary(ClusterChangedEvent event) {
+        // Wait until every node in the cluster is upgraded to 8.4.0 or later
+        if (event.state().nodesIfRecovered().getMinNodeVersion().onOrAfter(Version.V_8_4_0)) {
+            if (event.localNodeMaster() && publishedAfterElection == false) {
+                submitHealthMetadata("health-metadata-update-master-election");
+            }
+            // If the node is not the elected master anymore
```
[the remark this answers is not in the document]
gmarouli: Yes, functionally, that's exactly what I was going for. I think I cannot reuse the same code because we wait to check for the version of the cluster `event.state().nodesIfRecovered().getMinNodeVersion().onOrAfter(Version.V_8_4_0)`. If I am not mistaken this happens after the master node is elected so we do not publish the health metadata during the start up phase of a cluster. 

Do you have any suggestions to capture this scenario?

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
gmarouli: Indeed, I could get the last one like we did before and make it final, but I like that this is more inline with code in other similar cases.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
gmarouli: Fixed

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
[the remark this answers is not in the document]
gmarouli: I will try to use `LocalNodeMasterListener`. Maybe it will produce some cleaner code.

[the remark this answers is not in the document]
gmarouli: No, that wouldn't work because still need to check the version `event.state().nodesIfRecovered().getMinNodeVersion().onOrAfter(Version.V_8_4_0)`.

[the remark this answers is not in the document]
gmarouli: I am working on something, I will ping you back when it's ready :)

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
gmarouli: Settled offline, we expect that the changes will not be in high frequency so parsing the settings will not be a big burden and it remove the need of a cache.

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    public static List<NamedWriteableRegistry.Entry> getNamedWriteables() {
+        return List.of(
+            new NamedWriteableRegistry.Entry(Metadata.Custom.class, HealthMetadata.TYPE, HealthMetadata::new),
+            new NamedWriteableRegistry.Entry(NamedDiff.class, HealthMetadata.TYPE, HealthMetadata::readDiffFrom)
+        );
+    }
+
+    static class UpsertHealthMetadataTask implements ClusterStateTaskListener {
```
andreidan, line 156: Shall we document this?

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
```diff
@@ -0,0 +1,241 @@
...
+
+        private final HealthMetadata healthMetadata;
+
+        private UpsertHealthMetadataTask(HealthMetadata healthMetadata) {
+            this.healthMetadata = healthMetadata;
+        }
+
+        static UpsertHealthMetadataTask createInsertTask(Settings settings) {
```
andreidan, line 164: Shall we document this?

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
```diff
@@ -0,0 +1,241 @@
...
+                            CLUSTER_ROUTING_ALLOCATION_DISK_FLOOD_STAGE_WATERMARK_SETTING.getKey()
+                        ),
+                        new HealthMetadata.Disk.Threshold(CLUSTER_ROUTING_ALLOCATION_DISK_FLOOD_STAGE_FROZEN_SETTING.get(settings)),
+                        CLUSTER_ROUTING_ALLOCATION_DISK_FLOOD_STAGE_FROZEN_MAX_HEADROOM_SETTING.get(settings)
+                    )
+                )
+            );
+        }
+
+        static UpsertHealthMetadataTask createUpdateTask(ClusterState clusterState, String setting, String value) {
```
andreidan, line 187: Shall we document this?

andreidan (approved):
LGTM, thanks for iterating on this Mary

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
gmarouli: While documenting this, I noticed that this had a bug. The code was applying the changes on the cluster state it from the caller. Since we are working with batched updates, the last update applied would override all the previous ones. 

Fixed and now the health metadata is constructed in the execute method which makes much more sense.

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadata.java
```diff
@@ -0,0 +1,329 @@
...
+        diskMetadata.toXContent(builder, params);
+        builder.endObject();
+        return builder;
+    }
+
+    public static HealthMetadata fromXContent(XContentParser parser) throws IOException {
+        return PARSER.parse(parser, null);
+    }
+
+    public static HealthMetadata getHealthCustomMetadata(ClusterState clusterState) {
```
andreidan, line 93: This can be nullable. Shall we mark this method with `@Nullable` ?

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
```diff
@@ -0,0 +1,260 @@
...
+        public void clusterStateProcessed(ClusterState oldState, ClusterState newState) {
+            assert false : "never called";
+        }
+
+        @Override
+        public void onFailure(@Nullable Exception e) {
+            logger.error("failure during health metadata update", e);
+        }
+
+        abstract HealthMetadata execute(HealthMetadata initialHealthMetadata);
```
andreidan, line 170: Should the param be marked with @Nullable?

andreidan: Or maybe avoid the ignored parameter in the `Insert...` implementation by having the `ClusterState currentState` as parameter? (each implementation can then look for the `health` custom and add it/modify it)

Maybe even return `ClusterState` ?

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
andreidan: ++ good catch 

I've just added one more suggestion to have the `execute` method be `ClusterState -> ClusterState` (this would avoid an ignored parameter and having it @Nullable)

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/metadata/HealthMetadataService.java
gmarouli: Ah, even better!!! Thank you for the suggestion!

server/src/internalClusterTest/java/org/elasticsearch/health/HealthMetadataServiceIT.java
gmarouli: @andreidan So, I am reconsidering this choice since distributed team will be working soon(-ish) on #81406. I am proposing to already use `RelativeByteSizeValue` instead of the `Threshold`. The reason is that it has the `calculateValue` method which calculates the percentage and takes into account the `headroom`, so for the frozen tier it will take the headroom into account and for the others we will use `-1` for now until the headroom is implemented. This will make easier the transition when headroom  is implemented everywhere. Any objections?

gmarouli:
@elasticmachine update branch

gmarouli:
_Dotting the i's and crossing the t's_
- Removed the `low watermark` from the `HealthMetadata`, we do not use it in the disk health calculation so there is no need to store it there.
- Reverted the `DiskThresholdSettingParser` in course of this PR it became unnecessary (at least for now) to decouple the settings and their parsing.

server/src/internalClusterTest/java/org/elasticsearch/health/HealthMetadataServiceIT.java
gmarouli: Proposal applied

server/src/main/java/org/elasticsearch/common/unit/RelativeByteSizeValue.java
```diff
@@ -95,4 +97,17 @@ public String getStringRep() {
...
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        RelativeByteSizeValue that = (RelativeByteSizeValue) o;
+        return Objects.equals(absolute, that.absolute) && Objects.equals(ratio, that.ratio);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(absolute, ratio);
+    }
```
andreidan, line 112: I don't think this is needed ?

gmarouli: Correct, I just saw it. Removing it.

gmarouli:
@elasticmachine update branch

PR #88176, opened 2022-06-29
[8.3] Use the provided SAS token without SDK sanitation that can produce invalid signatures (#88155)

fcofdez:
 - Use the provided SAS token without SDK sanitation that can produce invalid signatures (#88155)


fcofdez:
@elasticmachine update branch

PR #88178, opened 2022-06-29
Seq no shrink maybe

nik9000:
This shrinks the points for the `_seq_no` field from `562mb` to `449mb`,
that's an `20%` reduction in size. The whole index is about `7gb` so this
clocks in around `1.6%` of the entire index.

The price of this size reduction is a slow down in queries for `_seq_no`.
Specifically, we have to double check some documents against doc values.
More specifically, a `range` query has to double check at most 2048
doc values, though commonly will only have to double check 1024. This is
important because `range` queries are used for recovery and any "change
capture" style use case like transform and fleet.

It's also possible to perform `term`, `match`, and `terms` queries against
`_seq_no`. `term` and `match` queries will have to double check at most
1024 doc values. `terms` will have to double check at most 1024 docs per
term - though any terms that are close to each other will be able to
"share" the double checking.


nik9000:
run 
elasticsearch-ci/part-1

nik9000:
run elasticsearch-ci/part-1

nik9000:
The bwc test failures were real here - I was treating indices created before 8.4 as though they were also shifting their seq_nos

nik9000:
I had expected the bwc tests to fail with an actual test failure, but the failure was that ES wouldn't start.

PR #88181, opened 2022-06-29
Log ingestion API

### LINKED ISSUE #88777: Better default mappings for logs
There are several pitfalls when using the default mappings for logs (the [`logs-*-*` index template](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/src/main/resources/logs-template.json)):

- Data loss
  - Mapping issues due to object vs scalar conflicts (`"host": "foo", "host.name": "foo"`)
  - Mapping issues due to conflicting types (`"foo": 42, "foo": "bar"`)
  - Mapping explosions
- Ingestion and disk overhead because all fields are indexed by default, even for fields that are never or rarely searched by (for example `process.argv`)

felixbarny:
* Indexes to data stream based on dataset and namespace
  * Sets timestamp to now if now provided
  * Indexes fields that are important for correlation, filtering, and slice/dice
* Indexes to fallback data stream
  * If JSON parsing of the event fails
  * If indexing to requested data stream has failed 
  * Fallback data stream has lenient mappings and can accept any document

[Docs preview](https://elasticsearch_88181.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/logs-api.html)


docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,153 @@
...
+The dataset and namespace parameters determine the <<data-streams, data stream>> in which the events will be indexed `logs-<dataset>-<namespace>`. This can be overridden on a per-event basis.
+
+[discrete]
+[[logs-api-query-params]]
+=== {api-query-parms-title}
+
+Any provided query parameter will be added to each log line.
+For example, `/_logs/myapp?service.name=myapp` will add `"data_stream.dataset": "myapp"` and `"service.name": "myapp"` to all logs.
+
+
```
felixbarny, line 62: [discrete]

server/src/main/java/org/elasticsearch/rest/action/logs/RestLogsAction.java
```diff
@@ -0,0 +1,257 @@
...
+                        return;
+                    }
+                    BulkRequest retryBulk = Requests.bulkRequest();
+                    Arrays.stream(bulkItemResponses.getItems()).filter(BulkItemResponse::isFailed).forEach(failedRequest -> {
+                        IndexRequest originalRequest = indexRequests.get(failedRequest.getItemId());
+                        Map<String, Object> doc = originalRequest.sourceAsMap();
+                        BulkItemResponse.Failure failure = failedRequest.getFailure();
+                        if (failure.getStatus() == RestStatus.TOO_MANY_REQUESTS) {
+                            // looks like a transient error; re-try as-is
+                            retryBulk.add(Requests.indexRequest(originalRequest.index()).opType(DocWriteRequest.OpType.CREATE).source(doc));
```
felixbarny, line 128: There's a risk that re-trying in the target index leads to a persistent error.
I'm inclined to keep the re-try logic simple for now and just always send documents to the DLQ, no matter the reason for ingestion failure. Once we introduce something like a buffer/queue, we could do more clever re-trying on back pressure, such as exponentially backing off the re-tries.

@jpountz, wdyt?

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+If an error happens during ingestion,
+the logs are sent to the `logs-dlq-<namespace>` data stream that acts as a dead letter queue for failed events.
+However, log ingestion should rarely fail as the mappings in the built-in index template for the `logs-*-*` data streams are designed to minimize mapping conflicts.
+
+A <<logs-api-request-body, couple of fields>> from the {ecs-ref}[Elastic Common Schema (ECS)] are indexed by default that are commonly used to search or filter logs.
+All other fields are not indexed by default.
+But you can still add any top-level fields and use them in searches and aggregations as the default index template for logs
+<<dynamic-mapping-runtime-fields, maps dynamic fields as runtime fields>>.
+
+If there are custom fields that you frequently use in searches or aggregations, you might want to leverage the speed benefits of indexing the field.
```
graphaelli, line 51:
```suggestion
For custom fields that are frequently used in searches or aggregations, you might want to leverage the speed benefits of indexing the field.
```

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+This comes at the expense of a larger index size more processing at ingest time.
+To index a custom field, <<create-index-template, create an index template>> for your dataset `logs-<dataset>-*` based on the built-in index template for `logs-*-*` and add your custom field to the mappings.
+If wanted, you can <<manually-roll-over-a-data-stream, roll over the data stream>> to immediately apply the setting to the data stream’s write index.
```
graphaelli, line 54:
```suggestion
To immediately apply the setting, <<manually-roll-over-a-data-stream, roll over the data stream>>
```

server/src/main/java/org/elasticsearch/rest/action/logs/RestLogsAction.java
jpountz: That makes sense to me.

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+`POST /_logs/<dataset>/<namespace>`
+
+[discrete]
+[[logs-api-prereqs]]
+=== {api-prereq-title}
+* If the {es} {security-features} are enabled, you must have the `create`
+<<privileges-list-indices,index privileges>> for the target data stream.
+* As the target stream depends on routing rules and the data provided in the log events,
+it's recommended to grant permissions for `logs-*-*`.
+* Automatic data stream creation requires a matching index template with data
```
ruflin, line 26: As the `logs-*-*` template already exist, I assume a matching template always exists either the global one or one created by users?

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+`POST /_logs/<dataset>`
+
+`POST /_logs/<dataset>/<namespace>`
+
+[discrete]
+[[logs-api-prereqs]]
+=== {api-prereq-title}
+* If the {es} {security-features} are enabled, you must have the `create`
+<<privileges-list-indices,index privileges>> for the target data stream.
+* As the target stream depends on routing rules and the data provided in the log events,
```
ruflin, line 24: I assume if a user selects directly a `dataset`, only has permissions there and no routing happens, things would just work. The broader permissions are only needed for routing.

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+the logs are sent to the `logs-dlq-<namespace>` data stream that acts as a dead letter queue for failed events.
+However, log ingestion should rarely fail as the mappings in the built-in index template for the `logs-*-*` data streams are designed to minimize mapping conflicts.
+
+A <<logs-api-request-body, couple of fields>> from the {ecs-ref}[Elastic Common Schema (ECS)] are indexed by default that are commonly used to search or filter logs.
+All other fields are not indexed by default.
+But you can still add any top-level fields and use them in searches and aggregations as the default index template for logs
+<<dynamic-mapping-runtime-fields, maps dynamic fields as runtime fields>>.
+
+For custom fields that are frequently used in searches or aggregations, you might want to leverage the speed benefits of indexing the field.
+This comes at the expense of a larger index size more processing at ingest time.
```
ruflin, line 52:
```suggestion
This comes at the expense of a larger index size and more processing at ingest time.
```

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+To index a custom field, <<create-index-template, create an index template>> for your dataset `logs-<dataset>-*` based on the built-in index template for `logs-*-*` and add your custom field to the mappings.
```
ruflin, line 53: Ideally that is a part we would have an API for, user should not have to think about templates etc. Same for the rollover below.

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+To immediately apply the setting, <<manually-roll-over-a-data-stream, roll over the data stream>>
+This affects any new data added to the stream after the rollover.
+However, it does not affect the data stream’s existing backing indices or existing data.
+
+All fields, aside from the `@timestamp` field, are configured to <<ignore-malformed, ignore malformed>> values.
+This means that if a log event contains a field whose type is incompatible with the type of the field that exists in the mapping,
+{es} will ignore the field instead of rejecting the whole document.
```
ruflin, line 60: Will the field still be available in `_source`?

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+  (Optional, string)
+  Defaults to `default`.
+  A user-configurable arbitrary grouping, such as an environment (dev, prod, or qa), a team, or a strategic business unit.
+
+[discrete]
+[[logs-api-query-params]]
+=== {api-query-parms-title}
+
+Any provided query parameter will be added to each log line.
+For example, `/_logs?service.name=myapp` will add `"service.name": "myapp"` to all logs.
```
ruflin, line 85: For later: I have the suspicion we will need the query params to set some other flags then adding data. Even though it looks very convenient, I wonder if this is the expected way of adding data. As we follow bulk requests, it does not fit into the body but it would likely be a better place for it?

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+The request body contains a newline-delimited list of log events to ingest.
+The individual events don't have any required fields and can contain arbitrary JSON content.
+There is no required structure for the log events, and you can add any top-level fields.
+However, it is recommended to follow the {ecs-ref}[Elastic Common Schema (ECS)] to structure your logs.
+
+TIP: Use the {ecs-logging-ref}/intro.html[ECS logging libraries] to create ECS-compliant JSON logs.
+
+Only the following fields, which are commonly used for searching, filtering, and correlating logs, are indexed by default:
+
+* {ecs-ref}/ecs-base.html#field-timestamp[`@timestamp`] +
```
ruflin, line 99: The list is pretty long. Maybe we can get it even shorter?

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+* {ecs-ref}/ecs-host.html#field-host-name[`host.name`]
+* {ecs-ref}/ecs-container.html#field-container-id[`container.id`]
+* {ecs-ref}/ecs-container.html#field-container-name[`container.name`]
+* {ecs-ref}/ecs-orchestrator.html#field-orchestrator-namespace[`orchestrator.namespace`]
+* {ecs-ref}/ecs-orchestrator.html#field-orchestrator-cluster-id[`orchestrator.cluster.id`]
+* {ecs-ref}/ecs-orchestrator.html#field-orchestrator-cluster-name[`orchestrator.cluster.name`]
+* {ecs-ref}/ecs-orchestrator.html#field-orchestrator-resource-id[`orchestrator.resource.id`]
+* {ecs-ref}/ecs-orchestrator.html#field-orchestrator-resource-name[`orchestrator.resource.name`]
+
+Dotted field names are expanded to objects so that they can be used interchangeably with nested objects. For example, the following documents are treated equally: `{"log.level": "INFO"}`, `{"log": { "level": "INFO"} }`.
```
ruflin, line 133: Should we directly start to use subobject false to prevent conflicts?

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+
+`_metadata`::
```
ruflin, line 135: Would be nice to show here an example event.

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+[discrete]
+[[logs-api-response-body]]
+==== {api-response-body-title}
+
+The log API's response body is always empty.
+
+Status
+
+* 202 Accepted: The log events have been received and are processed in the background. They should be searchable after a short while.
+* 500 Internal Server Error: There was an error while processing the log events. Some logs may have been lost.
```
ruflin, line 152: What is the expected client behaviour in this case? Resend all data?

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+// including the logs API which interprets it as global metadata that's added to every event
+
+'''
+
+Ingests a single log into the `logs-myapp-default` data stream.
+
+[source,console]
+------------------------------------------------------------
+POST _logs/myapp
+{"_metadata": {"service.name": "myapp"}} <1>
```
ruflin, line 245: Here we go with the example :-)

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+{"@timestamp": "2016-05-23T08:05:34.853Z", "message": "Hello World"} <2>
+------------------------------------------------------------
+
+<1> Provides global metadata that applies to all log events in the request via a global metadata event.
+<2> Specifies the timestamp as an ISO date string.
+
+'''
+
+Ingests a two log events into the `logs-myapp-default` and `logs-my_other_app-default` data stream, respectively.
```
ruflin, line 254:
```suggestion
Ingests two log events into the `logs-myapp-default` and `logs-my_other_app-default` data stream, respectively.
```

docs/reference/rest-api/logs.asciidoc
```diff
@@ -0,0 +1,278 @@
...
+------------------------------------------------------------
+POST _logs
+{"_metadata": {"service.name": "myapp"}} <1>
+{"_metadata": {"data_stream.dataset": "myapp"}} <2>
+{"@timestamp": "2016-05-23T08:05:34.853Z", "message": "hello app"} <3>
+{"_metadata": {"data_stream.dataset": "my_other_app"}} <4>
+{"@timestamp": "2016-05-23T08:05:34.853Z", "message": "other app"} <5>
+------------------------------------------------------------
+
+<1> Provides global metadata that applies to all log events in the request via a global metadata event.
```
ruflin, line 267: I understand the logic but not sure if this is intuitive. Maybe instead call it `_meta_global` and `_meta_local` or similar to make it more describtive

ruflin (commented):
Left some comments on the docs, will have a closer look at the code soonish.

docs/reference/rest-api/logs.asciidoc
felixbarny: Yes

docs/reference/rest-api/logs.asciidoc
felixbarny: > I have the suspicion we will need the query params to set some other flags then adding data.

What I had in mind is that these flags would be prefixed by an underscore.

docs/reference/rest-api/logs.asciidoc
felixbarny: Shorter, because it takes up a lot of space in the docs or because you think some of the fields should not be indexed by default? If so, which ones and why?

docs/reference/rest-api/logs.asciidoc
felixbarny: I think we should but there are some missing features. See: https://github.com/elastic/elasticsearch/issues/88934

docs/reference/rest-api/logs.asciidoc
felixbarny: There are some examples at the bottom of this page.

docs/reference/rest-api/logs.asciidoc
felixbarny: Fair point

docs/reference/rest-api/logs.asciidoc
felixbarny: Another reason for expanding dots is to be able to access the fields in a consistent way in ingest pipelines and scripts so that you don't need to have something like `_doc['log']['level']` and `_doc['log.level']`.

ruflin: That argument can be taken both ways. I think the important bit is that it is consistent, either all flat or all expanded so that we don't have something like `_doc['foo']['hello.world']`

rest-api-spec/src/main/resources/rest-api-spec/api/logs.json
```diff
@@ -0,0 +1,57 @@
...
+        },
+        {
+          "path":"/_logs/{dataset}",
+          "methods":[
+            "POST"
+          ],
+          "parts":{
+            "dataset":{
+              "type":"string",
+              "description":"Default index for items which don't provide one"
```
ruflin, line 29: I'm leaving some comments on this PR which do not have to be addressed now but is rather a note for later when we get into the details of implementation.

For later: We need to discuss the exact descriptions for users here. Index sounds off. Below you have a different one.

server/src/main/java/org/elasticsearch/rest/action/logs/DataStreamNamingUtils.java
```diff
@@ -0,0 +1,44 @@
...
+        return sanitizeDataStreamField(namespace, DISALLOWED_IN_NAMESPACE);
+    }
+
+    private static String sanitizeDataStreamField(String s, char[] disallowedInDataset) {
+        if (s == null) {
+            return null;
+        }
+        s = s.toLowerCase(Locale.ROOT);
+        s = s.substring(0, Math.min(s.length(), MAX_LENGTH));
+        for (char c : disallowedInDataset) {
```
ruflin, line 39: Would be nice to have some debug logging around this at least the first time it has to be done. Or some way to let users know that they ingest data that has to be sanitised. They might be able to fix the source.

server/src/main/java/org/elasticsearch/rest/action/logs/RestLogsAction.java
```diff
@@ -0,0 +1,252 @@
...
+                    });
+                }
+            });
+        };
+    }
+
+    private Map<String, Object> createDlqDoc(Map<String, Object> doc, Exception cause) {
+        Map<String, Object> dlqDoc = new HashMap<>();
+        dlqDoc.put("@timestamp", Instant.now().toString());
+        addPath(dlqDoc, "event.original", doc);
```
ruflin, line 156: The problem with this and the `error.*` is on what happens, if these fields already existed in the event? Will these be overwritten? Ideally we would store this data in a predefined structure in the document that can't conflict, something more "internal". Maybe we could start it with `_` or `.` like `_internal.event.orginal` and same for error. This has some disadvantages that it makes querying the dlq data harder but some magic on the ES side could help here assuming this is a special "index".

x-pack/plugin/core/src/main/resources/logs-dlq-mappings.json
```diff
@@ -0,0 +1,33 @@
...
+            "type": {
+              "type": "constant_keyword",
+              "value": "logs"
+            },
+            "dataset": {
+              "type": "constant_keyword",
+              "value": "dlq"
+            },
+            "namespace": {
+              "type": "constant_keyword"
```
ruflin, line 21: Does the dlq data stream need a namespace?

x-pack/plugin/core/src/main/resources/logs-dlq-template.json
```diff
@@ -0,0 +1,15 @@
+{
+  "index_patterns": ["logs-dlq-*"],
+  "priority": 200,
```
ruflin, line 3: Ideally users could never overwrite these templates so ♾️  would be here :-)

x-pack/plugin/core/src/main/resources/logs-mappings.json
```diff
@@ -1,14 +1,312 @@
...
+          "log_level": {
+            "path_match": "log.level",
+            "mapping": {
+              "type": "keyword",
+              "ignore_above": 1024
+            }
+          }
+        },
+        {
+          "log_logger": {
```
ruflin, line 32: Do we need this indexed by default? Runtime field instead?

x-pack/plugin/core/src/main/resources/logs-mappings.json
```diff
@@ -1,14 +1,312 @@
...
+            "match_mapping_type": "string",
+            "match": "ip",
+            "runtime": {
+              "type": "ip"
+            }
+          }
+        },
+        {
+          "message": {
+            "path_match": "message",
```
ruflin, line 16: Lets only index `message` on the top level. There is a good chance that also for example `error.message` exists but I don't think we should index it by default.

x-pack/plugin/core/src/main/resources/logs-mappings.json
```diff
@@ -1,14 +1,312 @@
...
+          "log_logger": {
+            "path_match": "log.logger",
+            "mapping": {
+              "type": "keyword",
+              "ignore_above": 1024
+            }
+          }
+        },
+        {
+          "trace_id": {
```
ruflin, line 41: If it is a trace, don't we expect it in most cases that it goes to `trace-*-*` and likely already has a more specific mapping?

x-pack/plugin/core/src/main/resources/logs-mappings.json
ruflin: Hm, I might mix `match` and `patch_match`?

x-pack/plugin/core/src/main/resources/logs-mappings.json
```diff
@@ -1,14 +1,312 @@
...
+            "path_match": "service.name",
+            "mapping": {
+              "type": "keyword",
+              "ignore_above": 1024
+            }
+          }
+        },
+        {
+          "service_version": {
+            "path_match": "service.version",
```
ruflin, line 78: Runtime field by default? Could we use https://www.elastic.co/guide/en/elasticsearch/reference/current/version.html ?

x-pack/plugin/core/src/main/resources/logs-mappings.json
```diff
@@ -1,14 +1,312 @@
...
+            "mapping": {
+              "type": "keyword",
+              "ignore_above": 1024
+            }
+          }
+        },
+        {
+          "service_environment": {
+            "path_match": "service.environment",
```
ruflin, line 87: Before I comment on each of these, I think we need to go through each one and discuss if we really not to index it. I think we can start with a smaller subset.

x-pack/plugin/core/src/main/resources/logs-mappings.json
```diff
@@ -1,14 +1,312 @@
...
+            }
+          }
+        }
+      ],
+      "dynamic": "runtime",
+      "date_detection": false,
       "properties": {
+        "@timestamp": {
+          "type": "date",
+          "ignore_malformed": false
```
ruflin, line 251: If `@timestamp` is malformed, the `_logs` endpoint handles this error and ingests the data again?

server/src/main/java/org/elasticsearch/rest/action/logs/RestLogsAction.java
felixbarny: > Will these be overwritten?

No, the original event is stored under event.original so that there's no opportunity for a conflict:

```json
{
  "error": {
    "type": "mapper_parsing_exception",
    "message": "..."
  },
  "event": {
    "original": {
      "@timestamp": "...",
      "message": "Hello World"
    }
  }
}
```

It means that the original event is not on the top level anymore but I think that's a fine tradeoff.

x-pack/plugin/core/src/main/resources/logs-dlq-mappings.json
felixbarny: Yes, I think every data stream should have a namespace. The question is whether it should be customizable or be fixed to `default`. I think a dlq per namespace makes sense. It helps with managing ILM and permissions without the dlqs being too fine-grained and therefore hard to manage, such as having one per data stream.

x-pack/plugin/core/src/main/resources/logs-mappings.json
felixbarny: Yes, `"path_match": "message"` only indexes `message` and doesn't index `error.message`. If I used `"match": "message"` or "path_match": "*message", `error.message` would be indexed, too.

x-pack/plugin/core/src/main/resources/logs-mappings.json
felixbarny: It's not used as often as the `log.level`, but it is useful to be able to filter for logs of a specific sub-section of the app. I think it's debatable. But it's also not set by all logger frameworks. So if it's set, why not index it? Not feeling strongly about this field, though.

x-pack/plugin/core/src/main/resources/logs-mappings.json
felixbarny: No, this is about trace/log correlation. APM Agents can inject the `trace.id` to logs so that you can navigate to the APM trace that correlates to this log.

x-pack/plugin/core/src/main/resources/logs-mappings.json
felixbarny: Good idea about using the version field! But I think it should be indexed by default. It can be really useful in automated correlations to see if a new version is more prone to logging errors, for example.

x-pack/plugin/core/src/main/resources/logs-mappings.json
felixbarny: Sure, let's do that. But I don't think we should subtract a lot here.

x-pack/plugin/core/src/main/resources/logs-mappings.json
felixbarny: Yes, events with a malformed timestamp will land in the dlq. The dlq always sets the top-level `@timestamp` to the ingest time. The `@timestamp` from the even is stored in `event.original.@timestamp`.

x-pack/plugin/core/src/main/resources/logs-mappings.json
```diff
@@ -1,14 +1,312 @@
...
+              "type": "match_only_text"
+            }
+          }
+        },
+        {
+          "log_level": {
+            "path_match": "log.level",
+            "mapping": {
+              "type": "keyword",
+              "ignore_above": 1024
```
javanna, line 27: do we need to specify ignore_above here?

x-pack/plugin/core/src/main/resources/logs-mappings.json
```diff
@@ -1,14 +1,312 @@
...
+        },
+        "cloud": {
+          "type": "object"
+        },
+        "container": {
+          "type": "object"
+        },
+        "host": {
+          "type": "object"
+        },
```
javanna, line 293: I take that we would like to use subobjects:false here but we first need to automatically collapse objects in documents (https://github.com/elastic/elasticsearch/issues/88934) ? It would be nice to not have to map these objects, which I understand are mapped only to prevent users from ingesting docs with fields that have values for host, container etc.

x-pack/plugin/core/src/main/resources/logs-mappings.json
felixbarny: Good question. That's what we previously did for all keyword fields and I didn't what to change it.

https://github.com/elastic/elasticsearch/blob/5d6af5890f70cfdb234240f7a1d8c65c6339315d/x-pack/plugin/core/src/main/resources/data-streams-mappings.json#L23-L31

x-pack/plugin/core/src/main/resources/logs-mappings.json
felixbarny: > I take that we would like to use subobjects:false here but we first need to automatically collapse objects in documents (#88934) ?

Yes :)

> It would be nice to not have to map these objects, which I understand are mapped only to prevent users from ingesting docs with fields that have values for host, container etc.

Without this, the dynamic templates don't work unfortunately, it's because dynamic mapping is disabled so when ingesting `log.level`, there's no object for `log`. Probably a non-issue once we use `subobjects: false`.

x-pack/plugin/core/src/main/resources/logs-mappings.json
ruflin: In theory no, in practice maybe. Lets assume for whatever reason log.level value is > 32KB, in this case, ES rejects the document and we never want this to happen. The part I don't know is what ignore_malformed does in this scenario, likely worth to test.

felixbarny:

PR #88183, opened 2022-06-29
Ingest: RandomDocumentPicks.randomExistingFieldName doesn't return _version

stu-elastic:
When `RandomDocumentPicks.randomExistingFieldName` selects a field, the
caller expects to be able to manipulate that field in any manner.  `_version` has
special validation, must fit in an `long` and may not be `null`.

This change no longer returns `_version` from `randomExistingFieldName` unless it is the only field in the document.

Related to change: https://github.com/elastic/elasticsearch/pull/88102


martijnvg (approved):
LGTM

PR #88184, opened 2022-06-29
Merge creating indices and shards methods in IndicesClusterStateService

original-brownbear:
Motivated by the fact that with recent improvements to shard allocation, mapping parsing etc., data node cluster application time now started to become the slowest part of bootstrapping many-shards benchmarks. This helps by a couple percent and there's obvious follow-ups to do similar things to other parts of `applyClusterState` which do a lot of redundant work similar to what's fixed here.

-> Looping over the local routing node can be quite expensive for e.g.
large warm/cold/frozen nodes. We can save one full iteration as well
as a couple of map lookups by dealing with index and shard creation+updates
in a single loop.

relates #77466


tlrx (approved):
LGTM (assuming CI is happy)

original-brownbear:
Jenkins test this?

original-brownbear:
Jenkins test this?

original-brownbear:
Thanks Tanguy!

PR #88186, opened 2022-06-29
Support updates of API key attributes [REST and transport layer]

n1v0lg:
REST and transport layer implementation to add support for updating
attributes of existing API keys. This allows end-users to modify
privileges and metadata associated with API keys dynamically, without
requiring rolling out new API keys every time there is a change.

The new route supports updates to one API key, given its ID:

`PUT /_security/api_key/{id}`

The request body consists of optional fields `role_descriptors` and
`metadata`. If a request field is absent, the existing value of the
field on the given API key is retained. If a request field is set to
`{}` it replaces the existing value with `{}`. Explicit null-values for
request fields are not allowed and will produce a 400.
`limited_by_role_descriptors`, `creator`, and `version` are
automatically updated on every call. Attributes a replaced, not merged.

Only the owner user of an API key can update it. API keys cannot update
themselves, nor can other users (even users with `all` or
`manage_security` cluster privileges).

Relates: #87870


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/apikey/TransportUpdateApiKeyAction.java
```diff
@@ -0,0 +1,80 @@
...
+            listener.onFailure(new IllegalStateException("authentication is required"));
+            return;
+        } else if (authentication.isApiKey()) {
+            listener.onFailure(new IllegalArgumentException("authentication via an API key is not supported for updating API keys"));
+            return;
+        }
+
+        // Don't resolve and validate owner roles if the service is not enabled; this avoids a costly operation and also provides
+        // a clearer error message (service disabled error is prioritized over role descriptor validation errors)
+        apiKeyService.ensureEnabled();
```
n1v0lg, line 65: This block of code is quite similar to what we have in `ApiKeyGenerator`. Re-using it however would require refactoring around the `ApiKeyGenerator` class and I haven't come up with a good approach for that yet. One idea is to generalize `ApiKeyGenerator` to something like `ApiKeyManager` with `create` and `update` methods. Not convinced yet that's the way to go -- the class loses it's single responsibility that way.

I'm wondering around leaving the current implementation and revisiting a possible refactor. WYDT?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/apikey/TransportUpdateApiKeyAction.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    protected void doExecute(Task task, UpdateApiKeyRequest request, ActionListener<UpdateApiKeyResponse> listener) {
+        final var authentication = securityContext.getAuthentication();
+        if (authentication == null) {
+            listener.onFailure(new IllegalStateException("authentication is required"));
+            return;
+        } else if (authentication.isApiKey()) {
+            listener.onFailure(new IllegalArgumentException("authentication via an API key is not supported for updating API keys"));
```
n1v0lg, line 59: I'm adding the check here and treating it as a "bad" request, instead failing authorization. To me, it feels more appropriate since in principle an API key is, for instance, "authorized" to update itself, it's just that we don't support that functionality. Think this also leads to a clearer error message. Let me know if you disagree!

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/privilege/ManageOwnApiKeyClusterPrivilege.java
```diff
@@ -93,6 +94,13 @@ protected boolean extendedCheck(String action, TransportRequest request, Authent
                 }
             } else if (request instanceof final QueryApiKeyRequest queryApiKeyRequest) {
                 return queryApiKeyRequest.isFilterForCurrentUser();
+            } else if (request instanceof UpdateApiKeyRequest) {
+                // Note: this returns true even for requests authenticated via API keys.
```
n1v0lg, line 98: Wanted to check on this: does the reasoning here make sense?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/rest/action/apikey/RestUpdateApiKeyAction.java
```diff
@@ -0,0 +1,72 @@
...
+        PARSER.declareObject(optionalConstructorArg(), (p, c) -> p.map(), new ParseField("metadata"));
+    }
+
+    public RestUpdateApiKeyAction(final Settings settings, final XPackLicenseState licenseState) {
+        super(settings, licenseState);
+    }
+
+    @Override
+    public List<Route> routes() {
+        return List.of(new Route(POST, "/_security/api_key/_update/{id}"));
```
n1v0lg, line 52: I think we currently have a bug in our route validation: `PUT /_security/api_key/{id}` is not possible here since it clashes with `POST /_security/api_key/{ids}/_clear_cache` because of the different wildcard path parameter names. Starting the node with the PUT route in place leads to an illegal arg exception.

```
fatal exception while booting Elasticsearchjava.lang.IllegalArgumentException: Trying to use conflicting wildcard names for same path: ids and id
```

I don't think this is correct behavior -- the path suffix should be sufficient to distinguish those two routes. I'll reproduce in a test and open an issue.

In any case, that's why I'm going for `POST /_security/api_key/_update/{id}` here for now. Don't think this is a blocker for this PR -- we can always tweak the route, as long as it's before 8.4 release.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/apikey/TransportUpdateApiKeyAction.java
ywangd: I'd prefer to _not_ duplicating the code here. We had previous bugs in handling limited role descriptors. So my instinct would be reusing the logic in ApiKeyGenerator with some refactoring. The reusable part is mainly about getting the list of limited role descriptors for creating API keys (now updating as well). I think it can be refactored into its own method (still within ApiKeyGenerator) relatively easily, e.g.:

```java
public void generateApiKey(Authentication authentication, CreateApiKeyRequest request, ActionListener<CreateApiKeyResponse> listener) {
    if (authentication == null) {
        listener.onFailure(new ElasticsearchSecurityException("no authentication available to generate API key"));
        return;
    }
    apiKeyService.ensureEnabled();

    getUserRoleDescriptors(authentication, ActionListener.wrap(roleDescriptors -> {
        apiKeyService.createApiKey(authentication, request, roleDescriptors, listener);
    }, listener::onFailure));
}

private void getUserRoleDescriptors(Authentication authentication, ActionListener<Set<RoleDescriptor>> listener) {
    final ActionListener<Set<RoleDescriptor>> roleDescriptorsListener = ActionListener.wrap(roleDescriptors -> {
        for (RoleDescriptor rd : roleDescriptors) {
            try {
                DLSRoleQueryValidator.validateQueryField(rd.getIndicesPrivileges(), xContentRegistry);
            } catch (ElasticsearchException | IllegalArgumentException e) {
                listener.onFailure(e);
                return;
            }
        }
        listener.onResponse(roleDescriptors);
    }, listener::onFailure);

    final Subject effectiveSubject = authentication.getEffectiveSubject();

    // Retain current behaviour that User of an API key authentication has no roles
    if (effectiveSubject.getType() == Subject.Type.API_KEY) {
        roleDescriptorsListener.onResponse(Set.of());
        return;
    }

    rolesStore.getRoleDescriptorsList(effectiveSubject, ActionListener.wrap(roleDescriptorsList -> {
        assert roleDescriptorsList.size() == 1;
        roleDescriptorsListener.onResponse(roleDescriptorsList.iterator().next());

    }, roleDescriptorsListener::onFailure));
}
```

The above refactoring should not cause any cascading changes for both main and test code. One advantage of the separate and reusable method is that it can be easily moved to another place if necessary (e.g. if we want to merge it into ApiKeyService). I am not concerned about its single responsibility since I don't think we particularly wanted it. My guess is that it existed because we didn't want to expose RolesStore to ApiKeyService.

We might want to have more refactoring around ApiKeyService, either merging ApiKeyGenerator entirely into ApiKeyService, or as you mentioned, enhance ApiKeyGenerator to be something like ApiKeyManager as a high level facade of the Service class. In either case, we probably won't have this logic inside the transport action.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/privilege/ManageOwnApiKeyClusterPrivilege.java
ywangd: Trying to update the API key using the API key itself should be an authorization error, _not_ because of technical constraint. An API key is _not_ the owner of itself. An analogy is that a user cannot update its own roles. It is a common use case that an API key is created by a higher privileged user (e.g. a superuser), and gets handed to a different user (or application) to use. That different user should not be able to update the key using the key itself since that could lead to privilege escalation.

What we haven't decided yet is whether an API key can own its derived keys. Right now, derived keys are owned by no one. If we decide a derived key should be owned by the API key that creates it, we may need revise the logic here to allow it. But this does not change the fact that an API key does _not_ own itself (and can _never_ update itself).

If we want to consider all the nuisances, we should:
1. AuthZ (403) error if `isCurrentAuthenticationUsingSameApiKeyIdFromRequest == true`.
2. 400 error if the authentication is an API Key

But I think simply having 403 for both cases is easier (both explanation and implementation). It is also a better consistency if you consider the future bulk update API.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/rest/action/apikey/RestUpdateApiKeyAction.java
ywangd: It's not a bug. The issue is that we have to use the same placeholder name for different APIs that share the same prefix. So it will work if we have the following instead
```
new Route(POST, "/_security/api_key/{ids}")
```
(note `ids` vs `id`)

However, the new API guideline recommends singular form over plural in URL path. So we should instead change `POST /_security/api_key/{ids}/_clear_cache` to `POST /_security/api_key/{id}/_clear_cache`. This is not really a user-facing change and should be fine in a minor.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/apikey/TransportUpdateApiKeyAction.java
ywangd: Please see my [other comment](https://github.com/elastic/elasticsearch/pull/88186#discussion_r910521670) on the same topic.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/rest/action/apikey/RestUpdateApiKeyAction.java
n1v0lg: Yup, I understand the fix for this -- my point is that it seems like an arbitrary limitation. Routes with the same prefix but different suffix could realistically co-exist with different wildcards, and don't need to be coupled. If this is known and accepted behavior it's all good though! Not a big limitation.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/apikey/TransportUpdateApiKeyAction.java
n1v0lg: I'm good with the refactoring 👍 I really like how focused and small `ApiKeyGenerator` is but I agree that the code re-use makes sense here.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/rest/action/apikey/RestUpdateApiKeyAction.java
ywangd: > If this is known and accepted behavior

It's definitely know. Maybe largely accepted as well. But probably not set in stone. core-infra will be able to give out better advices.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/privilege/ManageOwnApiKeyClusterPrivilege.java
n1v0lg: We had a sync on this on slack and decided to return `true` here and go for a `400`:

Fundamentally, the limitation that only an owner can update an API key sits outside the authorization layer: there are other privileges like `all` or `manage_security` that, from an access perspective, would enable an API key to update itself. However, we still prevent the behavior because of the agreed upon semantics of the update API key operation, namely that only the owner *user* can directly (things like run-as are still possible) update an API key and no other entity.

As such, a `400` is the more appropriate response here.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -1230,50 +1231,47 @@ private static VersionedApiKeyDoc single(final String apiKeyId, final Collection
     }
 
     private BulkRequest buildBulkRequestForUpdate(
-        final Collection<VersionedApiKeyDoc> currentVersionedDocs,
+        final VersionedApiKeyDoc apiKeyDoc,
```
n1v0lg, line 1234: I couldn't resist a small refactor here: we already enforce that we have a single api key doc before calling this method. It makes little sense to go back to suddenty handling a collection.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -1497,71 +1495,73 @@ public void testUpdateApiKey() throws ExecutionException, InterruptedException,
...
-                RoleDescriptorTests.randomRoleDescriptor()
-            );
-            case 2 -> null;
-            default -> throw new IllegalStateException("unexpected case no");
-        };
-    }
-
     public void testUpdateApiKeyNotFoundScenarios() throws ExecutionException, InterruptedException {
-        final Tuple<CreateApiKeyResponse, Map<String, Object>> createdApiKey = createApiKey(ES_TEST_ROOT_USER, null);
+        final Tuple<CreateApiKeyResponse, Map<String, Object>> createdApiKey = createApiKey(TEST_USER_NAME, null);
```
n1v0lg, line 1499: Switching users here because the role of `ES_TEST_ROOT_USER` does not pass validation that happens during role parsing under the hood of `expectRoleDescriptorsForApiKey`

x-pack/plugin/security/qa/security-trial/src/javaRestTest/java/org/elasticsearch/xpack/security/apikey/ApiKeyRestIT.java
```diff
@@ -49,22 +52,28 @@ public class ApiKeyRestIT extends SecurityOnTrialLicenseRestTestCase {
     private static final SecureString SYSTEM_USER_PASSWORD = new SecureString("system-user-password".toCharArray());
     private static final String END_USER = "end_user";
     private static final SecureString END_USER_PASSWORD = new SecureString("end-user-password".toCharArray());
+    private static final String MANAGE_OWN_API_KEY_USER = "manage_own_api_key_user";
```
n1v0lg, line 55: Multiple tests using this user now, so lifting it out of the specific test into suite scope

n1v0lg:
@elasticmachine run elasticsearch-ci/part-2-fips plz

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/rest/action/apikey/RestUpdateApiKeyActionTests.java
```diff
@@ -0,0 +1,57 @@
...
+            requestHolder.set((UpdateApiKeyRequest) actionRequest);
+            return new UpdateApiKeyResponse(true);
+        }));
+    }
+
+    public void testAbsentRoleDescriptorsAndMetadataSetToNull() {
+        final var apiKeyId = "api_key_id";
+        final FakeRestRequest restRequest = new FakeRestRequest.Builder(xContentRegistry()).withMethod(RestRequest.Method.PUT)
+            .withPath("/_security/api_key/" + apiKeyId)
+            .withContent(new BytesArray("{}"), XContentType.JSON)
```
ywangd, line 50: I think we should randomise this so that sometimes the request has no content at all.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/rest/action/apikey/RestUpdateApiKeyAction.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public String getName() {
+        return "xpack_security_update_api_key";
+    }
+
+    @Override
+    protected RestChannelConsumer innerPrepareRequest(final RestRequest request, final NodeClient client) throws IOException {
+        final var apiKeyId = request.param("ids");
+        final var payload = PARSER.parse(request.contentParser(), null);
```
ywangd, line 62: I think we should allow "no request body" since all fields are optional and the request makes sense as long as it has an ID (from the path parameter).

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -1497,71 +1499,119 @@ public void testUpdateApiKey() throws ExecutionException, InterruptedException,
...
+            new UpdateApiKeyRequest(apiKeyForNativeRealmUser.getId(), request.getRoleDescriptors(), request.getMetadata())
         );
     }
 
     public void testInvalidUpdateApiKeyScenarios() throws ExecutionException, InterruptedException {
-        final Tuple<CreateApiKeyResponse, Map<String, Object>> createdApiKey = createApiKey(ES_TEST_ROOT_USER, null);
-        final var apiKeyId = createdApiKey.v1().getId();
+        final List<String> apiKeyPrivileges = new ArrayList<>(randomSubsetOf(ClusterPrivilegeResolver.names()));
+        // At a minimum include `manage_own_api_key` to ensure no 403
+        apiKeyPrivileges.add("manage_own_api_key");
```
ywangd, line 1594:
```suggestion
        apiKeyPrivileges.add(randomFrom("manage_own_api_key", "manage_api_key"));
```

ywangd (approved):
LGTM

No major comments. But I appreciate if you could address them before merging. Thanks!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -1230,50 +1234,47 @@ private static VersionedApiKeyDoc single(final String apiKeyId, final Collection
     }
 
     private BulkRequest buildBulkRequestForUpdate(
-        final Collection<VersionedApiKeyDoc> currentVersionedDocs,
+        final VersionedApiKeyDoc versionedDoc,
```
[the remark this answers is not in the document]
n1v0lg: Yup that was my thinking originally but the method was essentially still specialized for a single update and couldn't be directly re-used for bulk. I see more refactoring happening here once we get to bulk updates either way.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/security/authz/privilege/ManageOwnApiKeyClusterPrivilegeTests.java
```diff
@@ -69,8 +87,10 @@ public void testAuthenticationWithUserAllowsAccessToApiKeyActionsWhenItIsOwner()
 
         TransportRequest getApiKeyRequest = GetApiKeyRequest.usingRealmAndUserName(realmRef.getName(), "joe");
         TransportRequest invalidateApiKeyRequest = InvalidateApiKeyRequest.usingRealmAndUserName(realmRef.getName(), "joe");
+        TransportRequest updateApiKeyRequest = UpdateApiKeyRequest.usingApiKeyId(randomAlphaOfLength(10));
         assertTrue(clusterPermission.check("cluster:admin/xpack/security/api_key/get", getApiKeyRequest, authentication));
         assertTrue(clusterPermission.check("cluster:admin/xpack/security/api_key/invalidate", invalidateApiKeyRequest, authentication));
+        assertTrue(clusterPermission.check("cluster:admin/xpack/security/api_key/update", updateApiKeyRequest, authentication));
```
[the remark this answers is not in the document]
n1v0lg: Indeed.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/rest/action/apikey/RestUpdateApiKeyAction.java
```diff
@@ -0,0 +1,71 @@
...
+        PARSER.declareObject(optionalConstructorArg(), (p, c) -> p.map(), new ParseField("metadata"));
+    }
+
+    public RestUpdateApiKeyAction(final Settings settings, final XPackLicenseState licenseState) {
+        super(settings, licenseState);
+    }
+
+    @Override
+    public List<Route> routes() {
+        return List.of(new Route(PUT, "/_security/api_key/{ids}"));
```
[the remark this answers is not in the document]
n1v0lg: 👍 will add a comment

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
n1v0lg: This is odd: I thought the leading `_` in the role name (`_es_test_root`) was failing the validation but that's not the case apparently. I think the original failure was due to the random role descriptors being generated failing for a different reason which I have since fixed. 

**tl;dr** it was actually fine to use the root user here. I'll keep `TEST_USER_NAME` here since the code is already rewritten to use that name. Sorry for the confusion!

ywangd: Thanks for the explanation. All good.

ylasri:
We got this error when using this API, on version 7.17.6


```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "illegal_argument_exception",
        "reason" : "request [/_security/api_key/yAlcM4QBdS8O9IF71bZo] contains unrecognized parameter: [ids] -> did you mean [id]?"
      }
    ],
    "type" : "illegal_argument_exception",
    "reason" : "request [/_security/api_key/yAlcM4QBdS8O9IF71bZo] contains unrecognized parameter: [ids] -> did you mean [id]?"
  },
  "status" : 400
}
```

ywangd:
> We got this error when using this API, on version 7.17.6

This API is for version 8.4.0 and later. You must have confused it with some other APIs. 

Also, this appears to be a user question, and we'd like to direct these kinds of things to the [Elasticsearch forum](https://discuss.elastic.co/c/elasticsearch). If you can stop by there, we'd appreciate it. This allows us to use GitHub for verified bug reports, feature requests, and pull requests. There's an active community in the [forum](https://discuss.elastic.co/c/elasticsearch) that should be able to help get an answer to your question.

PR #88188, opened 2022-06-29
Allow easing testing of java ea versions using gradle java tool chain support

breskeby:
This introduces the ability to simply configure a java tool chain for elasticsearch java projects to be used.
If an environment variable `JAVA_TOOLCHAIN_HOME `is declared, this JDK will be used as toolChain in elasticsearch.java projects.

This should make testing our build with java `ea` versions easier and allows detangling the used compiler jdk from the gradle java runtime.

### DIFF
--- a/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaBasePlugin.java
+++ b/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaBasePlugin.java
@@ -26,7 +26,6 @@ import org.gradle.api.tasks.compile.AbstractCompile;
 import org.gradle.api.tasks.compile.CompileOptions;
 import org.gradle.api.tasks.compile.GroovyCompile;
 import org.gradle.api.tasks.compile.JavaCompile;
-import org.gradle.jvm.toolchain.JavaToolchainSpec;
 
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaBasePlugin.java
```diff
@@ -106,11 +106,10 @@ private static void disableTransitiveDependenciesForSourceSet(Project project, S
      */
     public static void configureCompile(Project project) {
         project.getExtensions().getExtraProperties().set("compactProfile", "full");
-
         JavaPluginExtension java = project.getExtensions().getByType(JavaPluginExtension.class);
+        BuildParams.getJavaToolChainSpec().map(spec -> java.toolchain(spec)).getOrNull();
```
mark-vieira, line 110: Having side-effects in a `map()` call seems a bit like bad practice. I think being explicit here and just putting this behind a call to `ifPresent()` seems more idiomatic.

mark-vieira (commented):
Question: Is this toolchain used for both compilation and test execution? If so, is that the behavior we want?

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaBasePlugin.java
breskeby: I wish gradle would have a more functional api here. Changed.

breskeby:
@mark-vieira yes declaring a tool chain at the java extension level actually applies that toolchain to all compile, test and javadoc tasks. We could split this up and only apply it to compile tasks if required. I think having all set at once seems the reasonable default. We can put more effort into this and split this up or only apply it to compile tasks. What would you think is a most useful default here @ChrisHegarty @rjernst?

rjernst:
> What would you think is a most useful default here

I think apply to compile, test and javadoc is a reasonable default.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/ElasticsearchJavaBasePlugin.java
mark-vieira: Agreed. Might be a good simple submission. An `ifPresent()` method like exists on `java.util.Optional` that accepts a `Consumer` would be super useful.

mark-vieira:
If that's the case then it could be confusing if combined with `RUNTIME_JAVA_HOME`. We should make that explicit and enforce mutual exclusivity or at least warn the user that `RUNTIME_JAVA_HOME` will be ignored.

rjernst:
> We should make that explicit and enforce mutual exclusivity or at least warn the user that RUNTIME_JAVA_HOME will be ignored.

I don't think it needs to be a warning or ignored. Most of the time RUNTIME_JAVA_HOME is not set. Can we make it so that RUNTIME_JAVA_HOME is still used for tests *if* it is specified?

mark-vieira:
> Can we make it so that RUNTIME_JAVA_HOME is still used for tests _if_ it is specified?

This would be a good option if possible but I'm not sure the toolchain apis make that distinction.

breskeby:
From looking at the code using RUNTIME_JAVA_HOME for tests seems like a breaking change. 

Currently we only configure runtime java home for the test cluster runtime and this is not affected by this change here.

What I changed here is that when resolving RUNTIME_JAVA I fallback to toolchain java version if set and if not we fall back as usual to the jdk used by the gradle runtime (as we did before).

rjernst (approved):
I think the behavior here is what I would expect. Runtime Java home is used if it is found, otherwise fallback to the tool chain. Thanks Rene.

ChrisHegarty (approved):
LGTM.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/info/GlobalBuildInfoPlugin.java
```diff
@@ -164,6 +181,10 @@ private void logGlobalBuildInfo() {
             LOGGER.quiet("  JDK Version           : " + gradleJvmImplementationVersion + " (" + gradleJvmVendorDetails + ")");
             LOGGER.quiet("  JAVA_HOME             : " + gradleJvm.getJavaHome());
         }
+        String javaToolchainHome = System.getenv("JAVA_TOOLCHAIN_HOME");
+        if (javaToolchainHome != null) {
+            LOGGER.quiet("  JAVA TOOLCHAIN HOME   : " + javaToolchainHome);
```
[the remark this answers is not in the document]
breskeby: 👍 fixed

PR #88209, opened 2022-06-30
Rework shard snapshot workers

### LINKED ISSUE #83408: Order shards by "changes" when snapshoting
### Description

Snapshotting process when interrupted will return all not copied yet shards as PARTIAL.
If shards were ordered by what changes they have, the snapshoting process could work with "unchanged" shards first and finish them immediately. Then when working on copying a bigger "changed" shard, during interruption (node with shard is shut down) only the shards with changes will be marked as PARTIAL. 
This could be an improvement untill https://github.com/elastic/elasticsearch/issues/71333 is implemented.

pxsalehi:
Currently, when starting the shard snapshot tasks, we loop through the
shards to be snapshotted back-to-back. This will calculate whether/which
changes in the shard there are to be snapshotted, write some metadata,
and ONLY if there are changes and there are files to be uploaded, fork
off the file upload calls. This could be further improved by parallelizing 
shard snapshot tasks (currently, only file uploads happen in parallel). 

This change uses a worker pool and a queue to:

1. Paralleize (with limited number of workers) shard snapshotting, and 
    limit the number of the concurrently running snapshot tasks.
2. Prioritize shard snapshot tasks over file snapshot tasks.

Closes #83408


server/src/main/java/org/elasticsearch/snapshots/SnapshotShardsService.java
```diff
@@ -248,9 +248,8 @@ private void startNewSnapshots(List<SnapshotsInProgress.Entry> snapshotsInProgre
         }
     }
 
-    private void startNewShards(SnapshotsInProgress.Entry entry, Map<ShardId, IndexShardSnapshotStatus> startedShards) {
+    private void startNewShardSnapshots(SnapshotsInProgress.Entry entry, Map<ShardId, IndexShardSnapshotStatus> startedShards) {
```
idegtiarenko, line 251: May be `snapshotNewShards`? I believe we use `snapshot` as a verb quiet often. Also all shards belong to the same snapshot?

pxsalehi: I just tried to keep the names aligned with the existing methods. The caller of this is called `startNewSnapshots` and since each snapshot is broken down into snapshots of the shards (of the indices), I thought `startNewShardSnapshot` might fit better. I kept the `start*` naming since it conveys that the tasks are forked off. Is the naming confusing?

I think the `new` here just means those listed in the snapshotsInProgress, and I think, yes, all shards called here are within the same snapshot.

henningandersen:
I wonder if this approach is necessarily better. IIUC, we currently keep the one thread looping until the end. In the new approach, we risk the first (or an early) shard filling up the snapshot queue before the last tasks are added to the queue. This could mean that shards with no changes being scheduled after large file uploads?

original-brownbear:
> This could mean that shards with no changes being scheduled after large file uploads?

We figured this was somewhat unlikely (to happen a lot) and outweigh by the fact that we now multi-thread the starting of shard snapshots (which is the main motivation for this change, as we realized that running this single-threaded could take a long time on large data nodes).
But I guess if we want to be stricter about this, we could implement this differently by putting all the shards to start in a queue and then running threads (one of which would be the original thread that we fork from) that poll from that queue. That way we get even better ordering at the price of a little more code? Think it's worth it?

pxsalehi:
(Apologies for using too many vague/relative terms like many/few/large! But I need to somehow ask my question!)

For me one missing piece to evaluate the worst case impact of a file upload on unchanged shard snapshots, is the following: could the size of single files (to be uploaded) be extremely large? What would be a realistic "average" case? Or is it hard to tell since it could be wildly different depending on the settings/load? It seemed to me that many not-huge files to uploads is the common case. (Also another assumption from our talk, it seems often what happens during a snapshot is that most shards have no changes, and only few of them need to upload files.)

henningandersen:
We generally try to not fill the queue with all tasks, for instance in `BlobStoreRepository.snapshotShard` which only schedules as many threads as there are. I assume to allow other work to interleave?

If we do all shards in parallel, we risk getting `#shards * thread_pool_size` into the queue. It seems like the "throttling" we do per shard is no longer useful then? This could mean a duration of many minutes if not hours of no response on the `SNAPSHOT` thread pool. I think that could block restores, new followers and single-index snapshots? Am I missing something essential here?

I think this would be true also if we queue up the shard level execution and only do n jobs in parallel.

original-brownbear:
> Am I missing something essential here?

You're not directly missing anything actually. But I think there is a tradeoff here. This change makes it so that running a snapshot on a 10k shard warm-node may block the snapshot pool for ~15 min (assuming ~0.5s per shard snapshot and 5 snapshot threads max) vs. taking 75min to run a single snapshot without blocking the pool.
Taking the 15 min over the 75 min and accepting the blocking seems like the right tradeoff to me, especially when it's this easy to implement?
Also, note that we are doing the delete side where we're not being clever about doing any sort of hand-crafted work stealing queue like we do for file uploads either and no complaints about it have come in ever as far as I'm aware :)

> I think this would be true also if we queue up the shard level execution and only do n jobs in parallel.

True, the advantage of that approach would be that we'd have a harder guarantee on running the metadata work right before the file uploads (and it's a little faster potentially I guess).

pxsalehi:
I'm trying to understand how much of this discussion relates to the original issue (https://github.com/elastic/elasticsearch/issues/83408). 

If blocking the snapshot threadpool during a snapshot is a concern, then what would be an acceptable approach that addresses the issue and avoids blocking the threadpool? One thread going through the shard snapshot tasks, calculating which snapshots do not upload a file, doing all of those in the same thread, and then again in the same thread going through the shard snapshots that do have a change, and only fork-off the file-uploads? 

Basically, everything else other than file upload should be done on the thread going through the shards, and actually in two passes? No snapshotting of unchanged shards in parallel? Considering having many thousand unchanged shard snapshots and few shards actually uploading files is a normal occurrence, as Armin mentioned, parallelizing snapshotting of unchanged shards seems to reduce snapshot time, which I guess is an improvement.

My question is that if we do not want to parallelize shard snapshot tasks, then would the approach I mentioned above have any real improvement for the issue? Currently, all unchanged snapshots happen in the same thread, which doesn't seem that different than the two phase approach!

henningandersen:
Armin and I had a brief conversation on this and while the existing mechanism for limiting the number of upload threads is somewhat broken, running the outer level tasks in parallel will slightly increase this brokennes. We prefer to instead fix this for good by maintaining enough data structures to just fill the snapshot pool with the right amount of work without exhausting it for potentially hours.

Something like 2 queues (the other level work and the actual file upload work) and a counter for how many jobs are active should do. When any job finishes it checks the outer level queue first and once that is depleted it does the actual file upload jobs.

This also ensures that all no-change shards are done prior to any file uploads. I think Armin intends to chat sync with you on this.

pxsalehi:

(I think, test failure was unrelated to the PR. I opened: https://github.com/elastic/elasticsearch/issues/88615)

pxsalehi (commented):
@original-brownbear ~While still a draft PR,~ I have some questions about the implementation. Could you please take a look at what I have so far?

pxsalehi:
@elasticmachine please update branch

pxsalehi:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
```diff
@@ -3517,26 +3505,16 @@ public int getReadBufferSizeInBytes() {
...
-            this.newGeneration = newGeneration;
-            this.blobsToDelete = blobsToDelete;
-        }
-    }
+    private record ShardSnapshotMetaDeleteResult(
+        IndexId indexId,
+        int shardId,
+        ShardGeneration newGeneration,
+        Collection<String> blobsToDelete
+    ) {}
```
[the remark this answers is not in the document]
DaveCTurner: workflow nit: it's _so_ much easier on reviewers to make cosmetic changes like this as a separate PR. I'm definitely +1 on making cosmetic improvements as we find them, but -1 on piling them in to more complex changes like this.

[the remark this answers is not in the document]
pxsalehi: https://github.com/elastic/elasticsearch/pull/88721

server/src/test/java/org/elasticsearch/repositories/fs/FsRepositoryTests.java
```diff
@@ -149,8 +148,8 @@ public void testSnapshotAndRestore() throws IOException, InterruptedException {
             IndexCommit incIndexCommit = Lucene.getIndexCommit(Lucene.readSegmentInfos(store.directory()), store.directory());
             Collection<String> commitFileNames = incIndexCommit.getFileNames();
             final PlainActionFuture<ShardSnapshotResult> future2 = PlainActionFuture.newFuture();
+            IndexShardSnapshotStatus snapshotStatus = IndexShardSnapshotStatus.newInitializing(shardGeneration);
```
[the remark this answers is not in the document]
pxsalehi: Since these changes are also unrelated, I moved them out to https://github.com/elastic/elasticsearch/pull/88723.

pxsalehi:
I have also undone the changes to `SnapshotShardsService` in favour of https://github.com/elastic/elasticsearch/pull/88707.

server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
```diff
@@ -415,6 +418,12 @@ protected void doStart() {
         if (chunkSize != null && chunkSize.getBytes() <= 0) {
             throw new IllegalArgumentException("the chunk size cannot be negative: [" + chunkSize + "]");
         }
+        shardSnapshotWorkerPool = new ShardSnapshotWorkerPool(
+            threadPool.info(ThreadPool.Names.SNAPSHOT).getMax(),
```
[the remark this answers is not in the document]
original-brownbear: We want to upload at the maximum parallelism possible so max pool size like you use it here is what we want.

server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
```diff
@@ -2882,46 +2895,20 @@ public void snapshotShard(SnapshotShardContext context) {
                 snapshotStatus.moveToDone(threadPool.absoluteTimeInMillis(), shardSnapshotResult);
                 context.onResponse(shardSnapshotResult);
             }, context::onFailure);
-            if (indexIncrementalFileCount == 0) {
+            if (indexIncrementalFileCount == 0 || filesToSnapshot.isEmpty()) {
```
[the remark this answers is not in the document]
original-brownbear: Makes sense, neat fix saving some redundant tasks :)

server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
```diff
@@ -415,6 +418,12 @@ protected void doStart() {
         if (chunkSize != null && chunkSize.getBytes() <= 0) {
             throw new IllegalArgumentException("the chunk size cannot be negative: [" + chunkSize + "]");
         }
+        shardSnapshotWorkerPool = new ShardSnapshotWorkerPool(
```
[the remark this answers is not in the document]
original-brownbear: ++ makes sense

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkerPool.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+
+    public int size() {
+        return workerCount;
+    }
+
+    private void startWorker(final String workerId) {
+        logger.debug("starting snapshot pool worker {}", workerId);
+        executor.execute(() -> {
```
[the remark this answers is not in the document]
pxsalehi: Thank you both for the feedback.

> I think the answer here is "mostly no".

That's the part I am not entirely sure about when using two queues. I don't like the fact that it is not entirely predictable! It would be great to make sure there is a certain order regarding the sub-tasks, that will always be maintained when snapshotting. I am just not sure if it's an overkill, and don't want to add unnecessary code.

If it is not too much unnecessary code, I could use the `startTime` that is recorded in the `SnapshotsInProgress.Entry`. I think this is set once on master when the snapshot request gets processed, so it seems to stay unchanged afterwards. I could use a priority queue that sorts by snapshot `startTime` -> `snapshotId` -> task type(metadata work or file upload). I'm adding `startTime` to the mix, since ordering by `snapshotId` might still result in some weird corner case, where some snapshot keeps getting pushed back, due to ordering by  names/UUIDs. 

Would there be a downside to adding `startTime` to the ordering?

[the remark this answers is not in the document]
DaveCTurner: > this only matters if snapshots are triggered at a rate that's higher than the rate at which they can be completed.

I don't think that's quite true - there can temporarily be a bit spike in snapshot work to do even if in the long run we will be able to get through it all. For instance if a hot node fails and a bunch of replicas get promoted then that's going to cause a one-off bigger snapshot job. By prioritising this work over the work for later-starting ones we would typically finish the big snapshot sooner. Snapshots that finish sooner are kind of more useful because of how that reduces the window of uncertainty about which data is/isn't included.

server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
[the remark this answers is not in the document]
pxsalehi: Well, with the current PR this was necessary, as otherwise, we'd be creating a `GroupedActionListener` of size 0.

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkerPool.java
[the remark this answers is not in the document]
DaveCTurner: > To me it seems preferable to just always do the meta work first, then the file uploads.

I think I agree here, but that means we should also check for new metadata work between file uploads right?

server/src/test/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkersTests.java
```diff
@@ -0,0 +1,208 @@
...
+        assertThat(workers.size(), equalTo(maxSize));
+        snapshotBlocker.countDown();
+        // Eventually all workers exit
+        assertBusy(() -> assertThat(workers.size(), equalTo(0)));
+        assertThat(repo.finishedFileSnapshotTasks(), equalTo(repo.expectedFileSnapshotTasks()));
+        assertThat(repo.finishedShardSnapshotTasks(), equalTo(enqueuedSnapshots));
+        assertThat(repo.finishedShardSnapshots(), equalTo(enqueuedSnapshots));
+    }
+
+    public void testCompareToShardSnapshotTask() {
```
pxsalehi, line 183: To test that ShardSnapshotTask has priority over file snapshots, I tried to add some tests beyond just this compare test, but it turned out quite tricky to verify that. If you have any suggestions, please let me know.

pxsalehi (commented):
@henningandersen @DaveCTurner @original-brownbear 
This is now ready for review. Please have another look. After our discussion [here](https://github.com/elastic/elasticsearch/pull/88209#discussion_r925485272), I've decided to prioritize "meta" work over file uploads only within each snapshot, not across all snapshots. IMO, this would avoid parallel snapshots taking longer, and within each snapshot it would still finish unchanged shards first.

pxsalehi:
@elasticmachine update branch

original-brownbear (commented):
I left a couple of questions + comments mostly on the mechanical details of things. Overall this looks nice and exactly as we want it IMO :) So mainly just the sort order left to discuss here I think.

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
```diff
@@ -0,0 +1,179 @@
...
+        synchronized (mutex) {
+            snapshotTasks.add(new FileSnapshotTask(context, fileInfo, fileUploadListener));
+            ensureEnoughWorkers();
+        }
+    }
+
+    private void ensureEnoughWorkers() {
+        assert Thread.holdsLock(mutex);
+
+        int workersToCreate = Math.min(maxWorkers - workerCount, snapshotTasks.size());
```
henningandersen, line 135: I think we would max add one worker here ever? Also, using `snapshotTasks.size()` seems wrong, since the task is polled by a worker, but that worker is not free to do other work until it completes it's task (and thus we may still want a new worker even if there is only 1 task in the queue).

Also, as pointed out by Armin, we poll outside the mutex anyway.

I wonder if we could instead just start a worker every time as long as we have budget for it?

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
```diff
@@ -0,0 +1,179 @@
...
+            fileInfo.name()
+        );
+        synchronized (mutex) {
+            snapshotTasks.add(new FileSnapshotTask(context, fileInfo, fileUploadListener));
+            ensureEnoughWorkers();
+        }
+    }
+
+    private void ensureEnoughWorkers() {
+        assert Thread.holdsLock(mutex);
```
henningandersen, line 133: I wonder if this method would be simpler to reason about if it was lock-free, i.e., using an `AtomicInteger` for the worker count and just spawn one new thread if there is budget.

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
```diff
@@ -0,0 +1,179 @@
...
+        public void run() {
+            shardSnapshotter.accept(context);
+        }
+
+        public short priority() {
+            return 1;
+        }
+
+        @Override
+        public int compareTo(ShardSnapshotTask other) {
```
henningandersen, line 62: I think we should mark this method `final` since overriding is likely bug-prone.

server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
```diff
@@ -376,6 +377,8 @@ public abstract class BlobStoreRepository extends AbstractLifecycleComponent imp
      */
     private final int maxSnapshotCount;
 
+    private ShardSnapshotWorkers shardSnapshotWorkerPool;
```
henningandersen, line 380: nit: I'd prefer to name this identical to the type:

```suggestion
    private ShardSnapshotWorkers shardSnapshotWorkers;
```

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
original-brownbear: > I wonder if we could instead just start a worker every time as long as we have budget for it?

++ there isn't much value in being tricky here, just starting up to pool-size workers seems like a good idea and removes the rather hard to reason about `size()` call.

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
```diff
@@ -0,0 +1,179 @@
...
+
+    public int size() {
+        return workerCount;
+    }
+
+    private void startWorker(final String workerId) {
+        logger.debug("starting snapshot pool worker {}", workerId);
+        executor.execute(() -> {
+            try {
+                while (true) {
```
henningandersen, line 164: I wonder if this should instead respawn to the back of the queue when there is more work, allowing other work to use the snapshot thread pool.

I suggest something like:

`startWorker` takes the `ShardSnapshotTask` as argument and executes it. It checks the queue afterwards and respawns (`executor.execute`) if there is more work to do - with the new task as argument.

We only queue excess tasks. I.e., if there is budget in `workerCount`, we always call `startWorker`, otherwise we queue it. We recheck `workerCount` after queueing to catch the edge case of the last worker exiting - and spawns a new task in that case (polling out the added task from the queue).

henningandersen (commented):
One more comment. I think there could be a liveness issue with current implementation (though it might go away with more mutex or lock-free). I think this suggestion fixes it and is more fair to other ongoing work.

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
```diff
@@ -0,0 +1,179 @@
...
+            shardSnapshotter.accept(context);
+        }
+
+        public short priority() {
+            return 1;
+        }
+
+        @Override
+        public int compareTo(ShardSnapshotTask other) {
+            int res = context.snapshotId().compareTo(other.context.snapshotId());
```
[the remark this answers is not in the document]
pxsalehi: There were no more replies to my question/suggestion [here](https://github.com/elastic/elasticsearch/pull/88209#discussion_r927741572), so I went for this as a first shot. Changing it would be easy once we agree on it. (I would need a bit more code since startTime is not part of the context).

So it should be ALL meta work (across all snapshots of the same repository) first, and then among file snapshots, sort by start time? No need to sort by snapshot name/id? Would start time be a replacement for sorting by snapshot name to prevent concurrent snapshots getting interleaved?

My understanding is that it is desirable to have predictable ordering here, and also finish shard snapshots as quick as possible. If a shard snapshot is unchanged, prioratizing it, and when it comes to file snapshots, finish the relevant ones (for the same snapshot) first, so that snapshots are marked as finished one at a time and do not get interleaved.

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
pxsalehi: @henningandersen Thanks for the clarification. TBH I think we misunderstood [your original suggestion](https://github.com/elastic/elasticsearch/pull/88209#issuecomment-1177696199)! What you are suggesting here makes way more sense that what I have in the PR! :-D  [I was wondering](https://github.com/elastic/elasticsearch/pull/88209#discussion_r925487903) whether we'd need to at least not use the same number of workers as the threadpool size since the workers would keep the threadpool busy as long as there are tasks in the queue!

Putting the lock-free part aside, I've tried out a draft of your suggestion, please see `ShardSnapshotWorkers.java` again. I've done some renaming as well, I guess the name worker doesn't have much meaning anymore. It's more about controlling max concurrent snapshot tasks and the spawning order. Is that correct?

Please let me know if this is in the right direction. I'll keep addressing the rest of the comments.

> We only queue excess tasks. I.e., if there is budget in workerCount, we always call startWorker, otherwise we queue it. We recheck workerCount after queueing to catch the edge case of the last worker exiting - and spawns a new task in that case (polling out the added task from the queue).

Not sure I completely understand this part. We'd have to always queue first, so that we run the tasks in the priority we want.

henningandersen: > I think we misunderstood https://github.com/elastic/elasticsearch/pull/88209#issuecomment-1177696199

I think you got it just right, I think I educated my thinking here based on your work.

> Not sure I completely understand this part. We'd have to always queue first, so that we run the tasks in the priority we want.

I do not think this is important. It will be right most of the time if not all the time (depending on details in the code). Let me take a look at the change you did.

henningandersen: I think the version you have could work out.

Anyway, I wanted to illustrate the lock-free approach. I think this preserves order just as well as the mutex one. Notice that if there is free capacity we always execute. We only queue if there is not. Meaning the queue is empty if there is free capacity. This is not entirely true in the mutex variant (but could be modified there too).

I think I prefer the lock-free version, but I'd also be ok with the mutex version (have comments for it though).

```
public class ShardSnapshotWorkers {
    private static final Logger logger = LogManager.getLogger(ShardSnapshotWorkers.class);

    private final int maxRunningTasks;
    private final Object mutex = new Object();
    private AtomicInteger runningTasksCount = new AtomicInteger();
    private final BlockingQueue<SnapshotTask> snapshotTasks = new PriorityBlockingQueue<>();
    private final Executor executor;
    private final Consumer<SnapshotShardContext> shardSnapshotter;
    private final CheckedBiConsumer<SnapshotShardContext, FileInfo, IOException> fileSnapshotter;

    abstract class SnapshotTask implements Comparable<SnapshotTask>, Runnable {
        protected final SnapshotShardContext context;

        SnapshotTask(SnapshotShardContext context) {
            this.context = context;
        }

        public abstract short priority();

        public SnapshotShardContext context() {
            return context;
        }

        @Override
        public final int compareTo(SnapshotTask other) {
            int res = context.snapshotId().compareTo(other.context.snapshotId());
            if (res != 0) {
                return res;
            }
            return Integer.compare(priority(), other.priority());
        }
    }

    class ShardSnapshotTask extends SnapshotTask {
        ShardSnapshotTask(SnapshotShardContext context) {
            super(context);
        }

        @Override
        public void run() {
            shardSnapshotter.accept(context);
        }

        @Override
        public short priority() {
            return 1;
        }

    }

    class FileSnapshotTask extends SnapshotTask {
        private final FileInfo fileInfo;
        private final ActionListener<Void> fileUploadListener;

        FileSnapshotTask(SnapshotShardContext context, FileInfo fileInfo, ActionListener<Void> fileUploadListener) {
            super(context);
            this.fileInfo = fileInfo;
            this.fileUploadListener = fileUploadListener;
        }

        @Override
        public void run() {
            ActionRunnable.run(fileUploadListener, () -> fileSnapshotter.accept(context, fileInfo)).run();
        }

        @Override
        public short priority() {
            return 2;
        }
    }

    public ShardSnapshotWorkers(
        final int maxRunningTasks,
        final Executor executor,
        final Consumer<SnapshotShardContext> shardSnapshotter,
        final CheckedBiConsumer<SnapshotShardContext, FileInfo, IOException> fileSnapshotter
    ) {
        assert maxRunningTasks > 0;

        logger.info("starting shard snapshot worker pool of max size {}", maxRunningTasks);
        this.maxRunningTasks = maxRunningTasks;
        this.executor = executor;
        this.shardSnapshotter = shardSnapshotter;
        this.fileSnapshotter = fileSnapshotter;
    }

    public void enqueueShardSnapshot(final SnapshotShardContext context) {
        logger.trace("enqueuing shard snapshot task [snapshotID={}, indexID={}]", context.snapshotId(), context.indexId());
        runOrQueueTask(new ShardSnapshotTask(context));
    }

    public void enqueueFileSnapshot(
        final SnapshotShardContext context,
        final FileInfo fileInfo,
        final ActionListener<Void> fileUploadListener
    ) {
        logger.trace(
            "enqueuing shard snapshot file upload task [snapshotID={}, indexID={}, file={}]",
            context.snapshotId(),
            context.indexId(),
            fileInfo.name()
        );
        FileSnapshotTask task = new FileSnapshotTask(context, fileInfo, fileUploadListener);
        runOrQueueTask(task);
    }

    private void runOrQueueTask(SnapshotTask task) {
        if (incrementRunningTasks()) {
            spawnTask(task);
        } else {
            snapshotTasks.add(task);
            if (incrementRunningTasks()) {
                pollAndSpawn();
            }
        }
    }

    private void pollAndSpawn() {
        SnapshotTask polledTask = snapshotTasks.poll();
        if (polledTask != null) {
            spawnTask(polledTask);
        } else {
            int decremented = runningTasksCount.decrementAndGet();
            assert decremented >= 0;
        }
    }

    private void spawnTask(SnapshotTask task) {
        executor.execute(() -> runOneTask(task));
    }

    private boolean incrementRunningTasks() {
        return runningTasksCount.getAndUpdate(v -> v < maxRunningTasks ? v + 1 : v) < maxRunningTasks;
    }

    // for testing
    int size() {
        synchronized (mutex) {
            return runningTasksCount.get();
        }
    }

    private void runOneTask(final SnapshotTask task) {
        try {
            logger.trace("running snapshot task {}", task); // TODO: toString()?
            task.run();
        } finally {
            pollAndSpawn();
        }
    }

}
```

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
pxsalehi: Hmm. I don't really see that happening as of https://github.com/pxsalehi/elasticsearch/blob/5397f5207da55b41b33afb10046d3079833829a1/server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java. Maybe I'm missing something. In your example, since `maxWorker` is 2 and one is busy, when the second task gets queued we'd call `ensureEnoughWorkers` and we'd end up with `min(2-1, 1)` workers, so a new one will be started. (I think [this test](https://github.com/pxsalehi/elasticsearch/blob/3649729263f79e3d80d191406008e4a6e22df688/server/src/test/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkersTests.java#L157-L181) kind of covers a similar case).

server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
```diff
@@ -415,6 +418,12 @@ protected void doStart() {
         if (chunkSize != null && chunkSize.getBytes() <= 0) {
             throw new IllegalArgumentException("the chunk size cannot be negative: [" + chunkSize + "]");
         }
+        shardSnapshotWorkerPool = new ShardSnapshotWorkers(
```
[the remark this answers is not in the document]
pxsalehi: done in 7ce5c419373effbd3172f6a088c7f83d7098533a

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
```diff
@@ -0,0 +1,179 @@
...
+        final ActionListener<Void> fileUploadListener
+    ) {
+        logger.trace(
+            "enqueuing shard snapshot file upload task [snapshotID={}, indexID={}, file={}]",
+            context.snapshotId(),
+            context.indexId(),
+            fileInfo.name()
+        );
+        synchronized (mutex) {
+            snapshotTasks.add(new FileSnapshotTask(context, fileInfo, fileUploadListener));
```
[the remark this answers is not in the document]
pxsalehi: Outdated

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
```diff
@@ -0,0 +1,179 @@
...
+    private void workerDone() {
+        synchronized (mutex) {
+            assert workerCount > 0;
+
+            workerCount--;
+            ensureEnoughWorkers();
+        }
+    }
+
+    public int size() {
```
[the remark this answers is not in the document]
pxsalehi: done

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
```diff
@@ -0,0 +1,179 @@
...
+
+        ShardSnapshotTask(SnapshotShardContext context) {
+            this.context = context;
+        }
+
+        public SnapshotShardContext context() {
+            return context;
+        }
+
+        public void run() {
```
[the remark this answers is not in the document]
pxsalehi: Yeah. That sounds better. Done in https://github.com/elastic/elasticsearch/pull/88209/files#diff-a7a97862ca5689e6d801f92ba8b1255b68bc4464ede9755efbfb65685752a966

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private void ensureEnoughWorkers() {
+        assert Thread.holdsLock(mutex);
+
+        int workersToCreate = Math.min(maxWorkers - workerCount, snapshotTasks.size());
+        if (workersToCreate > 0) {
+            logger.debug("starting {} shard snapshot workers", workersToCreate);
+        }
+        while (workersToCreate > 0) {
```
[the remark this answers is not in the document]
pxsalehi: outdated

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
```diff
@@ -0,0 +1,179 @@
...
+public class ShardSnapshotWorkers {
+    private static final Logger logger = LogManager.getLogger(ShardSnapshotWorkers.class);
+
+    private final int maxWorkers;
+    private final Object mutex = new Object();
+    private final BlockingQueue<ShardSnapshotTask> snapshotTasks = new PriorityBlockingQueue<>();
+    private final Executor executor;
+    private final Consumer<SnapshotShardContext> shardSnapshotter;
+    private final CheckedBiConsumer<SnapshotShardContext, FileInfo, IOException> fileSnapshotter;
+    private volatile int workerCount = 0;
```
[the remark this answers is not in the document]
pxsalehi: outdated

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkers.java
pxsalehi: Done!

pxsalehi:
@henningandersen @original-brownbear Thank you for the reviews. Please have another look. I've addressed all comments.

I haven't added that case in Henning's code sample where we spawn directly if workers are available. IMO, it is an optimization for a very specific corner case which I don't know if we need. Always queueing and polling seems simpler (unless it has some disadvantage I'm not aware of).

docs/changelog/88209.yaml
```diff
@@ -0,0 +1,6 @@
+pr: 88209
+summary: Parallelize unchanged shard snapshots and limit number of parallel file snapshots
```
pxsalehi, line 2: I guess this is not clear enough now. I'll change it.

pxsalehi: Done!

test/framework/src/main/java/org/elasticsearch/index/shard/IndexShardTestCase.java
```diff
@@ -1010,6 +1010,7 @@ protected ShardGeneration snapshotShard(final IndexShard shard, final Snapshot s
                     snapshotStatus,
                     Version.CURRENT,
                     Collections.emptyMap(),
+                    threadPool.absoluteTimeInMillis(),
```
henningandersen, line 1013: Also randomize here instead? I would think we do not rely on the specific time here.

x-pack/plugin/core/src/test/java/org/elasticsearch/snapshots/sourceonly/SourceOnlySnapshotShardTests.java
```diff
@@ -134,6 +134,7 @@ public void testSourceIncomplete() throws IOException {
                         indexShardSnapshotStatus,
                         Version.CURRENT,
                         Collections.emptyMap(),
+                        threadPool.absoluteTimeInMillis(),
```
henningandersen, line 137: Randomize instead?

x-pack/plugin/core/src/test/java/org/elasticsearch/snapshots/sourceonly/SourceOnlySnapshotShardTests.java
```diff
@@ -207,6 +209,7 @@ public void testIncrementalSnapshot() throws IOException {
...
```
henningandersen, line 212: Randomize instead?

x-pack/plugin/core/src/test/java/org/elasticsearch/snapshots/sourceonly/SourceOnlySnapshotShardTests.java
```diff
@@ -238,6 +241,7 @@ public void testIncrementalSnapshot() throws IOException {
...
```
henningandersen, line 244: Randomize instead?

x-pack/plugin/core/src/test/java/org/elasticsearch/snapshots/sourceonly/SourceOnlySnapshotShardTests.java
```diff
@@ -299,6 +303,7 @@ public void testRestoreMinmal() throws IOException {
...
```
henningandersen, line 306: Randomize instead?

x-pack/plugin/searchable-snapshots/src/test/java/org/elasticsearch/xpack/searchablesnapshots/store/SearchableSnapshotDirectoryTests.java
```diff
@@ -632,6 +632,7 @@ protected void assertSnapshotOrGenericThread() {
                             snapshotStatus,
                             Version.CURRENT,
                             emptyMap(),
+                            threadPool.absoluteTimeInMillis(),
```
henningandersen, line 635: Randomize instead?

henningandersen (commented):
This direction looks good.

I'd like to see a test case that stresses the multi-threading aspects of the `ShardSnapshotWorkers` class. I have wondered if we could refactor it into something capable of running a fixed number of tasks on a thread-pool, i.e., not know about snapshots. I think that could be useful elsewhere. I mention it mostly in case that makes the multi-threading test easier, that can also be refactored as a follow-up for when we need a similar functionality next time.

pxsalehi:
@elasticmachine test elasticsearch-ci/part-1
(Test failure doesn't seem to be related to the PR: https://github.com/elastic/elasticsearch/issues/89623)

pxsalehi:
@elasticmachine please test elasticsearch-ci/part-1

pxsalehi:

test/framework/src/main/java/org/elasticsearch/index/shard/IndexShardTestCase.java
pxsalehi: I don't think the test relies on the time. I've changed it to random, here and also all the following ones.

server/src/test/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkersTests.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.lessThan;
+
+public class ShardSnapshotWorkersTests extends ESTestCase {
+
+    private static ThreadPool threadPool;
+    private static Executor executor;
```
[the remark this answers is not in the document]
pxsalehi: done

server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
```diff
@@ -3097,10 +3084,10 @@ void ensureNotClosing(final Store store) throws AlreadyClosedException {
 
     private static ActionListener<Void> fileQueueListener(
         BlockingQueue<BlobStoreIndexShardSnapshot.FileInfo> files,
-        int workers,
+        int groupSize,
```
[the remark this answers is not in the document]
pxsalehi: done

server/src/test/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkersTests.java
```diff
@@ -0,0 +1,213 @@
...
+        // Adding at least one new shard snapshot would create a new worker
+        int newTasks = randomIntBetween(1, 10);
+        for (int i = 0; i < newTasks; i++) {
+            workers.enqueueShardSnapshot(dummyContext());
+        }
+        enqueuedSnapshots += newTasks;
+        assertThat(workers.size(), equalTo(maxSize));
+        snapshotBlocker.countDown();
+        // Eventually all workers exit
+        assertBusy(() -> assertThat(workers.size(), equalTo(0)));
```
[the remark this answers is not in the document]
pxsalehi: I've done this for `ThrottledTaskRunnerTests` since the code is a bit differently structured now.

server/src/test/java/org/elasticsearch/repositories/blobstore/ShardSnapshotWorkersTests.java
```diff
@@ -0,0 +1,213 @@
...
+        int maxSize = randomIntBetween(1, threadPool.info(ThreadPool.Names.SNAPSHOT).getMax());
+        CountDownLatch snapshotBlocker = new CountDownLatch(1);
+        MockedRepo repo = new MockedRepo(snapshotBlocker);
+        ShardSnapshotWorkers workers = new ShardSnapshotWorkers(maxSize, executor, repo::snapshotShard, repo::snapshotFile);
+        repo.setWorkers(workers);
+        int enqueuedSnapshots = maxSize - 1; // So that it is possible to create at least one more worker
+        for (int i = 0; i < enqueuedSnapshots; i++) {
+            workers.enqueueShardSnapshot(dummyContext());
+        }
+        assertBusy(() -> assertThat(workers.size(), equalTo(maxSize - 1)));
```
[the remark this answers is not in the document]
pxsalehi: Done

server/src/main/java/org/elasticsearch/common/util/concurrent/ThrottledTaskRunner.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    private boolean incrementRunningTasks() {
+        int preUpdateValue = runningTasks.getAndUpdate(v -> v < maxRunningTasks ? v + 1 : v);
+        assert preUpdateValue <= maxRunningTasks;
+        return preUpdateValue < maxRunningTasks;
+    }
+
+    public int runningTasks() {
```
pxsalehi, line 63: Technically, this would return the number of the running task runners! since whenever a task is finished, we call `pollAndSpawn` again, we could have a task runner that just exists since it finds no tasks to run. But I don't think this would worth adding another counter. Do we care about that distinction?

pxsalehi (commented):
@henningandersen thanks for the suggestions. This is ready for review now. As suggested, I have refactored some parts to a more generic `ThrottledTaskRunner`, added more tests for the multi-threading aspects, and covered the corner case you mentioned for the lock-free approach.

server/src/test/java/org/elasticsearch/repositories/blobstore/ShardSnapshotTaskRunnerTests.java
```diff
@@ -0,0 +1,195 @@
...
+        int enqueuedSnapshots = randomIntBetween(maxTasks * 2, maxTasks * 10);
+        for (int i = 0; i < enqueuedSnapshots; i++) {
+            threadPool.generic().execute(() -> taskRunner.enqueueShardSnapshot(dummyContext()));
+        }
+        // Eventually all snapshots are finished
+        assertBusy(() -> {
+            assertThat(repo.finishedShardSnapshots(), equalTo(enqueuedSnapshots));
+            assertThat(taskRunner.runningTasks(), equalTo(0));
+        });
+        assertThat(taskRunner.queueSize(), equalTo(0));
```
henningandersen, line 160: I wonder if there is a risk of the `assertBusy` exiting too early, since `runningTasks` could in theory be 0 with many more queued items. Suppose all your runners has just done the decrement and now wants to do the increment. At that time, the `runningTasks()` could be 0?

You will likely have to include either the queue size or the file-snapshot task check in the assert busy. If you include queue-size it must be checked before the `runningTasks` check.

server/src/test/java/org/elasticsearch/common/util/concurrent/ThrottledTaskRunnerTests.java
```diff
@@ -0,0 +1,123 @@
...
+
+        private final Runnable runnable;
+
+        TestTask(Runnable runnable) {
+            this.runnable = runnable;
+        }
+
+        @Override
+        public int compareTo(TestTask o) {
+            return randomIntBetween(0, 1000);
```
henningandersen, line 50: I'd think this should be between (-1000, 1000).

But I can imagine `PrioritizedBlockingQueue` not working with such a compare. I think it expects objects to adhere to the `compare` contract, which requires stability and transitivity.

Can we assign a random number to the task and compare the tasks using that number instead?

Can we also validate that we execute tasks in order?

henningandersen (commented):
Thanks for extracting the logic, looks great and will definitely be useful elsewhere.

I left more comments. I'd like validation of the order of execution of tasks in `ThrottledTaskRunner` too.

pxsalehi:
@elasticmachine update branch

pxsalehi:
I looked into the [previous failures](https://gradle-enterprise.elastic.co/s/lpiy4ql6ecvxc/tests/:server:internalClusterTest/org.elasticsearch.snapshots.CloneSnapshotIT/testRemoveFailedCloneFromCSWithQueuedSnapshotInProgress?top-execution=1), since they were both related to snapshotting. From the previous test failures, it seems `ConcurrentSnapshotsIT.testMasterFailoverOnFinalizationLoop` was failing due to recent changes on main, and is now fixed on main and doesn't fail here anymore. However, `CloneSnapshotIT.testRemoveFailedCloneFromCSWithQueuedSnapshotInProgress` seems to rarely fail on this PR. Not sure why! I can make it time out on a slow system, but can't tell whether the test assumption doesn't match anymore with how we schedule the snapshotting tasks, or there is an issue in the PR. It seems that the snapshot cloning step of the test doesn't return and times out. I'd have to spend some more time on it. 

CC: @original-brownbear maybe we could look into it together.

server/src/test/java/org/elasticsearch/repositories/blobstore/ShardSnapshotTaskRunnerTests.java
```diff
@@ -0,0 +1,195 @@
...
+
+        public void setTaskRunner(ShardSnapshotTaskRunner taskRunner) {
+            this.taskRunner = taskRunner;
+        }
+
+        public void snapshotShard(SnapshotShardContext context) {
+            int filesToUpload = randomIntBetween(0, 10);
+            if (filesToUpload == 0) {
+                finishedShardSnapshots.incrementAndGet();
+            } else {
```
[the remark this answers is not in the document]
pxsalehi: No, I think it's different. When there is no file to upload, we mark the snapshot as done/finished. The increment at the end of the if/else is for the task, not the snapshot.

server/src/test/java/org/elasticsearch/common/util/concurrent/ThrottledTaskRunnerTests.java
```diff
@@ -0,0 +1,123 @@
...
+                try {
+                    taskBlocker.await();
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+                executed.incrementAndGet();
+            }));
+            assertThat(taskRunner.runningTasks(), equalTo(maxTasks));
+        }
+        taskBlocker.countDown();
```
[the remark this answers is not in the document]
pxsalehi: Why? as the comment suggests, the point is to make sure we don't go over maxTasks. `newTasks -1` is just a random value between 0 and 9.

server/src/test/java/org/elasticsearch/repositories/blobstore/ShardSnapshotTaskRunnerTests.java
pxsalehi: Maybe there is a misunderstanding but `finishedShardSnapshots` means all file/meta tasks are finished, not just the meta work. I think the assertBusy there waits until all tasks (of both type) are finished.

server/src/test/java/org/elasticsearch/repositories/blobstore/ShardSnapshotTaskRunnerTests.java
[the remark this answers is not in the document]
henningandersen: Thanks, I missed that.

server/src/test/java/org/elasticsearch/common/util/concurrent/ThrottledTaskRunnerTests.java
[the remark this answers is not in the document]
henningandersen: Yes, we expect one task to execute and the rest to be queued. I see no reason not to validate that they are in the queue?

Just like the validation at the end that the queue is now empty.

server/src/main/java/org/elasticsearch/common/util/concurrent/ThrottledTaskRunner.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Executor;
+import java.util.concurrent.PriorityBlockingQueue;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * ThrottledTaskRunner performs the enqueued tasks in the order dictated by the natural ordering
+ * of the tasks, limiting the max number of concurrently running tasks.
+ */
+public class ThrottledTaskRunner<T extends Comparable<T> & Runnable> {
```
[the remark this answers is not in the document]
pxsalehi: Thanks for the suggestion. Done in 3a4a28eca0d3b264eda25da74390a96b11b7d648

server/src/main/java/org/elasticsearch/common/util/concurrent/ThrottledTaskRunner.java
pxsalehi: done a4ef329d46fb4383a181bb56434febdfab2a33ec

server/src/main/java/org/elasticsearch/common/util/concurrent/ThrottledTaskRunner.java
```diff
@@ -0,0 +1,81 @@
...
+        int preUpdateValue = runningTasks.getAndUpdate(v -> v < maxRunningTasks ? v + 1 : v);
+        assert preUpdateValue <= maxRunningTasks;
+        return preUpdateValue < maxRunningTasks;
+    }
+
+    public int runningTasks() {
+        return runningTasks.get();
+    }
+
+    public int queueSize() {
```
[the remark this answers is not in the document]
pxsalehi: doen a4ef329d46fb4383a181bb56434febdfab2a33ec

server/src/test/java/org/elasticsearch/common/util/concurrent/ThrottledTaskRunnerTests.java
```diff
@@ -0,0 +1,123 @@
...
+            return randomIntBetween(0, 1000);
+        }
+
+        @Override
+        public void run() {
+            runnable.run();
+        }
+    }
+
+    public void testThrottledTaskRunner() throws Exception {
```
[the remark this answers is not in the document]
pxsalehi: done 865ceca6e6f0a7021b9db7439071d0fb7c9370cd

server/src/test/java/org/elasticsearch/common/util/concurrent/ThrottledTaskRunnerTests.java
```diff
@@ -0,0 +1,123 @@
...
+        // Eventually all tasks are executed
+        assertBusy(() -> {
+            assertThat(executed.get(), equalTo(enqueued));
+            assertThat(taskRunner.runningTasks(), equalTo(0));
+        });
+        assertThat(taskRunner.queueSize(), equalTo(0));
+    }
+
+    public void testEnqueueSpawnsNewTasksUpToMax() throws Exception {
+        int maxTasks = randomIntBetween(1, threadPool.info(ThreadPool.Names.SNAPSHOT).getMax());
```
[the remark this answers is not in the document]
pxsalehi: I changed it in the whole file to GENERIC. I guess I forgot to change it once taking it out of the snapshot context.

server/src/test/java/org/elasticsearch/common/util/concurrent/ThrottledTaskRunnerTests.java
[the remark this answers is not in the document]
pxsalehi: ok, now I know what you mean! :-) 79331f810c3a4f21152d57ad3d99a0356fdbf5af

server/src/test/java/org/elasticsearch/common/util/concurrent/ThrottledTaskRunnerTests.java
pxsalehi: done in 56619c0eb675ec0401f7ddf73b7a1950bf03677c
I added a new test for the ordering in 79331f810c3a4f21152d57ad3d99a0356fdbf5af

pxsalehi:
@henningandersen Thanks for the detailed feedback. I addressed all your comments. Please have another look. Meanwhile I'll look into that previous CI failure to see if it is related or I was just paranoid!

pxsalehi:
> @henningandersen Thanks for the detailed feedback. I addressed all your comments. Please have another look. Meanwhile I'll look into that previous CI failure to see if it is related or I was just paranoid!

I am not able to reproduce that issue on the branch. For the record, it used to timeout on different asserts in the test, the couple of times that it happened (not just a specific one), and that was on a pretty slow system. I think, it is safe to dismiss that.

server/src/test/java/org/elasticsearch/common/util/concurrent/PrioritizedThrottledTaskRunnerTests.java
```diff
@@ -0,0 +1,177 @@
...
+        private final int priority;
+
+        TestTask(Runnable runnable, int priority) {
+            this.runnable = runnable;
+            this.priority = priority;
+        }
+
+        @Override
+        public int compareTo(TestTask o) {
+            return Integer.compare(priority, o.priority);
```
henningandersen, line 55: Can we return `priority - o.priority` instead? `Integer.compare`normalizes to `-1, 0, 1`, thus we do not test that any other values has the right meaning.

henningandersen (commented):
This looks good to me now, but I'd like @original-brownbear to do the final review on this.

server/src/test/java/org/elasticsearch/common/util/concurrent/PrioritizedThrottledTaskRunnerTests.java
pxsalehi: @henningandersen the `compareTo` is only used by the priority queue, which doesn't care for the actual distance between the two priorities, I think! So I don't understand why does this matter? Could you please elaborate why is this important?

server/src/main/java/org/elasticsearch/common/util/concurrent/PrioritizedThrottledTaskRunner.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public void enqueueTask(final T task) {
+        logger.trace("[{}] enqueuing task {}", taskRunnerName, task);
+        tasks.add(task);
+        pollAndSpawn();
+    }
+
+    // visible for testing
+    protected void pollAndSpawn() {
```
original-brownbear, line 46: Can we add some commentary on how this loop works. In particular, on why we need to peek the queue. This is somewhat hard to follow for me and will be even harder to follow for future readers of this code.

server/src/main/java/org/elasticsearch/repositories/blobstore/ShardSnapshotTaskRunner.java
```diff
@@ -0,0 +1,140 @@
...
+    private final CheckedBiConsumer<SnapshotShardContext, FileInfo, IOException> fileSnapshotter;
+
+    abstract static class SnapshotTask implements Comparable<SnapshotTask>, Runnable {
+        protected final SnapshotShardContext context;
+
+        SnapshotTask(SnapshotShardContext context) {
+            this.context = context;
+        }
+
+        public abstract short priority();
```
original-brownbear, line 40: Should we just use `int` here. `short` is kind of pointless isn't it, might even be a net negative for comparison performance?

original-brownbear (commented):
This looks good to me, just one point where I asked for some docs because I'm having trouble understanding the code and one rather trivial detail and this should be good to go :)

server/src/main/java/org/elasticsearch/common/util/concurrent/PrioritizedThrottledTaskRunner.java
pxsalehi: Good point!

pxsalehi:
@original-brownbear @henningandersen All done! Please check again.

server/src/test/java/org/elasticsearch/common/util/concurrent/PrioritizedThrottledTaskRunnerTests.java
henningandersen: This is part of a test verifying that `PrioritizedThrottledTaskRunner` executes tasks in the right order when presented with *any* comparable object. But the test here is restricted to only verify using a sub-set of such comparable objects that use "normalized" return values in `compareTo`.

Arguing in terms of the implementation seems invalid, the purpose of the tests is to demonstrate that the implementation works under as many circumstances as possible. Given the simple change to demonstrate this under more circumstances here, I think we should make the change.

pxsalehi: I've done the change already. Thanks for the explanation! :)

server/src/test/java/org/elasticsearch/common/util/concurrent/PrioritizedThrottledTaskRunnerTests.java
```diff
@@ -0,0 +1,177 @@
...
+    public void testMultiThreadedEnqueue() throws Exception {
+        final int maxTasks = randomIntBetween(1, threadPool.info(ThreadPool.Names.GENERIC).getMax());
+        PrioritizedThrottledTaskRunner<TestTask> taskRunner = new PrioritizedThrottledTaskRunner<>("test", maxTasks, executor);
+        final int enqueued = randomIntBetween(2 * maxTasks, 10 * maxTasks);
+        AtomicInteger executed = new AtomicInteger();
+        CountDownLatch threadBlocker = new CountDownLatch(enqueued);
+        for (int i = 0; i < enqueued; i++) {
+            new Thread(() -> {
+                try {
+                    threadBlocker.countDown();
```
original-brownbear, line 73: NIT: could use `CyclicBarrier` here instead which seems to be the correct primitive?

original-brownbear (approved):
LGTM, I'm good with this one now as is. Performance seems alright too, gave it a quick test run benchmarking. Unless @henningandersen has anything open? I think we're good to go here.

pxsalehi:
Thanks Henning and Armin!

server/src/test/java/org/elasticsearch/common/util/concurrent/PrioritizedThrottledTaskRunnerTests.java
pxsalehi: We could. But since we need only a one-time barrier, the latch is enough I think. I've seen it used like that in several places in the code.

henningandersen: I'd agree with Armin here, `CyclicBarrier` is good for a rendezvous interaction. And saves a line of code too.

But fine to leave as is...

PR #88210, opened 2022-06-30
Save loop over all local shards in IndicesClusterService.applyClusterState

original-brownbear:
Similar to #88184 We can save another two loops here by checking for shards to fail in the same loop that updates or creates shards.
Also, we only need to loop over all indices services locally once for deleting indices as a whole or just shards
out of existing indices.

relates #77466


original-brownbear:

DaveCTurner (commented):
I'm wary of reordering the steps like this - there are certain requirements for having these things in the order they're in. Before I dive too deeply, can you quantify the benefits of this change?

original-brownbear:
> I'm wary of reordering the steps like this - there are certain requirements for having these things in the order they're in.

I took that into account when making these changes, I don't think anything changes logically here. IMO things are clearer after this change since it's more directly visible what state a shard/index will end up in after the loop from a single block of logic than before isn't it?

> can you quantify the benefits of this change?

We're reducing the factor on one of the largest factor O(N) (N ... number of shards/indices) here so the benefit will become bigger the larger the index count. It's around a 1-2% time improvement end-to-end when bootstrapping a 50k indices cluster and (even though I admittedly didn't test this case yet but it's obvious from profiling) more than that when testing larger cases.

original-brownbear:
@DaveCTurner @tlrx ping I think this one is still quite worthwhile, maybe even more so now than a month ago with other improvements making this relatively more important :)

server/src/main/java/org/elasticsearch/indices/cluster/IndicesClusterStateService.java
```diff
@@ -481,8 +440,19 @@ private void createIndicesAndUpdateShards(final ClusterState state) {
         final Map<Index, List<ShardRouting>> indicesToCreate = new HashMap<>();
         for (ShardRouting shardRouting : localRoutingNode) {
             if (failedShardsCache.containsKey(shardRouting.shardId()) == false) {
+                ShardId shardId = shardRouting.shardId();
```
DaveCTurner, line 443: nit:

```suggestion
            final ShardId shardId = shardRouting.shardId();
            if (failedShardsCache.containsKey(shardId) == false) {
```

server/src/main/java/org/elasticsearch/indices/cluster/IndicesClusterStateService.java
```diff
@@ -481,8 +440,19 @@ private void createIndicesAndUpdateShards(final ClusterState state) {
...
+                if (shardRouting.initializing() == false && (indexService == null || indexService.getShardOrNull(shardId.id()) == null)) {
+                    // the master thinks we are active, but we don't have this shard at all, mark it as failed
+                    sendFailShard(
+                        shardRouting,
+                        "master marked shard as active, but shard has not been created, mark shard as failed",
+                        null,
+                        state
+                    );
+                    continue;
+                }
```
DaveCTurner, line 455: nit: bit odd to `continue` past an `if/else`...

```suggestion
                } else
```

server/src/main/java/org/elasticsearch/indices/cluster/IndicesClusterStateService.java
```diff
@@ -390,78 +387,40 @@ private void removeIndices(final ClusterChangedEvent event) {
...
+                            currentRoutingEntry,
+                            newShardRouting
+                        );
+                        indexService.removeShard(shardId.id(), "removing shard (stale copy)");
+                    } else if (newShardRouting.initializing() && currentRoutingEntry.active()) {
+                        // this can happen if the node was isolated/gc-ed, rejoins the cluster and a new shard with the same allocation id
+                        // is assigned to it. Batch cluster state processing or if shard fetching completes before the node gets a new
+                        // cluster
+                        // state may result in a new shard being initialized while having the same allocation id as the currently started
+                        // shard.
```
DaveCTurner, line 414: nit: formatting

```suggestion
                        // this can happen if the node was isolated/gc-ed, rejoins the cluster and a new shard with the same allocation id
                        // is assigned to it. Batch cluster state processing or if shard fetching completes before the node gets a new
                        // cluster state may result in a new shard being initialized while having the same allocation id as the currently
                        // started shard.
```

server/src/main/java/org/elasticsearch/indices/cluster/IndicesClusterStateService.java
```diff
@@ -390,78 +387,40 @@ private void removeIndices(final ClusterChangedEvent event) {
             if (reason != null) {
                 logger.debug("{} removing index ({})", index, reason);
                 indicesService.removeIndex(index, reason, "removing index (" + reason + ")");
-            }
-        }
-    }
-
-    /**
-     * Notifies master about shards that don't exist but are supposed to be active on this node.
```
DaveCTurner, line 398: `createIndicesAndUpdateShards` deserves javadoc (incorporating this sentence)

DaveCTurner (commented):
Looks good, IntelliJ does a better job than Github when it comes to ignoring the whitespace/formatting changes. I left only nits.

original-brownbear:
Thanks David, all suggestions applied :)

original-brownbear:
@elasticmachine update branch

DaveCTurner (approved):
LGTM

original-brownbear:
Thanks David!

PR #88211, opened 2022-06-30
Add 'mode' option to `_source` field mapper

romseygeek:
Currently we have two parameters that control how the source of a document
is stored, `enabled` and `synthetic`, both booleans.  However, there are only
three possible combinations of these, with `enabled:false` and `synthetic:true`
being disallowed.  To make this easier to reason about, this commit replaces
the `enabled` parameter with a new `mode` parameter, which can take the values
`stored`, `synthetic` and `disabled`.  The `mode` parameter cannot be set
in combination with `enabled`, and we will subsequently move towards 
deprecating `enabled` entirely.


romseygeek:
This is still a draft, as I want to get CI passing before I start looking at documentation etc.

romseygeek:

romseygeek:

server/src/test/java/org/elasticsearch/index/mapper/SourceFieldMapperTests.java
```diff
@@ -42,9 +42,25 @@ protected void registerParameters(ParameterChecker checker) throws IOException {
...
+            topMapping(b -> b.startObject(SourceFieldMapper.NAME).field("mode", "synthetic").endObject()),
+            dm -> assertTrue(dm.metadataMapper(SourceFieldMapper.class).isSynthetic())
+        );
+        /*
+        checker.registerUpdateCheck(
+            topMapping(b -> b.startObject(SourceFieldMapper.NAME).field("enabled", false).endObject()),
+            topMapping(b -> b.startObject(SourceFieldMapper.NAME).field("mode", "synthetic").endObject()),
+            dm -> assertTrue(dm.metadataMapper(SourceFieldMapper.class).isSynthetic())
+        );
+         */
```
javanna, line 56: this throws exception right? did you mean to test that behaviour?

server/src/test/java/org/elasticsearch/index/mapper/SourceFieldMapperTests.java
```diff
@@ -172,13 +188,38 @@ public void testSourceObjectContainsExtraTokens() throws Exception {
...
+        assertTrue(mapper.enabled());
+        assertTrue(mapper.isSynthetic());
+
+        ParsedDocument doc = mapperService.documentMapper().parse(source("{}"));
+        assertNull(doc.rootDoc().get(SourceFieldMapper.NAME));
+
+        Exception e = expectThrows(IllegalArgumentException.class, () -> merge(mapperService, """
+            { "_doc" : { "_source" : { "mode" : "stored" } } }
+            """));
+        assertThat(e.getMessage(), containsString("Cannot update parameter [mode] from [synthetic] to [stored]"));
```
javanna, line 223: do we also need to test other updates like going to DISABLED or from STORED to SYNTHETIC?

server/src/main/java/org/elasticsearch/index/mapper/SourceFieldMapper.java
```diff
@@ -36,17 +38,23 @@ public class SourceFieldMapper extends MetadataFieldMapper {
     public static final String CONTENT_TYPE = "_source";
     private final XContentFieldFilter filter;
 
+    /** The source mode */
+    public enum Mode {
```
javanna, line 42: does it need to be public considering that consumers rely mostly on the existing enabled() method?

javanna (commented):
This looks reasonable to me, I left a couple of questions. Would it be possible to update the description of the PR including the reasoning behind this change? Looks like we are thinking about possible future options, but then I wonder what we would call a synthetic option that automatically adjusts some of the field types to be "synthetic source compliant". Could you raise awareness within the search team about this change / the way forward and gather feedback about it? Is it also the plan to one day remove the enabled:true|false ?

server/src/test/java/org/elasticsearch/index/mapper/SourceFieldMapperTests.java
romseygeek: I was trying to see if I could get this update to work, but it isn't possible with the way that merging and validation is implemented at present.  I'll just remove the commented-out lines.

server/src/test/java/org/elasticsearch/index/mapper/SourceFieldMapperTests.java
romseygeek: There's already a test that going from STORED to SYNTHETIC works.  I'll add a check that you can always go to DISABLED.

server/src/main/java/org/elasticsearch/index/mapper/SourceFieldMapper.java
romseygeek: ++, will make it private

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/indices.put_mapping/10_basic.yml
```diff
@@ -205,15 +204,14 @@
         id:      1
         refresh: true
         body:
-          kwd: foo
+          value: 4
```
ywelsch, line 207: what required this change?

server/src/main/java/org/elasticsearch/index/mapper/SourceFieldMapper.java
```diff
@@ -140,32 +167,47 @@ public Query termQuery(Object value, SearchExecutionContext context) {
         }
     }
 
-    private final boolean enabled;
+    private final Mode mode;
```
ywelsch, line 170: mark this one as `@Nullable`? Might not be immediately clear why it can be null (bwc reasons).

ywelsch (approved):
I've evaluated the change from a technical perspective (looks good to me). I'm unsure how this fits into the overall TSDB story and whether it makes sense to expose this capability as a standalone feature (given the current limitations that come with synthetic source).

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/indices.put_mapping/10_basic.yml
romseygeek: `kwd: foo` would dynamically create a text field + keyword field with `ignore_above` set, which is a combination that does not currently work with synthetic source (you need to have no `ignore_above` limit on the keyword subfield for it to be accepted).

PR #88214, opened 2022-06-30
[ML] Adds basic notifications for trained model deployments

dimitris-athanasiou:
For specific models:
  - deployment started
  - deployment stopped

System notifications when rebalance occurrs with reasons:
  - model deployment started
  - model deployment stopped
  - nodes changed


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/common/notifications/AbstractAuditMessage.java
```diff
@@ -154,6 +157,7 @@ public boolean equals(Object obj) {
     /**
      * @return resource id field name used when storing a new message
      */
+    @Nullable
     protected abstract String getResourceField();
```
benwtrent, line 161: There are many things that assume this is not nullable 
```
if (resourceId != null) {
            builder.field(getResourceField(), resourceId);
        }
```
In the builder for one. 

Could you change the abstraction here to be `Optional<String>` to ensure we handle all possible NPEs?

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/notifications/SystemAuditor.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class SystemAuditor extends AbstractMlAuditor<SystemAuditMessage> {
+
+    public SystemAuditor(Client client, ClusterService clusterService) {
+        super(
+            client,
+            (resourceId, message, level, timestamp, nodeName) -> new SystemAuditMessage(message, level, timestamp, nodeName),
+            clusterService
+        );
+    }
```
benwtrent, line 22: I think this should override all the audit message levels like
```
public void info(String message) {
}
public void info(String resourceId, String message) {
info(message); // Or throw unsupported?
    }
```

This way we enforce the resourceId is always passed as null and don't rely on the caller knowing this (I see mistakes in the future resulting in weird NPEs or something).

dimitris-athanasiou:
@elasticmachine update branch

x-pack/plugin/ml/qa/native-multi-node-tests/src/javaRestTest/java/org/elasticsearch/xpack/ml/integration/PyTorchModelIT.java
```diff
@@ -725,6 +725,27 @@ public void testStoppingDeploymentShouldTriggerRebalance() throws Exception {
...
+        putModelDefinition(modelId2);
+        putVocabulary(List.of("these", "are", "my", "words"), modelId2);
+        startDeployment(modelId2);
+
+        stopDeployment(modelId1);
+        stopDeployment(modelId2);
+
+        assertNotificationsContain(modelId1, "Started deployment", "Stopped deployment");
+        assertNotificationsContain(modelId2, "Started deployment", "Stopped deployment");
+        assertSystemNotificationsContain("Rebalanced trained model allocations because [model deployment started]");
```
davidkyle, line 746: Shouldn't there also be a `Rebalanced trained model allocations because [model deployment STOPPED]` message can you test for that.

davidkyle (approved):
LGTM

~Will you add system audit messages for node changes?~

Sorry I missed that the message was already there

x-pack/plugin/ml/qa/native-multi-node-tests/src/javaRestTest/java/org/elasticsearch/xpack/ml/integration/PyTorchModelIT.java
dimitris-athanasiou: No, because the remaining model is satisfied when the other gets stopped. We only proceed with the rebalance if there are unsatisfied models after a deployment is stopped.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -414,7 +415,12 @@ public ClusterState execute(ClusterState currentState) {
 
                     if (areClusterStatesCompatibleForRebalance(clusterState, currentState)) {
                         isUpdated = true;
-                        return update(currentState, rebalancedMetadata);
+                        ClusterState updatedState = update(currentState, rebalancedMetadata);
+                        if (TrainedModelAssignmentMetadata.fromState(currentState)
```
davidkyle, line 419: Can the `update()` method return a `Tuple<Boolean, ClusterState>` where the boolean indicates a change. It saves extracting the metadata from cluster state again and the `equals()` check

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/TrainedModelAssignmentClusterService.java
```diff
@@ -414,7 +415,12 @@ public ClusterState execute(ClusterState currentState) {
...
+                            .equals(TrainedModelAssignmentMetadata.fromState(updatedState)) == false) {
+                            systemAuditor.info(Messages.getMessage(Messages.INFERENCE_DEPLOYMENT_REBALANCED, reason));
```
davidkyle, line 421: It's better to be safe and not do the auditing inside the `ClusterStateUpdateTask::execute` when it could be done in `clusterStateProcessed` using the ML utility thread pool. 

The auditor can potentially do a lot of work if it is the first time a message is audited

https://github.com/elastic/elasticsearch/blob/898d84998b081d47ec6a9c2305f9bf4486d69be3/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/common/notifications/AbstractAuditor.java#L122

davidkyle (approved):
LGTM2

PR #88215, opened 2022-06-30
[Transform] Make default transform scheduler frequency 1s again

przemekwitek:
This PR partially reverts https://github.com/elastic/elasticsearch/pull/88153. As the effect the default scheduler frequency is made to be `1s` again and this is most likely a correct value.
Additionally, this PR fixes the placeholders in `Strings.format`.

Relates https://github.com/elastic/elasticsearch/issues/88063


hendrikmuhs (approved):
LGTM

PR #88217, opened 2022-06-30
Add 'filter' to 'knn' search section

jtibshirani:
This PR introduces a 'filter' subsection where filters can be provided to the
kNN search:

```
POST index/_search
{
  "knn": {
    "field": "image_vector",
    "query_vector": [0.3f, 0.1f, ...],
    "k": 10,
    "num_candidates": 100,
    "filter": {
      "term": {
       "file_type": "jpeg"
      }
    }
  }
}
```

Like all other queries, the filters are rewritten on the coordinating nodes as
well as data nodes.

Addresses #87625.

### DIFF
--- a/server/src/main/java/org/elasticsearch/search/vectors/KnnSearchBuilder.java
+++ b/server/src/main/java/org/elasticsearch/search/vectors/KnnSearchBuilder.java
@@ -173,7 +173,7 @@ public class KnnSearchBuilder implements Writeable, ToXContentFragment, Rewritea
 
     @Override
     public int hashCode() {
-        return Objects.hash(field, k, numCands, Arrays.hashCode(queryVector), boost);
+        return Objects.hash(field, k, numCands, Arrays.hashCode(queryVector), Objects.hashCode(filterQueries), boost);
     }
 
     @Override
@@ -204,10 +204,7 @@ public class KnnSearchBuilder implements Writeable, ToXContentFragment, Rewritea
         out.writeVInt(k);
         out.writeVInt(numCands);
         out.writeFloatArray(queryVector);
-        out.writeVInt(filterQueries.size());
-        for (QueryBuilder query : filterQueries) {
-            out.writeNamedWriteable(query);
-        }
+        out.writeNamedWriteableList(filterQueries);
         out.writeFloat(boost);
     }
 }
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/elasticsearch/search/vectors/KnnSearchBuilder.java
```diff
@@ -123,6 +167,7 @@ public boolean equals(Object o) {
             && numCands == that.numCands
             && Objects.equals(field, that.field)
             && Arrays.equals(queryVector, that.queryVector)
+            && Objects.equals(filterQueries, that.filterQueries)
```
mayya-sharipova, line 170: Should we add `filterQueries` to `hashCode()` as well?

server/src/main/java/org/elasticsearch/search/vectors/KnnSearchBuilder.java
```diff
@@ -151,6 +204,10 @@ public void writeTo(StreamOutput out) throws IOException {
         out.writeVInt(k);
         out.writeVInt(numCands);
         out.writeFloatArray(queryVector);
+        out.writeVInt(filterQueries.size());
```
mayya-sharipova, line 207: Should we write filterQueries the same way we read them:
`out.writeNamedWriteableList(filterQueries);`?

mayya-sharipova (approved):
@jtibshirani Thanks, great work! I left a couple of comments.

server/src/main/java/org/elasticsearch/search/vectors/KnnSearchBuilder.java
jtibshirani: Yes, thanks for catching that!

server/src/main/java/org/elasticsearch/search/vectors/KnnSearchBuilder.java
jtibshirani: Good point, not sure why I did this manually.

jtibshirani:
Thanks @mayya-sharipova for the review. I addressed your comments and also updated the feature branch `knn-search` with the latest commits from `main` (that's why there are force pushes).

PR #88220, opened 2022-06-30
Allow configuring snyk target reference and lifecycle properties

breskeby:
We will use target reference to distinguish between different versions or branches of our elasticsearch project
to be able to trace vulnerable dependencies down to the version.

snyk lifecycle property allows filtering the project overview by `production` or `development`. When version
is ending with SNAPSHOT we configure the lifecycle as development. Otherwise its production.

Related to #87620

### DIFF
--- a/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/snyk/GenerateSnykDependencyGraph.java
+++ b/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/snyk/GenerateSnykDependencyGraph.java
@@ -22,11 +22,11 @@ import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.InputFiles;
 import org.gradle.api.tasks.OutputFile;
 import org.gradle.api.tasks.TaskAction;
-import org.gradle.initialization.layout.BuildLayout;
 
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.StandardOpenOption;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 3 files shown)


build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/snyk/SnykDependencyMonitoringGradlePlugin.java
```diff
@@ -37,8 +37,11 @@ public void apply(Project project) {
             .register("generateSnykDependencyGraph", GenerateSnykDependencyGraph.class, generateSnykDependencyGraph -> {
                 generateSnykDependencyGraph.getProjectPath().set(project.getPath());
                 generateSnykDependencyGraph.getProjectName().set(project.getName());
-                generateSnykDependencyGraph.getVersion().set(project.getVersion().toString());
+                String projectVersion = project.getVersion().toString();
+                generateSnykDependencyGraph.getVersion().set(projectVersion);
                 generateSnykDependencyGraph.getGradleVersion().set(project.getGradle().getGradleVersion());
+                generateSnykDependencyGraph.getTargetReference()
+                    .set(providerFactory.gradleProperty("snykTargetReference").orElse(projectVersion));
```
breskeby, line 44: This allows us to tweak the targetReference from CI instead of relying on the version property there. Kibana for example is just using `main` and `7.x` to distinguish between development branches

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/snyk/GenerateSnykDependencyGraph.java
```diff
@@ -96,7 +96,22 @@ private Map<String, Object> generateGradleGraphPayload() {
...
+            buildTargetData(),
+            "targetReference",
+            targetReference.get(),
+            "projectAttributes",
+            projectAttributesData()
+        );
+    }
+
+    private Map<String, List<String>> projectAttributesData() {
+        return Map.of("lifecycle", List.of(version.map(v -> v.endsWith("SNAPSHOT") ? "development" : "production").get()));
```
breskeby, line 114: This allows filtering in the snyk UI by project lifecycle

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/snyk/SnykDependencyMonitoringGradlePlugin.java
mark-vieira: Agreed. I think using branches probably makes the most sense here as that's how developers are used to thinking.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/snyk/GenerateSnykDependencyGraph.java
mark-vieira: Realistically will we _ever_ publish "production" reports? Does it really matter? If so we'd have to integrate this into the release build somehow and I don't think it's really worth doing so.

We may want to check with what other teams are doing here. InfoSec may want release build reports for auditing purposes and if so we'll need to sort out how we do that.

mark-vieira (approved):

breskeby:
@mark-vieira see my other email on about the proposal for snyk monitoring setup. Somehow we definitely should track the versions of a release IMO

PR #88224, opened 2022-06-30
Immutable cluster state controller

grcevski:
[Making this PR obsolete, splitting further]

This PR adds the immutable cluster state controller, which part of the File based settings PR (#86224).

This is a reusable component that has number of Immutable State Handler components that know how to update parts of the Elasticsearch state. This component is only used by the File Settings Service at the moment, but the intent is for modules and plugins to be able to call it to save settings and entities in the cluster state that cannot be changed by the end user. For example work on using this component to update state from plugins/modules, please see https://github.com/elastic/elasticsearch/issues/88341


server/src/test/java/org/elasticsearch/immutablestate/ImmutableClusterStateHandlerTests.java
```diff
@@ -44,53 +42,6 @@ public TransformState transform(Object source, TransformState prevState) throws
         );
     }
 
-    public void testAsMapAndFromMap() throws IOException {
```
grcevski, line 47: Not needed anymore. After I refactored the parsing and typed the handlers by the type they receive in transform, I don't need the mapTo methods anymore

server/src/main/java/org/elasticsearch/immutablestate/service/ImmutableClusterStateController.java
```diff
@@ -0,0 +1,302 @@
...
+            return;
+        }
+
+        ClusterState state = clusterService.state();
+        ImmutableStateMetadata existingMetadata = state.metadata().immutableStateMetadata().get(namespace);
+        if (checkMetadataVersion(existingMetadata, packageVersion, errorListener) == false) {
+            return;
+        }
+
+        // Do we need to retry this, or it retries automatically?
```
grcevski, line 158: I have an outstanding question if there's anything else I need to do here to ensure the cluster state update is retried if it fails.

grcevski:
@DaveCTurner I hope you don't mind me tagging you for review here, I want to make sure I did the updating of the cluster state correctly, and if I need to do anything special to retry.

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/action/ImmutableLifecycleAction.java
```diff
@@ -60,18 +59,12 @@ public String name() {
     @SuppressWarnings("unchecked")
     public Collection<PutLifecycleAction.Request> prepare(Object input) throws IOException {
         List<PutLifecycleAction.Request> result = new ArrayList<>();
+        List<LifecyclePolicy> policies = (List<LifecyclePolicy>) input;
 
-        Map<String, ?> source = asMap(input);
```
grcevski, line 64: With this PR, I'm splitting the parsing of XContent from the transform step, so that we can construct immutable cluster state updates without having to supply JSON, e.g. we can send plain Java objects. Example of this can be found here: https://github.com/elastic/elasticsearch/pull/88224/files#diff-2c79b9f3f906ff489178070b9a25efc8db03fe98234156ffb8ab19ab05b8c7d7R384. The ILM plugin will use this to save plugin state packages.

server/src/main/java/org/elasticsearch/immutablestate/ImmutableClusterStateHandler.java
```diff
@@ -95,4 +93,18 @@ default void validate(MasterNodeRequest<?> request) {
             throw new IllegalStateException("Validation error", exception);
         }
     }
+
```
grcevski, line 96: Splitting off the parsing part of handling immutable cluster state, so that we can reuse the immutable cluster state update controller to work with plain Java objects. This will come in handy when plugins/modules will want to save state on initialization.

server/src/main/java/org/elasticsearch/immutablestate/service/ImmutableClusterStateController.java
```diff
@@ -0,0 +1,302 @@
...
+
+                    @Override
+                    public void onFailure(Exception e) {
+                        logger.error("Failed to apply immutable cluster state", e);
+                        errorListener.accept(e);
+                    }
+                }
+            ),
+            ClusterStateTaskConfig.build(Priority.URGENT),
+            new ImmutableStateUpdateStateTask.ImmutableUpdateStateTaskExecutor(namespace, clusterService.getRerouteService())
```
DaveCTurner, line 182: Using a new executor instance each time will mean that these tasks are unbatched. Probably not a big deal in this context but as a rule we should re-use executor instances so that state updates are always batched.

server/src/main/java/org/elasticsearch/immutablestate/service/ImmutableClusterStateController.java
DaveCTurner: Yes, there are no retries here, this state update will be attempted just once. Typically you would be calling this from a `TransportMasterNodeAction` which does handle certain retryable failures (i.e. `NotMasterException` and `FailedToCommitClusterStateException`) up to a timeout (defaulting to `30s`). This particular update can also fail in ways that don't make sense to retry. At the moment this doesn't seem to be called from any production code, retrying or otherwise.

grcevski:

PR #88231, opened 2022-07-01
Use faster maths to project WGS84 to mercator

iverase:
This commit changes the from strict maths to sloppy maths  when we compute the projection from WGS84 to spherical mercator. Precision is more than enough and test shows a meaningful speed up when doing it for big geometries. the test performed locally shows a ~15% improvement.

### DIFF
--- a/x-pack/plugin/build.gradle
+++ b/x-pack/plugin/build.gradle
@@ -143,6 +143,7 @@ tasks.named("yamlRestTestV7CompatTransform").configure { task ->
   task.skipTest("indices.freeze/10_basic/Test index options", "#70192 -- the freeze index API is removed from 8.0")
   task.skipTest("sql/sql/Paging through results", "scrolling through search hit queries no longer produces empty last page in 8.2")
   task.skipTest("service_accounts/10_basic/Test get service accounts", "new service accounts are added")
+  task.skipTest("spatial/70_script_doc_values/diagonal length", "precision changed in 8.4.0")
 
   task.replaceValueInMatch("_type", "_doc")
   task.addAllowedWarningRegex("\\[types removal\\].*")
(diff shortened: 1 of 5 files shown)


x-pack/plugin/build.gradle
```diff
@@ -143,6 +143,7 @@ tasks.named("yamlRestTestV7CompatTransform").configure { task ->
   task.skipTest("indices.freeze/10_basic/Test index options", "#70192 -- the freeze index API is removed from 8.0")
   task.skipTest("sql/sql/Paging through results", "scrolling through search hit queries no longer produces empty last page in 8.2")
   task.skipTest("service_accounts/10_basic/Test get service accounts", "new service accounts are added")
+  task.skipTest("spatial/70_script_doc_values/diagonal length", "precision changed in 8.4.0")
```
craigtaverner, line 146: Do we permanently skip this test? I see you changed the test result in the yaml file so do we need to skip this at all?

craigtaverner (approved):
I did have a question about skipping a test, but otherwise I think this looks good.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/spatial/70_script_doc_values.yml
```diff
@@ -143,4 +143,4 @@ setup:
               script:
                 source: "doc['geo_shape'].getMercatorHeight()"
   - match: { hits.hits.0.fields.width.0: 389.62170283915475 }
-  - match: { hits.hits.0.fields.height.0: 333.37976840604097 }
+  - match: { hits.hits.0.fields.height.0: 333.37976841442287 }
```
craigtaverner, line 146: Would be nice if we could assert on values with a precision.

x-pack/plugin/build.gradle
iverase: We are skipping the V7 compatibility test (yamlRestTestV7CompatTransform) . It seems this mode is running the test on the v7 branch against this cluster. That currently fails because the change in precision which is fine.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/spatial/70_script_doc_values.yml
iverase: It does not help here.

PR #88238, opened 2022-07-04
Deduplicate field-caps fields in node and ccs responses

dnhatn:
We can reduce the memory usage of field-caps responses from node and remote-cluster responses by sharing field responses with the same name.  I think we can also remove the type deduplicating as we now deduplicate the whole field response.

This change can handle more concurrent field-caps requests targeting many indices.

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/fieldcaps/FieldCapabilitiesFetcher.java
+++ b/server/src/main/java/org/elasticsearch/action/fieldcaps/FieldCapabilitiesFetcher.java
@@ -39,9 +39,9 @@ class FieldCapabilitiesFetcher {
     private final IndicesService indicesService;
     private final IndexFieldCapabilities.Deduplicator fieldDeduplicator;
 
-    FieldCapabilitiesFetcher(IndicesService indicesService, boolean dedupFields) {
+    FieldCapabilitiesFetcher(IndicesService indicesService) {
         this.indicesService = indicesService;
-        this.fieldDeduplicator = dedupFields ? IndexFieldCapabilities.deduplicatorWithMap() : f -> f;
+        this.fieldDeduplicator = IndexFieldCapabilities.deduplicatorWithMap();
     }
 
     public FieldCapabilitiesIndexResponse fetch(final FieldCapabilitiesIndexRequest request) throws IOException {
--- a/server/src/main/java/org/elasticsearch/action/fieldcaps/FieldCapabilitiesFetcher.java
+++ b/server/src/main/java/org/elasticsearch/action/fieldcaps/FieldCapabilitiesFetcher.java
@@ -41,7 +41,7 @@ class FieldCapabilitiesFetcher {
 
     FieldCapabilitiesFetcher(IndicesService indicesService) {
         this.indicesService = indicesService;
-        this.fieldDeduplicator = IndexFieldCapabilities.deduplicatorWithMap();
+        this.fieldDeduplicator = new IndexFieldCapabilities.Deduplicator();
     }
 
     public FieldCapabilitiesIndexResponse fetch(final FieldCapabilitiesIndexRequest request) throws IOException {
--- a/server/src/main/java/org/elasticsearch/action/fieldcaps/FieldCapabilitiesFetcher.java
+++ b/server/src/main/java/org/elasticsearch/action/fieldcaps/FieldCapabilitiesFetcher.java
@@ -37,9 +37,11 @@ import java.util.function.Predicate;
  */
 class FieldCapabilitiesFetcher {
     private final IndicesService indicesService;
+    private final IndexFieldCapabilities.Deduplicator fieldDeduplicator;
 
-    FieldCapabilitiesFetcher(IndicesService indicesService) {
+    FieldCapabilitiesFetcher(IndicesService indicesService, boolean dedupFields) {
         this.indicesService = indicesService;
+        this.fieldDeduplicator = dedupFields ? IndexFieldCapabilities.deduplicatorWithMap() : f -> f;
     }
 
     public FieldCapabilitiesIndexResponse fetch(final FieldCapabilitiesIndexRequest request) throws IOException {
@@ -80,7 +82,7 @@ class FieldCapabilitiesFetcher {
                         ft.isAggregatable(),
                         ft.meta()
                     );
-                    responseMap.put(field, fieldCap);
+                    responseMap.put(field, fieldDeduplicator.deduplicate(fieldCap));
                 } else {
                     continue;
                 }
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/elasticsearch/action/fieldcaps/IndexFieldCapabilities.java
```diff
@@ -62,7 +60,7 @@ public class IndexFieldCapabilities implements Writeable {
     IndexFieldCapabilities(StreamInput in) throws IOException {
         if (in.getVersion().onOrAfter(Version.V_7_7_0)) {
             this.name = in.readString();
-            this.type = typeStringDeduplicator.deduplicate(in.readString());
+            this.type = in.readString();
```
original-brownbear, line 63: I get where this is coming from and like the idea here a lot, but isn't type still going to be a massively repeated string potentially if we have a ton of different fields in a large use-case? Might still be worthwhile to dedup this?

original-brownbear (commented):
I like this and it makes sense to me with my understanding of what fields will look like on average, but it's probably best for someone from search to do a real review :)
Thanks Nhat!

server/src/main/java/org/elasticsearch/action/fieldcaps/IndexFieldCapabilities.java
dnhatn: I think we should keep the type deduplicator. I restored it in https://github.com/elastic/elasticsearch/pull/88238/commits/4b9c1aad17746ddfd21192b0d66ce89e3450a70f.

dnhatn:
Thanks Armin. @javanna Would you mind taking a look?

javanna:
hey @dnhatn do you have numbers on the improvement that this brings? I am assuming that this is not needed on 8.x because we already deduplicate based on the mappings hash while in this case we ensure reusing the same instances for fields with same name and definition across multiple indices?

server/src/main/java/org/elasticsearch/action/fieldcaps/FieldCapabilitiesFetcher.java
```diff
@@ -37,9 +37,11 @@
  */
 class FieldCapabilitiesFetcher {
     private final IndicesService indicesService;
+    private final IndexFieldCapabilities.Deduplicator fieldDeduplicator;
 
-    FieldCapabilitiesFetcher(IndicesService indicesService) {
+    FieldCapabilitiesFetcher(IndicesService indicesService, boolean dedupFields) {
```
javanna, line 42: why do we need to enable deduplication conditionally?

dnhatn:
@javanna Thanks for reviewing. 

> do you have numbers on the improvement that this brings

Without this change, a 3-nodes cluster with an 8G heap couldn't handle 3 concurrent field-caps requests targeting 5000 `metricbeat` indices. With this patch, that cluster handled those field-caps requests without issues. This change also speeds up field-caps requests a bit but not significant.

>  I am assuming that this is not needed on 8.x because we already deduplicate based on the mappings hash while in this case we ensure reusing the same instances for fields with same name and definition across multiple indices

I think we shouldn't push this improvement into 8.x as it won't save much memory and can slow down the serialization.

server/src/main/java/org/elasticsearch/action/fieldcaps/FieldCapabilitiesFetcher.java
dnhatn: I disabled the deduplication for index-level requests from old nodes (i.e., before 7.16) as the deduplication won't help. However, I think we should always enable the deduplication for simplicity. I pushed https://github.com/elastic/elasticsearch/pull/88238/commits/76a0bd169fb488c279482484111e761f57ff3074.

server/src/main/java/org/elasticsearch/action/fieldcaps/TransportFieldCapabilitiesAction.java
```diff
@@ -337,6 +337,7 @@ public void messageReceived(FieldCapabilitiesNodeRequest request, TransportChann
                 final Map<String, List<ShardId>> groupedShardIds = request.shardIds()
                     .stream()
                     .collect(Collectors.groupingBy(ShardId::getIndexName));
+                final FieldCapabilitiesFetcher fieldCapabilitiesFetcher = new FieldCapabilitiesFetcher(indicesService);
```
javanna, line 340: should we move this back where it used to be?

dnhatn: We should one Fetcher for each field-caps request as we shouldn't keep the deduplication map after the request completes.

server/src/main/java/org/elasticsearch/action/fieldcaps/RequestDispatcher.java
```diff
@@ -227,7 +227,7 @@ private void sendRequestToNode(String nodeId, List<ShardId> shardIds) {
                     shardRequest,
                     parentTask,
                     TransportRequestOptions.EMPTY,
-                    new ActionListenerResponseHandler<>(listener, FieldCapabilitiesIndexResponse::new)
+                    new ActionListenerResponseHandler<>(listener, is -> new FieldCapabilitiesIndexResponse(is, f -> f))
```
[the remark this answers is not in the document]
dnhatn: Yes, I have removed the interface in https://github.com/elastic/elasticsearch/pull/88238/commits/c184ff3726eea59dd154263eca0e29e8ac3e38f3. The impact should be minimal as the no-op deduplication is used only when interacting with old-nodes (before 7.16).

server/src/main/java/org/elasticsearch/action/fieldcaps/FieldCapabilitiesIndexResponse.java
```diff
@@ -34,10 +34,10 @@ public class FieldCapabilitiesIndexResponse extends ActionResponse implements Wr
         this.originVersion = Version.CURRENT;
     }
 
-    FieldCapabilitiesIndexResponse(StreamInput in) throws IOException {
+    FieldCapabilitiesIndexResponse(StreamInput in, IndexFieldCapabilities.Deduplicator fieldDeduplicator) throws IOException {
```
javanna, line 37: Does it need to be provided as a constructor argument? Isn't it always the same type, just created externally? Could we create it within readFields?

javanna (approved):
left a small comment, LGTM otherwise

server/src/main/java/org/elasticsearch/action/fieldcaps/FieldCapabilitiesIndexResponse.java
dnhatn: We can't create the deduplicator inside `readFields` as we need to deduplicate fields between index responses. We can avoid passing the deduplicator in the constructor by deferring the deduplication until we have constructed an index response. However, this requires us to iterate and recreate a list of fields.

dnhatn:
Thanks Luca.

server/src/main/java/org/elasticsearch/action/fieldcaps/FieldCapabilitiesIndexResponse.java
javanna: thanks for explaining :)

PR #88241, opened 2022-07-04
Log unsuccessful attempts to get credentials from web identity tokens

arteam:
Currently, we only verify that local environment for web identity tokens is correctly set up, but we don't verify whether it's
possible to exchange the token to credentials from the STS. If we can't get credentials from the STS, we silently fall back
to the EC2 credentials provider. Let's try to log the web identity token auth errors, so the users get a clear message in the logs in case the STS is unavailable for the ES server.


arteam:
@elasticmachine 
run elasticsearch-ci/part-1
run elasticsearch-ci/part-2
run elasticsearch-ci/bwc

arteam:
@elasticmachine
run elasticsearch-ci/part-1
run elasticsearch-ci/part-2

arteam:
@elasticmachine  run elasticsearch-ci/bwc

arteam:
@elasticmachine update branch

modules/repository-s3/src/main/java/org/elasticsearch/repositories/s3/S3Service.java
```diff
@@ -223,7 +224,9 @@ static AWSCredentialsProvider buildCredentials(
             if (webIdentityTokenCredentialsProvider.isActive()) {
                 logger.debug("Using a custom provider chain of Web Identity Token and instance profile credentials");
                 return new PrivilegedAWSCredentialsProvider(
-                    new AWSCredentialsProviderChain(webIdentityTokenCredentialsProvider, new EC2ContainerCredentialsProviderWrapper())
+                    new CustomAWSCredentialsProviderChain(
```
tlrx, line 227: Instead of implementing another `AWSCredentialsProvider` I wonder if we could instead catch and log an error in `CustomWebIdentityTokenCredentialsProvider#getCredentials?`

And rely on `AWSCredentialsProviderChain` as it was before.

arteam:
@elasticmachine update branch

modules/repository-s3/src/main/java/org/elasticsearch/repositories/s3/S3Service.java
arteam: The catch is that `AWSCredentialsProviderChain` catches errors and silently logs them at the `DEBUG` level,, so we won't get an exception if the first provider fails, but the second succeeds. Hence, we the custom provider that logs individual errors as warnings.

tlrx: I agree and sorry if I wasn't clear. What I'm suggesting is too implement a `FilterAWSCredentialsProvider` that delegates calls to a `AWSCredentialsProvider` but catches and logs at the level we want (I suppose `ERROR`) before rethrowing. 

Someting like:
```java
private static class LoggingCredentialsProvider implements AWSCredentialsProvider {

        private final AWSCredentialsProvider delegate;

        private LoggingCredentialsProvider(AWSCredentialsProvider delegate) {
            this.delegate = Objects.requireNonNull(delegate);
        }

        @Override
        public AWSCredentials getCredentials() {
            try {
                return delegate.getCredentials();
            } catch (Exception e) {
                LOGGER.error("Unable to load credentials from " + delegate, e);
                throw e;
            }
        }

        @Override
        public void refresh() {
            try {
                delegate.refresh();
            } catch (Exception e) {
                LOGGER.error("Unable to refresh " + delegate, e);
                throw e;
            }
            
        }
    }
```

And then just wrap the low-levels providers with it:

```java
logger.debug("Using a custom provider chain of Web Identity Token and instance profile credentials");
                return new AWSCredentialsProviderChain(
                    new LoggingCredentialsProvider(webIdentityTokenCredentialsProvider),
                    new LoggingCredentialsProvider(new EC2ContainerCredentialsProviderWrapper())
                );
```

That would provide the behavior we want without having to reimplement and maintain the logic of AWSCredentialsProviderChain?

tlrx (commented):
I left a comment https://github.com/elastic/elasticsearch/pull/88241#discussion_r951367487

arteam:
@elasticmachine update branch

modules/repository-s3/src/main/java/org/elasticsearch/repositories/s3/S3Service.java
arteam: That's a good option :+1:

arteam:
@elasticmachine update branch

modules/repository-s3/src/main/java/org/elasticsearch/repositories/s3/S3Service.java
```diff
@@ -375,6 +381,37 @@ public void shutdown() throws IOException {
...
+            this.delegate = Objects.requireNonNull(delegate);
+            this.logger = Objects.requireNonNull(logger);
+        }
+
+        @Override
+        public AWSCredentials getCredentials() {
+            try {
+                return delegate.getCredentials();
+            } catch (Exception e) {
+                logger.error("Unable to load credentials from " + delegate, e);
```
tlrx, line 399:
```suggestion
                logger.error(() -> "Unable to load credentials from " + delegate, e);
```

modules/repository-s3/src/main/java/org/elasticsearch/repositories/s3/S3Service.java
```diff
@@ -375,6 +381,37 @@ public void shutdown() throws IOException {
...
+                throw e;
+            }
+        }
+
+        @Override
+        public void refresh() {
+            try {
+                delegate.refresh();
+            } catch (Exception e) {
+                logger.error("Unable to refresh " + delegate, e);
```
tlrx, line 409:
```suggestion
                logger.error(() -> "Unable to refresh " + delegate, e);
```

modules/repository-s3/src/test/java/org/elasticsearch/repositories/s3/AwsS3ServiceImplTests.java
```diff
@@ -229,4 +233,48 @@ private void assertEndpoint(Settings repositorySettings, Settings settings, Stri
...
+    public void testLoggingCredentialsProviderCatchesErrors() {
+        var mockProvider = Mockito.mock(AWSCredentialsProvider.class);
+        String mockProviderErrorMessage = "mockProvider failed to generate credentials";
+        Mockito.when(mockProvider.getCredentials()).thenThrow(new IllegalStateException(mockProviderErrorMessage));
+        var mockLogger = Mockito.mock(Logger.class);
+
+        var credentialsProvider = new S3Service.ErrorLoggingCredentialsProvider(mockProvider, mockLogger);
+        try {
+            credentialsProvider.getCredentials();
+            fail("Shouldn't successfully get credentials");
```
tlrx, line 245: Can we use expectThrows() instead?

modules/repository-s3/src/test/java/org/elasticsearch/repositories/s3/AwsS3ServiceImplTests.java
```diff
@@ -229,4 +233,48 @@ private void assertEndpoint(Settings repositorySettings, Settings settings, Stri
...
+    public void testLoggingCredentialsProviderCatchesErrorsOnRefresh() {
+        var mockProvider = Mockito.mock(AWSCredentialsProvider.class);
+        String mockProviderErrorMessage = "mockProvider failed to refresh";
+        Mockito.doThrow(new IllegalStateException(mockProviderErrorMessage)).when(mockProvider).refresh();
+        var mockLogger = Mockito.mock(Logger.class);
+
+        var credentialsProvider = new S3Service.ErrorLoggingCredentialsProvider(mockProvider, mockLogger);
+        try {
+            credentialsProvider.refresh();
+            fail("Shouldn't successfully refresh");
```
tlrx, line 267: Can we use expectThrows instead?

tlrx (approved):
LGTM, I left minor comments that must be addressed.

modules/repository-s3/src/test/java/org/elasticsearch/repositories/s3/AwsS3ServiceImplTests.java
arteam: Thank you! I keep forgetting about the assertion helpers from the Lucene test utils framework.

arteam:
@elasticmachine update branch

PR #88242, opened 2022-07-04
[Transform] unmuting TransformContinuousIT to for debug output.

hendrikmuhs:
#88215 fixes debug logging. This PR unmutes the failing test to collect logs and proof or disproof the root cause.

relates #88063


dimitris-athanasiou (approved):
LGTM

hendrikmuhs:
Luckily the failure happened in the PR build, so it wasn't necessary to merge this.

Debugging the logs showed that the problem does not seem to be a slow threadpool, but rather the opposite: The transforms were stopped _before_ a checkpoint was able to finish. It is strange that this error did not happen for the old scheduler. 

(The checked property `last_search_time` gets updated at the beginning of a checkpoint. For a fix I additionally added a check to stop the transform after it isn't indexing anymore)

hendrikmuhs:
run elasticsearch-ci/part-2

Issue #88090, opened 2022-06-27
[CI] DebMetadataTests test05CheckLintian failing on libtorch_cpu.so


grcevski:
### CI Link

https://gradle-enterprise.elastic.co/s/2bezdlvc2ylgg/tests/:qa:os:destructiveDistroTest.default-deb/org.elasticsearch.packaging.test.DebMetadataTests/test05CheckLintian?top-execution=1

### Repro line

[bash, -c, lintian  /var/lib/jenkins/workspace/elastic+elasticsearch+pull-request+packaging-tests-unix-sample/PACKAGING_TASK/destructiveDistroTest.packages/os/ubuntu-20.04-packaging/distribution/packages/deb/build/distributions/elasticsearch-8.4.0-SNAPSHOT-amd64.deb] 

### Does it reproduce?

Didn't try

### Applicable branches

master

### Failure excerpt

I think this issue is similar to https://github.com/elastic/elasticsearch/issues/87632, except this time lintian is complaining about libtorch_cpu.so and libstdc++.so.6.

mark-vieira:
@droberts195 is this a change to included ML dependencies or is this another possible update to Lintian itself just complaining about more stuff?

droberts195:
The 3 errors Lintian found were:

```
E: elasticsearch: binary-or-shlib-defines-rpath usr/share/elasticsearch/modules/x-pack-ml/platform/linux-x86_64/lib/libstdc++.so.6 '$ORIGIN'
E: elasticsearch: binary-or-shlib-defines-rpath usr/share/elasticsearch/modules/x-pack-ml/platform/linux-x86_64/lib/libtorch_cpu.so '$ORIGIN'
W: elasticsearch: shared-lib-without-dependency-information usr/share/elasticsearch/modules/x-pack-ml/platform/linux-x86_64/lib/libmkl_cdft_core.so
```

For the ones about the `RPATH` being `'$ORIGIN'` the thing that's wrong is not crystal clear. However, comparing what things look like before and after https://github.com/elastic/ml-cpp/pull/2316 shows the problem.

Before:

```
[dave@marple lib]$ readelf -a libtorch_cpu.so | grep 'R.*PATH'
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN]
[dave@marple lib]$ readelf -a libMlApi.so | grep 'R.*PATH'
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN]
```

[...]

PR #88228, opened 2022-07-01
Unmute Lintian test

droberts195:
The fix for the problems identified in #88090 is in
elastic/ml-cpp#2337, so the test can be unmuted now.

Closes #88090

PR #88243, opened 2022-07-04
[MA] Unmute test05CheckLintian

edsavage:
Relates #88090, elastic/ml-cpp#2337


droberts195:
This is the same as #88228, but that one has already passed CI.

PR #88250, opened 2022-07-04
Make JdkMapDiff return immutable map

### LINKED ISSUE #88019: DiffableUtils.diff should return an Immutable Map when used with a JDK Map
### Description

When using the DiffableUtils.diff with a vanilla OpenJDK Map, instead of the legacy ImmutableOpenMap interface, we return a mutable map (generic HashMap), which isn't the expected behaviour of the API:

https://github.com/elastic/elasticsearch/blob/45c56b22f946733712ad6d92600411c0885594c5/server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java#L85

We should wrap the map returned by the diff API in an unmodifiable version.

grcevski:
When using the DiffableUtils.diff with a vanilla OpenJDK Map, instead of the legacy ImmutableOpenMap interface, we return a mutable map (generic HashMap), which isn't the expected behaviour of the API:

[elasticsearch/server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java](https://github.com/elastic/elasticsearch/blob/45c56b22f946733712ad6d92600411c0885594c5/server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java#L85)
Line 85 in [45c56b2](https://github.com/elastic/elasticsearch/commit/45c56b22f946733712ad6d92600411c0885594c5)
 public static <K, T extends Diffable<T>> MapDiff<K, T, Map<K, T>> diff( 

This PR wraps the Map returned from the `diff.apply()` method to return an immutable map collection.

Closes https://github.com/elastic/elasticsearch/issues/88019

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java
+++ b/server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java
@@ -16,6 +16,7 @@ import org.elasticsearch.common.io.stream.Writeable.Reader;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -242,7 +243,7 @@ public final class DiffableUtils {
                 builder.put(upsert.getKey(), upsert.getValue());
             }
 
-            return builder;
+            return Collections.unmodifiableMap(builder);
         }
     }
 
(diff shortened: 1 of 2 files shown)


grcevski:

rjernst (approved):
LGTM

ChrisHegarty (approved):
LGTM.

PR #88253, opened 2022-07-04
Rework the test to reduce code duplication

idegtiarenko:


x-pack/plugin/shutdown/src/internalClusterTest/java/org/elasticsearch/xpack/shutdown/NodeShutdownShardsIT.java
```diff
@@ -167,202 +124,159 @@ public void testNotStalledIfAllShardsHaveACopyOnAnotherNode() throws Exception {
...
-        PutShutdownNodeAction.Request putShutdownRequest = new PutShutdownNodeAction.Request(
-            nodeAId,
-            SingleNodeShutdownMetadata.Type.REPLACE,
-            this.getTestName(),
-            null,
-            nodeB
+        String node0 = internalCluster().startNode();// node_t0
+        String node0Id = getNodeId(node0);
+        String node1 = "node_t1";
+        String node2 = "node_t2";
```
DaveCTurner, line 139: It's a bit yucky that we hard-code the node naming convention like this, although I see why we need to know the node names before they start up. Could we at least assert that the nodes have the right names when they start up?

Nothing to do with this change obviously, but still while we're here it would be nice to clear this up.

DaveCTurner (approved):
LGTM with one observation.

x-pack/plugin/shutdown/src/internalClusterTest/java/org/elasticsearch/xpack/shutdown/NodeShutdownShardsIT.java
```diff
@@ -167,202 +124,159 @@ public void testNotStalledIfAllShardsHaveACopyOnAnotherNode() throws Exception {
...
+
+        createIndex(
+            "myindex",
+            Settings.builder().put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 3).put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 0).build()
```
DaveCTurner, line 143: As discussed on Zoom, it was important to have replicas in this test.

DaveCTurner (changes requested):

x-pack/plugin/shutdown/src/internalClusterTest/java/org/elasticsearch/xpack/shutdown/NodeShutdownShardsIT.java
```diff
@@ -536,9 +334,9 @@ public void testAutoExpandDuringRestart() throws Exception {
             );
         });
 
-        client().prepareIndex("myindex").setSource("field", "value");
+        indexRandomData("myindex");
```
DaveCTurner, line 337: Hmm I think the previous line of code was basically a no-op whereas with this change we're now actually indexing stuff.

x-pack/plugin/shutdown/src/internalClusterTest/java/org/elasticsearch/xpack/shutdown/NodeShutdownShardsIT.java
```diff
@@ -167,202 +122,148 @@ public void testNotStalledIfAllShardsHaveACopyOnAnotherNode() throws Exception {
...
+    }
 
-        assertBusy(() -> {
-            GetShutdownStatusAction.Response shutdownStatus = client().execute(
-                GetShutdownStatusAction.INSTANCE,
-                new GetShutdownStatusAction.Request(nodeAId)
-            ).get();
-            assertThat(shutdownStatus.getShutdownStatuses().get(0).migrationStatus().getStatus(), equalTo(COMPLETE));
-        });
+    public void testNodeReplacementOverridesFilters() throws Exception {
```
DaveCTurner, line 160: This test seems to have seen some rather major changes.

DaveCTurner (changes requested):
I've spent quite some time on this and I think that this PR actually makes significant changes to these tests, some of which are not even running right now due to AwaitsFix annotations. It's also adding some extra dependence on the node naming convention even where that wasn't particularly important before.

I'm not convinced this is worth pursuing right now. At the very least, I'd like this to be a sequence of more self-contained and obviously-correct changes to simplify the review process.

x-pack/plugin/shutdown/src/internalClusterTest/java/org/elasticsearch/xpack/shutdown/NodeShutdownShardsIT.java
idegtiarenko: I confirmed with original author that this was intended to index

PR #88254, opened 2022-07-04
Use pre-parsed index mode when figuring out timestamp range

martijnvg:
Use `indexMode` field instead of parsing IndexMode from settings in order
to determine time series timestamp range. Parsing the index mode isn't
necessary, because we read the index mode when IndexMetadata is constructed,
and this just burns unnecessary CPU cycles.


original-brownbear (approved):
LGTM assuming CI is happy :) Thanks for the quick turnaround Martijn!

PR #88255, opened 2022-07-04
Support async for disk usage API

dnhatn:
This change adds async support to the DiskUsage API - which is useful when analyzing a large index. It might be better to have separate REST endpoints for retrieving and deleting responses like async_search. However, this PR relies on the `tasks` index instead as the implementation is simpler.

### DIFF
--- a/docs/changelog/88255.yaml
+++ b/docs/changelog/88255.yaml
@@ -1,0 +1,5 @@
+pr: 88255
+summary: Support async for disk usage API
+area: Search
+type: enhancement
+issues: []
--- a/docs/reference/indices/diskusage.asciidoc
+++ b/docs/reference/indices/diskusage.asciidoc
@@ -180,3 +180,13 @@ files are ignored and some parts of data files might not be scanned by the API.
 together in a compressed format, the estimated sizes of stored fields are
 best efforts and can be inaccurate. The stored size of the `_id` field
 is likely underestimated while the `_source` field is overestimated.
+
+
+===== Running disk usage API asynchronously
+
+If the request contains `wait_for_completion=false`, {es}
+performs some preflight checks, launches the request, and returns a
+<<tasks,`task`>> you can use to cancel or get the status of the task.
+{es} creates a record of this task as a document at `_tasks/<task_id>`.
+When you are done with a task, you should delete the task document so
+{es} can reclaim the space.
--- a/rest-api-spec/src/main/resources/rest-api-spec/api/indices.disk_usage.json
+++ b/rest-api-spec/src/main/resources/rest-api-spec/api/indices.disk_usage.json
@@ -32,6 +32,10 @@
         "type": "boolean",
         "description": "Must be set to [true] in order for the task to be performed. Defaults to false."
       },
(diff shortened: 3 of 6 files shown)


nik9000 (commented):
Makes sense to me. I wonder if the clients folks should look at the change to see if it works right for them. It solves that problems I have though.

sethmlarson:
Our guidelines recommend a separate API endpoint for async execution due to the response type being very different than the synchronous API endpoint depending on the value of `wait_for_completion`. Means we'd need to make all properties on the endpoint response optional.

javanna:
@nik9000 @dnhatn what is the problem that we are solving? Sorry if I missed previous discussion or a related issue around it, could you point me to it so I can catch up?

nik9000:
Yeah. I was trying to run this in cloud and it always timed out. Turns out
I could find someone with ssh access to run it, but that's the only way.
That's lame.

javanna:
I wonder if adding a new async API is the right trade-off. Was this long running request putting pressure on the server or on the client? Or is the hard proxy timeout the only reason to add async support?

dnhatn:
> Was this long running request putting pressure on the server or on the client?

The async request will add an extra load to the server for storing the results in ES indices. But, this will be small because we limit 1 disk_usage execution per node and the size of the responses is small.

> Or is the hard proxy timeout the only reason to add async support?

Yes, I wasn't aware of other reasons.

javanna:
I would start a conversation with cloud about the hard proxy timeout and see what can be done around that. I am not sure that this is a good reason to add a new API.

nik9000:
In my experience any http request that can take more than a minute or so
can't really be trusted. Cloud's timeout is longer but proxies and
firewalls and stuff tend to chop pretty quick. In an ideal world, yeah, no
new API. But I've seen so many timeouts.

sethmlarson:
> In my experience any http request that can take more than a minute or so
can't really be trusted.

Big ++ on this, Nik. There are so many things that are outside our control that can cause timeouts or transient issues, it's better to have quick response times and rely on retries to "wait" on longer operations.

javanna:
I see both sides, but the development and maintenance cost for a new async API here, as well as async APIs for anything that may need it are too high in my opinion.  The API surface of Elasticsearch is already huge. are there ways to make this API faster? How often is it slow enough to require async support? let's have a separate conversation about this. I think anyways this PR won't go in as-is because a separate API was requested, which is a bigger effort and won't fit in the current PR.

nik9000:
Just the proxy.

kunisen:
Can we make it behave like scroll API and do it async? 

Currently in cloud we have a proxy timeout (X* min) and also an ES API timeout in ESS API console (1min) which we will likely extend it to proxy timeout in the not-far future, however, per a check from our cloud dev, it took 10+min for a 150gb index, which X min is not enough. And timeout X min from proxy level is hard to extend too from infra perspective.

We want to avoid the situation that every time when we need to run this, we need to have a cloud engineer ssh into the instance and check it for us...

* 1 < X < 10. Since it's a public repo, I would avoid putting the real number here.

PR #88257, opened 2022-07-04
INFO logging of snapshot restore and completion

### LINKED ISSUE #86610: Log snapshot restores at INFO level
Today we don't record the start or end of a snapshot restore in the server logs by default (or even at `DEBUG` level). In contrast we do record regular index creation, and also the start and end of every snapshot. Snapshot restores are relatively rare, and it would definitely be helpful to see them in the logs.

I think we wouldn't want to see restores related to mounting searchable snapshots, nor creation of CCR followers, so we'd need a `silent` parameter rather like we have for index creation requests. I think we would usually want to see details of the restore request (e.g. the index pattern) in the log message.

kingherc:
But DEBUG (silent) logging of snapshot restore/completion when
done in the context of CCR or searchable snapshots.

Fixes #86610


docs/reference/snapshot-restore/apis/restore-snapshot-api.asciidoc
```diff
@@ -144,6 +144,10 @@ backing indices are configured using the data stream's matching index template.
 If `true`, the request restores aliases for any restored data streams and
 indices. If `false`, the request doesn't restore aliases. Defaults to `true`.
 
+`silent`::
+(Optional, Boolean)
+If `true`, logging happens at a lower level. Defaults to `false`.
```
kingherc, line 149: Expression is got from the similar "silent" parameter of https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/cluster/metadata/MetadataCreateIndexService.java#L416 . Not sure if it's super useful, as it does not mention what the lower level is (DEBUG or INFO). But I guessed the original description did not want to make promises about the log level, and that's why I also adopted a similar description. Feedback welcome!

server/src/main/java/org/elasticsearch/snapshots/RestoreService.java
```diff
@@ -1051,6 +1053,7 @@ public ClusterState execute(ClusterState currentState) {
                 boolean changed = false;
                 for (RestoreInProgress.Entry entry : currentState.custom(RestoreInProgress.TYPE, RestoreInProgress.EMPTY)) {
                     if (entry.state().completed()) {
+                        logger.log(entry.silent() ? Level.DEBUG : Level.INFO, "restore of snapshot [{}] completed", entry.snapshot());
```
kingherc, line 1056: @fcofdez suggested to me this is a nice place to log the completion of a snapshot restore. This meant passing the silent property from the `RestoreSnapshotRequest` to `RestoreInProgress.Entry` so that we can read the silent parameter here.

kingherc:
Hi. Adding @DaveCTurner as reviewer (since he created the issue) and @original-brownbear as another reviewer (since you have experience with snapshot restores AFAIK). I manually tested this in this way:

- Executed org.elasticsearch.snapshots.RestoreSnapshotIT#testParallelRestoreOperations (for regular snapshot restore use case) and saw that it emits the INFO log messages
- Executed org.elasticsearch.xpack.ccr.AutoFollowIT#testAutoFollow (for CCR use case) and saw it does not emit the INFO log messages
- Executed org.elasticsearch.xpack.searchablesnapshots.SearchableSnapshotsIntegTests#testCreateAndRestoreSearchableSnapshot (for searchable snapshot use case) and saw it does not emit the INFO log messages

Feel free to tell me:

- if only one of you is needed as reviewer
- whether we would like to add some sort of logging testing (if it exists for this part of the code, similar to logging tests that I saw in CoordinatorTests.java).

kingherc:
OK, I understand I should have added you as reviewers after the CI passes -- will keep that in mind.

After conversing a bit, I also understand that in order to fix the CI bwc failures, I should (a) not expose the silent parameter in XContent, and (b) add `ifs` around the serialization parts. I'll try that 🤞 .

DaveCTurner:
++ yes we shouldn't see this parameter in anything user-facing (particularly this means neither in docs nor XContent[^1]) and the wire-format change needs version guards. When receiving a transport message from an older node I suggest just treating the missing `silent` parameter as `true`.

[^1] NB XContent is not always user-facing, it's also used for writing the cluster metadata to disk (e.g. for cluster coordination and snapshots) but in this case neither of those cases apply.

kingherc:
Hi @original-brownbear and @DaveCTurner , CI passed. Feel free to review.

server/src/main/java/org/elasticsearch/action/admin/cluster/snapshots/restore/RestoreSnapshotRequest.java
```diff
@@ -383,6 +394,26 @@ public boolean includeAliases() {
         return includeAliases;
     }
 
+    /**
+     * If set to true the logging should be at a lower level
```
kingherc, line 398: Expression is got from the similar "silent" parameter of https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/cluster/metadata/MetadataCreateIndexService.java#L416 . Not sure if it's super useful, as it does not mention what the lower level is (DEBUG or INFO). But I guessed the original description did not want to make promises about the log level, and that's why I also adopted a similar description. Feedback welcome!

DaveCTurner: I guess if we're logging at `DEBUG` still then `silent` is too much of an absolute. Maybe we should call it `quiet` instead? And yes I agree we should specify the levels, and maybe something about what is logged too, something like this?

```suggestion
     * If {@code false}, report the completion of the restore at `INFO` level. If {@code true}, report no logs about this restore at `INFO` level.
```

server/src/main/java/org/elasticsearch/action/admin/cluster/snapshots/restore/RestoreSnapshotRequest.java
```diff
@@ -383,6 +394,26 @@ public boolean includeAliases() {
...
+     */
+    public RestoreSnapshotRequest silent(boolean silent) {
+        this.silent = silent;
+        return this;
+    }
+
+    /**
+     * Returns a boolean for whether logging should be at a lower or higher level
+     *
+     * @return true if logging should be at a lower level
```
DaveCTurner, line 411: No real need to repeat yourself here, I typically just use the `@return` tag.

server/src/main/java/org/elasticsearch/action/admin/cluster/snapshots/restore/RestoreSnapshotRequest.java
```diff
@@ -383,6 +394,26 @@ public boolean includeAliases() {
         return includeAliases;
     }
 
+    /**
+     * If set to true the logging should be at a lower level
+     *
+     * @param silent true if logging should be at a lower level
```
DaveCTurner, line 400: Likewise here, you could just use the `@param` tag, no need for the repetition.

server/src/main/java/org/elasticsearch/action/admin/cluster/snapshots/restore/RestoreSnapshotRequestBuilder.java
```diff
@@ -162,6 +162,17 @@ public RestoreSnapshotRequestBuilder setIncludeAliases(boolean restoreAliases) {
         return this;
     }
 
+    /**
+     * If set to true the logging should be at a lower level
+     *
+     * @param silent true if logging should be at a lower level
+     * @return this builder
+     */
+    public RestoreSnapshotRequestBuilder setSilent(boolean silent) {
```
DaveCTurner, line 171: I don't think this is used, so it can be dropped. These request builders are a bit of a legacy thing for use in clients that no longer exist. Mostly it's fine just to create the request directly.

server/src/main/java/org/elasticsearch/cluster/RestoreInProgress.java
```diff
@@ -342,10 +359,21 @@ public RestoreInProgress(StreamInput in) throws IOException {
             uuid = in.readString();
             Snapshot snapshot = new Snapshot(in);
             State state = State.fromValue(in.readByte());
+            boolean silent = true;
```
DaveCTurner, line 362: Suggest a comment about why this is the default in BwC situations (i.e. older versions never did this logging anyway)

also a style nit: I have a slight preference for doing this assignment in an else branch, especially if we're adding a comment about it.

DaveCTurner (commented):
Looks good, I left a couple of small suggestions.

I'd like to have a test which asserts that we do actually report the log message we're adding at the right level. I think you can add a test case in `RestoreSnapshotIT` for that.

kingherc:
@DaveCTurner incorporated your feedback. Feel free to review. @original-brownbear feel free to also review if you'd like. Thanks!

kingherc:
Gentle reminder for review @original-brownbear , @DaveCTurner

server/src/internalClusterTest/java/org/elasticsearch/snapshots/RestoreSnapshotIT.java
```diff
@@ -158,6 +161,73 @@ public void testParallelRestoreOperationsFromSingleSnapshot() throws Exception {
...
+                    }
+                }
+
+                @Override
+                public void assertMatched() {
+                    assertThat(startMessageSnapshot, startsWith(snapshotName));
+                    assertEquals(Arrays.asList(indexName), startMessageIndicesList);
+                    assertThat(completionMessageSnapshot, startsWith(snapshotName));
+                }
+            });
```
DaveCTurner, line 204: I think it'd be clearer to write this as two regular `SeenEventExpectation` expectations with the actual message we expect (maybe with some `*` wildcards) instead of doing this. But maybe I've missed some subtlety about why we're doing it this way.

Done this way I think the fields would need to be `volatile` right?

DaveCTurner (commented):
Sorry for the delay here @kingherc, this all looks good although I've one question about the test you added.

DaveCTurner (approved):
LGTM

PR #85162, opened 2022-03-21
Skip backing indices with a disjoint range on @timestamp field.

martijnvg:
Implicitly skip backing indices with a time series range that doesn't match with
a required filter on @timestamp field.

Relates to #74660

PR #88258, opened 2022-07-04
Collapse IndexMode#getConfiguredTimestampRange(...) into IndexMode#getTimestampBound(...)

martijnvg:
`IndexMode#getConfiguredTimestampRange(...)` functionally overlaps a lot with `IndexMode#getTimestampBound(...)`.
In order to reuse `IndexMode#getTimestampBound(...)` for the use case `IndexMode#getConfiguredTimestampRange(...)`
is reused a change had to be made to TimestampBounds.

During query rewrite and in `TimestampFieldMapperService` no `IndexScopedSettings` is available,
so made this an optional parameter. Only the usage in `IndexSettings` needs this, so that the
instance is updated in the case `index.time_series.end_time` setting is updated. In the case of
query rewrite and in `TimestampFieldMapperService` this isn't needed,  since the instance is
kept around and reused.

Followup from #85162

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/IndexSettings.java
+++ b/server/src/main/java/org/elasticsearch/index/IndexSettings.java
@@ -698,7 +698,7 @@ public final class IndexSettings {
         numberOfShards = settings.getAsInt(IndexMetadata.SETTING_NUMBER_OF_SHARDS, null);
         mode = isTimeSeriesModeEnabled() ? scopedSettings.get(MODE) : IndexMode.STANDARD;
         this.timestampBounds = mode.getTimestampBound(indexMetadata);
-        if (mode != IndexMode.TIME_SERIES) {
+        if (timestampBounds != null) {
             scopedSettings.addSettingsUpdateConsumer(
                 IndexSettings.TIME_SERIES_END_TIME,
                 endTime -> { this.timestampBounds = new TimestampBounds(this.timestampBounds, endTime); }
--- a/server/src/main/java/org/elasticsearch/index/IndexSettings.java
+++ b/server/src/main/java/org/elasticsearch/index/IndexSettings.java
@@ -697,7 +697,7 @@ public final class IndexSettings {
         this.indexMetadata = indexMetadata;
         numberOfShards = settings.getAsInt(IndexMetadata.SETTING_NUMBER_OF_SHARDS, null);
         mode = isTimeSeriesModeEnabled() ? scopedSettings.get(MODE) : IndexMode.STANDARD;
-        this.timestampBounds = mode.getTimestampBound(scopedSettings);
+        this.timestampBounds = mode.getTimestampBound(indexMetadata, scopedSettings);
         this.searchThrottled = INDEX_SEARCH_THROTTLED.get(settings);
         this.queryStringLenient = QUERY_STRING_LENIENT_SETTING.get(settings);
         this.queryStringAnalyzeWildcard = QUERY_STRING_ANALYZE_WILDCARD.get(nodeSettings);
--- a/server/src/main/java/org/elasticsearch/index/IndexSettings.java
+++ b/server/src/main/java/org/elasticsearch/index/IndexSettings.java
@@ -698,10 +698,11 @@ public final class IndexSettings {
         numberOfShards = settings.getAsInt(IndexMetadata.SETTING_NUMBER_OF_SHARDS, null);
         mode = isTimeSeriesModeEnabled() ? scopedSettings.get(MODE) : IndexMode.STANDARD;
         this.timestampBounds = mode.getTimestampBound(indexMetadata);
-        if (mode == IndexMode.TIME_SERIES) {
-            scopedSettings.addSettingsUpdateConsumer(IndexSettings.TIME_SERIES_END_TIME, endTime -> {
-                this.timestampBounds = new TimestampBounds(this.timestampBounds, endTime);
-            });
+        if (mode != IndexMode.TIME_SERIES) {
+            scopedSettings.addSettingsUpdateConsumer(
+                IndexSettings.TIME_SERIES_END_TIME,
+                endTime -> { this.timestampBounds = new TimestampBounds(this.timestampBounds, endTime); }
+            );
         }
         this.searchThrottled = INDEX_SEARCH_THROTTLED.get(settings);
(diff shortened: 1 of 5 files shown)


martijnvg:
Not super happy about the fact that `IndexScopedSettings` is an optional parameter in `TimestampBounds`'s constructor. In the case where it is really need (usage in `IndexSettings`) it really needs to be provided, otherwise weird bugs can occur.

server/src/main/java/org/elasticsearch/index/TimestampBounds.java
```diff
@@ -18,10 +20,12 @@ public class TimestampBounds {
...
-    TimestampBounds(IndexScopedSettings scopedSettings) {
-        startTime = scopedSettings.get(IndexSettings.TIME_SERIES_START_TIME).toEpochMilli();
-        endTime = scopedSettings.get(IndexSettings.TIME_SERIES_END_TIME).toEpochMilli();
-        scopedSettings.addSettingsUpdateConsumer(IndexSettings.TIME_SERIES_END_TIME, this::updateEndTime);
+    TimestampBounds(IndexMetadata indexMetadata, @Nullable IndexScopedSettings scopedSettings) {
+        startTime = indexMetadata.getTimeSeriesStart().toEpochMilli();
+        endTime = indexMetadata.getTimeSeriesEnd().toEpochMilli();
+        if (scopedSettings != null) {
+            scopedSettings.addSettingsUpdateConsumer(IndexSettings.TIME_SERIES_END_TIME, this::updateEndTime);
+        }
```
nik9000, line 28: Ah! I see. What a pain.

I wonder if it's better to have the bounds object be immutable and to move the replacement of the end time to the `IndexSettings`. Or wrap this in something mutable over there when in tsdb mode. I'd wanted to avoid having a `volatile` read from `IndexSettings` for every document write which is why I made this mutable. The way it is now you only do the volatile read when you are in tsdb mode.

The mutability here is annoying. Sorry!

server/src/main/java/org/elasticsearch/index/mapper/DataStreamTimestampFieldMapper.java
```diff
@@ -217,8 +218,9 @@ public void postParse(DocumentParserContext context) throws IOException {
         if (first == null) {
             throw new IllegalArgumentException("data stream timestamp field [" + DEFAULT_PATH + "] is missing");
         }
-        TimestampBounds bounds = context.indexSettings().getTimestampBounds();
-        if (bounds != null) {
+        var indexMode = context.indexSettings().getMode();
+        if (indexMode == IndexMode.TIME_SERIES) {
+            TimestampBounds bounds = context.indexSettings().getTimestampBounds();
```
martijnvg, line 223: This avoids doing a volatile read in case of non tsdb indices. Seems like a good idea?
I think this is the only place where `getTimestampBounds()` getter is being used.

server/src/main/java/org/elasticsearch/index/TimestampBounds.java
martijnvg: I tried to the following: ba5460c10e0cd1dfcaa654ebe4a6ada8eac12bfb
Which updates the timestampBounds in `IndexSettings` if end_time setting changes and makes `TimestampBounds` immutable.

server/src/main/java/org/elasticsearch/index/IndexSettings.java
```diff
@@ -697,7 +697,12 @@ public IndexSettings(final IndexMetadata indexMetadata, final Settings nodeSetti
         this.indexMetadata = indexMetadata;
         numberOfShards = settings.getAsInt(IndexMetadata.SETTING_NUMBER_OF_SHARDS, null);
         mode = isTimeSeriesModeEnabled() ? scopedSettings.get(MODE) : IndexMode.STANDARD;
-        this.timestampBounds = mode.getTimestampBound(scopedSettings);
+        this.timestampBounds = mode.getTimestampBound(indexMetadata);
+        if (mode == IndexMode.TIME_SERIES) {
```
nik9000, line 701: I think it's enough to do this if it isn't null, right?

nik9000: Again, I'm just paranoid about every hard check here. I *think* it'll make the code easier to reason about.

martijnvg: If someone uses `IndexMode.STANDARD` then we also attempt to get the time bounds. But that is ok?

server/src/main/java/org/elasticsearch/index/IndexSettings.java
```diff
@@ -697,7 +697,12 @@ public IndexSettings(final IndexMetadata indexMetadata, final Settings nodeSetti
...
+            scopedSettings.addSettingsUpdateConsumer(IndexSettings.TIME_SERIES_END_TIME, endTime -> {
+                this.timestampBounds = new TimestampBounds(this.timestampBounds, endTime);
```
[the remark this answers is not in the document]
nik9000: oh, you do it in the ctor. Maybe make that a static method with a name? That way some dummy like me scanning the code doesn't get confused.

server/src/main/java/org/elasticsearch/index/IndexSettings.java
martijnvg: o wait, just doing `timestampBounds != null` is fine here.

server/src/main/java/org/elasticsearch/index/mapper/DataStreamTimestampFieldMapper.java
martijnvg: I pushed this: 8ae391a1f792adf7f74566ffdd7ba13e52b3a2eb

nik9000: :+1:

server/src/main/java/org/elasticsearch/index/IndexSettings.java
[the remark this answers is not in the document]
martijnvg: pushed: 1a433e8ec07bf615f70371e0a0f7fc97062be282

nik9000 (approved):
LGTM.

PR #88262, opened 2022-07-05
Change UBI docker user to `elasticsearch`

### LINKED ISSUE #88218: Elasticsearch 7.17.x Docker image does not declare an explicit non-root user
@mark-vieira @pugnascotia The Elasticsearch image failed the red hat scan, and so cannot be published.

We were not able to see it before the day of the release because of issues with the Red Hat API/workflow.
The goal is now to make sure we can fix it before 7.17.6. I do not think we need to move 7.17.6 earlier because the security issue that led to 7.17.6 was not related to Elasticsearch. cc @elastic/release-eng 

-----

Here is the report from the Red Hat system

```
Checking if container runs as the root user because a container that does not specify a non-root user will fail the automatic certification, and will be subject to a manual review before the container can be approved for publication

Help
Check RunAsNonRoot encountered an error. Please review the preflight.log file for more information.

Suggestion
Indicate a specific USER in the dockerfile or containerfile

Knowledge base
https://access.redhat.com/documentation/en-us/red_hat_openshift_certification/4.9/html-single/red_hat_openshift_software_certification_policy_guide/index#assembly-requirements-for-container-images_openshift-sw-cert-policy-introduction
```
Screen Shot 2022-06-30 at 10 14 50 AM

Screen Shot 2022-06-30 at 10 14 50 AM

_Originally posted by @mgreau in https://github.com/elastic/dev/issues/2052#issuecomment-1171281509_

pugnascotia:
Closes #88218.

Docker image scans have flagged up the `USER` that the UBI Docker image
runs with. Switch to `elasticsearch:root`, which is what the Iron Bank
image also uses, and is what we use for all images from
8.0 onwards.

### DIFF
--- a/distribution/docker/src/docker/Dockerfile
+++ b/distribution/docker/src/docker/Dockerfile
@@ -250,7 +250,7 @@ RUN mkdir /licenses && cp LICENSE.txt /licenses/LICENSE
 COPY LICENSE /licenses/LICENSE.addendum
 <% } %>
 
-<% if (docker_base == 'iron_bank') { %>
+<% if (docker_base == 'ubi' || docker_base == 'iron_bank') { %>
 USER elasticsearch:root
 <% } %>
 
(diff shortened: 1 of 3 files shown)


breskeby (approved):
lgtm

PR #88266, opened 2022-07-05
Skip expensive name collision checks when building metadata from diff

original-brownbear:
We can skip the name collision check here as we can and must assume that the incoming diff
won't be corrupted and gets us the correct next state.
This saves quite a bit of work on data node during state publication because the check gets
expensive for large states.

relates #77466


martijnvg (approved):
Nice idea, LGTM.

original-brownbear:
Thanks Martijn!

PR #88267, opened 2022-07-05
Clarify what to do if repo analysis fails

DaveCTurner:
The docs for the repo analysis API don't really say how to react on a
failure. This commit adds a note about this case.


fcofdez (approved):
LGTM

PR #76670, opened 2021-08-18
[DOCS] Add stub docs for QueryApiKey API

lcawl:
This PR addresses the following documentation build failure:

> 09:39:38 INFO:build_docs:Bad cross-document links:
09:39:38 INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/elasticsearch/client/javascript-api/7.x/api-reference.html contains broken links to:
09:39:38 INFO:build_docs:   - en/elasticsearch/reference/7.x/security-api-query-api-key.html
09:39:38 INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/elasticsearch/client/javascript-api/current/api-reference.html contains broken links to:
09:39:38 INFO:build_docs:   - en/elasticsearch/reference/7.x/security-api-query-api-key.html
09:39:38 INFO:build_docs:  /tmp/docsbuild/target_repo/html/en/elasticsearch/client/javascript-api/master/api-reference.html contains broken links to:
09:39:38 INFO:build_docs:   - en/elasticsearch/reference/master/security-api-query-api-key.html
09:39:51 runbld>>> <<<<<<<<<<<< SCRIPT EXECUTION END <<<<<<<<<<<<

It is related to a URL that was added to the rest api spec in https://github.com/elastic/elasticsearch/pull/76238, but which won't exist until https://github.com/elastic/elasticsearch/pull/76521 is merged.  This PR creates a stop-gap stub of that page.

PR #88270, opened 2022-07-05
Updatable API keys - REST API spec and tests

n1v0lg:
This PR adds REST API spec and YAML test files for the UpdateApiKey
operation.


rest-api-spec/src/main/resources/rest-api-spec/api/security.update_api_key.json
```diff
@@ -0,0 +1,38 @@
+{
+  "security.update_api_key": {
+    "documentation": {
+      "url": "https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-update-api-key.html",
```
n1v0lg, line 4: Need to write the docs still 😅

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/30_update.yml
```diff
@@ -0,0 +1,355 @@
+---
```
n1v0lg, line 1: I'm going for happy path tests here since negative tests (e.g., non-owner or invalid updates) are covered quite thoroughly in `ApiKeyRestIT` and `ApiKeyIntegTests` already.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/30_update.yml
```diff
@@ -0,0 +1,363 @@
...
+        body: {}
+  - match: { updated: true }
+
+  # Check update works without a body
+  - do:
+      headers:
+        Authorization: "Basic YXBpX2tleV91c2VyXzE6eC1wYWNrLXRlc3QtcGFzc3dvcmQ=" # api_key_user_1
+      security.update_api_key:
+        id: "$user1_key_id"
+  - match: { updated: true }
```
n1v0lg, line 231: This will be `false` once we have the noop update logic in place.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/20_query.yml
```diff
@@ -62,7 +62,7 @@ teardown:
 
   - do:
       security.delete_role:
-        name: "use_role"
+        name: "user_role"
```
n1v0lg, line 65: Spotted this typo while copy&pasting.

n1v0lg:
@elasticmachine run elasticsearch-ci/part-2 plz. Unrelated failure.

sethmlarson:
@n1v0lg Looks pretty good so far! Thanks for adding so many YAML tests for this API.

rest-api-spec/src/main/resources/rest-api-spec/api/security.update_api_key.json
ywangd: Merging the REST spec  without a doc page will cause downstream doc build to fail. We should add stub docs with this PR. An example is https://github.com/elastic/elasticsearch/pull/76670

docs/changelog/88270.yaml
```diff
@@ -0,0 +1,5 @@
+pr: 88270
+summary: Updatable API keys - REST API spec and tests
+area: Authentication
```
ywangd, line 3: I am curious why you picked `Authentication`. I'd go with the generic `Security`. It's not great either. I don't have a strong opinion, but would like to know your thoughts.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/30_update.yml
```diff
@@ -0,0 +1,363 @@
...
+      "api_keys.0.metadata": {
+          "letter": "a",
+          "number": 42
+      }
+    }
+
+  - do:
+      security.clear_api_key_cache:
+        ids: "*"
+  - match: { _nodes.failed: 0 }
```
ywangd, line 167: I don't think we should clear API key cache after each test since this does not match the real usage pattern, i.e. it is unlikely that users would clear cache every often. I understand this is copied from `10_basic.yml`. But those calls were added when API key cache clearing API was initially introduced. So they intentionally excercise the API. But outside that, there is no need (and actually better not to) call this API.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/30_update.yml
ywangd: The YAML tests look fine to me. That said, I just noticed two things in Java tests:
1. There is a bug in `ApiKeyIntegTests.testUpdateApiKeyAutoUpdatesUserRoles` in that the role definition most often do not change because the `putRoleWithClusterPrivileges` method [only honors the last cluster privilege](https://github.com/elastic/elasticsearch/blob/9016883e1c5fe1304b89394a1164531ea7cc50e9/x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java#L2048-L2050).
2. We don't seem to have tests for auto-update user information other than role definition changes, e.g. `full_name` or `email` changes. Role name change (as opposed to role definition changes) is not covered either.

ywangd (approved):
LGTM pending stub doc page as suggested below.

docs/changelog/88270.yaml
n1v0lg: I really wasn't sure here. I went with authentication because API keys generally are related to authentication but it's a bit of a stretch for this PR. I think `Security` might make more sense.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/30_update.yml
n1v0lg: Ah, good catch(es)!

I already have partial tests for (2.) in the noop PR. I'll address the points you've identified in that PR. Hope that's all good.

rest-api-spec/src/main/resources/rest-api-spec/api/security.update_api_key.json
n1v0lg: Thanks for the pointer! Will do.

PR #88271, opened 2022-07-05
fix: extract matrix stats using bucket_selector buckets_path

### LINKED ISSUE #87454: matrix_stats aggregation doesn't work with bucket_selector aggregation
### Elasticsearch Version

latest unreleased (main), latest released - 8.2.2

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

irrelevant

### Problem Description

It is not possible to reference the `kurtosis` calculation from a `matrix_stats` aggregation in a `bucket_selector` aggregation.

### Steps to Reproduce

```
POST /test/_bulk?refresh

{"index":{"_id":1}}
{"name":"test","numeric":123}
```

Running the following query

```
{
    "size": 0,
    "_source": false,
    "aggs": {
        "matrix": {
        	"filters": {
        		"filters": [
        			{"match_all":{}}]
        	},
        	"aggs": {
        		"my_stats": {
		            "matrix_stats": {
		                "fields": [
		                    "numeric"
		                ],
		                "mode": "AVG"
		            }
		        },
		        "filter_stats": {
		            "bucket_selector": {
		                "buckets_path": {
		                    "a0": "stats['numeric'].kurtosis"
		                },
		                "script": {
		                    "source": "params.a0 > params.v0",
		                    "lang": "painless",
		                    "params": {
		                        "v0": 1
		                    }
		                }
		            }
		        }
        	}
    	}
    }
}
```

Results in this error message

```
{
    "error": {
        "root_cause": [
            {
                "type": "aggregation_execution_exception",
                "reason": "Invalid path element [stats['numeric'].kurtosis] in path [stats['numeric'].kurtosis]"
            }
        ],
        "type": "aggregation_execution_exception",
        "reason": "Invalid path element [stats['numeric'].kurtosis] in path [stats['numeric'].kurtosis]"
    },
    "status": 500
}
```

According to https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-pipeline.html#buckets-path-syntax where one can use the syntax `sale_type['hat']>sales` where `hat` is an actual term from the index, I should be able to reference a specific statistic for a specific field (since `matrix_stats` allows multiple field names in its definition) with a syntax like `agg_name['field_name'].statistical_value` (ie `stats['numeric'].kurtosis`).

### Logs (if relevant)

_No response_

salvatore-campagna:
When using a `bucket_selector` and the `buckets_path`
to refer a matrix stats aggregation we need to use the
metric stats element and the field name to correctly extract
the referred value. Note also that, when extracting the
covariance and correlation values with notation
`stats[value1].covariance` or `stats[value1].correlation`
we extract `covariance[value1, value1]` and
`correlation[value1, value1]`. This patch does not address
extraction of arbitrary covariance and correlation values
such as `covariance[value1, value2]` or
`correlation[value1, value2]`.

Resolves #87454

### DIFF
--- a/docs/changelog/88271.yaml
+++ b/docs/changelog/88271.yaml
@@ -1,0 +1,6 @@
+pr: 88271
+summary: "Fix: extract matrix stats using `bucket_selector` `buckets_path`"
+area: Aggregations
+type: bug
+issues:
+ - 87454
--- a/modules/aggs-matrix-stats/src/main/java/org/elasticsearch/search/aggregations/matrix/stats/InternalMatrixStats.java
+++ b/modules/aggs-matrix-stats/src/main/java/org/elasticsearch/search/aggregations/matrix/stats/InternalMatrixStats.java
@@ -204,6 +204,26 @@ public class InternalMatrixStats extends InternalAggregation implements MatrixSt
     public Object getProperty(List<String> path) {
         if (path.isEmpty()) {
             return this;
+        } else if (path.size() == 2) {
+            if (results == null) {
+                return emptyMap();
+            }
+            final String field = path.get(0)
+                .replaceAll("^\"", "") // remove leading "
+                .replaceAll("^'", "") // remove leading '
+                .replaceAll("\"$", "") // remove trailing "
+                .replaceAll("'$", ""); // remove trailing '
+            final String element = path.get(1);
+            return switch (element) {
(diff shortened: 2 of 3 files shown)


not-napoleon (approved):
LGTM, thanks for fixing this.

Issue #81820, opened 2021-12-16
Java 17 features in Elasticsearch code base


astefan:
This is a meta issue for tracking Java 17 features implementation in Elasticsearch. The work is composed of the following tasks:
* [ ] Pattern matching for `instanceof`. Tasks are divided between folders under https://github.com/elastic/elasticsearch. The split also takes into consideration that some of the above mentioned folders do not contain any `instanceof` occurrences that can be transitioned to pattern matching. 
  * [ ]  `x-pack/plugin/analytics` through `x-pack/plugin/enrich` #https://github.com/elastic/elasticsearch/pull/81812
  * [ ] `x-pack/plugin/eql` through `x-pack/plugin/searchable-snapshots`
  * [ ] `x-pack/plugin/security` through `x-pack/plugin/wildcard`
  * [ ] `build-conventions` through `modules`
  * [ ] `plugins` through `qa`, `server/internalClusterTest`
  * [ ] `server/main`
  * [ ] `server/test`
  * [ ] `test`, `x-pack/qa`
* [ ] Improved switch expressions
* [ ] Pattern matching for switch
* [ ] Sealed classes
* [ ] Records

PR #88276, opened 2022-07-05
Updatable API keys - logging audit trail event

n1v0lg:
This PR adds a new audit trail event for when API keys are updated.


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/audit/logfile/LoggingAuditTrail.java
```diff
@@ -1228,6 +1244,18 @@ private void withRequestBody(XContentBuilder builder, CreateApiKeyRequest create
...
 
+        private void withRequestBody(final XContentBuilder builder, final UpdateApiKeyRequest updateApiKeyRequest) throws IOException {
+            builder.startObject("apikey").field("id", updateApiKeyRequest.getId());
+            if (updateApiKeyRequest.getRoleDescriptors() != null) {
+                builder.startArray("role_descriptors");
+                for (RoleDescriptor roleDescriptor : updateApiKeyRequest.getRoleDescriptors()) {
+                    withRoleDescriptor(builder, roleDescriptor);
+                }
+                builder.endArray();
+            }
```
n1v0lg, line 1255: Not including metadata here since it doesn't seem relevant to the security change event itself.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/audit/logfile/LoggingAuditTrail.java
```diff
@@ -747,6 +750,9 @@ public void accessGranted(
                 } else if (msg instanceof final SetProfileEnabledRequest setProfileEnabledRequest) {
                     assert SetProfileEnabledAction.NAME.equals(action);
                     securityChangeLogEntryBuilder(requestId).withRequestBody(setProfileEnabledRequest).build();
+                } else if (msg instanceof UpdateApiKeyRequest) {
```
ywangd, line 753: We should use `instanceof` pattern matching, i.e.:
```suggestion
                } else if (msg instanceof final UpdateApiKeyRequest updateApiKeyRequest) {
```
There is a meta issue (#81820) that tracks the effort for it as part of rolling out new Java features in the code base. Apparently this class has not been updated yet. But any new code should leverage the new feature.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/audit/logfile/LoggingAuditTrail.java
ywangd: It's an oversight that `metadata` is not included in the auditing message because `metadata` support was added after audit for security-config-change. Our view has contiuned to be that audit message should be as complete as possible. It can get very verbose. But we generally consider the leaving out information without clear context is more risky. Taking the API key metadata as an example, Fleet uses it to identify the host where an agent runs on and this is actually an important piece of information. 

That said, I am happy for it to be a separate PR. In fact, I can take this one so that you can focus on the main tasks.

ywangd (commented):
We [document](https://www.elastic.co/guide/en/elasticsearch/reference/current/audit-event-types.html) all events belongs to the `security_config_change` event type. Do you plan to have it as a separate PR? I think we can have it here since it is a relative small change and does not belong to the main API doc.

n1v0lg:
@ywangd 

> We [document](https://www.elastic.co/guide/en/elasticsearch/reference/current/audit-event-types.html) all events belongs to the security_config_change event type. Do you plan to have it as a separate PR?

Happy to add in this PR 👍 

> That said, I am happy for it to be a separate PR. In fact, I can take this one so that you can focus on the main tasks.

Awesome, appreciated!

x-pack/docs/en/security/auditing/event-types.asciidoc
```diff
@@ -231,6 +231,29 @@ event action.
...
+[source,js]
+{"type":"audit", "timestamp":"2020-12-31T00:33:52,521+0200", "node.id":
+"9clhpgjJRR-iKzOw20xBNQ", "event.type":"security_config_change", "event.action":
+"update_apikey", "request.id":"9FteCmovTzWHVI-9Gpa_vQ", "change":{"apikey_update":
+{"id":"zcwN3YEBBmnjw-K-hW5_","role_descriptors":[{"cluster":
+["monitor","manage_ilm"],"indices":[{"names":["index-a*"],"privileges":
+["read","maintenance"]},{"names":["in*","alias*"],"privileges":["read"],
+"field_security":{"grant":["field1*","@timestamp"],"except":["field11"]}}],
+"applications":[],"run_as":[]},{"cluster":["all"],"indices":[{"names":
+["index-b*"],"privileges":["all"]}],"applications":[],"run_as":[]}]}}}
```
n1v0lg, line 254: Note: no metadata here - will to add it when we add metadata to the actual event.

x-pack/docs/en/security/auditing/event-types.asciidoc
```diff
@@ -231,6 +231,29 @@ event action.
 ["index-b*"],"privileges":["all"]}],"applications":[],"run_as":[]}]}}}
 ====
 
+[[event-update-apikey]]
+`update_apikey`::
```
ywangd, line 235: We might want change this to `change_apikey` since `change` seems to be the word of choice with the current naming convention for audit events. @albertzaharovits Your view is appreciated.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/audit/logfile/LoggingAuditTrail.java
```diff
@@ -1228,6 +1244,18 @@ private void withRequestBody(XContentBuilder builder, CreateApiKeyRequest create
                 .endObject(); // apikey
         }
 
+        private void withRequestBody(final XContentBuilder builder, final UpdateApiKeyRequest updateApiKeyRequest) throws IOException {
+            builder.startObject("apikey").field("id", updateApiKeyRequest.getId());
```
ywangd, line 1248: I am thinking how this should play when we have the bulk update API. My intuition would be having a separate `apikeys` schema to avoid having both `id` and `ids` in the same schema. What do you think?

ywangd (approved):
LGTM

x-pack/docs/en/security/auditing/event-types.asciidoc
n1v0lg: I'm good with `change_apikey`. I went with `update_apikey` based on `update_user_profile_data` but no strong preference either way.

n1v0lg: I've updated to `change_apikey`, provisionally.

albertzaharovits: The current convention for auditing security config changes, is that for `event.action`  of type `x_y`, the audited config content starts with `"x": { "y":" (where is `x` can be one of `put`, `delete`, `change`, `create` or `invalidate`). Accoding to that, `change_apikey` goes with `"change": {"api_key"`.

n1v0lg: Going with `change_apikey`. Thanks!

PR #88283, opened 2022-07-05
Fix checkstyle version drift and API change

pugnascotia:
It turns out we have the Checkstyle version in 2 places:

   * `build-tools-internal/version.properties`
   * `gradle/build.versions.toml`

This concealed that there was an API change between 10.1 and 10.3, and we needed to update one of our custom checks.


breskeby (approved):
lgtm

pugnascotia:
We don't apply the custom Checkstyle in the CLI build, only in the IDE. We would have caught this if we didn't duplicate the Checkstyle version declaration. @breskeby said he'd look into consolidating them.

mark-vieira (approved):

PR #88284, opened 2022-07-05
Upgrade to Lucene-9.3.0-snapshot-2d05f5c623e

dnhatn:
To include [LUCENE-10620](https://issues.apache.org/jira/browse/LUCENE-10620) - which passes Weight to Collector.


server/src/main/java/org/elasticsearch/search/internal/ContextIndexSearcher.java
```diff
@@ -162,7 +162,10 @@ public Weight createWeight(Query query, ScoreMode scoreMode, float boost) throws
         }
     }
 
+    @Override
     public void search(List<LeafReaderContext> leaves, Weight weight, Collector collector) throws IOException {
+        weight = wrapWeight(weight);
+        collector.setWeight(weight);
```
dnhatn, line 168: This is the main change in this upgrade.

jpountz (approved):
Thanks Nhat! To get the full benefits of this new `Collector#setWeight` API, I think we'll also need to propagate `Weight#count` to sub weights in Weight wrappers that do not change the set of matching docs like the one in `ContextIndexSearcher#wrapWeight`. Maybe something we can look into in a follow-up?

PR #88285, opened 2022-07-05
Remove ImmutableOpenMap from most tests

rjernst:
This commit removes ImmutableOpenMap from most test cases from server,
where the type is no longer needed because Map is used by the class
being constructed.

relates #86239

### DIFF
--- a/server/src/test/java/org/elasticsearch/cluster/routing/allocation/ThrottlingAllocationTests.java
+++ b/server/src/test/java/org/elasticsearch/cluster/routing/allocation/ThrottlingAllocationTests.java
@@ -29,7 +29,6 @@ import org.elasticsearch.cluster.routing.allocation.command.MoveAllocationComman
 import org.elasticsearch.cluster.routing.allocation.decider.Decision;
 import org.elasticsearch.cluster.routing.allocation.decider.ThrottlingAllocationDecider;
 import org.elasticsearch.common.UUIDs;
-import org.elasticsearch.common.collect.ImmutableOpenMap;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.index.Index;
 import org.elasticsearch.index.shard.ShardId;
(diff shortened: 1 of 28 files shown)


server/src/test/java/org/elasticsearch/cluster/routing/allocation/ThrottlingAllocationTests.java
```diff
@@ -415,7 +414,7 @@ private ClusterState createRecoveryStateAndInitializeAllocations(
                 snapshot,
                 RestoreInProgress.State.INIT,
                 new ArrayList<>(snapshotIndices),
-                restoreShards.build()
+                restoreShards
```
grcevski, line 417: Another unmodifiable

server/src/test/java/org/elasticsearch/cluster/routing/allocation/ThrottlingAllocationTests.java
```diff
@@ -424,7 +423,7 @@ private ClusterState createRecoveryStateAndInitializeAllocations(
             .nodes(DiscoveryNodes.builder().add(node1))
             .metadata(metadataBuilder.build())
             .routingTable(routingTable)
-            .customs(restores.build())
+            .customs(restores)
```
grcevski, line 426: Here too.

server/src/test/java/org/elasticsearch/indices/ShardLimitValidatorTests.java
```diff
@@ -253,12 +254,12 @@ public static ClusterState createClusterForShardLimitTest(
...
 
     public static DiscoveryNodes createDiscoveryNodes(int nodesInCluster, String group) {
-        ImmutableOpenMap.Builder<String, DiscoveryNode> dataNodes = ImmutableOpenMap.builder();
+        Map<String, DiscoveryNode> dataNodes = new HashMap<>();
         for (int i = 0; i < nodesInCluster; i++) {
             dataNodes.put(randomAlphaOfLengthBetween(5, 15), createNode(group));
         }
         DiscoveryNodes nodes = mock(DiscoveryNodes.class);
-        when(nodes.getDataNodes()).thenReturn(dataNodes.build());
+        when(nodes.getDataNodes()).thenReturn(dataNodes);
```
grcevski, line 262: Unmodifiable map here too.

server/src/test/java/org/elasticsearch/snapshots/InternalSnapshotsInfoServiceTests.java
```diff
@@ -452,7 +452,7 @@ private ClusterState addUnassignedShards(final ClusterState currentState, String
                 recoverySource.snapshot(),
                 RestoreInProgress.State.INIT,
                 Collections.singletonList(indexName),
-                shards.build()
+                shards
```
grcevski, line 455: Unmodifiable map here too.

grcevski (commented):
This looks great, I left few comments where I think some of these test helper methods should preserve the immutable properties of the maps we pass in.

rjernst:
> I think some of these test helper methods should preserve the immutable properties of the maps we pass in

There are lots of other cases in existing tests where we don't make the passed Map immutable. I think this is ok. What is important is that if we have a non-mocked object, the value is immutable. For example, if we call have a Node instance (a higher level test) that gets at the DiscoveryNodes, the members should be immutable. But I don't think mocking these to be immutable is important, they are only mocked because some other system expects to read them.

grcevski (approved):
LGTM!

PR #88286, opened 2022-07-05
Don't fail validation on archived settings

### LINKED ISSUE #86022: Archived cluster settings should not block cluster's upgrade, or we should describe the impact more clear
## Background

Per manual, 
https://www.elastic.co/guide/en/elasticsearch/reference/8.1/archived-settings.html#archived-cluster-settings

There are 2 kinds of archived settings, 
- (A) Archived cluster settings
- (B) Archived index settings

For `(B) Archived index settings`, it gives a clear description below, that these archived settings can block upgrade:

```
Before you upgrade, remove any unsupported index settings from index and component templates. 
Elasticsearch doesn’t archive unsupported index settings in templates during an upgrade. 
Attempts to use a template that contains an unsupported index setting will fail and return an error. 
This includes automated operations, such the ILM rollover action.
```

For `(A) Archived cluster settings`, there's no such info, and also the manual shows we can simply remove the cluster settings by:

```
PUT _cluster/settings
{
  "persistent": {
    "archived.*": null
  }
}
```

## What's the problem?

We found that, some `archived cluster settings` like the following, can block upgrade on ESS.

- `archived.xpack.monitoring.exporters.found-user-defined.auth.password`
- `archived.xpack.monitoring.exporters.monitoring_user_defined.auth.password `

We understand the workaround is simply removing the archived settings, as mentioned above.

But it's against [the description in the manual](https://www.elastic.co/guide/en/elasticsearch/reference/8.1/archived-settings.html#archived-cluster-settings), because it's blocking the upgrade, thus the user cannot remove the settings "after upgrading".

```
We recommend you remove any archived settings after upgrading
```

## Suggested improvements

Logically thinking, depending on whether we want to keep those archived settings on purpose, there could be 2 independent improvements:

### [1]

If there's no need to keep those `Archived cluster settings`, then they should not block the upgrade.

Additionally, whether we should remove those settings automatically during the upgrade, or we put the instructions into i.e. Kibana upgrade assistant and let the user remove them, might be worth discussing later.

### [2]

If we want to keep them on purpose, then we should make it more clear, similar to `Archived index settings`, we should put a banner into the documentation, and also put the instructions into i.e. Kibana upgrade assistant, to let user handle these settings before the upgrading.

grcevski:
Skip validating existing archived settings in SettingsUpdater. This is consistent with how
we treat any newly archived settings.

Closes #86022

### DIFF
--- a/docs/changelog/88286.yaml
+++ b/docs/changelog/88286.yaml
@@ -1,0 +1,6 @@
+pr: 88286
+summary: Don't fail validation on archived settings
+area: Infra/Settings
+type: bug
+issues:
+ - 86022
--- a/server/src/internalClusterTest/java/org/elasticsearch/cluster/settings/ClusterSettingsIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/cluster/settings/ClusterSettingsIT.java
@@ -402,17 +402,6 @@ public class ClusterSettingsIT extends ESIntegTestCase {
         state = client().admin().cluster().prepareState().get().getState();
         assertTrue(state.getMetadata().persistentSettings().getAsBoolean("archived.this.is.unknown", false));
 
-        // cannot remove read only block due to archived settings
-        final IllegalArgumentException e1 = expectThrows(IllegalArgumentException.class, () -> {
-            Settings.Builder builder = Settings.builder();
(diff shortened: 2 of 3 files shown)


grcevski:
We are not taking this approach. Closing.

PR #88287, opened 2022-07-05
[ML] add new optional cache_hit field for pytorch results

benwtrent:
This commit adds the ground work for accepting a `cache_hit` boolean field that indicates if the response was cached or not.

User facing statistics are not updated with cache hit/miss information yet. That will need to be done when that information is actually provided from the native process.

This is related to: https://github.com/elastic/ml-cpp/pull/2305


davidkyle (approved):
LGTM

PR #88292, opened 2022-07-05
Autoscaling during shrink

### LINKED ISSUE #85480: Autoscaling may scale too far during index shrink
### Elasticsearch Version

7.13.3

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

ubuntu

### Problem Description

When doing a shrink, a copy of each shard is first reallocated to the one node. Then a shrink is initiated, which generates new shards. In case the chosen node is above the low watermark now, the shrunk index cannot be allocated to the node. Currently, this can result in an unnecessary scale up when using autoscaling.

This relates to the work in #76206 in that fixing that one would reduce the risk of this occurring. We also have thoughts around retries in ILM to fix a stuck shrink. However, we would still need to ensure autoscaling would not scale up unnecessarily in this case (but still scale up if that is theonly  way to ensure the shrink succeeds).

### Steps to Reproduce

NA

### Logs (if relevant)

_No response_

henningandersen:
Fix autoscaling during shrink to disregard pinning to nodes for the total tier size.
Instead signal that we need a minimum node size to hold the entire shrink
operation. This avoids scaling far higher than necessary when cluster
balancing does not allow a shrink to proceed. It is considered a
(separate) balancing issue when a shrink cannot complete with enough
space in the tier.

This changes autoscaling in general for node pinning filters (based on
`_id`, `_name` or `name` filters).

Clone and split also pins to the shards they clone or split, similarly
this is changed to ignore that pinning when calculating the total tier size.

Closes #85480


x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -324,8 +401,16 @@ private boolean cannotAllocateDueToStorage(ShardRouting shard, RoutingAllocation
...
-                return nodesInTier(allocation.routingNodes()).map(node -> allocationDeciders.canAllocate(shard, node, allocation))
-                    .anyMatch(ReactiveStorageDeciderService::isDiskOnlyNoDecision);
+                boolean diskOnly = nodesInTier(allocation.routingNodes()).map(
+                    node -> allocationDeciders.canAllocate(shard, node, allocation)
+                ).anyMatch(ReactiveStorageDeciderService::isDiskOnlyNoDecision);
+                if (diskOnly && shard.unassigned() && shard.recoverySource().getType() == RecoverySource.Type.LOCAL_SHARDS) {
+                    // For resize shards only allow autoscaling if there is no other node where the shard could fit had it not been
+                    // a resize shard. Notice that we already removed any initial_recovery filters.
+                    diskOnly = nodesInTier(allocation.routingNodes()).map(node -> allocationDeciders.canAllocate(shard, node, allocation))
+                        .anyMatch(ReactiveStorageDeciderService::isResizeOnlyNoDecision) == false;
```
fcofdez, line 411: Can we add a test where there's enough space to hold the resize shard so we verify that we don't request more capacity in that case?

fcofdez: I find double negations a bit trappy 😅

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -222,7 +298,8 @@ public static class AllocationState {
             Set<DiscoveryNode> nodes,
             Set<DiscoveryNodeRole> roles
         ) {
-            this.state = state;
+            this.state = removeNodeLockFilters(state);
```
fcofdez, line 301: I wonder if removing the node lock filter in all cases couldn't lead to a different allocation outcome? Maybe we should only remove the `initial_recovery` setting here? I'm not 100% sure though.

fcofdez (commented):
Looks mostly good, I left a few comments 👍

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
henningandersen: The idea I followed was that we should ensure that we have a node that is large enough to hold the node-locked data, hence the addition to ensure we deliver a proper node-level size.

With that done, we can assume that it is an allocation problem if it cannot fit. Hence it seems fair to remove the node locking here. I am aware that our allocation system is not yet sophisticated enough, but I'd rather not autoscale to a too large setup (since that may be multiple steps too large) in that case. For ILM controlled shrink, it will eventually fail and retry. Manual intervention may be necessary until our allocation system can handle this.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -638,6 +775,68 @@ ClusterInfo info() {
...
+        private static Settings removeNodeLockFilters(Settings settings, Predicate<String> predicate, DiscoveryNodeFilters filters) {
+            // only filter if it is a single node filter - otherwise removing it risks narrowing legal nodes for OR filters.
+            if (filters != null && filters.isSingleNodeFilter()) {
+                return settings.filter(predicate);
+            } else {
+                return settings;
+            }
+        }
+
+        private static boolean isCloneOrSplit(IndexMetadata indexMetadata, Metadata metadata) {
```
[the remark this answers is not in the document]
henningandersen: fixed in 109f82003d727e75080997af7a001877e94b9e2d

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -638,6 +775,68 @@ ClusterInfo info() {
...
+            return isNodeLocked(indexMetadata.requireFilters())
+                || isNodeLocked(indexMetadata.includeFilters())
+                || isNodeLocked(indexMetadata.getInitialRecoveryFilters());
+        }
+
+        private static boolean isNodeLocked(DiscoveryNodeFilters filters) {
+            return filters != null && filters.isSingleNodeFilter();
+        }
+
+        private static ShardRouting removeResizeRecoverySource(ShardRouting shardRouting) {
```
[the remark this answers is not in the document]
henningandersen: fixed in 109f82003d727e75080997af7a001877e94b9e2d

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -150,6 +189,28 @@ static boolean isDiskOnlyNoDecision(Decision decision) {
         return singleNoDecision(decision, single -> true).map(DiskThresholdDecider.NAME::equals).orElse(false);
     }
 
+    static boolean isResizeOnlyNoDecision(Decision decision) {
+        return singleNoDecision(decision, single -> true).map(ResizeAllocationDecider.NAME::equals).orElse(false);
+    }
+
+    enum DiskResizeNoDecision {
```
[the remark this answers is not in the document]
henningandersen: fixed in 109f82003d727e75080997af7a001877e94b9e2d

henningandersen:
Thanks @fcofdez for the review and sorry about the left-over code from previous generations of this PR. This is ready for another round (commented on the two other outstanding comments).

henningandersen:
@elasticmachine update branch

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
fcofdez: That makes sense, thanks for clarifying.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
fcofdez: Sorry, I missed that.

fcofdez (approved):
LGTM 👍

henningandersen:
Thanks Francisco!

PR #88295, opened 2022-07-05
Enforce max values limit only when running a script

javanna:
Runtime fields scripts have a hard maximum number of values that they can emit. This was accidentally inherited by some doc_value based queries that reuse the existing runtime fields infrastructure (in absence of a corresponding already existing lucene doc_value based query) as well as by script-less runtime fields that load from _source.

This limit was introduced to prevent mistakes when writing a script hence when we are loading from either _source or doc_values the same limit should not be enforced. This commit addresses this, and applies the same to the max chars limit for string fields.

### DIFF
--- a/docs/changelog/88295.yaml
+++ b/docs/changelog/88295.yaml
@@ -1,0 +1,5 @@
+pr: 88295
+summary: Enforce max values limit only when running a script
+area: Mapping
+type: bug
+issues: []
--- a/server/src/main/java/org/elasticsearch/script/AbstractFieldScript.java
+++ b/server/src/main/java/org/elasticsearch/script/AbstractFieldScript.java
@@ -99,8 +99,12 @@ public abstract class AbstractFieldScript extends DocBasedScript {
             return;
         }
         for (Object value : values) {
-            emitFromObject(value);
+            emitValueFromCompositeScript(value);
         }
+    }
+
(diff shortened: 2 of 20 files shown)


ywelsch (commented):
I'm wondering why you went with the current solution, and not add the limits as an explicit parameter to the StringFieldScript / AbstractFieldScript classes? That would make each instantiator of the class aware of setting a proper limit (or setting unbounded)?

javanna:
Good question. I went down this route because I realized that loading from _source was affected by the same problem as loading from doc_values. I came to the conclusion that we only need to have limits enforced when emit is called from within a script, which we can do by leaning on the Emit internal class that is the bridge between the script and the actual emit code. This way all the other places where we call emit programmatically hence outside of a script don't perform any check. Composite scripts did make this a bit tricky because they are the exception: it's a script that calls emit programmatically.

What you propose makes sense too, though I was hoping that we do not need to push the decision to consumers (who initializes the script class).

nik9000:
I might have tried to encode this information into the class hierarchy rather than the call path. I feel like we're sort of torturing the concept of scripts using them programmatically. There's just paths in there that only make sense to call from user code or because of the way painless operates. But both of your proposed changes are smaller and get the job done though.

ywelsch (approved):
LGTM

PR #88297, opened 2022-07-05
Print full exception when console is non-interactive

rjernst:
The console logger truncates stack traces so a user is not bombarded
with enormouse messages on startup. However, when the output is
redirected to a file, there is no need to avoid large messages, and in
fact it is a hinderance to debugging. This commit adds an internal flag
to the console logging exception convert which disables the truncation
when attached to a non-interactive console.

### DIFF
--- a/server/src/main/java/org/elasticsearch/common/logging/ConsoleThrowablePatternConverter.java
+++ b/server/src/main/java/org/elasticsearch/common/logging/ConsoleThrowablePatternConverter.java
@@ -14,17 +14,25 @@ import org.apache.logging.log4j.core.config.plugins.Plugin;
 import org.apache.logging.log4j.core.pattern.ConverterKeys;
 import org.apache.logging.log4j.core.pattern.PatternConverter;
 import org.apache.logging.log4j.core.pattern.ThrowablePatternConverter;
+import org.elasticsearch.bootstrap.BootstrapInfo;
 import org.elasticsearch.bootstrap.StartupException;
 import org.elasticsearch.common.inject.CreationException;
 
 /**
- * Outputs a very short version of exceptions for the console, pointing to full log for details.
+ * Outputs a very short version of exceptions for an interactive console, pointing to full log for details.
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/elasticsearch/common/logging/ConsoleThrowablePatternConverter.java
```diff
@@ -14,17 +14,25 @@
...
+ * Outputs a very short version of exceptions for an interactive console, pointing to full log for details.
+ *
+ * <p> If a non-interactive console is attached, the full exception is always printed.
  */
 @Plugin(name = "consoleException", category = PatternConverter.CATEGORY)
 @ConverterKeys({ "consoleException" })
 public class ConsoleThrowablePatternConverter extends ThrowablePatternConverter {
-    private ConsoleThrowablePatternConverter(String[] options, Configuration config) {
+
+    // true if exceptions shoudwhether exception formatting should always be delegated to the super class
```
grcevski, line 30: Nit: text in comment.

grcevski (approved):
LGTM!

server/src/main/java/org/elasticsearch/common/logging/ConsoleThrowablePatternConverter.java
```diff
@@ -34,20 +42,26 @@ private ConsoleThrowablePatternConverter(String[] options, Configuration config)
...
+    // package private for tests
+    static ConsoleThrowablePatternConverter newInstance(final Configuration config, final String[] options, boolean enabled) {
+        return new ConsoleThrowablePatternConverter(options, config, enabled);
     }
 
     @Override
     public void format(final LogEvent event, final StringBuilder toAppendTo) {
         Throwable error = event.getThrown();
-        if (error == null) {
+        if (enabled == false | error == null) {
```
pugnascotia, line 56: @rjernst did you mean to use a bitwise OR here?

PR #88302, opened 2022-07-06
Save loop over all indices in Metadata Builder

original-brownbear:
No need to run the mapping cleanup in a separate loop. We can just collect
the mappings right away when we loop the indices in the builder.
Also, no need for our own inline copy of `retainAll` actually.


martijnvg (approved):
Nice simplification! LGTM

PR #88304, opened 2022-07-06
Pre-size Metadata builder when applying diffs

original-brownbear:
Small but non-trivial speedup for very large cluster states where the indices map
is built one-by-one.


martijnvg (approved):
LGTM

PR #88307, opened 2022-07-06
[ML] {DRAFT} _ml_search endpoint

benwtrent:
This PR contains experimental work on an tech-preview feature `_ml_search`. This code may never be merged. It is mostly for simplifying testing.

For now, all this does is infer against a text embedding model and run KNN search.

It can be used as follows: 
```
POST my-index/_ml_search
{
  "_source": false,
  "track_total_hits": false,
  "fields": ["my_text"],
  "knn": {
    "num_candidates": 2,
    "model_id": "sentence-transformers__all-mpnet-base-v2",
    "query_string": "That is a very happy person" ,
    "k": 2,
    "field": "my_vector"
  },
  "query": {"match_all": {}}
}
```
 - `"query"` is just filters
 - `"k"` relates to knn hits returned
 - `"query_string"` is the string to embed
 - `"model_id"` is the model to infer (must be deployed and assigned to nodes already)
 - `"field"` is the indexed dense vector field


davidkyle:
I tested this and cloud and it worked really well for me  🙇. There are obvious advantages to making a single request rather than round tripping the embedding back to the client only for it to be sent back in the KNN search. 

Note: if we do decide to implement this the `took` time is not being accurately reported, I think it accounts for just the KNN search `took` time.

benwtrent:
> Note: if we do decide to implement this the took time is not being accurately reported, I think it accounts for just the KNN search took time.

Yeah, we would need to adjust for that :)

benwtrent:
@elasticmachine update branch

droberts195:
Superseded by #90450

PR #88311, opened 2022-07-06
Use chunked REST serialization for large REST responses

original-brownbear:
Adds chunked rest serialization infrastructure that tries to serialize
only what can be flushed to the channel right away instead of fully
materializing a response on heap first and then writing it to the channel.

Makes use of the new infrastructure for get-snapshots as an example use case.
Follow-ups will extend the usage of this logic to other problematic APIs like
`/_cluster/state`.

A working version of this with the cluster state crudely chunked can be found in https://github.com/original-brownbear/elasticsearch/pull/new/chunked-rest-response-with-cs 

relates #77466


server/src/main/java/org/elasticsearch/rest/action/admin/cluster/RestGetSnapshotsAction.java
```diff
@@ -85,9 +96,52 @@ public RestChannelConsumer prepareRequest(final RestRequest request, final NodeC
...
+                                        GetSnapshotsResponse.toXContentStart(builder);
+                                    }
+                                    while (snapshotInfoIterator.hasNext() && chunkStream.size() < sizeHint) {
+                                        snapshotInfoIterator.next().toXContentExternal(builder, request);
+                                    }
+                                    if (snapshotInfoIterator.hasNext() == false) {
+                                        getSnapshotsResponse.toXContentEnd(builder);
+                                        builder.close();
+                                    }
+                                    out.clearTarget();
```
DaveCTurner, line 135: I don't like that this logic lives in the REST layer. It bypasses the `toXContent` implementation in `GetSnapshotsResponse` and I expect that over time we'll introduce bugs where these things diverge. Please could we push this logic down to the data object itself somehow (maybe giving it a different interface from `ToXContentObject`) such that the only way to build its `XContent` representation is using this chunking behaviour.

I would also find it preferable to be able to test this kind of thing without involving the REST layer.

original-brownbear: > It bypasses the toXContent implementation in GetSnapshotsResponse and I expect that over time we'll introduce bugs where these things diverge. Please could we push this logic down to the data object itself somehow

Sure thing, this wasn't' supposed to be the final state of things :) The plan was definitely to introduce a clean interface here and then just use that in the non-chunked `toXContent`. On it today though :)

DaveCTurner: Sounds great (no rush tho)

original-brownbear:
A working version of this with the cluster state crudely chunked can be found in https://github.com/original-brownbear/elasticsearch/pull/new/chunked-rest-response-with-cs

qa/smoke-test-http/src/javaRestTest/java/org/elasticsearch/http/snapshots/RestGetSnapshotsIT.java
```diff
@@ -322,6 +322,33 @@ public void testSortAfterStartTime() throws Exception {
         assertThat(allBeforeStartTimeDescending(startTime1 - 1), empty());
     }
 
+    public void testLargeChunkedResponses() throws Exception {
```
original-brownbear, line 325: This is a bit of a nasty test admittedly, but I really like having at least one integration test that actually forces this chunked response implementation to run end-to-end through the networking code and not just testing networking and chunked serialization in isolation each.

server/src/main/java/org/elasticsearch/rest/AbstractRestChannel.java
```diff
@@ -99,6 +98,26 @@ public XContentBuilder newBuilder(
...
+        );
+    }
+
+    /**
+     * Creates a new {@link XContentBuilder} for a response to be sent using this channel. The builder's type can be sent as a parameter,
+     * through {@code responseContentType} or it can fallback to {@link #newBuilder(XContentType, boolean)} logic if the sent type value
+     * is {@code null}.
+     */
+    @Override
+    public XContentBuilder newBuilder(
```
original-brownbear, line 115: Had to add an override that would allow passing a user defined stream. This was the shortest path that I could see for making all the various tricky filtering and content type decisions we have in this method available to both chunked and standard encoding code. This should really get a cleanup eventually: The logic for how the lifecycle of these builders is managed is already rather complicated and with this override we add yet another piece of complexity by having chunked encoding manage memory in yet another way.

original-brownbear:
This should be good to review now at long last, did my best to keep the changes short, there's certainly some follow-ups possible here to streamline the code a little. Let me know if you'd like it better if I broke this apart, I could create a separate PR for all the x-content changes which are rather simple to follow (IMO) to keep the rather complicated network changes in an isolated PR.

server/src/main/java/org/elasticsearch/common/xcontent/ChunkedToXContent.java
```diff
@@ -0,0 +1,58 @@
...
+public interface ChunkedToXContent extends ToXContent {
+
+    /**
+     * Create a serialization session for the implementation instance that will flush to the given {@code builder} and use the given
+     * {@code params} the same way {@link ToXContent#toXContent(XContentBuilder, Params)} would.
+     * @param builder builder to flush to
+     * @param params params for serialization
+     * @return serialization session
+     */
+    ChunkedXContentSerialization toXContentChunked(XContentBuilder builder, ToXContent.Params params);
```
henningandersen, line 31: I wonder if we should instead add a method that returns the chunks as iterator or stream over`ToXContent` (or a new interface if there are compelling reasons). The individual chunks could then also be chunked again.

I think this will simplify the logic in the code that implements this, since they no longer need booleans for whether a chunk has been written or not (`wroteStart`) and also do not need to handle the nested chunking.

server/src/main/java/org/elasticsearch/common/xcontent/ChunkedToXContent.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    default XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+        ChunkedXContentSerialization serialization = toXContentChunked(builder, params);
```
henningandersen, line 35: Can we add a minimal test of this default implementation?

henningandersen (commented):
Mainly looked at the xcontent part for now (need more time to digest the netty part). I think we can simplify how this works for those implementing chunked xcontent.

server/src/main/java/org/elasticsearch/common/xcontent/ChunkedToXContent.java
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * The state of a chunked x-content serialization.
+     */
+    interface ChunkedXContentSerialization {
+
+        /**
+         * Writes a single chunk to the underlying {@link XContentBuilder}. Must be called repeatedly until a non-null return to complete
+         * the serialization.
+         *
+         * @return the {@link XContentBuilder} that was used for this serialization once finished
```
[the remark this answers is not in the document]
original-brownbear: Yea true, somehow my head liked doing this for some similarity with the existing x-content APIs but you are right, that's quite pointless :) Will adjust.

server/src/main/java/org/elasticsearch/common/xcontent/ChunkedToXContent.java
original-brownbear: That's what I had at first. What I really disliked about that was, that you'd have to pass the params in over and over when they never change across chunks, so I felt it was cleaner to make this clear via the interface.
I don't really think we'd save any complexity with this move anyway, you'd still have to build custom iterators for everything? You'll still have to have nested chunking in so far that one iterator will have to have logic for when to delegate to another one or when to return its own the last chunk etc?

Last but not least, using `Iterator` and returning some `ToXContent` might behave quite a bit worse performance wise because it won't compile as nicely as having all of this internal to the implementation (won't matter if chunks are large, but will matter more if they are small and it's not always trivial to predict chunk size).
-> I kinda like this better tbh.

server/src/main/java/org/elasticsearch/common/xcontent/ChunkedToXContent.java
original-brownbear: Seems kind of redundant doesn't it? We currently test this for the one implementation we have in `org.elasticsearch.action.admin.cluster.snapshots.get.GetSnapshotsResponseTests#testToChunkedXContent`. What does it add if I add a specific test with some (I presume) made up chunked x-content implementation?

original-brownbear:
Thanks Henning I adjusted the API as you suggested in one case and commented on the other now, let me know what you think :)

original-brownbear:
@henningandersen gentle ping here :)

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -181,6 +209,27 @@ private void doWrite(ChannelHandlerContext ctx, Netty4HttpResponse readyResponse
...
 
+    private void doWrite(ChannelHandlerContext ctx, Netty4ChunkedHttpResponse readyResponse, ChannelPromise promise) throws IOException {
+        assert currentChunkedWrite == null : "unexpected existing write [" + currentChunkedWrite + "]";
+        final PromiseCombiner combiner = new PromiseCombiner(ctx.executor());
+        currentChunkedWrite = new ChunkedWrite(combiner, promise, readyResponse);
+        final ChannelPromise first = ctx.newPromise();
+        combiner.add((Future<Void>) first);
+        enqueueWrite(ctx, readyResponse, first);
+        while (ctx.channel().isWritable()) {
+            if (writeChunk(ctx, combiner, readyResponse.body())) {
```
henningandersen, line 220: This looks slightly odd to me, might be me not fully understanding if a write can happen between the channel becoming writable and the `channelWritabilityChanged` method being called. But if `enqueueWrite` ever enqueues the `readyResponse`, it seems wrong to start writing the chunk now? I suppose you assume that `queuedWrites` is empty, but I am not sure that is always the case. If a split write happens first and then a chunked write (due to pipelining), would we then not risk the chunk being written here first?

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -227,8 +276,28 @@ private boolean doFlush(ChannelHandlerContext ctx) {
...
+                }
+                currentWrite = queuedWrites.poll();
+            }
+            if (currentWrite == null) {
+                // no write from a non-chunked message was found queued to be written, check if a chunked message might have become
+                // writable
+                if (currentChunkedWrite != null) {
+                    if (writeChunk(ctx, currentChunkedWrite.combiner, currentChunkedWrite.response.body())) {
+                        finishChunkedWrite();
+                        continue;
```
henningandersen, line 298: I'd think we would want to continue rather than break also if `writeChunk` returns false?

henningandersen (commented):
Went through some of it, have a few comments and questions.

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
original-brownbear: Right :facepalm: that was a real bug, thanks for catching. I added a specific test with a message that outputs small chunks to reproduce and cover now (see https://github.com/elastic/elasticsearch/pull/88311/commits/9b0680009263bfb991833bc5c30b2d9ec6489293).

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
original-brownbear: :facepalm: again, you're right this might be an issue we could run into. I pushed https://github.com/elastic/elasticsearch/pull/88311/commits/ee09bb92b7e185130d2050bb55187d1de04a0f5e to as you point out, cover the case where we only queue the first chunk because there's a pending split write of a huge non-chunked message.

original-brownbear:
Thanks Henning all points addressed again I think :)

server/src/test/java/org/elasticsearch/rest/RestControllerTests.java
```diff
@@ -697,6 +697,11 @@ public HttpResponse createResponse(RestStatus status, BytesReference content) {
                 return null;
             }
 
+            @Override
+            public HttpResponse createResponse(RestStatus status, ChunkedRestResponseBody content) {
+                return null;
```
henningandersen, line 702: Can we throw or assert false?

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -181,6 +210,28 @@ private void doWrite(ChannelHandlerContext ctx, Netty4HttpResponse readyResponse
         writeSequence++;
     }
 
+    private void doWrite(ChannelHandlerContext ctx, Netty4ChunkedHttpResponse readyResponse, ChannelPromise promise) throws IOException {
+        final PromiseCombiner combiner = new PromiseCombiner(ctx.executor());
+        currentChunkedWrite = new ChunkedWrite(combiner, promise, readyResponse);
```
henningandersen, line 215: nit: I'd find it more intuitive to move this two lines down, setting up the combiner first.

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -227,8 +278,28 @@ private boolean doFlush(ChannelHandlerContext ctx) {
...
+                    ctx.flush();
+                    needsFlush = false;
+                }
+                if (channel.isWritable() == false) {
+                    break;
+                }
+                currentWrite = queuedWrites.poll();
+            }
+            if (currentWrite == null) {
+                // no write from a non-chunked message was found queued to be written, check if a chunked message might have become
```
henningandersen, line 295: I think this comment is now misleading, since we also queue chunked writes?

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -227,8 +278,28 @@ private boolean doFlush(ChannelHandlerContext ctx) {
...
+                // writable
+                if (currentChunkedWrite != null) {
+                    if (writeChunk(ctx, currentChunkedWrite.combiner, currentChunkedWrite.response.body())) {
+                        finishChunkedWrite();
+                    }
+                    continue;
```
henningandersen, line 301: Do we not need a flush here too if the channel is not writeable?

I wonder if we can simplify the logic here, perhaps splitting into 3 methods with the 3 concerns (outer queue, queued writes and chunked writes)?

henningandersen (commented):
Went a bit deeper this round, left more comments.

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
original-brownbear: heh I knew you'd notice :) I think the whole business if flushing in the loop is nonsense for normal messages right now and just a waste of cycles. I didn't add it to the chunked write loop on purpose and was going to remove it for normal writes in a follow-up.
Calling flush here will never make the channel writable again, all it will do is enqueue a redundant flush task on the executor. If we aren't writable any more we should exit the loop, pass a single flush along and wait for the writability-changed callback to be called again. I think with that in mind, it's simple enough to keep the code as currently is? (at least IMO here it's slightly easier to read having everything in one method body)

server/src/main/java/org/elasticsearch/rest/ChunkedRestResponseBody.java
```diff
@@ -0,0 +1,107 @@
...
+                final RecyclerBytesStreamOutput chunkStream = new RecyclerBytesStreamOutput(recycler);
+                assert this.target == null;
+                this.target = chunkStream;
+                while (serialization.hasNext()) {
+                    serialization.next().toXContent(builder, params);
+                    if (chunkStream.size() >= sizeHint) {
+                        break;
+                    }
+                }
+                if (serialization.hasNext() == false) {
```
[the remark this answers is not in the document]
original-brownbear: You don't have to flush the builder manually, closing should flush everything from it. We don't really limit the size of the underlying stream as well so even if we overshoot here it's not so bad except for a bit of wasted memory?
Maybe I'm getting the question wrong?

server/src/main/java/org/elasticsearch/common/xcontent/ChunkedToXContent.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Create an iterator of {@link ToXContent} chunks, that must be serialized individually with the same {@link XContentBuilder} and
+     * {@link ToXContent.Params} for each call until it is fully drained.
+     * @return iterator over chunks of {@link ToXContent}
+     */
+    Iterator<ToXContent> toXContentChunked();
+
+    @Override
+    default XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
```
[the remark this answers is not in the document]
original-brownbear: Yea I think I mentioned this elsewhere as well, the extension really is just there for the tests at the moment. Might still be nicer to remove it in a follow-up and adjust tests there to keep this a little shorter? I kinda hate mixing snapshot and tricky network changes in one PR :)

server/src/main/java/org/elasticsearch/rest/RestResponse.java
```diff
@@ -68,13 +74,27 @@ public RestResponse(RestStatus status, String responseMediaType, String content)
...
      * Creates a binary response.
      */
-    public RestResponse(RestStatus status, String responseMediaType, BytesReference content) {
+    private RestResponse(
+        RestStatus status,
+        String responseMediaType,
+        @Nullable BytesReference content,
+        @Nullable ChunkedRestResponseBody responseBody
+    ) {
         this.status = status;
```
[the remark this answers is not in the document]
original-brownbear: Done

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
henningandersen: Can we remove the flush first for the normal writes, in a separate PR? That would simplify this PR a bit and also ensure that not flushing is sound.

original-brownbear: Sure :) https://github.com/elastic/elasticsearch/pull/89647

server/src/main/java/org/elasticsearch/http/DefaultRestChannel.java
```diff
@@ -102,24 +102,28 @@ public void sendResponse(RestResponse restResponse) {
...
-            }
-            toClose.add(this::releaseOutputBuffer);
-
-            BytesReference finalContent = content;
-            try {
-                if (request.method() == RestRequest.Method.HEAD) {
-                    finalContent = BytesArray.EMPTY;
+            final HttpResponse httpResponse;
+            if (restResponse.isChunked()) {
+                httpResponse = httpRequest.createResponse(restResponse.status(), restResponse.chunkedContent());
```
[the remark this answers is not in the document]
original-brownbear: Uff yes, implemented a solution for that now and more importantly added a test (we had 0 coverage of the `HEAD` handling)

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
original-brownbear: ++ reworded

qa/smoke-test-http/src/javaRestTest/java/org/elasticsearch/http/snapshots/RestGetSnapshotsIT.java
original-brownbear: Not sure that'll add much coverage that we don't already have in a more pinpointed way in `Netty4HttpPipeliningHandlerTests`?

original-brownbear:
Thanks Henning, I think I addressed all points (hope I didn't miss anything :crossed_fingers: ). This should be good for another round I think.

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandlerTests.java
```diff
@@ -197,6 +209,205 @@ public void testPipeliningRequestsAreReleased() throws InterruptedException {
...
+        final EmbeddedChannel embeddedChannel = new EmbeddedChannel(capturingHandler(messagesSeen), getTestHttpHandler());
+        embeddedChannel.writeInbound(createHttpRequest("/chunked"));
+        final Netty4HttpRequest request1 = embeddedChannel.readInbound();
+        embeddedChannel.writeInbound(createHttpRequest("/chunked2"));
+        final Netty4HttpRequest request2 = embeddedChannel.readInbound();
+
+        final int chunks2 = randomIntBetween(2, 10);
+        final HttpResponse response1 = request1.createResponse(
+            RestStatus.OK,
+            new BytesArray(randomByteArrayOfLength(Math.toIntExact(embeddedChannel.bytesBeforeUnwritable() * 2)))
```
henningandersen, line 299: Also just +1 here?

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandlerTests.java
```diff
@@ -197,6 +209,205 @@ public void testPipeliningRequestsAreReleased() throws InterruptedException {
...
+        final Netty4HttpRequest request1 = embeddedChannel.readInbound();
+        final Netty4HttpRequest request2 = embeddedChannel.readInbound();
+
+        final int chunks1 = randomIntBetween(2, 10);
+        final int chunks2 = randomIntBetween(2, 10);
+        final BytesReference chunk = new BytesArray(randomByteArrayOfLength(Netty4WriteThrottlingHandler.MAX_BYTES_PER_WRITE * 2));
+        final HttpResponse response1 = request1.createResponse(RestStatus.OK, getRepeatedChunkResponseBody(chunks1, chunk));
+        final HttpResponse response2 = request2.createResponse(RestStatus.OK, getRepeatedChunkResponseBody(chunks2, chunk));
+        final ChannelPromise promise1 = embeddedChannel.newPromise();
+        embeddedChannel.write(response1, promise1);
```
henningandersen, line 245: Can we make this (and the other tests) randomly write the second response first?

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandlerTests.java
```diff
@@ -197,6 +209,205 @@ public void testPipeliningRequestsAreReleased() throws InterruptedException {
...
+        assertTrue(promise1.isDone());
+        assertThat(messagesSeen, hasSize(chunks1 + chunks2 + 2));
+        assertThat(messagesSeen.get(0), instanceOf(Netty4ChunkedHttpResponse.class));
+        assertThat(messagesSeen.get(chunks1), instanceOf(LastHttpContent.class));
+        assertThat(messagesSeen.get(chunks1 + 1), instanceOf(Netty4ChunkedHttpResponse.class));
+        assertThat(messagesSeen.get(chunks1 + chunks2 + 1), instanceOf(LastHttpContent.class));
+        assertTrue(promise2.isDone());
+    }
+
+    public void testChunkedResumesAfterSingleMessage() {
```
henningandersen, line 260: Do we need the opposite too, i.e., that single message is sent after a chunked completes?

server/src/main/java/org/elasticsearch/rest/ChunkedRestResponseBody.java
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Create a chunked response body to be written to a specific {@link RestChannel} from a {@link ChunkedToXContent}.
+     *
+     * @param chunkedToXContent chunked x-content instance to serialize
+     * @param params parameters to use for serialization
+     * @param channel channel the response will be written to
+     * @return chunked rest response body
+     */
+    static ChunkedRestResponseBody fromXContent(ChunkedToXContent chunkedToXContent, ToXContent.Params params, RestChannel channel)
```
henningandersen, line 55: Can we add a test of this method as well as the `encodeChunk` method?

qa/smoke-test-http/src/javaRestTest/java/org/elasticsearch/http/snapshots/RestGetSnapshotsIT.java
henningandersen: I think there is still value in this. For one, `encodeChunk` does not run there. Not saying that that integration will lead to issues (I see none), but some verification of pipelining with chunking end-to-end seems valuable.

henningandersen (commented):
I went through one more round here and left a few more comments now that I looked at tests too.

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -239,6 +304,19 @@ private boolean doFlush(ChannelHandlerContext ctx) {
...
     }
 
+    private boolean writeChunk(ChannelHandlerContext ctx, PromiseCombiner combiner, ChunkedRestResponseBody body) throws IOException {
+        final ReleasableBytesReference bytes = body.encodeChunk(
+            Netty4WriteThrottlingHandler.MAX_BYTES_PER_WRITE,
+            serverTransport.recycler()
+        );
+        final ByteBuf content = Netty4Utils.toByteBuf(bytes);
+        final boolean done = body.isDone();
+        final ChannelFuture f = ctx.write(done ? new DefaultLastHttpContent(content) : new DefaultHttpContent(content));
```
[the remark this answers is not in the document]
original-brownbear: We can be more extreme IMO and just assert `length > 0` and `done == false` before that. We shouldn't be calling this once we're done and should never emit an empty buffer not even for the last invocation because that'll just be needless overhead for an empty last chunk. Adding those assertions

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandlerTests.java
```diff
@@ -197,6 +209,205 @@ public void testPipeliningRequestsAreReleased() throws InterruptedException {
...
+        final List<Object> messagesSeen = new ArrayList<>();
+        final EmbeddedChannel embeddedChannel = new EmbeddedChannel(capturingHandler(messagesSeen), getTestHttpHandler());
+        embeddedChannel.writeInbound(createHttpRequest("/chunked"));
+        final Netty4HttpRequest request1 = embeddedChannel.readInbound();
+        embeddedChannel.writeInbound(createHttpRequest("/chunked2"));
+        final Netty4HttpRequest request2 = embeddedChannel.readInbound();
+
+        final int chunks2 = randomIntBetween(2, 10);
+        final BytesReference chunk = new BytesArray(randomByteArrayOfLength(Math.toIntExact(embeddedChannel.bytesBeforeUnwritable() * 2)));
```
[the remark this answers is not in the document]
original-brownbear: Makes sense + 1 and high watermark it is now

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandlerTests.java
original-brownbear: ++ done

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandlerTests.java
```diff
@@ -197,6 +209,205 @@ public void testPipeliningRequestsAreReleased() throws InterruptedException {
...
+        final List<Object> messagesSeen = new ArrayList<>();
+        final EmbeddedChannel embeddedChannel = new EmbeddedChannel(capturingHandler(messagesSeen), getTestHttpHandler());
+        embeddedChannel.writeInbound(createHttpRequest("/chunked"));
+        embeddedChannel.writeInbound(createHttpRequest("/chunked2"));
+        final Netty4HttpRequest request1 = embeddedChannel.readInbound();
+        final Netty4HttpRequest request2 = embeddedChannel.readInbound();
+
+        final int chunks1 = randomIntBetween(2, 10);
+        final int chunks2 = randomIntBetween(2, 10);
+        final BytesReference chunk = new BytesArray(randomByteArrayOfLength(Netty4WriteThrottlingHandler.MAX_BYTES_PER_WRITE * 2));
```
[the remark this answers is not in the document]
original-brownbear: ++ adjusted also

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandlerTests.java
```diff
@@ -197,6 +209,205 @@ public void testPipeliningRequestsAreReleased() throws InterruptedException {
...
+        final HttpResponse response = request.createResponse(RestStatus.OK, getRepeatedChunkResponseBody(chunks, chunk));
+        final ChannelPromise promise = embeddedChannel.newPromise();
+        embeddedChannel.write(response, promise);
+        assertFalse("should not be fully flushed right away", promise.isDone());
+        assertThat(messagesSeen, hasSize(2));
+        embeddedChannel.flush();
+        assertTrue(promise.isDone());
+        assertThat(messagesSeen, hasSize(chunks + 1));
+        assertThat(messagesSeen.get(0), instanceOf(Netty4ChunkedHttpResponse.class));
+        assertThat(messagesSeen.get(chunks), instanceOf(LastHttpContent.class));
```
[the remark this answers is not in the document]
original-brownbear: Pushed dcf73e420dd for this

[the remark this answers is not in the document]
original-brownbear: Yes pushed dcf73e420dd for this that adds that helper method

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandlerTests.java
original-brownbear: Sure added some randomisation to that effect in 911e91d4181

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandlerTests.java
original-brownbear: ++ added in 98ff648833e we obviously cover this case endlessly indirectly but it's nice to lock this down directly for sure :)

server/src/main/java/org/elasticsearch/rest/ChunkedRestResponseBody.java
original-brownbear: Sure thing, how aboout 2943f1b3f98 ?

qa/smoke-test-http/src/javaRestTest/java/org/elasticsearch/http/snapshots/RestGetSnapshotsIT.java
original-brownbear: It's not easy to add this kind of test though and it'll again be unit-testy like `Netty4HttpPipeliningHandlerTests` because I can't really force pipelining via out of order messages directly in any sort of systematic fashion. Do you really think this is worth the effort? We never added any such testing for the existing pipelining and single rest requests directly, it's all just unit-tests pretty much here as well.

original-brownbear:
Thanks again Henning! I think I once again addressed all points :)

server/src/main/java/org/elasticsearch/common/xcontent/ChunkedToXContent.java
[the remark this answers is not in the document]
henningandersen: Fine to do in a follow-up too, thanks.

server/src/main/java/org/elasticsearch/rest/ChunkedRestResponseBody.java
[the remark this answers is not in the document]
henningandersen: Sorry for missing this comment in my previous review. I were not after the case where we close (which obviously flushes too), rather the case where we do not close it. That would ensure a "whole" object is returned per chunk rather than a chunk splitting across json. Might not be a big deal at all and I'd be ok to omit this, just feels nicer.

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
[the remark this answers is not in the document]
henningandersen: I am not sure we can always have that, but as long as we emit json it sounds plausible so ok by me for now.

server/src/main/java/org/elasticsearch/rest/ChunkedRestResponseBody.java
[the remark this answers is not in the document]
original-brownbear: Ah understood. I'm not sure we actually want this. In the end, the current state of affairs where the chunks are somewhat aligned with the JSON objects is more of a side-effect of not wanting to make this too complicated in step 1. It seems like we'd rather want to size the chunks as evenly as possible to avoid overhead from too many chunks on the one hand and get smooth memory usage on the other hand.

I was to some degree planning to do a follow-up here that adds a buffering layer in the pipeline to give us nice and evenly sized chunks if I ever found the time :)

henningandersen (approved):
LGTM.

Left a few comments that I hope you will address too.

server/src/main/java/org/elasticsearch/http/DefaultRestChannel.java
```diff
@@ -101,25 +101,29 @@ public void sendResponse(RestResponse restResponse) {
...
+        } catch (IllegalArgumentException ignored) {
+            assert restResponse.status() == RestStatus.METHOD_NOT_ALLOWED
+                : "request HTTP method is unsupported but HTTP status is not METHOD_NOT_ALLOWED(405)";
+        }
+        try {
+            final HttpResponse httpResponse;
+            if (isHeadRequest == false && restResponse.isChunked()) {
+                httpResponse = httpRequest.createResponse(restResponse.status(), restResponse.chunkedContent());
+            } else {
+                final BytesReference content = restResponse.content();
```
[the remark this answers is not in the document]
original-brownbear: No this will currently just return `null` for the content, which will then be ignored below transparently (because for head we replace it with `BytesArray.EMPTY` and only do an `instanceof` check on the variable otherwise).

server/src/main/java/org/elasticsearch/rest/action/admin/cluster/RestGetSnapshotsAction.java
```diff
@@ -85,9 +86,18 @@ public RestChannelConsumer prepareRequest(final RestRequest request, final NodeC
...
-                new DispatchingRestToXContentListener<>(threadPool.executor(ThreadPool.Names.MANAGEMENT), channel, request)
-            );
+            .getSnapshots(getSnapshotsRequest, new RestActionListener<>(channel) {
+                @Override
+                protected void processResponse(GetSnapshotsResponse getSnapshotsResponse) throws IOException {
+                    ensureOpen();
+                    channel.sendResponse(
+                        new RestResponse(
+                            RestStatus.OK,
+                            (Objects.requireNonNullElse(request.getXContentType(), XContentType.JSON)).mediaType(),
```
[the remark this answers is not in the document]
original-brownbear: Interestingly enough, this isn't so easy without adding the content type to the api for the `ChunkedRestResponseBody` which I'd personally rather not do.
Would you be ok with leaving this as is and cleaning this up in a follow-up that move the logic for determining the type out of `newBuilder` where it doesn't really belong to begin with?

modules/transport-netty4/src/main/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandler.java
```diff
@@ -181,6 +210,28 @@ private void doWrite(ChannelHandlerContext ctx, Netty4HttpResponse readyResponse
...
+                if (writeChunk(ctx, combiner, readyResponse.body())) {
+                    finishChunkedWrite();
+                    return;
+                }
+            }
+        }
+    }
+
+    private void finishChunkedWrite() {
+        currentChunkedWrite.combiner.finish(currentChunkedWrite.onDone);
```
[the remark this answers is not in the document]
original-brownbear: Yea unless we have a bug where we finish the same thing twice we shouldn't get any throwing here, but makes sense to me still -> put a finally here.

modules/transport-netty4/src/test/java/org/elasticsearch/http/netty4/Netty4HttpPipeliningHandlerTests.java
original-brownbear: Right good point probably not with pipelining :)

server/src/main/java/org/elasticsearch/http/DefaultRestChannel.java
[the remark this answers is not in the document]
henningandersen: ah, ofc, thanks.

DaveCTurner (commented):
Sorry I haven't had a chance to look at this in more detail. I have one question tho: what happens if we add a response header (especially a deprecation warning) during REST serialization, which might now be after we've finished sending response headers to the client? I don't think we can do much in practice but I think it'd be wise to at least cause a test failure if this happens.

original-brownbear:
@DaveCTurner 

> if we add a response header (especially a deprecation warning) during REST serialization, which might now be after we've finished sending response headers to the client

Good point. We could add these as trailers though, which should be understood by most clients? Maybe in a follow-up that I could get on asap?

original-brownbear:
Jenkins run elasticsearch-ci/part-1  ( known + unrelated)

DaveCTurner:
> We could add these as trailers though

TIL yet another dark corner of the HTTP spec...

sethmlarson:
@original-brownbear @DaveCTurner I would avoid using trailers if possible, many HTTP libraries don't process them or make them available to users so wouldn't be actionable from clients team.

DaveCTurner:
> avoid using trailers if possible

Let's continue this on https://github.com/elastic/elasticsearch/issues/89839.

PR #88312, opened 2022-07-06
Replace usages of deprecated specialized field exists queries

javanna:
DocValueFieldExistsQuery, NormsFieldExistsQuery as well as KnnVectorFieldExistsQuery are deprecated in Lucene in favour of FieldExistsQuery which combines the three into a single query.

This commit updates Elasticsearch to no longer rely on such deprecated queries.

see https://issues.apache.org/jira/browse/LUCENE-10436


romseygeek (approved):
LGTM

PR #88314, opened 2022-07-06
Update CorruptedFileIT so that it passes with new allocation strategy

idegtiarenko:
New allocation strategy is not going to retry failed shards. Update the
test to not rely on that behavior

Rel: https://github.com/elastic/elasticsearch/issues/86429


idegtiarenko:
Just confirmed this also passes on the feature branch

idegtiarenko:

server/src/internalClusterTest/java/org/elasticsearch/index/store/CorruptedFileIT.java
```diff
@@ -438,91 +414,88 @@ public void testCorruptionOnNetworkLayer() throws ExecutionException, Interrupte
...
+        client().admin()
+            .indices()
+            .prepareUpdateSettings("test")
+            .setSettings(
+                Settings.builder()
+                    .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, "1")
+                    .put("index.routing.allocation.include._name", primariesNode.getName() + "," + unluckyNode.getName())
+            )
+            .get();
+        ensureYellow("test");
```
DaveCTurner, line 460: Hmm this would pass even if the index reaches `green` health. I think we want to wait for no initialising shards and then genuinely assert that none of the shards are on the unlucky node.

server/src/internalClusterTest/java/org/elasticsearch/index/store/CorruptedFileIT.java
```diff
@@ -438,91 +414,88 @@ public void testCorruptionOnNetworkLayer() throws ExecutionException, Interrupte
...
+            .prepareUpdateSettings("test")
+            .setSettings(
+                Settings.builder()
+                    .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, "0")
+                    .put("index.routing.allocation.include._name", primariesNode.getName())
+            )
+            .get();
+        ensureGreen("test");
+
+        // can allocate on any other data node
```
DaveCTurner, line 474: Rather than removing and adding replicas I think it should work to call the reroute API with `?retry_failed`.

DaveCTurner (commented):
Looks good, I left a couple of comments.

server/src/internalClusterTest/java/org/elasticsearch/index/store/CorruptedFileIT.java
idegtiarenko: I believe this call is waiting for no initializing shards internally. Added more detailed assertion below.

DaveCTurner (approved):
LGTM

PR #88316, opened 2022-07-06
Update NodeShutdownShardsIT to actually index docs

idegtiarenko:
Replace a noop method call with actual document indexing.


henningandersen (approved):
LGTM.

PR #88319, opened 2022-07-06
Speed up put index to metadata builder

original-brownbear:
This method is very hot when applying metadata diffs so it's worthwhile
to skip the redundant `get` here when not incrementing the version.

relates #77466


martijnvg (approved):
LGTM

original-brownbear:
Thanks Martijn :)

PR #88320, opened 2022-07-06
Add test case for a replace shutdown

idegtiarenko:
This test case verifies that the index is going to be moved to the
replacement node even if it could not be allocated anywhere in
the cluster according to the allocation settings.

Related to: https://github.com/elastic/elasticsearch/issues/86429


DaveCTurner (approved):
LGTM

PR #88323, opened 2022-07-06
[ML] Rebalance model allocations when an ML job is stopped

dimitris-athanasiou:
When ML jobs get stopped (ie. anomaly detection, data frame analytics, etc.)
it could be that memory has been freed up and that we can now assign allocations
for a model deployment.

This commit adds to the `TrainedModelAssignmentClusterService` so that when
a cluster state update is observed we also look whether a persistent task
associated to an ML job has been stopped. If so, we trigger rebalance.


davidkyle (approved):
LGTM

benwtrent (approved):

dimitris-athanasiou:
@elasticmachine update branch

PR #88329, opened 2022-07-06
File Settings Service

grcevski:
Watches a directory under Elasticsearch config for changes to a file called settings.json. This is a JSON file describing the settings that the operator wants to set that will be immutable through the REST API. Any changes to these settings will only be allowed through modifying the file. As configured by default we watch for a path operator/settings.json under the config directory. The name of the directory operator can be changed via settings.

The File Settings Service will notice updates to the file and perform an update. The operator directory doesn't have to exist, the file setting service will notice when it's created. Recommended way to change operator file based settings is through modifying the settings.json file in a brand new location and then performing a symlink under the Elasticsearch config directory.

Relates to https://github.com/elastic/elasticsearch/pull/86224


grcevski:

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,309 @@
...
+        if (watching() || active == false) {
+            // already watching or inactive, nothing to do
+            return;
+        }
+
+        logger.info("starting file settings watcher ...");
+
+        Path settingsDir = operatorSettingsDir();
+
+        /**
```
rjernst, line 159: trivial: not a javadoc, so no need for the extra `*`

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,309 @@
...
+                Path settingsFilePath = operatorSettingsFile();
+                if (Files.exists(settingsFilePath)) {
+                    logger.info("found initial operator settings file [{}], applying...", settingsFilePath);
+                    // we make a distinction here for startup, so that if we had operator settings before the node started
+                    // we would fail startup.
+                    processFileSettings(settingsFilePath, onStartup);
+                }
+                enableSettingsWatcher(settingsDir);
+            } else {
+                logger.info("operator settings directory [{}] not found, will watch for its creation...", settingsDir);
```
rjernst, line 176: nit: This seems more like a debug level message?

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,309 @@
...
+         * We essentially watch for two things:
+         *  - the creation of the operator directory (if it doesn't exist), symlink changes to the operator directory
+         *  - any changes to files inside the operator directory if it exists, filtering for settings.json
+         */
+        try {
+            this.watchService = PathUtils.getDefaultFileSystem().newWatchService();
+            if (Files.exists(settingsDir)) {
+                Path settingsFilePath = operatorSettingsFile();
+                if (Files.exists(settingsFilePath)) {
+                    logger.info("found initial operator settings file [{}], applying...", settingsFilePath);
```
rjernst, line 169: nit: This seems more like a debug level message?

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,309 @@
...
+                    // we make a distinction here for startup, so that if we had operator settings before the node started
+                    // we would fail startup.
+                    processFileSettings(settingsFilePath, onStartup);
+                }
+                enableSettingsWatcher(settingsDir);
+            } else {
+                logger.info("operator settings directory [{}] not found, will watch for its creation...", settingsDir);
+                enableSettingsWatcher(environment.configFile());
```
rjernst, line 177: If stashing the expected operator path as a member instead of environment, then this can be `operatorSettingsDir.getParent()`

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,309 @@
...
+                            Path path = operatorSettingsFile();
+
+                            if (logger.isDebugEnabled()) {
+                                key.pollEvents().stream().forEach(e -> logger.debug("{}:{}", e.kind().toString(), e.context().toString()));
+                            }
+
+                            key.pollEvents();
+                            key.reset();
+
+                            enableSettingsWatcher(settingsDir);
```
rjernst, line 221: Is there any cost or potential negative behavior from calling register after every take()? ie, why not only do this when the key indicates the parent dir was created?

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,309 @@
...
+            StandardWatchEventKinds.ENTRY_MODIFY,
+            StandardWatchEventKinds.ENTRY_CREATE,
+            StandardWatchEventKinds.ENTRY_DELETE
+        );
+    }
+
+    void processFileSettings(Path path, boolean onStartup) {
+        logger.info("processing path [{}] for [{}]", path, NAMESPACE);
+        try (
+            XContentParser parser = XContentType.JSON.xContent().createParser(XContentParserConfiguration.EMPTY, Files.newInputStream(path))
```
rjernst, line 276: nit: maybe do some static imports and/or set the json xcontent so that the try can be on one line (for readability)

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,309 @@
...
+        ) {
+            stateService.process(NAMESPACE, parser, (e) -> {
+                if (e != null) {
+                    // If we encountered an exception trying to apply the operator state at
+                    // startup time, we throw an error to force Elasticsearch to exit.
+                    if (onStartup) {
+                        throw new OperatorConfigurationError("Error applying operator settings", e);
```
rjernst, line 283: It looks like this gets propagated because the exception consumer is not guarded. That's fine, but then we catch the exception down below, so it seems like we would not fail startup as intended? Am I missing something?

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,309 @@
...
+                            key.pollEvents();
+                            key.reset();
+
+                            enableSettingsWatcher(settingsDir);
+
+                            if (watchedFileChanged(path)) {
+                                processFileSettings(path, false);
+                            }
+                        } catch (Exception e) {
+                            logger.warn("unable to watch or read operator settings file", e);
```
rjernst, line 227: Why the "watch" part of this message? There shouldn't even be an exception out of asking to watch the settings dir right?

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,309 @@
...
+                    // startup time, we throw an error to force Elasticsearch to exit.
+                    if (onStartup) {
+                        throw new OperatorConfigurationError("Error applying operator settings", e);
+                    } else {
+                        logger.error("Error processing operator settings json file", e);
+                    }
+                }
+            });
+        } catch (Exception e) {
+            logger.error("Error processing operator settings json file", e);
```
rjernst, line 290: process doesn't throw anything checked. So why is this guarded? There is already a catch outside of this method that should handled logging the exception?

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,309 @@
...
+                    } else {
+                        key.pollEvents();
+                        key.reset();
+                    }
+                }
+            } catch (Exception e) {
+                if (logger.isDebugEnabled()) {
+                    logger.debug("encountered exception watching", e);
+                }
+                logger.info("shutting down watcher thread");
```
rjernst, line 238: This would only happen when an exception occurs, so is this useful as an info message? Why not log the exception at error level, stating the watcher thread is shutting down? Isn't this always an exception case (not how the thread would be shutdown)?

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,309 @@
...
+    synchronized void stopWatcher() {
+        logger.debug("stopping watcher ...");
+        if (watching()) {
+            try {
+                watchService.close();
+                if (watcherThreadLatch != null) {
+                    watcherThreadLatch.await();
+                }
+            } catch (IOException | InterruptedException e) {
+                logger.info("encountered exception while closing watch service", e);
```
rjernst, line 254: why is this info instead of error or warn?

server/src/test/java/org/elasticsearch/action/ActionModuleTests.java
```diff
@@ -116,7 +119,9 @@ public void testSetupRestHandlerContainsKnownBuiltin() {
             null,
             null,
             usageService,
-            null
+            null,
+            mock(ClusterService.class),
+            Collections.emptyList()
```
rjernst, line 124: trivial: all these `Collections.emptyList()` can be `List.of()`

test/framework/src/main/java/org/elasticsearch/plugins/MockPluginsService.java
```diff
@@ -76,4 +80,43 @@ public PluginsAndModules info() {
             List.of()
         );
     }
+
+    @Override
+    public <T> List<? extends T> loadServiceProviders(Class<T> service) {
```
rjernst, line 85: Can this be moved out to a separate PR, with a unit test?

rjernst (commented):
Looks pretty good! I have several comments, mostly nits

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
ChrisHegarty: Unless I'm mistaken we do not set `StreamReadFeature.AUTO_CLOSE_SOURCE`, so we need to explicitly close the input stream returned from Files::newInputStream, right? Maybe just refactor into the first declaration in the try-with-resources, then pass to createParser. And maybe wrap in a BufferedInputStream ?

test/framework/src/main/java/org/elasticsearch/plugins/MockPluginsService.java
grcevski: Sounds good, I made a new PR here https://github.com/elastic/elasticsearch/pull/88690. Once that's merged I'll rebase and remove this change.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
grcevski: I'll switch to warn.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
grcevski: I didn't want to pollute the logs with an exception trace when the watching is interrupted on the take() method call. It's how we interrupt the watcher when we stop watching, e.g. we are no longer the master node. So what I'll do here is log info when ClosedWatchServiceException is thrown and log error on any other exception.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
grcevski: Good point, I'll make it generic, it's a remnant of a refactor I had done.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
grcevski: OperatorConfigurationError is of type Error, so it's never caught. I wasn't sure if there's a nicer way to cause the exception to skip all the exception handling we do and terminate the node.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
grcevski: I'll double check to see what happens when the watcher is over and over registered on the same dir, I didn't see an issue when testing, but it doesn't mean it doesn't leak somehow. 

The reason I enable the settings watcher every time is because of symlinks. It's platform specific whether the watcher continues to watch the original symlinked directory or the new one after a change to a link. I couldn't think of another way to make this more reliable, other than repeat that I want to watch the same dir.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
rjernst: We could have a separate catch for InterruptedException so that any unknown errors are logged at the appropriate level?

server/src/internalClusterTest/java/org/elasticsearch/threadpool/SimpleThreadPoolIT.java
```diff
@@ -81,8 +81,10 @@ public void testThreadNames() throws Exception {
             // or the ones that are occasionally come up from ESSingleNodeTestCase
             if (threadName.contains("[node_s_0]") // TODO: this can't possibly be right! single node and integ test are unrelated!
                 || threadName.contains("Keep-Alive-Timer")
+                || threadName.contains("readiness-service")
                 || threadName.contains("JVMCI-native") // GraalVM Compiler Thread
-                || threadName.contains("readiness-service")) {
+                || threadName.contains("file-settings-watcher")
+                || threadName.contains("FileSystemWatchService")) {
```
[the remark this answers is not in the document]
grcevski: I agree that's this is tedious to modify each time. We need to explicitly add each thread that's not in any threadpool. When I worked on the readiness-service, I was advised not to use the thread pools for threads that will constantly be running, so I followed the same approach here. `FileSystemWatchService` comes from the JDK, so there's no avoiding adding that one.

Is it OK if I followed up with a new PR to tackle this problem, rather than extend this PR further? I think I should be able to ensure all standalone Elasticsearch threads are registered in a way that this list doesn't have to be manually modified.

grcevski:
I think I addressed all of the feedback and I added an integration test to ensure it all is wired up together and it's working.

server/src/internalClusterTest/java/org/elasticsearch/threadpool/SimpleThreadPoolIT.java
```diff
@@ -81,8 +81,10 @@ public void testThreadNames() throws Exception {
             // or the ones that are occasionally come up from ESSingleNodeTestCase
             if (threadName.contains("[node_s_0]") // TODO: this can't possibly be right! single node and integ test are unrelated!
                 || threadName.contains("Keep-Alive-Timer")
+                || threadName.contains("readiness-service")
                 || threadName.contains("JVMCI-native") // GraalVM Compiler Thread
-                || threadName.contains("readiness-service")) {
+                || threadName.contains("file-settings-watcher")
```
grcevski, line 86: I will tackle cleaning up this manual registration in a separate PR.

server/src/internalClusterTest/java/org/elasticsearch/threadpool/SimpleThreadPoolIT.java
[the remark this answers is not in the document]
rjernst: A followup is great

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,330 @@
...
+                if (settingsDirWatchKey != null) {
+                    settingsDirWatchKey.cancel();
+                    settingsDirWatchKey = null;
+                }
+                fileUpdateState = null;
+                watchService.close();
+                if (watcherThreadLatch != null) {
+                    watcherThreadLatch.await();
+                }
+            } catch (IOException | InterruptedException e) {
```
rjernst, line 273: Stop happens on another ES thread, not one controlled here. So the interrupt should be restored for InterruptedException. (perhaps capture InterruptedException in another catch block). Note I think it is ok to not restore above since the thread is owned by this service and the thread will end after the interrupted exception is logged there.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,330 @@
...
+                            // if the file name maps to the same native file system file id. Symlinks
+                            // are one potential cause of inconsistency here, since their handling by
+                            // the WatchService is platform dependent.
+                            settingsDirWatchKey = enableSettingsWatcher(settingsDirWatchKey, settingsDir);
+
+                            if (watchedFileChanged(path)) {
+                                processFileSettings(path, false);
+                            }
+                        } catch (Exception e) {
+                            logger.warn("error processing operator settings file", e);
```
rjernst, line 243: How is this possible given processFileSettings logs any errors? It seems like the only possible error here is IOException? Could this be narrowed to that exception, and the warning message be distinguished from that on line 307? They are very similarly worded, yet this one, I don't think is specific to processing the file, it's just some IO problem happened while trying read it.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,330 @@
...
+     * Holds information about the last known state of the file we watched. We use this
+     * class to determine if a file has been changed.
+     */
+    record FileUpdateState(long timestamp, String path, Object fileKey) {}
+
+    /**
+     * Error subclass that is thrown when we encounter a fatal error while applying
+     * the operator cluster state at Elasticsearch boot time.
+     */
+    public static class OperatorConfigurationError extends Error {
```
rjernst, line 325: We shouldn't be extending Error (and should be renamed to avoid configuration, perhaps mentioning this is a startup problem in the name). This can extend RuntimeException so it can be propagated easily.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,330 @@
...
+
+    void processFileSettings(Path path, boolean onStartup) throws IOException {
+        logger.info("processing path [{}] for [{}]", path, NAMESPACE);
+        try (var json = new BufferedInputStream(Files.newInputStream(path))) {
+            try (XContentParser parser = JSON.xContent().createParser(XContentParserConfiguration.EMPTY, json)) {
+                stateService.process(NAMESPACE, parser, (e) -> {
+                    if (e != null) {
+                        // If we encountered an exception trying to apply the operator state at
+                        // startup time, we throw an error to force Elasticsearch to exit.
+                        if (onStartup) {
```
rjernst, line 304: The logic here is a little difficult to reason about since it depends on this boolean (but the error here is then not possible outside of startup). Perhaps a more direct approach would be to pass a Consumer<Exception> so the two callers can decide how to handle it.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,330 @@
...
+    private WatchService watchService; // null;
+    private CountDownLatch watcherThreadLatch;
+
+    private volatile FileUpdateState fileUpdateState = null;
+    private volatile WatchKey settingsDirWatchKey = null;
+
+    private volatile boolean active = false;
+    private volatile boolean initialState = true;
+
+    public static final Setting<String> OPERATOR_DIRECTORY = Setting.simpleString(
```
rjernst, line 66: Why is this configurable? I think we could start with it hardcoded? I much prefer it is always the same, so that we know it will always be directly under the config dir.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -0,0 +1,330 @@
...
+
+        Path settingsDir = operatorSettingsDir();
+
+        /*
+         * We essentially watch for two things:
+         *  - the creation of the operator directory (if it doesn't exist), symlink changes to the operator directory
+         *  - any changes to files inside the operator directory if it exists, filtering for settings.json
+         */
+        try {
+            this.watchService = PathUtils.getDefaultFileSystem().newWatchService();
```
rjernst, line 171: Rather than grab the default filesystem, it is probably better to grab it from the config dir?

rjernst (approved):
A few more comments. LGTM, no need for anymore review if you agree with them.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
grcevski: Sure thing, I'll fix it.

PR #88330, opened 2022-07-06
Remove unused InternalPlugin interface

mark-vieira:
The `InternalPlugin` interface is legacy from when all `build-tools` code was colocated with internal plugins. We now separate `build-tools` from `build-tools-internal` so this marker interface is no longer needed.


mark-vieira:

pugnascotia (approved):

PR #88331, opened 2022-07-06
Fixed NullPointerException on bulk request

catap:
Java's `ArrayList.toArray()` returns provided array when collection is empty.

Here is created a one-element array which contains null element.

Thus, returned `BulkResponse` may contains a null element as `BulkItemResponse`.

How to achieve:
1. Sent a request to `/_bulk?filter_path=took,errors`
2. call inside `BulkProcessor.Listener` a `BulkResponse.hasFailures()`


catap:
Reopened as https://github.com/elastic/elasticsearch/pull/88358

PR #88333, opened 2022-07-06
Script: Metadata for update context

stu-elastic:
Adds the `metadata()` API call and a Metadata class for the Update context.

There are different metadata available in the update context depending
on whether it is an update or an insert (via upsert).

For update, scripts can read `index`, `id`, `routing`, `version` and `timestamp`.

For insert, scripts can read `index`, `id` and `timestamp`.

Scripts can always read and write the `op` but the available ops are different.

Updates allow 'noop', 'index' and 'delete'.
Inserts allow 'noop' and 'create'.

Refs: #86472


jdconrad (commented):
I realized we have talked about this, but after viewing the code I was wondering again what if each context did have its own Metadata and the base class for this took in the Map of validators? The numerous comments in the Metadata class about how something is needed only for one context, but not another seems like it'll be hard to maintain. We also wouldn't have to have methods on the class that don't make sense for a specific context. You still wouldn't have to alias the name because it's part of the allow list specific to each context.

I can see there's a lot of refactoring work as well as you mentioned to me. Since I'm just coming back into this project it's tricky for me to differentiate what's refactoring versus what's added here that's new. Would you be able to split the refactoring into a first PR please?

stu-elastic:
@jdconrad here's the work to pull Metadata into it's own class, as requested: https://github.com/elastic/elasticsearch/pull/88401

stu-elastic:
@rjernst and @jdconrad, I've changed this PR to use infrastructure from https://github.com/elastic/elasticsearch/pull/88401 and https://github.com/elastic/elasticsearch/pull/88458.  Can you take another pass?

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.update.txt
```diff
@@ -18,3 +18,17 @@ class org.elasticsearch.painless.api.Json {
   String dump(def)
   String dump(def, boolean)
 }
+
+class org.elasticsearch.script.Metadata {
+    String getIndex()
+    String getId()
+    String getRouting()
+    long getVersion()
+    String getOp()
```
rjernst, line 27: We discussed having get and set op outside of the metadata on each script type that needs it, as it isn't really metadata about the document. What happened with that?

server/src/main/java/org/elasticsearch/script/UpdateMetadata.java
```diff
@@ -0,0 +1,113 @@
...
+        SET_ONCE_STRING,
+        TYPE,
+        SET_ONCE_STRING,
+        OP,
+        new FieldProperty<>(String.class, true, true, setValidator(VALID_UPDATE_OPS)),
+        TIMESTAMP,
+        SET_ONCE_LONG
+    );
+
+    protected static BiConsumer<String, String> setValidator(Set<String> valid) {
```
rjernst, line 57: nit: the naming here makes it look like a setter. perhaps `stringSetValidator`?

server/src/main/java/org/elasticsearch/script/UpdateMetadata.java
```diff
@@ -0,0 +1,113 @@
...
+        metadata.put(TYPE, type);
+        metadata.put(OP, op);
+        metadata.put(TIMESTAMP, timestamp);
+        return metadata;
+    }
+
+    @Override
+    public String getOp() {
+        String op = super.getOp();
+        if (LEGACY_NOOP_STRING.equals(op) || op == null) {
```
rjernst, line 98: I don't think we can do this translation for the "legacy" value here, it would be breaking. We should look at how/when to deprecate the two noop values outside of this PR.

server/src/main/java/org/elasticsearch/script/UpsertMetadata.java
```diff
@@ -0,0 +1,52 @@
...
+        metadata.put(INDEX, index);
+        metadata.put(ID, id);
+        metadata.put(OP, op);
+        metadata.put(TIMESTAMP, timestamp);
+        return metadata;
+    }
+
+    @Override
+    public String getRouting() {
+        throw new IllegalArgumentException("routing is unavailable for insert");
```
rjernst, line 45: UnsupportedOperationException would be more typical. We should use the same anywhere else we are failing a get/set method (a user should never see it since we woulnd't have allowed the method on scripting to begin with).

server/src/main/java/org/elasticsearch/script/UpdateCtxMap.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Map<String, Object> getSource() {
+        Map<String, Object> wrapped = super.getSource();
+        Object rawSource = wrapped.get(SOURCE);
+        if (rawSource instanceof Map<?, ?> map) {
+            return (Map<String, Object>) map;
+        }
+        throw new IllegalArgumentException(
```
rjernst, line 52: How could this occur? Since we control the ctx map, wouldn't this be a coding error? Should this be an assert instead?

rjernst (commented):
Thanks for the update Stu! After the other PRs this is much more straightforward and easier to understand. I think there is a backcompat issue to address, but otherwise I just have a few minor comments.

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.update.txt
stu-elastic: If it's on the script, then it conflicts with any variables named `op`, so it would be a backwards compatibility issue.

After thinking about it some more, it seems like a high cost to introduce another concept in addition to `field` and `metadata` for this one-off.  Maybe there's a better way to handle it?

server/src/main/java/org/elasticsearch/script/UpdateMetadata.java
```diff
@@ -0,0 +1,113 @@
...
+            // it to VALID_UPDATE_OPS
+            return "noop";
+        }
+        return op;
+    }
+
+    @Override
+    public void setOp(String op) {
+        if (LEGACY_NOOP_STRING.equals(op)) {
+            throw new IllegalArgumentException(LEGACY_NOOP_STRING + " is deprecated, use 'noop' instead");
```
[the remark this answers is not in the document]
stu-elastic: It is supported in the map, so there's no bwc issue, what do you think about removing the `" is deprecated"` language, disallow it in set but translate it in `getOp`?

That's asymmetric, but the addition of this class interface seems like the time and place to make that change.

server/src/main/java/org/elasticsearch/script/UpdateCtxMap.java
stu-elastic: An `assert` brings down the node.  Do you think we'd get the same feedback with an Exception without the customer pain? Perhaps `IllegalStateException`.

server/src/main/java/org/elasticsearch/script/UpsertMetadata.java
rjernst: > If we split the Update context then a script

I thought we were already planning to do that?

server/src/main/java/org/elasticsearch/script/UpdateMetadata.java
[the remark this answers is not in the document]
rjernst: Ok, I see now. I think your suggestion is good.

server/src/main/java/org/elasticsearch/script/UpdateCtxMap.java
stu-elastic: Changed to assert

server/src/main/java/org/elasticsearch/script/UpdateMetadata.java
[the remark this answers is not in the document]
stu-elastic: Ok, keeping it.

jdconrad (approved):
The new changes look fine to me. Please wait for @rjernst to have another look before committing.

server/src/main/java/org/elasticsearch/script/UpdateMetadata.java
stu-elastic: I suppose we should also restrict `null` in `setOp` like we're already doing with "none".

stu-elastic: In fact, there shouldn't be validation for `op` at all in the map, as it all gets translated to `UpdateOp.NONE`.

rjernst (approved):
LGTM.

PR #88334, opened 2022-07-07
Synthetic source error on script loads

tmgordeeva:
Scripts will now error out if they attempt to load source with synthetic source on. As part of this change, SourceLookup is now refactored to split the cases where we intend to actually load the source (ie fetch phase) and cases where we use a source already provided.


tmgordeeva:
I swore this PR existed somewhere, but checking on the status today I was surprised not to see it. If there is a duplicate somewhere somehow my bad!

tmgordeeva:
@romseygeek Nik's out, and I heard you might be able to take a look at this?

server/src/main/java/org/elasticsearch/index/query/SearchExecutionContext.java
```diff
@@ -476,9 +477,11 @@ public boolean containsBrokenAnalysis(String field) {
      */
     public SearchLookup lookup() {
         if (this.lookup == null) {
+            SourceFieldMapper sourceMapper = mappingLookup == null ? null : (SourceFieldMapper) mappingLookup.getMapper("_source");
```
nik9000, line 480: I might just do `boolean isSynthetic =` here so you don't have to reason about "null means false"

server/src/main/java/org/elasticsearch/search/lookup/SourceLookup.java
```diff
@@ -32,24 +32,10 @@
...
-    private XContentType sourceContentType;
-
-    public XContentType sourceContentType() {
-        return sourceContentType;
-    }
-
-    public int docId() {
-        return docId;
-    }
+    int docId();
```
nik9000, line 38: While you are here, can you add javadoc for these? It'll really help whoever comes after. I know I often don't. But I should...

server/src/main/java/org/elasticsearch/search/lookup/SourceLookup.java
```diff
@@ -32,24 +32,10 @@
 
 import static java.util.Collections.emptyMap;
 
-public class SourceLookup implements Map<String, Object> {
+public interface SourceLookup extends Map<String, Object> {
```
nik9000, line 35: I *wonder* if we should/can confine the `Map` extension to scripts. It'd an odd thing to do outside of script compatibility and this is used so many places. It's probably not worth it for this PR, but it's an interesting question.

server/src/main/java/org/elasticsearch/search/lookup/SearchLookup.java
```diff
@@ -53,7 +53,22 @@ public SearchLookup(
...
-        this.sourceLookup = new SourceLookup();
+        this.sourceLookup = new SourceLookup.Static();
+        this.fieldDataLookup = fieldDataLookup;
+    }
+
+    public SearchLookup(
+        Function<String, MappedFieldType> fieldTypeLookup,
+        BiFunction<MappedFieldType, Supplier<SearchLookup>, IndexFieldData<?>> fieldDataLookup,
+        boolean canReloadSource
+    ) {
```
nik9000, line 64: I tend not to like adding these extra ctors if I can get away without it. Is this temporary? I think we may have talked about this but I'm forgetting.

Either way, probably worth copying the javadoc from above.

server/src/main/java/org/elasticsearch/index/query/SearchExecutionContext.java
nik9000: Or, rather, "null means true"

server/src/main/java/org/elasticsearch/index/mapper/NestedValueFetcher.java
```diff
@@ -50,7 +50,7 @@ public List<Object> fetchValues(SourceLookup lookup, List<Object> includedValues
         for (Object entry : nestedValues) {
             // add this one entry only to the stub and use this as source lookup
             stub.put(nestedFieldName, entry);
-            SourceLookup nestedSourceLookup = new SourceLookup();
+            SourceLookup nestedSourceLookup = new SourceLookup.Static();
             nestedSourceLookup.setSource(filteredSource);
```
nik9000, line 54: Seeing this makes me think "SourceLookup.Static should take the source as a parameter - but I think that'd only work sometimes, right? In the fetch phase we don't have it available right away, I think.

docs/changelog/88334.yaml
```diff
@@ -0,0 +1,5 @@
+pr: 88334
+summary: Synthetic source error on script loads
+area: Geo
```
nik9000, line 3: Hmmm.... I think maybe the thing made a mistake. This isn't in Geo at all.

server/src/test/java/org/elasticsearch/search/aggregations/bucket/terms/SignificantTermsAggregatorTests.java
```diff
@@ -408,6 +408,49 @@ public void testFieldAlias() throws IOException {
         }
     }
 
+    public void testFieldBackground() throws IOException {
```
nik9000, line 411: I think this also snuck in.

server/src/test/java/org/elasticsearch/search/fetch/subphase/FieldFetcherTests.java
```diff
@@ -1162,7 +1162,7 @@ public void testFetchMetadataFieldWithSourceDisabled() throws IOException {
             IndexSearcher searcher = newSearcher(iw);
             LeafReaderContext readerContext = searcher.getIndexReader().leaves().get(0);
             fieldFetcher.setNextReader(readerContext);
-            SourceLookup sourceLookup = new SourceLookup();
+            SourceLookup sourceLookup = new SourceLookup.Static();
             sourceLookup.setSegmentAndDocument(readerContext, 0);
```
nik9000, line 1166: Oh this is fun. `Static` isn't going to be able to do anything with the `readerContext` or the docId. But this is how we pass those things into the fetching framework, right?

server/src/main/java/org/elasticsearch/search/lookup/SourceLookup.java
```diff
@@ -204,72 +94,256 @@ public List<Object> extractRawValuesWithoutCaching(String path) {
...
+        public void clear() {
+            throw new UnsupportedOperationException();
+        }
     }
 
-    @Override
-    @SuppressWarnings("rawtypes")
-    public void putAll(Map m) {
-        throw new UnsupportedOperationException();
+    class Static extends AbstractSourceLookup {
```
nik9000, line 305: These both deserve javadoc please.

I wonder if `PreLoaded` is a better name than `Static`. My mind jumps to java's `static` keyword and I keep thinking "there is only one of this". But static here means something more like "unchanging". 

Speaking of, how hard would be be to make the `Static` version truly immutable? Not worth it in this PR, I'd wager, but maybe actually good?

modules/percolator/src/main/java/org/elasticsearch/percolator/PercolatorHighlightSubFetchPhase.java
```diff
@@ -86,7 +87,7 @@ public void process(HitContext hit) throws IOException {
                                 percolatorLeafReaderContext,
                                 slot
                             );
-                            subContext.sourceLookup().setSource(document);
+                            subContext.sourceLookup().setSourceProvider(new SourceLookup.BytesSourceProvider(document));
```
romseygeek, line 90: One of these days we will get rid of these setters.  Ah well, progress not perfection!

server/src/main/java/org/elasticsearch/index/query/SearchExecutionContext.java
```diff
@@ -481,12 +483,20 @@ public boolean containsBrokenAnalysis(String field) {
...
```
romseygeek, line 486: Can we make this a new `getSourceProvider()` method on `MappingLookup`?  That way we can easily add in future enhancements like a nested source provider.

server/src/main/java/org/elasticsearch/search/lookup/SearchLookup.java
romseygeek: From a quick look it appears that all of the uses of the extra constructor are in test code, which makes me agree with Nik.  Let's nuke it - it makes the PR a bit noisier but I think it will save confusion further down the line.

server/src/main/java/org/elasticsearch/search/lookup/SearchLookup.java
```diff
@@ -53,14 +53,22 @@ public class SearchLookup {
      */
     public SearchLookup(
         Function<String, MappedFieldType> fieldTypeLookup,
-        TriFunction<MappedFieldType, Supplier<SearchLookup>, MappedFieldType.FielddataOperation, IndexFieldData<?>> fieldDataLookup
+        TriFunction<MappedFieldType, Supplier<SearchLookup>, MappedFieldType.FielddataOperation, IndexFieldData<?>> fieldDataLookup,
+        SourceLookup sourceLookup
```
romseygeek, line 57: I think I'd prefer to pass the `SourceProvider` as a parameter here, rather than the `SourceLookup` directly

server/src/main/java/org/elasticsearch/search/lookup/SourceLookup.java
```diff
@@ -62,101 +58,32 @@ public int docId() {
...
-            if (context.reader()instanceof SequentialStoredFieldsLeafReader lf) {
-                // Avoid eagerly loading the stored fields reader, since this can be expensive
-                Supplier<StoredFieldsReader> supplier = new MemoizedSupplier<>(lf::getSequentialStoredFieldsReader);
-                fieldReader = (d, v) -> supplier.get().visitDocument(d, v);
-            } else {
-                fieldReader = context.reader()::document;
-            }
+        if (sourceProvider instanceof ReaderSourceProvider) {
+            ((ReaderSourceProvider) sourceProvider).setSegmentAndDocument(context, docId);
         }
```
romseygeek, line 71: Let's add a `setSegmentAndDocument()` method to the `SourceProvider` interface as well.

romseygeek (changes requested):
This looks great, thanks @tmgordeeva!  I left a few comments but I think we're close.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/terms/SignificanceLookup.java
```diff
@@ -74,9 +75,22 @@ interface BackgroundFrequencyForLong extends Releasable {
...
+            Query rewrittenMatchAll = context.filterQuery(matchAllDocsQuery);
+            if (rewrittenMatchAll != matchAllDocsQuery) {
+                this.backgroundFilter = rewrittenMatchAll;
+            } else {
+                this.backgroundFilter = null;
+            }
+        } else {
+            Query rewrittenBackground = context.filterQuery(backgroundQuery);
+            this.backgroundFilter = rewrittenBackground;
+        }
```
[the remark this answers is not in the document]
romseygeek: Oops, leftover from a previous round.  Ignore this comment!

docs/changelog/88334.yaml
```diff
@@ -0,0 +1,5 @@
...
+type: bug
```
romseygeek, line 4: Let's make this `area: Search` and `type: enhancement`

server/src/main/java/org/elasticsearch/search/lookup/SourceLookup.java
```diff
@@ -24,27 +24,23 @@
...
 import java.util.Set;
 import java.util.function.Supplier;
 
-import static java.util.Collections.emptyMap;
-
-public class SourceLookup implements Map<String, Object> {
-
-    private LeafReader reader;
-    private CheckedBiConsumer<Integer, FieldsVisitor, IOException> fieldReader;
+public class SourceLookup {
```
romseygeek, line 33: Can we double-check that removing the Map interface here doesn't break scripts that look things up via `_source['foo']`? There's a lot of dynamic typing that goes on in scripts and I'm a bit nervous that this will silently cause some issues.  Maybe @stu-elastic or @jdconrad have a view? If it works, then that's awesome as having it here has been annoying me for several years now :)

romseygeek (commented):
Nearly there, I would like to double check one thing re script access to `_source`.

server/src/main/java/org/elasticsearch/search/lookup/SourceLookup.java
jdconrad: I looked through this with @stu-elastic, and this looks like this should be a safe change. Everywhere we allow source access we do through a `DynamicMap` where we call the `source` method to get the source `Map` directly instead of using `SourceLookup`. The one edge case I can see is a custom plugin may have its own specially defined context; however, we don't guarantee any stable api right now.

romseygeek: I think we need to also change the construction of the `asMap` member of `LeafSearchLookup`.  Although this hasn't caused any test failures, which either means we have a gap in test coverage or that this isn't actually used anymore.

romseygeek (approved):
Let's get this merged! Thanks for all the iterations @tmgordeeva

PR #88335, opened 2022-07-07
Support run_as another user when granting API keys

ywangd:
Previoulsy, in order to grant an API key to a user, the user's
credentials (either username/password or access token) must be provided.
However, this is not always possible for certain setups, e.g. when
all end-users are proxied by a service user with the run_as feature.
This PR adds run_as support for granting API keys so that the user whose
credentials are provided (in the request body) can run-as another user
for whom the API key will be eventually granted.

The run-as behaviour is stricter in that it reports error if it is not supported 
by the provided authentication, e.g. anonymous, chained run-as. This is
different that used in authenticating requests, where unsupported run-as is
ignored and logged.


ywangd:
This PR intentionally skips the following two things to reduce scope (also because it is self-sufficient in its current state):
1. Support `client_credentials` for `grant_type`. This is potentially a convenience feature when the user specified in the request body is the same as the request caller. It is left out because (a) It has implications in that it enables more authentication types (e.g. service account, kerberos realm) to grant API keys; (b) we don't have code to support run-as from an existing authentication object.
2. Similar run-as support for activate user profile API. This will reuse most of the code changes in this PR. But it requires changes to a different REST point and associated tests, docs, therefore better to be a separate follow-up.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/apikey/ApiKeySingleNodeTests.java
```diff
@@ -238,6 +255,133 @@ public void testGetApiKeyWorksForTheApiKeyItself() {
...
+        grantApiKeyRequest3.getGrant().setAccessToken(new SecureString(oAuth2Token3.accessToken().toCharArray()));
+        grantApiKeyRequest3.getGrant().setRunAsUsername("user2");
+        final ElasticsearchStatusException e3 = expectThrows(
+            ElasticsearchStatusException.class,
+            () -> client().execute(GrantApiKeyAction.INSTANCE, grantApiKeyRequest3).actionGet()
+        );
+        assertThat(e3.getMessage(), containsString("the provided grant credentials do not support run-as"));
+
+        // Failure 4: user1 run-as user4 and creates a token. The token is used for GrantApiKey. But the token loses the run-as
+        // privileges when it is used.
```
ywangd, line 326: This can happen today. So this PR is also a minor bug fix.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/TransportGrantAction.java
```diff
@@ -50,11 +59,48 @@ protected void executeWithGrantAuthentication(GrantRequest grantRequest, ActionL
...
+                    }
+                }
+            }, listener::onFailure);
+
+            if (runAsUsername != null) {
+                threadContext.putHeader(AuthenticationServiceField.RUN_AS_USER_HEADER, runAsUsername);
+            }
             authenticationService.authenticate(
                 actionName,
                 grantRequest,
```
ywangd, line 101: I'd like to change this `grantRequest` to an `AuthenticateRequest` similar to how it is done in the authZ code. But it is not possible currently because `AuthenticateRequest` requires username and it may not be known if the provided credential is an accessToken. I think the right thing to do is to refactor `AuthenticateRequest` to drop the `username` field. We no longer support transport client. It is not possible for end users to set this field to a different value. `AuthenticateRequest` should always work for the current authentication object and this logic should just be enforced regardless of the username. I will raise another PR to make this change if you agree with the proposal.

tvernum: The change to drop username from `AuthenticateRequest` seems reasonable.
I'm not sure that I see the significance of using Authenticate here though. We could, but it doesn't seem to be strictly better.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/apikey/ApiKeySingleNodeTests.java
```diff
@@ -238,6 +255,133 @@ public void testGetApiKeyWorksForTheApiKeyItself() {
...
+        final GrantApiKeyRequest grantApiKeyRequest = buildGrantApiKeyRequest("user1", user1Password, "user2", securityClient);
+        final CreateApiKeyResponse createApiKeyResponse = client().execute(GrantApiKeyAction.INSTANCE, grantApiKeyRequest).actionGet();
+        final String apiKeyId = createApiKeyResponse.getId();
+        final String base64ApiKeyKeyValue = Base64.getEncoder()
+            .encodeToString((apiKeyId + ":" + createApiKeyResponse.getKey().toString()).getBytes(StandardCharsets.UTF_8));
+        assertThat(securityClient.getApiKey(apiKeyId).getUsername(), equalTo("user2"));
+        // The API key has privileges (inherited from user2) to check cluster health
+        client().filterWithHeader(Map.of("Authorization", "ApiKey " + base64ApiKeyKeyValue))
+            .execute(ClusterHealthAction.INSTANCE, new ClusterHealthRequest())
+            .actionGet();
```
tvernum, line 277: Can we have a negative test here too?
That is, give `user1_role` a privilege that is easy to test for. Don't give that privilege to `user2_role`. Check that the API Key doesn't have the privilege.

One of the possible bugs that could exist (or be introduced in the future) here is that the API key has the `monitor` privilege due to the facilitating user (`test_user`) which has cluster `ALL`.
Testing that the API Key _doesn't_ have something it isn't supposed to have helps minimize that risk.

tvernum: And can we also have a test with limiting roles? When there's 3 sets of users involved, there's a risk that something causes us to pull in the wrong set of roles, and I'd like to test more extensively.

tvernum: > So I assume you mean the API key should have its own role descriptors?

Yes, that's what I meant. Did I use the wrong words? I can never remember which one we call `limiting` and `limited`.

tvernum (approved):

ywangd:

PR #88336, opened 2022-07-07
Add build_flavor back to info api rest response

### LINKED ISSUE #88318: Regression in `info` API response by removing `version.build_flavor` field
In https://github.com/elastic/elasticsearch/pull/85796 the `version.build_flavor` field was removed from the `/` API response. This API is marked as `stable` in the [API spec stub](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/info.json) and thus this removal should be considered a regression. The removal has started [causing issues for users of clients](https://github.com/elastic/elasticsearch-java/issues/335) when they connect to 8.3.0+ of Elasticsearch.

My proposed fix is to add back the field with a static value of `default` and to change the removal of the field to the next major version instead.

rjernst:
The build_flavor was previously removed since it is no longer relevant;
only the default distribution now exists. However, the removal of build
flavor included removing it from the version information on the info
response for the root path. This API is supposed to be stable, so
removing that key was a compatibility break. This commit adds the
build_flavor back to that API, hardcoded to `default`. Additionally, a
test is added to ensure the key exists going forward, until it can be
properly deprecated.

closes #88318


grcevski (approved):
LGTM!

sethmlarson (approved):
LGTM

PR #88339, opened 2022-07-07
Remove duplicate definition of checkstyle version in use

breskeby:
We only rely on the checkstyle version in the buildLibs.toml gradle version catalogue with this change.
Also added some hints for gradle best practices.

This is an aftermath of #88283

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -48,6 +48,16 @@ in our build logic to resolve this.
 
 **The Elasticsearch build will fail if any deprecated Gradle API is used.**
 
+### Follow Gradle best practices
+
+Tony Robalik has compiled a good list of rules that aligns with ours when it comes to writing and maintaining elasticsearch
+gradle build logic at http://autonomousapps.com/blog/rules-for-gradle-plugin-authors.html.
+Our current build does not yet tick off all those rules everywhere but the ultimate goal is to follow these principles.
+The reasons for following those rules besidesbetter readability or maintenance are also the goal to support newer gradle
+features that we will benefit from in terms of performance and reliability.
+E.g. [configuration-cache support](https://github.com/elastic/elasticsearch/issues/57918), [Project Isolation]([https://gradle.github.io/configuration-cache/#project_isolation) or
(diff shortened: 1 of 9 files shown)


BUILDING.md
```diff
@@ -48,6 +48,16 @@ in our build logic to resolve this.
 
 **The Elasticsearch build will fail if any deprecated Gradle API is used.**
 
+### Follow Gradle best practices
+
+Tony Robalik has compiled a good list of rules that aligns with ours when it comes to writing and maintaining elasticsearch
+gradle build logic at http://autonomousapps.com/blog/rules-for-gradle-plugin-authors.html.
+Our current build does not yet tick off all those rules everywhere but the ultimate goal is to follow these principles.
+The reasons for following those rules besidesbetter readability or maintenance are also the goal to support newer gradle
```
pugnascotia, line 56:
```suggestion
The reasons for following those rules besides better readability or maintenance are also the goal to support newer gradle
```

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/precommit/CheckstylePrecommitPlugin.java
```diff
@@ -85,11 +86,13 @@ public void execute(Task task) {
...
 
         DependencyHandler dependencies = project.getDependencies();
-        String checkstyleVersion = VersionProperties.getVersions().get("checkstyle");
         Provider<String> conventionsDependencyProvider = project.provider(
             () -> "org.elasticsearch:build-conventions:" + project.getVersion()
         );
-        dependencies.add("checkstyle", "com.puppycrawl.tools:checkstyle:" + checkstyleVersion);
+        dependencies.addProvider("checkstyle", project.provider(() -> {
+            var versionCatalog = project.getExtensions().getByType(VersionCatalogsExtension.class).named("buildLibs");
+            return versionCatalog.findLibrary("checkstyle").get().get();
```
mark-vieira, line 94: Why did they wrap a `Provider` in an `Optional` when the `Provider` API itself already has the notion of a "missing" value 🤦

pugnascotia (approved):
LGTM

PR #88340, opened 2022-07-07
Fix performance regression in DataTierAllocationDecider

original-brownbear:
In #87735 we moved to using streams here. This results in a visible
regression in the many shards benchmarks that gets bigger the larger
the cluster is.
Using streams here adds a lot of alloction and indirection but does not
come with any advantages since `DiscoveryNodes` does not implement
`stream` so it's just an indirection on top of the existing iterator
that gets instantiated for every shard and then runs a compartively
short loop over the nodes.

-> move this back to iterator

relates #77466


fcofdez (approved):
LGTM. Thanks for fixing this @original-brownbear!

PR #88344, opened 2022-07-07
Speed up creating new IndexMetaDataGenerations without removed snapshots

arteam:
* Pre-collect all new blobUuids to a set before removing stale identifiers. This avoids iterating over the `lookup` map and its each entry's map for every identifier (which results in cubic time complexity!).
* Avoid creating copies of maps just to stale data, instead populate them on the go.


server/src/main/java/org/elasticsearch/repositories/IndexMetaDataGenerations.java
```diff
@@ -133,11 +135,18 @@ public IndexMetaDataGenerations withAddedSnapshot(
...
+        for (Map.Entry<SnapshotId, Map<IndexId, String>> e : lookup.entrySet()) {
+            if (snapshotIds.contains(e.getKey()) == false) {
+                updatedIndexMetaLookup.put(e.getKey(), e.getValue());
+                updatedBlobUuids.addAll(e.getValue().values());
+            }
+        }
+        final Map<String, String> updatedIndexMetaIdentifiers = identifiers.entrySet()
+            .stream()
+            .filter(e -> updatedBlobUuids.contains(e.getKey()))
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```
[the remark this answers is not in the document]
arteam: Unfortunately, It *is* immutable https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/repositories/IndexMetaDataGenerations.java#L50

We need to create a new instance.

original-brownbear (approved):
LGTM but maybe rename that variable to make this a little easier to read.
I can see how this is running for longer than necessary for extreme deployments.

server/src/main/java/org/elasticsearch/repositories/IndexMetaDataGenerations.java
```diff
@@ -133,11 +134,18 @@ public IndexMetaDataGenerations withAddedSnapshot(
...
      */
     public IndexMetaDataGenerations withRemovedSnapshots(Collection<SnapshotId> snapshotIds) {
-        final Map<SnapshotId, Map<IndexId, String>> updatedIndexMetaLookup = new HashMap<>(lookup);
-        updatedIndexMetaLookup.keySet().removeAll(snapshotIds);
-        final Map<String, String> updatedIndexMetaIdentifiers = new HashMap<>(identifiers);
-        updatedIndexMetaIdentifiers.keySet()
-            .removeIf(k -> updatedIndexMetaLookup.values().stream().noneMatch(identifiers -> identifiers.containsValue(k)));
+        final Map<SnapshotId, Map<IndexId, String>> updatedIndexMetaLookup = lookup.entrySet()
+            .stream()
+            .filter(e -> snapshotIds.contains(e.getKey()) == false)
```
henningandersen, line 139: This is in essence `List.contains`. I wonder if this might be slower than copying it to a Set and then looking up against that? A delete of a big list of snapshots could hurt here maybe?

henningandersen (commented):
I wonder if we can improve this a bit more?

server/src/main/java/org/elasticsearch/repositories/IndexMetaDataGenerations.java
arteam: Yeah, I was thinking about fixing it in a separate PR by creating a `Set` in SnapshotsService https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/snapshots/SnapshotsService.java#L2420 where `deleteSnapshots` gets called.

server/src/main/java/org/elasticsearch/snapshots/SnapshotsService.java
```diff
@@ -2401,7 +2401,7 @@ private void deleteSnapshotsFromRepository(
     ) {
         if (repositoryOperations.startDeletion(deleteEntry.uuid())) {
             assert currentlyFinalizing.contains(deleteEntry.repository());
-            final List<SnapshotId> snapshotIds = deleteEntry.getSnapshots();
+            final Set<SnapshotId> snapshotIds = new HashSet<>(deleteEntry.getSnapshots());
```
original-brownbear, line 2404: This is really subtle now, I think if we want to optimize this we should either make the signature of `withRemovedSnapshots` take a set or copy to a set in that method. Making the signature a set is probably cleanest

server/src/main/java/org/elasticsearch/repositories/IndexMetaDataGenerations.java
```diff
@@ -133,12 +134,19 @@ public IndexMetaDataGenerations withAddedSnapshot(
...
-        updatedIndexMetaLookup.keySet().removeAll(snapshotIds);
-        final Map<String, String> updatedIndexMetaIdentifiers = new HashMap<>(identifiers);
-        updatedIndexMetaIdentifiers.keySet()
-            .removeIf(k -> updatedIndexMetaLookup.values().stream().noneMatch(identifiers -> identifiers.containsValue(k)));
-        return new IndexMetaDataGenerations(updatedIndexMetaLookup, updatedIndexMetaIdentifiers);
+        final Map<SnapshotId, Map<IndexId, String>> retainedIndexMetaLookup = lookup.entrySet()
+            .stream()
+            .filter(e -> snapshotIds.contains(e.getKey()) == false)
+            .collect(Collectors.toUnmodifiableMap(Map.Entry::getKey, Map.Entry::getValue));
+        final Set<String> retainedBlobUUIDs = retainedIndexMetaLookup.values()
```
original-brownbear, line 141: We did you change this to two loops now, seems the first version that I LGTMed was more efficient because we wouldn't loop `retainedIndexMetaLookup` twice?

server/src/main/java/org/elasticsearch/snapshots/SnapshotsService.java
henningandersen: I agree. I would also prefer to just copy the `snapsthotIds` inside `withRemovedSnapshots` for now, I do not think it really matters that we copy it there?

original-brownbear: I don't think performance here matters that much, that method is not running all that often either. Lets keep it simple IMO

original-brownbear (approved):
LGTM though no need to backport to 8.3 IMO

arteam:
Thanks Armin!

PR #88346, opened 2022-07-07
Updatable API keys - noop check

n1v0lg:
This PR adds a noop check for API key updates. If we detect a noop
update, i.e., an update that does not result in any changes to the
existing doc, we skip the index step and return `updated = false` in
the response.

This PR also extends test coverage around various corner cases.


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -969,6 +966,71 @@ public void logRemovedField(String parserName, Supplier<XContentLocation> locati
         }
     }
 
+    private void doUpdateApiKey(
```
n1v0lg, line 969: One more iteration on this: I'm coming around to inlining 😅 

I've inlined the `executeRequest` method and in-lined building the request here. The only "functional" change is the noop change further down.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -969,6 +966,71 @@ public void logRemovedField(String parserName, Supplier<XContentLocation> locati
...
+                    targetDocVersion,
+                    request.getMetadata()
+                )
+            )
+            .setIfSeqNo(currentVersionedDoc.seqNo())
+            .setIfPrimaryTerm(currentVersionedDoc.primaryTerm())
+            .setOpType(DocWriteRequest.OpType.INDEX)
+            .request();
+
+        final boolean noop = indexRequest.source().equals(currentVersionedDoc.source());
```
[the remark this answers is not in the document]
n1v0lg: I'm going for this approach for a noop check because it's very simple, and robust. If we change what fields can be updated or evolve the API key doc model, we can still rely on the noop check working here.

We could for sure optimize this (for instance by weaving the noop check into `buildUpdatedDocument`) but my intuition is that would be premature. Could likewise avoid passing around the source for the doc with the doc (it's redundant logically and on the heap) but then we'd have some form of conversion here.

I'm open to changing this but wanted to get your thoughts first.

n1v0lg:
Build failures unrelated.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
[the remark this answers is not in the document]
ywangd: > it's very simple, and robust

It's simple. But I don't think it's robust. This relies on the order of `buildUpdatedDocument` to match exactly the index mapping for API Key document. They are probably identical right now. But we never intend for them to be consistent. There is also no guarantee that no optimisation on lucene side could re-arrange how bytes are read from the disc. Overall I think matching these two (potentially big) string blobs for equality is not desirable.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -2041,13 +2248,27 @@ private void createNativeRealmUser(
...
+    }
+
     private RoleDescriptor putRoleWithClusterPrivileges(final String nativeRealmRoleName, String... clusterPrivileges)
         throws InterruptedException, ExecutionException {
         final PutRoleRequest putRoleRequest = new PutRoleRequest();
         putRoleRequest.name(nativeRealmRoleName);
-        for (final String clusterPrivilege : clusterPrivileges) {
-            putRoleRequest.cluster(clusterPrivilege);
-        }
+        putRoleRequest.cluster(clusterPrivileges);
```
n1v0lg, line 2271: Fix test to use all cluster privs, not just last.

n1v0lg:
@elasticmachine run elasticsearch-ci/part-2-fips plz. Unrelated failure.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -490,7 +492,87 @@ static XContentBuilder buildUpdatedDocument(
...
+                )
+            );
+            if (sourceRealm.getDomain().equals(currentRealmDomain) == false) {
+                return false;
+            }
+        } else {
+            if (currentCreator.get("realm_domain") != null) {
+                return false;
+            }
+        }
```
ywangd, line 541: In a future PR, it is worthwhile to extract some of this logic to a separate method (something like `ApiKeyDoc.getOwnerSubject`) which can be reused in other places like `validateApiKeyExpiration`.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -490,7 +492,87 @@ static XContentBuilder buildUpdatedDocument(
...
+                RoleReference.ApiKeyRoleType.ASSIGNED
+            );
+            if (false == (newRoleDescriptors.size() == currentRoleDescriptors.size()
+                && new HashSet<>(newRoleDescriptors).equals(new HashSet<>(currentRoleDescriptors)))) {
+                return false;
+            }
+        }
+
+        assert userRoles != null;
+        final List<RoleDescriptor> currentLimitedByRoleDescriptorRoles = parseRoleDescriptorsBytes(
```
ywangd, line 569: Nit: The name `currentLimitedByRoleDescriptorRoles` is a bit awkward. It should just be `currentLimitedByRoleDescriptors`. Also, `userRoles` really should be `userRoleDescriptors` since the code has distinct concepts for role, role name and role descriptor.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -490,7 +492,87 @@ static XContentBuilder buildUpdatedDocument(
 
         addCreator(builder, authentication);
 
-        return builder.endObject();
+        return new ApiKeyDocBuilderWithNoopFlag(
+            builder.endObject(),
+            isNoop(currentApiKeyDoc, targetDocVersion, authentication, request, userRoles)
+        );
```
ywangd, line 498: This works. But I think it can be simplified a bit if:
1. Perform `isNoop` check in the beginning of the method. So we can avoid creating and preparing the builder all together if no change.
2. Remove the `ApiKeyDocBuilderWithNoopFlag` record type and just return `null` if there is no change or the actual builder if there are any changes.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -1679,6 +1725,140 @@ public void testUpdateApiKeyAccountsForSecurityDomains() throws ExecutionExcepti
...
+                        new String[] { "ALL" },
+                        new RoleDescriptor.IndicesPrivileges[] {
+                            RoleDescriptor.IndicesPrivileges.builder()
+                                .indices("*")
+                                .allowRestrictedIndices(true)
+                                .privileges("ALL")
+                                .build() },
+                        null
+                    )
+                ),
```
ywangd, line 1856: For this assumption to uphold for future changes, it would be better if we check that it indeed results into a noop update before this check. Otherwise, any future change to the test role will essentially make this test not doing anything.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -490,7 +492,87 @@ static XContentBuilder buildUpdatedDocument(
...
+                && new HashSet<>(newRoleDescriptors).equals(new HashSet<>(currentRoleDescriptors)))) {
+                return false;
+            }
+        }
+
+        assert userRoles != null;
+        final List<RoleDescriptor> currentLimitedByRoleDescriptorRoles = parseRoleDescriptorsBytes(
+            request.getId(),
+            apiKeyDoc.limitedByRoleDescriptorsBytes,
+            RoleReference.ApiKeyRoleType.LIMITED_BY
```
ywangd, line 572: This might be OK though the behaviour is not perfect.

The method `parseRoleDescriptorsBytes` replaces old superuser role definition (7.x) with the new one (8.0). This means it can return a role descriptor that is different from what is actually stored. If we only test against the returned role descriptor, we may miss an update because the equality does not apply to the actual stored descriptor. 

In practice this can only happen in a mix version cluster (8.4 + 7.17). If update skips for a `superuser` limitedByRoleDescriptor. It may not be too big of a deal especially since the legacy descriptor will be handled correctly when loaded (at authentication time). 

Alternatively, we can check whether the current doc has version 7.x and the returned a descriptor is a `superuser`. If answers to both are `true`, it is _not_ a noop (because we will be updating the legacy superuser descriptor to the new one). This would also means after update, the key will lose the legacy superuser behaviour even when it is used on a 7.17 node. But that is probably OK (and even desired due to consistency). 

It is definitely an edge case. I don't really have a strong opinion. But at least it worths some code level comment. What do you think?

ywangd (commented):
I have a few comments.

I appreciate your exhaustive work for the `isNoop` method.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
n1v0lg: Performing the noop check at the beginning of the method is how I had this structured initially, so I'm happy to switch it back to that. My thinking around always building was that it makes the method more generic and gives it simpler semantics: it creates the object and tells the caller if it's a noop -- then it's up to the caller to decide what to do with it.

This is just to give context. I'm good with your suggestion 👍

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
n1v0lg: Oh whoops `currentLimitedByRoleDescriptorRoles` is definitely a copy&mishap. Thanks for catching 👍 

> Also, userRoles really should be userRoleDescriptors since the code has distinct concepts for role, role name and role descriptor.

I'm good with that, I went with `userRoles` because that's what it was called before the work on updatable API keys elsewhere in this class.

ywangd: > I went with userRoles because that's what it was called before the work on updatable API keys elsewhere in this class.

I figured. I wasn't a fan of this name in other places as well. But we never bothered to change them ...

n1v0lg:

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -433,23 +431,31 @@ static XContentBuilder newDocument(
...
         return builder.endObject();
     }
 
-    static XContentBuilder buildUpdatedDocument(
+    // package private for testing
+
+    /**
+     * @return `null` if the update is a noop, i.e., if no changes to `currentApiKeyDoc` are required
+     */
+    XContentBuilder buildUpdatedDocument(
```
ywangd, line 448: As suggested in the other comment: I think it's helpful to rename this to `maybeBuildUpdatedDocument`.

ywangd (approved):
LGTM

PR #88347, opened 2022-07-07
[ML] fix NLP question_answering task when best answer is only one token

benwtrent:
There are scenarios when question_answering find the best start/end token and they are the same token. An example of this is:

`context: "My name is Ben and I live in London"` `question: "Where do I live?"`

The correct answer here is `London` and its a single token. Without this fix, we will return `in London` with a lower probability.


benwtrent:
@elasticmachine update branch

droberts195 (approved):
LGTM

PR #88362, opened 2022-07-08
Fix test memory leak

### LINKED ISSUE #88337: [CI] MlDistributedFailureIT testClusterWithTwoMlNodes_StopsDatafeed_GivenJobFailsOnReassign failing
There have been two instances of this today (both in PR builds) that blew up due to an OOME.

**Build scan:**
https://gradle-enterprise.elastic.co/s/ryfsljqpbyjlm/tests/:x-pack:plugin:ml:internalClusterTest/org.elasticsearch.xpack.ml.integration.MlDistributedFailureIT/testClusterWithTwoMlNodes_StopsDatafeed_GivenJobFailsOnReassign

**Reproduction line:**
`./gradlew ':x-pack:plugin:ml:internalClusterTest' --tests "org.elasticsearch.xpack.ml.integration.MlDistributedFailureIT.testClusterWithTwoMlNodes_StopsDatafeed_GivenJobFailsOnReassign" -Dtests.seed=A3E13461A939BDF6 -Dtests.locale=ar-IQ -Dtests.timezone=America/Porto_Velho -Druntime.java=17`

**Applicable branches:**
master

**Reproduces locally?:**
No

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.xpack.ml.integration.MlDistributedFailureIT&tests.test=testClusterWithTwoMlNodes_StopsDatafeed_GivenJobFailsOnReassign

**Failure excerpt:**
```
java.lang.Exception: Test abandoned because suite timeout was reached.

  at __randomizedtesting.SeedInfo.seed([A3E13461A939BDF6]:0)

```

grcevski:
I had incorrectly made the assumption that the createComponents will only be called once for ILM plugin initialization. It appears that this isn't the case when we run tests.

This PR fixes how we initialize the ILM immutable state handlers, if the LifecyclePlugin gets re created, only the latest state is reflected in the immutable state handler provider.

Fixes https://github.com/elastic/elasticsearch/issues/88337

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/ActionModule.java
+++ b/server/src/main/java/org/elasticsearch/action/ActionModule.java
@@ -262,9 +262,6 @@ import org.elasticsearch.common.settings.SettingsFilter;
 import org.elasticsearch.gateway.TransportNodesListGatewayStartedShards;
 import org.elasticsearch.health.GetHealthAction;
 import org.elasticsearch.health.RestGetHealthAction;
-import org.elasticsearch.immutablestate.ImmutableClusterStateHandler;
-import org.elasticsearch.immutablestate.ImmutableClusterStateHandlerProvider;
-import org.elasticsearch.immutablestate.action.ImmutableClusterSettingsAction;
 import org.elasticsearch.index.seqno.GlobalCheckpointSyncAction;
 import org.elasticsearch.index.seqno.RetentionLeaseActions;
 import org.elasticsearch.indices.SystemIndices;
@@ -276,7 +273,6 @@ import org.elasticsearch.persistent.StartPersistentTaskAction;
 import org.elasticsearch.persistent.UpdatePersistentTaskStatusAction;
 import org.elasticsearch.plugins.ActionPlugin;
 import org.elasticsearch.plugins.ActionPlugin.ActionHandler;
-import org.elasticsearch.plugins.PluginsService;
 import org.elasticsearch.plugins.interceptor.RestInterceptorActionPlugin;
 import org.elasticsearch.rest.RestController;
 import org.elasticsearch.rest.RestHandler;
@@ -890,23 +886,6 @@ public class ActionModule extends AbstractModule {
             }
         }
         registerHandler.accept(new RestCatAction(catActions));
-    }
-
-    /**
-     * Initializes the immutable cluster state handlers for Elasticsearch and it's modules/plugins
-     *
-     * @param pluginsService needed to load all modules/plugins immutable state handlers through SPI
-     */
-    public void initImmutableClusterStateHandlers(PluginsService pluginsService) {
-        List<ImmutableClusterStateHandler<?>> handlers = new ArrayList<>();
-
-        List<? extends ImmutableClusterStateHandlerProvider> pluginHandlers = pluginsService.loadServiceProviders(
-            ImmutableClusterStateHandlerProvider.class
-        );
-
-        handlers.add(new ImmutableClusterSettingsAction(clusterSettings));
-        pluginHandlers.forEach(h -> handlers.addAll(h.handlers()));
-        // Initialize the controller when merged
     }
 
     @Override
--- a/server/src/main/java/org/elasticsearch/action/ActionModule.java
+++ b/server/src/main/java/org/elasticsearch/action/ActionModule.java
@@ -276,7 +276,6 @@ import org.elasticsearch.persistent.StartPersistentTaskAction;
 import org.elasticsearch.persistent.UpdatePersistentTaskStatusAction;
 import org.elasticsearch.plugins.ActionPlugin;
 import org.elasticsearch.plugins.ActionPlugin.ActionHandler;
-import org.elasticsearch.plugins.Plugin;
 import org.elasticsearch.plugins.PluginsService;
 import org.elasticsearch.plugins.interceptor.RestInterceptorActionPlugin;
 import org.elasticsearch.rest.RestController;
@@ -421,7 +420,6 @@ import org.elasticsearch.usage.UsageService;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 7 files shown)


rjernst:
Is there a way we can avoid making production code mutable just for tests?

rjernst:
I think the issue is the use of statics. We don't use them in other places for such registration. These handlers should be captured in instance variables specific to the node, so that many node instances (in tests) can be created.

grcevski:
So I flipped this around a bit and I removed the use of statics. I also added an example code of how this is intended to work in ActionModule.

Essentially, the SPI creates the service provides independently of the plugin constructor, so if we want to avoid any statics, we need a way to pass down what the handlers need to be constructed with. However, the best place to construct the handlers is during the plugin constructor (or createComponents), since then we have all what we'd need to make the handler work.

To avoid passing a superset of what we might need for construction of the immutable state handlers, I did the following:

- Let the plugin create the immutable handler just like any other action handler during construction and keep a reference to it
- When and if a handler provider is created, it will ask the plugin to deliver a collection of the handlers it likes to expose.
- The provider now receives a list of plugins it 'serves' and it asks the plugins for what handlers they'll expose. We have the plugin instance at the time we call for the providers to be created, since we have already loaded the plugins by then.

server/src/main/java/org/elasticsearch/immutablestate/ImmutableClusterStateHandlerProvider.java
```diff
@@ -21,5 +23,7 @@ public interface ImmutableClusterStateHandlerProvider {
      *
      * @return a list of ${@link ImmutableClusterStateHandler}s
      */
-    Collection<ImmutableClusterStateHandler<?>> handlers();
+    Collection<ImmutableClusterStateHandler<?>> handlers(Collection<? extends Plugin> loadedPlugins);
+
+    Collection<Class<? extends Plugin>> providedPlugins();
```
grcevski, line 28: I need to add docs here.

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/ILMImmutableStateHandlerProvider.java
```diff
@@ -9,24 +9,30 @@
...
 /**
  * ILM Provider implementation for the {@link ImmutableClusterStateHandlerProvider} service interface
  */
 public class ILMImmutableStateHandlerProvider implements ImmutableClusterStateHandlerProvider {
-    private static final Set<ImmutableClusterStateHandler<?>> handlers = ConcurrentHashMap.newKeySet();
-
     @Override
-    public Collection<ImmutableClusterStateHandler<?>> handlers() {
-        return handlers;
+    public Collection<ImmutableClusterStateHandler<?>> handlers(Collection<? extends Plugin> loadedPlugins) {
```
grcevski, line 23: The provider receives a reference back of its related plugins, which will supply it with what handlers they want to expose.

grcevski:
I think also, maybe I don't need to worry about a collection of plugins here. It should be probably be only one per plugin.

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/ILMImmutableStateHandlerProvider.java
```diff
@@ -10,23 +10,24 @@
...
  * ILM Provider implementation for the {@link ImmutableClusterStateHandlerProvider} service interface
  */
 public class ILMImmutableStateHandlerProvider implements ImmutableClusterStateHandlerProvider {
-    private static final Set<ImmutableClusterStateHandler<?>> handlers = ConcurrentHashMap.newKeySet();
+    private final IndexLifecycle plugin;
 
-    @Override
-    public Collection<ImmutableClusterStateHandler<?>> handlers() {
-        return handlers;
+    public ILMImmutableStateHandlerProvider() {
```
grcevski, line 21: Declaring this constructor to make sure module-info compiles properly, since SPI requires that we use a no-arg constructor. If we switch to using the META-INF/services file, this problem goes away, however our build 'validateModule' task fails, requiring that we also expose this service provider in our module provides section.

server/src/main/java/org/elasticsearch/action/ActionModule.java
```diff
@@ -888,6 +892,23 @@ public void initRestHandlers(Supplier<DiscoveryNodes> nodesInCluster) {
...
+    public void initImmutableClusterStateHandlers(PluginsService pluginsService) {
+        List<ImmutableClusterStateHandler<?>> handlers = new ArrayList<>();
+
+        List<? extends ImmutableClusterStateHandlerProvider> pluginHandlers = pluginsService.loadServiceProviders(
+            ImmutableClusterStateHandlerProvider.class
+        );
+
+        handlers.add(new ImmutableClusterSettingsAction(clusterSettings));
+        pluginHandlers.forEach(h -> handlers.addAll(h.handlers()));
+        // Initialize the controller when merged
```
rjernst, line 909: The two collections here seem to be locals. Shouldn't they be captured somewhere?

server/src/main/java/org/elasticsearch/node/Node.java
```diff
@@ -1055,6 +1055,8 @@ protected Node(
 
             logger.debug("initializing HTTP handlers ...");
             actionModule.initRestHandlers(() -> clusterService.state().nodesIfRecovered());
+            logger.debug("initializing operator handlers ...");
+            actionModule.initImmutableClusterStateHandlers(pluginsService);
```
rjernst, line 1059: This doesn't seem like the right place in node construction to init these. The action module is here in the line above for other reasons (and can actually be moved up, it only needs the cluster service created). I would init these when the action module is created, as part of the ctor.

server/src/main/java/org/elasticsearch/action/ActionModule.java
```diff
@@ -888,6 +892,23 @@ public void initRestHandlers(Supplier<DiscoveryNodes> nodesInCluster) {
...
 
+    /**
+     * Initializes the immutable cluster state handlers for Elasticsearch and it's modules/plugins
+     *
+     * @param pluginsService needed to load all modules/plugins immutable state handlers through SPI
+     */
+    public void initImmutableClusterStateHandlers(PluginsService pluginsService) {
+        List<ImmutableClusterStateHandler<?>> handlers = new ArrayList<>();
+
+        List<? extends ImmutableClusterStateHandlerProvider> pluginHandlers = pluginsService.loadServiceProviders(
```
rjernst, line 903: I think I would rather see loadServiceProviders called in Node, and pass in the List of providers to ActionModule. That better matches the other calls in Node to eg filterPlugins. In general, we try to avoid passing the plugin service itself.

server/src/test/java/org/elasticsearch/plugins/PluginsServiceTests.java
```diff
@@ -531,6 +531,10 @@ public TestExtension() {}
             public TestExtension(TestPlugin plugin) {
 
             }
+
+            public TestExtension(TestPlugin plugin, String anotherArg) {
```
rjernst, line 535: Is this used somewhere?

rjernst (approved):
I left a couple asks on how this looks with Node/ActionModule. LGTM if you agree on the points I left.

server/src/main/java/org/elasticsearch/action/ActionModule.java
grcevski: Yeah, this is unused, I brought in the code to demonstrate how we'd use these when I had the earlier approach. I'll remove this code altogether. The values are locals because the initialization of the controller is not here yet, as in the other PR: https://github.com/elastic/elasticsearch/pull/88224/files#diff-b56bed42f5b0025886eb243ecb48678ac048bfba0ef047545dbb6504e357edf2R915.

I agree with your comments, so I'll refactor the code in the other PR to apply your suggestions on how to initialize the immutable handlers in the other PR.

server/src/test/java/org/elasticsearch/plugins/PluginsServiceTests.java
grcevski: Yes, I added a constructor to make the test check that we disallow a service provider with more than 2 constructors. Previously this class was testing for existence of more than 1.

grcevski:

PR #88363, opened 2022-07-08
New setting to close idle connections in OIDC back-channel (#87773)

ywangd:

the timeout issue.

NOTE: This is a "safe" backport of #87773. The key difference here is
that the new setting is by default not configured, which means the PR
introduces *zero* behaviour change by default. Users need to actively
configure the new setting (to a value other than -1) to enable the new 
behaviour ("automatically closing idle connections).

Backport: #87773

### DIFF
--- a/x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticatorTests.java
+++ b/x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticatorTests.java
@@ -57,13 +57,14 @@ import org.apache.http.impl.nio.conn.PoolingNHttpClientConnectionManager;
 import org.apache.http.message.BasicHeader;
 import org.apache.http.message.BasicHttpResponse;
 import org.apache.http.message.BasicStatusLine;
-import org.elasticsearch.ElasticsearchSecurityException;
-import org.elasticsearch.action.ActionListener;
-import org.elasticsearch.action.support.PlainActionFuture;
 import org.apache.http.protocol.HTTP;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.elasticsearch.ElasticsearchSecurityException;
+import org.elasticsearch.action.ActionListener;
+import org.elasticsearch.action.support.PlainActionFuture;
+import org.elasticsearch.common.Strings;
 import org.elasticsearch.common.logging.Loggers;
 import org.elasticsearch.common.network.InetAddresses;
 import org.elasticsearch.common.settings.SecureString;
@@ -119,6 +120,7 @@ import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.nullValue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
@@ -1000,7 +1002,6 @@ public class OpenIdConnectAuthenticatorTests extends OpenIdConnectTestCase {
             getBasicRealmSettings().put(getFullSettingKey(REALM_NAME, OpenIdConnectRealmSettings.HTTP_CONNECTION_POOL_TTL), "1s").build(),
             threadContext
         );
-        authenticator = new OpenIdConnectAuthenticator(config, getOpConfig(), getDefaultRpConfig(), new SSLService(env), null);
 
         // In addition, capture logs to show that kept alive (TTL) is honored
         final Logger logger = LogManager.getLogger(PoolingNHttpClientConnectionManager.class);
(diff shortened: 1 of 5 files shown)


ywangd:
A few things for a "safe" backport:
1. Default of the new setting is now -1
2. No behaviour change if either the setting is *not* configured or configured to be something like -1.
3. Removed docs since the behavior is temporary on the patch
4. Changed PR title and description as well as the change log summary.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticatorTests.java
```diff
@@ -941,6 +969,238 @@ public void testHandleUserinfoResponseFailure() throws Exception {
...
+            authenticator = new OpenIdConnectAuthenticator(config, getOpConfig(), getDefaultRpConfig(), new SSLService(env), null);
+            appender.assertAllExpectationsMatched();
+            assertThat(authenticator.getKeepAliveStrategy(), nullValue());
+        } finally {
+            Loggers.removeAppender(logger, appender);
+            appender.stop();
+            Loggers.setLevel(logger, (Level) null);
+            authenticator.close();
+        }
+    }
```
ywangd, line 1091: This is a new test for keeping the existing behaviour, i.e. no keep-alive strategy is configured.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticatorTests.java
```diff
@@ -941,6 +969,238 @@ public void testHandleUserinfoResponseFailure() throws Exception {
...
+
+    public void testKeepAliveStrategyNotConfigured() throws IllegalAccessException, URISyntaxException {
+        final Settings.Builder settingsBuilder = getBasicRealmSettings();
+        if (randomBoolean()) {
+            // Only negative time value allowed is -1, but it can take many forms, e.g. -1, -01, -1s, -001nanos etc.
+            settingsBuilder.put(
+                getFullSettingKey(REALM_NAME, OpenIdConnectRealmSettings.HTTP_CONNECTION_POOL_TTL),
+                "-"
+                    + Strings.collectionToDelimitedString(randomList(0, 10, () -> "0"), "")
```
tvernum, line 1062: Nit: 
```suggestion
                    + "0".repeat(randomIntBetween(0,10))
```

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticatorTests.java
```diff
@@ -941,6 +969,238 @@ public void testHandleUserinfoResponseFailure() throws Exception {
...
+        appender.start();
+        Loggers.addAppender(logger, appender);
+        Loggers.setLevel(logger, Level.DEBUG);
+        try {
+            appender.addExpectation(
+                new MockLogAppender.UnseenEventExpectation(
+                    "configure keep-alive strategy",
+                    logger.getName(),
+                    Level.DEBUG,
+                    "configuring keep-alive strategy for http client used by oidc back-channel"
```
tvernum, line 1079: My concern here is that testing `Unseen` is high risk. A tiny change in the message in the authenticator will mean this test no longer achieves anything.

The best suggestion I have is to extract this test into a method that takes a `boolean shouldSeeLogMessage` parameter. And then call it with `true` in the first test, and `false` here. Then we know that changes to the log message will cause one test to fail.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticatorTests.java
ywangd: Neat. Didn't know this method.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/oidc/OpenIdConnectAuthenticatorTests.java
ywangd: The PR effectively does this already. The postive assertion for the logging message is in another test method https://github.com/elastic/elasticsearch/pull/88363/files#diff-85fbc5dd474fc40cca2ada8babe33ea07b5e0db7a432d2424534f74235a5d7ffR1184

Alternatively, I can also add another log message in main code when keep-alive strategy is null. So testing this new message will be a postive assertion. What do you think?

ywangd: Another two things to consider:
1. The test also asserts getKeepaliveStrategy returns null. So it's not completely useless if the logging message is changed.
2. This is only for the patch release. So it probably won't see much change in this area if at all.

tvernum: I would be happy with dropping the logging based check entirely. I don't think it's really needed when we have the `null` check. But I think in its current state it's fragile, so if we want to keep it we should make it reliable.

ywangd: I dropped logging check as suggested. Thanks!

tvernum (approved):
LGTM

ywangd:

PR #88365, opened 2022-07-08
Drop username from AuthenticateRequest

ywangd:
Since transport client is no longer supported for 8.x, the username field
in AuthenticateRequest is not useful at all. At REST layer, the API
never requires passing the username. Authentication should always be
performed for the current authenticating/effective subject. This logic
does not need to depend on the username. This PR drops the username
field and makes the Request class a singleton.

Relates: #88335


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/AuthenticateRequest.java
```diff
@@ -6,50 +6,38 @@
...
-        return new String[] { username };
-    }
-
     @Override
     public void writeTo(StreamOutput out) throws IOException {
         super.writeTo(out);
-        out.writeString(username);
+        if (out.getVersion().before(Version.V_8_4_0)) {
+            throw new IllegalStateException("cannot send authenticate request to a node of earlier version");
+        }
```
ywangd, line 41: Throwing exception here since it is better to fail explicitly than guessing. But in practice this will never happen because AuthenticateRequest does not get sent across nodes.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/AuthenticateRequest.java
```diff
@@ -6,50 +6,38 @@
...
+public class AuthenticateRequest extends ActionRequest {
 
-    private String username;
+    public static final AuthenticateRequest INSTANCE = new AuthenticateRequest();
 
     public AuthenticateRequest(StreamInput in) throws IOException {
         super(in);
-        username = in.readString();
+        if (in.getVersion().before(Version.V_8_4_0)) {
+            in.readString();
```
tvernum, line 24:
```suggestion
            // Older versions included the username as a field
            in.readString();
```

It can be helpful to include an explanation for these seemingly meaningless reads.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/AuthenticateRequest.java
```diff
@@ -6,50 +6,38 @@
...
-public class AuthenticateRequest extends ActionRequest implements UserRequest {
+public class AuthenticateRequest extends ActionRequest {
 
-    private String username;
+    public static final AuthenticateRequest INSTANCE = new AuthenticateRequest();
 
     public AuthenticateRequest(StreamInput in) throws IOException {
         super(in);
-        username = in.readString();
+        if (in.getVersion().before(Version.V_8_4_0)) {
```
tvernum, line 23: Sorry, I sat on this too long and you'll need to change the constants.

tvernum (approved):
LGTM

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/user/AuthenticateRequest.java
ywangd: No worries. I updated them.

PR #88369, opened 2022-07-08
Use consistent map type in IndicesService

original-brownbear:
Iteration and lookup from this map get quite hot at times for large data nodes.
We shouldn't have different map types here depending on whether or not the latest operation
was an index delete or create on the node.
Also, the wrapped `HashMap` is slower to iterate than the immutable collections map.

relates #77466


idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

PR #88378, opened 2022-07-08
[ML] indicate overall deployment failure if all node routes are failed

benwtrent:
If all node routes are failed, we should indicate that the whole deployment is failed through its assignment state. 

The failures could be due to multiple reasons, so for detailed information, the individual node routing reasons should be investigated.


davidkyle (approved):
LGTM

benwtrent:
@elasticmachine update branch

PR #88383, opened 2022-07-08
Extract utility for asserting current thread pool

DaveCTurner:
It's useful to assert we're running on a particular thread pool. Today
we do this by checking the current thread's name, but the actual
implementation of this check varies quite a bit across the codebase.
Some of them are quite verbose, and others fail to include the
delimeters around the thread name. This commit introduces a utility
method to standardize this kind of assertion.


pugnascotia (approved):
LGTM

Issue #88296, opened 2022-07-05
Clarify that lookup runtime sub-fields can't be queried nor aggregated on


javanna:
The lookup runtime field enriches the top hits with fields retrieved from a secondary index. This is purely a fetch phase enrichment at the moment, and users may get surprised that the sub-fields that are fetched from the secondary index are not searchable nor aggregatable. Effectively those are treated like unmapped fields. Would it make sense to throw an error when they get queried or aggregated on?

PR #88389, opened 2022-07-08
Add tests for query/agg on lookup runtime fields

dnhatn:
Adds tests to ensure that querying and aggregating on lookup runtimes aren't supported.

Relates #88296


javanna (approved):
Do we need both yaml tests as well as IT tests? I was thinking either of the two is enough? A unit test that calls the relevant MappedFieldType methods would also be enough I think. Up to you though.

dnhatn:
@javanna Thanks for review. I removed the IT tests and will keep only YAML tests.

PR #88390, opened 2022-07-08
Health API - Monitoring local disk health

gmarouli:
This PR introduces the local health monitoring functionality needed for #84811 . The monitor uses the `NodeService` to get the disk usage stats and determines the node's disk health.

When a change in the disk's is detected or when the health node changes, this class would be responsible to send the node's health to the health node. Currently this is simulated with a method that just logs the current health.

The monitor keeps the last reported health, this way, if something fails on the next check it will try to resend the new health state.


gmarouli:
@elasticmachine update branch

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,251 @@
...
+        if (currentHealth.equals(previousHealth) == false) {
+            sendHealth(currentHealth);
+        }
+        scheduleNextRunIfEnabled(monitorInterval);
+    }
+
+    // This method is synchronized to ensure that we keep track of the last reported health state to the health node.
+    // Note: this is for demonstration purposes, when we will be actually sending the information to the health node we might
+    // choose a different way to keep it thread safe.
+    private synchronized void sendHealth(IndividualNodeHealth currentHealth) {
```
andreidan, line 143: No synchronisation is needed as the cluster applier thread is .. just one thread (cluster changed is called sequentially) 
I see visibility is ensured by the volatile variables - could you please document the reason why each volatile variable is needed to be volatile? ie. I think for some it's "safe publication of objects written to them" 

Alternatively, should we leave the "send of health" part to another PR to keep this one limited in scope?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,251 @@
...
+        if (event.state().nodesIfRecovered().getMinNodeVersion().onOrAfter(Version.V_8_4_0)) {
+            // Wait until the health metadata is available in the cluster state
+            if (healthMetadataInitialized == false) {
+                healthMetadataInitialized = HealthMetadata.getFromClusterState(event.state()) != null;
+                if (healthMetadataInitialized) {
+                    scheduleNextRunIfEnabled(TimeValue.timeValueMillis(1));
+                }
+            } else if (newHealthNodeSelected(event)) {
+                sendHealth(lastReportedHealth);
+            }
```
andreidan, line 111: should this be in a separate PR that handles "sending of health data" ?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,251 @@
...
+
+    @Override
+    public void clusterChanged(ClusterChangedEvent event) {
+        // Wait until every node in the cluster is upgraded to 8.4.0 or later
+        if (event.state().nodesIfRecovered().getMinNodeVersion().onOrAfter(Version.V_8_4_0)) {
+            // Wait until the health metadata is available in the cluster state
+            if (healthMetadataInitialized == false) {
+                healthMetadataInitialized = HealthMetadata.getFromClusterState(event.state()) != null;
+                if (healthMetadataInitialized) {
+                    scheduleNextRunIfEnabled(TimeValue.timeValueMillis(1));
```
andreidan, line 107: I think this should schedule based the monitoring based on the normal interval

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,251 @@
...
+            }
+
+            FsInfo.Path leastAvailablePath = null;
+            for (FsInfo.Path info : nodeStats.getFs()) {
+                if (leastAvailablePath == null) {
+                    leastAvailablePath = info;
+                } else if (leastAvailablePath.getAvailable().getBytes() > info.getAvailable().getBytes()) {
+                    leastAvailablePath = info;
+                }
+            }
```
andreidan, line 228: would we be interested in the path ES is configured to store data? ie. `path.data` ?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,251 @@
...
+/**
+ * This class monitors the health of the node regarding the load on several resources.
+ * Currently, it only checks for available disk space. Furthermore, it informs the health
+ * node about the local health upon change or when a new node is detected.
+ */
+public class LocalHealthMonitor implements ClusterStateListener {
+
+    private static final Logger logger = LogManager.getLogger(LocalHealthMonitor.class);
+
+    public static final Setting<TimeValue> INTERVAL_SETTING = Setting.timeSetting(
```
andreidan, line 46: Should we document this new setting?

andreidan (commented):
Thanks for working on this Mary

I left a first set of comments (and a suggestion to break this PR up)

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: I have a task in https://github.com/elastic/elasticsearch/issues/84811 to document all settings introduced as part of this and to add the troubleshooting page. I prefer to do all documentation at the end to avoid going live before we remove the feature flag.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: Hm.... But that wouldn't work for non-data nodes, right?

But even for data nodes, the `DiskThresholdMonitor` also checks for the least available path and I think it would be nice to be consistent. 

What do you think?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: hm, you mean that if it gets rescheduled it should have the initial delay? 

I have to admit that when I see something executed at an interval I do not expect to have an initial delay. I expect that it's going to be executed asap and then wait for the interval before the next execution. For example let's say that the interval is 10 minutes. I would think it's wrong to wait for 10 minutes before running it for the first time.

What do you think?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: I can put it in the next PR that I am working on, thinking about it kind of helped me set up the code but I can remove this and just log a line upon a change of the node's health. 

I will remove it.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: I will document the volatiles and leave this to the next PR, I still think there is a race but I can elaborate more at the next PR.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,251 @@
...
+            scheduled.cancel();
+            if (enabled) {
+                scheduleNextRunIfEnabled(new TimeValue(1));
+            }
+        }
+    }
+
+    private void scheduleNextRunIfEnabled(TimeValue time) {
+        if (enabled && threadPool.scheduler().isShutdown() == false) {
+            scheduled = threadPool.schedule(this::monitorHealth, time, ThreadPool.Names.MANAGEMENT);
```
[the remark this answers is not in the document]
gmarouli: `scheduleUnlessShuttingDown` is not returning the `Scheduler.ScheduledCancellable` so we cannot cancel a scheduled run. However, I do think that id I rework the code a bit I can use it a make something even nicer than what we already have. So, I will give it a try.

gmarouli:
Update:
- The sending is removed and left for the next PR.
- We use `scheduleUnlessShuttingDown` and used an `AtomicBoolean` to ensure the single execution of the monitoring. I like this approach more. I think this reads better and it doesn't do extra cancelling and scheduling. If there are multiple runs scheduled at the same time only one will run, the other ones will just do nothing. What do you think?

gmarouli:
Investigating the test failure. I can reproduce the failure locally too with this seed:
```


./gradlew ':server:test' --tests "org.elasticsearch.health.node.LocalHealthMonitorTests.testEnablingAndDisabling" -Dtests.seed=C307AA9DA1F7BA82 -Dtests.locale=es-CO -Dtests.timezone=Etc/Greenwich -Druntime.java=17 |  
-- | --



```

gmarouli:
Test fixed. The issue was that we removed the `scheduled` field that was used both to cancel a run but also to signal not allow setting changes to schedule runs. Fixed by using `healthMetadataInitialized` for the same purpose.

gmarouli:
@elasticmachine update branch

gmarouli:

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,233 @@
...
+    private void scheduleNowIfEnabled() {
+        scheduleNextRunIfEnabled(new TimeValue(1));
+    }
+
+    @Override
+    public void clusterChanged(ClusterChangedEvent event) {
+        // Wait until every node in the cluster is upgraded to 8.4.0 or later
+        if (event.state().nodesIfRecovered().getMinNodeVersion().onOrAfter(Version.V_8_4_0)) {
+            // Wait until the health metadata is available in the cluster state
+            if (healthMetadataInitialized == false) {
```
andreidan, line 106: This flag is a little misleading as it waits for a bit more than just the metadata to be initialised (ie. nodes versions). Could we rename it to reflect that?

Also, if the service is enabled,disabled, and enabled again we might end up with 2 scheduled monitoring tracks? (since `setEnabled` doesn't check if we're already running?)

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,233 @@
...
+            }
+
+            FsInfo.Path leastAvailablePath = null;
+            for (FsInfo.Path info : nodeStats.getFs()) {
+                if (leastAvailablePath == null) {
+                    leastAvailablePath = info;
+                } else if (leastAvailablePath.getAvailable().getBytes() > info.getAvailable().getBytes()) {
+                    leastAvailablePath = info;
+                }
+            }
```
andreidan, line 210: Could this be a standalone method that encapsulates this logic? (and we can unit test)

andreidan (commented):
Thanks for iterating on this Mary

Left a few more suggestions

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,233 @@
...
+
+    private void scheduleNextRunIfEnabled(TimeValue time) {
+        if (enabled && healthMetadataInitialized) {
+            threadPool.scheduleUnlessShuttingDown(time, ThreadPool.Names.MANAGEMENT, this::monitorHealth);
+        }
+    }
+
+    // Helper method that starts the monitoring without a delay.
+    private void scheduleNowIfEnabled() {
+        scheduleNextRunIfEnabled(new TimeValue(1));
```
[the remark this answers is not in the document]
gmarouli: Yes, good point!

server/src/main/java/org/elasticsearch/health/node/IndividualNodeHealth.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.elasticsearch.health.node;
+
+import org.elasticsearch.health.HealthStatus;
+
+/**
+ * The health of an individual node which consists by the health status of different resources, currently only disk space.
+ */
+record IndividualNodeHealth(Disk disk) {
```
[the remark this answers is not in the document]
gmarouli: Yes, for sure, we can always figure out a better set-up when we need to add another source.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: About the scheduling two monitoring tasks. That's true, but then one will see that the `inProgress` flag is not false and it will just exit without doing anything or scheduling anything. I thought it was good trade-off compared to locking. What do you think?

server/src/main/java/org/elasticsearch/health/node/selection/HealthNode.java
```diff
@@ -43,4 +44,13 @@ public static PersistentTasksCustomMetadata.PersistentTask<?> findTask(ClusterSt
         PersistentTasksCustomMetadata taskMetadata = clusterState.getMetadata().custom(PersistentTasksCustomMetadata.TYPE);
         return taskMetadata == null ? null : taskMetadata.getTask(TASK_NAME);
     }
+
+    @Nullable
+    public static DiscoveryNode findHealthNode(ClusterState clusterState) {
```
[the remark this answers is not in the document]
gmarouli: Correct I will clean it up

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,233 @@
...
+    // Visible for testing
+    void monitorHealth() {
+        if (inProgress.compareAndSet(false, true)) {
+            ClusterState clusterState = clusterService.state();
+            HealthMetadata healthMetadata = HealthMetadata.getFromClusterState(clusterState);
+            assert healthMetadata != null : "health metadata should have been initialized.";
+            IndividualNodeHealth previousHealth = this.lastReportedHealth;
+            IndividualNodeHealth currentHealth = new IndividualNodeHealth(diskCheck.getHealth(healthMetadata, clusterState));
+            if (currentHealth.equals(previousHealth) == false) {
+                logger.info("Sending node health [{}] to health node", currentHealth);
```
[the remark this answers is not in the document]
gmarouli: Done.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    // Helper method that starts the monitoring without a delay.
+    private void scheduleNowIfEnabled() {
+        scheduleNextRunIfEnabled(new TimeValue(1));
+    }
+
+    @Override
+    public void clusterChanged(ClusterChangedEvent event) {
+        // Wait until every node in the cluster is upgraded to 8.4.0 or later
```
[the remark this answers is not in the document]
gmarouli: Ah, I thought it was already covered by the check for the version. I will take a look and if it's not then we should change it in all the other places we have been checking the version. Right?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,233 @@
...
+            }
+        }
+    }
+
+    // Visible for testing
+    void monitorHealth() {
+        if (inProgress.compareAndSet(false, true)) {
+            ClusterState clusterState = clusterService.state();
+            HealthMetadata healthMetadata = HealthMetadata.getFromClusterState(clusterState);
+            assert healthMetadata != null : "health metadata should have been initialized.";
```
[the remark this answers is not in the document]
gmarouli: Before we schedule any monitoring task we check the flag `healthMetadataInitialized` so theoretically this should hold true. The only case I can think of, is if someone removes the health metadata. But I think we do not give that option programmatically. Would you prefer to have an `if` here?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
andreidan: Yes ! That's a good trade-off 🚀  Can you please document it?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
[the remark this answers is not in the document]
andreidan: Ah, right. The `HealthMetadataService` is controlled by a different setting - enabled/disabled via `health.node.enabled` so that made me think things can diverge. 
I prefer the explicit checks, but it's a personal preference so feel free to ignore

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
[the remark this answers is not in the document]
andreidan: Ah righto - `nodesIfRecovered` does this check already TIL

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: on it!

gmarouli: See: https://github.com/elastic/elasticsearch/pull/88996

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: Done, let me know if it's clear.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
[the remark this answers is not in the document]
gmarouli: They are all controlled by the same flag `health.node.enabled`. The reason I have done this, is that if the health node is disabled, there is no point to monitor the disk health locally, or to publish health metadata right?

On a second thought, depending on the use case @jbaiera is working on we might want to decouple this and have the health metadata always be published. They might be useful for the coordinator node too.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: Also renamed `healthMetadataInitialized` :)

andreidan (commented):
Thanks for iterating on this. 

Left a could more comments

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,201 @@
...
+        scheduleNextRunIfEnabled(TimeValue.timeValueMillis(1));
+    }
+
+    @Override
+    public void clusterChanged(ClusterChangedEvent event) {
+        if (prerequisitesFulfilled == false) {
+            prerequisitesFulfilled = event.state().nodesIfRecovered().getMinNodeVersion().onOrAfter(Version.V_8_5_0)
+                && HealthMetadata.getFromClusterState(event.state()) != null;
+            scheduleNowIfEnabled();
+        }
```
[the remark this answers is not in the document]
gmarouli: Hm, I think I need to rename this method. We check the prerequisites in the method `scheduleNowIfEnabled`. This was a choice I made to ensure the prerequisites are always checked before any scheduling. For this reason it felt redundant to check it again here. What do you think about renaming to `maybeScheduleNow()` indicating that it is not a given that it will schedule it?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -0,0 +1,201 @@
...
+    private volatile TimeValue monitorInterval;
+    private volatile boolean enabled;
+    // Signals that all the prerequisites have been fulfilled and the monitoring task can be scheduled.
+    private volatile boolean prerequisitesFulfilled;
+    // Ensures that only one monitoring task will be in progress at any moment in time.
+    // It removes the need to synchronize scheduling since at the event that there are two
+    // monitoring tasks scheduled, one of them will be no-op.
+    private final AtomicBoolean inProgress = new AtomicBoolean();
+    // Keeps the latest health state that was successfully reported.
+    private DiskHealthInfo lastReportedDiskHealthInfo = null;
```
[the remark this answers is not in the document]
gmarouli: Yeap, I probably removed it accidentally at some point.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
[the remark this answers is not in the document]
gmarouli: About `8.5.0` you are right the `healthMetadata` do exists in `8.4.0` but this local monitor will be pushing changes to the `HealthNode` and that endpoint is not implemented in `8.4.0`. That's why I thought that the minimum required version should be `8.5.0`. Does this make sense?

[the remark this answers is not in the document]
andreidan: Ah ++ fair enough

andreidan (approved):
LGTM thanks for iterating on this Mary 🚀

gmarouli:
@elasticmachine update branch

gmarouli:

gmarouli:
@andreidan thanks for the feedback!

PR #88393, opened 2022-07-08
[DOCS] Add minimal security steps back to docs

### LINKED ISSUE #85046: 8.x security setup docs don't describe how to set up kibana
### Description

As part of security on by default we removed a lot of the existing "setting up security" documentation, because it was no longer relevant.

However, that means that we have a gap in our docs for people who want to turn on security on an existing cluster.

In particular the 7.x docs have a page for [setting up passwords and configuring Kibana](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/security-minimal-setup.html#security-create-builtin-users) but there is no obvious replacement in 8.x  
The docs jump straight into setting up TLS, and skip the Kibana authentication process.

lockewritesdocs:
Adds the steps into 8.x for [enabling security and setting passwords](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/security-minimal-setup.html) from the 7.x documentation for users who want to enable security on an existing cluster where security was previously disabled.

Preview links:
* [Secure the Elastic Stack](https://elasticsearch_88393.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/secure-cluster.html)
* [Manually configure security](https://elasticsearch_88393.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/manually-configure-security.html)
* [Set up minimal security](https://elasticsearch_88393.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/security-minimal-setup.html)

Closes #85046


yang-wei:
haha I think you mean @ywangd

lockewritesdocs:
> haha I think you mean @ywangd

Yes! Apologies for the mistype.

x-pack/docs/en/security/authentication/built-in-users.asciidoc
```diff
@@ -77,24 +77,23 @@ password.
...
-the command in an "interactive" mode, which prompts you to enter new passwords
+the command in an interactive mode, which prompts you to enter new passwords
 for the `elastic`, `kibana_system`, `logstash_system`, `beats_system`, `apm_system`,
 and `remote_monitoring_user` users:
 
 [source,shell]
 --------------------------------------------------
-bin/elasticsearch-setup-passwords interactive
+bin/elasticsearch-reset-passwords interactive
 --------------------------------------------------
```
ywangd, line 90: The usage of these commands are not the same. The interactive mode of the `reset` tool only works for a single user and you need specify it. I am thinking that we may want to leave these changes out of this PR because it needs to be largely rewritten for recommending the new `reset` tool.

x-pack/docs/en/security/securing-communications/security-minimal-setup.asciidoc
```diff
@@ -0,0 +1,168 @@
...
+first time. 
+
+If you're running an existing {es} cluster where security is disabled, you can 
+manually enable the {es} {security-features} and then create passwords for 
+built-in users. You can add more users later, but using the built-in users 
+simplifies the process of enabling security for your cluster.
+
+include::../security-manual-configuration.asciidoc[tag=minimal-security-note]
+
+The basic license includes minimal security settings for the {stack}, so
```
ywangd, line 18: Using `minimal` here is a bit confusing: This page's title is about `minimal security`. It is the least amount of security that a user can enable for some protection. But It is definitely *not* what a basic license offers, which is much more than `minimal`.

ywangd (commented):
It might be helpful to emphasize a bit more that security auto-configuration is the recommended approach for new clusters. It may be helpful to add a callout on the top of the `security-manual-configuration` page to state that one should only follow these steps if they have an existing, unsecured cluster. Also the page's title says `Manually configure security`, but on the side navigation menu, it is `Configure security` (missing the "manually" bit, which is important). The previous page is titled `Start the Elastic Stack with security enabled` which does not reflect the important information that it is "automatic". It would be great if we can contrast these two approaches a bit more so that users can more easily pick the right approach for them.

x-pack/docs/en/security/securing-communications/security-minimal-setup.asciidoc
lockewritesdocs: I think that this section about licenses doesn't make sense here anymore. I'm going to remove it.

lockewritesdocs:
@ywangd, I think that I addressed all of your comments. Let me know if you spot anything else.

ywangd (approved):
LGTM

PR #88398, opened 2022-07-08
Track the count of failed invocations since last successful policy snapshot

jbaiera:
When an automated snapshot fails, the last failure for a policy is captured and stored in the cluster state. Similarly, we store the last successful snapshot invocation as well. We do not track how many invocations have passed between a successful snapshot and the most recent failure. These stats would be helpful for reporting on SLM policy health. 

Instead of a fixed delay, snapshot lifecycle policies are scheduled using a cron expression which can produce variable execution times between snapshot attempts. This makes it difficult to select a window of time where continuous snapshot failure becomes indicative of a problem instead of a transient issue. By including the count of failed invocations since last success we can provide health reporting logic that allows for some transient failures while remaining agnostic of variable execution times that cron can produce.

### DIFF
--- a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/slm/SnapshotLifecyclePolicyMetadata.java
+++ b/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/slm/SnapshotLifecyclePolicyMetadata.java
@@ -7,6 +7,7 @@
 
 package org.elasticsearch.xpack.core.slm;
 
+import org.elasticsearch.Version;
 import org.elasticsearch.cluster.SimpleDiffable;
 import org.elasticsearch.common.io.stream.StreamInput;
 import org.elasticsearch.common.io.stream.StreamOutput;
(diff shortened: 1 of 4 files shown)


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/slm/SnapshotLifecyclePolicyMetadata.java
```diff
@@ -51,6 +53,8 @@ public class SnapshotLifecyclePolicyMetadata implements SimpleDiffable<SnapshotL
     private final SnapshotInvocationRecord lastSuccess;
     @Nullable
     private final SnapshotInvocationRecord lastFailure;
+    @Nullable
+    private final Long invocationsSinceLastSuccess;
```
dakrone, line 57: How do you feel about making this non-nullable and instead treating a missing value as 0? I think that would be nicer code-wise since consumers don't have to handle a null value (and 0 invocations would be the "default" empty value)

dakrone (commented):
This generally looks good to me, but I think we should make it non-null and treat the default missing value as 0 invocations, what do you think?

jbaiera:

dakrone (approved):
LGTM

PR #88399, opened 2022-07-08
Improve error when sorting on incompatible types

### LINKED ISSUE #73146: ClassCastException when sorting on incompatible fields
**Elasticsearch version** (`bin/elasticsearch --version`): 7.12.1

Elasticsearch returns a ClassCastException when trying to sort on incompatible fields across indexes. Discussing this with @romseygeek, we should be returning a specific error instead.

**Steps to reproduce**:
```
PUT index_1
{
  "mappings": {
    "properties": {
      "date": {
        "type": "keyword"
      }
    }
  }
}

PUT index_2
{
  "mappings": {
    "properties": {
      "date": {
        "type": "date"
      }
    }
  }
}

POST index_1/_doc
{ "date": "2021-05-17" }

POST index_2/_doc
{ "date": "2021-05-17" }

GET index_1,index_2/_search
{ "sort": "date" }
```
Last query returns:
```json
{
  "error" : {
    "root_cause" : [ ],
    "type" : "search_phase_execution_exception",
    "reason" : "",
    "phase" : "fetch",
    "grouped" : true,
    "failed_shards" : [ ],
    "caused_by" : {
      "type" : "class_cast_exception",
      "reason" : "class java.lang.Long cannot be cast to class org.apache.lucene.util.BytesRef (java.lang.Long is in module java.base of loader 'bootstrap'; org.apache.lucene.util.BytesRef is in unnamed module of loader 'app')"
    }
  },
  "status" : 500
}
```

mayya-sharipova:
Currently when sorting on incompatible types, we get
class_cast_exception error (500). This patch improves
the error to explain that the problem is because
of incompatible sort types for the field across
different shards and returns user error (400).

Closes #73146

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/search/SearchPhaseController.java
+++ b/server/src/main/java/org/elasticsearch/action/search/SearchPhaseController.java
@@ -14,6 +14,8 @@ import org.apache.lucene.search.FieldDoc;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.Sort;
 import org.apache.lucene.search.SortField;
+import org.apache.lucene.search.SortedNumericSortField;
+import org.apache.lucene.search.SortedSetSortField;
 import org.apache.lucene.search.TermStatistics;
 import org.apache.lucene.search.TopDocs;
 import org.apache.lucene.search.TopFieldDocs;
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/elasticsearch/action/search/SearchPhaseController.java
```diff
@@ -207,6 +210,56 @@ static TopDocs mergeTopDocs(Collection<TopDocs> results, int topN, int from) {
         return mergedTopDocs;
     }
 
+    private static void assertSameSortTypes(Collection<TopDocs> results, SortField[] firstSortFields) {
```
romseygeek, line 213: Can we call this 'check' rather than 'assert'? Assert implies that it would be disabled if assertions are turned off.

romseygeek (approved):
One nit, but LGTM otherwise

mayya-sharipova:

PR #88403, opened 2022-07-09
Make DiffableUtils.diff implementation agnostic

rjernst:
DiffableUtils has methods for reading and creating diffs of maps. The
two implementation types, a JDK Map or ImmutableOpenMap, have different
methods and implementations. However, the format of the diff is always
the same, so the diff creation can be shared. This commit creates an
internal MapBuilder abstraction to allow both Map and ImmutableOpenMap
to produce diffs from the same MapDiff implementation.

relates #86239


original-brownbear (approved):
LGTM, nice cleanup!

PR #88404, opened 2022-07-09
Simplify azure discovery installation docs

rjernst:
The installation docs for azure discovery suggest installing
Elasticsearch, and then verifying it is running. However, installation
packages do not generally automatically start an installed service. For
example, with systemd, once a package is installed it must be enabled
and/or started explicitly. This commit removes wording about verifying
Elasticsearch is running, as well as adjusts later docs about restarting
it. Additionally, the command to start Elasticsearch is fixed to use
systemctl, as SystemV init is no longer supported by our packages.


debadair (approved):
LGTM

PR #88410, opened 2022-07-11
Fix RealmIdentifier XContent parser

ywangd:
RealmIdentifier's constructor has type before name. But the parser had
it reserved. This PR fixes the parser to have the same argument order as
the constructor. It also adds relevant tests.

Since the parser is only used internally for the User Profile feature
which is not official released yet. This fix is a non-issue.


n1v0lg (approved):
Thanks for the fix!

PR #88412, opened 2022-07-11
New setting to close idle connections in OIDC back-channel (#87773)

ywangd:

the timeout issue.

NOTE: This is a "safe" backport of #87773. The key difference here is
that the new setting is by default not configured, which means the PR
introduces *zero* behaviour change by default. Users need to actively
configure the new setting to enable the new behaviour ("automatically
closing idle connections).

Backport: #87773

### DIFF
--- a/docs/changelog/87773.yaml
+++ b/docs/changelog/87773.yaml
@@ -1,0 +1,5 @@
+pr: 87773
+summary: New setting to close idle connections in OIDC back-channel
+area: Security
+type: enhancement
+issues: []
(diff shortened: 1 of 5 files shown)


ywangd:
This is almost identical to the safe backport to 8.3.3 (https://github.com/elastic/elasticsearch/pull/88363). The only difference is to replace some Java APIs that are not available in 8.

tvernum (approved):
LGTM

ywangd:
> I think this means we'll also need to allow-list this setting on ESS. Can you cover that?

Yes. This is on my todo list. Support folks are also asking about it in the relevant SDH.

PR #88414, opened 2022-07-11
Remove usages of BucketCollector#getLeafCollector(LeafReaderContext)

iverase:
The method BucketCollector#getLeafCollector(LeafReaderContext) should be removed in favour of BucketCollector#getLeafCollector(AggregationExecutionContext). This PR prepares that action by removing the usages of the first method with the second.


server/src/main/java/org/elasticsearch/search/aggregations/bucket/BestBucketsDeferringCollector.java
```diff
@@ -119,8 +118,8 @@ public LeafBucketCollector getLeafCollector(AggregationExecutionContext aggCtx)
 
             @Override
             public void collect(int doc, long bucket) throws IOException {
-                if (context == null) {
-                    context = aggCtx.getLeafReaderContext();
+                if (BestBucketsDeferringCollector.this.aggCtx == null) {
```
not-napoleon, line 121: Thanks for clarifying this on slack.  I think renaming would be clearer here.

not-napoleon (approved):
LGTM, thanks for cleaning this up!

PR #88416, opened 2022-07-11
TSDB: RollupShardIndexer improve logger

weizijun:
1. add execute took
2. add trace check to avoid high cost method


csoulios:
@elasticmachine ok to test

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
```diff
@@ -180,7 +182,13 @@ public void afterBulk(long executionId, BulkRequest request, BulkResponse respon
                             )
                         );
                     numFailed.addAndGet(failures.size());
-                    logger.error("Shard [{}] failed to populate rollup index. Failures: [{}]", indexShard.shardId(), failures);
+                    logger.info(
```
csoulios, line 185: If indexing rollup documents fail, we should fail the all the task. I think this is an error case and should be logged as error.

csoulios (commented):
Thanks for submitting those improvements. 

I have added a couple minor comments.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
weizijun: Do you means that, if there are failed items, the total rollup task will be cancelled?

csoulios: Yes, otherwise we will end up with a wrong rollup index that does not contain all rollup documents. If we fail the whole task, ILM will handle the failure and retry

weizijun: got it, I will revert this change, and add the cancelled code when bulk failed

csoulios (approved):
LGTM!

PR #88418, opened 2022-07-11
Reduce map lookups

idegtiarenko:
This change replcase 2 hash map operations with a single one and a null
check.


DaveCTurner (approved):
LGTM

PR #88420, opened 2022-07-11
TSDB: downsampling  cache histoTimestamp to reduce rounding cost

weizijun:
Cache histoTimestamp to reduce rounding cost.
The current tests will cover all the modified case.


csoulios:
@elasticmachine ok to test

csoulios (approved):
LGTM!

PR #88421, opened 2022-07-11
Fix merge

idegtiarenko:


server/src/main/java/org/elasticsearch/cluster/routing/RoutingNodes.java
```diff
@@ -275,7 +275,7 @@ public RoutingNode node(String nodeId) {
 
     public Set<String> getAttributeValues(String attributeName) {
         // Only ever accessed on the master service thread so no need for synchronization
-        assert MasterService.assertMasterUpdateOrTestThread();
+        // assert MasterService.assertMasterUpdateOrTestThread();
```
idegtiarenko, line 278: This is now called by `AwarenessAllocationDecider` from a continuous computation that is reusing generic thread pool. 
It also looks like all data used here is immutable and final so I am not sure above comment is relevant and assertion is needed (unless I miss something)

server/src/main/java/org/elasticsearch/common/util/Maps.java
```diff
@@ -298,4 +298,14 @@ static int capacity(int expectedSize) {
...
 
+    /**
+     * This method creates a "deep" of the {@code source} map using on {@code copyValueFunction} each value
+     */
+    public static <K, V> Map<K, V> copy(Map<K, V> source, Function<V, V> copyValueFunction) {
+        var copy = Maps.<K, V>newHashMapWithExpectedSize(source.size());
+        for (var entry : source.entrySet()) {
+            copy.put(entry.getKey(), copyValueFunction.apply(entry.getValue()));
+        }
+        return copy;
```
idegtiarenko, line 309: I wonder if it is worth implementing this change on master and then sync and fix this branch with it?

server/src/main/java/org/elasticsearch/common/util/Maps.java
```diff
@@ -298,4 +298,14 @@ static int capacity(int expectedSize) {
         throw new IllegalStateException("Unsupported input format");
     }
 
+    /**
+     * This method creates a "deep" of the {@code source} map using on {@code copyValueFunction} each value
```
DaveCTurner, line 302:
```suggestion
     * This method creates a "deep copy" of the {@code source} map using {@code copyValueFunction} on each value
```

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNodes.java
```diff
@@ -152,27 +153,14 @@ private RoutingNodes(RoutingNodes routingNodes) {
...
         this.relocatingShards = routingNodes.relocatingShards;
-        this.attributeValuesByAttribute = Maps.newMapWithExpectedSize(routingNodes.attributeValuesByAttribute.size());
-        for (Map.Entry<String, Set<String>> entry : routingNodes.attributeValuesByAttribute.entrySet()) {
-            this.attributeValuesByAttribute.put(entry.getKey(), new HashSet<>(entry.getValue()));
-        }
-        this.recoveriesPerNode = Maps.newMapWithExpectedSize(routingNodes.recoveriesPerNode.size());
-        for (Map.Entry<String, Recoveries> entry : routingNodes.recoveriesPerNode.entrySet()) {
-            this.recoveriesPerNode.put(entry.getKey(), entry.getValue().copy());
-        }
+        this.attributeValuesByAttribute = new ConcurrentHashMap<>(Maps.copy(routingNodes.attributeValuesByAttribute, HashSet::new));
```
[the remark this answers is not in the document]
idegtiarenko: Surprisingly we do not yet have anything that creates a CHM with expected size.
I wonder if it is okay to use `java.util.Collections#synchronizedMap` here? I will take a look into CHM with expected size in a bit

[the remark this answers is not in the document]
DaveCTurner: I don't think there's much contention on this map so yeah probably that's fine.

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNodes.java
```diff
@@ -74,7 +75,7 @@ public class RoutingNodes extends AbstractCollection<RoutingNode> {
 
     private int relocatingShards = 0;
 
-    private final Map<String, Set<String>> attributeValuesByAttribute;
+    private final ConcurrentMap<String, Set<String>> attributeValuesByAttribute;
```
DaveCTurner, line 78: This can remain as a `Map` right?

```suggestion
    private final Map<String, Set<String>> attributeValuesByAttribute;
```

DaveCTurner (approved):
LGTM with one question

PR #88432, opened 2022-07-11
Simplify map copying

idegtiarenko:
This commit introduces a method that simplifies creating a map deep copy

### DIFF
--- a/server/src/main/java/org/elasticsearch/common/util/Maps.java
+++ b/server/src/main/java/org/elasticsearch/common/util/Maps.java
@@ -298,4 +298,14 @@ public class Maps {
         throw new IllegalStateException("Unsupported input format");
     }
 
+    /**
+     * This method creates a copy of the {@code source} map using {@code copyValueFunction} to create a defensive copy of each value.
+     */
+    public static <K, V> Map<K, V> copyOf(Map<K, V> source, Function<V, V> copyValueFunction) {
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/elasticsearch/common/util/Maps.java
```diff
@@ -298,4 +298,14 @@ static int capacity(int expectedSize) {
         throw new IllegalStateException("Unsupported input format");
     }
 
+    /**
+     * This method creates a copy of the {@code source} map using {@code copyValueFunction} to create a defensive copy of each value.
+     */
+    public static <K, V> Map<K, V> copyOf(Map<K, V> source, Function<V, V> copyValueFunction) {
```
original-brownbear, line 304: Maybe better call if `deepCopy` to make it clearer from the name what this is for?

original-brownbear (approved):
LGTM but maybe rename including a "deep" in the name of the new method.

server/src/main/java/org/elasticsearch/common/util/Maps.java
original-brownbear: That was kinda my point actually. I'd rather not have the same name as the JDK method here since this does something quite different :) But either way, this was more of a NIT, go with it if you disagree, it was more of a NIT :)

PR #88437, opened 2022-07-11
Don't index geo_shape field in AbstractBuilderTestCase

iverase:
The geo_shape field mapper is currently register in the spatial x-pack module. For historical reason we are adding a geo_shape field on the AbstractBuilderTestCase by default which means all the implementation needs to support geo_shape fields. In order to achieve that we are registering a (sort of) fake mapper. This PR stops adding the geo_shape field mapper by default and adds the mapper only when it is needed.


javanna (approved):
LGTM, I was wondering what the plan is with the test plugin. I see that it's deprecated, and its javadocs refers to the need to migrate tests to spatial-extras. Is that still the case?

iverase:
@javanna The idea is to remove it completely, this is the first baby step in that direction.

PR #88440, opened 2022-07-11
Remove usages of TestGeoShapeFieldMapperPlugin from enrich module

iverase:
We should be using the spatial module where the real mapper is being registered.


jbaiera (approved):
LGTM

PR #88443, opened 2022-07-11
Provide tracing implementation using OpenTelemetry +  APM agent

pugnascotia:
Part of #84369. Split out from #87696. Implement the `Tracer` interface by providing an X-Pack module that uses OpenTelemetry, along with Elastic's APM agent for Java.

See the file `TRACING.md` for background on the changes and the reasoning for some of the implementation decisions.

The configuration mechanism is the most fiddly part of this PR. The Security Manager permissions required by the APM Java agent make it prohibitive to start an agent from within Elasticsearch programmatically, so it must be configured when the ES JVM starts. That means that the startup CLI needs to assemble the required JVM options.

To complicate matters further, the APM agent needs a secret token in order to ship traces to the APM server. We can't use Java system properties to configure this, since otherwise the secret will be readable to all code in Elasticsearch. It therefore has to be configured in a dedicated config file. This in itself is awkward, since we don't want to leave secrets in config files. Therefore, we pull the APM secret token from the keystore, write it to a config file, then delete the config file after ES starts.

There's a further issue with the config file. Any options we set in the APM agent config file cannot later be reconfigured via system properties, so we need to make sure that only "static" configuration goes into the config file.

I generated most of the files under `qa/apm` using an APM test utility (I can't remember which one now, unfortunately). The goal is to setup up a complete system so that traces can be captured in APM server, and the results in Elasticsearch inspected.


x-pack/plugin/apm-integration/src/main/java/org/elasticsearch/xpack/apm/APMTracer.java
```diff
@@ -0,0 +1,388 @@
...
+
+    void setExcludeNames(List<String> excludeNames) {
+        this.excludeNames = excludeNames;
+        this.filterAutomaton = buildAutomaton(includeNames, excludeNames);
+    }
+
+    @Override
+    protected void doStart() {
+        if (enabled) {
+            this.services = createApmServices();
```
grcevski, line 114: I think there might be a possible race condition here. If one thread was doing `setEnabled(false)` and we had already read `enabled` here we might leave APMServices as an instance with enabled false.

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/APMJvmOptions.java
```diff
@@ -0,0 +1,269 @@
...
+     * @throws IOException if a problem occurs reading the filesystem
+     */
+    private static Optional<Path> findAgentJar() throws IOException {
+        final Path apmModule = Path.of("modules/x-pack-apm-integration");
+
+        if (Files.isDirectory(apmModule) == false) {
+            return Optional.empty();
+        }
+
+        try (var apmStream = Files.list(apmModule)) {
```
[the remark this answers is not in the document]
pugnascotia: There should never be multiple agent jars unless the installation had been corrupted somehow, but I've added some logic to check.

server/src/main/java/org/elasticsearch/common/util/concurrent/ThreadContext.java
```diff
@@ -116,9 +119,21 @@ public StoredContext stashContext() {
                 }
             }
         }
-        if (hasHeadersToCopy) {
-            Map<String, String> map = headers(context);
-            ThreadContextStruct threadContextStruct = DEFAULT_CONTEXT.putHeaders(map);
+
+        boolean hasTransientHeadersToCopy = context.transientHeaders.containsKey(Task.APM_TRACE_CONTEXT);
+
+        if (hasHeadersToCopy || hasTransientHeadersToCopy) {
```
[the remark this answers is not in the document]
pugnascotia: Oh. You're right. I don't know how I didn't spot that 🙈

server/src/main/java/org/elasticsearch/tasks/TaskManager.java
```diff
@@ -175,43 +175,45 @@ public <Request extends ActionRequest, Response extends ActionResponse> Task reg
...
-                try {
-                    release();
-                } finally {
-                    taskListener.onResponse(response);
-                }
+
+        try (var ignored = threadPool.getThreadContext().newTraceContext()) {
+            final Task task;
+            try {
+                task = register(type, action.actionName, request);
```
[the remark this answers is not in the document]
pugnascotia: Added an assert.

x-pack/plugin/apm-integration/src/main/java/org/elasticsearch/xpack/apm/APMTracer.java
pugnascotia: I see what you're getting at, but in reality `doStart()` will only be called from `start()`, which is called from `Node.start()` in a single thread. There shouldn't be any cluster state modifications going on at this point that would cause `enabled` to change value.

grcevski: OK, cool.

grcevski (approved):
This looks good to me! LGTM!

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ServerCli.java
```diff
@@ -80,25 +79,31 @@ public void execute(Terminal terminal, OptionSet options, Environment env, Proce
...
-            return;
-        }
+            ServerArgs args = createArgs(options, env, keystore, keystorePassword, processInfo);
+            this.server = startServer(terminal, processInfo, args);
 
-        // we are running in the foreground, so wait for the server to exit
-        int exitCode = server.waitFor();
-        if (exitCode != ExitCodes.OK) {
-            throw new UserException(exitCode, "Elasticsearch exited unexpectedly");
+            if (options.has(daemonizeOption)) {
```
rjernst, line 97: The keystore should be closed once we reach here. The detach/wait should happen outside of the try.

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/JvmOptionsParser.java
```diff
@@ -133,12 +131,19 @@ private List<String> jvmOptions(final Path config, Path plugins, final String es
         final List<String> ergonomicJvmOptions = JvmErgonomics.choose(substitutedJvmOptions);
         final List<String> systemJvmOptions = SystemJvmOptions.systemJvmOptions();
 
+        final List<String> apmOptions = APMJvmOptions.apmJvmOptions(
+            args.nodeSettings(),
+            args.keystore(),
+            Path.of(substitutions.get("ES_TMPDIR"))
```
rjernst, line 137: We have the tmpDir outside of the call to jvmOptions. Let's pass it through instead of assuming the env var name?

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/APMJvmOptions.java
```diff
@@ -0,0 +1,279 @@
...
+     * for the APM Java agent and CLI options to the JVM to configure APM. The config file is temporary
+     * because it will be deleted once Elasticsearch starts.
+     *
+     * @param settings the Elasticsearch settings to consider
+     * @param keystore a wrapper to access the keystore, or null if there is no keystore
+     * @param tmpdir Elasticsearch's temporary directory, where the config file will be written
+     */
+    static List<String> apmJvmOptions(Settings settings, @Nullable KeyStoreWrapper keystore, Path tmpdir) throws UserException,
+        IOException {
+        final boolean enabled = settings.getAsBoolean("xpack.apm.enabled", false);
```
rjernst, line 140: Is this a dynamic setting (cluster setting) or node local? I thought from reading other code it was dynamic, so it could be disabled on the fly.

rjernst (commented):
I left some comments, but it's only a partial review, I wasn't able to get through it all. I leave it to other reviewers to give a final ok.

I also have some general comments:
* This PR seems to have both the imlementation using APM as well as additional uses of tracing within ES. For example, I see changes to MasterService adding tracing. Can those uses be split out? They should be reviewed by the appropriate team owning that code, like distributed team in the case of MasterService, and having an isolated PR (which should just depend on the Tracer that already exists in main) would make reviewing for them much easier than looking at this large PR.
* The cli changes are a little clunky. In particular needing to delete the configuration file generated in the cli later during startup. Additionally node settings are pulled without validation. It would be nicer if this could all be done in node startup...
* I am concerned about how this agent will interact with our future planned agent that will help replacing the java security manager. They will have to behave together, and I don't know about passing them both as command line agents. I think if this PR were to dynamically load the agent in process, instead of passing via the command line, it would allow these changes to be more isolated, and thus less invasive. We can then look at the most effective way to make these agents play nicely with each other in the future, without be bogged down by cli clunkiness that must then be refactored anyways.

/cc @ChrisHegarty who probably has more thoughts on the agent options

x-pack/plugin/apm-integration/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+ * 2.0; you may not use this file except in compliance with the Elastic License
+ * 2.0.
+ */
+apply plugin: 'elasticsearch.internal-es-plugin'
+
+esplugin {
+  name 'x-pack-apm-integration'
+  description 'Provides APM integration for Elasticsearch'
+  classname 'org.elasticsearch.xpack.apm.APM'
+  extendedPlugins = ['x-pack-core']
```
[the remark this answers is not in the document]
pugnascotia: I tried moving it to `server` without a separate loader, and the agent code (not the server code) exploded all over the place with security manager exceptions. It's not obvious to be what is the problem, or why using another loader would help, since we don't directly use the agent jar, though I wonder if having it in the server classpath is enough to breaking things.

In the meantime, I've moved the APM code to be a regular module. Maybe this is something we can improve later.

x-pack/plugin/apm-integration/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+}
+
+dependencies {
+  implementation "io.opentelemetry:opentelemetry-api:${versions.opentelemetry}"
+  implementation "io.opentelemetry:opentelemetry-context:${versions.opentelemetry}"
+  implementation "io.opentelemetry:opentelemetry-semconv:${versions.opentelemetry}-alpha"
+  runtimeOnly    "co.elastic.apm:elastic-apm-agent:${versions.apm_agent}"
+
+  // OTel context needs these, I think
```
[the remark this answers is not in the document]
pugnascotia: This is a leftover. I've removed the logging deps.

x-pack/plugin/apm-integration/build.gradle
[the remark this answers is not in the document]
rjernst: A regular module is fine for now. We can iterate later to load it differently.

server/src/main/java/org/elasticsearch/tasks/TaskManager.java
```diff
@@ -175,43 +175,45 @@ public <Request extends ActionRequest, Response extends ActionResponse> Task reg
...
-        action.execute(task, request, new ActionListener<Response>() {
-            @Override
-            public void onResponse(Response response) {
-                try {
-                    release();
-                } finally {
-                    taskListener.onResponse(response);
-                }
+
+        try (var ignored = threadPool.getThreadContext().newTraceContext()) {
```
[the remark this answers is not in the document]
pugnascotia: I'm in the process on splitting out instrumenting changes into separate PRs.

x-pack/plugin/apm-integration/build.gradle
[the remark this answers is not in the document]
rjernst: However, note that I don't think it would be possible to load apm early enough through a dynamic agent if it's in an ES module.

> I tried moving it to server without a separate loader

By loading directly from server I meant in a separate loader, similar to how we load x-content. This would mean the apm code would be under libs, need to be in a separate directory under libs in the distribution, and we would then create the loader on startup and reflectively add the agent.

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/APMJvmOptions.java
```diff
@@ -0,0 +1,292 @@
...
+        "service_node_name",
+        "service_version",
+        "stress_monitoring_interval",
+        "trace_methods_duration_threshold",
+        "use_jaxrs_path_as_transaction_name",
+        "verify_server_cert"
+    );
+
+    static {
+        STATIC_CONFIG = new HashMap<>();
```
original-brownbear, line 87: It's always a little weird to have mutable constants, maybe better use `Map.of(` or `Map.copyOf` for the assignment of the constants here?

modules/apm/src/main/java/org/elasticsearch/tracing/apm/APMTracer.java
```diff
@@ -0,0 +1,394 @@
...
+ * This is an implementation of the {@link org.elasticsearch.tracing.Tracer} interface, which uses
+ * the OpenTelemetry API to capture spans.
+ * <p>
+ * This module doesn't provide an implementation of the OTel API. Normally that would mean that the
+ * API's default, no-op implementation would be used. However, when the APM Java is attached, it
+ * intercepts the {@link GlobalOpenTelemetry} class and provides its own implementation instead.
+ */
+public class APMTracer extends AbstractLifecycleComponent implements org.elasticsearch.tracing.Tracer {
+
+    private static final Logger LOGGER = LogManager.getLogger(APMTracer.class);
```
original-brownbear, line 58: I think we started using lowercase `logger` throughout at some point didn't we?

modules/apm/src/main/java/org/elasticsearch/tracing/apm/APMTracer.java
```diff
@@ -0,0 +1,394 @@
...
+        if (span != null) {
+            span.recordException(throwable);
+        }
+    }
+
+    @Override
+    public void setAttribute(String spanId, String key, boolean value) {
+        final var span = Span.fromContextOrNull(spans.get(spanId));
+        if (span != null) {
+            span.setAttribute(key, value);
```
original-brownbear, line 299: How does the `Span` work? Is this a blocking IO call? Or is it non-blocking, in which case what happens if the other side is slow and we start queuing up work here?

original-brownbear (commented):
I left one important question for now (and two nits :)).

server/src/main/java/org/elasticsearch/node/Node.java
```diff
@@ -1096,14 +1101,44 @@ protected Node(
         }
     }
 
-    private Tracer getTracer(PluginsService pluginsService, ClusterService clusterService, Settings settings) {
+    /**
+     * If the JVM was started with the Elastic APM agent and a config file argument was specified, then
+     * delete the config file. The agent only reads it once, when supplied in this fashion, and it
+     * may contain a secret token.
+     */
+    private void deleteTemporaryApmConfig(JvmInfo jvmInfo) {
```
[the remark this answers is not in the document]
pugnascotia: I was trying to delete the config as early as possible, which is when the JVM arguments are available, which is too soon to load the tracer plugin. The file is actually for the Java agent, not the module. I timed the different between the current delete point and where I create the tracer, and on my laptop it was of the order of 1500ms.

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/APMJvmOptions.java
pugnascotia: I was trying to avoid the formatting getting mangled, but actually all I need to do is use the right enable / disable comments. Now we just have `Map.of()`.

modules/apm/src/main/java/org/elasticsearch/tracing/apm/APMTracer.java
pugnascotia: I didn't think so, but a quite check showed me that lower-case is way more popular, so I've changed it.

modules/apm/src/main/java/org/elasticsearch/tracing/apm/APMTracer.java
pugnascotia: It's implementation-dependent, but it absolutely shouldn't be blocked. The APM Java agent buffers spans and send them on a separate thread. Spans will get dropped if there's slowdown on the APM server. From the [docs FAQ](https://www.elastic.co/guide/en/apm/agent/java/master/faq.html):

> It’s important to note that the latency of an application won’t be affected much by the agent (in the order of single-digit microseconds), even if you sample at 100%. However, the background reporter thread has some work to do for serializing and gzipping events. If your application is not CPU bound, this shouldn’t matter much. Note that if the APM Server can’t handle all the events, the agent will drop data to not crash your application. It will then also not serialize and gzip the events.

We'll ship with a sample rate of 20% by default, possibly lower after we've tried it out with internal clusters.

TRACING.md
```diff
@@ -0,0 +1,156 @@
...
+      https://localhost:9200/_cluster/settings
+
+
+### More details about configuration
+
+For context, the APM agent pulls configuration from [multiple
+sources][agent-config], with a hierarchy that means, for example, that options
+set in the config file cannot be overridden via system properties.
+
+Now, in order to send tracing data to the APM server, ES needs to configured with
```
jpountz, line 48: I think you meant "needs to **be** configured"?

jpountz:
Wonderful! Should we add the `release highlight` label?

pugnascotia:
> Wonderful! Should we add the `release highlight` label?

I'm not honestly sure. We haven't written externally-facing docs yet, as we're focussed on using it for ourselves right now. I'll leave that up to the Core/Infra team.

Issue #31096, opened 2018-06-05
Add mixed-cluster tests for X-Pack


ywelsch:
We need tests that check that mixed-version X-Pack clusters are working correctly, in particular now that they're becoming the default distribution. Having such tests can uncover X-Pack BWC issues such as found in #30859.

gwbrown:
To resolve confusion for anyone who, like me, saw the [rolling upgrade x-pack tests](https://github.com/elastic/elasticsearch/commits/91f4b58bf71249cf81562bd6069c64aafa46591b/x-pack/qa/rolling-upgrade) and thought this might have been addressed already, per @jakelandis (thanks!):
> There is a difference between "mixed cluster" in a rolling upgrade and a "mixed cluster" test. The former (rolling upgrade) only tests the explicit things that are written for the rolling upgrade, and the latter (mixed cluster test) runs all of the YAML REST tests.  We have rolling upgrade tests for both OSS and x-pack, but we only have "mixed cluster" tests that run the REST tests for OSS. We do not run the x-pack YAML REST tests in a mixed cluster as far as I know.

cbuescher:
I'm just stubling about this long-open issue again reading some details of #80608 again. Looking at the [comments](https://github.com/elastic/elasticsearch/pull/80608#discussion_r747708564) and following discussion it might be worth bumping this since we might have a larger blind spot in coverage here.

mark-vieira:
@nik9000 also brought up adding mixed-cluster testing to other modules. Essentially we want a generic way to run any given test suite in a mixed-cluster scenario. While initially investigating that it's actually quite complex, perhaps just adding coverage for the core x-pack tests in a mixed-cluster is a good start.

mosche:
@elastic/es-delivery We discussed this in our team sync. We would need your help for the infrastructure part on this.

mark-vieira:
> [@elastic/es-delivery](https://github.com/orgs/elastic/teams/es-delivery) We discussed this in our team sync. We would need your help for the infrastructure part on this.

Happy to sync on this to discuss what we need. I don't think we need any additional infra here, as we have plenty of mixed cluster tests suites already. It's just a matter of identifying the gaps we want to fill here.

PR #88445, opened 2022-07-11
add issuer to GET _ssl/certificates

jakelandis:
This commit adds the certificate issuer to the values returned by _ssl/certificates.
This can be useful to better understand the certificate chain in use.


jakelandis:
@elasticmachine update branch

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ssl/cert/CertificateInfo.java
```diff
@@ -129,30 +145,23 @@ public String toString() {
     }
 
     @Override
-    public boolean equals(Object other) {
```
jakelandis, line 132: equals and hash code updated using Intellij (nothing custom in here)

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/ssl/cert/CertificateInfoTests.java
```diff
@@ -22,8 +22,11 @@
 
 public class CertificateInfoTests extends ESTestCase {
 
+    private static final String selfSignedCertPath = "/org/elasticsearch/xpack/security/transport/ssl/certs/simple/testnode.crt";
+    private static final String rootSignedCertPath = "/org/elasticsearch/xpack/security/transport/ssl/certs/simple/openldap.crt";
```
gwbrown, line 26: I was gonna comment that this is weird and ask you what's up, but I see you just copied in from below looks like? If you know, what's up with this? Having hardcoded paths in a test like this is kinda weird.

Actually, if we have that explanation, can we put it in a comment?

AthenaEryma (approved):
Looks good, left a couple tiny comments but honestly I'd be fine with this as-is too.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ssl/cert/CertificateInfo.java
```diff
@@ -64,6 +67,11 @@ public CertificateInfo(StreamInput in) throws IOException {
         this.serialNumber = in.readString();
         this.hasPrivateKey = in.readBoolean();
         this.expiry = Instant.ofEpochMilli(in.readLong()).atZone(ZoneOffset.UTC);
+        if (in.getVersion().onOrAfter(Version.V_8_4_0)) {
+            this.issuer = in.readString();
+        } else {
+            this.issuer = "[unknown]";
```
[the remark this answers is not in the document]
ywangd: This object does not get sent across the wire. So this is indeed just nitpicking. That said, I'd suggest setting the issuer to `null` in this case and skip it in `toXContent` method when it is `null` to (1) avoid having an arbitrary default value; (2) closer to the existing behaviour.

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/ssl/cert/CertificateInfoTests.java
jakelandis: Not sure I follow ..

Using absolute paths here is the full path to the test location
```
/home/jakelandis/workspace/elasticsearch/main/x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/ssl/cert/CertificateInfoTests.java
```
and here is the the full path to the resource
```
/home/jakelandis/workspace/elasticsearch/main/x-pack/plugin/core/src/test/resources/org/elasticsearch/xpack/security/transport/ssl/certs/simple/openldap.crt
```
I think these are just really deep paths to the resources (not sure why , there is a bit amount ES specific plumbing to loading these resouces)

gwbrown: Ah yeah I gotcha, not sure what I was thinking here. Nevermind!

jakelandis:
@elasticmachine update branch

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ssl/cert/CertificateInfo.java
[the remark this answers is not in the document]
ywangd: First of all, thanks for indulging me with this discussion. We all agree that it does not really matter for this PR (it is nitpicking). Therefore I am not trying to make a strong case here and also do not want to block it from merging as is. 

> If we threw an exception here, for 8.7+ you could no longer use that existing API in an 8.4-,8.7+ mixed clusters. That would likely be undesirable and possibly a bug, so in 8.7 we would need add support for mixed clusters for these version ranges

That was my intention. My thinking is that we address it when it is actually needed. Ideally a test case should catch it in 8.7 before releasing. So it won't become a production bug. But there is a good chance that the test coverage is not sufficient and it gets slipped into a released version. So I admit it is a bit risky.

> It can be argued that we could remove the code/test for transport serialization here

I did not mean that. The proposed exception is only thrown if serialisation is performed cross specific versions. Please correct me if I am wrong, I think most serialisation tests do not test for cross versions? So no need to change the serialisation tests.

> Keeping the "" as default from elder versions, in future cases, you may actually get back an "" as the value

I do not particularly like using empty string as the special value. I feel it has pros and cons compared to `null`. I don't think `null` is great. But as a special value, `null` is at the language level (and fails loudly) while the empty string is our choosing. 

Again, I have no strong views, just clarifying my thoughts. Hopefully it could be more relevant in some future context. Thanks!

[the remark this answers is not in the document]
jakelandis: tl;dr - will update PR to conditionally omit the empty string return value

@ywangd and I met in person yesterday and here are the notes and outcome. 

We agreed:
* it doesn't really matter for this PR but this conversation helps with future alignment where it could matter
* (we agreed) that the purpose of the transport serialization while not in use toady is added to classes like these to help enable future cases 
* (we agreed) this pattern will likely repeat itself since much of the security code runs on the coordinating node (i.e. no transport serialization) 

We ruled out:
* throwing an exception since it runs counter to the goal to enable future use cases. Additionally our mixed cluster testing for x-pack testing is also [lacking](https://github.com/elastic/elasticsearch/issues/31096). However in general, throwing exceptions when the decision is unclear and can be deferred is generally a good approach, but we don't want to defer here. 

So this leave two areas of discussion:
* external serialization
* internal representation

For external serialization we can emit a key/value pairing where the value is null, "", or something like "unknown". In this context, the object we are serializing represents a certificate and the field is the issuer. It is not possible for a certificate to not have issuer, so strictly speaking emitting anything but the acutal value for the issuer is incorrect. So we believe correct answer for how to represent this externally is to omit the value if can not determine the value. It adds a tiny amount of additional code that is may never be used. This is OK because but it is easily testable and is in line with the goals of ensuring that these POJO type classes are ready for future expanded use cases. I will update the PR with this change.

For internal representation, it does not really matter. It probably makes the most sense to represent the variable as null since it is missing, but that forces a mismatch of the constructors. Where one constructor sets the variable to be null and the other enforces it as non-null. Since the non-null expectations are the primary and current workflow we will optimize for that an internally represent as "". 

I know this is way overkill for this PR, but getting pedantic now on an issue that does not matter means that we don't need to distract from future more meaningful PRs. I will look to updating some internal documentation with this kind of guidance. 

Thanks again @yangd and @gwbrown for the conversation here, and @sethmlarson for updating our internal REST API guidance for how to handle null/empty values.

PR #88447, opened 2022-07-11
Make ClusterInfo use immutable maps in all cases

original-brownbear:
This class's maps are used very hot in the disk threshold allocation
decider. Moving them from hppc maps to unmodifiable map wrapping
`HashMap` has led to a measurable slowdown in the many-shards benchmark
bootstrapping. Lets use immutable map copies here exclusively to make
performance outright better and more predictable via a single implementation.

relates #77466

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/InternalClusterInfoService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/InternalClusterInfoService.java
@@ -39,7 +39,6 @@ import org.elasticsearch.monitor.fs.FsInfo;
 import org.elasticsearch.threadpool.ThreadPool;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


original-brownbear:
Jenkins run elasticsearch-ci/part-2 (unrelated)

server/src/main/java/org/elasticsearch/cluster/InternalClusterInfoService.java
```diff
@@ -262,10 +261,10 @@ public void onResponse(IndicesStatsResponse indicesStatsResponse) {
...
 
                     indicesStatsSummary = new IndicesStatsSummary(
-                        Collections.unmodifiableMap(shardSizeByIdentifierBuilder),
-                        Collections.unmodifiableMap(shardDataSetSizeBuilder),
-                        Collections.unmodifiableMap(dataPathByShardRoutingBuilder),
-                        Collections.unmodifiableMap(rsrvdSpace)
+                        Map.copyOf(shardSizeByIdentifierBuilder),
+                        Map.copyOf(shardDataSetSizeBuilder),
+                        Map.copyOf(dataPathByShardRoutingBuilder),
+                        Map.copyOf(rsrvdSpace)
```
idegtiarenko, line 267: Since type of the map is critical, would it be better to move this inside of the constructor so that it is impossible to create the object with wrong (slower) map implementation?

idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

GrandFisher:
> This class's maps are used very hot in the disk threshold allocation decider. Moving them from hppc maps to unmodifiable map wrapping `HashMap` has led to a measurable slowdown in the many-shards benchmark bootstrapping. Lets use immutable map copies here exclusively to make performance outright better and more predictable via a single implementation.
> 
> relates #77466

@original-brownbear 
Hi, How much of a performance improvement is this, and what is the minimum supported JDK version?  
I'm curious about the performance difference between these two different map（hppc and immutable map） implementations.（hppc and immutable map）

PR #88450, opened 2022-07-11
[ML] add new cache_size parameter to trained_model deployments API

benwtrent:
With: https://github.com/elastic/ml-cpp/pull/2305 we now support caching pytorch inference responses per node per model.

By default, the cache will be the same size has the model on disk size. This is because our current best estimate for memory used (for deploying) is 2*model_size + constant_overhead. 

This is due to the model having to be loaded in memory twice when serializing to the native process. 

But, once the model is in memory and accepting requests, its actual memory usage is reduced vs. what we have "reserved" for it within the node.

Consequently, having a cache layer that takes advantage of that unused (but reserved) memory is effectively free. When used in production, especially in search scenarios, caching inference results is critical for decreasing latency.


benwtrent:
@elasticmachine update branch

davidkyle (approved):
LGTM

I'd prefer a more conservative default cache size. Consider an rather optimal case where we are caching the floats at 4bytes each and the result tensor contains 512 elements = 2048 bytes = 2Kb 

A 2MB cache could contain 1000 hits, and 20MB 10,000 hits. 

200MB is ~ 100,000 hits, I worry we are storing the long tail at that point. 

10-20MB seems a reasonable default, we can always increase the default harder to decrease it.

docs/reference/ml/trained-models/apis/start-trained-model-deployment.asciidoc
```diff
@@ -55,6 +55,11 @@ include::{es-repo-dir}/ml/ml-shared.asciidoc[tag=model-id]
 [[start-trained-model-deployment-query-params]]
 == {api-query-parms-title}
 
+`cache_size`::
+(Optional, <<byte-units,byte value>>)
+The inference cache size (off JVM heap) per node for the model. The default value is the same size as the model
```
lcawl, line 60: Drive-by suggestions: "off ... heap" is unclear to me. If it's part of the heap, "in" seems better to me:
```suggestion
The inference cache size (in the JVM heap) per node for the model. The default value is the same size as the model
```
```suggestion
The inference cache size (off JVM heap) per node for the model. The default value is the same size as the model
```

docs/reference/ml/trained-models/apis/start-trained-model-deployment.asciidoc
```diff
@@ -55,6 +55,11 @@ include::{es-repo-dir}/ml/ml-shared.asciidoc[tag=model-id]
...
+size estimate. To disable cache, `0b` can be provided.
```
lcawl, line 61:
```suggestion
size estimate. To disable the cache, use `0b`.
```
```suggestion
size estimate. To disable cache, `0b` can be provided.
```

lcawl, line 61: > model size estimate. 

Is that the same as the `model_size_bytes` returned by the get trained models API? If yes, it might be helpful to reference that here.

rest-api-spec/src/main/resources/rest-api-spec/api/ml.start_trained_model_deployment.json
```diff
@@ -28,6 +28,12 @@
       ]
     },
     "params":{
+      "cache_size": {
+        "type": "string",
+        "description": "A byte-size value (e.g. 20mb) for configuring the inference cache size",
```
lcawl, line 33:
```suggestion
        "description": "A byte-size value for configuring the inference cache size. For example, 20mb. ",
```
```suggestion
        "description": "A byte-size value (e.g. 20mb) for configuring the inference cache size",
```

docs/reference/ml/trained-models/apis/start-trained-model-deployment.asciidoc
benwtrent: I mean "outside" the jvm heap, not part of the heap

droberts195: I think we need to explain that this comes out of ML native memory (using whatever terminology we use for that elsewhere), but also explain that values up to the on-disk size of the model are effectively free.

lcawl (approved):
One suggestion, otherwise docs LGTM

benwtrent:
@elasticmachine update branch

benwtrent:
Here are some numbers:
```
| task (model size)             | base    | full-cached  | partial-cache   |
|-------------------------------|---------|--------------|-----------------|
| ner model (253.2mb)           | 1700555 | x            | 1498322 (1.13x) |
| text_classification (475.5mb) | 8038894 | 43292 (185x) | x               |
```
NER classification output is much larger as every token gets a classification value. Consequently the cache can be filled up much quicker, but even in partial caching, there is a speed improvement when reindexing the same docs multiple times.

Text classification is one of the best examples of caching. The model response is much smaller than the input. Consequently many more items can be in cache. In this particular test, we got 100% cache hit (it was only over 38k docs) these numbers are on just a single thread && node.

docs/reference/ml/trained-models/apis/start-trained-model-deployment.asciidoc
benwtrent: I will look at our docs @droberts195 and update this line

benwtrent:
@elasticmachine update branch

PR #88453, opened 2022-07-11
Pass IndexMetadata to AllocationDecider.canRemain

original-brownbear:
We need the metadata in a number of allocation deciders and pass it to other allocation methods.
Passing it here avoids redundant lookups across deciders.

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/AllocationDecider.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/AllocationDecider.java
@@ -42,7 +42,7 @@ public abstract class AllocationDecider {
      * Returns a {@link Decision} whether the given shard routing can be remain
      * on the given node. The default is {@link Decision#ALWAYS}.
      */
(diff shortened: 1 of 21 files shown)


DaveCTurner:
Looks like it might make sense to add this to `canAllocate` too?

original-brownbear:
> Looks like it might make sense to add this to canAllocate too?

Maybe. If we want it I'd do it in a follow-up though. This one is already quite noisy and doing it in `canAllocate` will be even noisier. I wonder if it's worth it though, `canAllocate` is obviously not nearly as hot as `canRemain` and in fact I've never seen it show up much in profiling?

DaveCTurner (approved):
That is a little surprising - if rebalancing is allowed then every rebalance attempt should call `canAllocate` on every ⟨index,node⟩ pair. Maybe rebalancing isn't allowed much in these benchmarks?

That said, the balancer uses the `canAllocate` variant which already accepts an `IndexMetadata`.

Anyway LGTM.

original-brownbear:
Thanks David!

> Maybe rebalancing isn't allowed much in these benchmarks?

That's what I would expect yes.

PR #88458, opened 2022-07-12
Script: Ingest Metadata and CtxMap

stu-elastic:
Create a `Metadata` superclass for ingest and update contexts.

Create a `CtxMap` superclass for `ctx` backwards compatibility in ingest and update contexts.  `script.CtxMap` was moved from `ingest.IngestSourceAndMetadata`

`CtxMap` takes a `Metadata` subclass and validates update via the `FieldProperty`s passed in.

`Metadata` provides typed getters and setters and implements a `Map`-like interface, making it easy for a class containing `CtxMap` to implement the full `Map` interface.

The `FieldProperty` record that configures how to validate fields. Fields have a `type`, are `writeable` or read-only, and `nullable` or not and may have an additional validation useful for Set/Enum validation.


stu-elastic:

server/src/main/java/org/elasticsearch/ingest/IngestCtxMap.java
```diff
@@ -0,0 +1,150 @@
...
+        Object ts = ingestMetadata.get(IngestDocument.TIMESTAMP);
+        if (ts instanceof ZonedDateTime timestamp) {
+            return timestamp;
+        } else if (ts instanceof String str) {
+            return ZonedDateTime.parse(str);
+        }
+        return null;
+    }
+
+    static class IngestMetadata extends Metadata {
```
rjernst, line 79: Since this is static, let's move it out to a tope level class. It can still be package private.

server/src/main/java/org/elasticsearch/script/Metadata.java
```diff
@@ -50,80 +46,31 @@ public class Metadata {
...
-            metadata.put(IngestDocument.Metadata.ROUTING.getFieldName(), routing);
-        }
-        if (versionType != null) {
-            metadata.put(IngestDocument.Metadata.VERSION_TYPE.getFieldName(), VersionType.toString(versionType));
-        }
-        return metadata;
-    }
-
     /**
      * Check that all metadata map contains only valid metadata and no extraneous keys and source map contains no metadata
```
rjernst, line 60: nit: I think this comment is off? there is no source map

rjernst (approved):
This looks fine. I don't quite see yet how this will get to what we discussed before (subclasses of Metadata passing the operations they support for keys, ie read-only vs read-write), but I know there are more followups. This PR on it's own looks like an improvement.

server/src/main/java/org/elasticsearch/ingest/IngestCtxMap.java
stu-elastic: There is already an [IngestMetadata](https://github.com/elastic/elasticsearch/blob/f8f440af95c68e8934e2472b7472f77fcf1a1923/server/src/main/java/org/elasticsearch/ingest/IngestMetadata.java) that "Holds the ingest pipelines that are available in the cluster".

I'm going to rename this to `IngestDocMetadata` when I make it top-level.

server/src/main/java/org/elasticsearch/script/Metadata.java
```diff
@@ -172,7 +119,7 @@ public void setVersion(long version) {
     }
 
     public ZonedDateTime getTimestamp() {
```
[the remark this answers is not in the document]
stu-elastic: It is available in 2 of 4 write contexts, ingest and update.  It is missing in reindex and update by query

It's here because that allows scripts to do `Metadata m = metadata(); m.timestamp`.  Otherwise, the script would have to name the subtype.

It is possible to add timestamp to reindex and update by query, both have access to the thread pool and pull a long supplier from it.

server/src/main/java/org/elasticsearch/script/Metadata.java
stu-elastic: Updated.

server/src/main/java/org/elasticsearch/ingest/IngestCtxMap.java
```diff
@@ -0,0 +1,76 @@
...
+    IngestCtxMap(
+        String index,
+        String id,
+        long version,
+        String routing,
+        VersionType versionType,
+        ZonedDateTime timestamp,
+        Map<String, Object> source
+    ) {
+        super(new HashMap<>(source), new IngestDocMetadata(index, id, version, routing, versionType, timestamp));
```
jdconrad, line 46: Since we do a deepcopy for source when calling this constructor, do we then again need to wrap source in a new HashMap?

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
```diff
@@ -763,7 +751,7 @@ public static Object deepCopy(Object value) {
             for (Map.Entry<?, ?> entry : mapValue.entrySet()) {
                 copy.put(entry.getKey(), deepCopy(entry.getValue()));
             }
-            // TODO(stu): should this check for IngestSourceAndMetadata in addition to Map?
+            // TODO(stu): should this check for IngestCtxMap in addition to Map?
```
jdconrad, line 754: Did you mean to leave this as part of the PR?

server/src/main/java/org/elasticsearch/script/Metadata.java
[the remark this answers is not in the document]
jdconrad: We could duck-type the Metadata subclasses just for timestamp.

jdconrad (approved):
LGTM. A few minor comments.

server/src/main/java/org/elasticsearch/ingest/IngestDocument.java
stu-elastic: It shows up because of the rename, it's still relevant question for now.

server/src/main/java/org/elasticsearch/script/Metadata.java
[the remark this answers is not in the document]
rjernst: If it’s relevant to all the other contexts, then it makes sense to stay completely in this class. But I don’t think we should have it defined here but implemented in subclasses.

server/src/main/java/org/elasticsearch/ingest/IngestCtxMap.java
stu-elastic: A bunch of tests call into IngestDocument which calls this constructor, those test all do Map.of or SingletonMap and fail if we don't perform this copy.

server/src/main/java/org/elasticsearch/script/Metadata.java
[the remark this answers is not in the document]
stu-elastic: I'll change the implementation here in the Update PR.

PR #84757, opened 2022-03-08
SQL: Pin specific py modules used by Tableau connector SDK

bpintea:
This pins certain Python modules used by Tableau's connector SDK to
certain versions. This will ensure build success consistency.

PR #88462, opened 2022-07-12
Update Tableau connector to use connection dialog v2

luigidellaquila:


luigidellaquila:

luigidellaquila:
@elasticmachine update branch

x-pack/plugin/sql/connectors/tableau/package.sh
```diff
@@ -78,7 +78,6 @@ function package() {
     # install environment
     cd $SDK_DIR/connector-packager
     # apply pip-installed modules version pinning
-    patch setup.py $MY_TOP_DIR/setup.py.diff
```
bpintea, line 81: Why has this been removed? It might be better to just update the pinned versions, if we know the build works. The reason it has been introduced was because the pip modules will be requiring the latest versions and these can sometimes come with conflicts (https://github.com/elastic/elasticsearch/pull/84757 links to further details).

luigidellaquila: Ok, now I see where it comes from (the patch did not work with the newer versions of course).
I'll update it pinning to latest version

luigidellaquila: Restored the patch and updated the dependencies to latest version

bpintea (approved):
LGTM

astefan (approved):
LGTM

PR #88463, opened 2022-07-12
[ML] Frequent items Use bitsets for faster de-duplication

hendrikmuhs:
**Draft**

Enhancement investigation on top of #83055

By using bitsets instead of lists of longs item sets can be faster de-duplicated. A bit is set according to the order of top items (by count). #83055 isn't merged yet, the change is only in the last commit: https://github.com/elastic/elasticsearch/commit/2b2c14c46102baaf7c7362d3cd46ff17c9aad5bb#diff-36b5a961dff736a5ac8f86ae93d4a86ff32780b76d8811b5e674d73aaebce82c

performance comparison: A first testrun indicates a runtime improvement 10-20%.

Notes: 
 - the bitset might be useful for transactions
 - bitsets reduce memory requirements(memory for remembering collected sets)


hendrikmuhs:
superseeded by #88943

PR #88465, opened 2022-07-12
Use consistent shard map type in IndexService

original-brownbear:
It's faster and easier to reason about if we always have
an immutable collections map here and not have the type depend
on what the last operation on the index service was.


idegtiarenko (approved):

original-brownbear:

Issue #88408, opened 2022-07-10
ctx.execution_time yields strange results in watcher compare condition


sakurai-youhei:
### Elasticsearch Version

7.17.5 (since 6.6.0 but 8.x is not affected)

### Installed Plugins

n/a

### Java Version

_bundled_

### OS Version

doesn't matter

### Problem Description

The following example from [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/condition-compare.html) results in `"met": true` on 8.x but doesn't on 7.x

```json
{
  "condition" : {
    "compare" : {
      "ctx.execution_time" : {
        "gte" : "<{now-5m}>"
      }
    }
  }
}
```

Getting to the bottom, the following test case succeeds in master branch but fails in 7.17. branch.

https://github.com/sakurai-youhei/elasticsearch/blob/58d57bed48d31b1c9d842d70889003106477b9a4/x-pack/plugin/watcher/src/test/java/org/elasticsearch/xpack/watcher/condition/CompareConditionTests.java#L191-L200

```
    public void testExecuteDateMathExecutionContext() throws Exception {
        ClockMock clock = ClockMock.frozen();
        boolean met = true;
        Op op = CompareCondition.Op.GTE;
        String value = "<{now-5m}>";

        CompareCondition condition = new CompareCondition("ctx.execution_time", op, value, clock);
        WatchExecutionContext ctx = mockExecutionContext("_name", null);
        assertThat(condition.execute(ctx).met(), is(met));
    }
```

That's because [LenientCompare.java](https://github.com/elastic/elasticsearch/blob/v7.17.5/x-pack/plugin/watcher/src/main/java/org/elasticsearch/xpack/watcher/condition/LenientCompare.java#L59-L77) doesn't expect `org.elasticsearch.script.JodaCompatibleZonedDateTime` of which `ctx.execution_time` is an instance in 7.x.

### Steps to Reproduce

1. Spin up a new Elasticsearch cluster.
2. Start a trial by `curl -XPOST "http://localhost:9200/_license/start_trial?acknowledge=true"` if needed.
3. Run the following two commands with `watch.json` file.

```shell
curl -XPUT "http://localhost:9200/_watcher/watch/test" -H "Content-Type: application/json" -d@watch.json
curl -XPOST "http://localhost:9200/_watcher/watch/test/_execute" | jq .watch_record.result.condition
```

watch.json
```json
{
  "trigger": {
    "schedule": {
      "interval": "10s"
    }
[...]

PR #88466, opened 2022-07-12
Add one test for execution context in watcher compare condition

sakurai-youhei:
Issue: https://github.com/elastic/elasticsearch/issues/88408

This PR adds one test for execution context in the watcher compare condition, which is equivalent to the below compare condition.

```json
{
  "condition" : {
    "compare" : {
      "ctx.execution_time" : {
        "gte" : "<{now-5m}>"
      }
    }
  }
}
```

This test should always pass and actually passes in master branch but fails in 7.17 branch. So, backport PR will append necessary adjustment (bug fix) on top of this.

**Question**: I don't think this PR needs a changelog, right?


stu-elastic (approved):
lgtm

Issue #86440, opened 2022-05-04
Duplicate mappers consume too much heap memory


original-brownbear:
Even with recent fixes to the heap consumption of field mappers, large frozen nodes require a lot of heap memory for duplicate mappers. An example from benchmarking today's master branch shows 67 different mappings taking up only ~100kb compressed in the cluster state on a data node holding ~16k indices/shards.
On the other hand the actual `MappingLookup` instances (and thus the mapper service instances) consume 6.8G of heap. 

![image](https://user-images.githubusercontent.com/6490959/166806416-8f4ebc69-ba2a-4e68-a682-0e6668f2b1e2.png)

In fact, in the many shards benchmark that this is extracted from (that has this data node run indexing and searching at the time the dump was taken), the `MappingLookup` instances are the single largest consumer of heap.
It seems to me there's no technical reason why we should not be able to massively deduplicate objects here when all the mappings are the same. We already do a lot of deduplication via tricks like https://github.com/elastic/elasticsearch/pull/86301 that helped bring down the consumption a lot recently but it seems we should go a little further here and tackle this at a higher level given the numbers?

Marking/suggesting this as ">bug" since it diminishes the scalability of frozen/warm/cold nodes greatly and should be somewhat straigh-forward to fix hopefully that classification is obviously debatable.

ywelsch:
I've looked at bit deeper at the memory usage here. `MappingLookup` is indeed holding onto the majority of memory when having lots of indices on a frozen node. I've analyzed where and why so much memory is being spent there. What plays a factor here is:
- the amount of memory taken by `MappedFieldType` and `Mapper` objects. For each field that's defined in the mapping, there is a corresponding  `MappedFieldType` and `Mapper` object. The memory of these depend on the respective implementation, e.g. `NumberFieldMapper`/`NumberFieldType` take less memory than the `TextFieldMapper`/`TextFieldType` implementations. Typically the space used by these implementations ranges from 10 to 40 bytes per field definition.
- the amount of memory taken by the lookup structures in `MappingLookup`. The `MappingLookup` class is comprised of a couple of lookup structures that allows looking up fields by their fully expanded path name. 
[...]

PR #88475, opened 2022-07-12
Extract field name from MappedFieldType

ywelsch:
Separates the field name from MappedFieldType, making MappedFieldType just a container for type-related information.

The combination of field name and typing information is now available via a new MappedField class.

As follow-up, we would like to implement equals/hashCode on this new name-stripped MappedFieldType, so that we can deduplicate MappedFieldType instances. This should dramatically lower the memory usage of large mappings on data nodes.

Relates https://github.com/elastic/elasticsearch/issues/86440


modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/RankFeatureQueryBuilder.java
```diff
@@ -378,25 +378,25 @@ protected void doXContent(XContentBuilder builder, Params params) throws IOExcep
 
     @Override
     protected Query doToQuery(SearchExecutionContext context) throws IOException {
-        final MappedFieldType ft = context.getFieldType(field);
+        final MappedField mappedField = context.getMappedField(field);
 
-        if (ft instanceof final RankFeatureFieldType fft) {
-            return scoreFunction.toQuery(RankFeatureMetaFieldMapper.NAME, field, fft.positiveScoreImpact());
-        } else if (ft == null) {
+        if (mappedField == null) {
```
ywelsch, line 383: I reordered the if/else clauses here to move the null check to the top (as we can't otherwise call `mappedField.type()` without a NPE)

server/src/main/java/org/elasticsearch/index/mapper/MappedFieldType.java
```diff
@@ -58,22 +58,19 @@
...
 
     public MappedFieldType(
-        String name,
         boolean isIndexed,
         boolean isStored,
         boolean hasDocValues,
         TextSearchInfo textSearchInfo,
         Map<String, String> meta
     ) {
-        this.name = Mapper.internFieldName(name);
```
ywelsch, line 76: This has moved to the MappedField class now.

server/src/main/java/org/elasticsearch/index/mapper/TextFieldMapper.java
```diff
@@ -962,59 +970,73 @@ public Query fuzzyQuery(
...
-        public Query existsQuery(SearchExecutionContext context) {
+        public Query existsQuery(String name, SearchExecutionContext context) {
             if (context.allowExpensiveQueries() == false) {
                 throw new ElasticsearchException(
                     "runtime-computed exists query cannot be executed while [" + ALLOW_EXPENSIVE_QUERIES.getKey() + "] is set to [false]."
                 );
             }
-            return existQueryFieldType.existsQuery(context);
+            // norms are not available, neither are doc-values, so fall back to _source to run exists query
+            final MappedField existQueryField = KeywordScriptFieldType.sourceOnly(name).asMappedFields().findFirst().get();
```
ywelsch, line 1038: this was moved from the constructor of LegacyTextFieldType now to the invocation site here, as we need the `name` to construct the existQueryField

romseygeek (approved):
LGTM.  It would be good to have some javadoc on the new FieldMapper class too.

javanna:
We did not go ahead with this change, see https://github.com/elastic/elasticsearch/issues/86440#issuecomment-1188978720 .

Issue #86724, opened 2022-05-12
[CI] SnapshotStressTestsIT testRandomActivities failing


pugnascotia:
**Build scan:**
https://gradle-enterprise.elastic.co/s/anyfhhx6i7naw/tests/:server:internalClusterTest/org.elasticsearch.snapshots.SnapshotStressTestsIT/testRandomActivities

**Reproduction line:**
`./gradlew ':server:internalClusterTest' --tests "org.elasticsearch.snapshots.SnapshotStressTestsIT.testRandomActivities" -Dtests.seed=EB88608F2635B7C9 -Dtests.locale=tr-TR -Dtests.timezone=Pacific/Wake -Druntime.java=17`

**Applicable branches:**
8.2

**Reproduces locally?:**
No

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.snapshots.SnapshotStressTestsIT&tests.test=testRandomActivities

**Failure excerpt:**
```
java.lang.AssertionError: failed to acquire all permits: [repo-0, index-0, index-1, index-2, node_s5, node_s4, node_s3, node_s2, node_s1, node_s0]

  at org.junit.Assert.fail(Assert.java:88)
  at org.elasticsearch.snapshots.SnapshotStressTestsIT$TrackedCluster.run(SnapshotStressTestsIT.java:345)
  at org.elasticsearch.snapshots.SnapshotStressTestsIT.testRandomActivities(SnapshotStressTestsIT.java:84)
  at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(NativeMethodAccessorImpl.java:-2)
  at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
  at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.lang.reflect.Method.invoke(Method.java:568)
  at com.carrotsearch.randomizedtesting.RandomizedRunner.invoke(RandomizedRunner.java:1758)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$8.evaluate(RandomizedRunner.java:946)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$9.evaluate(RandomizedRunner.java:982)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$10.evaluate(RandomizedRunner.java:996)
[...]

PR #88476, opened 2022-07-12
Avoid noisy exceptions on data nodes when aborting snapshots

original-brownbear:
Currently, an abort (especially when triggered an index delete) can
manifest as either an aborted snapshot exception, a missing index exception or
an NPE. The latter two show up as noise in logs.
This change catches effectively all of these cleanly as aborted snapshot
exceptions so they don't get logged as warnings and avoids the NPE if
a shard was removed from the index service concurrently by using the
API that throws on missing shards to look it up.

Seen in this noisy failure https://github.com/elastic/elasticsearch/issues/86724#issuecomment-1180783691


idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

PR #87126, opened 2022-05-25
Deduplicate mappings in persisted cluster state

DaveCTurner:
WIP

PR #82608, opened 2022-01-14
More Compact Serialization of Metadata

original-brownbear:
Serialize the map of hashes to mappings and then lookup from the map instead
of serializing them over and over for each index to make full cluster state
transport messages much smaller in the common case of many duplicate mappings.

This should make the master node impact of requests for the full cluster state (or at least the state including mappings) quite a bit cheaper memory+cpu+network wise. Also it saves lots of buffers on the coordinating/sending node as well as CPU for deduplicating mappings.

relates #77466

PR #88479, opened 2022-07-12
Deduplicate mappings in persisted cluster state

joegallo:
Replacement for #87126, related to #77466

Following up on https://github.com/elastic/elasticsearch/pull/82608, which compacted the network serialization of mappings in the cluster state's metadata. In that PR, mappings are serialized as a map keyed by hash, and then when serializing index metadata the appropriate mapping can be referenced by its hash only, saving network traffic in the (expected common) case of indices with duplicate mappings.

This PR extends that same logic to the on-disk serialization of a cluster state via the PersistedClusterStateService. The efficiency gain in this case is greater, however, because in the case of an index settings change, for example, the mapping and associated hash are unchanged and so the on-disk mapping doesn't need to be re-persisted.

/cc @DaveCTurner


joegallo:
I think this is ready for review, @original-brownbear if you have the time and inclination to take a look. I won't merge it without an explicit ✅ from @DaveCTurner, though, and I know I won't get that until he's back early next week.

server/src/test/java/org/elasticsearch/gateway/PersistedClusterStateServiceTests.java
```diff
@@ -1629,6 +1894,18 @@ private NodeEnvironment newNodeEnvironment(Path[] dataPaths) throws IOException
         );
     }
 
+    private static MappingMetadata randomMappingMetadata(boolean allowNull) {
```
DaveCTurner, line 1897: nit: slight preference for this being two methods `randomMappingMetadata` vs `randomMappingMetadataOrNull`

DaveCTurner (approved):
LGTM with one style nit in the tests

PR #88481, opened 2022-07-12
Rename immutable cluster state to reserved cluster state

grcevski:
For the purpose of implementing file based settings we had to create the concept of cluster state that isn't modifiable by the REST handlers. We initially called that state 'immutable' cluster state, since it can't be changed by the REST handlers. However, the naming is confusing for developers that make state handlers which modify this state.

To remedy this we are renaming the immutable cluster state to 'reserved', meaning the cluster state is reserved by someone and it cannot be modified by the REST handlers. Typically, that 'someone' that reserves the cluster state is the cloud operator through file based settings, or plugins/modules when they install their initial cluster settings.


rjernst (approved):
Lgtm

grcevski:
Thanks Ryan!

PR #88485, opened 2022-07-12
[ML] adjust internal inference object to accept request_id in outer object

benwtrent:
Since `request_id` is effectively required for all messages sent to the native process, this bubbles up the request_id to the outer object.

It is currently optional (until the native side is changed). 

If an inner object has the request_id set, and the outer object does not, the outer object request ID is set to that inner one.


davidkyle (approved):
LGTM

Please hold of merging for now in case we have to format again - the back end change isn't as simple as I imagined

droberts195:
To make this most useful the cache hit indicator also needs to be on the top level.

I'd also prefer to make the Java and C++ changes in parallel and mute tests during the few hours where they're out-of-sync.  It will make the code history cleaner in the future if anyone wants to look back at what was changed.

Since you're going to be on leave soon and this PR already has lots of merge conflicts I will close it and copy the relevant bits to a new PR in my fork, which can then get tested in combination with a C++ side PR in the C++ CI.

PR #88487, opened 2022-07-12
Support cartesian shape with doc values

craigtaverner:
Also refactoring testing infrastructure to make testing easier to share across different test classes.


x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/index/mapper/ShapeDocValuesQuery.java
```diff
@@ -0,0 +1,197 @@
...
+        }
+        this.field = field;
+        this.encoder = encoder;
+        this.relation = relation;
+        this.geometries = geometries;
+    }
+
+    protected abstract Component2D create(GEOMETRY geometry);
+
+    protected abstract Component2D create(GEOMETRY[] geometries);
```
iverase, line 53: this is nice, do we need two methods? would not be enough with something like:
```
protected abstract Component2D create(GEOMETRY... geometry);
```

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/index/mapper/ShapeDocValuesQuery.java
```diff
@@ -0,0 +1,197 @@
...
+            public boolean isCacheable(LeafReaderContext ctx) {
+                return DocValues.isCacheable(ctx, field);
+            }
+
+        };
+    }
+
+    /** Override if special cases, like dateline support, should be considered */
+    protected boolean addSpecialCase(List<Component2D> components2D, GEOMETRY geometry) {
+        return false;
```
iverase, line 143: Not strong opinion but I wonder if we should change the API to somethng like:

```
    protected abstract Component2D create(GEOMETRY... geometry);

    protected abstract void add(List<Component2D> components2D, GEOMETRY geometry);
```

So we don't need this special case.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/index/fielddata/LabelPositionVisitor.java
```diff
@@ -7,54 +7,58 @@
...
 
 /**
- * Get the first node of the tree and provide a point in that gemetry (point, line or triangle)
+ * Get the first node of the tree and provide a point in that geometry (point, line or triangle)
  * as a suggested label position likely to be somewhere in the middle of the entire geometry.
  *
  * TODO: We could instead choose the point closer to the centroid which improves unbalanced trees
  */
-public class LabelPositionVisitor implements TriangleTreeReader.Visitor {
+public class LabelPositionVisitor<T extends ToXContentFragment> implements TriangleTreeReader.Visitor {
```
iverase, line 20: Do we need to extend ToXContentFragment here? I think we don't need any method from that interface.

server/src/main/java/org/elasticsearch/index/mapper/ShapeIndexer.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+package org.elasticsearch.index.mapper;
+
+import org.apache.lucene.index.IndexableField;
+import org.elasticsearch.geometry.Geometry;
+
+import java.util.List;
+
+public interface ShapeIndexer {
```
iverase, line 16: Add some java docs?

x-pack/plugin/spatial/src/test/java/org/elasticsearch/xpack/spatial/index/mapper/ShapeFieldMapperTests.java
```diff
@@ -72,6 +72,12 @@ protected void registerParameters(ParameterChecker checker) throws IOException {
         });
     }
 
+    @Override
```
iverase, line 75: I am missing here explicit test for doc_values, like that they are generated when set the flag to true or that the default value is true in current version but. false in previous versions.

x-pack/plugin/spatial/src/test/java/org/elasticsearch/xpack/spatial/index/mapper/ShapeFieldMapperTests.java
```diff
@@ -72,6 +72,12 @@ protected void registerParameters(ParameterChecker checker) throws IOException {
...
+    protected boolean supportsSearchLookup() {
+        // TODO: is this really true? We have failing tests, but they look like test issues not core issues
+        return false;
```
iverase, line 78: It seems that this was not needed before because we did not have doc values so it was equivalent to set this to false. Now we need to do it explicitly.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/index/fielddata/ShapeValues.java
```diff
@@ -0,0 +1,259 @@
...
+ *   if (values.advanceExact(docId)) {
+ *     GeoValue value = values.value()
+ *     final int numValues = values.count();
+ *     // process value
+ *   }
+ * </pre>
+ *
+ * There is just one value for one document.
+ */
+public abstract class ShapeValues<T extends ToXContentFragment> {
```
iverase, line 41: Do we need this class in this PR?

iverase (changes requested):
This is looking pretty solid, just left some small comments. 

The most important thing to improve is the test coverage of the doc_value flag in ShapeFieldMapperTests, in particular making sure that set the flag to true, it generates doc_values and that the default value changes depending on the versions.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/common/stats/EnumCounters.java
```diff
@@ -118,11 +118,9 @@ public String toString() {
         StringBuilder buf = new StringBuilder("[");
         boolean first = true;
         for (E e : enums) {
+            if (first == false) buf.append(", ");
+            first = false;
```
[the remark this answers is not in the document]
craigtaverner: https://github.com/elastic/elasticsearch/pull/88505

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/index/fielddata/ShapeValues.java
craigtaverner: Done in e52ba263c3d118caf23d1da4358b9f8d95076702

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/index/mapper/ShapeDocValuesQuery.java
craigtaverner: That is what I did originally, but it turns out varargs does not play with generics, so in the generic class you have to use the two methods to get the same effect.

craigtaverner: I think it is because with type erasure we cannot tell the difference between Object[] and Object (since Object[] is also an Object).

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/index/mapper/ShapeDocValuesQuery.java
craigtaverner: Interesting approach. Is easier to read and reason about, so I've changed it to that, sort-of.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/index/fielddata/ShapeValues.java
craigtaverner: Good catch, indeed. I originally pulled in some of the other generic refactoring from the prototype, realised I did not need it and reverted the classes that extend this class, but forgot to remove this class itself!

x-pack/plugin/spatial/src/test/java/org/elasticsearch/xpack/spatial/index/mapper/ShapeFieldMapperTests.java
craigtaverner: Added in 1066e01c98b3d8f5bd9873803366d5a2d7785f80

iverase (approved):
LGTM

PR #88491, opened 2022-07-13
Add transport action immutable state checks

grcevski:
This PR is splitting the commit to add the code that verifies if a master node action is modifying state that was set in operator mode. 

Relates to #88224


server/src/main/java/org/elasticsearch/action/support/master/TransportMasterNodeAction.java
```diff
@@ -172,9 +175,39 @@ protected Set<String> modifiedKeys(Request request) {
...
+        for (ImmutableStateMetadata metadata : state.metadata().immutableStateMetadata().values()) {
+            Set<String> conflicts = metadata.conflicts(handlerName.get(), modified);
+            if (conflicts.isEmpty() == false) {
+                errors.add(format("[%s] set as read-only by [%s]", String.join(",", conflicts), metadata.namespace()));
+            }
+        }
+
+        if (errors.isEmpty() == false) {
+            throw new IllegalArgumentException(
+                format("Failed to process request [%s] with errors: %s", request, String.join(System.lineSeparator(), errors))
```
rjernst, line 195: I don't know if we have multiline error messages in other exceptions. Will this IAE propgate through a user request (that would return a 400)? eg a user tries to add an ILM policy with an already reserved name, then they get this exception. If we want to go with newlines, I think we need to make sure this gets serialized correctly (ie the newline needs to be escaped in the json response).

Also, I'm not sure System.lineSeparator is necessary? At least for json only `\n` would be expected.

server/src/test/java/org/elasticsearch/action/support/master/TransportMasterNodeActionTests.java
```diff
@@ -697,6 +759,54 @@ protected ClusterBlockException checkBlock(Request request, ClusterState state)
...
+            clusterService,
+            threadPool,
+            ThreadPool.Names.SAME
+        );
+
+        assertTrue(action.supportsImmutableState());
+
+        assertTrue(
+            expectThrows(IllegalArgumentException.class, () -> action.validateForImmutableState(request, clusterState)).getMessage()
+                .contains("with errors: [a] set as read-only by [namespace_one]\n" + "[e] set as read-only by [namespace_two]")
```
rjernst, line 799: This assertion will break on windows since `\n` is hardcoded (at least with the current exception message)

rjernst (commented):
One thought about the exception generated.

server/src/main/java/org/elasticsearch/action/support/master/TransportMasterNodeAction.java
grcevski: OK, sounds good. I'll wrap the errors as an array with [] and commas. That should solve the windows line separator test issue too.

grcevski: > Will this IAE propgate through a user request (that would return a 400)?

Yes it does go all the way back to the user that issued the REST request.

grcevski:
I switched it to use similar multi line reporting as in PluginsService.java

rjernst (approved):
LGTM

PR #88492, opened 2022-07-13
Refactor json parsing in immutable handlers

grcevski:
The current immutable handlers assume json input, which is inefficient for use by plugins and modules. This PR refactors the xcontent parsing in a separate method to be used only when we actually have json input (file based settings).

This way the immutable state controller can work with POJOs and xcontent too.

This PR is split off from https://github.com/elastic/elasticsearch/pull/88224 to reduce the PR size.

### DIFF
--- a/server/src/main/java/org/elasticsearch/common/util/Maps.java
+++ b/server/src/main/java/org/elasticsearch/common/util/Maps.java
@@ -284,18 +284,4 @@ public class Maps {
         return expectedSize < 2 ? expectedSize + 1 : (int) (expectedSize / 0.75 + 1.0);
     }
 
-    /**
-     * Convenience method to convert the passed in input object as a map with String keys.
-     *
-     * @param input the input passed into the operator handler after parsing the content
-     * @return
-     */
-    @SuppressWarnings("unchecked")
-    public static Map<String, ?> asMap(Object input) {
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/elasticsearch/common/util/Maps.java
```diff
@@ -284,18 +284,4 @@ static int capacity(int expectedSize) {
         return expectedSize < 2 ? expectedSize + 1 : (int) (expectedSize / 0.75 + 1.0);
     }
 
-    /**
```
grcevski, line 287: Not needed anymore with the way I do parsing. This method was added previously by me.

x-pack/plugin/ilm/src/main/resources/META-INF/services/org.elasticsearch.immutablestate.ImmutableClusterStateHandlerProvider
```diff
@@ -0,0 +1,8 @@
+#
```
grcevski, line 1: I added this file because we need it to expose the service provider when we load them in the future using PluginsService createExtension.

rjernst (approved):
LGTM

ChrisHegarty (approved):

PR #88496, opened 2022-07-13
TSDB: Downsampling support cancelled

weizijun:
support cancel downsampling task:

- POST _tasks/oTUltX4IQMOUUVeiohTt8A:12345/_cancel
- POST _tasks/_cancel?actions=\*rollup_indexer\*
- POST _tasks/_cancel? parent_task_id=xxx


x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
```diff
@@ -172,17 +195,23 @@ public void beforeBulk(long executionId, BulkRequest request) {
             public void afterBulk(long executionId, BulkRequest request, BulkResponse response) {
                 numIndexed.addAndGet(request.numberOfActions());
                 if (response.hasFailures()) {
-                    Map<String, String> failures = Arrays.stream(response.getItems())
+                    List<BulkItemResponse> failedItems = Arrays.stream(response.getItems())
```
salvatore-campagna, line 198: I don't get why we need this change here. `failedItems.size()` should be equal to `failures.size()`. Isn't enough to just call `status.setCancelled()`?

x-pack/plugin/rollup/src/test/java/org/elasticsearch/xpack/rollup/v2/RollupActionSingleNodeTests.java
```diff
@@ -114,7 +120,7 @@ public void setup() {
         rollupIndex = "rollup-" + sourceIndex;
         startTime = randomLongBetween(946769284000L, 1607470084000L); // random date between 2000-2020
         docCount = randomIntBetween(10, 9000);
-        numOfShards = randomIntBetween(1, 4);
+        numOfShards = 4;// randomIntBetween(1, 4);
```
salvatore-campagna, line 123: Why not setting this value randomly?

x-pack/plugin/rollup/src/test/java/org/elasticsearch/xpack/rollup/v2/RollupActionSingleNodeTests.java
```diff
@@ -335,6 +341,67 @@ public void testRollupDatastream() throws Exception {
...
+            .field(FIELD_DIMENSION_1, randomAlphaOfLength(1))
+            .field(FIELD_NUMERIC_1, randomDouble())
+            .endObject();
+        bulkIndex(sourceSupplier);
+        prepareSourceIndex(sourceIndex);
+
+        IndicesService indexServices = getInstanceFromNode(IndicesService.class);
+        Index srcIndex = resolveIndex(sourceIndex);
+        IndexService indexService = indexServices.indexServiceSafe(srcIndex);
+        IndexShard shard = indexService.getShard(0);
```
salvatore-campagna, line 359: Should we select the shard randomly?

x-pack/plugin/rollup/src/test/java/org/elasticsearch/xpack/rollup/v2/RollupActionSingleNodeTests.java
weizijun: > Why not setting this value randomly?

sorry, I will revert it

x-pack/plugin/rollup/src/test/java/org/elasticsearch/xpack/rollup/v2/RollupActionSingleNodeTests.java
weizijun: > Should we select the shard randomly?

ok

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
weizijun: when the _id is null, the map key is the same, so the failures.size() is 1, so I use a list to same the failures

salvatore-campagna:
@weizijun do you mind updating this PR so to resolve conflicts? We are reviewing it and considering merging it. Thanks.

weizijun:
> @weizijun do you mind updating this PR so to resolve conflicts? We are reviewing it and considering merging it. Thanks.

Okay!

martijnvg:

weizijun:
@salvatore-campagna @martijnvg there is a notice: the RollupShardStatus is referenced by RollupShardIndexer. I'm not sure if RollupShardStatus will change the reference.
If the reference changes, the new RollupShardStatus changed to ABORT, but the old RollupShardStatus reference by RollupShardIndexer don't changed, so the task can not be stopped.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
```diff
@@ -170,9 +179,12 @@ protected RollupIndexerAction.Response newResponse(
     private class Async extends AsyncBroadcastAction {
         private final RollupIndexerAction.Request request;
         private final ActionListener<RollupIndexerAction.Response> listener;
+        private final Task task;
+        private boolean hasCancelled = false;
```
martijnvg, line 183: I don't think I fully understand that the changes to the `Async` inner class.
It seems to always cancel a shard level rollup before actually performing it?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/RollupShardIndexer.java
```diff
@@ -90,11 +92,15 @@ class RollupShardIndexer {
     private final String[] metricFields;
     private final String[] labelFields;
     private final Map<String, FieldValueFetcher> fieldValueFetchers;
+    private final AtomicLong numReceived = new AtomicLong();
```
martijnvg, line 95: Maybe these counters should be moved to the  `RollupShardTask` class? If this class has access to the task then it can just increments these counters that are part of the task. I think makes managing these counters easier.

martijnvg (changes requested):
Thanks @weizijun for working on this. I left comments around the integration with the task framework.

martijnvg:
> the RollupShardStatus is referenced by RollupShardIndexer. I'm not sure if RollupShardStatus will change the reference.
If the reference changes, the new RollupShardStatus changed to ABORT, but the old RollupShardStatus reference by RollupShardIndexer don't changed, so the task can not be stopped.

I think status classes aren't meant to be passed around to other components and keep track of state. This what the task class is for. When an action executes there should be only one task instance and that should exist until the operation completed (either successfully or failed). In this case for downsampling, a task (created by RollupIndexerAction) will be created for each shard and also a task instance when `TransportRollupAction` invokes the rollup indexer action action.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/action/RollupShardStatus.java
```diff
@@ -0,0 +1,166 @@
...
+    private static final ParseField SHARD_FIELD = new ParseField("shard");
+    private static final ParseField START_TIME_FIELD = new ParseField("start_time");
+    private static final ParseField IN_NUM_DOCS_RECEIVED_FIELD = new ParseField("in_num_docs_received");
+    private static final ParseField OUT_NUM_DOCS_SENT_FIELD = new ParseField("out_num_docs_sent");
+    private static final ParseField OUT_NUM_DOCS_INDEXED_FIELD = new ParseField("out_num_docs_indexed");
+    private static final ParseField OUT_NUM_DOCS_FAILED_FIELD = new ParseField("out_num_docs_failed");
+
+    private final ShardId shardId;
+    private final long rollupStart;
+    private final AtomicLong numReceived;
```
martijnvg, line 38: Change the num* fields to type long?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/action/RollupShardTask.java
```diff
@@ -0,0 +1,72 @@
...
+import org.elasticsearch.tasks.TaskId;
+import org.elasticsearch.xpack.core.downsample.DownsampleConfig;
+import org.elasticsearch.xpack.core.rollup.RollupField;
+
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class RollupShardTask extends CancellableTask {
+    private String rollupIndex;
+    private DownsampleConfig config;
```
martijnvg, line 21: I think `rollupIndex ` and `config ` fields can also be final fields?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/RollupShardIndexer.java
```diff
@@ -90,11 +92,16 @@ class RollupShardIndexer {
     private final String[] metricFields;
     private final String[] labelFields;
     private final Map<String, FieldValueFetcher> fieldValueFetchers;
-    private final AtomicLong numSent = new AtomicLong();
-    private final AtomicLong numIndexed = new AtomicLong();
-    private final AtomicLong numFailed = new AtomicLong();
+    private final AtomicLong numReceived;
```
martijnvg, line 95: Maybe use the use getters on the `task` object directly? Instead of also defining these fields here.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/action/RollupShardTask.java
```diff
@@ -0,0 +1,72 @@
...
+        return rollupIndex;
+    }
+
+    public DownsampleConfig config() {
+        return config;
+    }
+
+    @Override
+    public Status getStatus() {
+        return status;
```
martijnvg, line 54: Instead of creating a status object in the constructor. The status object should be created in this method.
This way a read only copy of the status of the task is created. And then there is no need to pass the atomic longs to the status.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
martijnvg: I think I understand now, why this logic is added and how it works. This for cancelling other shard level operations in case one fails. Downsampling may take a long time to complete and if a downsample fails in one shard, this can speed up the response to the client.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
```diff
@@ -186,5 +198,32 @@ protected void finishHim() {
...
+            super.onOperation(shard, shardIt, shardIndex, e);
+        }
+
+        private void cancelOtherShardIndexers() {
+            if (false == hasCancelled) {
+                cancelTasksAction.execute(
+                    task,
+                    new CancelTasksRequest().setTargetParentTaskId(new TaskId(clusterService.localNode().getId(), task.getId())),
+                    ActionListener.wrap(r -> {
+                        logger.info("[{}] rollup cancel other shard indexers", request.getRollupRequest().getSourceIndex());
```
martijnvg, line 215: Maybe check the `r` response here? To log whether tasks are really cancelled?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
martijnvg: Should the  `hasCancelled` field be volatile? I think it will read / updated from different threads.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
```diff
@@ -50,6 +55,7 @@ public class TransportRollupIndexerAction extends TransportBroadcastAction<
     private final Client client;
     private final ClusterService clusterService;
     private final IndicesService indicesService;
+    private final TransportCancelTasksAction cancelTasksAction;
```
martijnvg, line 58: Maybe use the `client` field here instead of adding `cancelTasksAction` here? The cancel task action is available in `ClusterAdminClient`.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
```diff
@@ -186,5 +198,32 @@ protected void finishHim() {
                 listener.onFailure(e);
             }
         }
+
+        @Override
+        protected void onOperation(@Nullable ShardRouting shard, final ShardIterator shardIt, int shardIndex, Exception e) {
+            // when this shard operation failed, cancel other shard operations
+            cancelOtherShardIndexers();
```
martijnvg, line 205: In case a shard has replicas the behaviour of the super class is to retry on the replica shard.
With the cancel logic, I don't think we want this?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/RollupShardIndexer.java
```diff
@@ -90,11 +92,16 @@ class RollupShardIndexer {
...
+    private final AtomicLong numSent;
+    private final AtomicLong numIndexed;
+    private final AtomicLong numFailed;
+
+    private final RollupShardTask task;
+    private volatile boolean abort = false;
```
martijnvg, line 101: If we set `abort` then the task cancel field isn't set to true. I think this field and task cancel field are very overlapping in term of usage. I think we should remove this field and use task cancel only here.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/RollupShardIndexer.java
```diff
@@ -179,17 +204,23 @@ public void beforeBulk(long executionId, BulkRequest request) {
...
                                 BulkItemResponse::getId,
                                 BulkItemResponse::getFailureMessage,
                                 (msg1, msg2) -> Objects.equals(msg1, msg2) ? msg1 : msg1 + "," + msg2
                             )
                         );
-                    numFailed.addAndGet(failures.size());
                     logger.error("Shard [{}] failed to populate rollup index. Failures: [{}]", indexShard.shardId(), failures);
+
+                    // cancel rollup task
+                    abort = true;
```
martijnvg, line 223: I think we can just call `task.cancel(...)` here and maybe use the log message as reason.

martijnvg (commented):
Thanks @weizijun. I did another review round.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
weizijun: done

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/action/RollupShardTask.java
weizijun: done

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/action/RollupShardTask.java
weizijun: Yeah, I change the code.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
weizijun: done

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
weizijun: When do you want cancelOtherShardIndexers to be called? I change the code, when the shard failed, to avoid retry, I set the shardIt to the last shard.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/RollupShardIndexer.java
weizijun: because the `task.cancel()` is not a public method, so I add a new variable to do the abort logic internal.
The abort reason is from the bulk failed, so the RollupShardIndexer will throw exception. 
It's not from a user call to the cancel API.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/action/RollupShardTask.java
martijnvg: This method should return a new status instance:

```
new RollupShardStatus(shardId, numReceived, numSent, numIndexed, numFailed);
```

and then the status field in this class can be removed.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/action/RollupShardTask.java
martijnvg: These fields are still not final?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/RollupShardIndexer.java
martijnvg: Ah, yes. I forgot about that. Yes, then the `abort` field makes sense here.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
martijnvg: I think the `cancelTasksAction` field can be removed here and just use `client` here to invoke the cancel action.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
martijnvg: I also don't see this change. Maybe you forgot to push it?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
martijnvg: I also don't see this change. Maybe you forgot to push it?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/RollupShardIndexer.java
martijnvg: There is a reference to the `task` here. So I think the atomic long fields can be removed here and
instead the atomic longs from the `RollupShardTask` can be used directly in this class.

martijnvg (commented):
I did another review round. I think some changes were not pushed? I don't see some of the changes you said you've done.

weizijun:
> I did another review round. I think some changes were not pushed? I don't see some of the changes you said you've done.

Sorry, I lost some changes, now I have pushed them.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/action/RollupShardStatus.java
martijnvg: These fields are still `AtomicLong`, can you change the type of the fields field `long`?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/RollupShardIndexer.java
martijnvg: Can you also make this (^) change?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
martijnvg: > I change the code, when the shard failed, to avoid retry, I set the shardIt to the last shard.

👍

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
```diff
@@ -170,9 +177,12 @@ protected RollupIndexerAction.Response newResponse(
     private class Async extends AsyncBroadcastAction {
         private final RollupIndexerAction.Request request;
         private final ActionListener<RollupIndexerAction.Response> listener;
+        private final Task task;
+        private volatile boolean hasCancelled = false;
```
martijnvg, line 181: Maybe rename to `hasCancelledOtherShardOperations`?

weizijun: done

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
```diff
@@ -186,5 +196,42 @@ protected void finishHim() {
                 listener.onFailure(e);
             }
         }
+
```
martijnvg, line 199: After some more thinking, I'm not comfortable this particular change (cancelling other shard level downsamples incase a shard level downsample fails). I think in the current state it is fragile. I don't have a good suggestion how to make this logic more robust.

martijnvg (commented):
I left one smaller comment and a bigger comment around cancelling shard level downsamples in case a single shard level downsample fails. I think it is fragile, and I don't have a good idea how to make this more robust. Maybe leave this out of this PR? So that we can get the rest of the PR merged? If you like then the shard level downsample cancelling in case of failure can be explored/developed in another PR.

martijnvg:

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/action/RollupShardTask.java
```diff
@@ -0,0 +1,90 @@
...
+        return rollupIndex;
+    }
+
+    public DownsampleConfig config() {
+        return config;
+    }
+
+    @Override
+    public Status getStatus() {
+        return new RollupShardStatus(shardId, rollupStartTime, numReceived, numSent, numIndexed, numFailed);
```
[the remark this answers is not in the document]
weizijun: done

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/downsample/TransportRollupIndexerAction.java
weizijun: okay, I removed these code.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/rollup/action/RollupShardTask.java
```diff
@@ -0,0 +1,90 @@
...
+    public DownsampleConfig config() {
+        return config;
+    }
+
+    @Override
+    public Status getStatus() {
+        return new RollupShardStatus(shardId, rollupStartTime, numReceived.get(), numSent.get(), numIndexed.get(), numFailed.get());
+    }
+
+    public long getNumReceived() {
```
martijnvg, line 59: Maybe just have getter methods for the `AtomicLong` fields?

martijnvg (approved):
Thanks @weizijun! I left one minor comment, but other than that this looks good to me.

martijnvg:
@elasticmachine update branch

martijnvg:

martijnvg:

PR #88498, opened 2022-07-13
[ML] Add exception detail to auditor failure to install template

droberts195:
While looking at a different issue I noticed cases of the
ML auditor failing to install its template. Unfortunately
this log message didn't include the exception, so it was
not possible to say for sure why this happened. This PR
adds the exception.


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/common/notifications/AbstractAuditor.java
```diff
@@ -158,7 +159,7 @@ protected void indexDoc(ToXContent toXContent) {
                     }
                     backlog.add(toXContent);
                 } else {
-                    logger.error("Latest audit template missing but the back log has been written");
+                    logger.error("Latest audit template missing and audit message cannot be added to the backlog");
```
droberts195, line 162: This should never happen, due to the assert on line 155, but since I'm changing the file I thought I'd correct the message anyway.

benwtrent (approved):

PR #88499, opened 2022-07-13
[DOCS] Update API key API

n1v0lg:
API docs for the new `PUT /_security/api_key/<id>` API.

Link to [doc preview](https://elasticsearch_88499.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/security-api-update-api-key.html).


x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,270 @@
...
+[[security-api-update-api-key-desc]]
+==== {api-description-title}
+
+The API keys created by <<security-api-create-api-key,create API Key>> or <<security-api-grant-api-key,grant API Key>> can be updated using this API.
+It's not possible to update expired API keys, or API keys that have been invalidated by <<security-api-invalidate-api-key,invalidate API Key>>.
+
+This API supports updates to an API key's access scope, i.e., its permissions, and its metadata.
+Other properties such as `name` or `expiration` cannot be updated.
+
+IMPORTANT: An API key's permissions are always restricted by the permissions of the authenticated user creating or updating the API key.
```
n1v0lg, line 34: This callout is likely confusing. Would love your take on it, and more than happy to go over it in detail on Slack or Zoom or here on GH, whichever you prefer.

lockewritesdocs: Perhaps we can focus less on restriction and more on inheritance. For example:

> The API key inherits the permissions of the authenticated user that creates or updates the API key.

Provided that this ^ statement is technically valid.

n1v0lg: I like inherit but it loses the following nuance: the API key's permissions are the intersection of the roles a user assigns to it and the user's permissions. "Inherit" seems to imply that the API key simply has the same permissions as the user it inherits from. What we should get across is the API key's permissions are "at most" those of the user.

"Limited by" is another phrase that came to mind but it clashes a bit with a statement further down:

> If you supply role descriptors, the API key's permissions will be the intersection of the supplied permissions and the authenticated user's current permissions, **thereby limiting the API key's access scope**.

I want to reserve using "limiting" to talking about the "assigned" role descriptors limiting the API key's permissions, and so went with "reserved" for the inverse case.

n1v0lg:
@elasticmachine run docs build

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
n1v0lg: I like the idea of a broader statement here. Essentially what I want to get across is that any call to this API may change the access scope of the API key even if an end-user does not specify any `role_descriptors`. I'll think about how I can keep make the statement broader and possibly move some of the context into the `role_descriptors` sub-section.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,273 @@
...
+  "metadata": {
+    "environment": {
+       "level": 2,
+       "trusted": true,
+       "tags": ["production"]
+    }
+  }
+}
+----
+// TEST[s/VuaCfGcBCdbkQm-e5aOx/\${body.id}/]
```
n1v0lg, line 161: Turns out I was missing a `\` in my attempts to get this to work. Thanks @ywangd for figuring this out 🚀

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,270 @@
...
+The following example updates the API key, replacing the API key's assigned permissions with the <<security-api-update-api-key-examples-user-permissions, authenticated user's>>:
+
+[source,console]
+----
+PUT /_security/api_key/VuaCfGcBCdbkQm-e5aOx
+{
+  "role_descriptors": {}
+}
+----
+// TEST[skip:api key id not available anymore]
```
n1v0lg, line 196: We might be able to work around this, but settling for this for now. Might revisit in a follow up PR.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,270 @@
...
+IMPORTANT: It's not possible to use an API key as the authentication credential for this API.
+To update an API key, user password or <<security-api-get-token, token credentials>> are required.
+
+[[security-api-update-api-key-desc]]
+==== {api-description-title}
+
+The API keys created by <<security-api-create-api-key,create API Key>> or <<security-api-grant-api-key,grant API Key>> can be updated using this API.
+It's not possible to update expired API keys, or API keys that have been invalidated by <<security-api-invalidate-api-key,invalidate API Key>>.
+
+This API supports updates to an API key's access scope, i.e., its permissions, and its metadata.
```
ywangd, line 31: We should also explicitly list the auto-synced properties, especially the owner user's privileges. This makes the following callout more readable.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,270 @@
...
+
+IMPORTANT: A call to <<security-api-update-api-key,update API Key>> may result in a change in the API key's access scope, even if you don't specify new <<security-api-update-api-key-api-key-role-descriptors,`role_descriptors`>> in the request.
+This is because the API key's access scope is <<security-api-update-api-key-api-key-role-descriptors,derived and updated>> based on a snapshot of the authenticated user's permissions at the time of the request.
```
ywangd, line 34: I understand `authenticated user` is what we use in the create-api-key page. But I don't think we should stick to it. The "owner user" is what we currently settled with and its usage should be promoted.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,270 @@
...
+==== {api-request-body-title}
+
+The request body is optional.
+Following parameters can be specified in it:
+
+[[security-api-update-api-key-api-key-role-descriptors]]
+`role_descriptors`::
+(Optional, array-of-role-descriptor) An array of new role descriptors for this API key.
+This parameter is optional.
+When the parameter is supplied but empty, the API key's new permissions will be the current _point in time snapshot of permissions of the authenticated user_.
```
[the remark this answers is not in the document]
ywangd: If we separate the two following two discussions, it might help make things easier to explain:
1. The API key's effective permission is *always* the intersection of its own privileges and owner user's snapshotted privileges
2. How the privileges can be updated, i.e. a) Always (auto) updating owner user's privileges b) optionally update the key's own privileges. 

Item 2 is what we really care here. But we have to explain 1 because currently we have no overview doc for API keys. We can keep 1 simple to just state the rule is always intersection (no matter how its updated or captured). 

For 2, the tricky bit is the `empty` `role_descriptors` means inherting owner user's privileges. Otherwise, it is not that complicated. So I think we can state the simple rule (optional update) first and leave the special `empty` input as a callout. 

This should let us avoid mentioned things like "previously assigned ..." and not trying to list all combinations.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,270 @@
...
+If you supply role descriptors, the API key's permissions will be the intersection of the supplied permissions and the authenticated user's current permissions, thereby limiting the API key's access scope.
+If you do not supply this parameter, the permissions will be the user's current permissions, limited by permissions previously assigned to the API key, if any.
+The structure of the role descriptor is the same as the request for the create role API.
+For more details, see <<security-api-put-role, create or update roles API>>.
+
+`metadata`::
+(Optional, object) Arbitrary metadata that you want to associate with the API key.
+It supports nested data structure.
+Within the `metadata` object, keys beginning with `_` are reserved for system usage.
```
ywangd, line 61:
```suggestion
Within the `metadata` object, top level keys beginning with `_` are reserved for system usage.
```

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,270 @@
...
+When specified, this fully replaces metadata previously associated with the API key.
+
+[[security-api-update-api-key-response-body]]
+==== {api-response-body-title}
+
+`updated`::
+(boolean) `true` if the API key was updated, `false` if the API key did not change.
```
ywangd, line 68:
```suggestion
(boolean) `true` if the API key was updated, `false` if the API key did not change (because no change is needed).
```

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,270 @@
...
+A successful call returns a JSON structure indicating that the API key was updated:
+
+[source,console-result]
+----
+{
+  "updated": true
+}
+----
+
+The API key's effective permissions after the update will be the intersection of the supplied role descriptors and the <<security-api-update-api-key-examples-user-permissions, authenticated user's permissions>>:
```
ywangd, line 172: This and below are largely trying to explain how API key's permission is an intersection. Given how dense they are and its importance, we might want just add a basic version of intro page for API keys and have this explain over there. The API itself is not a great place for explain this concept. WDYT? Also @lockewritesdocs

ywangd (commented):
I have a few comments

lockewritesdocs (commented):
I left a few suggested changes, plus some comments and questions. Happy to re-review after you've addressed feedback!

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
ywangd: My idea was to have a "basic" overview doc page, not "explaining in greater detail". That said, it's probably hard to come up with a "basic" yet sufficient overview. I also agree it's outside of the scope for this PR.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,270 @@
...
+
+`PUT /_security/api_key/<id>`
+
+[[security-api-update-api-key-prereqs]]
+==== {api-prereq-title}
+
+* To use this API, you must have at least the `manage_own_api_key` cluster privilege.
+Users can only update their own API keys, i.e., API keys they created.
+To update another user's API key, you must use the run as feature.
+For more information, see <<run-as-privilege>>.
```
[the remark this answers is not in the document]
n1v0lg: Flows better, thanks!

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,270 @@
...
+==== {api-path-parms-title}
+
+`id`::
+(Required, string) The ID of the API key to be updated.
+
+[[security-api-update-api-key-request-body]]
+==== {api-request-body-title}
+
+The request body is optional.
+Following parameters can be specified in it:
```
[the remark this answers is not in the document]
n1v0lg: Nit but I think it's worth highlighting that the request body is optional altogether.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
n1v0lg: I considered that as well, and am up for including them.

My reservation is that updating version and creator info is, in a sense, is an implementation detail -- these fields are not exposed via a public API and should be largely opaque to users.

Likewise, "updates to an API key's access scope" _includes_ the auto update to the user perm snapshot.

That said, I think including details about the auto-synced user perms makes a lot of sense. To avoid redundancy, I've tweak the text here as well as the callout. LMKWYT

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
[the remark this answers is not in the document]
n1v0lg: I've addressed this in combination with @ywangd's [comment](https://github.com/elastic/elasticsearch/pull/88499#discussion_r922947063).

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,276 @@
...
+
+This API supports updates to an API key's access scope, i.e., its permissions, and its metadata.
+The access scope of an API key is derived from the <<security-api-update-api-key-api-key-role-descriptors,`role_descriptors`>> you specify in the request, and a snapshot of the owner user's permissions at the time of the request.
+The snapshot of the owner's permissions is updated automatically on every call.
+
+[IMPORTANT]
+====
+A call to this API might change the API key's access scope, even if you don't specify new
+<<security-api-update-api-key-api-key-role-descriptors,`role_descriptors`>> in the request.
+This behavior can occur if the owner user's permissions have changed since the creation or last update to the API key.
```
n1v0lg, line 39: Instead of "creation or last update" could also go with "last modification of the API key".

The preceding paragraph now explains that the snapshot is auto-updated so I've removed the now redundant explanation from the callout.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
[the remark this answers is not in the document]
n1v0lg: Absolutely up for simplifying. I provided the "verbose" full context here so that we have the complete picture for discussion and it's easier to talk about what should and shouldn't be included.

> The API key's permissions are a combination of it's privileges, plus the owner user's current permissions at the current time when the request is issued.

Is still a little confusing since we talk about the API key's permissions and the API key's privileges which are only very subtly different (if at all). I think using "assigned permissions" or "assigned privileges" works better here.

> The API key's effective permission is always the intersection of its own privileges and owner user's snapshotted privileges

Is not exactly right since if the assigned privileges are empty, it's not the intersection anymore (it would be empty) so we still need some sort of qualifier here.

What about s.th. like:

> The API key's permissions are an intersection of its assigned privileges, plus the owner user's current permissions at the time the request is issued. If no privileges are assigned, the API key has the owner user's full current permissions.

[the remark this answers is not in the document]
n1v0lg: > However, I think that we need an example to clarify and solidify this statement.

What if we link to the example requests further down? I've intentionally included examples that highlight the nuances of this.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,270 @@
...
+[[security-api-update-api-key-request-body]]
+==== {api-request-body-title}
+
+The request body is optional.
+Following parameters can be specified in it:
+
+[[security-api-update-api-key-api-key-role-descriptors]]
+`role_descriptors`::
+(Optional, array-of-role-descriptor) An array of new role descriptors for this API key.
+This parameter is optional.
```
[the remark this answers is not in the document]
n1v0lg: I dropped the `array-of-role-descriptor` bit and the `This parameter is optional.` bit (this were copy and paste from the create API key API). I've tweaked the remaining text slightly to better align with my most recent edits.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,277 @@
...
+To update an API key, the owner user's credentials are required.
+
+[[security-api-update-api-key-desc]]
+==== {api-description-title}
+
+Use this API to update API keys created by the <<security-api-create-api-key,create API Key>> or <<security-api-grant-api-key,grant API Key>> APIs.
+It's not possible to update expired API keys, or API keys that have been invalidated by <<security-api-invalidate-api-key,invalidate API Key>>.
+
+This API supports updates to an API key's access scope, i.e., its permissions, and its metadata.
+The access scope of an API key is derived from the <<security-api-update-api-key-api-key-role-descriptors,`role_descriptors`>> you specify in the request, and a snapshot of the owner user's permissions at the time of the request.
```
ywangd, line 32: We can be more explicit here. Instead of saying "is derived", we can just say "is an intersection of ...".

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,277 @@
...
+The snapshot of the owner's permissions is updated automatically on every call.
+
+[IMPORTANT]
+====
+A call to this API might change the API key's access scope, even if you don't specify new
+<<security-api-update-api-key-api-key-role-descriptors,`role_descriptors`>> in the request.
+This behavior can occur if the owner user's permissions have changed since the last modification of the API key.
```
ywangd, line 39: My attempt to cover both creation and update:
```suggestion
It can occur if the owner user's permissions have changed and are different from the snapshot version stored as part of the API key.
```

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,277 @@
...
+
+[[security-api-update-api-key-request-body]]
+==== {api-request-body-title}
+(Optional)
+
+You can specify the following optional parameters in the request body.
+
+[[security-api-update-api-key-api-key-role-descriptors]]
+`role_descriptors`::
+(Optional, array) An array of role descriptors to assign to this API key.
```
ywangd, line 56: This is technically incorrect because this parameter does not accept array, but must be an object. Create API Key page has the same issue.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,277 @@
...
+}
+--------------------------------------------------
+// NOTCONSOLE
+
+For the next example, assume that the owner user's permissions have changed from <<security-api-update-api-key-examples-user-permissions, the original permissions>> to:
+
+[source,js]
+--------------------------------------------------
+{
+  "cluster": ["manage_own_api_key"],
```
ywangd, line 235: Nit:
Can we use `manage_security` instead of `manage_own_api_key`? I am worried that too many `api_key` is confusing at times.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,277 @@
...
+Which returns the response:
+
+[source,console-result]
+----
+{
+  "updated": true
+}
+----
+
+Resulting in the following permissions for the API key:
```
ywangd, line 263:
```suggestion
Resulting in the following effective permissions for the API key:
```

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,277 @@
...
+
+[source,js]
+--------------------------------------------------
+{
+  "cluster": ["manage_own_api_key"],
```
ywangd, line 268: Same here: can we replace it with `manage_security`?

ywangd (approved):
LGTM

The content looks correct to me. I'll defer to @lockewritesdocs for polishing the writing. Thanks!

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
n1v0lg: I don't think "intersection of" is technically correct here without the disclaimer that when `role_descriptors` is empty (or not assigned) it's not the intersection. I went for "derived" because it side-steps that problem, and references the role desc section where we accurately describe how the permissions are derived.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
n1v0lg: What about "have changed since the creation or last modification of the API key"? It keeps things more high level, without going into detail about the snapshot.

Alternatively:

"can occur if the snapshot of the owner permissions on the API key is stale, i.e., the owner permissions have since diverged (changed?)."

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
n1v0lg: Indeed, forgot to comment on this. Thanks for flagging! I can change this to `(Optional, object) The role descriptors to assigned to this API key` and likewise tweak the create and grant docs.

n1v0lg:
@elasticmachine update branch plz

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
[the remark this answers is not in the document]
lockewritesdocs: I don't think that we need the additional (Optional) tag. What if we changed this to read:

>You can specify the following parameters in the request body, which is optional.

Each of the parameters have an (Optional) tag already, so I think that this statement highlights that the entire request body is optional without explicitly including an additional tag.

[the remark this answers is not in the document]
lockewritesdocs: I don't think that we need the additional (Optional) tag. What if we changed this to read:

>You can specify the following parameters in the request body, which is optional.

Each of the parameters have an (Optional) tag already, so I think that this statement highlights that the entire request body is optional without explicitly including an additional tag.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
lockewritesdocs: +1 for including this information for now so that we have it somewhere. We can move it in a subsequent PR for creating an overview of API keys. Thanks @n1v0lg!

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
lockewritesdocs: What if we flipped this statement to move the conditional part about `role_descriptors` to the end? I think that moving the explanation of why this can occur helps to clarify the initial sentence:

>A call to this API might change the API key's access scope, which can occur if the owner user's
>permissions have changed since the API key was created or last modified. This change can occur
>even if you don't specify new <<security-api-update-api-key-api-key-role-descriptors,`role_descriptors`>> in the request.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
```diff
@@ -1,5 +1,277 @@
...
+IMPORTANT: It's not possible to use an API key as the authentication credential for this API.
+To update an API key, the owner user's credentials are required.
+
+[[security-api-update-api-key-desc]]
+==== {api-description-title}
+
+Use this API to update API keys created by the <<security-api-create-api-key,create API Key>> or <<security-api-grant-api-key,grant API Key>> APIs.
+It's not possible to update expired API keys, or API keys that have been invalidated by <<security-api-invalidate-api-key,invalidate API Key>>.
+
+This API supports updates to an API key's access scope, i.e., its permissions, and its metadata.
```
lockewritesdocs, line 31:
```suggestion
This API supports updates to an API key's access scope, such as its permissions and metadata.
```

lockewritesdocs (approved):
Left a few suggested non-blocking changes. Thanks for iterating on this one @n1v0lg!

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
n1v0lg: Sorry for the confusing wording/punctuation on my part. I meant this to read as:

> This API supports updates to: 1. an API key's access scope, i.e., its permissions; and 2. its metadata.

I don't mind ditching the i.e. part and making this:

> This API supports updates to an API key's access scope and metadata.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
n1v0lg: I like it. Thanks!

n1v0lg: Discussed on slack, and going with a different tweak which keeps the callout specific to the "access change despite omitted role descriptors" issue.

x-pack/docs/en/rest-api/security/update-api-key.asciidoc
lockewritesdocs: +1 to that change

PR #50277, opened 2019-12-17
Make the Docker build more re-usable in Cloud

pugnascotia:
Closes #49926. Closes #46166.

   * Add a mini-init process
   * Don't use root at all when running the container
   * Ensure no files in the image have the setuid flag

Also improve dependency tracking in the build.

PR #88502, opened 2022-07-13
Fix passing positional args to ES in Docker

pugnascotia:
As part of #50277, we removed the `TAKE_FILE_OWNERSHIP` option from the
Docker entrypoint script in and the associated chroot calls, and instead
just defaulted to running the image as `elasticsearch` instead of
`root`.

However, we didn't check that it was still possible to pass CLI options
to Elasticsearch via CLI arguments, and broke this by mistake.  The
impact is apparently minor, since we only just noticed the problem, and
supplying CLI args in this way is probably an uncommon pattern, versus
environment variables or a config file. Nevertheless, it is supposed to
be possible and is mentioned in the documentation.

Fix the problem by suppling the missing positional params when calling
Elasticsearch, and add a test case so that we don't break it again.

This doesn't impact the `7.x` series.

### DIFF
--- a/distribution/docker/src/docker/bin/docker-entrypoint.sh
+++ b/distribution/docker/src/docker/bin/docker-entrypoint.sh
@@ -81,4 +81,4 @@ fi
 
 # Signal forwarding and child reaping is handled by `tini`, which is the
 # actual entrypoint of the container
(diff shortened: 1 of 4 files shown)


mark-vieira (approved):
Good catch. LGTM.

pugnascotia:
Backported to `8.3` in 3776923e458.

PR #88504, opened 2022-07-13
Disallow explicitly setting search_type with knn search

jtibshirani:
When kNN search is enabled, it automatically uses the `dfs_query_then_fetch`
search type. The user can no longer control the search type that's used. This
commit throws an error if `search_type` is explicitly set when `knn` search is
used.

Addresses #87625.

### DIFF
--- a/server/src/main/java/org/elasticsearch/rest/action/search/RestSearchAction.java
+++ b/server/src/main/java/org/elasticsearch/rest/action/search/RestSearchAction.java
@@ -201,7 +201,6 @@ public class RestSearchAction extends BaseRestHandler {
         searchRequest.indicesOptions(IndicesOptions.fromRequest(request, searchRequest.indicesOptions()));
 
         validateSearchRequest(request, searchRequest);
-        checkSearchType(request, searchRequest);
 
         if (searchRequest.pointInTimeBuilder() != null) {
             preparePointInTime(searchRequest, request, namedWriteableRegistry);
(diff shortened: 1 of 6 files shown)


jtibshirani:
I initially tried a strategy where we check this inside `SearchRequest#validate`. But it is not possible to tell whether the search type was explicitly set by the user, or just set to the default value. We could update `SearchRequest` to use `Explicit<SearchType>`, but this was intrusive -- there are questions over wire serialization, and we need multiple different setters for setting "explicitly" (for user requests) vs. "implicitly" (for internal optimizations).

server/src/main/java/org/elasticsearch/rest/action/search/RestSearchAction.java
```diff
@@ -200,7 +200,8 @@ public static void parseSearchRequest(
         searchRequest.preference(request.param("preference"));
         searchRequest.indicesOptions(IndicesOptions.fromRequest(request, searchRequest.indicesOptions()));
 
-        checkRestTotalHits(request, searchRequest);
+        validateSearchRequest(request, searchRequest);
```
mayya-sharipova, line 203: If `validateSearchRequest` already includes `checkSearchType` why do we need to call it again after `validateSearchRequest`?

mayya-sharipova (approved):
@jtibshirani  Thanks, this overall LGTM, except the removal of double checking in `RestSearchAction`.  After fixing it, this PR can be merged without further review.

server/src/main/java/org/elasticsearch/rest/action/search/RestSearchAction.java
jtibshirani: Thanks @mayya-sharipova , this was just an oversight. Will fix.

PR #88505, opened 2022-07-13
Trivial code cleanup

craigtaverner:
This tiny cleanup was done as part of https://github.com/elastic/elasticsearch/pull/88487, but is clearly not related to that PR. It should have no behavioural or performance implications and is simply a very small cleanup to a method that does a kind of `String.join` function.


not-napoleon (approved):
LGTM

PR #88506, opened 2022-07-13
Avoid needless index metadata builders during reroute

original-brownbear:
This set of changes makes `org.elasticsearch.cluster.routing.allocation.IndexMetadataUpdater#applyChanges` essentially free even in clusters of O(100k) indices compared to using a disproportionately increasing amount of CPU as the cluster grows (about 1% of CPU time while bootstrapping many shards at 25k indices benchmarks and increasing from there).
It also appears to have additional benefits end-to-end in those benchmarks, likely as a result of making diffing metadata cheaper by retaining more instance equality across the board.

relates #77466


server/src/main/java/org/elasticsearch/common/util/Maps.java
```diff
@@ -64,13 +58,25 @@ public static <K, V> Map<K, V> copyMapWithAddedEntry(final Map<K, V> map, final
      * @param <V>   the type of the values in the map
      * @return an immutable map that contains the items from the specified map and a mapping from the specified key to the specified value
      */
+    @SuppressWarnings("unchecked")
     public static <K, V> Map<K, V> copyMapWithAddedOrReplacedEntry(final Map<K, V> map, final K key, final V value) {
```
original-brownbear, line 62: Changes here aren't 100% necessary for this to work, but they cause the whole thing to compile far better.
Without this change there's not really any reason to improve this code though so I put it in here.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetadata.java
```diff
@@ -654,7 +655,8 @@ private IndexMetadata(
         this.autoExpandReplicas = autoExpandReplicas;
         this.isSearchableSnapshot = isSearchableSnapshot;
         this.isPartialSearchableSnapshot = isPartialSearchableSnapshot;
-        this.indexCompatibilityVersion = SETTING_INDEX_VERSION_COMPATIBILITY.get(settings);
+        this.indexCompatibilityVersion = indexCompatibilityVersion;
+        assert indexCompatibilityVersion.equals(SETTING_INDEX_VERSION_COMPATIBILITY.get(settings));
```
original-brownbear, line 659: This snuck in recently making the constructor here needlessly heavy, fixed here so the copy-constructors are as fast as possible again.

server/src/main/java/org/elasticsearch/common/util/Maps.java
```diff
@@ -41,9 +38,6 @@ public class Maps {
      */
     @SuppressWarnings("unchecked")
     public static <K, V> Map<K, V> copyMapWithAddedEntry(final Map<K, V> map, final K key, final V value) {
-        Objects.requireNonNull(map);
```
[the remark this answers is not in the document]
idegtiarenko: May be keep them ass `assert` so that it is clear for the reader?

server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
```diff
@@ -397,6 +398,47 @@ public Metadata withCoordinationMetadata(CoordinationMetadata coordinationMetada
...
     }
 
+    /**
+     * Creates a copy of this instance updated with the given {@link IndexMetadata} that must only contain changes to primary terms
+     * and in-sync allocation ids relative to the existing entries. This method is only used by
+     * {@link org.elasticsearch.cluster.routing.allocation.IndexMetadataUpdater#applyChanges(Metadata, RoutingTable)}.
+     * @param updates map of index name to {@link IndexMetadata}.
+     * @return updated metadata instance
+     */
+    public Metadata withAllocationAndTermUpdatesOnly(Map<String, IndexMetadata> updates) {
```
idegtiarenko, line 408: This sounds very specific. Would it make sense to have it as a `private static Metadata withAllocationAndTermUpdatesOnly(Metadata original, Map<String, IndexMetadata> updates)` in `IndexMetadataUpdater` so nobody misuses it somewhere else?

idegtiarenko (approved):

server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
original-brownbear: I agree, it's a little messy. The problem is, we don't really get anything out of moving to this kind of private method because then we'd have to expose the raw constructor of `Metadata` to do the work elsewhere wouldn't we?

server/src/main/java/org/elasticsearch/common/util/Maps.java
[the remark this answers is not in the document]
original-brownbear: ++ Sure thing added :)

PR #88514, opened 2022-07-13
Updatable API keys - auto-update legacy superuser RDs

n1v0lg:
API keys created in `7.x` may have legacy `superuser` user role
descriptors. In `8.x` this is handled by translating these to `8.x`
`superuser` role descriptors when they are read. Instead, we can
automatically update them (once) when an API key is first updated. This
PR tweaks our noop detection logic to enable this.

Labeling `non-issue` since this is an implementation detail of a
not-yet released feature.


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -705,6 +701,14 @@ public List<RoleDescriptor> parseRoleDescriptorsBytes(
...
         BytesReference bytesReference,
         RoleReference.ApiKeyRoleType roleType
+    ) {
+        return parseRoleDescriptorsBytes(apiKeyId, bytesReference, roleType == RoleReference.ApiKeyRoleType.LIMITED_BY);
+    }
+
+    private List<RoleDescriptor> parseRoleDescriptorsBytes(
+        final String apiKeyId,
+        BytesReference bytesReference,
+        final boolean replaceLegacySuperuserRoleDescriptor
```
ywangd, line 711: This was actually the initial version that I had for this method. It got changed to using an Enum for better semantics. Since this is now a private method, I think it is OK.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -569,17 +570,12 @@ private boolean isNoop(
...
-        //
-        // This does *not* present a functional issue, since whenever a `LEGACY_SUPERUSER_ROLE_DESCRIPTOR` is loaded at authentication time,
-        // it is likewise automatically transformed to `ReservedRolesStore.SUPERUSER_ROLE_DESCRIPTOR`.
         final List<RoleDescriptor> currentLimitedByRoleDescriptors = parseRoleDescriptorsBytes(
             request.getId(),
             apiKeyDoc.limitedByRoleDescriptorsBytes,
-            RoleReference.ApiKeyRoleType.LIMITED_BY
+            // We want the 7.x `LEGACY_SUPERUSER_ROLE_DESCRIPTOR` role descriptor to be returned here to auto-update
+            // `LEGACY_SUPERUSER_ROLE_DESCRIPTOR` to `ReservedRolesStore.SUPERUSER_ROLE_DESCRIPTOR`, when we update a 7.x API key.
+            false
```
ywangd, line 578: I think we should also use this version of the method in the above (Line 563) where we parse the key's descriptors. It is better consistency. In this context, we never want to any sorta of transform. So rather relying on `ApiKeyRoleType`, we should just directly say `false` to the transform.

ywangd (approved):
LGTM

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
n1v0lg: Agreed that the flag is suitable for a private method whereas enum is better for public API 👍

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -560,26 +561,21 @@ private boolean isNoop(
             final List<RoleDescriptor> currentRoleDescriptors = parseRoleDescriptorsBytes(
                 request.getId(),
                 apiKeyDoc.roleDescriptorsBytes,
-                RoleReference.ApiKeyRoleType.ASSIGNED
+                false
             );
             if (false == (newRoleDescriptors.size() == currentRoleDescriptors.size()
-                && Set.copyOf(newRoleDescriptors).containsAll(new HashSet<>(currentRoleDescriptors)))) {
```
n1v0lg, line 566: The `HashSet` bit here is actually redundant: we check that the list sizes are equal and that one collection (as a set) contains the other.

ywangd: Indeed. My suggestion on the previous PR did not have it.

ywangd: https://github.com/elastic/elasticsearch/pull/88346#discussion_r918543380

n1v0lg:

PR #88515, opened 2022-07-13
[DOCS] disable the ILM history store on full cluster restore

andreidan:
When restoring a full cluster there might be issues with the snapshot containing the
ilm-history data stream and, even though deleted from the elasticsearch cluster and ILM stopped, the history
data stream reappearing due to late processing of ILM history events.

This documents disabling the history index so that the data stream is not recreated and restoring
the snapshot doesn't fail due to the data stream already being present in the cluster.


dakrone (changes requested):
looks good, but on line 435 we need to add it to the list of settings to re-enable after the restore has completed

andreidan:
> looks good, but on line 435 we need to add it to the list of settings to re-enable after the restore has completed

Crikey! Good spot. I missed that even though I ctrl-f-ed for "start ILM" :)

dakrone (approved):
LGTM, thanks!

andreidan:
@elasticmachine update branch

PR #88522, opened 2022-07-13
Setup periodic snyk monitoring per branch

breskeby:
Adds a new ci Jenkins job configuration for running snyk dependency monitoring on a daily basis. We setup a service account in snyk and resolve the api token for publishing in vault. 

Related to #87620

### DIFF
--- a/.ci/jobs.t/elastic+elasticsearch+periodic+snyk-dependency-monitoring.yml
+++ b/.ci/jobs.t/elastic+elasticsearch+periodic+snyk-dependency-monitoring.yml
@@ -1,0 +1,23 @@
+---
+- job:
+    name: elastic+elasticsearch+%BRANCH%+snyk-dependency-monitoring
+    workspace: /dev/shm/elastic+elasticsearch+%BRANCH%+snyk-dependency-monitoring
+    display-name: "elastic / elasticsearch # %BRANCH% - snyk dependency monitoring"
+    description: "Publishing of the Elasticsearch %BRANCH% dependencies graph to snyk dependency monitoring"
+    builders:
+      - inject:
+          properties-file: '.ci/java-versions.properties'
+          properties-content: |
+            JAVA_HOME=$HOME/.java/$ES_BUILD_JAVA
+            RUNTIME_JAVA_HOME=$HOME/.java/$ES_RUNTIME_JAVA
+      - shell: |
+          #!/usr/local/bin/runbld --redirect-stderr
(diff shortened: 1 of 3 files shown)


.ci/jobs.t/elastic+elasticsearch+periodic+snyk-dependency-monitoring.yml
```diff
@@ -0,0 +1,23 @@
...
+            JAVA_HOME=$HOME/.java/$ES_BUILD_JAVA
+            RUNTIME_JAVA_HOME=$HOME/.java/$ES_RUNTIME_JAVA
+      - shell: |
+          #!/usr/local/bin/runbld --redirect-stderr
+          set -euo pipefail
+          set +x
+          VAULT_TOKEN=$(vault write -field=token auth/approle/login role_id=$VAULT_ROLE_ID secret_id=$VAULT_SECRET_ID)
+          export VAULT_TOKEN
+          export data=$(vault read -format=json secret/elasticsearch-ci/snyk)
+          export snyk_token=$(echo $data | jq -r .data.token)
```
mark-vieira, line 20: We can simplify this to `snyk_token=$(vault read -field=token secret/elasticsearch-ci/snyk)`.

.ci/jobs.t/elastic+elasticsearch+periodic+snyk-dependency-monitoring.yml
```diff
@@ -0,0 +1,23 @@
...
+          unset VAULT_TOKEN data
+          set -x
+          $WORKSPACE/.ci/scripts/run-gradle.sh uploadSnykDependencyGraph -PsnykToken=$snyk_token
```
mark-vieira, line 23: We can't do `set -x` and then include the token in the command line as this is going to leak the secret. We'll either need to tweak the build to pull this from an env var, or omit the `set -x` bit so we don't print this command. I'm inlined to do the former, as it's probably less likely that we'll make a mistake and accidentally leak the token in console output when using environment variables as we do for most other secrets.

mark-vieira (changes requested):

.ci/jobs.t/elastic+elasticsearch+periodic+snyk-dependency-monitoring.yml
breskeby: neat. done

mark-vieira (approved):
LGTM

Issue #88341, opened 2022-07-07
Improve how builtin resources are being installed


martijnvg:
Today our builtin resources (component/composable index/legacy templates and ilm policies) that ship by default with Elasticsearch get installed and updated via `IndexTemplateRegistry` infrastructure. This installs or updates the builtin resources using the put transport actions that exist for templates and ilm policies by checking the cluster state on each update.

Reusing these put transport actions that are also used by the rest APIs have downsides.
- There is nothing preventing admin users from removing or updating the builtin resources. When this happens, the `IndexTemplateRegistry` detects this and reverts the change, but for a short moment, these builtin resources are removed or in an unexpected state.
- An info log line is printed for each template/policy being installed/updates, which is very verbose at cluster startup. This is because printing info log is hardcoded in the put transport actions.
- Each time a template/policy is installed or updated this then results in a separate cluster state update. This slows down initializing Elasticsearch quite a bit. Ideally installing/updating builtin resources can be performed in a single cluster state update.

With the development of the operator and plugin read only settings project, infrastructure will be added that we can reuse to install and update our builtin resources. This new infrastructure will allow builtin resources to be protected from modification via API. Any attempt to modify the builtin resources via API will be forbidden. Also the other mentioned downside will be addressed by reusing the new infrastructure for operator and plugin read only settings project. 

At same time we should also formalise the notion of a package, which are templates and ilm policies (and other config resources) that have a dependency on each other and belong together. Part of the notion of a package will be defined in the operator and plugin read only settings project. The builtin resources that we ship by default should be defined as a package as well.

We will be needing a way for modules and plugins to be able to define the builtin resources that need to be installed/updated. This was previously done with the `IndexTemplateRegistry` infrastructure. This can be something lightweight that builds on top of the infrastructure built for the operator and plugin read only settings project. For example an interface (`BuiltinPackage` or `PackageProvider`) that plugins and modules can implement that is loaded via SPI.

### Tasks

- [ ] Build infrastructure that allows for modules and plugins to define legacy, component, composable index templates and ilm policies. An interface should be added that allows modules/plugins to define a package. A component should added that loads all instances of this interface defined in modules/plugins and manages the installation of all the defined resources.
- [ ] Implement operator handlers for legacy, component and composable index templates. Currently the operator and plugin read only settings project infrastructure only has support for defining cluster settings and ilm templates as part of packages.
- [ ] Migrate stack module away from IndexTemplateRegistry
[...]

PR #88527, opened 2022-07-13
Reserved cluster state service

grcevski:
This PR adds the reserved cluster state controller, which part of the File based settings PR (https://github.com/elastic/elasticsearch/pull/86224).

This is a reusable component that has number of Reserved State Handler components that know how to update parts of the Elasticsearch state. This component is only used by the File Settings Service at the moment, but the intent is for modules and plugins to be able to call it to save settings and entities in the cluster state that cannot be changed by the end user. For example work on using this component to update state from plugins/modules, please see https://github.com/elastic/elasticsearch/issues/88341


server/src/main/java/org/elasticsearch/ExceptionsHelper.java
```diff
@@ -262,6 +263,26 @@ public static void maybeDieOnAnotherThread(final Throwable throwable) {
         });
     }
 
+    /**
+     * Unwraps a throwable and calls an error listener with every unwrapped throwable
+     * @param t the starting throwable
+     * @param errorListener a listener function that will be called with every unwrapped throwable we find
+     * @param limit after how many encountered throwables should we stop unwrapping. Prevents stack overflows. 10 is reasonable max.
+     */
+    public static void unwrap(Throwable t, Consumer<Throwable> errorListener, int limit) {
```
grcevski, line 272: I added this helper because I couldn't find a suitable one here that would return each throwable message. When we parse JSON, the messages of errors are reported on each throwable as it goes down in subsections of the JSON body.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedClusterStateController.java
```diff
@@ -0,0 +1,232 @@
...
+import static org.elasticsearch.core.Strings.format;
+
+/**
+ * Controller class for storing and reserving a portion of the {@link ClusterState}
+ * <p>
+ * This class contains the logic about validation, ordering and applying of
+ * the cluster state specified in a file or through plugins/modules. Reserved cluster state
+ * cannot be modified throught the REST APIs, only through this controller class.
+ */
+public class ReservedClusterStateController {
```
rjernst, line 45: nit: we've typically used the "Service" suffix for other similar "things" in a node, any reason not to use that here in place of Controller?

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedStateUpdateTask.java
```diff
@@ -0,0 +1,138 @@
...
+
+        var reservedMetadataBuilder = new ReservedStateMetadata.Builder(namespace).version(reservedStateVersion.version());
+        List<String> errors = new ArrayList<>();
+
+        for (var handlerName : orderedHandlers) {
+            ReservedClusterStateHandler<?> handler = handlers.get(handlerName);
+            try {
+                Set<String> existingKeys = keysForHandler(existingMetadata, handlerName);
+                TransformState transformState = handler.transform(reservedState.get(handlerName), new TransformState(state, existingKeys));
+                state = transformState.state();
```
rjernst, line 90: Each handler is create it's own ClusterState. Can we do this more efficiently? Building cluster state is not cheap. Could we instead create a ClusterState.Builder here (or perhaps just the customs and metadata builders), and that is passed to the transform method? I think the TransformState would not be needed, the method could just return the new set of keys.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedStateUpdateTask.java
```diff
@@ -0,0 +1,138 @@
...
+            // Check if we had previous error metadata with version information, don't spam with cluster state updates, if the
+            // version hasn't been updated.
+            if (existingMetadata != null
+                && existingMetadata.errorMetadata() != null
+                && existingMetadata.errorMetadata().version() >= reservedStateVersion.version()) {
+                logger.error("Error processing state change request for [{}] with the following errors [{}]", namespace, errors);
+
+                throw new ReservedStateVersion.IncompatibleVersionException(
+                    format(
+                        "Not updating error state because version [%s] is less or equal to the last state error version [%s]",
```
rjernst, line 107: Why is this an error? Isn't this part of the idempotent nature of the reserved state design?

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedStateVersion.java
```diff
@@ -0,0 +1,58 @@
...
+            return new ReservedStateVersion(updateId, minCompatVersion);
+        }
+    );
+
+    static {
+        PARSER.declareString(ConstructingObjectParser.constructorArg(), VERSION);
+        PARSER.declareString(ConstructingObjectParser.constructorArg(), COMPATIBILITY);
+    }
+
+    public static ReservedStateVersion parse(XContentParser parser, Void v) {
```
rjernst, line 40: Could the implementation specific Void here could be hidden from the signature? Remove from the method and pass null to PARSER?

rjernst (commented):
Looks pretty good. I left some thoughts.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedStateErrorTaskExecutor.java
```diff
@@ -0,0 +1,42 @@
...
+            taskContext.success(
+                () -> taskContext.getTask().listener().delegateFailure((l, s) -> l.onResponse(ActionResponse.Empty.INSTANCE))
+            );
+        }
+        return currentState;
+    }
+
+    @Override
+    public void clusterStatePublished(ClusterState newClusterState) {
+        logger.info("Wrote new error state in immutable metadata");
```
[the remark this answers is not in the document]
grcevski: I think it's useful to know that we've recorded the error state in the cluster state. Updating the state can fail here, maybe we are not the master anymore, or ES terminated before we got to write the error state. By seeing the message in the logs, we definitely know we wrote what's wrong with the file settings. Thanks for catching the wording bug.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedStateUpdateTask.java
grcevski: I can't fully use ClusterState.Builder without more significant refactor of how the TransportActions we are wrapping for this work. All of the transport actions we use for the reserved state handlers take cluster state in and return a new state back. However, I really like the idea of using the builder instead of the TransportState. This way we can remove that TransportState class, we can start using the builder and then we can slowly improve the transport actions to maybe work with a builder in the future, to improve efficiency.

grcevski: I think I said that too quickly, I think using the builder at the moment without refactoring all of the transport actions will be more overhead. I'll rip out transport state though, I can simply use Tuple.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedStateUpdateTask.java
rjernst: What about if the same operator settings file has an update triggered, but for the same version? Wouldn't this cause an internal error, when it should be a no-op?

Separate from this PR, I think we should consider how plugins can use this functionality without directly calling the reserved state service. As you point out, it can be error prone, so perhaps we can redesign the pluggable portion to declaratively specify their resources, and the server is actually what does the registration.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedStateUpdateTask.java
rjernst: I'm ok with waiting on this optimization, but I do think it is something we should look into. Creating cluster state is relatively expensive.

grcevski: I'll follow-up with a PR to optimize this after I check that it's OK to refactor the Transport node actions to use a builder.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedStateErrorTaskExecutor.java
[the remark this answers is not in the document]
grcevski: Yeah that's a valid point, it's not useful to the end user as is the confirmation of new cluster state being set. I'll change this to a debug message instead, so we can use it while working on the feature.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedStateUpdateTask.java
grcevski: Fair point, I throw a specific exception here that would be caught and handled differently by the caller of the service, but it's not ideal, it should be a no-op and a log message only. I'll fix this.

grcevski: > Separate from this PR, I think we should consider how plugins can use this functionality without directly calling the reserved state service. As you point out, it can be error prone, so perhaps we can redesign the pluggable portion to declaratively specify their resources, and the server is actually what does the registration.

I like this idea, perhaps though SPI we can ask all plugins if they want to declare any state chunks that need to be applied and we call the update service internally. This way no plugins will have access to the update service and we'll apply it using our rules.

grcevski:
Thanks again for the review @rjernst, I removed that extra exception and changed such that we return the original cluster state reference to avoid unnecessary update.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedClusterStateService.java
```diff
@@ -179,31 +177,23 @@ public void onFailure(Exception e) {
...
         if (Version.CURRENT.before(reservedStateVersion.minCompatibleVersion())) {
-            errorListener.accept(
-                new ReservedStateVersion.IncompatibleVersionException(
-                    format(
-                        "Cluster state version [%s] is not compatible with this Elasticsearch node",
-                        reservedStateVersion.minCompatibleVersion()
-                    )
+            logger.warn(
+                () -> format(
+                    "Cluster state version [%s] is not compatible with this Elasticsearch node",
```
rjernst, line 184: Perhaps this should mention the reserved state name?

rjernst (approved):
LGTM

grcevski:
@elasticmachine update branch

PR #84657, opened 2022-03-04
[Transform] Introduce TransformScheduler central service

przemekwitek:
This PR introduces `TransformScheduler` class as a central place for scheduling transform tasks. The outline of how this mechanism works is:
- There is a queue of transform tasks, sorted by their `next_scheduled_time`
- On transform `_start` a transform task is added to the queue
- Every `schedulerFrequency` (defaulting to `5s`) the first task is being picked up from the queue and processed, i.e.: its listener is being triggered (indirectly triggering the indexer)
- On transform `_stop` a transform task is removed from the queue and no longer processed
- On transform failure (transient failure, to be precise), the task's `next_scheduled_time` is updated so that it is processed before it normally would. We are using exponential backoff scheme to calculate the appropriate delays
- `SchedulerEngine` is no longer used

Additionally, this PR implements unit tests for `TransformTaskParams` and `TransformContext` classes.

The following work items are **not** covered by this PR. These will be handled separately, in the subsequent PRs:
- configurable per-transform setting: `numFailureRetries`
- support for infinite `numFailureRetries`
- turning some of the irrecoverable errors into recoverable
- increase maximum possible transform frequency

Relates to https://github.com/elastic/elasticsearch/issues/85016

PR #88531, opened 2022-07-14
[Transform] Finetune Schedule to be less noisy on retry and retry slower

hendrikmuhs:
reduce amount of log and audits if the same failure happens in a row
and change the minimum wait time for retrying to 5s

This finetunes changes done in #84657 which makes retrying independent of frequency, which as a result triggers retry faster than before with the side-effect of producing more noise. For this change I investigated the log/audit behavior and adjusted it.

Notes:

 - the change contains mostly test changes, the important code changes are:
   -  [set MIN_DELAY_MILLIS to 5s (from 1s)](https://github.com/elastic/elasticsearch/compare/master...hendrikmuhs:transform-scheduler-reduce-noise?expand=1#diff-654470ab2a56583eaa9410e6f943ad5c41cc66111a0382b9a6704642465d91c0)
   - [dedup logs/audits based on exception class](https://github.com/elastic/elasticsearch/compare/master...hendrikmuhs:transform-scheduler-reduce-noise?expand=1#diff-84dcc4eb8ca941f4a67a6e32bb071461b20338c64a47d1cabba32b26ca8de1d3)
      - the 1st (retry-able) failure gets logged/audited, consequent failures only if they differ from the last one. The last retry is logged as well
      - exceptions are usually wrapped in a `SearchPhaseExecutionException`, so it is important to unwrap this as a 1st step. Just taking the message from the unwrapped class would still produce a lot of noise, because the messages can slightly differ although they have the same root cause. Using the class name of the unwrapped exception seems like a pragmatic way to dedup them.
      - I moved the last error into the context, note: there is no need to persist this as we do not stop the transform after such a failure, but keep the error in-memory only
 - Testing:
   - because testing this change as a system test isn't possible with the current infrastructure, I created mock module tests, [see TransformIndexerFailureHandlingTests](https://github.com/elastic/elasticsearch/compare/master...hendrikmuhs:transform-scheduler-reduce-noise?expand=1#diff-31a950d0df14254a662a2f4ca810d717049c8cfcdfdaba97e3c5da05ba6c0934)
 - Log/Audit message:
   - "Transform encountered an exception: [{Exception}]; Will automatically retry [1/10]"

relates #84657

(marked as non-issue as this is an addition to #84657)

### DIFF
--- a/x-pack/plugin/transform/src/test/java/org/elasticsearch/xpack/transform/notifications/MockTransformAuditor.java
+++ b/x-pack/plugin/transform/src/test/java/org/elasticsearch/xpack/transform/notifications/MockTransformAuditor.java
@@ -7,6 +7,8 @@
 
 package org.elasticsearch.xpack.transform.notifications;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.elasticsearch.client.internal.Client;
 import org.elasticsearch.cluster.ClusterState;
 import org.elasticsearch.cluster.metadata.IndexTemplateMetadata;
@@ -14,7 +16,9 @@ import org.elasticsearch.cluster.metadata.Metadata;
 import org.elasticsearch.cluster.service.ClusterService;
 import org.elasticsearch.common.regex.Regex;
 import org.elasticsearch.xpack.core.common.notifications.Level;
+import org.elasticsearch.xpack.core.transform.notifications.TransformAuditMessage;
 
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
@@ -35,6 +39,7 @@ import static org.mockito.Mockito.when;
 public class MockTransformAuditor extends TransformAuditor {
 
     private static final String MOCK_NODE_NAME = "mock_node_name";
+    private static final Logger logger = LogManager.getLogger(MockTransformAuditor.class);
 
     @SuppressWarnings("unchecked")
     public static MockTransformAuditor createMockAuditor() {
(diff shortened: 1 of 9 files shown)


x-pack/plugin/transform/src/test/java/org/elasticsearch/xpack/transform/notifications/MockTransformAuditor.java
```diff
@@ -134,28 +147,52 @@ public boolean innerMatch(final Level level, final String resourceId, final Stri
...
 
         public SeenAuditExpectation(String expectedName, Level expectedLevel, String expectedResourceId, String expectedMessage) {
-            super(expectedName, expectedLevel, expectedResourceId, expectedMessage);
+            super(expectedName, expectedLevel, expectedResourceId, expectedMessage, 1);
         }
 
         @Override
         public void assertMatched() {
-            assertThat("expected to see " + expectedName + " but did not", saw, equalTo(true));
+            assertThat("expected to see " + expectedName + " but did not", count, equalTo(expectedCount));
```
droberts195, line 155: If would be good to distinguish the case where we don't see the message at all to the case where we see it too many times.  Otherwise, if we ever get this test failure we won't know which scenario was to blame.

In fact, just reading further the error message in `MultipleSeenAuditExpectation` is ideal, so maybe just delete this class and always use `MultipleSeenAuditExpectation` with expected count 1.

hendrikmuhs: I realize that I changed the way it worked, before it just flipped a bool, so it was "seen at least once", with the count I changed it to "exactly once". However the tests pass and "exactly once" is probably what we want anyway.

I will merge the 2 classes and keep the constructor that defaults to "seen once" as convenience. That way it does not trigger further refactorings.

droberts195 (approved):
LGTM

PR #88534, opened 2022-07-14
Support aggregate_double_metric fields in the Field API

salvatore-campagna:
For metric fields of type `aggregate_double_metric`, we would like to return all metric values such as
`max`, `min`, `sum` and `value_count` for a `gauge`, instead of returning just the value for the `default` metric.

## Context

In the past, we had taken the decision to expose fields of type `aggregate_metric_double` as regular `double` fields in the [fields API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html#search-fields-param) by only returning the `default_value` metric sub-field. This decision intended to hide the complexity of this field from Kibana and only expose it as a simple numeric field. After recent discussions with the Kibana team, we decided that Elasticsearch should be transparent as to when a field is an `aggregate_metric_double` field. Elasticsearch should return all metric sub-fields and Kibana will deal with managing the field type.

## Consequences
**This is a breaking change**


romseygeek (approved):
LGTM - a lot simpler than I thought it would need to be!

salvatore-campagna:
I will merge this after the other one...because this will give me some merge conflicts and I prefer to fix merge conflicts here rather than in the other just because this is simpler.

x-pack/plugin/rollup/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/rollup/10_basic.yml
```diff
@@ -119,6 +122,60 @@ setup:
       indices.get:
         index: test
 
+  - do:
```
csoulios, line 125: It's good to have those tests in the rollups code, but perhaps it would be simpler to test `aggregate_metric_double` field in isolation at the more [basic aggregate-metrics yaml tests](https://github.com/elastic/elasticsearch/blob/master/x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/aggregate-metrics/10_basic.yml)

x-pack/plugin/rollup/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/rollup/10_basic.yml
```diff
@@ -119,6 +122,60 @@ setup:
...
+  - match: { hits.hits.2.fields.k8s\.pod\.network\.tx.0.sum: 2.869099752E9 }
+  - match: { hits.hits.2.fields.k8s\.pod\.network\.tx.0.value_count: 2 }
+  - match: { hits.hits.2.fields.k8s\.pod\.network\.rx.0.min: 5.30575198E8 }
+  - match: { hits.hits.2.fields.k8s\.pod\.network\.rx.0.max: 5.30600088E8 }
+  - match: { hits.hits.2.fields.k8s\.pod\.network\.rx.0.sum: 1.061175286E9 }
+  - match: { hits.hits.2.fields.k8s\.pod\.network\.rx.0.value_count: 2 }
+
+  - match: { hits.hits.3.fields.metricset.0: pod }
+  - match: { hits.hits.3.fields.count.0: 16 }
+  - match: { hits.hits.3.fields.k8s\.pod\.network\.tx.0.min: 1.434587694E9 }
```
csoulios, line 170: Also, those tests should be skipped on all versions before 8.4.0

x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
```diff
@@ -494,18 +494,7 @@ public BucketedSort newBucketedSort(
...
-
-            return new SourceValueFetcher(name(), context) {
-                @Override
-                @SuppressWarnings("unchecked")
-                protected Object parseSourceValue(Object value) {
-                    Map<String, Double> metrics = (Map<String, Double>) value;
-                    return metrics.get(defaultMetric.name());
-                }
-            };
+            return SourceValueFetcher.identity(name(), context, format);
```
csoulios, line 497: Should we require that format is null for this field type? 
@romseygeek can you help here?

csoulios: I just noticed  [`FieldValueFetcher#identity` method](https://github.com/elastic/elasticsearch/blob/50793a68a8dfaca22ba1b552523baf4d5e2e15ea/server/src/main/java/org/elasticsearch/index/mapper/SourceValueFetcher.java#L107-L109) does not support `format` param and throws and exception if not null. 

So, I think we are good with this change

salvatore-campagna: That is already done in the `identity` method.

x-pack/plugin/rollup/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/rollup/10_basic.yml
```diff
@@ -23,6 +23,9 @@ setup:
               metricset:
                 type: keyword
                 time_series_dimension: true
+              count:
```
csoulios, line 26: I don't see this field being used anywhere in this test. Is this a left over?

csoulios (approved):
LGTM!

x-pack/plugin/rollup/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/rollup/10_basic.yml
salvatore-campagna: I added it because I saw there was no metric of type `counter`. Will add a few values and test it together with `gauge`.

salvatore-campagna:
@elasticmachine update branch

csoulios:
@elasticmachine merge upstream

PR #88535, opened 2022-07-14
Remove unused datastream snapshot utility from Metadata

original-brownbear:
This method was introduced to fix datastream snapshots during
concurrent index/datastream changes but was never actually
used because we went with a different approach in the end.
=> remove it and its tests


original-brownbear:
Jenkins run elasticsearch-ci/packaging-tests-unix-sample

dakrone (approved):
LGTM

original-brownbear:
Thanks Lee!

PR #88538, opened 2022-07-14
Improve EQL Sequence circuit breaker precision

### LINKED ISSUE #88300: `eql_sequence` circuit breaker sometimes has negative `used` value
Seen in several production clusters running 8.3 (thanks to the changes in #87881) but this also explains #86059 in earlier versions:

    [instance-0000000011] negative value in circuit breaker: [eql_sequence,limit=8589934592/8gb,estimated=-7961288

My guess is that somehow this circuit breaker is releasing bytes it didn't already hold. Seems pretty rare, and not covered by tests since it would trip assertions there.

luigidellaquila:
Fixes  #88300

~~The problem seems to be very sporadic though, I couldn't find a reproducer.~~

There are two good candidates as a cause for this problem, this PR addresses both:
- The [new memory calculation](https://github.com/elastic/elasticsearch/pull/88538/files#r921040374) is added to the total count even if the CB fails, so a cleanup after a failure could lead to a negative value of the CB
- The [delta memory](https://github.com/elastic/elasticsearch/pull/88538/files#r921039661) is calculated only between the beginning and the end of `match()` execution, but memory consumption change also happens outside that method and the initial value could be greater than the final one, so the delta added to the CB could be negative. In general, this second approach makes the memory calculation more accurate.

[edit] It turned out the first hypothesis was correct: the problem was on cleanup.


x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sequence/SequenceMatcher.java
```diff
@@ -114,9 +119,6 @@ private void trackSequence(Sequence sequence) {
      * Returns false if the process needs to be stopped.
      */
     boolean match(int stage, Iterable<Tuple<KeyAndOrdinal, HitReference>> hits) {
-        long ramBytesUsedInFlight = ramBytesUsedInFlight();
```
luigidellaquila, line 117: These values are > 0, if they reduce for some reason during the execution, at the end the memory diff will be negative.
Setting them to 0 as a global value, and avoiding to re-calculate them at the beginning of `match()`, guarantees that the memory added to the circuit breaker is always positive (unless Accountable objects return negative values, but checking the implementations it's not the case)

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sequence/SequenceMatcher.java
```diff
@@ -305,21 +307,23 @@ public void clear() {
...
         return RamUsageEstimator.sizeOf(keyToSequences) + RamUsageEstimator.sizeOf(stageToKeys);
     }
 
-    private long ramBytesUsedCompleted() {
+    protected long ramBytesUsedCompleted() {
         return RamUsageEstimator.sizeOfCollection(completed);
     }
 
     private void addMemory(long bytes, String label) {
-        totalRamBytesUsed += bytes;
```
luigidellaquila, line 317: In case of breaker failure, the memory is not added to the circuit breaker, so the addition to `totalRamBytesUsed` should be done only after `addEstimateBytesAndMaybeBreak` succeeds.

bpintea (commented):
Left some comments.

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sequence/SequenceMatcher.java
bpintea: This is a good point, should this be relatively easy reproducible?

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sequence/SequenceMatcher.java
```diff
@@ -328,11 +332,13 @@ private void clearCircuitBreaker() {
     // expensive, so we just calculate the difference in bytes of the total memory that the matcher's
     // structure occupy for the in-flight tracking of sequences, as well as for the list of completed
     // sequences.
-    private void trackMemory(long prevRamBytesUsedInflight, long prevRamBytesUsedCompleted) {
-        long bytesDiff = ramBytesUsedInFlight() - prevRamBytesUsedInflight;
+    private void trackMemory() {
+        long bytesDiff = ramBytesUsedInFlight() - prevRamBytesUsedInFlight;
         addMemory(bytesDiff, CB_INFLIGHT_LABEL);
+        prevRamBytesUsedInFlight += bytesDiff;
```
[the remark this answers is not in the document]
luigidellaquila: That's exactly the reason for this change: if we calculate the memory footprint change as the difference between the beginning and the end of `match()` execution, we risk to miss everything that happens outside of that.
Eg. 

- a match() is invoked, it calculates a delta of 10k: the real memory footprint is 10KB, the CB is updated with 10KB
- a trim() is invoked: the new real memory footprint is 0KB (the trim released all the memory), the CB is still 10KB (!!!) as `trim()` does not update the CB
- a match is invoked, it calculates another delta of 10k: the real memory footprint is 10KB again (the memory was cleared by the trim()), but the CB is updated to 20KB (!!!) because the new delta (10KB) is added to the CB value (10KB, not updated by trim)

Repeat this pattern for 100 times and at the end you'll have 10KB of real memory used, but the CB thinks you are using 1MB

The same could happen at the opposite, eg. something (outside of `match()` execution) allocates memory, but it does not update the CB (so the CB value is still zero!!!); after that, a `match()` execution REDUCES the memory footprint, so the delta is NEGATIVE. As soon as `match()` tries to update the CB with a negative value, you have the error we are chasing. 

Not sure how likely it is to happen with current implementation, but I see it as a weakness of current algorithm.

With a global variable (that is what this PR does) you solve all of this: you just calculate the total memory every time and you compare it with the total memory calculated at previous round. As long as Accountable objects cannot return negative values, you are always guaranteed that the CB cannot go below zero

luigidellaquila:
Finally reproduced (and added a test case), the problem happened on `clearCircuitBreaker()` after a circuit breaker failure happening *in the parent circuit breaker service*

x-pack/plugin/eql/src/test/java/org/elasticsearch/xpack/eql/execution/sequence/CircuitBreakerTests.java
```diff
@@ -306,6 +223,107 @@ private void assertMemoryCleared(int sequenceFiltersCount, BiFunction<CircuitBre
         }
     }
 
+    public void testParentCircuitBreakerOnClean() {
```
astefan, line 226: The test name imo is not descriptive enough. I'd suggest something like `testEqlCBCleanedUp_on_ParentCBBreak`.

Also, please add a comment to the method:
`// test covering fix for https://github.com/elastic/elasticsearch/issues/88300`

x-pack/plugin/eql/src/test/java/org/elasticsearch/xpack/eql/execution/sequence/CircuitBreakerTests.java
```diff
@@ -306,6 +223,107 @@ private void assertMemoryCleared(int sequenceFiltersCount, BiFunction<CircuitBre
...
+        int sequenceFiltersCount = 2;
+        final int SEARCH_REQUESTS_EXPECTED_COUNT = 2;
+        List<BreakerSettings> eqlBreakerSettings = Collections.singletonList(
```
astefan, line 229: The `eqlBreakerSettings` can be extracted in a separate method. This code block is used in two places in this class.

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sequence/SequenceMatcher.java
```diff
@@ -305,34 +305,42 @@ public void clear() {
         clearCircuitBreaker();
     }
```
astefan, line 307: Add a comment to these two methods stating that they are protected for testing purposes.

astefan (approved):
LGTM. Good job on creating a test for this exact (hard to investigate) failure.
I've left two small comments.

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sequence/SequenceMatcher.java
```diff
@@ -317,14 +315,17 @@ protected long ramBytesUsedCompleted() {
 
     private void addMemory(long bytes, String label) {
         circuitBreaker.addEstimateBytesAndMaybeBreak(bytes, label);
-        totalRamBytesUsed += bytes;
+        if (CB_COMPLETED_LABEL.equals(label)) {
+            prevRamBytesUsedCompleted += bytes;
+        } else {
+            prevRamBytesUsedInFlight += bytes;
+        }
```
bpintea, line 322: This isn't needed, as the values are assigned after the function call to the actual memory consumption.

x-pack/plugin/eql/src/test/java/org/elasticsearch/xpack/eql/execution/sequence/CircuitBreakerTests.java
```diff
@@ -308,6 +223,107 @@ private void assertMemoryCleared(int sequenceFiltersCount, BiFunction<CircuitBre
         }
     }
 
+    public void testParentCircuitBreakerOnClean() {
+        int sequenceFiltersCount = 2;
+        final int SEARCH_REQUESTS_EXPECTED_COUNT = 2;
```
bpintea, line 228: Nit: I guess this is c&p'd and that's why it's final/capitalized, but it looks a bit off that, comparatively, the var above is only effectively final and not actually so.

bpintea (approved):
Nice catch.
I've left one comment, otherwise lgtm!

x-pack/plugin/eql/src/main/java/org/elasticsearch/xpack/eql/execution/sequence/SequenceMatcher.java
luigidellaquila: Good point, it's completely redundant. We don't even need an addMemory() method actually

luigidellaquila:
@elasticmachine update branch

PR #88539, opened 2022-07-14
TSDB: RollupShardIndexer decrease tsid deep copy

weizijun:
RollupShardIndexer decrease tsid deep copy

### DIFF
--- a/x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
+++ b/x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
@@ -279,9 +279,9 @@ class RollupShardIndexer {
 
                         // Create new rollup bucket
                         if (tsidChanged) {
-                            rollupBucketBuilder.init(tsid, lastHistoTimestamp);
+                            rollupBucketBuilder.resetTsid(tsid, lastHistoTimestamp);
                         } else {
-                            rollupBucketBuilder.init(lastHistoTimestamp);
+                            rollupBucketBuilder.resetTimestamp(lastHistoTimestamp);
                         }
 
                         bucketsCreated++;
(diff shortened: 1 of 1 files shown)


csoulios:
@elasticmachine ok to test

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
```diff
@@ -279,7 +278,12 @@ public void collect(int docId, long owningBucketOrd) throws IOException {
                         }
 
                         // Create new rollup bucket
-                        rollupBucketBuilder.init(tsid, lastHistoTimestamp);
+                        if (tsidChanged) {
+                            rollupBucketBuilder.init(tsid, lastHistoTimestamp);
+                        } else {
+                            rollupBucketBuilder.init(lastHistoTimestamp);
```
csoulios, line 284: I understand the point of minimizing the number of `BytesRef.deepCopyOf()` calls, but I don't like the fact that we "initialize" a bucket without passing the tsid.

Maybe we can handle this in the `RollupBucketBuilder` constructor by

```java
public RollupBucketBuilder init(BytesRef tsid, long timestamp) {
  if (this.tsid == null || this.tsid.equals(tsid) == false) {
    this.tsid = BytesRef.deepCopyOf(tsid);
  }
  ......
}
```

I understand you would still want to avoid the `BytesRef.equals()` at every bucket creation, but I think the performance penalty should be negligible

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
```diff
@@ -268,7 +268,6 @@ public void collect(int docId, long owningBucketOrd) throws IOException {
                             + "] -> ["
                             + timestampFormat.format(timestamp)
                             + "]";
-                    lastTsid = BytesRef.deepCopyOf(tsid);
```
csoulios, line 271: This tsid copy had been originally added to support the assertions and debug the code. I would rather remove the assertions altogether than keep the copy at every iteration. 

Thanks for fixing this

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
weizijun: init has two case, one is the tsid changed, the timestamp will also changed. another is timestamp change, it's more common case, if the `BytesRef.equals()` is true ,it will cost more cpu.

I rename the method, and add some comment.

csoulios (approved):
LGTM!

Thank you for this contribution!

weizijun:
> LGTM!
> 
> Thank you for this contribution!

Thanks @csoulios !

PR #88544, opened 2022-07-14
Fix testPreferCopyWithHighestMatchingOperations.

idegtiarenko:
This test arranges replicas that fall behind master on specific nodes.
New allocator might break assertion by shuffling the shards across the cluster.
This change disables rebalancing to fix the test.


DaveCTurner (approved):
LGTM

PR #88547, opened 2022-07-14
Simplify node stopping call

idegtiarenko:
There is no need to use randomNode/predicate combination to stop a node by its
name.


DaveCTurner (commented):
Nicely spotted.

However this loses the check that we're not stopping the last remaining master-eligible node. Could we keep that?

I wonder if we can get rid of `stopRandomNode` entirely, replacing it with a function that selects a random node's name according to the given settings predicate?

idegtiarenko:
>However this loses the check that we're not stopping the last remaining master-eligible node. Could we keep that?

Added the check to existing `stopNode(String nodeName)`.

>I wonder if we can get rid of stopRandomNode entirely, replacing it with a function that selects a random node's name according to the given settings predicate?

Would you mind if I do it in a separate pr to make this change smaller?

DaveCTurner (approved):
LGTM

idegtiarenko:

PR #88548, opened 2022-07-14
fix: use status code 500 for errors if no shard failed

### LINKED ISSUE #20004: Search reduce failures map to 503 status code
[`SearchPhaseExecutionException#status()`](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/action/search/SearchPhaseExecutionException.java#L87-L90) returns a `503 Unavailable` status code if there are no shard failures added to the exception:

```
        if (shardFailures.length == 0) {
            // if no successful shards, it means no active shards, so just return SERVICE_UNAVAILABLE
            return RestStatus.SERVICE_UNAVAILABLE;
        }
```

If an Exception is thrown in the reduce phase (e.g. from the reduce of an aggregation) this means that it will be given a 503 status code since if there were no failures returned from the shards. 

It looks like this is primarily caused by the fact that [`ReduceSearchPhaseException`](https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/action/search/ReduceSearchPhaseException.java) extends `SearchPhaseExecutionException` so maybe either we should override the `status()` method in `ReduceSearchPhaseException` or we should not have `ReduceSearchPhaseException` extend `SearchPhaseExecutionException`?

salvatore-campagna:
If there is no shard failure we would like to use a generic
`Internal Server Error` other than a `Service Unavailable`.
Moreover, if the cause is known we use a status code that
reflects the cause.

Resolves #20004

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/search/SearchPhaseExecutionException.java
+++ b/server/src/main/java/org/elasticsearch/action/search/SearchPhaseExecutionException.java
@@ -70,7 +70,7 @@ public class SearchPhaseExecutionException extends ElasticsearchException {
         if (shardFailures.length == 0) {
             // if no successful shards, the failure can be due to EsRejectedExecutionException during fetch phase
             // on coordinator node. so get the status from cause instead of returning SERVICE_UNAVAILABLE blindly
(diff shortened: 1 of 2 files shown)


salvatore-campagna:
Misunderstood the solution decided for this issue.

PR #88551, opened 2022-07-14
Use status code 500 for errors if no shard failed

### LINKED ISSUE #20004: Search reduce failures map to 503 status code

salvatore-campagna:
If there is no shard failure we would like to use a generic
Internal Server Error other than a Service Unavailable.
Moreover, if the cause is known we use a status code that
reflects the cause.

Resolves #20004

### DIFF
--- a/docs/changelog/88551.yaml
+++ b/docs/changelog/88551.yaml
@@ -1,5 +1,5 @@
 pr: 88551
-summary: "Fix: use status code 500 for errors if no shard failed"
+summary: "Fix: use status code 500 for aggregation reduce phase errors if no shard failed"
 area: Search
 type: bug
 issues:
--- a/docs/changelog/88551.yaml
+++ b/docs/changelog/88551.yaml
@@ -1,0 +1,6 @@
+pr: 88551
+summary: "Fix: use status code 500 for errors if no shard failed"
+area: Search
+type: bug
+issues:
+ - 20004
--- a/server/src/main/java/org/elasticsearch/action/search/ReduceSearchPhaseException.java
+++ b/server/src/main/java/org/elasticsearch/action/search/ReduceSearchPhaseException.java
@@ -8,7 +8,9 @@
 
 package org.elasticsearch.action.search;
 
+import org.elasticsearch.ExceptionsHelper;
 import org.elasticsearch.common.io.stream.StreamInput;
+import org.elasticsearch.rest.RestStatus;
 
 import java.io.IOException;
 
@@ -26,5 +28,22 @@ public class ReduceSearchPhaseException extends SearchPhaseExecutionException {
 
     public ReduceSearchPhaseException(StreamInput in) throws IOException {
         super(in);
+    }
+
+    @Override
+    public RestStatus status() {
+        ShardSearchFailure[] shardFailures = getShardFailures();
+        if (shardFailures.length == 0) {
+            return getCause() == null ? RestStatus.INTERNAL_SERVER_ERROR : ExceptionsHelper.status(getCause());
+        }
+        RestStatus status = shardFailures[0].status();
+        if (shardFailures.length > 1) {
+            for (int i = 1; i < shardFailures.length; i++) {
+                if (shardFailures[i].status().getStatus() >= 500) {
+                    status = shardFailures[i].status();
+                }
+            }
+        }
+        return status;
     }
 }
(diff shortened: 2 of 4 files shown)


server/src/main/java/org/elasticsearch/action/search/ReduceSearchPhaseException.java
```diff
@@ -27,4 +29,21 @@ public ReduceSearchPhaseException(String phaseName, String msg, Throwable cause,
...
+    @Override
+    public RestStatus status() {
+        ShardSearchFailure[] shardFailures = getShardFailures();
+        if (shardFailures.length == 0) {
+            return getCause() == null ? RestStatus.INTERNAL_SERVER_ERROR : ExceptionsHelper.status(getCause());
+        }
+        RestStatus status = shardFailures[0].status();
+        if (shardFailures.length > 1) {
+            for (int i = 1; i < shardFailures.length; i++) {
+                if (shardFailures[i].status().getStatus() >= 500) {
```
csoulios, line 42: I am not a big fan of the hardcoded `500` value here. 
Is there any way to add a constant such as `RestStatus.INTERNAL_SERVER_ERROR.getStatus()`

Also, should you perhaps delegate this to the `RestStatus.status()` method?

Would running something like `RestStatus.status(0, shardFailures.length, shardFailures)` return the same result? Again, I am not an expert on this part of the code, so take this with a grain of salt.

csoulios (commented):
Generally the code looks good. I have left a minor comment. 

However, I must admit I am not an expert on the related code. So @nik9000 or @not-napoleon could have a better opinion here.

docs/changelog/88551.yaml
```diff
@@ -0,0 +1,6 @@
+pr: 88551
+summary: "Fix: use status code 500 for errors if no shard failed"
```
not-napoleon, line 2:
```suggestion
summary: "Fix: use status code 500 for aggregation reduce phase errors if no shard failed"
```

server/src/main/java/org/elasticsearch/action/search/SearchPhaseExecutionException.java
```diff
@@ -161,4 +161,8 @@ public String toString() {
     public String getPhaseName() {
         return phaseName;
     }
+
+    public ShardSearchFailure[] getShardFailures() {
```
not-napoleon, line 165: I think this could be package scoped?

not-napoleon (approved):
LGTM.  Left a couple of nits, but I think this is a fine approach.

server/src/main/java/org/elasticsearch/action/search/ReduceSearchPhaseException.java
```diff
@@ -32,18 +32,10 @@ public ReduceSearchPhaseException(StreamInput in) throws IOException {
...
-        RestStatus status = shardFailures[0].status();
-        if (shardFailures.length > 1) {
-            for (int i = 1; i < shardFailures.length; i++) {
-                if (shardFailures[i].status().getStatus() >= RestStatus.INTERNAL_SERVER_ERROR.getStatus()) {
-                    status = shardFailures[i].status();
-                }
-            }
-        }
-        return status;
+        return super.status();
```
csoulios, line 39: That's much better :)

server/src/main/java/org/elasticsearch/action/search/SearchPhaseExecutionException.java
salvatore-campagna: It looks like there is already a method `shardFailures()` which I didn't see...so I will use the existing one and get rid of the one I added.

javanna:
Heya, have we thought about the consequence in clients code around changing the status code here? I've seen this is marked bug hence it is not considered a breaking change, I assume. But was this discussed before merging?

PR #88554, opened 2022-07-14
Move geo_shape percolator test to the spatial module

iverase:
The test is moved to the percolator module so we can remove the dependency on `TestGeoShapeFieldMapperPlugin`


romseygeek (approved):
LGTM

PR #88556, opened 2022-07-14
Move geo_shape query string test to the spatial module

iverase:
Adds a new SpatialQueryStringIT test in the spatial module to test fields that the module adds.


romseygeek (approved):
LGTM

PR #88559, opened 2022-07-14
Fix Azure repository tests randomly taking many minutes

original-brownbear:
If we run into a seed that causes many fake exceptions and thus retries,
a 100ms retry interval will add up to minutes of test time for tests like
`testLargeBlobCountDeletion` that trigger thousands of requests.
There's no reason not to speed this up by 10x via more aggressive retry
timings as far as I can see so I reduced the timings to avoid randomly
blocked tests.


idegtiarenko (approved):

fcofdez (approved):
LGTM 👍

PR #88560, opened 2022-07-14
Always close directory streams

ChrisHegarty:
This change fixes a couple of directory streams that were not being explicitly closed - which could lead to a descriptor/memory leak.

There are other instances of this issue, but I want so quash these quickly so as to eliminate them from being test failure suspects (on Windows). Other instances of this issue can be done as a follow-up.


grcevski (approved):
LGTM!

rjernst (approved):
LGTM

PR #88561, opened 2022-07-14
Verify that 'details' aren't sent when explain=false

sethmlarson:
Adds an additional test case for the `explain` parameter to ensure indicators `details` field isn't sent when `explain=false`.


masseyke (approved):
LGTM

PR #88562, opened 2022-07-14
Polling for cluster diagnostics information

masseyke:
As shown in step 1.2.2.3 in the diagram at https://github.com/elastic/elasticsearch/issues/85624#issuecomment-1175421820, if there has not been a node elected master for more than 30 seconds, then when a user queries a non-master-eligible node it is supposed to reach out to a master-eligible node to gather information about what that node thinks of master stability. Since we don't want to block a user request while making those remote calls, it should be done speculatively so that we have the information when it is needed. If a non-master-eligible node gets a cluster state changed event that the master is null, then it will wait 10 seconds (in case it gets another cluster state changed event that the master has flipped to non-null quickly, which is the most likely scenario), then reach out to a random master-eligible node using a transport action (https://github.com/elastic/elasticsearch/pull/87984) to get its cluster diagnostics data. This data is stashed away in case a user request comes in. Since the user won't hit this path for 30 seconds after a master has become null (because we tell the user they have a stable master unless it's been missing for more than 30 seconds), we have plenty of time for this -- 10 seconds of delay in case we're notified that there is a new master plus 10 seconds to get the result back from the master-eligible node before timing out, plus 10 seconds to spare.
Note that the pattern used in this PR for doing asynchronous polling is very similar to what was done in #88397. The main differences are that we're polling a single master node here instead of all of them, and we're using a different transport action.
Also note that there will be a follow-up PR that actually uses this information to make decisions about master stability. This PR only keeps it for later use.


server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -588,6 +613,159 @@ private Scheduler.Cancellable fetchClusterFormationInfo(
...
+     * @param cancellableConsumer A consumer for any Cancellable tasks produced by this method
+     */
+    // Non-private for testing
+    void beginPollingRemoteStableMasterHealthIndicatorService(
+        Collection<DiscoveryNode> masterEligibleNodes,
+        Consumer<RemoteMasterHealthResult> responseConsumer,
+        Consumer<Scheduler.Cancellable> cancellableConsumer
+    ) {
+        masterEligibleNodes.stream().findAny().ifPresentOrElse(masterEligibleNode -> {
+            cancellableConsumer.accept(
```
jbaiera, line 645: It looks like this list will grow indefinitely while the master is missing. Should we simplify here? Is there ever a time where the list will contain more than one validly cancellable object?

masseyke: That's a really good point, and the same thing impacts #88397 (it was an unintended side effect of a late change on that one). I'm pretty sure I could just keep one, and whenever I'm about to set the value I cancel the existing one. I think that would be better than potentially leaving a cancellable task running, right?

jbaiera: It only reschedules the execution of the remote diagnostic after it receives a response from the remote server right? At that point the item isn't really cancellable any more, it should be done. I think you're safe to just replace it in the regular rescheduling case, and leave any cancellation logic to code outside the regular reschedule loop?

jbaiera: Also it looks like the only thing that is actually cancellable is the thread pool task that is waiting 10 seconds. If a master node comes back and it attempts to cancel the task but it has already started the remote diagnostic, does the cancellation get ignored and the rescheduling continue?

masseyke: Oh you're right -- we don't schedule this every 10 seconds. We schedule it every 10 seconds after the previous one completes (or errs out). So we just need to keep track of a single cancellable, and we only need to cancel it in the `cancelPollingRemoteStableMasterHealthIndicatorService` method. I'll update that now. Thanks.
```
If a master node comes back and it attempts to cancel the task but it has already started the remote diagnostic, does the cancellation get ignored
```
I believe I looked into this a few weeks ago, and I think I convinced myself that calling cancel issues an interrupt that interrupts network requests, but I could be wrong. Either way, I don't think there's much else we can do about it.

jbaiera: > I believe I looked into this a few weeks ago, and I think I convinced myself that calling cancel issues an interrupt that interrupts network requests, but I could be wrong. Either way, I don't think there's much else we can do about it.

We might want to move the rescheduling logic into the runnable that is cancellable then - If we wait until after the results are in to reschedule the remove diagnostic then we might miss the cancellation when a new master shows up that says to stop collecting remote diagnostics.

Even if we use an atomic reference to store the active cancellable it's possible that we can still end up in a runaway execution I think. Example:
- Thread A is running the remote diagnostic logic it is almost finished and about to reschedule the next execution
- Thread B is trying to cancel the remote diagnostic
- Thread B gets the cancellable task out of the atomic reference
- Thread A schedules the next execution and replaces the old cancellable in the atomic reference
- Thread B calls cancel() on the old cancellable and moves on
- The remote diagnostic check will continue its execution 10 seconds later because the new cancellable has not been cancelled.

A couple ways we might be able to guard against this: We could mutex the cancellation and scheduling logic so we can't schedule anything new while something is being cancelled. Once cancelled, the scheduling logic will need to check something to see that it is indeed cancelled to avoid it scheduling anyway once it obtains the lock.

It might be even safer to modify the rescheduling logic to check if it should not reschedule any more. Either the current node has a master or the diagnostic response indicates there is a master, or something else?

masseyke: OK I think I've taken care of this risk now. I'm putting an AtomicBoolean `isCancelled` variable on the stack each time we begin polling, and I'm synchronizing access in a few places. I've commented in the code where I'm not synchronizing access and why I think it's OK.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -588,6 +622,223 @@ private Scheduler.Cancellable fetchClusterFormationInfo(
...
+                    ConnectionProfile.buildDefaultConnectionProfile(clusterService.getSettings()),
+                    connectionListener
+                );
+            }
+        }, new TimeValue(10, TimeUnit.SECONDS), ThreadPool.Names.SAME);
+    }
+
+    private void cancelPollingRemoteStableMasterHealthIndicatorService() {
+        synchronized (remoteDiagnosticsMutex) {
+            if (remoteStableMasterHealthIndicatorTask != null && remoteCoordinationDiagnosticsCancelled.getAndSet(true) == false) {
```
andreidan, line 811: `remoteCoordinationDiagnosticsCancelled.getAndSet(true) == false`
Is `compareAndSwap` a better option here?

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -588,6 +622,223 @@ private Scheduler.Cancellable fetchClusterFormationInfo(
...
+             * Note that PeerFinder (the source of master eligible nodes) could be updating the master eligible nodes on a different
+             * thread, so making a copy here so that it doesn't change for the short duration of this method.
+             */
+            List<DiscoveryNode> masterEligibleNodes = new ArrayList<>(getMasterEligibleNodes());
+            if (masterEligibleNodes.isEmpty()) {
+                cancelPollingRemoteStableMasterHealthIndicatorService();
+            } else {
+                beginPollingRemoteStableMasterHealthIndicatorService(masterEligibleNodes);
+            }
+        }
```
andreidan, line 839: I'm wondering. Are these the correct start/stop triggers for this polling? (even further, for the `cancel...` trigger - is the last call for `onFoundPeersUpdated` always coinciding with an empty list of `masterEligibleNodes` in order to satisfy the cancel condition `if (masterEligibleNodes.isEmpty())` ?)

Would it be simpler to start the polling of **one** random master eligible node (as indicated [here](https://user-images.githubusercontent.com/453603/177401768-adda0bc7-112b-4b39-943f-202cfc4d142e.png)) from the cluster state change event notification that illustrates we lost the master node ? And stop the polling when we've got a master? 
(similar to how we do it for the cluster formation fetch/cancel but in the case when the local node is _not_ master eligible)

The `onFoundPeersUpdated` could be used to keep track of the discovered peers in a `CopyOnWriteArrayList` and the polling track could pick a random master eligible node from this list to poll.
(just one would be enough I think - based on the reference diagram - as long as one is reachable, and when it is not, we can source pick another one )

andreidan (commented):

I left a few comments and questions

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: I'm pretty sure it's still a problem (and actually a bigger one since there we're polling all nodes instead of one). But I'll put that into a separate PR.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -588,6 +622,223 @@ private Scheduler.Cancellable fetchClusterFormationInfo(
         }, new TimeValue(10, TimeUnit.SECONDS), ThreadPool.Names.SAME);
     }
 
+    private void beginPollingRemoteStableMasterHealthIndicatorService(Collection<DiscoveryNode> masterEligibleNodes) {
+        synchronized (remoteDiagnosticsMutex) {
+            if (remoteCoordinationDiagnosticsCancelled.get()) { // Don't start a 2nd one if the last hasn't been cancelled
```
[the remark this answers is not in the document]
masseyke: I'll add more comments. `remoteCoordinationDiagnosticsCancelled` is a reference to the current/most recent poll. It is kept as an object-level variable so that `cancelPollingRemoteStableMasterHealthIndicatorService()` can cancel it. `isCancelled` is a reference to the one for a specific poll, which is why it's kept on stack. So if we get to a point where there are two polls running at once, the older one will check its copy of `isCancelled` and see that `cancelPollingRemoteStableMasterHealthIndicatorService()` has cancelled it (even if the current value of `remoteCoordinationDiagnosticsCancelled` is `true`.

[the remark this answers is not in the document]
masseyke: > The synchronized block can be avoided here by doing a `compareAndSwap` on the `AtomicBolean` - I renamed `remoteCoordinationDiagnosticsCancelled` to `isRemoteFetchRunning` but some better naming is needed here :)

The reason for the mutex is to protect the assignment of 3 variables atomically -- `remoteStableMasterHealthIndicatorTask`, `remoteCoordinationDiagnosisResult`, and `remoteCoordinationDiagnosticsCancelled`, all while knowing that `remoteCoordinationDiagnosticsCancelled` is `true`. I don't think your suggestion covers that, does it?

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -588,6 +622,223 @@ private Scheduler.Cancellable fetchClusterFormationInfo(
...
+                AtomicReference<Scheduler.Cancellable> cancellableReference = new AtomicReference<>();
+                AtomicReference<RemoteMasterHealthResult> resultReference = new AtomicReference<>();
+                AtomicBoolean isCancelled = new AtomicBoolean(false);
+                beginPollingRemoteStableMasterHealthIndicatorService(masterEligibleNodes, resultReference::set, task -> {
+                    synchronized (remoteDiagnosticsMutex) {
+                        if (isCancelled.get()) {
```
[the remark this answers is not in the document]
masseyke: I think what you're missing is that it's not just protecting the read of one variable -- it's protecting the read and assignment of 3 variables -- `remoteStableMasterHealthIndicatorTask`, `remoteCoordinationDiagnosisResult`, and `remoteCoordinationDiagnosticsCancelled`

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -588,6 +622,223 @@ private Scheduler.Cancellable fetchClusterFormationInfo(
...
+        DiscoveryNode node,
+        Consumer<RemoteMasterHealthResult> responseConsumer,
+        AtomicBoolean isCancelled
+    ) {
+        StepListener<Releasable> connectionListener = new StepListener<>();
+        StepListener<CoordinationDiagnosticsAction.Response> fetchCoordinationDiagnosticsListener = new StepListener<>();
+        long startTime = System.nanoTime();
+        connectionListener.whenComplete(releasable -> {
+            if (isCancelled.get()) {
+                IOUtils.close(releasable);
```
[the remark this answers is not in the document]
masseyke: It could. But previously we were using `Releasables.close()`, which also can throw an exception (it catches the IOException and rethrows it as an `UncheckedIOException` so that it doesn't cause compilation errors with the streams API but the result is the same. In this block I think it won't matter because this poll was cancelled anyway, so we don't need to make sure to reschedule anything. In the block below where `Releasables.close()` is used, I believe the exception will be caught by `fetchCoordinationDiagnosticsListener`'s exception handler, and the task will be rescheduled.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: I don't think so, for the same reasons given above.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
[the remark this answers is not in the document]
andreidan: These assignments happen inside the proposed CAS operation 
ie. 
```
 if (isRemoteFetchRunning.compareAndSwap(false, true)) {
```
The **one thread** that gets to flip this boolean will be granted access to the contents of this if-block. 
I believe it would be covered.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: The problem is that non-master-eligible nodes do not have _any_ peers in PeerFinder at the time that `clusterChanged` is called with a notification that the master is null. PeerFinder's peers list is populated at some point after that in another thread. I had originally done what you say here, but realized in an integration test (in an upcoming PR) that I would _never_ get any remote diagnostics back because the list of master eligible nodes was _always_ empty, so the polling would stop immediately.

andreidan: Wouldn't the polling retry 10 seconds later? 
ie. if there's nothing in the list of discovered peers we should just retry in 10 seconds and pick a random node from the list of peers at that point ... if any available

masseyke:
Closing this in favor of #89014

PR #88563, opened 2022-07-15
TSDB: fix rollup test testRollupSparseMetrics

weizijun:
fix rollup test testRollupSparseMetrics.
The test failed reason is that the last_value field (numeric_2) use the max aggs, so it must make the numeric_2 field in the desc order.

### DIFF
--- a/x-pack/plugin/rollup/src/test/java/org/elasticsearch/xpack/rollup/v2/RollupActionSingleNodeTests.java
+++ b/x-pack/plugin/rollup/src/test/java/org/elasticsearch/xpack/rollup/v2/RollupActionSingleNodeTests.java
@@ -6,7 +6,6 @@
  */
 package org.elasticsearch.xpack.rollup.v2;
 
-import org.apache.lucene.tests.util.LuceneTestCase;
 import org.elasticsearch.ElasticsearchException;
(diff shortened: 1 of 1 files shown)


csoulios:
This test has been fixed for PR #87929

csoulios:
@elasticmachine ok to test

salvatore-campagna:
Thanks @weizijun for your contribution. Indeed the issue was with using the Max aggregation. This has been fixed replacing the max aggregation with a top metrics aggregation in PR #87929 . That allows us to use values that don't need to be ordered anymore.

weizijun:
> Thanks @weizijun for your contribution. Indeed the issue was with using the Max aggregation. This has been fixed replacing the max aggregation with a top metrics aggregation in PR #87929 . That allows us to use values that don't need to be ordered anymore.

yeah, it need a Last aggregation, Now it hasn't. So the values will be desc when using the Max aggregation.

csoulios:
@weizijun if you don't mind I will close this PR in favor of https://github.com/elastic/elasticsearch/pull/87929 that follows a more generic  and correct approach to solving this problem.

weizijun:
> @weizijun if you don't mind I will close this PR in favor of #87929 that follows a more generic and correct approach to solving this problem.

ok!

PR #88564, opened 2022-07-15
TSDB: Fixed tsid parse unsigned_long value

weizijun:
Fixed tsid parse unsigned_long value: Use BigInteger to save unsigned_long


csoulios:
@elasticmachine ok to test

nik9000 (approved):

server/src/main/java/org/elasticsearch/search/DocValueFormat.java
```diff
@@ -712,19 +712,10 @@ public BytesRef parseBytesRef(Object value) {
...
-                    // For a long encoded number, we must check if the number can be the encoded value
-                    // of an unsigned_long.
-                    Number ul = (Number) UNSIGNED_LONG_SHIFTED.format(l);
-                    if (l == ul) {
-                        builder.addLong(f, l);
-                    } else {
-                        long ll = UNSIGNED_LONG_SHIFTED.parseLong(String.valueOf(l), false, () -> 0L);
-                        builder.addUnsignedLong(f, ll);
-                    }
+                    builder.addLong(f, (Long) v);
```
nik9000, line 715: The test failures come because this is `Integer`. Could you break arm of the if statement up into `else if (v instanceof Long l)` and `else if (v instanceof Integer i)`? I think that'd fix it.

server/src/main/java/org/elasticsearch/index/mapper/TimeSeriesIdFieldMapper.java
```diff
@@ -264,8 +264,14 @@ public void addLong(String fieldName, long value) {
...
-                out.write((byte) 'u');
-                out.writeLong(value);
+                Object ul = DocValueFormat.UNSIGNED_LONG_SHIFTED.format(value);
+                if (ul instanceof Long l) {
+                    out.write((byte) 'l');
+                    out.writeLong(l);
+                } else {
+                    out.write((byte) 'u');
+                    out.writeLong(value);
+                }
```
weizijun, line 274: @nik9000 I change the code, if the value is Long, use 'l' to encode it. If the value is too long, the value is a BigInteger type, so use the 'u' to encode it.
So, when decode the value, small value will decode to be Long, long value will decode to be BigInteger.

weizijun:
@elasticmachine update branch

weizijun:
@elasticmachine update branch

nik9000:
@csoulios, could you have a look at this one? I remember you having an opinion here, but I don't remember what it was.

weizijun:
@elasticmachine update branch

csoulios:
@elasticmachine merge upstream

csoulios (commented):
Change looks good. 
Can you please update the changelog file?

Thanks

csoulios (approved):
LGTM!

PR #88565, opened 2022-07-15
TSDB: Downsampling copies all settings from source to rollup index

csoulios:
This PR modified the rollup action so that all index settings are copied from the  source index to the rollup index. 

Relates to https://github.com/elastic/elasticsearch/pull/85708


x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -427,42 +432,42 @@ public static String createRollupIndexMapping(
...
-                    .put(IndexMetadata.INDEX_ROLLUP_SOURCE_UUID.getKey(), originalIndexUuid)
-                    .put(IndexMetadata.INDEX_HIDDEN_SETTING.getKey(), sourceIndexMetadata.isHidden())
-                    // Add the time series index settings
-                    .put(IndexSettings.MODE.getKey(), indexMode)
-                    .putList(IndexMetadata.INDEX_ROUTING_PATH.getKey(), indexRoutingPath)
-                    .put(IndexSettings.TIME_SERIES_START_TIME.getKey(), startTime.toString())
-                    .put(IndexSettings.TIME_SERIES_END_TIME.getKey(), endTime.toString())
-            );
+        if (IndexMetadata.INDEX_ROLLUP_SOURCE_UUID.exists(sourceIndexMetadata.getSettings()) == false
+            || IndexMetadata.INDEX_ROLLUP_SOURCE_UUID.exists(sourceIndexMetadata.getSettings()) == false) {
```
dakrone, line 464: This is the same check twice, is one of these supposed to be checking the `*_NAME` instead of `*_UUID`?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -427,42 +432,42 @@ public static String createRollupIndexMapping(
...
+            Index sourceIndex = sourceIndexMetadata.getIndex();
+            targetSettings.put(IndexMetadata.INDEX_ROLLUP_SOURCE_NAME.getKey(), sourceIndex.getName())
+                .put(IndexMetadata.INDEX_ROLLUP_SOURCE_UUID.getKey(), sourceIndex.getUUID());
```
dakrone, line 467: Do we already have a test for doing a rollup-of-a-rollup where the original index is maintained in the rollup source/uuid? If not, could you add one?

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -242,19 +246,20 @@ protected void masterOperation(
...
-                            final Settings settings = Settings.builder()
-                                .put(IndexMetadata.SETTING_BLOCKS_WRITE, true)
-                                .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, sourceIndexMetadata.getNumberOfReplicas())
-                                .build();
-                            UpdateSettingsRequest updateSettingsReq = new UpdateSettingsRequest(settings, rollupIndexName);
+                            final Settings.Builder settings = Settings.builder().put(IndexMetadata.SETTING_BLOCKS_WRITE, true);
+                            // Number of replicas had been previously set to 0 to speed up index population
+                            if (sourceIndexMetadata.getNumberOfReplicas() > 0) {
+                                settings.put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, sourceIndexMetadata.getNumberOfReplicas());
+                            }
```
dakrone, line 253: I don't understand this piece, why do we copy replicas only if they are greater than 0? How would this behave if someone had something like `index.auto_expand_replicas: 0-2` where the number of replicas is variable? Shouldn't we be consistent and maintain the number of replicas as defined?

dakrone (changes requested):
Thanks! I left a couple of comments

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: Right now rollup-of-rollups does not fully work. I plan to add support for rollup-of-rollup in a following PR. 
I will definitely add tests in that PR

This part is only plumbing that is going to be needed for rollups of rollups

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: Good catch, it was `_NAME`and `_UUID` . Fixing it now.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: The idea behind this is that when we create the rollup index we explicitly [set that `num_of_replicas=0`](https://github.com/elastic/elasticsearch/blob/52cf0f7de4d941a8e68d8fdb413fcf23334e9e2c/x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java#L508). We do so, because we don't want the index to replicate while we populate the index.

When all documents have been written in the index, we set the num_of_replicas equal to the num_of_replicas of the source index. Of course,  if num_of_replicas of the source index is 0, we don't have to do anything, hence the conditional.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: Fixed in [02c856d](https://github.com/elastic/elasticsearch/pull/88565/commits/02c856deba6986f5199c928b1dd829a667e53abc)

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
dakrone: > Right now rollup-of-rollups does not fully work.

As some subsequent work (not in this PR), does it make sense to reject the request if someone tries to rollup an index that's already been rolled up?

dakrone (approved):
LGTM, thanks!

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: In data streams this cannot happen because no two indices for the same period can co-exist.

When it comes to concrete indices, there are very valid use cases for allowing multiple rollup indices of the same source index. For example having multiple rollup intervals (hourlies, dailies etc) or having multiple timezones (right now we support only UTC, but tin the future we should support more timezones).

csoulios:
Thanks a lot for the review!

PR #88577, opened 2022-07-18
Remove random runner from build tools (internal) tests

breskeby:
This removes unrequited overhead for writing build tool and build tool internal tests and mechanically ports all junit3 tests to junit4.


build-tools-internal/src/test/java/org/elasticsearch/gradle/internal/test/rest/transform/match/AddMatchTests.java
```diff
@@ -34,7 +40,7 @@ public void testAddAllNotSupported() throws Exception {
         JsonNode addNode = MAPPER.convertValue("_doc", JsonNode.class);
         assertEquals(
             "adding matches is only supported for named tests",
-            expectThrows(
+            Assert.assertThrows(
```
pugnascotia, line 43: Should we statically import `assertThrows`, as with other methods on `Assert`?

pugnascotia (approved):
LGTM.

So do we not need things like `@ThreadLeakFilters` now?

breskeby:
> ThreadLeakFilters



> LGTM.
> 
> So do we not need things like `@ThreadLeakFilters` now?

Newer versions of gradle should handle this properly. Also we have ported / wrote most integration tests now using Spock not using that filter for a while. I haven't seen any issues. Unit tests should not have ran into problems here anyhow.

mark-vieira (approved):
LGTM

PR #88583, opened 2022-07-18
Remove geo_shape dependency on FieldFilterMapperPluginTests

iverase:
The test only relies on field names so we don't really need to have the dependency on geo_shape field.


romseygeek (approved):
LGTM

PR #88584, opened 2022-07-18
Fix docker positional params (take 2)

pugnascotia:
Follow-up to (and re-apply of) #88502, which we had to revert.

The change itself was good, but it turned out that the test we added only applies to non-Cloud images. This is because for the Cloud images, the `ENTRYPOINT` and `CMD` are different, such that defining the `ELASTIC_PASSWORD` doesn't set the `elastic` user's password. Since Cloud currently assumes control of image startup, there's little point working around this in the test since we have no control over what's happening at runtime in production.

@mark-vieira note that we didn't catch this during PRs, because we only run the `destructiveDistroTest.docker` task in CI, which only tests the default image. We only test all the Docker images on merged changes. What do you think about reworking this, so that we get better coverage when we're explicitly testing packaging changes?

### DIFF
--- a/distribution/docker/src/docker/bin/docker-entrypoint.sh
+++ b/distribution/docker/src/docker/bin/docker-entrypoint.sh
@@ -81,4 +81,4 @@ fi
 
 # Signal forwarding and child reaping is handled by `tini`, which is the
 # actual entrypoint of the container
-exec /usr/share/elasticsearch/bin/elasticsearch $POSITIONAL_PARAMETERS <<<"$KEYSTORE_PASSWORD"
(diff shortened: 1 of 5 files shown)


mark-vieira:
> What do you think about reworking this, so that we get better coverage when we're explicitly testing packaging changes?

If I recall, this is partly intentional because of the issue of downloading beats after a version bump. If we integrated cloud image testing into _all_ PR testing, folks would be blocked until a successfull snapshot was published. I think we can revisit this once we are consuming the DRA artifacts for beats.

mark-vieira (approved):

PR #88588, opened 2022-07-18
Move geo_shape query builder tests to spatial module

iverase:
Currently we are testing geo_points an  geo_shape fields for query builders in the same class in server. This PR extracts a TestCase for the query builder test (geo_bounding_box, geo_distance and geo_shape) and splits the testing between server (for geo_point) and spatial (for geo_shape).


not-napoleon (approved):
LGTM

PR #88590, opened 2022-07-18
Add shard explain info to ReactiveReason about unassigned shards

### LINKED ISSUE #85243: Provide more information about autoscaling policy decisions to improve debugging

arteam:
Add two new fields into the reactive autoscaling policy decision response.

* `unassigned_node_decisions` - `can_allocate` and `can_remain` decisions for  unassigned shards (limited by 5)
* `assigned_node_decisions` - `can_allocate` and `can_remain` decisions decisions for assigned shards (limited by 5)

```json
"unassigned_node_decisions": {
    "[MCVEOGORUQ][0]": {
      "can_allocate_decisions": [
        {
          "node_id": "node1",
          "node_name": "",
          "transport_address": "0.0.0.0:1",
          "deciders": [
            {
              "decider": "no_label",
              "decision": "NO",
              "explanation": "No space to allocate"
            }
          ]
        }
      ],
      "can_remain_decision": null
    }
  },
  "assigned_node_decisions": {
    "[MCVEOGORUQ][0]": {
      "can_allocate_decisions": [],
      "can_remain_decision": {
        "node_id": "node1",
        "node_name": "",
        "transport_address": "0.0.0.0:1",
        "deciders": [
          {
            "decider": "multi_throttle",
            "decision": "THROTTLE",
            "explanation": "is not active yet"
          },
          {
            "decider": "multi_no",
            "decision": "NO",
            "explanation": "No multi decision"
          },
          {
            "decider": "multi_yes",
            "decision": "YES",
            "explanation": "Yes multi decision"
          }
        ]
      }
    }
  }
```

Resolves #85243


x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -160,13 +170,19 @@ public AutoscalingDeciderResult scale(Settings configuration, AutoscalingDecider
                 unassignedBytes,
                 unassignedBytesUnassignedShards.shardIds(),
                 assignedBytes,
-                assignedBytesUnmovableShards.shardIds()
+                assignedBytesUnmovableShards.shardIds(),
+                unassignedBytesUnassignedShards.shards().size() > 0
```
fcofdez, line 174: I wonder if we should compute the allocation explanation here when we already computed the allocation decision in `ReactiveStorageDeciderService.AllocationState#storagePreventsAllocation` and `ReactiveStorageDeciderService.AllocationState#storagePreventsRemainOrMove` maybe we could use that information? That should simplify things. wdyt @henningandersen?

fcofdez (commented):
I left a few comments

server/src/main/java/org/elasticsearch/node/Node.java
```diff
@@ -625,6 +625,7 @@ protected Node(
             final RerouteService rerouteService = new BatchedRerouteService(clusterService, clusterModule.getAllocationService()::reroute);
             rerouteServiceReference.set(rerouteService);
             clusterService.setRerouteService(rerouteService);
+            clusterService.setAllocationService(clusterModule.getAllocationService());
```
fcofdez, line 628: `AllocationService` can be injected in `TransportGetAutoscalingCapacityAction`, so there's no need to modify `ClusterService`

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
```diff
@@ -22,19 +22,21 @@
...
+import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
 
 /**
  * {@link ShardRouting} immutably encapsulates information about shard
  * indexRoutings like id, state, version, etc.
  */
-public final class ShardRouting implements Writeable, ToXContentObject {
+public final class ShardRouting implements Writeable, ToXContentObject, Comparable<ShardRouting> {
```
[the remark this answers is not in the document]
arteam: Gotcha. That's what I originally did, but then moved to making `ShardRouting` extending `Comparable` for brevity.
I will revert it.

server/src/main/java/org/elasticsearch/node/Node.java
arteam: Hm, `ReactiveStorageDeciderService` gets created in `Autoscaling`, not in `TransportGetAutoscalingCapacityAction`, where we don't have access to the DI framework yet, so we have to pass `AllocationService` via `ClusterServce`.
Maybe I'm missing out something?

fcofdez: That's correct, but you can still provide a reference in `AutoscalingDeciderService#scale`? I think that the `ClusterService` shouldn't provide the `AllocationService`, I feel it's confusing.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
arteam: Yeah, we kind of do a double work here, but we also take advantage of the known JSON representation of `ShardAllocationDecision` that's used in the Cluster Allocation Explain API

henningandersen: I am not sure we need to avoid the double work (is only for one shard), but I do think we should consider to not use the `AllocationService` to determine the result.

The `AllocationService` interacts with the gateway service and might provoke a fetch of shard data from all nodes. I'd prefer to decouple all that code here, since we do not rely on the gateway service for the determination of autoscaling.

Also, the `AllocationService` uses a potentially newer `ClusterState` and `ClusterInfo`. So the reason obtained may be different from the reason we decided to autoscale (or no reason at all). Additionally, `AllocationState.state` has been fixed wrt. shrink/split/clone shards to avoid infinite scale up. We should therefore use that cluster state to get the reasons to avoid confusion.

I think we can still use `AllocateUnassignedDecision` to produce the JSON. Or we can simply output the decisions including reasons one by one from the autoscaling code. We could reuse `NodeAllocationResult` to produce the output per node at least, perhaps using `AbstractAllocationDecision.nodeDecisionsToXContent`.

Also, a specific tier only concerns itself with the nodes in the tier. I think we could limit the output to only have decisions for those nodes rather than all nodes. This might not be entirely true for the move case, but I think it is for the unassigned case.

This would also avoid the dependency on `AllocationService`, simplifying this PR (and the code in general).

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
```diff
@@ -430,29 +456,52 @@ private static void verifyScale(ClusterState state, long expectedDifference, Str
             assertThat(resultReason.summary(), equalTo(reason));
             assertThat(resultReason.unassignedShardIds(), equalTo(decider.allocationState(context).storagePreventsAllocation().shardIds()));
             assertThat(resultReason.assignedShardIds(), equalTo(decider.allocationState(context).storagePreventsRemainOrMove().shardIds()));
+            assertThat(
```
henningandersen, line 459: Can we add this validation to `ReactiveStorageIT.testScaleUp` too?

henningandersen (commented):
Thanks Artem, I think we need to take a slightly different direction here, see comments below.

arteam:
@elasticsearchmachine run elasticsearch-ci/docs-check

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
arteam: Hey @henningandersen, thanks for the feedback! I've reworked the PR to get rid of `AllocationService` and bubble `NodeAllocationResult` from `cannotAllocateDueToStorage` and `storagePreventsRemainOrMove` up to the `scale` method, and output them in `ReactiveReason`.

Let me know what you think about it!

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -915,6 +965,13 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
             builder.field("assigned", assigned);
             builder.field("assigned_shards", assignedShardIds.stream().limit(MAX_AMOUNT_OF_SHARDS).toList());
             builder.field("assigned_shards_count", assignedShardIds.size());
+            AbstractAllocationDecision.nodeDecisionsToXContent(
+                unassignedAllocationResults,
+                builder,
+                params,
+                "unassigned_allocation_results"
+            );
+            AbstractAllocationDecision.nodeDecisionsToXContent(assignedAllocationResults, builder, params, "assigned_allocation_results");
```
henningandersen, line 974: For the assigned shards, it is relevant to know both why it cannot remain on the current node and why it cannot be allocated to any node in the tier.

henningandersen (commented):
Thanks Artem, this direction looks good. I'd like a bit more explain info though.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -274,15 +279,26 @@ public static class AllocationState {
...
             return new ShardsSize(
-                unassignedShards.stream().mapToLong(this::sizeOf).sum(),
-                unassignedShards.stream().map(ShardRouting::shardId).collect(Collectors.toCollection(TreeSet::new))
+                unassignedShards.stream().map(e -> e.shard).mapToLong(this::sizeOf).sum(),
+                unassignedShards.stream().map(e -> e.shard.shardId()).collect(Collectors.toCollection(TreeSet::new)),
+                unassignedShards.stream()
+                    .filter(e -> e.nodeAllocationResults.size() > 0)
+                    .min(Comparator.comparing(e -> e.shard.shardId()))
+                    .map(e -> e.nodeAllocationResults)
+                    .orElse(null)
```
fcofdez, line 301: Can't we default here to an empty list when there aren't unassigned shards due to storage?

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveReasonTests.java
```diff
@@ -37,12 +44,21 @@ public void testXContent() throws IOException {
         String indexName = randomAlphaOfLength(10);
         SortedSet<ShardId> unassignedShardIds = new TreeSet<>(randomUnique(() -> new ShardId(indexName, indexUUID, randomInt(1000)), 600));
         SortedSet<ShardId> assignedShardIds = new TreeSet<>(randomUnique(() -> new ShardId(indexName, indexUUID, randomInt(1000)), 600));
+        DiscoveryNode discoveryNode = new DiscoveryNode("node1", buildNewFakeTransportAddress(), emptyMap(), emptySet(), Version.CURRENT);
+        List<NodeAllocationResult> unassignedShardAllocationDecision = List.of(
+            new NodeAllocationResult(discoveryNode, AllocationDecision.NO, Decision.NO, 1)
```
fcofdez, line 49: Can we add a Decision that includes more details using `org.elasticsearch.cluster.routing.allocation.decider.Decision#single`?

fcofdez (commented):
I left a few comments and I agree with Henning's points. I prefer this direction. Thanks!

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -274,15 +279,26 @@ public static class AllocationState {
             this.roles = roles;
         }
 
+        record DecisionAllocationResult(ShardRouting shard, boolean decision, List<NodeAllocationResult> nodeAllocationResults) {}
```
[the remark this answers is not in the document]
arteam: Renamed to `ShardAllocationResults`

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -301,28 +317,36 @@ && canAllocate(shard, allocation) == false) {
             }
 
             // track these to ensure we do not double account if they both cannot remain and allocated due to storage.
-            Set<ShardRouting> unmovableShards = candidates.stream()
+            List<DecisionAllocationResult> unmovableShardNodeAllocationResults = candidates.stream()
                 .filter(s -> allocatedToTier(s, allocation))
-                .filter(s -> cannotRemainDueToStorage(s, allocation))
-                .collect(Collectors.toSet());
+                .flatMap(s -> Optional.of(cannotRemainDueToStorage(s, allocation)).filter(e -> e.decision).stream())
```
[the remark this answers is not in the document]
arteam: It's intentional. It uses the pattern Optional.filter().stream() to convert a value with `e.decision` equal to `true` to a stream of one element. It's equivalent to 
```java
map(s -> cannotRemainDueToStorage(s, allocation))
                .flatMap(r -> r.decision ? Stream.of(r) : Stream.empty())
```

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderReasonWireSerializationTests.java
```diff
@@ -86,7 +136,16 @@ protected ReactiveStorageDeciderService.ReactiveReason createTestInstance() {
...
+            new TreeSet<>(randomUnique(() -> new ShardId(randomAlphaOfLength(8), UUIDs.randomBase64UUID(), randomInt(5)), 8)),
+            randomList(8, this::randomNodeDecision),
+            randomList(8, this::randomNodeDecision)
+        );
+    }
+
+    private NodeDecision randomNodeDecision() {
+        return new NodeDecision(
+            new DiscoveryNode(randomAlphaOfLength(6), buildNewFakeTransportAddress(), emptyMap(), emptySet(), Version.CURRENT),
+            randomFrom(Decision.NO, Decision.YES, Decision.THROTTLE)
```
fcofdez, line 148: can we add single and multiple decisions that include an explanation?

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderReasonWireSerializationTests.java
```diff
@@ -15,63 +19,109 @@
...
-                    new TreeSet<>(randomUnique(() -> new ShardId(randomAlphaOfLength(8), UUIDs.randomBase64UUID(), randomInt(5)), 8))
+                    new TreeSet<>(randomUnique(() -> new ShardId(randomAlphaOfLength(8), UUIDs.randomBase64UUID(), randomInt(5)), 8)),
+                    instance.unassignedAllocationResults(),
+                    instance.assignedAllocationResults()
                 );
             case 5:
                 return new ReactiveStorageDeciderService.ReactiveReason(
                     instance.summary(),
                     instance.unassigned(),
                     new TreeSet<>(randomUnique(() -> new ShardId(randomAlphaOfLength(8), UUIDs.randomBase64UUID(), randomInt(5)), 8)),
```
fcofdez, line 90: I think this should be `instance.unassignedShardIds()`?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/NodeDecision.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public NodeDecision(StreamInput in) throws IOException {
+        node = new DiscoveryNode(in);
+        decision = in.readOptionalWriteable(Decision::readFrom);
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        node.writeTo(out);
+        out.writeOptionalWriteable(decision);
```
fcofdez, line 42: is `decision` nullable here? we might want to annotate the field if that's the case.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/NodeDecision.java
```diff
@@ -0,0 +1,75 @@
...
+import org.elasticsearch.common.io.stream.Writeable;
+import org.elasticsearch.xcontent.ToXContentObject;
+import org.elasticsearch.xcontent.XContentBuilder;
+
+import java.io.IOException;
+import java.util.Objects;
+
+import static org.elasticsearch.cluster.routing.allocation.AbstractAllocationDecision.discoveryNodeToXContent;
+
+public class NodeDecision implements ToXContentObject, Writeable {
```
fcofdez, line 24: can we add serialization tests for this class?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -274,19 +285,27 @@ public static class AllocationState {
...
+                .flatMap(shard -> cannotAllocateDueToStorage(shard, allocation).stream())
                 .toList();
-            return new ShardsSize(
-                unassignedShards.stream().mapToLong(this::sizeOf).sum(),
-                unassignedShards.stream().map(ShardRouting::shardId).collect(Collectors.toCollection(TreeSet::new))
+            return new ShardsAllocationResults(
+                unassignedShards.stream().map(e -> e.shard).mapToLong(this::sizeOf).sum(),
+                unassignedShards.stream().map(e -> e.shard.shardId()).collect(Collectors.toCollection(TreeSet::new)),
+                unassignedShards.stream()
+                    .filter(e -> e.nodeAllocationResults.size() > 0)
```
fcofdez, line 300: nit: `s/e/shardNodeDecision/` or something more descriptive.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -917,6 +1007,20 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
             builder.field("assigned", assigned);
             builder.field("assigned_shards", assignedShardIds.stream().limit(MAX_AMOUNT_OF_SHARDS).toList());
             builder.field("assigned_shards_count", assignedShardIds.size());
+            if (unassignedAllocationResults != null && unassignedAllocationResults.size() > 0) {
```
fcofdez, line 1010: This can be simplified with `builder.xContentList("unassigned_allocation_results", unassignedAllocationResults);` and I think we should serialize it as long as it is not null.

arteam:
Hey @fcofdez and @henningandersen! I've addressed the comments from the last batch,

I created a new class `NodeDecision` to represent an abstraction holding a node and a decision instead of reusing `NodeAllocationResult`. `NodeAllocationResult` is suppose to work only with `can_allocate` decisions and we needed the class to hold also `can_remain` decisions.

I've also added suggestions from Henning for adding `can_allocate` decision for the nodes on the tier, if a shard can't be moved, and also I added a `can_remain` decision for the node if it can't be allocated due to storage.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -357,24 +417,26 @@ private boolean allocatedToTier(ShardRouting s, RoutingAllocation allocation) {
...
+                ).filter(nar -> ReactiveStorageDeciderService.isDiskOnlyNoDecision(nar.decision())).toList();
+                if (diskOnly.size() > 0 && shard.unassigned() && shard.recoverySource().getType() == RecoverySource.Type.LOCAL_SHARDS) {
                     // For resize shards only allow autoscaling if there is no other node where the shard could fit had it not been
                     // a resize shard. Notice that we already removed any initial_recovery filters.
-                    diskOnly = nodesInTier(allocation.routingNodes()).map(node -> allocationDeciders.canAllocate(shard, node, allocation))
-                        .anyMatch(ReactiveStorageDeciderService::isResizeOnlyNoDecision) == false;
+                    List<NodeDecision> resizeOnly = nodesInTier(allocation.routingNodes()).map(
+                        node -> new NodeDecision(node.node(), allocationDeciders.canAllocate(shard, node, allocation))
+                    ).filter(nar -> ReactiveStorageDeciderService.isResizeOnlyNoDecision(nar.decision())).toList();
+                    return resizeOnly.isEmpty() ? Optional.of(new ShardNodeDecisions(shard, resizeOnly)) : Optional.empty();
```
fcofdez, line 437: I think this is incorrect, we should return the diskOnly decision here as well, otherwise we won't know why we cannot allocate the shard, right?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -301,28 +320,69 @@ && canAllocate(shard, allocation) == false) {
...
+            long unallocatableBytes = unallocatedShardNodeAllocationResults.stream().map(e -> e.shard).mapToLong(this::sizeOf).sum();
 
-            return new ShardsSize(
+            return new ShardsAllocationResults(
                 unallocatableBytes + unmovableBytes,
-                Stream.concat(unmovableShards.stream(), unallocatedShards.stream())
-                    .map(ShardRouting::shardId)
-                    .collect(Collectors.toCollection(TreeSet::new))
+                Stream.concat(unmovableShardNodeAllocationResults.stream(), unallocatedShardNodeAllocationResults.stream())
+                    .map(e -> e.shard.shardId())
```
fcofdez, line 374: nit: can you use a more descriptive variable message?

fcofdez (commented):
This looks mostly good, I left a few more comments. Thanks for all the iterations!

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderReasonWireSerializationTests.java
```diff
@@ -86,7 +136,16 @@ protected ReactiveStorageDeciderService.ReactiveReason createTestInstance() {
             randomNonNegativeLong(),
             new TreeSet<>(randomUnique(() -> new ShardId(randomAlphaOfLength(8), UUIDs.randomBase64UUID(), randomInt(5)), 8)),
             randomNonNegativeLong(),
-            new TreeSet<>(randomUnique(() -> new ShardId(randomAlphaOfLength(8), UUIDs.randomBase64UUID(), randomInt(5)), 8))
+            new TreeSet<>(randomUnique(() -> new ShardId(randomAlphaOfLength(8), UUIDs.randomBase64UUID(), randomInt(5)), 8)),
+            randomList(8, this::randomNodeDecision),
```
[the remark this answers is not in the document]
arteam: I think we already include empty lists? `randomList` generates a list of the size between 0 and 9 inclusive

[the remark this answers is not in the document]
fcofdez: ah right 🤦 , I always use the method with explicit bounds and I got confused

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderReasonWireSerializationTests.java
arteam: I think it's needed for the case where we have both random unassigned and assigned shard ids

server/src/main/java/org/elasticsearch/cluster/routing/allocation/NodeDecision.java
arteam: Done!

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
arteam: You're totally right, I've changed the check, so we have one return statement with the disk only node decisions

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -274,19 +285,27 @@ public static class AllocationState {
...
+                .flatMap(shard -> cannotAllocateDueToStorage(shard, allocation).stream())
                 .toList();
-            return new ShardsSize(
-                unassignedShards.stream().mapToLong(this::sizeOf).sum(),
-                unassignedShards.stream().map(ShardRouting::shardId).collect(Collectors.toCollection(TreeSet::new))
+            return new ShardsAllocationResults(
+                unassignedShards.stream().map(e -> e.shard).mapToLong(this::sizeOf).sum(),
+                unassignedShards.stream().map(e -> e.shard.shardId()).collect(Collectors.toCollection(TreeSet::new)),
+                unassignedShards.stream()
+                    .filter(e -> e.nodeDecisions.size() > 0)
```
fcofdez, line 300: nit: can we use a more descriptive variable name here?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
fcofdez: I think this was resolved by mistake?

fcofdez (approved):
This looks good to me 👍 , I left a couple of nits. Thanks for all the iterations!. I think we should wait for @henningandersen +1.

arteam:

henningandersen (commented):
Left more comments.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
```diff
@@ -349,25 +386,40 @@ public void testStoragePreventsRemain() {
...
         verify(
             ReactiveStorageDeciderService.AllocationState::storagePreventsRemainOrMove,
-            new ShardsSize(nodes, shardIds),
+            0,
+            emptySortedSet(),
+            List::isEmpty,
+            List::isEmpty
+        );
+
+        // only consider it once (move case) if both cannot remain and cannot allocate.
```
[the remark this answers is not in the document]
arteam: Right! That's what we verify here. `canAllocateCheck` checks that we can't allocate on the other hot nodes, and the `canRemain` check checks verifies that we can't remain. I've added the similar checks for `verifyScale` where we check that these decisions get correctly concatenated as `assignedNodeDecisions`

server/src/main/java/org/elasticsearch/cluster/routing/allocation/NodeDecision.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.cluster.routing.allocation;
```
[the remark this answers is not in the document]
arteam: Done, moved to `autoscaling.storage`

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -162,7 +165,15 @@ public AutoscalingDeciderResult scale(Settings configuration, AutoscalingDecider
                 unassignedBytes,
                 unassignedBytesUnassignedShards.shardIds(),
                 assignedBytes,
-                assignedBytesUnmovableShards.shardIds()
+                assignedBytesUnmovableShards.shardIds(),
+                CollectionUtils.concatLists(
+                    unassignedBytesUnassignedShards.canAllocateDecisions(),
+                    assignedBytesUnmovableShards.canAllocateDecisions()
```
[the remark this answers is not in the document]
arteam: Right, I mixed up the semantics of the assigned and unassigned node decisions and `can_allocate` and `can_remain` decisions. Now  we concat `can_allocate` and `can_remain` for `storagePreventsAllocation` and `storagePreventsRemainOrMove` for unassigned and assigned node decisions respectively

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -357,24 +417,28 @@ private boolean allocatedToTier(ShardRouting s, RoutingAllocation allocation) {
...
+                    node -> new NodeDecision(node.node(), allocationDeciders.canAllocate(shard, node, allocation))
+                ).filter(nar -> ReactiveStorageDeciderService.isDiskOnlyNoDecision(nar.decision())).toList();
+                if (diskOnly.size() > 0 && shard.unassigned() && shard.recoverySource().getType() == RecoverySource.Type.LOCAL_SHARDS) {
                     // For resize shards only allow autoscaling if there is no other node where the shard could fit had it not been
                     // a resize shard. Notice that we already removed any initial_recovery filters.
-                    diskOnly = nodesInTier(allocation.routingNodes()).map(node -> allocationDeciders.canAllocate(shard, node, allocation))
-                        .anyMatch(ReactiveStorageDeciderService::isResizeOnlyNoDecision) == false;
+                    boolean hasResizeOnly = nodesInTier(allocation.routingNodes()).map(
+                        node -> new NodeDecision(node.node(), allocationDeciders.canAllocate(shard, node, allocation))
+                    ).anyMatch(nar -> ReactiveStorageDeciderService.isResizeOnlyNoDecision(nar.decision()));
```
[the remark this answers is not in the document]
arteam: Good catch!

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderDecisionTests.java
```diff
@@ -377,26 +429,66 @@ public void testStoragePreventsRemain() {
...
     }
 
-    private static ShardsSize emptyShardsSize() {
-        return new ShardsSize(0, Collections.emptySortedSet());
+    private static Decision singleNoDecision(NodeDecision nodeAllocationResult) {
+        return nodeAllocationResult.decision()
+            .getDecisions()
+            .stream()
+            .filter(d -> d.type() == Decision.Type.NO)
+            .findFirst()
```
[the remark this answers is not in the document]
arteam: I've added checks that we only have one single decisions.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -917,6 +1009,8 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
             builder.field("assigned", assigned);
             builder.field("assigned_shards", assignedShardIds.stream().limit(MAX_AMOUNT_OF_SHARDS).toList());
             builder.field("assigned_shards_count", assignedShardIds.size());
+            builder.xContentList("unassigned_node_decisions", unassignedNodeDecisions);
+            builder.xContentList("assigned_node_decisions", assignedNodeDecisions);
```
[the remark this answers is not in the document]
arteam: I've added a new constant `MAX_AMOUNT_OF_DECISIONS = 512` to control the amount of decisions for a shard.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ShardsAllocationResults.java
```diff
@@ -9,6 +9,12 @@
...
 
+import java.util.List;
 import java.util.SortedSet;
 
-record ShardsSize(long sizeInBytes, SortedSet<ShardId> shardIds) {}
+record ShardsAllocationResults(
+    long sizeInBytes,
+    SortedSet<ShardId> shardIds,
+    List<NodeDecision> canAllocateDecisions,
+    List<NodeDecision> canRemainDecisions
```
henningandersen, line 19: I wonder if we need to make this a `Map<ShardId, record<List<NodeDecision>, List<NodeDecision>>>` instead. That would make it possible to limit the number of shards to list to some count and to limit the number of decisions to list to a subset of those shards.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -831,30 +896,38 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
     public static class ReactiveReason implements AutoscalingDeciderResult.Reason {
 
         static final int MAX_AMOUNT_OF_SHARDS = 512;
+        static final int MAX_AMOUNT_OF_DECISIONS = 512;
```
henningandersen, line 899: This means we get a list of decisions that is

1. Fairly large.
2. Not covering shards in total, i.e., the last shard will have some decisions.
3. Does not include the shard-id in the decision.

I think we should move to a Map of shard-id to decision list(s). And then limit the number of shards to explain to something like 5.

henningandersen (commented):
The way the list is limited made me think deeper about the output. I think we need an output that makes it clear, which decision is for which shard rather than a flat list of decisions.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ShardsAllocationResults.java
arteam: I've created a new class `NodeDecisions` that holds `canAllocateDecisions` and `canRemainDecisions`. `ShardsAllocationResults` now holds these decisions by shard id as `Map<ShardId, NodeDecisions> shardNodeDecisions`.
This map gets stored as `unassignedNodeDecisions` and `assignedNodeDecisions` and then gets limited in the JSON output

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -274,19 +278,24 @@ public static class AllocationState {
...
+                .flatMap(shard -> cannotAllocateDueToStorage(shard, allocation).stream())
                 .toList();
-            return new ShardsSize(
-                unassignedShards.stream().mapToLong(this::sizeOf).sum(),
-                unassignedShards.stream().map(ShardRouting::shardId).collect(Collectors.toCollection(TreeSet::new))
+            return new ShardsAllocationResults(
+                unassignedShards.stream().map(e -> e.shard).mapToLong(this::sizeOf).sum(),
+                unassignedShards.stream().map(e -> e.shard.shardId()).collect(Collectors.toCollection(TreeSet::new)),
+                unassignedShards.stream()
+                    .filter(shardNodeDecisions -> shardNodeDecisions.nodeDecisions.size() > 0)
```
henningandersen, line 293: Can we impose the limit here rather than later. I am slightly (but not deeply) worried about the amounts of data we gather here and this could ease it slightly.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -301,28 +310,77 @@ && canAllocate(shard, allocation) == false) {
...
+            List<ShardNodeDecisions> otherNodesOnTierAllocationDecisions = unmovableShardNodeDecisions.stream()
+                .map(ShardNodeDecisions::shard)
+                .map(
+                    shard -> new ShardNodeDecisions(
+                        shard,
+                        state.getRoutingNodes()
+                            .stream()
+                            .filter(n -> nodeTierPredicate.test(n.node()))
+                            .filter(n -> shard.currentNodeId().equals(n.nodeId()) == false)
+                            .map(n -> new NodeDecision(n.node(), allocationDeciders.canAllocate(shard, n, allocation)))
```
henningandersen, line 326: Can we apply the limit when collecting these, both in order to limit memory use but also not calculate these decisions unnecesarily? We might need to collect both the cannot remain reason and the can allocate decision here then but I think that is fine?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -301,28 +310,77 @@ && canAllocate(shard, allocation) == false) {
...
-            List<ShardRouting> unallocatedShards = candidates.stream()
+            List<ShardNodeDecisions> unallocatedShardNodeDecisions = candidates.stream()
                 .filter(Predicate.not(unmovableShards::contains))
-                .filter(s1 -> cannotAllocateDueToStorage(s1, allocation))
+                .flatMap(shard -> cannotAllocateDueToStorage(shard, allocation).stream())
                 .toList();
-            long unallocatableBytes = unallocatedShards.stream().mapToLong(this::sizeOf).sum();
+            List<ShardNodeDecisions> canRemainDecisionsForUnallocatedShards = unallocatedShardNodeDecisions.stream()
+                .filter(snd -> snd.nodeDecisions.size() > 0)
+                .map(ShardNodeDecisions::shard)
```
henningandersen, line 346: Can we apply the limit when collecting these, both in order to limit memory use but also not calculate these decisions unnecesarily?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -301,28 +310,77 @@ && canAllocate(shard, allocation) == false) {
...
                 unallocatableBytes + unmovableBytes,
-                Stream.concat(unmovableShards.stream(), unallocatedShards.stream())
-                    .map(ShardRouting::shardId)
-                    .collect(Collectors.toCollection(TreeSet::new))
+                Stream.concat(unmovableShardNodeDecisions.stream(), unallocatedShardNodeDecisions.stream())
+                    .map(e -> e.shard.shardId())
+                    .collect(Collectors.toCollection(TreeSet::new)),
+                Stream.concat(unmovableShardNodeDecisions.stream(), canRemainDecisionsForUnallocatedShards.stream())
+                    .filter(shardNodeDecisions -> shardNodeDecisions.nodeDecisions.size() > 0)
+                    .collect(
```
henningandersen, line 374: Can we apply the limit when collecting these too?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -301,28 +310,77 @@ && canAllocate(shard, allocation) == false) {
...
+                .collect(toMap(snd -> snd.shard.shardId(), snd -> snd.nodeDecisions, (a, b) -> b));
+            return new ShardsAllocationResults(
                 unallocatableBytes + unmovableBytes,
-                Stream.concat(unmovableShards.stream(), unallocatedShards.stream())
-                    .map(ShardRouting::shardId)
-                    .collect(Collectors.toCollection(TreeSet::new))
+                Stream.concat(unmovableShardNodeDecisions.stream(), unallocatedShardNodeDecisions.stream())
+                    .map(e -> e.shard.shardId())
+                    .collect(Collectors.toCollection(TreeSet::new)),
+                Stream.concat(unmovableShardNodeDecisions.stream(), canRemainDecisionsForUnallocatedShards.stream())
```
henningandersen, line 372: I think this can be simplified if we let `canRemainDecisionsForUnallocatedShards` and `otherNodesOnTierAllocationDecisions` contain both the remain and allocate decision, might need to convert to map directly. We can then combine the maps here at the end.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -301,28 +310,77 @@ && canAllocate(shard, allocation) == false) {
...
+            List<ShardNodeDecisions> canRemainDecisionsForUnallocatedShards = unallocatedShardNodeDecisions.stream()
+                .filter(snd -> snd.nodeDecisions.size() > 0)
+                .map(ShardNodeDecisions::shard)
+                .map(
+                    shard -> new ShardNodeDecisions(
+                        shard,
+                        List.of(
+                            new NodeDecision(
+                                allocation.routingNodes().node(shard.currentNodeId()).node(),
+                                allocationDeciders.canRemain(shard, allocation.routingNodes().node(shard.currentNodeId()), allocation)
```
henningandersen, line 353: I think we need to enable debug decisions when calculating these. Same for the `canAllocate` part above.

Also, we should add a test that verifies that we have debug information included.

henningandersen (commented):
I ran out of time but thought it would be useful if you looked at this set of comments.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -357,24 +415,28 @@ private boolean allocatedToTier(ShardRouting s, RoutingAllocation allocation) {
...
-                        .anyMatch(ReactiveStorageDeciderService::isResizeOnlyNoDecision) == false;
+                    boolean hasResizeOnly = nodesInTier(allocation.routingNodes()).map(
+                        node -> allocationDeciders.canAllocate(shard, node, allocation)
+                    ).anyMatch(ReactiveStorageDeciderService::isResizeOnlyNoDecision);
+                    if (hasResizeOnly) {
+                        return Optional.empty();
+                    }
                 }
-                return diskOnly;
+                return diskOnly.size() > 0 ? Optional.of(new ShardNodeDecisions(shard, diskOnly)) : Optional.empty();
```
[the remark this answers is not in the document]
arteam: That's a good point. I've added all node decisions to the output.

arteam:

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
arteam: I don't think that this can be simplified, unfortunately. `unallocatedShardNodeDecisions` and `unmovableShardNodeDecisions` need to be separate collections for the calculation of the allocated and unallocated bytes.
I also didn't understand the part about `both the remain and allocate decision`. I thought the point was that we track them separately in different collections.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
arteam: Done! I've added a helper method `withAllocationDebugEnabled`, so we can turn the debug decision on and off and keep the calculation in the short lambda form.

arteam:
@elasticmachine update branch

arteam:
@elasticmachine update branch

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -357,24 +456,31 @@ private boolean allocatedToTier(ShardRouting s, RoutingAllocation allocation) {
          * Check that disk decider is only decider for a node preventing allocation of the shard.
          * @return true if and only if a node exists in the tier where only disk decider prevents allocation
```
henningandersen, line 457: Can we also update the javadoc here to reflect the new return value?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
```diff
@@ -301,31 +322,109 @@ && canAllocate(shard, allocation) == false) {
...
 
-            List<ShardRouting> unallocatedShards = candidates.stream()
+            List<ShardNodeDecisions> unallocatedShardNodeDecisions = candidates.stream()
                 .filter(Predicate.not(unmovableShards::contains))
-                .filter(s1 -> cannotAllocateDueToStorage(s1, allocation))
+                .flatMap(shard -> cannotAllocateDueToStorage(shard, allocation).stream())
                 .toList();
-            long unallocatableBytes = unallocatedShards.stream().mapToLong(this::sizeOf).sum();
+            List<ShardNodeDecisions> canRemainDecisionsForUnallocatedShards = unallocatedShardNodeDecisions.stream()
+                .filter(snd -> snd.nodeDecisions.size() > 0)
```
henningandersen, line 362: I think we should not filter this case out. This is the case where the target tier does not yet exist. And by filtering it out, we also omit the can-remain reason from the output. Instead we should keep it and accept that the list is empty (we can then find out by reading code that it has this special meaning).

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
henningandersen: This patch 
```
Index: x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java b/x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java
--- a/x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java	(revision 740451bc3c1615fd5e627f2f633298c43d0103b5)
+++ b/x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderService.java	(date 1667557190920)
@@ -286,7 +286,11 @@
             this.roles = roles;
         }
 
-        record ShardNodeDecisions(ShardRouting shard, List<NodeDecision> nodeDecisions) {}
+        record ShardNodeDecisions(ShardRouting shard, List<NodeDecision> nodeDecisions) {
+            static ShardNodeDecisions preferPrimary(ShardNodeDecisions snd1, ShardNodeDecisions snd2) {
+                return snd1.shard().primary() ? snd1 : snd2;
+            }
+        }
 
         public ShardsAllocationResults storagePreventsAllocation() {
             RoutingAllocation allocation = new RoutingAllocation(allocationDeciders, state, info, shardSizeInfo, System.nanoTime());
@@ -326,25 +330,17 @@
                 .filter(s -> allocatedToTier(s, allocation))
                 .flatMap(shard -> cannotRemainDueToStorage(shard, allocation).stream())
                 .toList();
-            List<ShardNodeDecisions> otherNodesOnTierAllocationDecisions = unmovableShardNodeDecisions.stream()
-                .map(ShardNodeDecisions::shard)
-                .map(
-                    shard -> new ShardNodeDecisions(
-                        shard,
-                        state.getRoutingNodes()
-                            .stream()
-                            .filter(n -> nodeTierPredicate.test(n.node()))
-                            .filter(n -> shard.currentNodeId().equals(n.nodeId()) == false)
-                            .map(
-                                n -> new NodeDecision(
-                                    n.node(),
-                                    withAllocationDebugEnabled(allocation, () -> allocationDeciders.canAllocate(shard, n, allocation))
-                                )
-                            .toList()
+            Map<ShardId, ShardNodeDecisions> perShardIdUnmovable = unmovableShardNodeDecisions.stream()
+                .collect(toMap(snd -> snd.shard().shardId(), Function.identity(), ShardNodeDecisions::preferPrimary, TreeMap::new));
+            Map<ShardId, NodeDecisions> otherNodesOnTierAllocationDecisions = perShardIdUnmovable.values()
+                .stream()
+                .limit(MAX_AMOUNT_OF_SHARD_DECISIONS)
+                .collect(
+                    toMap(
+                        snd -> snd.shard().shardId(),
+                        snd -> new NodeDecisions(canAllocateDecisions(allocation, snd.shard()), snd.nodeDecisions())
                     )
-                )
-                .toList();
+                );
             Set<ShardRouting> unmovableShards = unmovableShardNodeDecisions.stream().map(e -> e.shard).collect(Collectors.toSet());
 
             long unmovableBytes = unmovableShards.stream()
@@ -358,53 +354,60 @@
                 .filter(Predicate.not(unmovableShards::contains))
                 .flatMap(shard -> cannotAllocateDueToStorage(shard, allocation).stream())
                 .toList();
-            List<ShardNodeDecisions> canRemainDecisionsForUnallocatedShards = unallocatedShardNodeDecisions.stream()
-                .filter(snd -> snd.nodeDecisions.size() > 0)
-                .map(ShardNodeDecisions::shard)
-                .map(
-                    shard -> new ShardNodeDecisions(
-                        shard,
-                        List.of(
-                            new NodeDecision(
-                                allocation.routingNodes().node(shard.currentNodeId()).node(),
-                                withAllocationDebugEnabled(
-                                    allocation,
-                                    () -> allocationDeciders.canRemain(
-                                        shard,
-                                        allocation.routingNodes().node(shard.currentNodeId()),
-                                        allocation
-                                    )
-                                )
-                        )
-                    )
-                )
-                .toList();
+            Map<ShardId, ShardNodeDecisions> perShardIdUnallocated = unallocatedShardNodeDecisions.stream()
+                .collect(toMap(snd -> snd.shard().shardId(), Function.identity(), ShardNodeDecisions::preferPrimary, TreeMap::new));
+            Map<ShardId, NodeDecisions> canRemainDecisionsForUnallocatedShards = perShardIdUnallocated.values()
+                .stream()
+                .limit(MAX_AMOUNT_OF_SHARD_DECISIONS)
+                .collect(
+                    toMap(
+                        snd -> snd.shard().shardId(),
+                        snd -> new NodeDecisions(snd.nodeDecisions(), canRemainDecisions(allocation, snd.shard()))
+                    )
+                );
             long unallocatableBytes = unallocatedShardNodeDecisions.stream().map(e -> e.shard).mapToLong(this::sizeOf).sum();
 
-            Map<ShardId, List<NodeDecision>> shardAllocateDecisions = Stream.concat(
-                unallocatedShardNodeDecisions.stream(),
-                otherNodesOnTierAllocationDecisions.stream()
+            Map<ShardId, NodeDecisions> shardDecisions = Stream.concat(
+                canRemainDecisionsForUnallocatedShards.entrySet().stream(),
+                otherNodesOnTierAllocationDecisions.entrySet().stream()
             )
-                .filter(shardNodeDecisions -> shardNodeDecisions.nodeDecisions.size() > 0)
-                .collect(toMap(snd -> snd.shard.shardId(), snd -> snd.nodeDecisions, (a, b) -> b));
+                .collect(toSortedMap(Map.Entry::getKey, Map.Entry::getValue))
+                .entrySet()
+                .stream()
+                .limit(MAX_AMOUNT_OF_SHARD_DECISIONS)
+                .collect(toSortedMap(Map.Entry::getKey, Map.Entry::getValue));
             return new ShardsAllocationResults(
                 unallocatableBytes + unmovableBytes,
                 Stream.concat(unmovableShardNodeDecisions.stream(), unallocatedShardNodeDecisions.stream())
                     .map(e -> e.shard.shardId())
                     .collect(Collectors.toCollection(TreeSet::new)),
-                Stream.concat(unmovableShardNodeDecisions.stream(), canRemainDecisionsForUnallocatedShards.stream())
-                    .filter(shardNodeDecisions -> shardNodeDecisions.nodeDecisions.size() > 0)
-                    .collect(
-                        toSortedMap(
-                            snd -> snd.shard.shardId(),
-                            snd -> new NodeDecisions(shardAllocateDecisions.getOrDefault(snd.shard.shardId(), List.of()), snd.nodeDecisions)
-                        )
-                    )
-                    .entrySet()
-                    .stream()
-                    .limit(MAX_AMOUNT_OF_SHARD_DECISIONS)
-                    .collect(toSortedMap(Map.Entry::getKey, Map.Entry::getValue))
+                shardDecisions
+            );
+        }
+
+        private List<NodeDecision> canAllocateDecisions(RoutingAllocation allocation, ShardRouting shardRouting) {
+            return state.getRoutingNodes()
+                .stream()
+                .filter(n -> nodeTierPredicate.test(n.node()))
+                .filter(n -> shardRouting.currentNodeId().equals(n.nodeId()) == false)
+                .map(
+                    n -> new NodeDecision(
+                        n.node(),
+                        withAllocationDebugEnabled(allocation, () -> allocationDeciders.canAllocate(shardRouting, n, allocation))
+                    )
+                )
+                .toList();
+        }
+
+        private List<NodeDecision> canRemainDecisions(RoutingAllocation allocation, ShardRouting shard) {
+            return List.of(
+                new NodeDecision(
+                    allocation.routingNodes().node(shard.currentNodeId()).node(),
+                    withAllocationDebugEnabled(
+                        allocation,
+                        () -> allocationDeciders.canRemain(shard, allocation.routingNodes().node(shard.currentNodeId()), allocation)
+                    )
+                )
             );
         } 
```

demonstrates what I mean. First of all, we can limit the number of decisions we collect earlier. I also ensured we prefer primary over replica decisions. And removed the filter for empty node decisions list (new tier, see other comment). The latter requires a minor change to `testMoveToEmpty`.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/NodeDecisions.java
```diff
@@ -0,0 +1,70 @@
...
+    NodeDecisions(StreamInput in) throws IOException {
+        canAllocateDecisions = in.readList(NodeDecision::new);
+        canRemainDecisions = in.readList(NodeDecision::new);
+    }
+
+    List<NodeDecision> canAllocateDecisions() {
+        return canAllocateDecisions;
+    }
+
+    List<NodeDecision> canRemainDecisions() {
```
henningandersen, line 38: Does this have to be a list? I would imagine that there is only a decision for one node, the current node of the shard. I did not try to check this though, perhaps you can look into that.

henningandersen (commented):
Left more comments and a simplification suggestion.

arteam:
@elasticmachine update branch

arteam:
@elasticmachine update branch

arteam:
@elasticmachine update branch

arteam:

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/NodeDecisions.java
arteam: Yes, from the tests I can see that it we can have multiple `can_remain` decisions.

henningandersen: Can you explain how those multiple can remain decisions, i.e., for multiple nodes, comes about? Can you perhaps point me to how I can verify this easily using the test you mention here?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/NodeDecision.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    public int hashCode() {
+        return Objects.hash(node, decision);
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+        builder.startObject();
+        discoveryNodeToXContent(node, false, builder);
+        builder.field("node_decision", decision);
```
henningandersen, line 67: I think there is an array missing here, like in `NodeAllocationResult.toXContent`. I wonder if we can write a test verifying that toXContent works when the decision is a `Multi` decision.

arteam: Good catch! That's indeed should be a list. I've changed it to use `deciders` similar to `NodeAllocationResult`

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/NodeDecisions.java
arteam: @henningandersen I was wrong! The tests that produced multiple `can_remain` decisions were serialization tests with randomized input. There we no cases with multiple `can_remain` decisions in any integration tests. I've changed `NodeDecisions` to store a single instance of `canRemainDecision` and adjusted `ReactiveStorageDeciderService` to yield a single `canRemainDecision` or null instead of a singleton/empty list.

arteam:
@elasticmachine update branch

arteam:
@elasticmachine update branch

arteam:
@elasticmachine update branch

arteam:
@elasticmachine update branch

arteam:
@elasticmachine update branch

arteam:
@elasticmachine update branch

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveReasonTests.java
```diff
@@ -37,12 +43,34 @@ public void testXContent() throws IOException {
         String indexName = randomAlphaOfLength(10);
         SortedSet<ShardId> unassignedShardIds = new TreeSet<>(randomUnique(() -> new ShardId(indexName, indexUUID, randomInt(1000)), 600));
         SortedSet<ShardId> assignedShardIds = new TreeSet<>(randomUnique(() -> new ShardId(indexName, indexUUID, randomInt(1000)), 600));
+        DiscoveryNode discoveryNode = new DiscoveryNode("node1", buildNewFakeTransportAddress(), emptyMap(), emptySet(), Version.CURRENT);
+        Map<ShardId, NodeDecisions> unassignedShardAllocationDecision = Map.of(
+            unassignedShardIds.first(),
+            new NodeDecisions(
+                List.of(new NodeDecision(discoveryNode, Decision.single(Decision.Type.NO, "no_label", "No space to allocate"))),
```
henningandersen, line 50: Can we add both single and multi decision variants to be sure they both serialize correctly?

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveReasonTests.java
```diff
@@ -77,6 +105,55 @@ public void testXContent() throws IOException {
             );
             assertSorted(xContentAssignedShardIds.stream().map(ShardId::fromString).toList());
             assertEquals(assignedShardIds.size(), map.get("assigned_shards_count"));
+
+            Map<String, Object> unassignedNodeDecisions = (Map<String, Object>) ((Map<String, Object>) map.get("unassigned_node_decisions"))
+                .get(unassignedShardIds.first().toString());
+            for (Map<String, Object> canAllocateDecision : (List<Map<String, Object>>) unassignedNodeDecisions.get(
+                "can_allocate_decisions"
+            )) {
+                assertEquals("node1", canAllocateDecision.get("node_id"));
```
henningandersen, line 114: Thanks, the validation in this test case makes it easy to reason about the structure of the result.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveReasonTests.java
```diff
@@ -77,6 +105,55 @@ public void testXContent() throws IOException {
             );
             assertSorted(xContentAssignedShardIds.stream().map(ShardId::fromString).toList());
             assertEquals(assignedShardIds.size(), map.get("assigned_shards_count"));
+
+            Map<String, Object> unassignedNodeDecisions = (Map<String, Object>) ((Map<String, Object>) map.get("unassigned_node_decisions"))
+                .get(unassignedShardIds.first().toString());
+            for (Map<String, Object> canAllocateDecision : (List<Map<String, Object>>) unassignedNodeDecisions.get(
```
henningandersen, line 111: Can we check that the size of the list we loop over here. Just to avoid that the test passes if it is empty.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/storage/NodeDecisions.java
```diff
@@ -0,0 +1,72 @@
...
+    public void writeTo(StreamOutput out) throws IOException {
+        out.writeList(canAllocateDecisions);
+        out.writeOptionalWriteable(canRemainDecision);
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+        builder.startObject();
+        builder.xContentList("can_allocate_decisions", canAllocateDecisions);
+        builder.field("can_remain_decision", canRemainDecision);
```
henningandersen, line 55: Can we omit this if it is null?

In particular for the unassigned list, there will be no `can_remain_decision` and hence it may seem confusing and unnecessary to output it here.

henningandersen (approved):
LGTM.

Left 4 smaller things to address, but no need for another round unless bigger changes are mad.

arteam:
Thank you both!

PR #88592, opened 2022-07-18
Add variable shadowing to Painless

jdconrad:
This adds variable shadowing to Painless where a variable with the same name in a parent block may be re-declared as a different type and different value in a sub block.

WIP as this needs extensive testing added.


jdconrad:
Closing this as we have decided to encapsulate op for update/ingest contexts as part of metadata instead of a top-level variable on a script.

PR #88593, opened 2022-07-18
Reduce boilerplate in plugins utils tests

rjernst:
Tests for PluginsUtils need to create many different plugin descriptors.
But the vast majority of these just need a plugin name and dependencies.
This commit adds a utility method to simplify these tests.

### DIFF
--- a/server/src/test/java/org/elasticsearch/plugins/PluginsUtilsTests.java
+++ b/server/src/test/java/org/elasticsearch/plugins/PluginsUtilsTests.java
@@ -21,7 +21,6 @@ import java.io.InputStream;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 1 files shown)


jdconrad (approved):
LGTM

server/src/test/java/org/elasticsearch/plugins/PluginsUtilsTests.java
```diff
@@ -38,6 +37,21 @@
 @LuceneTestCase.SuppressFileSystems(value = "ExtrasFS")
 public class PluginsUtilsTests extends ESTestCase {
 
+    PluginDescriptor newTestDescriptor(String name, List<String> deps) {
```
stu-elastic, line 40: Maybe make deps variadic to avoid list creation by caller?

stu-elastic (approved):
lgtm, minor suggestion.

server/src/test/java/org/elasticsearch/plugins/PluginsUtilsTests.java
rjernst: We have to create a List for the descriptor regardless.

PR #88594, opened 2022-07-18
Better handling for missing bind_dn for AD impersonate

### LINKED ISSUE #81137: Run as / impersonate should give better error when used in conjunction with an Active Directory realm
**Elasticsearch version** (`bin/elasticsearch --version`): 7.15.0

**Plugins installed**: [none]

**JVM version** (`java -version`): bundled

**OS version** (`uname -a` if on a Unix-like system): CentOS 7

The error displayed (under Step 3), is not helpful. The [Active Directory Realm config](https://www.elastic.co/guide/en/elasticsearch/reference/current/active-directory-realm.html) states that

> The use of a bind user enables the run as feature to be used with the Active Directory realm and the ability to maintain a set of pooled connections to Active Directory. These pooled connection reduce the number of resources that must be created and destroyed with every user authentication.

The error message should reflect that statement.

**Steps to reproduce**:

 1. Configure an Active Directory Realm without a bind_dn
 2. Run the impersonate API against the `_security/_authenticate` API. using a curl similar to `curl -H "es-security-runas-user: active-directory-username" -u elastic .../_security/_authenticate`
 3. Now you should see this error
```
"action [cluster:admin/xpack/security/user/authenticate] is unauthorized for user [elastic] run as [active-directory-username] with roles [], this action is granted by the cluster privileges [manage_security,all]"
```

1. Add the bind_dn
2. Run the same command as above
3. Now you see the groups, roles and everything listed.

jakelandis:
This commit allows the authentication to stop processing if a setting is not 
compatible with the given workflow for LDAP/AD realms. Specifically, the
authentication can now be stopped when a bind_dn user is not configured 
for the AD realm when doing impersonation. 

Normally, we would not want to halt authentication for the in-ability to authenticate (or lookup)
to a realm. However, this is a special case such that if it were possible we would block
startup due to a mis configuration we would. However, in this case the configuration is correct
except for when you are impersonating which is only known at runtime. Without this 
change the authentication is succesful, lookup fails, and the user see the result of the 
authz (is unauthorized to run as) which is not very helpful nor correct error message. 

With this change a more specific and actionable error is returned. This does 
however slightly change the contract such that authentication is halted early.
I believe this is fine if used sparingly. The usage in this commit would only impact
a scenario where a user had 2 or more AD realms and was trying to authenticate 
to one and impersonate (or delegate authz) from another and only had the bind_dn 
setup for one of the AD realms. I think this would be a very rare case and a more 
meaningful error message should be preferred. 

closes #81137

### DIFF
--- a/docs/changelog/88594.yaml
+++ b/docs/changelog/88594.yaml
@@ -1,0 +1,5 @@
+pr: 88594
+summary: Better handling for missing `bind_dn` for AD impersonate
+area: Authentication
+type: bug
+issues: []
--- a/docs/changelog/88594.yaml
+++ b/docs/changelog/88594.yaml
@@ -2,4 +2,5 @@ pr: 88594
 summary: Better handling for missing `bind_dn` for AD impersonate
 area: Authentication
 type: bug
-issues: []
+issues:
+ - 81137
--- a/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ldap/ActiveDirectorySessionFactory.java
+++ b/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ldap/ActiveDirectorySessionFactory.java
@@ -26,6 +26,7 @@ import org.elasticsearch.common.logging.DeprecationCategory;
 import org.elasticsearch.common.logging.DeprecationLogger;
 import org.elasticsearch.common.settings.SecureString;
 import org.elasticsearch.common.settings.Setting;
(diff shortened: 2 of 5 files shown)


jakelandis:
Closing as after a bit of discussion this is actually a breaking change, multiple AD's with impersonate may not be that rare, and it also possible that could affect more than just multiple AD's configured but any delegated Authz lookup where AD is configured but not currently used (requiring adding a bind_dn for cases where it is not used).

PR #88603, opened 2022-07-19
Enable synthetic source support on constant keyword fields

romseygeek:
This commit implements synthetic source support on constant keyword fields,
which will now always emit their value as part of the retrieved source.


jpountz (commented):
I'm not qualified to do a full review, but the change makes sense to me and properly tests the corner case when the constant value has not been configured yet, so this is good as far as I can tell.

csoulios (approved):
LGTM

Nice improvement!

PR #88611, opened 2022-07-19
Warn if using reroute with listener that ignores the result

idegtiarenko:


DaveCTurner (commented):
Not sure what this is for - don't we already know all the places we use this listener by just looking at the code?

DaveCTurner (approved):
Ok, LGTM, we'll be removing this before the end anyway.

PR #88616, opened 2022-07-19
[ML] fix ml API cancellation when http connection is closed

benwtrent:
API cancellation was erroneously using the node-name for the parent task id. Task ids are constructed via the node ID, not name. 

This fixes that bug.

related to: 
 - https://github.com/elastic/elasticsearch/pull/88030
 - https://github.com/elastic/elasticsearch/pull/88142
 - https://github.com/elastic/elasticsearch/pull/88009


droberts195 (approved):
LGTM

PR #88621, opened 2022-07-19
Make NodeTaskRequest tasks cancellable

benwtrent:
We pass the actionTask to the subsequent task action so that the task action can be cancelled if the upstream tasks are cancelled.

But, the actionTask passed is created by the `NodeTaskRequest`, which isn't cancellable. 

This commit addresses this discrepancy and allows for down stream task actions to be cancelled if the upstream http connections are closed.

Related to: #88081

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/support/tasks/TransportTasksAction.java
+++ b/server/src/main/java/org/elasticsearch/action/support/tasks/TransportTasksAction.java
@@ -24,7 +24,9 @@ import org.elasticsearch.common.io.stream.StreamOutput;
 import org.elasticsearch.common.io.stream.Writeable;
 import org.elasticsearch.common.util.concurrent.AtomicArray;
 import org.elasticsearch.core.Tuple;
+import org.elasticsearch.tasks.CancellableTask;
 import org.elasticsearch.tasks.Task;
+import org.elasticsearch.tasks.TaskId;
 import org.elasticsearch.transport.TransportChannel;
 import org.elasticsearch.transport.TransportException;
 import org.elasticsearch.transport.TransportRequest;
@@ -307,6 +309,9 @@ public abstract class TransportTasksAction<
         }
 
         private void finishHim() {
+            if ((task instanceof CancellableTask t) && t.notifyIfCancelled(listener)) {
+                return;
+            }
             TasksResponse finalResponse;
             try {
                 finalResponse = newResponse(request, responses);
@@ -323,6 +328,9 @@ public abstract class TransportTasksAction<
 
         @Override
         public void messageReceived(final NodeTaskRequest request, final TransportChannel channel, Task task) throws Exception {
+            if (task instanceof CancellableTask cancellableTask) {
+                cancellableTask.ensureNotCancelled();
+            }
             nodeOperation(task, request, ActionListener.wrap(channel::sendResponse, e -> {
                 try {
                     channel.sendResponse(e);
@@ -335,7 +343,7 @@ public abstract class TransportTasksAction<
     }
 
     private class NodeTaskRequest extends TransportRequest {
-        private TasksRequest tasksRequest;
+        private final TasksRequest tasksRequest;
 
         protected NodeTaskRequest(StreamInput in) throws IOException {
             super(in);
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/elasticsearch/action/support/tasks/TransportTasksAction.java
```diff
@@ -323,6 +328,9 @@ class NodeTransportHandler implements TransportRequestHandler<NodeTaskRequest> {
 
         @Override
         public void messageReceived(final NodeTaskRequest request, final TransportChannel channel, Task task) throws Exception {
+            if (task instanceof CancellableTask cancellableTask) {
+                cancellableTask.ensureNotCancelled();
```
DaveCTurner, line 332: Is this needed? The transport layer will have just checked for cancellation before calling `messageReceived` I think.

DaveCTurner (commented):
I left one comment inline. Also could we have a test for this?

server/src/main/java/org/elasticsearch/action/support/tasks/TransportTasksAction.java
benwtrent: @DaveCTurner I am unfamiliar where this code is called within the transport layer. 

I copied this logic loosely based on NodeTransportHandler.

https://github.com/elastic/elasticsearch/blob/2f5ec518f349db7e446c0d4696666284bd983357/server/src/main/java/org/elasticsearch/action/support/nodes/TransportNodesAction.java#L327-L335

benwtrent: I will happily remove it. The big change I need for subsequent changes is that the task is cancellable, I thought it prudent to check here. But if its done for sure higher up the stack, then I am good with removing it

DaveCTurner: > I copied this logic loosely based on NodeTransportHandler.

`git blame` --> guilty as charged 🤡 

Yeah this is not necessary, we're called from `RequestHandlerRegistry#processMessageReceived` which just a few lines earlier obtained its task from `TaskManager#register`, so called `TaskManager#registerCancellableTask`, and this method would have thrown if the task were already cancelled.

benwtrent: @DaveCTurner I will remove it then :D thanks for the review!

server/src/main/java/org/elasticsearch/action/support/tasks/TransportTasksAction.java
```diff
@@ -323,6 +328,9 @@ class NodeTransportHandler implements TransportRequestHandler<NodeTaskRequest> {
...
+            }
```
benwtrent, line 333:
```suggestion
```

server/src/main/java/org/elasticsearch/action/support/tasks/TransportTasksAction.java
benwtrent: ok, removed

DaveCTurner (changes requested):
I still think this deserves some kind of test to verify that this is sufficient to make these actions cancellable.

benwtrent:
@DaveCTurner I will see if there is any testing framework that fits the bill. 

I have verified locally that it does (as long as the subsequent calls are cancellable or check for cancellation).

server/src/test/java/org/elasticsearch/action/admin/cluster/node/tasks/TransportTasksActionTests.java
```diff
@@ -620,6 +628,81 @@ private String getAllTaskDescriptions() {
...
+        CountDownLatch checkLatch = new CountDownLatch(1);
+        CountDownLatch taskLatch = new CountDownLatch(1);
+        ActionFuture<NodesResponse> future = startBlockingTestNodesAction(checkLatch);
+        int numNodes = 2;
+
+        CountDownLatch taskExecutesLatch = new CountDownLatch(numNodes);
+        TestTasksAction[] tasksActions = new TestTasksAction[numNodes];
+        for (int j = 0; j < numNodes; j++) {
+            final int nodeId = j;
+            // Simulate task action that fails on one of the tasks on one of the nodes
```
benwtrent, line 643:
```suggestion
```

benwtrent:
@DaveCTurner added a test that shows how the original action being cancelled propagates down to the node actions

DaveCTurner (approved):
Nice, LGTM

PR #88622, opened 2022-07-19
Use origin for the client when running _features/_reset

### LINKED ISSUE #88617: POST _features/reset doesn't use the system index origin
### Elasticsearch Version

8.3

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

any

### Problem Description

Calling POST _features/_reset should reset all system indices, provided the user calling the API has a 'superuser' role or at least cluster 'manage' role.

However, the code that handles calling the individual system index reset operations isn't properly passing the ORIGIN header value, therefore many of the system index reset operations are rejected. 

### Steps to Reproduce

Simplest way to reproduce is to add a security role of any kind, causing the security index to be created and then attempt to call POST _features/_reset with a superuser.

### Logs (if relevant)

_No response_

grcevski:
The current implementation of `POST /_features/_reset` doesn't pass the index descriptor origin when calling `DELETE` on the system indices, therefore, we fail to reset the respective indices.

Closes #88617

### DIFF
--- a/qa/system-indices/src/main/java/org/elasticsearch/system/indices/SystemIndicesQA.java
+++ b/qa/system-indices/src/main/java/org/elasticsearch/system/indices/SystemIndicesQA.java
@@ -38,6 +38,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.function.Supplier;
 
+import static org.elasticsearch.action.admin.cluster.node.tasks.get.GetTaskAction.TASKS_ORIGIN;
 import static org.elasticsearch.index.mapper.MapperService.SINGLE_MAPPING_NAME;
 import static org.elasticsearch.rest.RestRequest.Method.POST;
 import static org.elasticsearch.rest.RestRequest.Method.PUT;
@@ -73,7 +74,7 @@ public class SystemIndicesQA extends Plugin implements SystemIndexPlugin, Action
                         .put(IndexMetadata.SETTING_AUTO_EXPAND_REPLICAS, "0-1")
                         .build()
                 )
-                .setOrigin("net-new")
+                .setOrigin(TASKS_ORIGIN)
                 .setVersionMetaKey("version")
                 .setPrimaryIndex(".net-new-system-index-" + Version.CURRENT.major)
                 .build(),
@@ -81,7 +82,7 @@ public class SystemIndicesQA extends Plugin implements SystemIndexPlugin, Action
                 .setIndexPattern(INTERNAL_UNMANAGED_INDEX_NAME)
                 .setDescription("internal unmanaged system index")
                 .setType(SystemIndexDescriptor.Type.INTERNAL_UNMANAGED)
-                .setOrigin("qa")
+                .setOrigin(TASKS_ORIGIN)
                 .setAliasName(".internal-unmanaged-alias")
                 .build(),
             SystemIndexDescriptor.builder()
@@ -96,7 +97,7 @@ public class SystemIndicesQA extends Plugin implements SystemIndexPlugin, Action
                         .put(IndexMetadata.SETTING_AUTO_EXPAND_REPLICAS, "0-1")
                         .build()
                 )
-                .setOrigin("qa")
+                .setOrigin(TASKS_ORIGIN)
                 .setVersionMetaKey("version")
                 .setPrimaryIndex(".internal-managed-index-" + Version.CURRENT.major)
                 .setAliasName(".internal-managed-alias")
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/elasticsearch/indices/SystemIndices.java
```diff
@@ -864,39 +903,61 @@ public MigrationCompletionHandler getPostMigrationFunction() {
...
-                @Override
-                public void onFailure(Exception e) {
-                    listener.onResponse(ResetFeatureStateStatus.failure(name, e));
+                    if (errors.isEmpty()) {
+                        listener.onResponse(ResetFeatureStateStatus.success(name));
+                    } else {
+                        StringBuilder exceptions = new StringBuilder("[");
+                        exceptions.append(errors.stream().map(e -> e.getException().getMessage()).collect(Collectors.joining(", ")));
+                        exceptions.append(']');
+                        listener.onResponse(ResetFeatureStateStatus.failure(name, new Exception(exceptions.toString())));
```
gwbrown, line 940: Could you also add logging that captures the full stack traces? It's nitpicky but it's no fun when we hit a weird error and realize there's no way to get the actual stack trace.

server/src/main/java/org/elasticsearch/indices/SystemIndices.java
```diff
@@ -328,6 +331,11 @@ public ExecutorSelector getExecutorSelector() {
      * @return The matching {@link SystemIndexDescriptor} or {@code null} if no descriptor is found
      */
     public @Nullable SystemIndexDescriptor findMatchingDescriptor(String name) {
+        return findMatchingDescriptor(indexDescriptors, name);
+    }
+
+    @Nullable
+    static SystemIndexDescriptor findMatchingDescriptor(SystemIndexDescriptor[] indexDescriptors, String name) {
```
gwbrown, line 338: I'm not objecting necessarily, but why break this out? I assumed it was for testing in my first pass given it's pkg-private, but there's no tests that use this version looks like.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/integration/SecurityFeatureResetTests.java
```diff
@@ -0,0 +1,141 @@
...
+import org.elasticsearch.xpack.core.security.user.User;
+import org.junit.Before;
+
+import java.util.Collections;
+
+import static org.elasticsearch.xpack.core.security.authc.support.UsernamePasswordToken.BASIC_AUTH_HEADER;
+import static org.elasticsearch.xpack.core.security.authc.support.UsernamePasswordToken.basicAuthHeaderValue;
+import static org.hamcrest.Matchers.containsString;
+
+public class SecurityFeatureResetTests extends SecurityIntegTestCase {
```
gwbrown, line 28:
```suggestion
/** 
 * These tests ensure that the Feature Reset API works for users with default superuser and manage roles.
 * This can be complex due to restrictions on system indices and the need to use the correct origin for
 * each index. See also https://github.com/elastic/elasticsearch/issues/88617
 */
public class SecurityFeatureResetTests extends SecurityIntegTestCase {
```

Can you add Javadoc to this class? I put a suggested version above.

AthenaEryma (commented):
Left a few small comments but this looks like a good approach. Do we know what in particular requires the per-descriptor origin?

server/src/main/java/org/elasticsearch/indices/SystemIndices.java
grcevski: You are right, no need for that to be separate.

server/src/main/java/org/elasticsearch/indices/SystemIndices.java
grcevski: Sure sounds good.

grcevski:
> Left a few small comments but this looks like a good approach. Do we know what in particular requires the per-descriptor origin?

I'm not sure if there's a plugin where there's not 1-1 mapping of feature to origin. I simply erred on the side of caution since the origin is on the descriptor, I decided to split everything.

qa/system-indices/src/main/java/org/elasticsearch/system/indices/SystemIndicesQA.java
```diff
@@ -73,15 +74,15 @@ public Collection<SystemIndexDescriptor> getSystemIndexDescriptors(Settings sett
                         .put(IndexMetadata.SETTING_AUTO_EXPAND_REPLICAS, "0-1")
                         .build()
                 )
-                .setOrigin("net-new")
```
grcevski, line 76: Fake origins don't work anymore, since we actually switch the security. I picked tasks since it's not in a module.

AthenaEryma (approved):
LGTM, thanks for iterating!

albertzaharovits (approved):
LGTM

PR #88623, opened 2022-07-19
[ML] disallow autoscaling downscaling in two trained model assignment scenarios

benwtrent:
With 8.4, trained model assignments now take CPU into consideration. This changes our calculation for scaling down until we fully support autoscaling according to CPU requirements.

1. We shouldn't allow scaling down if there is ANY model assignment that isn't fully allocated (meaning, not enough CPUs)
2. We don't allow scaling down unless model assignments require less than half of the current scale's CPU count.

Point 2 is a place holder. Fix 1 will be a requirement even in the future with vCPU autoscaling.


droberts195 (commented):
I have added the `cloud-deploy` label.  We should re-run @wwang500's test that found the autoscaling loop using the image created by that label when the minor nits I mentioned are added.

droberts195 (approved):
LGTM apart from a couple of nits, but please deploy the Cloud image before merging and leave it with autoscaling enabled overnight with some models similar to the ones @wwang500 ran to see if it works.

wwang500:
> I have added the `cloud-deploy` label. We should re-run @wwang500's test that found the autoscaling loop using the image created by that label when the minor nits I mentioned are added.

Hi @benwtrent and @droberts195 I have tried the autoscaling by using this PR's docker build. there is one strange thing,

when I tried starting models with autoscaling ON, the models failed in starting, which was expected in `starting` state. Later on, the scale up event was still triggered, however the models are still in fail state.

Screen Shot 2022-07-21 at 12 41 23 AM
Screen Shot 2022-07-21 at 9 48 33 AM
Screen Shot 2022-07-21 at 9 47 40 AM

PR #88626, opened 2022-07-19
Adding the ability to register a PeerFinderListener to Coordinator

masseyke:
This came out of #88562. The problem there is that if we are a non-master-eligible node we want to start polling a master-eligible node whenever we realize that there is no elected master. We are notified that there is no elected master in a `ClusterChangedEvent`. We get the list of master-eligible nodes from `PeerFinder#getFoundPeers`. However at that the time that we are notified of the `ClusterChangeEvent`, `PeerFinder#getFoundPeers` returns an empty `Iterable`. That collection is populated in another thread, and there is currently no way to get notified of when it is populated. This PR adds the ability to register a `PeerFinderListener` with `Coordinator`. The listener has an `onFoundPeersUpdated` that is called whenever the collection of peers changes (whether added to or removed from).

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/coordination/ClusterBootstrapService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/coordination/ClusterBootstrapService.java
@@ -147,6 +147,7 @@ public class ClusterBootstrapService implements Coordinator.PeerFinderListener {
         }
     }
 
+    @Override
     public void onFoundPeersUpdated() {
         final Set<DiscoveryNode> nodes = getDiscoveredNodes();
         if (bootstrappingPermitted.get()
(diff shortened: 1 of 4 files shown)


masseyke:
One question I have is whether this is granular enough. For example, when a master node steps down and a new one is elected, `onFoundPeersUpdated` is called 3 times in rapid succession:

1. When the master steps down, `onLeaderFailure` calls `activate` on the PeerFinder, putting the name(s) of the other nodes in there
2. When it gets a response from each of these nodes after it tries to connect to them `PeerFinder#onFoundPeersUpdated` (and therefore this listener) is called again, even though the number of peers has not actually changed
3. Once the new master is elected everything is removed from the collection of peers and this listener is called again.

For #88562 I only care about the first one. It's possible the others could have uses too.

masseyke:
@DaveCTurner does this look like what you had in mind when we talked?

server/src/main/java/org/elasticsearch/cluster/coordination/ClusterBootstrapService.java
```diff
@@ -147,7 +147,7 @@ void logBootstrapState(Metadata metadata) {
         }
     }
 
-    void onFoundPeersUpdated() {
+    public void onFoundPeersUpdated() {
```
DaveCTurner, line 150:
```suggestion
    @Override
    public void onFoundPeersUpdated() {
```

DaveCTurner (approved):
Sure, makes sense. LGTM (one nit)

PR #88641, opened 2022-07-20
Replace health request with a state observer.

idegtiarenko:
This was using health request assuming yellow index is one with all primaries
initialized. This is not true as newly created index remain yellow when
primaries allocation is throttled. This was discovered while working on a new
shards allocator.

### DIFF
--- a/x-pack/plugin/fleet/src/internalClusterTest/java/org/elasticsearch/xpack/fleet/action/GetGlobalCheckpointsActionIT.java
+++ b/x-pack/plugin/fleet/src/internalClusterTest/java/org/elasticsearch/xpack/fleet/action/GetGlobalCheckpointsActionIT.java
@@ -31,6 +31,7 @@ import java.util.Comparator;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static org.elasticsearch.cluster.routing.allocation.decider.ThrottlingAllocationDecider.CLUSTER_ROUTING_ALLOCATION_NODE_INITIAL_PRIMARIES_RECOVERIES_SETTING;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.lessThan;
 import static org.hamcrest.Matchers.lessThanOrEqualTo;
@@ -70,9 +71,7 @@ public class GetGlobalCheckpointsActionIT extends ESIntegTestCase {
         );
         final GetGlobalCheckpointsAction.Response response = client().execute(GetGlobalCheckpointsAction.INSTANCE, request).get();
         long[] expected = new long[shards];
-        for (int i = 0; i < shards; ++i) {
-            expected[i] = -1;
-        }
+        Arrays.fill(expected, -1);
         assertArrayEquals(expected, response.globalCheckpoints());
 
         final int totalDocuments = shards * 3;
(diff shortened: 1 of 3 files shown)


DaveCTurner (commented):
LGTM but I'd like Tim to take a look too.

Tim-Brooks (commented):
The changes looks fine to me. Agree with David's comment.

x-pack/plugin/fleet/src/internalClusterTest/java/org/elasticsearch/xpack/fleet/action/GetGlobalCheckpointsActionIT.java
```diff
@@ -305,42 +304,63 @@ public void testWaitOnIndexCreated() throws Exception {
...
+            );
+        } finally {
+            client().admin()
+                .cluster()
+                .prepareUpdateSettings()
+                .setPersistentSettings(
+                    Settings.builder().putNull(CLUSTER_ROUTING_ALLOCATION_NODE_INITIAL_PRIMARIES_RECOVERIES_SETTING.getKey()).build()
+                )
+                .get();
+        }
```
idegtiarenko, line 354: Inline diff is totally confusing, consider reviewing in a split mode.

In short, this test:

* sets `cluster.routing.allocation.node_initial_primaries_recoveries=0` so that every created index is throttled
* creates an index
* executes `GetGlobalCheckpointsAction` with a timeout
* asserts it is complaining about unavailable shards rather then anything else
* reverts the setting to a default value

idegtiarenko: I think I could also rewrite it to revert the setting after executing `GetGlobalCheckpointsAction` and assert it could get result back but I am not sure it would make the test better

DaveCTurner: Yes I think we want that. As it stands the test suite still passes on `master` today (i.e. reverting the production-code changes in this PR). I think we have to check the success case here to see the value in this change.

DaveCTurner (changes requested):

x-pack/plugin/fleet/src/main/java/org/elasticsearch/xpack/fleet/action/GetGlobalCheckpointsAction.java
```diff
@@ -205,60 +210,61 @@ protected void doExecute(Task task, Request request, ActionListener<Response> li
...
+            var observer = new ClusterStateObserver(initialState, clusterService, request.timeout(), logger, threadPool.getThreadContext());
+
+            observer.waitForNextChange(new ClusterStateObserver.Listener() {
+                @Override
+                public void onNewClusterState(ClusterState state) {
+                    try {
+                        var index = resolver.concreteSingleIndex(state, request);
+                        long elapsedNanos = System.nanoTime() - startNanos;
                         long remainingNanos = request.timeout().nanos() - elapsedNanos;
-                        if (routingTable.allPrimaryShardsActive() && remainingNanos > 0) {
```
idegtiarenko, line 234: Looks like I forgot to check for `remainingNanos > 0` in a new code

idegtiarenko: Fixed

DaveCTurner (approved):
LGTM

PR #88643, opened 2022-07-20
Use custom task instead of generic AckedClusterStateUpdateTask

idegtiarenko:
Use custom task instead of generic AckedClusterStateUpdateTask that is not intended to be used with batching.

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/metadata/MetadataUpdateSettingsService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/metadata/MetadataUpdateSettingsService.java
@@ -75,8 +75,8 @@ public class MetadataUpdateSettingsService {
             for (final var taskContext : taskContexts) {
                 try {
                     final var task = taskContext.getTask();
-                    state = task.execute(state, taskContext);
-                    taskContext.success(task.getAckListener());
+                    state = task.execute(state);
+                    taskContext.success(task);
                 } catch (Exception e) {
                     taskContext.onFailure(e);
                 }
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/elasticsearch/cluster/metadata/MetadataUpdateSettingsService.java
```diff
@@ -56,223 +56,261 @@ public class MetadataUpdateSettingsService {
...
+                final IndexMetadata metadata = currentState.metadata().getIndexSafe(index);
+                if (metadata.getState() == IndexMetadata.State.OPEN) {
+                    openIndices.add(index);
+                } else {
+                    closedIndices.add(index);
                 }
+            }
 
-                if (skippedSettings.isEmpty() == false && openIndices.isEmpty() == false) {
-                    throw new IllegalArgumentException(
```
DaveCTurner, line 158: I think we can still throw here...

server/src/main/java/org/elasticsearch/cluster/metadata/MetadataUpdateSettingsService.java
```diff
@@ -56,223 +56,261 @@ public class MetadataUpdateSettingsService {
...
-                        final IndexMetadata currentMetadata = currentState.metadata().getIndexSafe(index);
-                        final IndexMetadata updatedMetadata = updatedState.metadata().getIndexSafe(index);
-                        // Verifies that the current index settings can be updated with the updated dynamic settings.
-                        indicesService.verifyIndexMetadata(currentMetadata, updatedMetadata);
-                        // Now check that we can create the index with the updated settings (dynamic and non-dynamic).
-                        // This step is mandatory since we allow to update non-dynamic settings on closed indices.
-                        indicesService.verifyIndexMetadata(updatedMetadata, updatedMetadata);
-                    }
-                } catch (IOException ex) {
-                    throw ExceptionsHelper.convertToElastic(ex);
```
DaveCTurner, line 265: ... and here. Then we don't need to pass in the `taskContext`.

server/src/main/java/org/elasticsearch/cluster/metadata/MetadataUpdateSettingsService.java
```diff
@@ -56,223 +56,261 @@ public class MetadataUpdateSettingsService {
...
+        public void onFailure(Exception e) {
+            listener.onFailure(e);
+        }
+
+        @Override
+        public void clusterStateProcessed(ClusterState oldState, ClusterState newState) {
+            assert false : "should not be called";
+        }
+
+        private ClusterStateAckListener getAckListener() {
```
DaveCTurner, line 111: Typically ack-sensitive tasks implement `ClusterStateAckListener` themselves rather than creating a new object. Not that it really matters, just a question of harmony with other implementations.

DaveCTurner (commented):
Looks good, I left a couple of (debatable) suggestions.

DaveCTurner (approved):
LGTM

PR #88649, opened 2022-07-20
[ML] make allocated trained model infer requests fully cancellable

benwtrent:
When an infer request is made, it may or may not be queued for later execution. If the caller making the inference request stops listening for the result, we should not execute action.

This commit allows for infer requests made to deployed models to be cancelled even after they are queued for inference.

Related to: https://github.com/elastic/elasticsearch/pull/88009


droberts195 (approved):
LGTM

PR #88655, opened 2022-07-20
[ML] make bucket_correlation aggregation generally available

benwtrent:
Originally released in 7.14, bucket_correlation is now generally available.


droberts195 (approved):
LGTM

droberts195:
We've decided to defer GA to 8.5, via #89172, so I've changed the labels on this issue to reflect that.

PR #88656, opened 2022-07-20
[ML] Fail model deployment if all allocations cannot be provided

dimitris-athanasiou:
When we cannot scale up (autoscaling is disabled) we should fail
requests to start a trained model deployment whose allocations
cannot be provided.

### DIFF
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportStartTrainedModelDeploymentAction.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportStartTrainedModelDeploymentAction.java
@@ -488,14 +488,10 @@ public class TransportStartTrainedModelDeploymentAction extends TransportMasterN
                 .filter(d -> nodesShuttingDown.contains(d.getId()) == false)
                 .filter(TaskParams::mayAssignToNode)
                 .toList();
-            OptionalLong smallestMLNode = nodes.stream().map(NodeLoadDetector::getNodeSize).flatMapToLong(OptionalLong::stream).min();
+            boolean isScalingPossible = isScalingPossible(nodes);
 
             // No nodes allocated at all!
-            if (nodeIdsAndRouting.isEmpty()
-                // We cannot scale horizontally
-                && maxLazyMLNodes <= nodes.size()
-                // We cannot scale vertically
-                && (smallestMLNode.isEmpty() || smallestMLNode.getAsLong() >= maxMLNodeSize)) {
+            if (nodeIdsAndRouting.isEmpty() && isScalingPossible == false) {
                 String msg = "Could not start deployment because no suitable nodes were found, allocation explanation ["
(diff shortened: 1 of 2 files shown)


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportStartTrainedModelDeploymentAction.java
```diff
@@ -509,6 +509,15 @@ public boolean test(ClusterState clusterState) {
...
             }
 
+            // We cannot add more nodes and the assignment is not satisfied
+            if (maxLazyMLNodes <= nodes.size()
+                && trainedModelAssignment.isSatisfied(nodes.stream().map(DiscoveryNode::getId).collect(Collectors.toSet())) == false) {
+                String msg = "Could not start deployment because there are not enough resources to provide all requested allocations";
+                logger.debug(() -> format("[%s] %s", modelId, msg));
+                exception = new ElasticsearchStatusException(msg, RestStatus.TOO_MANY_REQUESTS);
+                return true;
+            }
```
benwtrent, line 519: We should also check to make sure that the current node size is less than the max node size (vertical scaling vs horizontal).

dimitris-athanasiou: Well spotted. I've added a commit that factors out a `isScalingPossible` and uses it in both places we make that check.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/action/TransportStartTrainedModelDeploymentAction.java
```diff
@@ -527,6 +532,15 @@ public boolean test(ClusterState clusterState) {
             );
             return false;
         }
+
+        private boolean isScalingPossible(List<DiscoveryNode> nodes) {
```
droberts195, line 536: I think you should add a comment that this only considers memory, and in the future it would be nice to consider CPU too.

It means there's a discrepancy in how we handle different situations:

- If autoscaling is enabled, a cluster is scaled to maximum size, there are 2 free CPUs, and you ask to start a deployment that needs 3 CPUs then you get told you cannot.
- If autoscaling is enabled, a cluster is scaled to one step below its maximum size, there are no free CPUs, and you ask to start a deployment that needs 100000 CPUs then that's fine - we start it, the cluster scales to its maximum size and the deployment goes ahead but with far fewer than 100000 CPUs allocated to it.

While autoscaling doesn't understand CPU there's not a lot we can do about this, but it's worth at least adding comments to acknowledge where we are today.

dimitris-athanasiou: I have added a `TODO` comment.

benwtrent (approved):

PR #88657, opened 2022-07-20
[ML] Make bucket_count_ks_test aggregation generally available

benwtrent:
Initially released in 7.14, bucket_count_ks_test is now generally available.


droberts195:
We've decided to defer GA to 8.5, via #89172, so I've changed the labels on this issue to reflect that.

PR #88659, opened 2022-07-20
Simplify plugin descriptor tests

rjernst:
The plugin descriptor tests are extremely verbose, always specifying all
of the plugin properties to write to a descriptor file. Yet most of the
time only one property needs to be tested. This commit simplifies these
tests by having a set of template descriptor properties, and a helper
method to write and read those properties, allowing additional
properties to override those in the template.


jdconrad (approved):
LGTM

PR #88662, opened 2022-07-20
Refactor IndicesPermission authorize method

slobodanadamovic:
Separated authorization check from computation of index access control. 
The change is simply a preparation for allowing the access control to be [computed lazily](https://github.com/elastic/elasticsearch/pull/88708).


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/IndicesPermission.java
```diff
@@ -530,7 +520,90 @@ public IndicesAccessControl authorize(
...
+            // true if ANY group, which contains certain ingest privileges, covers the given index AND the action is a mapping update for
+            // an index or an alias (but not for a data stream)
+            boolean bwcGrantMappingUpdate = false;
+            final List<Runnable> bwcDeprecationLogActions = new ArrayList<>();
+
+            for (Group group : groups) {
+                // the group covers the given index OR the given index is a backing index and the group covers the parent data stream
+                if (resource.checkIndex(group)) {
+                    boolean actionCheck = group.checkAction(action);
+                    granted = granted || actionCheck;
```
albertzaharovits, line 550: Since we're in the performance realm with this PR, we could shortcircuit the check here:
```
if (actionCheck) {
  granted = true;
  break;
}
```

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/IndicesPermission.java
```diff
@@ -530,7 +520,90 @@ public IndicesAccessControl authorize(
...
+
+    /**
+     * Returns {@code true} if action is granted for all {@code requestedResources}.
+     * If action is not granted for at least one resource, this method will return {@code false}.
+     */
+    private boolean isActionGranted(final String action, final Map<String, IndexResource> requestedResources) {
+
+        if (requestedResources.isEmpty()) {
+            return true;
+        }
```
albertzaharovits, line 534: Please just use an assert here (since it is never the case). It is slightly distressing that an authz method `isActionGranted` has a special case for the granted case (ie it would be less distressing if the return was `false`).

albertzaharovits (approved):
Please address the shortcircuit suggestion, otherwise it's good to merge.

albertzaharovits:
@slobodanadamovic ,actually, please do not merge until the `8.4` branch is been cut (next Tuesday) because this change by itself is actually less performant because the permission groups are checked multiple times (once in the authz and then in the accesss control).

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/IndicesPermission.java
slobodanadamovic: That's a good suggestion. Thanks for suggesting it. 🙏 

Analogously, do we want to stop at first `bwcMappingActionCheck==true` or is there any benefit from logging all deprecated privileges?

slobodanadamovic: Suggestion applied in commit: https://github.com/elastic/elasticsearch/pull/88662/commits/5dc47b770abd99431a93573faf7f494f29b39868

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/IndicesPermission.java
slobodanadamovic: Added assert in [0a3dd7d](https://github.com/elastic/elasticsearch/pull/88662/commits/0a3dd7d05332ea431873cda303d04c5d00db9f91)

albertzaharovits (approved):
LGTM

PR #88663, opened 2022-07-20
Removing the notion of components from the health API

masseyke:
Users found the concept of "components" within the health API to be confusing and unhelpful. This PR removes them. That means that `components` is gone from being a top-level field in the response, and `indicators` is promoted into its place. Where the API previously returned something like:
```
{
    "status": "red",
    "cluster_name": "TEST-TEST_WORKER_VM=[--not-gradle--]-CLUSTER_SEED=[-6341988252367058471]-HASH=[11CC4EF6482AF0]-cluster",
    "components": {
        "cluster_coordination": {
            "status": "red",
            "indicators": {
                "master_is_stable": {
                    "status": "red",
                    "summary": "No master node observed in the last 30s, and some master eligible nodes are unable to discover other master eligible nodes",
                    "help_url": "https://ela.st/fix-master",
                    "details": {...},
                    "impacts": [...],
                    "user_actions": [...]
                }
            }
        },
        "data": {...},
        "snapshot": {...}
    }
}
```
It now returns something like this:
```
{
    "status": "red",
    "cluster_name": "TEST-TEST_WORKER_VM=[--not-gradle--]-CLUSTER_SEED=[-6341988252367058471]-HASH=[11CC4EF6482AF0]-cluster",
    "indicators": {
        "master_is_stable": {
            "status": "red",
            "summary": "No master node observed in the last 30s, and some master eligible nodes are unable to discover other master eligible nodes",
            "help_url": "https://ela.st/fix-master",
            "details": {...},
            "impacts": [...],
            "user_actions": [...]
        }
    }
}
```
All indicators are now returned in the single `indicators` block, and there is no such thing as a component name or status. This also impacts drilldown -- you can drill down by indicator name but not by component.


server/src/main/java/org/elasticsearch/health/GetHealthAction.java
```diff
@@ -45,17 +45,17 @@ public static class Response extends ActionResponse implements ToXContentObject
         private final ClusterName clusterName;
         @Nullable
         private final HealthStatus status;
-        private final List<HealthComponentResult> components;
+        private final List<HealthIndicatorResult> indicatorResults;
```
gmarouli, line 48: Nit: I did like the variable name `indicators` a bit more.

server/src/main/java/org/elasticsearch/health/GetHealthAction.java
```diff
@@ -109,35 +105,32 @@ public boolean equals(Object o) {
...
         }
     }
 
     public static class Request extends ActionRequest {
-        private final String componentName;
         private final String indicatorName;
         private final boolean explain;
 
         public Request(boolean explain) {
-            // We never compute details if no component name is given because of the runtime cost:
```
gmarouli, line 132: Should we preserve this comment? If I understand correctly the same applies for the indicators, right?

server/src/test/java/org/elasticsearch/health/HealthServiceTests.java
```diff
@@ -141,59 +115,24 @@ public void testPreflightIndicatorResultsPresent() {
         );
 
         // Get all indicators returns preflight result mixed in with appropriate component
```
gmarouli, line 117: I think this comment needs an update :)

server/src/test/java/org/elasticsearch/health/HealthServiceTests.java
```diff
@@ -141,59 +115,24 @@ public void testPreflightIndicatorResultsPresent() {
...
-            assertThat(component2.indicators(), is(notNullValue()));
-            assertThat(component2.indicators(), contains(shardsAvailable));
-        }
-
-        // Getting single component returns preflight result mixed in with appropriate component
-        health = service.getHealth("coordination", null, false);
-        assertThat(health.size(), is(equalTo(1)));
+        List<HealthIndicatorResult> health = service.getHealth(null, false);
+        assertThat(health.size(), is(equalTo(4)));
         {
```
gmarouli, line 120: Nit: Do you think we should keep this and the blocks bellow? I am not sure if they are adding anything. I am referring only to the one liners.

gmarouli (approved):
LGTM, small comments. Thank you for picking this up, I can imagine is not the most interesting work but it looks so much nicer with less nesting :)

server/src/main/java/org/elasticsearch/health/GetHealthAction.java
masseyke: Oh good point. I'll restore it.

jbaiera (approved):
_The fact that components are gone, that you no longer need to group them up, it fills you with determination._

PR #85141, opened 2022-03-20
Update to to Gradle wrapper 7.5

breskeby:
This updates the gradle wrapper to a 7.5

Fixes #85123

PR #88664, opened 2022-07-20
Reinstate test cluster throttling behavior

mark-vieira:
As a result of https://github.com/elastic/elasticsearch/pull/85141 upgrading to Gradle 7.5 we lost some previous behavior related to test cluster parallelism. Previously, when we registered a resource lock we also provided how many leases we were requesting. The reason for this is that some tests spin up a single node cluster and others multi-node clusters. To differentiate between these tasks we count each node as a lease towards to total allowed concurrent leases.

Since there's no longer an explicit API for this in Gradle 7.5, the ~~hack~~ workaround is simply to register multiple locks from the same shared resource. I've tested this locally in a simple standalone project to better demonstrate that the approach does indeed work as intended.

https://gradle-enterprise.elastic.co/s/2xgmqx2ika5si/timeline


mark-vieira:
@elasticmachine update branch

breskeby (approved):
neat!

PR #88667, opened 2022-07-21
Add documentation for `knn` option

jtibshirani:
This change adds docs for the new search `knn` option:
* Add `knn` option to `_search` API docs.
* Update the kNN search guide to use the `_search` endpoint instead of
`_knn_search` endpoint. Move relevant information from the `_knn_search` API
docs to this guide, like the filtering example.
* Add an example to the kNN search guide of configuring `index_options` (just
seemed nice to have, not core to this change).

The PR doesn't deprecate the `_knn_search` endpoint yet.

Addresses #87625.

### DIFF
--- a/docs/reference/search/search.asciidoc
+++ b/docs/reference/search/search.asciidoc
@@ -483,7 +483,8 @@ A boost value greater than `1.0` increases the score. A boost value between
 experimental::[]
 [[search-api-knn]]
 `knn`::
-(Required, object) Defines the kNN query to run.
+(Optional, object) Defines the <<approximate-knn, approximate kNN search>> to
+run.
 +
 .Properties of `knn` object
 [%collapsible%open]
--- a/docs/reference/search/search.asciidoc
+++ b/docs/reference/search/search.asciidoc
@@ -480,6 +480,39 @@ A boost value greater than `1.0` increases the score. A boost value between
 `0` and `1.0` decreases the score.
 ====
 
+experimental::[]
+[[search-api-knn]]
+`knn`::
+(Required, object) Defines the kNN query to run.
++
+.Properties of `knn` object
+[%collapsible%open]
+====
+`field`::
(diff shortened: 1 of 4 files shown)


docs/reference/search/search.asciidoc
```diff
@@ -480,6 +480,39 @@ A boost value greater than `1.0` increases the score. A boost value between
 `0` and `1.0` decreases the score.
 ====
 
+experimental::[]
```
mayya-sharipova, line 483: Is  the new`knn` section within `_search` API is also experimental?
I was under impression that it is not experimental anymore.

 cc @giladgal

docs/reference/search/search.asciidoc
```diff
@@ -480,6 +480,39 @@ A boost value greater than `1.0` increases the score. A boost value between
...
+[[search-api-knn]]
+`knn`::
+(Required, object) Defines the kNN query to run.
++
```
mayya-sharipova, line 487: Should we include a reference here to the `approximate-knn` section of the  `knn-search.asciidoc`? I was thinking if a user stumble on this page, they would not know that we have mode details on knn search.

mayya-sharipova (approved):
@jtibshirani Thanks, this LGTM. I left two small comments.

docs/reference/search/search.asciidoc
jtibshirani: In our discussions, we said that integrating ANN into the `_search` endpoint is a separate step from making the feature GA. Before GA, we still have a few items to tackle (like adding telemetry, improving index performance, etc.)

docs/reference/search/search.asciidoc
jtibshirani: Good idea, I'll add a link. I also realized this should be "Optional" instead of "Required".

jtibshirani:
Thanks for the review! I'm going to merge, but happy to discuss anything further.

PR #88675, opened 2022-07-21
Upgrade to OpenJDK 18.0.2+9

### LINKED ISSUE #88673: Upgrade OpenJDK to 18.0.2+9
OpenJDK 18.0.2+9 is now available. We need to upgrade the version that is bundled with Elasticsearch to the newer version. We currently bundle 18.0.1.1+2.

pugnascotia:
Closes #88673.

### DIFF
--- a/build-tools-internal/version.properties
+++ b/build-tools-internal/version.properties
@@ -2,7 +2,7 @@ elasticsearch     = 8.4.0
 lucene            = 9.3.0-snapshot-2d05f5c623e
 
 bundled_jdk_vendor = openjdk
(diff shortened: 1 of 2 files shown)


breskeby (approved):

mark-vieira:
@pugnascotia we should backport this to 8.3 as well.

pugnascotia:
Backported to `8.3` in 7ed2d9f1c6d. Depending on whether there's another BC, it may or may not make it into 8.3.3.

mark-vieira:
> Depending on whether there's another BC, it may or may not make it into 8.3.3.

Yep, that's fine. We don't want to block it but if it gets in, great.

PR #88677, opened 2022-07-21
[DOCS] Security domain splitting impacts API keys

n1v0lg:
This PR documents the impact of domain splitting on API keys. API key
ownership is determined via username and user realm information,
including the user's security domain. API key ownership is shared
across users with the same username that are part of the same security
domain.  A user loses ownership over an API key if their realm is
removed from the security domain that previously enabled ownership
through cross-realm resource sharing.


x-pack/docs/en/security/authentication/security-domain.asciidoc
```diff
@@ -22,7 +22,8 @@ Some types of resources in {es} are owned by a single user, such as
 <<async-search,async search contexts>>, <<security-api-create-api-key,API keys>>,
 and <<user-profile,user profiles>>. When a user creates a resource, {es}
 captures the user's username and realm information as part of the resource's
-metadata.
+metadata. Likewise, if a user updates a resource, such as an API key,
```
n1v0lg, line 25: Not strictly necessary but feels like a worthwhile callout. It makes the transition to the next paragraph slightly more clunky so I can also remove it/push it somewhere else.

x-pack/docs/en/security/authentication/security-domain.asciidoc
```diff
@@ -124,13 +125,15 @@ When adding realms to a security domain, avoid authenticating with a newly-added
...
-Resources created before the removal can be owned by different users depending on the resource type:
+Resources created or updated before the removal can be owned by different users depending on the resource type:
 
 - <<user-profile,User profiles>> are owned by the user for whom the profile was last
 <<security-api-activate-user-profile,activated>>.
 For users whose realms are no longer in the same domain as the owner user, a new user profile
 will be created for them next time the activate user profile API is called.
-- Resources such as API keys are owned by the user who originally created them.
+- An API key is owned by the user who originally <<security-api-create-api-key,created>> or last <<security-api-update-api-key,updated>> it.
+Users whose realms are no longer in the same domain as the API key's current owner won't be able to update the API key. Likewise, they won't be able to perform queries on the API key that rely on ownership (for example, calls to the <<security-api-get-api-key,get API key API>> with the `owner` parameter set to `true`).
```
n1v0lg, line 135: Happy to cut the "for example, calls to the <<security-api-get-api-key,get API key API>> with the `owner` parameter set to `true`" if it's TMI in this context

n1v0lg: Agreed, the specific API mentions feel out of place here.

> better establish what ownership means

Do you mean what ownership means for API keys, or more generally, what ownership means in the context of security domains?

I propose to go ahead in two parts:

For this PR, simply state:

> An API key is owned by the user who originally <<security-api-create-api-key,created>> or last <<security-api-update-api-key,updated>> it. Note that the original creator of an API key will lose ownership, if their realms are no longer in the same domain as those of the current API key owner.

In the future API key overview doc, discuss API key ownership in more detail.

WDYT?

n1v0lg:
@lockewritesdocs being mindful of on-week next week, I've requested your review now although there is still the possibility that the current content is not what we want. For your review, lets assume that it is. If @ywangd agrees with the current form on Monday, we will be able to merge before feature freeze. Otherwise, we will defer until after feature freeze. This should be fine, since it's a small doc change as opposed to feature work. Cheers!

x-pack/docs/en/security/authentication/security-domain.asciidoc
ywangd: Yep. Let's define what is ownership separately and more generally since it is not only just for API keys.

ywangd (approved):
LGTM

Content is accurate and follows the current way of explaining things. We agreed that ownership can and should be explained better in a separate PR.

x-pack/docs/en/security/authentication/security-domain.asciidoc
```diff
@@ -124,13 +125,15 @@ When adding realms to a security domain, avoid authenticating with a newly-added
...
-Resources created before the removal can be owned by different users depending on the resource type:
+Resources created or updated before the removal can be owned by different users depending on the resource type:
 
 - <<user-profile,User profiles>> are owned by the user for whom the profile was last
 <<security-api-activate-user-profile,activated>>.
 For users whose realms are no longer in the same domain as the owner user, a new user profile
 will be created for them next time the activate user profile API is called.
-- Resources such as API keys are owned by the user who originally created them.
+- An API key is owned by the user who originally <<security-api-create-api-key,created>> or last <<security-api-update-api-key,updated>> it.
+Users, including the original creator of the API key, will lose ownership, if their realms are no longer in the same domain as those of the current API key owner.
```
lockewritesdocs, line 135:
```suggestion
Users, including the original creator of the API key, will lose ownership if their realms are no longer in the same domain as those of the current API key owner.
```

lockewritesdocs (approved):
Minor nit -- LGTM otherwise 🧇

x-pack/docs/en/security/authentication/security-domain.asciidoc
n1v0lg: I had a suspicion that comma didn't belong there. Cheers!

PR #88678, opened 2022-07-21
Introduce step to separate explicit names from wildcards during index expression evaluation 

albertzaharovits:
This is a refactoring, that will enable the Security filter to reuse the index expression evaluation logic and ditch the historical alternative one.
It is still rather incomplete, in order to not complicate the review process too much.
The goal here is to have an isolated index expression processing step that separates explicit names from wildcard names. This new step fits (eventually) after the date math expression evaluation step (the first step), and before the wildcard evaluation step (the last step).

The main reason is that we need this extra step (separated from the wildcard expansion step) in the Security filter (`IndicesAndAliasesResolver#resolveIndicesAndAliases`, and without duplicating the logic there) in order to improve performance by running the authorization only on the explicit names.

Related: https://github.com/elastic/elasticsearch-team-planning/issues/168


server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1158,117 +1254,59 @@ public static List<String> resolve(Context context, List<String> expressions) {
...
+            ExplicitExpressionResolver.Resolved explicitsAndWildcards = ExplicitExpressionResolver.resolve(
+                context,
+                expressions,
+                options,
+                metadata
+            );
+            if (explicitsAndWildcards.wildcardsAndExclusions().isEmpty()) {
+                return new ArrayList<>(explicitsAndWildcards.explicitNames());
+            } else {
+                Set<String> expandedWildcards = innerResolve(context, explicitsAndWildcards.wildcardsAndExclusions(), options, metadata);
```
slobodanadamovic, line 1266: If I understand this correctly, we will always include explicitly mentioned names. Meaning, exclusions will never be evaluated agains explicit names.

I've run a simple test with the following example:
Given indices: `testXXX`, `testXYY` and `testYYY`
and expressions: `testYYY`, `testX*`, `-*`
before refactoring it would result in an `empty set` but after refactoring it returns `testYYY`.

slobodanadamovic: If it helps, this is the test I wrote in `WildcardExpressionResolverTests`:

```java
    public void testConvertWildcardsWithExclusion() {
        Metadata.Builder mdBuilder = Metadata.builder()
            .put(indexBuilder("testXXX"))
            .put(indexBuilder("testXYY"))
            .put(indexBuilder("testYYY"));
        ClusterState state = ClusterState.builder(new ClusterName("_name")).metadata(mdBuilder).build();

        IndexNameExpressionResolver.Context context = new IndexNameExpressionResolver.Context(
            state,
            IndicesOptions.lenientExpandOpen(),
            SystemIndexAccessLevel.NONE
        );

        assertThat(
            newHashSet(IndexNameExpressionResolver.WildcardExpressionResolver.resolve(context, Arrays.asList("testYYY", "testX*", "-*"))),
            equalTo(newHashSet())
        );
    }
```

albertzaharovits: 🤦 this is a problem, thanks for catching it!
I have pushed [7545d15](https://github.com/elastic/elasticsearch/pull/88678/commits/7545d1559d32a080aee426b72e15356f1ae044a8) .

albertzaharovits: It's a sign we need more tests, I will see to it.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1158,117 +1258,59 @@ public static List<String> resolve(Context context, List<String> expressions) {
...
+                if (expandedWildcards.isEmpty() && options.allowNoIndices() == false) {
+                    IndexNotFoundException infe = new IndexNotFoundException((String) null);
+                    infe.setResources("index_or_alias", explicitsAndWildcards.wildcardsAndExclusions().toArray(new String[0]));
+                    throw infe;
+                }
+                expandedWildcards.addAll(explicitsAndWildcards.explicitNames());
+                return new ArrayList<>(expandedWildcards);
```
slobodanadamovic, line 1277: Nit: As mentioned during our discussion, it might make sense to return immutable collection and change it to `Set` instead of `List`?

slobodanadamovic (approved):
LGTM

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
albertzaharovits: I now realize that this code in "core" permits exclusions of names by a different name.
For example, given two indices `index1` and `index2` and an alias, `alias` that refers to both, the expression `index1*,-alias1*` resolves to no indices when Security is disabled, but resolve to `index1` when Security is enabled.
We need to discuss what behavior we want to keep...
I'll organize an email/meeting.

rjernst (commented):
I don't have any comments on the specific refactoring here, but I have some general thoughts about the index name resolver, and making changes to it:

* We have many tests for the expression resolver, but they are long, and difficult to see the patterns being tested without reading through a lot of code. Are the cases you've described in several comments, or in general are meaning to reuse later, currently tested? In particular, it seems like it would be good to have unit tests for the new static resolver code added in this PR.
* The logic of the overall resolver is already difficult to follow with so many specialized methods. I wonder if this could be improved by reworking the resolver to have a stream-like API, similar to Lucene's analysis api token streams. This could allow building the chain of resolvers, and make it more clear which resolver "wins", as well as easier compositing of resolution. Such a refactoring is clearly beyond the scope of this PR, but I wanted to suggest it as a thought for future work here because the resolver has clearly become unwieldy.

albertzaharovits:
Per the comment in https://github.com/elastic/elasticsearch/pull/88678#discussion_r930887888 this is not a viable change to introduce. Closing.

PR #88686, opened 2022-07-21
Allow any map key type when serialising hash maps

### LINKED ISSUE #66057: Scripted Metric Aggregation Failure During Serialization
A change made in (https://github.com/elastic/elasticsearch/pull/62884) appears to have caused a regression. Though, I'm not sure if this is actually a "fixed" bug, and just happens to be a breaking change. A class cast exception appears when returning from a combine script as part of a scripted metric aggregation when the return is a Map<Long, Object> where serialization appears to only accept Map<String, Object>.

Steps to reproduce:
1. Run elasticsearch 7.10.0
2.
```
curl -X PUT "localhost:9200/testindex" -H 'Content-Type: application/json' -d'
{
  "mappings": {
    "properties": {
      "instanceId": {
        "store": true,
        "type": "long"
      }
    }
  }
}'

curl -X POST "localhost:9200/testindex/_doc" -H 'Content-Type: application/json' -d'
{
  "instanceId": 1
}
'
```
3.
```
curl -X POST "localhost:9200/_search" -H 'Content-Type: application/json' -d'
{
"query": {
    "match_all": {}
  },
  "aggs": {
    "profit": {
      "scripted_metric": {
        "init_script": "state.test = [:]",
        "map_script": "state.test[doc.instanceId.value] = 1",
        "combine_script": "return state.test",
        "reduce_script": "return 0"
      }
    }
  }
}'
```

Steps 2 and 3 are separate because it seems like the response was masked if done together.

I found this by bisecting with the last known good commit as `cd584d49dcc3819821c677973e388f4103f89418`
and the first bad commit as `78a93dc18fb3db9a1b9785de34753b499f8bcbf8`
from the root branch tag of `v7.10.0`

cc @jimczi

salvatore-campagna:
Before this change only maps with String keys
were supported. There is no specific reason
why we should not  support non-String keys
too, provided that they are serialisable. With
this PR we include support for other map key
types.

Resolves #66057


not-napoleon:
Is `StreamOutput` the right place to fix this issue?  I'm worried that will impact more than just scripted metric.  Maybe that's the right thing to do though, I haven't thought about it.  But I need a little more explanation of what's going on here.

nik9000:
I think it'd be kinder to catch this when the script tries to write a non-string key into the map. Painless doesn't support generics and the API we give for these is `Map` - which can have `Object` keys. We could call `toString` on all the keys on the way in. Or fail any that aren't strings. That'd be kindest, I think.

nik9000:
> That'd be kindest, I think.

@jdconrad mentioned it could be possible to make a `StringObjectMap implements Map<String, Object>`. Painless *would* see that type. It could just wrap a regular `HashMap`.

nik9000:
Er, well, maybe we shouldn't do this. If you had a stored script that required the old type signature then it won't compile and it'll fail to start the node. That's terrible.... Maybe better to just do it at runtime. It'll be more expensive, but scripted metric isn't fast.

rjernst:
What is the behavior we see today? We are already doing a hard cast to `Map<String, Object>`, so I think a class cast exception would already occur?

Technically we could support any key type, it just needs to be (1) hashable and (2) supported by StreamInput/StreamOutput. Today we assume String, but we could not do the cast to `Map<String,Object>`, and then call writeGenericValue on the key, just like we do for the value?

salvatore-campagna:
> Is `StreamOutput` the right place to fix this issue? I'm worried that will impact more than just scripted metric. Maybe that's the right thing to do though, I haven't thought about it. But I need a little more explanation of what's going on here.

I suspected this but I was not sure where those values are generated. I also tried a solution where I just `toString` the key no matter what the type is. In that case we don't have an error anymore but I didn't go for that because from the original issue my understanding is that we want to give back an error. Is my understanding wrong?

salvatore-campagna:
> What is the behavior we see today? We are already doing a hard cast to `Map<String, Object>`, so I think a class cast exception would already occur?
> 
> Technically we could support any key type, it just needs to be (1) hashable and (2) supported by StreamInput/StreamOutput. Today we assume String, but we could not do the cast to `Map<String,Object>`, and then call writeGenericValue on the key, just like we do for the value?

As it is now, without changes, this results in a `ClassCastException`.

```
Caused by: java.lang.ClassCastException: class java.lang.Long cannot be cast to class java.lang.String (java.lang.Long and java.lang.String are in module java.base of loader 'bootstrap')
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.StreamOutput.lambda$static$15(StreamOutput.java:695)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.StreamOutput.writeGenericValue(StreamOutput.java:820)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.StreamOutput.writeCollection(StreamOutput.java:1160)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.search.aggregations.metrics.InternalScriptedMetric.doWriteTo(InternalScriptedMetric.java:70)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.search.aggregations.InternalAggregation.writeTo(InternalAggregation.java:60)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.StreamOutput.writeNamedWriteable(StreamOutput.java:1091)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.StreamOutput.writeNamedWriteableList(StreamOutput.java:1213)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.search.aggregations.InternalAggregations.writeTo(InternalAggregations.java:67)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.DelayableWriteable.getSerializedSize(DelayableWriteable.java:218)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.io.stream.DelayableWriteable$Referencing.getSerializedSize(DelayableWriteable.java:121)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.action.search.QueryPhaseResultConsumer$PendingMerges.ramBytesUsedQueryResult(QueryPhaseResultConsumer.java:313)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.action.search.QueryPhaseResultConsumer$PendingMerges.consume(QueryPhaseResultConsumer.java:339)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.action.search.QueryPhaseResultConsumer.consumeResult(QueryPhaseResultConsumer.java:109)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.action.search.AbstractSearchAsyncAction.onShardResult(AbstractSearchAsyncAction.java:598)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.action.search.SearchQueryThenFetchAsyncAction.onShardResult(SearchQueryThenFetchAsyncAction.java:120)
        at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.action.search.AbstractSearchAsyncAction$1.innerOnResponse(AbstractSearchAsyncAction.java:338)
```

salvatore-campagna:
I understand that the problem is here:

```
"map_script": "state.test[doc.instanceId.value] = 1",
```

The script creates a map where `doc.InstanceId.value` is a Long...which then results in the creation of a `Map<Long, Object>` and now I am wondering. Why do we have to reject keys that are not strings? As long as keys are hash-able they should be fine right?

salvatore-campagna:
I have the feeling my last commit might break a few tests...

salvatore-campagna:
I had a look at `ScriptedMetricAggContexts`, `InternalScriptedMetric` and so on. I see they use `Map<String, Object>` to collect params and state. It is not clear to me why the use String as a key type. I also understand that this issue might be dealt with fixing it at "scripted metric level" other then in StreamOutput/StreamInput. I don't get, anyway, why in StreamOutput/StreamInput we only deal with maps whose key type is String.

salvatore-campagna:
Of course now a few tests are failing I think because of the expectation that the key type is String...at this point I am not sure anymore what is the right way to fix it...

nik9000:
> I don't get, anyway, why in StreamOutput/StreamInput we only deal with maps whose key type is String. Could it be that allowing any type other than String we might hit performance issues due to serialisation and/or hashing overhead?

Honestly if @rjernst is ok with changing the generic map serialization stuff I'd be fine with it changing the stream stuff to support all key types. I think it used a string key because it was usually serializing maps from json which has string keys. It'd probably be good to do a quick inventory of where we're using it to figure out what kind of overhead we're adding here. Scripted metric isn't efficient so I'm not worried about the overhead here, but if we do the map serialization in _search it might be worth asking if it's worth it. I do agree that the current state of the serialization logic is a bit funky.

One thing about fixing it on the serialization side - I'll be it'd still be possible to run into the issue when you try to serialize. It's just that we'd support many more key types. But I'll bet you could stick some kind of unsupported key type into that map with painless.

jdconrad:
I'm not sure `key.toString()` is the correct answer here. Wouldn't this produce asymmetric deserialization? If a user puts a key into the state Map in a Painless script that isn't a String, and they later go to retrieve something out of state with that non-String key, it's very likely it won't return the Object. There's no way during deserialization to know that key isn't really meant as a String.

nik9000:
I think `toString` during serialization would be pretty confusing. It'd be less confusing in scripted metric because you'd notice faster. But still confusing there too.

salvatore-campagna:

salvatore-campagna:
I had a look at `server/src/main/java/org/elasticsearch/script/Script.java` and it looks like supporting a map with generic keys requires quite a lot of work, considering that for BWC we need to support both types Map<String, Object> and Map<Object, Object> for the `params` field.

rjernst:
@salvatore-campagna The params map can stay keyed by string at the top level, no types need to change there. The existing `StreamInput.readMap()` already could use some better error handling. Right now, if a non string key were used, a class cast exception would occur, much like the exception you are trying to address elsewhere.

Separate from this PR, the readMap method (in addition to being renamed, maybe to readStringMap?) could be implemented with type safety like this:

```
readMap(StreamInput::readString, StreamInput::readGenericValue);
```

>  Isn't throwing an error enough as suggested in the original issue?

I don't think an error helps much. The user is still being restricted arbitrarily. A different error message doesn't change the fact they must then change their code to eg explicitly stringify a long value.

nik9000:
@rjernst would you be ok with a change like this one that serializes object keys? That's what folks are *trying* to do with scripted metric and it doesn't work now.

rjernst:
> would you be ok with a change like this one that serializes object keys?

Yes, that is my preferred approach. My comment above was about still allowing a more strongly typed map to be possible for eg params, where we know the keys will be string.

not-napoleon:
Looks like Nik and Ryan have got this, and I'm about to be out of office for a while, so I'm removing myself as a reviewer.  Whatever y'all decide is fine by me.

server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
```diff
@@ -738,8 +738,8 @@ public Object readGenericValue() throws IOException {
             case 6 -> readByteArray();
             case 7 -> readArrayList();
             case 8 -> readArray();
-            case 9 -> readLinkedHashMap();
-            case 10 -> readHashMap();
+            case 9 -> version.onOrAfter(Version.V_8_4_0) ? readObjectKeyLinkedHashMap() : readLinkedHashMap();
```
rjernst, line 741: Why are new methods necessary? I think even the existing linkedhashmap methods can be replaced:

generic linked map:
```
readOrderedMap(StreamInput::readGenericValue, StreamInput::readGenericValue);
```

string -> object linked map
```
readOrderedMap(StreamInput::readString, StreamInput::readGenericValue);
```

And for non linked, the same but with readMap

server/src/main/java/org/elasticsearch/common/io/stream/StreamOutput.java
```diff
@@ -770,6 +772,22 @@ public final void writeOptionalInstant(@Nullable Instant instant) throws IOExcep
         })
     );
 
+    private static void writeHashMap(StreamOutput o, Map<String, Object> map) throws IOException {
```
rjernst, line 775: Similar comment as to the read side. We already have write methods for maps, there should be no need for new methods.

rjernst (commented):
I just have one minor suggestion

server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
```diff
@@ -738,8 +738,8 @@ public Object readGenericValue() throws IOException {
...
+            case 10 -> version.onOrAfter(Version.V_8_4_0) ? readObjectKeyHashMap() : readHashMap();
```
nik9000, line 742: I think maybe we should modify `readLinkedHashMap` to do the object keys. So the version test would be inside those methods.

server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
```diff
@@ -829,6 +829,18 @@ private Map<String, Object> readLinkedHashMap() throws IOException {
         return map9;
     }
 
+    private Map<Object, Object> readObjectKeyLinkedHashMap() throws IOException {
+        int size9 = readArraySize();
```
nik9000, line 833: Maybe just `size` and `map`? I'm not sure where the numbers came from.

nik9000 (commented):
I think this deserves some unit tests in the StreamOutput StreamInput stuff.

server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
nik9000: Ah! I just left a similar comment. FWIW if there are lots of usages of `readLinkedHashMap` then I might bump removing the method to a follow up. But, yeah, this'd be fine:
```
readOrderedMap(StreamInput::readGenericValue, StreamInput::readGenericValue);
```

server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
nik9000: Ignore this comment - Ryan proposed a better solution using the generic map readers.

nik9000:
I think we should modify the signature of the `Map` on scripted metric - we've been running as though it were `Map<Object, Object>` for a long time. Let's just declare it that. Though I'd be quite happy doing that in a follow up to this.

server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
salvatore-campagna: The numbers come, I think, from the fact that 9 or 10 is the identifier used to serialize and deserialize the map distinguishing the two. I just kept the same naming used in the other two methods for "consistency".

salvatore-campagna:
> I think we should modify the signature of the `Map` on scripted metric - we've been running as though it were `Map<Object, Object>` for a long time. Let's just declare it that. Though I'd be quite happy doing that in a follow up to this.

What about BWC if we do that?

nik9000:
> What about BWC if we do that?

It'd hit folks using the API from java-land but I don't think that's particularly common.

server/src/main/java/org/elasticsearch/common/io/stream/StreamInput.java
nik9000: Ah. Well, it doesn't really matter anyway if we don't use a new method.

salvatore-campagna:
I don't manage to find a way to fix the compatibility issues. There are a lot of places where we have `Map<String, Object>` where those `readMap` methods are used.

salvatore-campagna:
I had a look at the BWC issues but I am not able to understand what is wrong with them. To me the code looks correct.

salvatore-campagna:
I think the BWC issue here is because we have a lot of places where maps are declared like
```
Map<String, Object>
```

while, according to this change we would need a
```
Map<Object, Object>
```

I tried to replace some of them but the change propagates to a very large number of classes.
Moreover, in some places we use `writeMpaWIthConsistentOrder` which sorts by key...keys that need to be `Comparable`, not the case for `Object`.

salvatore-campagna:
I have no clue why the BWC tests are failing...it looks correct to me.

rjernst:
This was a tricky issue to find! The mismatch is because you are using `version` directly. If you use `getVersion()`, the tests will pass. The reason is `FilterStreamInput` may is used, which overrides `getVersion()`. That can then create a mismatch of versions which cause writing one way but reading in the other way. IMO this is a bug (or trappiness) in StreamInput. While it would be better if we could enforce using getVersion(), I don't see any way to do that, so i've raised https://github.com/elastic/elasticsearch/pull/92422 to at least make it so `this.version` is correct.

salvatore-campagna:
> This was a tricky issue to find! The mismatch is because you are using `version` directly. If you use `getVersion()`, the tests will pass. The reason is `FilterStreamInput` is used, which overrides `getVersion()`. That can then create a mismatch of versions which cause writing one way but reading in the other way. IMO this is a bug (or trappiness) in StreamInput. While it would be better if we could enforce using getVersion(), I don't see any way to do that, so i've raised #92422 to at least make it so `this.version` is correct.

Ok now I understand. May I ask how did you find that? I spent quite some time debugging this without realizing that could be the issue.

salvatore-campagna:
🥳  thanks @rjernst

rjernst:
> May I ask how did you find that?

it was time consuming. I started by repeating one of the tests that failed. That test was failing on deserializing runtime mappings in a search source builder, so I added printouts on several Stream Input/Output methods regarding maps. Eventually I saw the write side was writing a generic string, but the read side was reading a boolean. I realized the boolean type id 5 was the size of the string, and then added a couple more printouts that showed it was using the new code on the write side, but then old code on the read side. After staring at the read code for a while, I wondered what would happen if getVersion were overridden, and that’s when I saw FilterStreamInput did just that. 

Debugging the transport protocol can be quite tedious. I’m sure there are more efficient ways than what I did above, but in any case I think long term we need a better way to debug (I have an idea for this for a future spacetime project).

salvatore-campagna:
@nik9000 @rjernst  do you think this should be backported?

salvatore-campagna:
@elasticsearchmachine run elasticsearch-ci/full-bwc please

rjernst:
I would not worry about backporting unless there is a compelling reason to do so (eg an anticipated PR that would want to use a map with an object key). However, without backporting it might make all future PRs (to 7.17) that have a map more difficult (though I expect since they are just calling the methods, there will not be a conflict). It really comes down to how easy the backport applies, if it's at all difficult, I wouldn't worry about it.

salvatore-campagna:
@rjernst @nik9000 to day I am going to have another quick review...anyway I need at least an approval to merge this.

server/src/main/java/org/elasticsearch/common/io/stream/StreamOutput.java
```diff
@@ -688,12 +692,14 @@ public final void writeOptionalInstant(@Nullable Instant instant) throws IOExcep
...
             }
-            @SuppressWarnings("unchecked")
-            final Map<String, Object> map = (Map<String, Object>) v;
-            o.writeVInt(map.size());
-            for (Map.Entry<String, Object> entry : map.entrySet()) {
-                o.writeString(entry.getKey());
-                o.writeGenericValue(entry.getValue());
+            if (o.getVersion().onOrAfter(Version.V_8_7_0)) {
+                @SuppressWarnings("unchecked")
+                final Map<Object, Object> map = (Map<Object, Object>) v;
```
martijnvg, line 697: maybe `final Map<?, ?> map = (Map<?, ?>) v;` ? Then `@SuppressWarnings("unchecked")` can be removed.

server/src/main/java/org/elasticsearch/common/io/stream/StreamOutput.java
```diff
@@ -688,12 +692,14 @@ public final void writeOptionalInstant(@Nullable Instant instant) throws IOExcep
...
+                o.writeMap(map, StreamOutput::writeGenericValue, StreamOutput::writeGenericValue);
+            } else {
+                @SuppressWarnings("unchecked")
+                final Map<String, Object> map = (Map<String, Object>) v;
```
martijnvg, line 701: same?

martijnvg (approved):
LGTM

I think fixing this in `InternalScriptedMetric` would be tedious (given that it just serialises a generic list). The fix in `StreamOutput`/`StreamInput` is smaller and shouldn't harm existing usages of maps in the internal protocol.

server/src/main/java/org/elasticsearch/common/io/stream/StreamOutput.java
salvatore-campagna: Here the key needs to be a `String` since we need to use `StreamOutput::writeString` below.

PR #88689, opened 2022-07-21
Allocation reason

idegtiarenko:
Record an allocation reason in `RoutingAllocation` objects so that it could be logged later while allocating for it, computing a balance or expiring outdated input. This should simplify debugging of some of the issues.

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
@@ -86,17 +86,6 @@ public class RoutingAllocation {
         this(deciders, null, clusterState, clusterInfo, shardSizeInfo, currentNanoTime, false, null);
     }
 
-    public RoutingAllocation(
-        AllocationDeciders deciders,
-        @Nullable RoutingNodes routingNodes,
-        ClusterState clusterState,
-        ClusterInfo clusterInfo,
-        SnapshotShardSizeInfo shardSizeInfo,
-        long currentNanoTime
-    ) {
-        this(deciders, routingNodes, clusterState, clusterInfo, shardSizeInfo, currentNanoTime, false, null);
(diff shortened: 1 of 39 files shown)


server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -92,16 +94,18 @@ public RoutingAllocation(
...
      */
     public RoutingAllocation(
         AllocationDeciders deciders,
         @Nullable RoutingNodes routingNodes,
         ClusterState clusterState,
         ClusterInfo clusterInfo,
         SnapshotShardSizeInfo shardSizeInfo,
-        long currentNanoTime
+        long currentNanoTime,
+        @Nullable String reason
```
DaveCTurner, line 106: Does it really need to be nullable? I only see a few ways it can be `null` in production code and all of them really could have a reason instead.

DaveCTurner (commented):
Seems reasonable but I left a question.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -81,7 +82,7 @@ public RoutingAllocation(
         SnapshotShardSizeInfo shardSizeInfo,
         long currentNanoTime
     ) {
-        this(deciders, null, clusterState, clusterInfo, shardSizeInfo, currentNanoTime);
+        this(deciders, null, clusterState, clusterInfo, shardSizeInfo, currentNanoTime, false, null);
```
DaveCTurner, line 85: Hmm it's still `null` here. Could we have callers of this constructor specify their reason too?

DaveCTurner (changes requested):

DaveCTurner (approved):
LGTM

PR #88690, opened 2022-07-21
[TEST] Add mock plugins service loadServiceProviders

grcevski:
This PR adds a bit of infrastructure to be able to correctly load service providers with our SPIClassIterator in MockNode. Namely, the mock node loads all plugins under the same classloader in tests, therefore we need special logic to find which is the correct plugin to use when constructing SPIs with a single argument constructor.

### DIFF
--- a/test/framework/src/main/java/org/elasticsearch/plugins/MockPluginsService.java
+++ b/test/framework/src/main/java/org/elasticsearch/plugins/MockPluginsService.java
@@ -91,7 +91,7 @@ public class MockPluginsService extends PluginsService {
             result.addAll(createExtensions(service, pluginTuple.instance()));
         }
 
-        return result.stream().toList();
+        return List.copyOf(result);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


test/framework/src/main/java/org/elasticsearch/plugins/MockPluginsService.java
```diff
@@ -76,4 +80,47 @@ public PluginsAndModules info() {
...
+    public <T> List<? extends T> loadServiceProviders(Class<T> service) {
+        // We use a set here to avoid duplicates because SPIClassIterator will match
+        // all plugins in MockNode, because all plugins are loaded by the same class loader.
+        Set<T> result = new HashSet<>();
+
+        for (LoadedPlugin pluginTuple : plugins()) {
+            result.addAll(createExtensions(service, pluginTuple.instance()));
+        }
+
+        return result.stream().toList();
```
rjernst, line 94: This could just be `List.copyOf(result)`

test/framework/src/main/java/org/elasticsearch/plugins/MockPluginsService.java
```diff
@@ -76,4 +80,47 @@ public PluginsAndModules info() {
...
+     */
+    static <T> List<? extends T> createExtensions(Class<T> extensionPointType, Plugin plugin) {
+        SPIClassIterator<T> classIterator = SPIClassIterator.get(extensionPointType, plugin.getClass().getClassLoader());
+        List<T> extensions = new ArrayList<>();
+        while (classIterator.hasNext()) {
+            Class<? extends T> extensionClass = classIterator.next();
+
+            @SuppressWarnings("unchecked")
+            Constructor<T>[] constructors = (Constructor<T>[]) extensionClass.getConstructors();
+            boolean compatible = true;
```
rjernst, line 112: Why this leniency?

rjernst (approved):
Looks good, a couple minor comments.

test/framework/src/main/java/org/elasticsearch/plugins/MockPluginsService.java
grcevski: When there's no constructor argument we simply fall though to load the plugin, there's no need to distinguish from which Plugin the load happened. We ensure we de-dup in the caller method by using a set, because the service provider will get loaded for every plugin.

I'll add a comment to this effect.

PR #88700, opened 2022-07-21
Add optional Authorization parameter to remote cluster settings

justincr-elastic:
The purpose of this PR is to add a new Authoriation parameter to remote cluster settings.

The `Authorization` key is case-insensitive.
The `Authorization` value is **optional**.
Support for this parameter is behind a feature flag.

Here is an example of how to add a remote cluster to Cluster Settings with the new `Authorization` setting.

```
curl -s -X PUT --cacert "/tmp/https-ca.crt" -H "Authorization: Basic `echo -n elastic:elastic|base64 -`" -H 'Content-Type: application/json' https://localhost:9211/_cluster/settings?pretty=true -d'
{
  "persistent" : {
    "cluster" : {
      "remote" : {
        "cluster2" : {
          "seeds" : [ "127.0.0.1:9321", "127.0.0.1:9322" ],
          "Authorization" : "value"
        }
      }
    }
  }
}
'
```

### DIFF
--- a/docs/changelog/90302.yaml
+++ b/docs/changelog/90302.yaml
@@ -1,0 +1,5 @@
+pr: 90302
+summary: Fix another potential race in the file settings watcher
+area: Infra/Core
+type: bug
+issues: [89500]
--- a/server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
+++ b/server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
@@ -341,6 +341,7 @@ public class FileSettingsService extends AbstractLifecycleComponent implements C
         logger.debug("stopping watcher ...");
         if (watching()) {
             try {
+                // make sure the watcher thread hits the processing latch correctly
                 cleanupWatchKeys();
                 fileUpdateState = null;
                 watchService.close();
@@ -350,6 +351,8 @@ public class FileSettingsService extends AbstractLifecycleComponent implements C
                 if (watcherThreadLatch != null) {
                     watcherThreadLatch.await();
                 }
+                // the watcher thread might have snuck in behind us and re-created the settings watch again
+                cleanupWatchKeys();
             } catch (IOException e) {
                 logger.warn("encountered exception while closing watch service", e);
             } catch (InterruptedException interruptedException) {


justincr-elastic:
@elasticmachine update branch

justincr-elastic:

server/src/main/java/org/elasticsearch/common/settings/ClusterSettings.java
```diff
@@ -525,7 +531,13 @@ public void apply(Settings value, Settings current, Settings previous) {
         MasterHistory.MAX_HISTORY_AGE_SETTING,
         ReadinessService.PORT,
         HealthNode.isEnabled() ? HealthNodeTaskExecutor.ENABLED_SETTING : null
-    ).filter(Objects::nonNull).collect(Collectors.toSet());
+    )
+        .filter(Objects::nonNull)
+        .filter(
```
gwbrown, line 536: I think I would prefer to check the feature flag inline with the setting, like line 533 above. I'll suggest what I mean on the line above, but that would mean we'd revert this hunk.

server/src/main/java/org/elasticsearch/common/settings/ClusterSettings.java
```diff
@@ -329,6 +334,7 @@ public void apply(Settings value, Settings current, Settings previous) {
         ProxyConnectionStrategy.SERVER_NAME,
         SniffConnectionStrategy.REMOTE_CLUSTERS_PROXY,
         SniffConnectionStrategy.REMOTE_CLUSTER_SEEDS,
+        SniffConnectionStrategy.REMOTE_CLUSTER_AUTHORIZATION,
```
gwbrown, line 337:
```suggestion
        ClusterSettings.CCX2_FEATURE_FLAG_ENABLED ? SniffConnectionStrategy.REMOTE_CLUSTER_AUTHORIZATION : null,
```

See my suggestion below. This way we keep the logic with the setting and can rely on the existing null filter.

server/src/main/java/org/elasticsearch/transport/SniffConnectionStrategy.java
```diff
@@ -64,6 +65,13 @@ public class SniffConnectionStrategy extends RemoteConnectionStrategy {
         }, new StrategyValidator<>(ns, key, ConnectionStrategy.SNIFF), Setting.Property.Dynamic, Setting.Property.NodeScope)
     );
 
+    public static final Setting.AffixSetting<String> REMOTE_CLUSTER_AUTHORIZATION = Setting.affixKeySetting(
+        "cluster.remote.",
+        "authorization",
```
gwbrown, line 70: You say in the PR description that the setting is `Authorization` and is case sensitive, but it's `authorization` (all lowercase) here. I think what's here currently is more closely aligned with our other settings, so can you update the PR description?

server/src/test/java/org/elasticsearch/transport/RemoteClusterServiceTests.java
```diff
@@ -81,6 +81,12 @@ public void testSettingsAreRegistered() {
         assertTrue(ClusterSettings.BUILT_IN_CLUSTER_SETTINGS.contains(RemoteConnectionStrategy.REMOTE_CONNECTION_MODE));
         assertTrue(ClusterSettings.BUILT_IN_CLUSTER_SETTINGS.contains(SniffConnectionStrategy.REMOTE_CONNECTIONS_PER_CLUSTER));
         assertTrue(ClusterSettings.BUILT_IN_CLUSTER_SETTINGS.contains(SniffConnectionStrategy.REMOTE_CLUSTER_SEEDS));
+        assertTrue(
+            (ClusterSettings.CCX2_FEATURE_FLAG_ENABLED
+                && ClusterSettings.BUILT_IN_CLUSTER_SETTINGS.contains(SniffConnectionStrategy.REMOTE_CLUSTER_AUTHORIZATION))
+                || (ClusterSettings.CCX2_FEATURE_FLAG_ENABLED == false
+                    && ClusterSettings.BUILT_IN_CLUSTER_SETTINGS.contains(SniffConnectionStrategy.REMOTE_CLUSTER_AUTHORIZATION) == false)
```
gwbrown, line 88: I think this would be clearer like:
```
if (ClusterSettings.CCX2_FEATURE_FLAG_ENABLED) {
  assertTrue(ClusterSettings.BUILT_IN_CLUSTER_SETTINGS.contains(SniffConnectionStrategy.REMOTE_CLUSTER_AUTHORIZATION))
} else {
  assertFalse(ClusterSettings.BUILT_IN_CLUSTER_SETTINGS.contains(SniffConnectionStrategy.REMOTE_CLUSTER_AUTHORIZATION)
}
```

server/src/test/java/org/elasticsearch/transport/RemoteClusterServiceTests.java
```diff
@@ -110,6 +116,37 @@ public void testRemoteClusterSeedSetting() {
         assertEquals("failed to parse port", e.getMessage());
     }
 
+    public void testRemoteClusterOptionalCredentialSettingValues() {
```
gwbrown, line 119: What is this test testing? I don't see any assertions.

server/src/test/java/org/elasticsearch/transport/RemoteClusterServiceTests.java
```diff
@@ -110,6 +116,37 @@ public void testRemoteClusterSeedSetting() {
...
+        Settings settings = Settings.builder()
+            .put("cluster.remote.foo.authorization", "ApiKey apikey1")
+            .put("cluster.remote.bar.authorization", "ApiKey apikey2")
+            .put("cluster.remote.emptystring.authorization", "")
+            .put("cluster.remote.nullstring.authorization", (String) null)
+            .build();
+        SniffConnectionStrategy.REMOTE_CLUSTER_AUTHORIZATION.getAllConcreteSettings(settings).forEach(setting -> setting.get(settings));
+    }
+
+    public void testRemoteClusterAuthorizationSettingDependencies() {
```
gwbrown, line 131: I wonder if `RemoteClusterSettingsTests` might be a better home for this, given it's testing properties of the remote cluster settings?

AthenaEryma (changes requested):
Left some comments, but this looks like a good start!

server/src/main/java/org/elasticsearch/transport/SniffConnectionStrategy.java
```diff
@@ -107,13 +115,13 @@ public class SniffConnectionStrategy extends RemoteConnectionStrategy {
...
         )
     );
-
     static final int CHANNELS_PER_CONNECTION = 6;
 
     private static final Predicate<DiscoveryNode> DEFAULT_NODE_PREDICATE = (node) -> Version.CURRENT.isCompatible(node.getVersion())
         && (node.isMasterNode() == false || node.canContainData() || node.isIngestNode());
 
     private final List<String> configuredSeedNodes;
+    private final String configuredAuthorization;
```
henningandersen, line 124: Perhaps just `authorization`, the `configured` prefix is there for seed nodes to distinguish from the resolved seed nodes, but here we only need the one field I think.

server/src/main/java/org/elasticsearch/transport/SniffConnectionStrategy.java
```diff
@@ -508,6 +525,14 @@ private static boolean seedsChanged(final List<String> oldSeedNodes, final List<
         return oldSeeds.equals(newSeeds) == false;
     }
 
+    private static boolean authorizationChanged(String oldOptionalCredential, String newOptionalCredential) {
```
henningandersen, line 528: `oldOptionalAuthorization` and `newOptionalAuthorization`? or just `old` and `current`?

server/src/main/java/org/elasticsearch/transport/SniffConnectionStrategy.java
```diff
@@ -508,6 +525,14 @@ private static boolean seedsChanged(final List<String> oldSeedNodes, final List<
...
+        if (oldOptionalCredential == null || oldOptionalCredential.isEmpty()) {
+            return (newOptionalCredential == null || newOptionalCredential.isEmpty()) == false;
+        }
+
+        return Objects.equals(oldOptionalCredential, newOptionalCredential) == false;
```
henningandersen, line 533: Can simplify to:

```
return Strings.isNullOrEmpty(oldOptionalCredential) != Strings.isNullOrEmpty(newOptionalCredential) || Objects.equals(oldOptionalCredential, newOptionalCredential) == false;
```

server/src/test/java/org/elasticsearch/transport/RemoteClusterServiceTests.java
henningandersen: For other feature flags, we ensure the flag is enabled when running most tests. I wonder if we should do the same here.

We could add a minimal test to verify that the setting is not available when the flag is not set. I think we have not done so for for instance tsdb though (@nik9000 may know if this is true).

I'd prefer to keep this similar, since otherwise test code is going to have these checks spread all over as we move forward with this feature. I think the feature is designed such that if `authorization` is not set, it does nothing and we can validate this in tests.

Also, validating it this way means it will pass our regular PR builds etc., but fail in release builds, which is sort of late in the process. If we think we need to validate that the setting is not registered when the feature is not enabled, I'd prefer a separate test for that (which can be done in a few different ways).

henningandersen (commented):
I left a few comments.

Terminology-wise, both credential and authorization is used, I'd like to stick to just authorization, also in the PR description. Also, can we remove the mention of new terminology (querying/fulfilling cluster), I think we should stick to the existing local/remote for now (at least it does not affect this PR). If we want new terminology we have to make a bigger change around this, including documentation and UIs.

The PR description seems to indicate that this PR takes it further than it does. Can we either move the parts that are not included in the PR to a meta-issue or make it clear in the PR description that those parts are for future follow-up PRs?

server/src/test/java/org/elasticsearch/transport/RemoteClusterServiceTests.java
nik9000: > We could add a minimal test to verify that the setting is not available when the flag is not set. I think we have not done so for for instance tsdb though (@nik9000 may know if this is true).

We haven't. I should do that though. It's kind of tricky to run it without the feature flags because we auto-enable all of the feature flags on `-SNAPSHOT` builds. So we'd have to perform a "release" build or skip the test. And we'd need a special `qa` style sub-module for it because we tend to enable the feature flag all over the place.

I do love the paranoia of making sure that we don't enable enable the option without the flag. @mark-vieira might be able to ballpark how much effort it'd take to set something like this up. I do like the idea.

nik9000: > Also, validating it this way means it will pass our regular PR builds etc., but fail in release builds, which is sort of late in the process. If we think we need to validate that the setting is not registered when the feature is not enabled, I'd prefer a separate test for that (which can be done in a few different ways).

We run the release builds nightly, just to catch stuff like this. *but* we're not as good at keeping them working. There is a `test-release` flag that runs the release build on PRs though! It just takes hours.

mark-vieira: We need to figure out what our policy is on feature flags. When I was sorting out issues with BWC testing I found that we have no real convention regarding their implementation (see [this comment](https://github.com/elastic/release-eng/issues/70#issuecomment-1199503493)).

How we handle and test feature flags needs some more discussion, and probably outside the context of this PR. Whatever we decide will likely involve refactoring existing feature flag usages as well. I'm not sure who the right audience is for that discussion though.

henningandersen: In order to unblock this PR, I suggest to simply do like tsdb until we have the proper test handling.

We can still add a unittest for a function deriving the contents of `ClusterSettings.BUILT_IN_CLUSTER_SETTINGS`, though we would have to rely on the input to it being correct. I think this is sufficient for now.

server/src/main/java/org/elasticsearch/common/settings/ClusterSettings.java
justincr-elastic: I agree. I made the change.

server/src/main/java/org/elasticsearch/transport/SniffConnectionStrategy.java
justincr-elastic: I agree. I made the change.

server/src/test/java/org/elasticsearch/transport/RemoteClusterServiceTests.java
justincr-elastic: I wrapped the last line with a try/catch. The catch has a fail message, instead of letting the exception bubble up JUnit. I tweaked the name of the test to hopefully clarify the test too.

server/src/test/java/org/elasticsearch/transport/RemoteClusterServiceTests.java
justincr-elastic: For now, I updated the test base based the example @gwbrown provided.

I don't have any context on TSDB. If you are able to point me to the example you had in mind, it would help me know what you intended. I can take a look and make adjustments.

server/src/test/java/org/elasticsearch/transport/RemoteClusterServiceTests.java
justincr-elastic: I moved two new methods `testRemoteClusterEmptyOrNullApiKey` and `testRemoteClusterAuthorizationSettingDependencies` to `RemoteClusterSettingsTests` as you suggested.

I left the pre-existing `testSettingsAreRegistered` in `RemoteClusterServiceTests`.

@gwbrown, do you think it makes sense to move the pre-existing `testSettingsAreRegistered` to `RemoteClusterSettingsTests` too? Or is what I did sufficient to address your feedback?

server/src/main/java/org/elasticsearch/common/settings/ClusterSettings.java
justincr-elastic: I applied the change.

server/src/main/java/org/elasticsearch/transport/SniffConnectionStrategy.java
justincr-elastic: I changed it from `configuredAuthorization` to authorization. I was following the `configuredSeeds` example, but it is OK to just call it `authorization`.

server/src/main/java/org/elasticsearch/transport/SniffConnectionStrategy.java
justincr-elastic: My intention was to copy `proxyChanged` and replace _proxy_ with _authorization_. If you want me to modify `authorizationChanged`, do you think it is worthwhile applying the same pattern to `proxyChanged`?

https://github.com/elastic/elasticsearch/blob/b1eecb712f5126c297bba20b08dd112aa69698d2/server/src/main/java/org/elasticsearch/transport/SniffConnectionStrategy.java#L528-L534

https://github.com/elastic/elasticsearch/blob/b1eecb712f5126c297bba20b08dd112aa69698d2/server/src/main/java/org/elasticsearch/transport/SniffConnectionStrategy.java#L536-L542

justincr-elastic:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/transport/SniffConnectionStrategy.java
justincr-elastic: I copied the existing proxyChanged method are changed `proxy` to `authorization`. Is that OK?

server/src/test/java/org/elasticsearch/transport/RemoteClusterServiceTests.java
gwbrown: I think leaving the existing test in place is OK. Maybe make a note of it for tech debt week

AthenaEryma (commented):
Left one comment, otherwise it looks good!

server/src/main/java/org/elasticsearch/transport/SniffConnectionStrategy.java
henningandersen: I would not mind either way.

justincr-elastic:
I tried to mergemultiple conflicts locally and pushing to the PR. Somehow that accidentally closed this PR. It was not intentional. I can't find a reopen option, so I think I need to resubmit as a new PR.

server/src/test/java/org/elasticsearch/transport/RemoteClusterServiceTests.java
justincr-elastic: No update for 2 weeks so I assume the change is OK and I will resolve this comment.

server/src/main/java/org/elasticsearch/common/settings/ClusterSettings.java
```diff
@@ -128,6 +129,10 @@
  */
 public final class ClusterSettings extends AbstractScopedSettings {
 
+    // CCX 2.0 = Cross Cluster Search (CCS) 2.0 || Cross Cluster Replication (CCR) 2.0
+    public static final boolean CCX2_FEATURE_FLAG_ENABLED = Build.CURRENT.isSnapshot()
+        || "true".equals(System.getProperty("es.ccx2_feature_flag_enabled"));
+
```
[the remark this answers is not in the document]
justincr-elastic: Done in #90312.

justincr-elastic:
This PR was accidentally closed, so I opened a new one #90312 to continue this work. It is ready for review again.

justincr-elastic:
I will add the same reviewed as this one, so everyone who commented here before can take a fresh look.

PR #88702, opened 2022-07-22
Fix validation of close_pit request

dnhatn:
`ClosePointInTimeRequest#validate` should return a validation error for an invalid pit_id, but it throws an IAE instead. This mistake prevents close_pit tasks of requests with empty pit_id from being removed.

I also looked at other `ActionRequest#validate` implementations. They do not throw exceptions.

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/support/TransportAction.java
+++ b/server/src/main/java/org/elasticsearch/action/support/TransportAction.java
@@ -40,12 +40,19 @@ public abstract class TransportAction<Request extends ActionRequest, Response ex
      * Use this method when the transport action should continue to run in the context of the current task
      */
     public final void execute(Task task, Request request, ActionListener<Response> listener) {
-        ActionRequestValidationException validationException = request.validate();
+        final ActionRequestValidationException validationException;
+        try {
+            validationException = request.validate();
+        } catch (Exception e) {
+            assert false : new AssertionError("validating of request [" + request + "] threw exception", e);
+            logger.warn("validating of request [" + request + "] threw exception", e);
+            listener.onFailure(e);
+            return;
+        }
         if (validationException != null) {
             listener.onFailure(validationException);
             return;
         }
-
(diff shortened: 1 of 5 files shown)


ywelsch:
Hi Nhat. Just wondering if we can also safeguard the calling code (TransportAction) from similar mistakes in the future? Can be a separate PR

dnhatn:
@ywelsch I added a safeguard, then backed it out after checking other `validate` implementations. I will add it back.

ywelsch (approved):
LGTM. Thanks @dnhatn!

server/src/main/java/org/elasticsearch/action/support/TransportAction.java
```diff
@@ -40,12 +40,19 @@ protected TransportAction(String actionName, ActionFilters actionFilters, TaskMa
      * Use this method when the transport action should continue to run in the context of the current task
      */
     public final void execute(Task task, Request request, ActionListener<Response> listener) {
-        ActionRequestValidationException validationException = request.validate();
+        final ActionRequestValidationException validationException;
+        try {
+            validationException = request.validate();
+        } catch (Exception e) {
```
mayya-sharipova, line 46: I am trying to understand this part of code. If assertion always `false`, it means we will always throw `AssertionError`, right?  So the code after `assert false` will never be executed?

Or are we checking assertions only during testing mode?

dnhatn: @mayya-sharipova Here we fail with AssertionError in tests, but log a warning message and fail the request with an exception in production.

mayya-sharipova: @dnhatn Thanks for clarifying, I was also thinking the same.

mayya-sharipova (approved):
@dnhatn Thanks for the fix, was also educational for me to study your changes.

dnhatn:
@ywelsch @mayya-sharipova Thanks for review.

PR #88703, opened 2022-07-22
Update indices permissions to Enterprise Search service account

wangch079:
Enterprise Search service account needs to create/delete an index, index documents, therefore we need to update the index privileges to `["read", "write", "manage"]`

Related: https://github.com/elastic/elasticsearch/pull/85726


x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/service/ElasticServiceAccountsTests.java
```diff
@@ -372,6 +372,7 @@ public void testElasticEnterpriseSearchServerAccount() {
         });
 
         final IndexAbstraction elasticsearchIndex = mockIndexAbstraction("search-" + randomAlphaOfLengthBetween(1, 20));
+        assertThat(role.indices().allowedIndicesMatcher("indices:foo").test(elasticsearchIndex), is(false));
```
ywangd, line 375: Privilege `all` allows any action begining with `indices:`. So this assertion is incorrect.

ywangd (commented):
@wangch079 I'd appreciate if you could fill the PR description with some more details instead of just linking to other issues/PRs. Having it available directly on the PR avoid having to go through other linkes. This is especially helpful when you read it as commit messages. Also, the linked enterprise search issue is private and we do not recommend linking private issues from public PRs/issues. The private issue also does not have necessary context on why this permission change is necessary. It basically just says it is needed but I cannot tell the "why". A brief context would be helpful. Thanks!

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/authc/service/ElasticServiceAccountsTests.java
wangch079: Thanks @ywangd , updated the list of index privileges.

ioanatia (approved):

ywangd:
@wangch079 The updated change looks fine to me. I was on PTO yesterday and could not take a look. That said, in the future, I'd appreciate if you could please wait for approval from a @elastic/es-security member before hitting the merge button. Thanks!

PR #88706, opened 2022-07-22
Upgrade to lucene snapshot lucene-9.3.0-snapshot-b8d1fcfd0ec

iverase:
New snapshot that removed lucene93 codec so we need to go back to lucene92 codec.


romseygeek (approved):
LGTM

PR #88707, opened 2022-07-22
Avoid capturing SnapshotsInProgress$Entry in queue

DaveCTurner:
Today each time there's shards to snapshot we enqueue a lambda which
captures the current `SnapshotsInProgress$Entry`. This is a pretty
heavyweight object, possibly several MB in size, most of which is not
necessary to capture, and with concurrent snapshots across thousands of
shards we may enqueue many hundreds of slightly different such objects.
With this commit we compute a more efficient representation of the work
to be done by each task in the queue instead.

Relates #77466

### DIFF
--- a/docs/changelog/88707.yaml
+++ b/docs/changelog/88707.yaml
@@ -1,0 +1,5 @@
+pr: 88707
+summary: Avoid capturing `SnapshotsInProgress$Entry` in queue
+area: Snapshot/Restore
+type: bug
+issues: []
--- a/server/src/main/java/org/elasticsearch/snapshots/SnapshotShardsService.java
+++ b/server/src/main/java/org/elasticsearch/snapshots/SnapshotShardsService.java
@@ -52,6 +52,7 @@ import org.elasticsearch.transport.TransportResponseHandler;
 import org.elasticsearch.transport.TransportService;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


pxsalehi:
We've had some discussions kind of related to this for https://github.com/elastic/elasticsearch/pull/88209. 

I thought the lambda that captures the Entry, is only one per snapshot not per shard snapshot, and that snapshot is just one entry for all the indices that have shards for the index on the node, meaning the number of those lambdas grow with concurrent snapshots in the cluster, which I am guessing is much smaller than number of concurrent shard snapshots.

DaveCTurner:
You're right that the lambda is not a per-shard thing but with concurrent snapshots each snapshot could result in potentially many tasks. Each queue entry will only snapshot the shards that are in state `INIT`, but that could potentially be a single shard that just moved from state `QUEUED` because it completed in an earlier snapshot. The heap dump in the attached case shows this effect.

But also we permit up to 1000 concurrent snapshots by default, so if each task takes a few MB of heap then that's a few GB of heap unnecessarily burned on this queue.

DaveCTurner:
I wasn't aware of #88209 tho, that looks like a good move but not something I'd want to backport to 7.17. And it suffers the same issue about capturing the `Entry` rather than just the few bytes needed to describe the task. This'll conflict with that PR but hopefully it's not too hard to resolve those conflicts - we can perhaps rework this to make the conflicts less severe too.

pxsalehi (approved):
I understand. Thanks for the explanation. 

No, the changes to `SnapshotShardsService` in https://github.com/elastic/elasticsearch/pull/88209 are just for readability. Your PR also addresses that. I could just undo changes to `SnapshotShardsService` in my PR. There shouldn't be so much conflicts.

original-brownbear (approved):
LGTM, nice one! I guess I should analyse a data node heap dump for many-shards snapshotting benchmarks at some point :)

Also makes me wonder if we should make the snapshots-in-progress a proper diffable given how large these objects can get ...

DaveCTurner:
> Also makes me wonder if we should make the snapshots-in-progress a proper diffable given how large these objects can get 

Yes some more sharing would have helped here too, although the `ShardSnapshotStatus` map is big and changes a lot so wouldn't be shared much.

DaveCTurner:
I opened https://github.com/elastic/elasticsearch/issues/88732 to capture https://github.com/elastic/elasticsearch/pull/88707#pullrequestreview-1047780909

Issue #88682, opened 2022-07-21
[CI] CartesianShapeQueryTests testQueryRandomGeoCollection failing


benwtrent:
**Build scan:**
https://gradle-enterprise.elastic.co/s/p7n7etkicebme/tests/:x-pack:plugin:spatial:internalClusterTest/org.elasticsearch.xpack.spatial.search.CartesianShapeQueryTests/testQueryRandomGeoCollection

**Reproduction line:**
`./gradlew ':x-pack:plugin:spatial:internalClusterTest' --tests "org.elasticsearch.xpack.spatial.search.CartesianShapeQueryTests.testQueryRandomGeoCollection" -Dtests.seed=D42035A03986D727 -Dtests.locale=lt -Dtests.timezone=Asia/Baku -Druntime.java=17`

**Applicable branches:**
master

**Reproduces locally?:**
No

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.xpack.spatial.search.CartesianShapeQueryTests&tests.test=testQueryRandomGeoCollection

**Failure excerpt:**
```
org.elasticsearch.index.mapper.MapperParsingException: failed to parse field [geo] of type [shape]

  at __randomizedtesting.SeedInfo.seed([D42035A03986D727:CCF71E7B64748804]:0)
  at org.elasticsearch.index.mapper.AbstractGeometryFieldMapper.lambda$parse$1(AbstractGeometryFieldMapper.java:191)
  at org.elasticsearch.xpack.spatial.index.mapper.ShapeParser.parse(ShapeParser.java:40)
  at org.elasticsearch.index.mapper.AbstractGeometryFieldMapper.parse(AbstractGeometryFieldMapper.java:187)
  at org.elasticsearch.index.mapper.DocumentParser.parseObjectOrField(DocumentParser.java:409)
  at org.elasticsearch.index.mapper.DocumentParser.parseObject(DocumentParser.java:461)
  at org.elasticsearch.index.mapper.DocumentParser.innerParseObject(DocumentParser.java:313)
  at org.elasticsearch.index.mapper.DocumentParser.parseObjectOrNested(DocumentParser.java:278)
  at org.elasticsearch.index.mapper.DocumentParser.internalParseDocument(DocumentParser.java:127)
  at org.elasticsearch.index.mapper.DocumentParser.parseDocument(DocumentParser.java:79)
[...]

PR #79152, opened 2021-10-14
Separate feature usage tracking for FLS and DLS

ywangd:
This PR ensures the feature usage is tracked separately for FLS and DLS.
Even though they still have identical license requirement. Separate
usage tracking is helpful for things like reporting.

The usage tracking is triggered if the incoming request touches any
indices that have DLS/FLS constraint. This is done in an request
interceptor which means it will only trigger if the request has passed
authorization. So if the user only has read with DLS/FLS for an index
and issues a write request, the request will fail at authorization and
usage tracking will not be triggered.

The request interceptor is chosen to be the single place for usage
tracking. As a result, license checking in all other places does not
trigger usage tracking (regardless whether they previously trigger it or
not).

PR #78378, opened 2021-09-28
Tighten API key behaviour with DLS and incompatible license

ywangd:
When the cluster license is incompatible with DLS and FLS, instead of
failing open, API keys now fail closed by throwing an error if any of
the target indices has DLS/FLS protection.

PR #70191, opened 2021-03-10
Support shard request cache for queries with DLS and FLS

ywangd:
Today the shard request cache is disabled by RequestInterceptor when it detects any DLS/FLS is involved in the query. We'd like to enable the cache with changes proposed by this PR. This is labelled as "draft" because we would like to validate the overall approach first before polishing it. In a nutshell, the changes are as the follows:
1. Add a new `cacheModifier` field to `ShardSearchRequest`. 
2. The `ShardSearchRequestInterceptor` sets the value of `cacheModifier` by essentially serialising the DLS queries and FLS field definitions.
3. The `cacheModifier` field becomes part of the `ShardSearchRequeset#cacheKey` method. Hence two otherwise identical queries will be cached with different keys if their `cacheModifiers` are different.

@tvernum came up with the overall idea and did most of the code changes. I have only done some tweaks so far. But I'll keep working on it and see it through (if we agree on the proposed changes).

Since this is just a draft, there are many things we glossed over. Other than tests and docs, here are a few things that we talked about didn't have time to work on:
* How does this work with aliases? For example, two aliases pointing to the same concrete index but have different DLS/FLS applied to them.
* We didn't attempt to normalise the DLS queries and FLS field definitions. Hence it is possible that two functional identical queries would result into different cache entries. The same thing can happen to FLS field definition as well. We generally consider this acceptable because it seems the overall cache stratgey is safety over efficiency.
* The computation of CacheKey itself may need to be cached if it is proved to be expensive enough
* We haven't evaluated whether increase the CacheKey size would negatively impact the existing cache
* Cache efficiency - Queries with DLS/FLS will increase the variety of queries which could lead to higher cache eviction rate or even cache thrashing (thinking templated DLS query). But we are not sure whether the concern is justified because similar issue could already happen today if users send in large number of different queries.

Resolves: #44581

PR #88708, opened 2022-07-22
Lazy compute the index access control

slobodanadamovic:
The access control was always computed eagerly, even in cases when 
it was not necessary (e.g. shard is not accessed on the node doing 
the authorization or in cases when authorization is denied). 
This commit defers the computation to when it's really needed and tries 
to avoid that the actual work is done on the network worker thread.


slobodanadamovic:
This PR is still a draft as it depends on changes made in https://github.com/elastic/elasticsearch/pull/88662.

slobodanadamovic:
The `elasticsearch-ci/part-2` build failure is caused by the same issue as reported in [[CI] CartesianShapeQueryTests testQueryRandomGeoCollection failing](https://github.com/elastic/elasticsearch/issues/88682).

slobodanadamovic:

slobodanadamovic:

slobodanadamovic:
@elasticmachine update branch

slobodanadamovic:
@elasticmachine update branch

jakelandis:
I stepped through an example with DLS and an example without DLS for a search request ...while i can see the new deferred/lazy loading, in both cases the supplier was realized/loaded as part of the SearchRequest task on the coordinating node (specifically in org.elasticsearch.xpack.security.authz.interceptor.FieldAndDocumentLevelSecurityRequestInterceptor#intercept and org.elasticsearch.xpack.security.authz.interceptor.DlsFlsLicenseRequestInterceptor#intercept) . I think this change only makes sense if deferring/lazy loading allow us to skip some expensive work or move the work to a better thread/context. 

As-is it seems that we are just delaying the same work till a bit later in the processing. (moving from authz code to request interceptors, but still on transport worker).  If I understand correctly, this information is is pertinent on the coordinating node for license and feature checks (fail fast), and pertinent on the shard level actions to find the dls/fls to apply. So I don't think we can skip the checks on the coordinating action nor or on the chlid actions (at least for search requests). 

--- (tangential) ---
Also, for my own education... part of this data structure the (now) lazy work contributes (org.elasticsearch.xpack.core.security.authz.accesscontrol.IndicesAccessControl.IndexAccessControl) determines access per index is granted or not ... do you know exactly how the granted information is used ? It appears to be tracked but I can't see how granted it is actually used (or how it could ever be false) as the requested index names have been replaced on the request object.

albertzaharovits:
We discussed but I wanted to verify it myself. It's possible to NOT evaluate the access control on the coordinating node:

```
diff --git a/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/DlsFlsLicenseRequestInterceptor.java b/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/DlsFlsLicenseRequestInterceptor.java
index 7fcf855d608..47c6eb81fdc 100644
--- a/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/DlsFlsLicenseRequestInterceptor.java
+++ b/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/DlsFlsLicenseRequestInterceptor.java
@@ -56,46 +56,52 @@ public class DlsFlsLicenseRequestInterceptor implements RequestInterceptor {
                 final IndicesAccessControl indicesAccessControl = threadContext.getTransient(INDICES_PERMISSIONS_KEY);
                 if (indicesAccessControl != null) {
                     final XPackLicenseState frozenLicenseState = licenseState.copyCurrentLicenseState();
-                    final IndicesAccessControl.DlsFlsUsage dlsFlsUsage = indicesAccessControl.getFieldAndDocumentLevelSecurityUsage();
-                    boolean incompatibleLicense = false;
-                    if (dlsFlsUsage.hasFieldLevelSecurity()) {
-                        logger.debug(
-                            () -> format(
-                                "User [%s] has field level security on [%s]",
-                                requestInfo.getAuthentication(),
-                                indicesAccessControl.getIndicesWithFieldLevelSecurity()
-                        );
-                        if (false == FIELD_LEVEL_SECURITY_FEATURE.check(frozenLicenseState)) {
-                            incompatibleLicense = true;
+                    if (logger.isDebugEnabled()) {
+                        IndicesAccessControl.DlsFlsUsage dlsFlsUsage = indicesAccessControl.getFieldAndDocumentLevelSecurityUsage();
+                        if (dlsFlsUsage.hasFieldLevelSecurity()) {
+                            logger.debug(
+                                () -> format(
+                                    "User [%s] has field level security on [%s]",
+                                    requestInfo.getAuthentication(),
+                                    indicesAccessControl.getIndicesWithFieldLevelSecurity()
+                                )
+                            );
+                        }
+                        if (dlsFlsUsage.hasDocumentLevelSecurity()) {
+                            logger.debug(
+                                () -> format(
+                                    "User [%s] has document level security on [%s]",
+                                    requestInfo.getAuthentication(),
+                                    indicesAccessControl.getIndicesWithDocumentLevelSecurity()
+                                )
+                            );
                         }
                     }
-                    if (dlsFlsUsage.hasDocumentLevelSecurity()) {
-                        logger.debug(
-                            () -> format(
-                                "User [%s] has document level security on [%s]",
-                                requestInfo.getAuthentication(),
-                                indicesAccessControl.getIndicesWithDocumentLevelSecurity()
-                        );
-                        if (false == DOCUMENT_LEVEL_SECURITY_FEATURE.check(frozenLicenseState)) {
+                    if (false == DOCUMENT_LEVEL_SECURITY_FEATURE.check(frozenLicenseState) ||
+                        false == FIELD_LEVEL_SECURITY_FEATURE.check(frozenLicenseState)) {
+                        boolean incompatibleLicense = false;
+                        IndicesAccessControl.DlsFlsUsage dlsFlsUsage = indicesAccessControl.getFieldAndDocumentLevelSecurityUsage();
+                        if (false == DOCUMENT_LEVEL_SECURITY_FEATURE.check(frozenLicenseState) &&
+                            dlsFlsUsage.hasDocumentLevelSecurity()) {
+                            incompatibleLicense = true;
+                        } else if (false == FIELD_LEVEL_SECURITY_FEATURE.check(frozenLicenseState) &&
+                            dlsFlsUsage.hasFieldLevelSecurity()) {
                             incompatibleLicense = true;
                         }
-                    }
-                    if (incompatibleLicense) {
-                        final ElasticsearchSecurityException licenseException = LicenseUtils.newComplianceException(
-                            "field and document level security"
-                        );
-                        licenseException.addMetadata(
-                            "es.indices_with_dls_or_fls",
-                            indicesAccessControl.getIndicesWithFieldOrDocumentLevelSecurity()
-                        );
-                        listener.onFailure(licenseException);
-                        return;
+                        if (incompatibleLicense) {
+                            final ElasticsearchSecurityException licenseException = LicenseUtils.newComplianceException(
+                                "field and document level security"
+                            );
+                            licenseException.addMetadata(
+                                "es.indices_with_dls_or_fls",
+                                indicesAccessControl.getIndicesWithFieldOrDocumentLevelSecurity()
+                            );
+                            listener.onFailure(licenseException);
+                            return;
+                        }
                     }
                 }
             }
-
         }
         listener.onResponse(null);
     }
diff --git a/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/SearchRequestInterceptor.java b/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/SearchRequestInterceptor.java
index 8e8f2bd6c77..f50c4bb5777 100644
--- a/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/SearchRequestInterceptor.java
+++ b/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/SearchRequestInterceptor.java
@@ -74,7 +74,8 @@ public class SearchRequestInterceptor extends FieldAndDocumentLevelSecurityReque

     @Override
     public boolean supports(IndicesRequest request) {
-        return request instanceof SearchRequest;
+        return request instanceof SearchRequest && ((SearchRequest) request).source() != null &&
+            (((SearchRequest) request).source().suggest() != null || ((SearchRequest) request).source().profile());
     }

     // package private for test
```

albertzaharovits:
> Also, for my own education... part of this data structure the (now) lazy work contributes (org.elasticsearch.xpack.core.security.authz.accesscontrol.IndicesAccessControl.IndexAccessControl) determines access per index is granted or not ... do you know exactly how the granted information is used ? It appears to be tracked but I can't see how granted it is actually used (or how it could ever be false) as the requested index names have been replaced on the request object.

I think it's possible to remove the granted flag on the access control. It is used in a few places (ie authz child actions, and authz bulk items) but I think we can find ways to not require it in those cases as well.

albertzaharovits (commented):
As we discussed it only makes sense to defer the access control computation if it is avoided in at least some common cases. I'm reluctant to merge this PR as is because it adds complication with no gain.
I have suggested a rough code diff that should avoid the access control computation on the coordinating node.

ywangd:
The `DlsFlsLicenseRequestInterceptor` does two things:
1. Prevent API keys with DLS/FLS from being used if license is not compatible (https://github.com/elastic/elasticsearch/pull/78378)
2. Feature usage tracking for DLS and FLS (https://github.com/elastic/elasticsearch/pull/79152)

The [diff](https://github.com/elastic/elasticsearch/pull/88708#issuecomment-1219327330) that @albertzaharovits proposed should work for 1. But it seems to need further tweaking for 2. The usage tracking requires:
1. Separate tracking for DLS and FLS
2. Track usage when license is compatible, e.g. Cloud deployments always have compatible licenses and we need track actual feature usages.
3. Track usage only when the feature is actually used. 

(Item 2 and 3 were the reason why existing code checks indicesAccessControl before license)

The interceptor was chosen to be the place for usage tracking because it is a choking point for all requests. So checking in the interceptor avoids having to check it in multiple places and is more future proof. But it probably makes authZ optimisation harder. In that case, we might have to think of different ways of handling usage tracking. I don't have strong opinions. Sorry to add more requirements to this issue.

albertzaharovits:
@ywangd Good point about feature usage tracking when license is compatible and the feature is actually used.
I would argue that the DLS/FLS feature is not actually used on the coordinating node.
In order to have some leeway in defering the access control map computation, would it be reasonable to track the DLS/FLS feature usage on the data nodes, ie inside the `SecurityIndexReaderWrapper`?

slobodanadamovic:

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/accesscontrol/IndicesAccessControl.java
```diff
@@ -36,11 +38,19 @@ public class IndicesAccessControl {
     public static final IndicesAccessControl DENIED = new IndicesAccessControl(false, Collections.emptyMap());
 
     private final boolean granted;
-    private final Map<String, IndexAccessControl> indexPermissions;
+    private final Supplier<Map<String, IndexAccessControl>> indexPermissions;
```
albertzaharovits, line 41: For clarity's sake, please rename `indexPermissions` to `indexPermissionsSupplier`, and change the field type to `CachedSupplier`. It helps with readability if the more concrete type is obvious.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/accesscontrol/IndicesAccessControl.java
```diff
@@ -36,11 +38,19 @@ public class IndicesAccessControl {
...
 
     public IndicesAccessControl(boolean granted, Map<String, IndexAccessControl> indexPermissions) {
+        this(granted, () -> Objects.requireNonNull(indexPermissions));
+    }
+
+    public IndicesAccessControl(boolean granted, Supplier<Map<String, IndexAccessControl>> indexPermissionsSupplier) {
         this.granted = granted;
-        this.indexPermissions = Objects.requireNonNull(indexPermissions);
+        this.indexPermissions = Objects.requireNonNull(indexPermissionsSupplier);
```
albertzaharovits, line 49: To me it looks clearer if we wrap the Supplier parameter in the `CachedSupplier` type at this stage, rather than delegate this responsability to the caller.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/permission/IndicesPermission.java
```diff
@@ -373,24 +374,24 @@ public IndicesAccessControl authorize(
         }
 
         final Map<String, IndexResource> resources = Maps.newMapWithExpectedSize(requestedIndicesOrAliases.size());
-        int totalResourceCount = 0;
+        final int[] totalResourceCountHolder = new int[] { 0 };
```
albertzaharovits, line 377: nit: please use an `AtomicInteger` this looks like a hack.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/integration/DocumentAndFieldLevelSecurityTests.java
```diff
@@ -42,7 +42,8 @@
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 
-public class DocumentAndFieldLevelSecurityTests extends SecurityIntegTestCase {
+@ESIntegTestCase.ClusterScope(numClientNodes = 1)
```
albertzaharovits, line 45: Can we move the tests in this class to the newly introduced [DocumentLevelSecurityFeatureUsageTests.java](https://github.com/elastic/elasticsearch/pull/88708/files#diff-d201e525f0cefb113fff77d4e78fe2bbcaa4d5d77267849e0ac34c8db8c1a888) one.

This test class already tests too much, and I'm worried that if we restrict the configuration (the number of client nodes) we might miss some esotheric test condition.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/SearchRequestInterceptor.java
```diff
@@ -69,7 +69,11 @@ void disableFeatures(
 
     @Override
     public boolean supports(IndicesRequest request) {
-        return request instanceof SearchRequest;
+        if (request instanceof SearchRequest searchRequest) {
+            return searchRequest.source() != null && (searchRequest.source().suggest() != null || searchRequest.source().profile());
+        } else {
+            return false;
+        }
```
albertzaharovits, line 76: This is a problem.
The interceptor also disables the request cache:
```
        final SearchRequest request = (SearchRequest) indicesRequest;
        if (hasRemoteIndices(request)) {
            request.requestCache(false);
        }
```

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java
```diff
@@ -613,6 +622,19 @@ public void onResponse(Void unused) {
...
+            || indicesAccessControl == IndicesAccessControl.ALLOW_NO_INDICES
+            || indicesAccessControl == IndicesAccessControl.DENIED
+            || indicesAccessControl == IndicesAccessControl.allowAll()
+            || indicesAccessControl.getClass().equals(IndicesAccessControl.class) == false) {
+            // Wrap only if it's not one of the statically defined nor already wrapped.
+            return indicesAccessControl;
+        } else {
+            return indicesAccessControlWrapper.wrap(indicesAccessControl);
+        }
+    }
```
albertzaharovits, line 636: I think it would be easier to follow if this method is a static one inside the DlsFlsFeatureTrackingIndicesAccessControlWrapper class.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java
```diff
@@ -613,6 +622,19 @@ public void onResponse(Void unused) {
         }
     }
 
+    private IndicesAccessControl wrapIndicesAccessControl(IndicesAccessControl indicesAccessControl) {
+        if (indicesAccessControl == null
+            || indicesAccessControl == IndicesAccessControl.ALLOW_NO_INDICES
+            || indicesAccessControl == IndicesAccessControl.DENIED
+            || indicesAccessControl == IndicesAccessControl.allowAll()
+            || indicesAccessControl.getClass().equals(IndicesAccessControl.class) == false) {
```
albertzaharovits, line 630: The last condition is not exactly a test for "already wrapped". It should test if the class is already an instance of `DlsFlsFeatureTrackingIndicesAccessControlWrapper`.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/accesscontrol/wrapper/IndicesAccessControlWrapper.java
```diff
@@ -0,0 +1,20 @@
...
+
+package org.elasticsearch.xpack.security.authz.accesscontrol.wrapper;
+
+import org.elasticsearch.xpack.core.security.authz.accesscontrol.IndicesAccessControl;
+
+/**
+ * A wrapper which allows adding additional functionality and intercepting calls to {@link IndicesAccessControl} methods.
+ */
+@FunctionalInterface
+public interface IndicesAccessControlWrapper {
```
albertzaharovits, line 16: An interface for a single class with a single method seems like overkill.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/BulkShardRequestInterceptor.java
```diff
@@ -48,17 +49,20 @@ public void intercept(
         ActionListener<Void> listener
     ) {
         if (requestInfo.getRequest()instanceof BulkShardRequest bulkShardRequest) {
+            final boolean isDlsLicensed = DOCUMENT_LEVEL_SECURITY_FEATURE.checkWithoutTracking(licenseState);
+            final boolean isFlsLicensed = FIELD_LEVEL_SECURITY_FEATURE.checkWithoutTracking(licenseState);
+            if (isDlsLicensed == false && isFlsLicensed == false) {
+                listener.onResponse(null);
+                return;
+            }
```
albertzaharovits, line 57: I understand that the intention is to only check the DLS/FLS permissions if the feature is licensed such that we avoid computing those otherwise. But this is hard to notice at this level.
What do you think if we move the check inside a `IndicesAccessControl#hasFieldLevelSecurity(String indexName)`? It should return `false` if it is not licensed, and otherwise trigger the access control computation and track the feature usage.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/DlsFlsLicenseRequestInterceptor.java
```diff
@@ -56,46 +56,52 @@ public void intercept(
...
-                                "User [%s] has document level security on [%s]",
-                                requestInfo.getAuthentication(),
-                                indicesAccessControl.getIndicesWithDocumentLevelSecurity()
-                            )
-                        );
-                        if (false == DOCUMENT_LEVEL_SECURITY_FEATURE.check(frozenLicenseState)) {
+                    if (false == DOCUMENT_LEVEL_SECURITY_FEATURE.checkWithoutTracking(frozenLicenseState)
+                        || false == FIELD_LEVEL_SECURITY_FEATURE.checkWithoutTracking(frozenLicenseState)) {
+                        boolean incompatibleLicense = false;
+                        IndicesAccessControl.DlsFlsUsage dlsFlsUsage = indicesAccessControl.getFieldAndDocumentLevelSecurityUsage();
```
albertzaharovits, line 83: This triggers the access control computatin but does not track the feature usage.
The code around it makes it to not be a problem.
BUT it would be much easier to follow if we knew that feature usage is tracked whenever the access control map is computed.
Only in this case are the IT tests that check the feature tracker also a test for the optimization that avoids the computation on coordinating nodes.

So my suggestion would be to hook the feature tracking inside the `DlsFlsFeatureTrackingIndicesAccessControlWrapper#getAllIndexPermissions` method. What do you think?

albertzaharovits (commented):
I think this looks promising!

I left a couple of comments around the areas of improvement. Mainly I would like to focus on the interceptors such that they do not trigger the access control map computation, and if they do that the FLS/DLS feature usage is tracked, which helps with testing.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/accesscontrol/wrapper/IndicesAccessControlWrapper.java
slobodanadamovic: There are two implementations: [anonymous](https://github.com/elastic/elasticsearch/blob/61e3075b77dbbe28b9f511c253622b759ea04a77/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationService.java#L179) and `DlsFlsFeatureTrackingIndicesAccessControlWrapper`. I don't feel strongly with this approach. I'm okay to remove it.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/SearchRequestInterceptor.java
slobodanadamovic: Good catch. I introduced a new interceptor (`SearchRequestCacheDisablingInterceptor`) which only disables the cache if the request has remote indices.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/accesscontrol/wrapper/IndicesAccessControlWrapper.java
slobodanadamovic: Removed https://github.com/elastic/elasticsearch/pull/88708/commits/d57f820122a39bb39cacdfda1ac88e6979edf28b

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/BulkShardRequestInterceptor.java
slobodanadamovic: I think this would cause issue in `DlsFlsLicenseRequestInterceptor`. There we check the compatibility of license and DLS/FLS usage.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/integration/DocumentAndFieldLevelSecurityTests.java
slobodanadamovic: Done. I also introduced a new test class `FieldLevelSecurityFeatureUsageTests`, to avoid changing `FieldLevelSecurityTests` for the same reason.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/accesscontrol/wrapper/DlsFlsFeatureTrackingIndicesAccessControlWrapper.java
```diff
@@ -0,0 +1,72 @@
...
+            return iac;
+        }
+
+        private void trackUsage(IndexAccessControl iac) {
+            if (iac.getDocumentPermissions().hasDocumentLevelPermissions()) {
+                DOCUMENT_LEVEL_SECURITY_FEATURE.check(licenseState);
+            }
+            if (iac.getFieldPermissions().hasFieldLevelSecurity()) {
+                FIELD_LEVEL_SECURITY_FEATURE.check(licenseState);
+            }
```
albertzaharovits, line 69: What do you think if we assert that the features are licensed and that we log WARN if it isn't?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/BulkShardRequestInterceptor.java
albertzaharovits: You're right, it's not a viable suggestion.

I think it's good the way you have it here, to first check if the features are licensed and only afterwards proceed to verify them (rather than the reverse in the original code).

One small nit though, could you please incorporate the `isDlsLicensed || isFlsLicensed` in the if above, I think it's easier on the eyes.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/DlsFlsLicenseRequestInterceptor.java
albertzaharovits: We discussed on another channel, my suggestion is not feasible.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/FieldAndDocumentLevelSecurityRequestInterceptor.java
```diff
@@ -50,8 +51,9 @@ public void intercept(
         if (requestInfo.getRequest()instanceof IndicesRequest indicesRequest
             && false == TransportActionProxy.isProxyAction(requestInfo.getAction())) {
             // TODO: should we check is DLS/FLS feature allowed here
-            if (supports(indicesRequest)) {
-                final boolean isDlsLicensed = DOCUMENT_LEVEL_SECURITY_FEATURE.checkWithoutTracking(licenseState);
+            final boolean isDlsLicensed = DOCUMENT_LEVEL_SECURITY_FEATURE.checkWithoutTracking(licenseState);
+            final boolean isFlsLicensed = FIELD_LEVEL_SECURITY_FEATURE.checkWithoutTracking(licenseState);
+            if ((isDlsLicensed || isFlsLicensed) && supports(indicesRequest)) {
```
albertzaharovits, line 56: nit: here I also think there would be easier to merge these two `if` conditions in a single one.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/SearchRequestInterceptor.java
albertzaharovits: LGTM!

Minor additional nit: can you please extract the two conditions `source != null && source.suggest() != null` and `source != null && source.profile()` in two private methods that you then reuse in the `supports` and `disableFeatures` methods?

albertzaharovits (approved):
LGTM nice job!

Left a couple of small issues. Please look into those, though they don't require me to take another look.

albertzaharovits:
@ycombinator We discussed some time ago that I'll ping you when we get ready to merge this change so that you can validate on the cloud side that all is well.

This change makes DLS/FLS feature tracking not report usage on the coordinating-only nodes, in most cases (when the feature is licensed). It will continue to report usage on the data nodes, as usual.

ycombinator:
> @ycombinator We discussed some time ago that I'll ping you when we get ready to merge this change so that you can validate on the cloud side that all is well.
> 
> This change makes DLS/FLS feature tracking not report usage on the coordinating-only nodes, in most cases (when the feature is licensed). It will continue to report usage on the data nodes, as usual.

Thanks for the ping, @albertzaharovits.  I want to make sure I understand the impact of the change correctly.  Before this change DLS/FLS feature tracking (if the feature was licensed) was being reported by all nodes?  And after this change this will only be reported by data nodes?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/interceptor/SearchRequestCacheDisablingInterceptor.java
```diff
@@ -0,0 +1,51 @@
...
+import org.elasticsearch.license.XPackLicenseState;
+import org.elasticsearch.threadpool.ThreadPool;
+import org.elasticsearch.xpack.core.security.authz.accesscontrol.IndicesAccessControl;
+
+import java.util.Arrays;
+import java.util.Map;
+
+import static org.elasticsearch.transport.RemoteClusterAware.REMOTE_CLUSTER_INDEX_SEPARATOR;
+
+public class SearchRequestCacheDisablingInterceptor extends FieldAndDocumentLevelSecurityRequestInterceptor {
```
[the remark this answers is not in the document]
slobodanadamovic: I tried to track the change down in order to better understand it, and if I parse [this comment](https://github.com/elastic/elasticsearch/pull/70191#issuecomment-855649284) correctly, we do want to extend `FieldAndDocumentLevelSecurityRequestInterceptor` and disable the cache only if DLS/FLS are used. But, maybe @ywangd could help us understand this better?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/accesscontrol/wrapper/DlsFlsFeatureTrackingIndicesAccessControlWrapper.java
slobodanadamovic: Adding `assert` sounds like a good idea. I'm not sure if logging a warning makes sense here, since `DlsFlsLicenseRequestInterceptor` should fail the request before even reaching this code.

In theory, if the license gets changed immediately after the `DlsFlsLicenseRequestInterceptor` has been run, we might spam the logs with WARN messages and I'm not sure if it's worth it since the very next requests will be simply rejected.

slobodanadamovic:
@ycombinator 

>  I want to make sure I understand the impact of the change correctly. Before this change DLS/FLS feature tracking (if the feature was licensed) was being reported by all nodes? And after this change this will only be reported by data nodes?

That is correct. 

With this change, the DLS/FLS feature usage will be reported by data nodes (as it did before), with just a small exception that it might not be reported by coordinating only nodes in some cases. Hence, I think it's fine to assume that coordinating-only nodes will not report the feature usage anymore and only data nodes will report it.

ycombinator:
> With this change, the DLS/FLS feature usage will be reported by data nodes (as it did before), with just a small exception that it might not be reported by coordinating only nodes in some cases. Hence, I think it's fine to assume that coordinating-only nodes will not report the feature usage anymore and only data nodes will report it.

Thanks for clarifying, @slobodanadamovic.  Just to confirm, this change will not impact the results of DLS/FLS feature usage for a cluster, right?  Meaning, whatever DLS/FLS feature usage the cluster was reporting before, it will report after this change as well --- the only difference is _what nodes_ may report the usage now, yes?

slobodanadamovic:
> > With this change, the DLS/FLS feature usage will be reported by data nodes (as it did before), with just a small exception that it might not be reported by coordinating only nodes in some cases. Hence, I think it's fine to assume that coordinating-only nodes will not report the feature usage anymore and only data nodes will report it.
> 
> Thanks for clarifying, @slobodanadamovic. Just to confirm, this change will not impact the results of DLS/FLS feature usage for a cluster, right? Meaning, whatever DLS/FLS feature usage the cluster was reporting before, it will report after this change as well --- the only difference is _what nodes_ may report that usage now, yes?

Correct.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/accesscontrol/wrapper/DlsFlsFeatureTrackingIndicesAccessControlWrapper.java
slobodanadamovic: We synced and concluded that it is not necessary to log WARN messages.

slobodanadamovic:
@elasticmachine run elasticsearch-ci/part-3-fips

slobodanadamovic:
@elasticmachine run elasticsearch-ci/part-3-fips

slobodanadamovic:
@elasticmachine update branch

PR #88710, opened 2022-07-22
Replace `stopRandomNode` call with more generic calls

idegtiarenko:
`stopRandomNode` with predicate is rather specific call. This change replaces it
 with more generic `stopNode` and `getNodeNameThat`.


test/framework/src/main/java/org/elasticsearch/test/InternalTestCluster.java
```diff
@@ -1997,30 +1982,18 @@ public String getMasterName(@Nullable String viaNode) {
...
+        return getNodeNameThat(ignored -> true);
     }
 
     /**
-     * Returns the name of the selected health node in the cluster and executes the request via the node specified
-     * in the viaNode parameter. If viaNode isn't specified a random node will be picked to send the request to. If
-     * there is no selected health node it returns null.
+     * @return the name of a random node in a cluster that match the {@code predicate}
      */
-    public String getHealthNodeName(@Nullable String viaNode) {
```
DaveCTurner, line 2011: Pinging @gmarouli about the removal of this unused code - not sure if the intention was to use it in a future PR or not, but if so then that PR will have to bring it back.

DaveCTurner (approved):
LGTM

PR #88715, opened 2022-07-22
Speed up DocumentParser.innerParseObject

original-brownbear:
Making this a little smaller so it inlines more often and more importantly
avoid getting the current field name redundantly before entering the loop which
is not free when the parser is a `DotExpandingXContentParser`.
This change gives a significant boost to `BeatsMapperBenchmark.benchmarkParseKeywordFields`.

Results below are pretty stable, running with a large number of iterations.

before:

```
Result "org.elasticsearch.benchmark.index.mapper.BeatsMapperBenchmark.benchmarkParseKeywordFields":
  9039.270 ±(99.9%) 28.001 ns/op [Average]
  (min, avg, max) = (9008.778, 9039.270, 9062.246), stdev = 18.521
  CI (99.9%): [9011.269, 9067.271] (assumes normal distribution)
```

after:

```
Result "org.elasticsearch.benchmark.index.mapper.BeatsMapperBenchmark.benchmarkParseKeywordFields":
  8645.649 ±(99.9%) 53.677 ns/op [Average]
  (min, avg, max) = (8568.319, 8645.649, 8688.210), stdev = 35.504
  CI (99.9%): [8591.972, 8699.327] (assumes normal distribution)
```

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
+++ b/server/src/main/java/org/elasticsearch/index/mapper/DocumentParser.java
@@ -294,17 +294,18 @@ public final class DocumentParser {
 
     private static void innerParseObject(DocumentParserContext context, ObjectMapper mapper) throws IOException {
 
-        XContentParser.Token token = context.parser().currentToken();
-        String currentFieldName = context.parser().currentName();
+        final XContentParser parser = context.parser();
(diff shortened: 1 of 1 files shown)


romseygeek (approved):
LGTM

original-brownbear:
Thanks Alan!

PR #88639, opened 2022-07-20
Introduce max headroom for disk watermark stages

kingherc:
Introduce max headroom settings for the low, high, and flood disk watermark stages, similar to the existing max headroom setting for the flood stage of the frozen tier. Also, convert the disk watermarks to RelativeByteSizeValue, similar to the existing setting for the flood stage of the frozen tier.

Introduce new max headrooms in HealthMetadata and in ReactiveStorageDeciderService.

Add multiple tests in DiskThresholdDeciderUnitTests, DiskThresholdDeciderTests and DiskThresholdMonitorTests.

Fixes #81406

PR #88719, opened 2022-07-22
Convert disk watermarks to RelativeByteSizeValues

kingherc:
Similar to the existing watermark setting for the frozen tier.

Pre-requisite for PR [88639](https://github.com/elastic/elasticsearch/pull/88639) that plans to introduce max headroom
settings for the disk watermarks, similar to the frozen tier max
headroom setting.


server/src/main/java/org/elasticsearch/common/unit/RelativeByteSizeValue.java
```diff
@@ -90,7 +90,7 @@ public static RelativeByteSizeValue parseRelativeByteSizeValue(String value, Str
 
     public String getStringRep() {
         if (ratio != null) {
-            return ratio.toString();
+            return RatioValue.formatNoTrailingZerosPercent(ratio.getAsPercent());
```
DaveCTurner, line 93: Seems weird that this isn't an instance method on `RatioValue`.

CONTRIBUTING.md
```diff
@@ -604,7 +604,7 @@ threshold has been breached:
 
     logger.warn(
         "flood stage disk watermark [{}] exceeded on {}, all indices on this node will be marked read-only",
-        diskThresholdSettings.describeFloodStageThreshold(),
+        diskThresholdSettings.describeFloodStageThreshold(total, false),
```
DaveCTurner, line 607: wow bonus points for thoroughness :)

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderServiceTests.java
```diff
@@ -355,39 +355,6 @@ public void testNodeLockSplitClone() {
         assertThat(createAllocationState(shardSizes, clusterState).maxNodeLockedSize(), equalTo(sourceSize * 2));
     }
 
-    public void testNodeSizeForDataBelowLowWatermark() {
```
DaveCTurner, line 358: Hmm this test still has meaning I think, although it fails sometimes as currently implemented. Can we fix it rather than just removing it? At least we can assert that we're no more than 1 off?

x-pack/plugin/searchable-snapshots/src/test/java/org/elasticsearch/xpack/searchablesnapshots/cache/shared/FrozenCacheServiceTests.java
```diff
@@ -210,7 +210,9 @@ public void testDecay() throws IOException {
     }
 
     public void testCacheSizeRejectedOnNonFrozenNodes() {
-        String cacheSize = randomBoolean() ? new ByteSizeValue(size(500)).getStringRep() : new RatioValue(between(1, 100)).toString();
+        String cacheSize = randomBoolean()
+            ? new ByteSizeValue(size(500)).getStringRep()
+            : RatioValue.formatNoTrailingZerosPercent(new RatioValue(between(1, 100)).getAsPercent()).toString();
```
DaveCTurner, line 215: Think this is calling `String#toString()`?

```suggestion
            : RatioValue.formatNoTrailingZerosPercent(new RatioValue(between(1, 100)).getAsPercent());
```

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
```diff
@@ -130,143 +125,89 @@ public DiskThresholdSettings(Settings settings, ClusterSettings clusterSettings)
...
+                        "high disk watermark ["
+                            + high.getStringRep()
+                            + "] less than flood stage disk watermark ["
+                            + flood.getStringRep()
+                            + "]"
+                    );
+                }
+            } else {
+                final String message = String.format(
+                    Locale.ROOT,
```
DaveCTurner, line 186: nit (not introduced in this change but since we're touching this line anyway...):

```suggestion
                final String message = Strings.format(
```

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
```diff
@@ -275,88 +216,78 @@ private void setEnabled(boolean enabled) {
...
+        return ByteSizeValue.ofBytes(total.getBytes() - watermark.calculateValue(total, maxHeadroom).getBytes());
     }
 
-    /**
-     * Gets the raw (uninterpreted) high watermark value as found in the settings.
-     */
-    public String getHighWatermarkRaw() {
-        return highWatermarkRaw;
+    public ByteSizeValue getFreeBytesThresholdLowStage(ByteSizeValue total) {
+        return getFreeBytesThreshold(total, lowStageWatermark, ByteSizeValue.ofBytes(-1));
```
DaveCTurner, line 248: nit:

```suggestion
        return getFreeBytesThreshold(total, lowStageWatermark, ByteSizeValue.MINUS_ONE);
```

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
```diff
@@ -367,128 +298,68 @@ public TimeValue getRerouteInterval() {
...
+            CLUSTER_ROUTING_ALLOCATION_DISK_FLOOD_STAGE_FROZEN_MAX_HEADROOM_SETTING.getKey()
+        );
     }
 
-    // Checks that a value is definitely not a percentage by testing if it ends on `b` which implies that it is probably a byte size value
-    // instead. This is used to make setting validation skip attempting to parse a value as a percentage/ration for the settings in this
-    // class that accept either a byte size value. The main motivation of this method is to make tests faster. Some tests call this method
-    // frequently when starting up internal cluster nodes and using exception throwing and catching when trying to parse as a ratio as a
-    // means of identifying that a string is not a ratio is quite slow.
-    private static boolean definitelyNotPercentage(String value) {
```
DaveCTurner, line 491: I think we should keep this optimization (i.e. implement it in `RelativeByteSizeValue`) since it's pretty powerful in tests.

kingherc:
Hi @DaveCTurner , managed to extricate the parts of https://github.com/elastic/elasticsearch/pull/88639 that related to converting the disk watermarks to RelativeByteSizeValues. Feel free to review them. Please note a couple of parts might feel a bit odd without the max headroom settings, but they prepare the ground for the other PR. E.g., there's a max headroom parameter in some calculation functions that are currently only useful for the frozen tier flood max headroom setting, but they will also be used by the new max headroom settings in the other PR.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
```diff
@@ -367,128 +298,68 @@ public TimeValue getRerouteInterval() {
...
+        if (watermark.isAbsolute()) {
+            return includeSettingKey ? watermarkSettingKey + "=" + watermark.getStringRep() : watermark.getStringRep();
+        } else if (watermark.calculateValue(total, maxHeadroom).equals(watermark.calculateValue(total, null))) {
+            String value = watermark.getStringRep();
+            return includeSettingKey ? watermarkSettingKey + "=" + value : value;
         } else {
-            return "max_headroom=" + maxHeadroom;
+            return includeSettingKey
+                ? maxHeadroomSettingKey + "=" + maxHeadroom.getStringRep()
+                : "max_headroom=" + maxHeadroom.getStringRep();
```
DaveCTurner, line 317: I don't think this case is covered by `DiskThresholdSettingsTests#doTestDescriptions` - could you add a few lines of test code about this branch there?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
```diff
@@ -275,88 +216,78 @@ private void setEnabled(boolean enabled) {
...
-    public ByteSizeValue getFreeBytesThresholdLow() {
-        return freeBytesThresholdLow;
+    public ByteSizeValue getFreeBytesThresholdFrozenFloodStage(ByteSizeValue total) {
+        return getFreeBytesThreshold(total, frozenFloodStageWatermark, frozenFloodStageMaxHeadroom);
     }
 
-    public ByteSizeValue getFreeBytesThresholdHigh() {
-        return freeBytesThresholdHigh;
-    }
+    private ByteSizeValue getMinimumTotalSizeForBelowWatermark(
```
DaveCTurner, line 263: This is only used for the low watermark (at least for now). Do you need it to be this general or can we inline it and then just add the headroom parameter in the next PR?

server/src/test/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDeciderTests.java
```diff
@@ -930,8 +915,8 @@ public void testCanRemainWithShardRelocatingAway() {
             decision.getExplanation(),
             containsString(
                 "the shard cannot remain on this node because it is above the high watermark cluster setting "
-                    + "[cluster.routing.allocation.disk.watermark.high=70%] and there is less than the required [30.0%] free disk on node, "
-                    + "actual free: [20.0%]"
+                    + "[cluster.routing.allocation.disk.watermark.high=70%] and there is less than the required [30b] free space "
+                    + "on node, actual free: [20b], actual used: [80%]"
```
DaveCTurner, line 919: I like that these messages now show the absolute byte values, but I think we shouldn't lose the percentages if the user is defining their watermarks in those terms. If it's too hard to make the message depend on whether the watermark is absolute or not then I would be ok with including both byte values and percentages in all such messages.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDecider.java
```diff
@@ -233,34 +226,34 @@ public Decision canAllocate(ShardRouting shardRouting, RoutingNode node, Routing
             && shardRouting.recoverySource().getType() == RecoverySource.Type.EMPTY_STORE;
 
         // checks for exact byte comparisons
```
DaveCTurner, line 228: nit: comment is obsolete now, this block does all the checks

```suggestion
```

DaveCTurner (commented):
Great, I like this change. I left some (mostly small) comments.

docs/reference/modules/cluster/disk_allocator.asciidoc
```diff
@@ -123,7 +123,7 @@ Controls the flood stage watermark for dedicated frozen nodes, which defaults to
 `cluster.routing.allocation.disk.watermark.flood_stage.frozen.max_headroom` {ess-icon}::
 (<<dynamic-cluster-setting,Dynamic>>)
 Controls the max headroom for the flood stage watermark for dedicated frozen
-nodes. Defaults to 20GB when
+nodes. Defaults to 20gb when
```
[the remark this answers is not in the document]
kingherc: Hi @DaveCTurner , I actually did this because I noticed that ByteSizeValue.toString() outputs it as `gb` rather than `GB` or `GiB`. So currently when the max headroom is output, it's `20gb` rather than the default `20GB`.

But I do agree with you. Maybe I leave this as it was, and I create a new minor enhancement issue to improve how ByteSizeValue/ByteSizeUnit output strings? So that they ideally output `GiB` or `GB` instead of `gb` (similar for other units of course)?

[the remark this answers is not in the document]
DaveCTurner: I think that this would be too large a change unfortunately, we've used lower-case byte-size units since forever (yes `mb` means megabytes not millibits). We probably could fix this given sufficient effort but it'd be a big breaking change so best left alone.

TBC `20gb` is fine, it's just an unformatted "20gb" that isn't.

server/src/main/java/org/elasticsearch/common/unit/RelativeByteSizeValue.java
kingherc: I see I can convert it to an instance method. I will do that. I am cc'ing @gmarouli (who implemented this) in case there is an objection.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderServiceTests.java
kingherc: Hi @DaveCTurner , this test has been moved to org.elasticsearch.cluster.routing.allocation.DiskThresholdSettingsTests#testMinimumTotalSizeForBelowLowWatermark , that is why you see it removed here. Please unresolve this conversation if you need more on this.

x-pack/plugin/searchable-snapshots/src/test/java/org/elasticsearch/xpack/searchablesnapshots/cache/shared/FrozenCacheServiceTests.java
kingherc: Resolved with the new instance method I made. You will see soon the new commits after I incorporate your feedback.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
kingherc: It's not used anymore and if not used, the checks will complain I believe. Should I keep it? And is there some qualifier I can add so checks do not complain?

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/storage/ReactiveStorageDeciderServiceTests.java
DaveCTurner: Ah sorry I missed that. Thanks for clarifying.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
DaveCTurner: Sure _this_ method isn't used any more because we're parsing these values differently, but the new parser doesn't have this optimization and is therefore going to slow tests down. I think we should move this optimization to the new parsing code.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
kingherc: Hi, yes it's covered in there if you see:

```
        String frozenFloodMaxHeadroomPrefix = includeKey
            ? DiskThresholdSettings.CLUSTER_ROUTING_ALLOCATION_DISK_FLOOD_STAGE_FROZEN_MAX_HEADROOM_SETTING.getKey() + "="
            : "max_headroom=";
```

It only applies to the frozen tier max headroom, and there is a relevant test:

```
        assertThat(
            diskThresholdSettings.describeFrozenFloodStageThreshold(thousandTb, includeKey),
            equalTo(frozenFloodMaxHeadroomPrefix + "10gb")
        );
```

Please unresolve this conversation if it doesn't suffice.

server/src/test/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDeciderTests.java
kingherc: With the headroom settings, it will be sometimes difficult to go back in figuring out which setting prevailed (the percentage or the headroom). So, I edited the message to include all relevant information that can be used to understand what happened. 

### Case where percentage prevails
This is the case of this example. You see that `actual used: [80%]` is over the `[cluster.routing.allocation.disk.watermark.high=70%]` watermark. So this should be fine. I do not think we also need the free percentage, since it is not used for the watermark calculation (the used percentage is over the watermark).

### Case where max headroom prevails
It will read something like:
```
[cluster.routing.allocation.disk.watermark.frozen.max_headroom=20gb] and there is less than the required [20gb] free space on node, actual free: [20b], actual used: [99%]"
```
This should also be fine. It shows that the max_headroom is violated since there is less than 20gb free space. We do not need a free percentage here because it is not taken into account (the bytes and headroom is taken into account).

Please tell me if this is not sufficient and unresolve the conversation.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
DaveCTurner: Sorry I see now, don't mind me. 😕

server/src/main/java/org/elasticsearch/common/unit/RelativeByteSizeValue.java
gmarouli: When we first created this it was processing doubles only, later as we adopted the `RelativeByteSizeValue` that wasn't necessary but we chose to preserve as static because we thought it would be useful for others too. I do not have strong objections.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
kingherc: @DaveCTurner had to keep this because it produced this:

```
> Task :server:forbiddenApisMain
Forbidden method invocation: java.lang.String#format(java.lang.String,java.lang.Object[]) [Uses default locale]
  in org.elasticsearch.cluster.routing.allocation.DiskThresholdSettings$WatermarkValidator (DiskThresholdSettings.java:185)
Scanned 7046 class file(s) for forbidden API invocations (in 2.12s), 1 error(s).

> Task :server:forbiddenApisMain FAILED
```

Please unresolve conversation if this is not sufficient.

server/src/main/java/org/elasticsearch/common/unit/RelativeByteSizeValue.java
kingherc: Oh thanks for the explanation! For now, I made it an instance method. But it can be converted back to a static method when somebody needs it. Please feel free to tell me if this is not sufficient and unresolve this conversation.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
kingherc: Ah thanks for clarifying, I now understood from reading your comment and the comment above this function. Indeed let me see if I can move it somewhere, e.g., in RelativeByteSizeValue. I will come back to you.

server/src/main/java/org/elasticsearch/common/unit/RelativeByteSizeValue.java
gmarouli: Great! No, it's sufficient, I just saw the comment and I wanted to give some context :)

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
kingherc: Hi @DaveCTurner , so after seeing the code in master, indeed this was used to speed up validations. But now with the RelativeByteValue class, we do not need to handle Strings in the validation, and the class already knows whether it's a percentage or absolute value. So there is no issue with validations.

However, I do see that this can speed up the parsing method of RelativeByteValue, so I moved this check and comment to org.elasticsearch.common.unit.RelativeByteSizeValue#parseRelativeByteSizeValue . Please check new commit and if it is not sufficient, comment and unresolve this conversation.
Thanks!

kingherc:
Hi @DaveCTurner , I think I may have handled all your feedback so far. Feel free to check once more. Thanks!

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
DaveCTurner: Yes you mustn't use `java.lang.String#format` but `org.elasticsearch.core.Strings#format` is preferable (it hard-codes the `Locale` and is exception-safe too).

kingherc: Thanks! done. Please unresolve conversation is it is not sufficient.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
DaveCTurner: sorry there's a few more of these still

server/src/test/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDeciderTests.java
DaveCTurner: Ok I see now, yes this makes sense to me.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/DiskThresholdSettings.java
kingherc: Thanks, I just did them.

DaveCTurner (approved):
Great, LGTM then (assuming CI is happy too)

PR #88723, opened 2022-07-22
Clean up FsRepositoryTests

pxsalehi:
I might be mistaken, but the CountDownLatch doesn't seem to be necessary.

### DIFF
--- a/server/src/test/java/org/elasticsearch/repositories/fs/FsRepositoryTests.java
+++ b/server/src/test/java/org/elasticsearch/repositories/fs/FsRepositoryTests.java
@@ -65,7 +65,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
-import java.util.concurrent.CountDownLatch;
 
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.emptySet;
@@ -149,8 +148,8 @@ public class FsRepositoryTests extends ESTestCase {
             IndexCommit incIndexCommit = Lucene.getIndexCommit(Lucene.readSegmentInfos(store.directory()), store.directory());
             Collection<String> commitFileNames = incIndexCommit.getFileNames();
             final PlainActionFuture<ShardSnapshotResult> future2 = PlainActionFuture.newFuture();
+            IndexShardSnapshotStatus snapshotStatus = IndexShardSnapshotStatus.newInitializing(shardGeneration);
             runGeneric(threadPool, () -> {
-                IndexShardSnapshotStatus snapshotStatus = IndexShardSnapshotStatus.newInitializing(shardGeneration);
                 repository.snapshotShard(
                     new SnapshotShardContext(
                         store,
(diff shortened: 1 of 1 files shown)


server/src/test/java/org/elasticsearch/repositories/fs/FsRepositoryTests.java
```diff
@@ -201,16 +200,8 @@ public void testSnapshotAndRestore() throws IOException, InterruptedException {
...
-        CountDownLatch latch = new CountDownLatch(1);
-        threadPool.generic().submit(() -> {
-            try {
-                runnable.run();
-            } finally {
-                latch.countDown();
-            }
-        });
-        latch.await();
+    private void runGeneric(ThreadPool threadPool, Runnable runnable) {
```
DaveCTurner, line 203: may as well just inline this now it's a one-liner?

DaveCTurner (changes requested):
Yep I think you're right. These APIs weren't always completely async, but now that it is we can do this. But I think we can go further: there's no need to run this stuff on a background thread  at all now, and we can even assert that each future is complete when the corresponding method returns.

server/src/test/java/org/elasticsearch/repositories/fs/FsRepositoryTests.java
DaveCTurner: ah sorry meant to delete this comment - there's no need to fork at all any more I think

pxsalehi: :facepalm: right! I've also done some renames! Please see https://github.com/elastic/elasticsearch/pull/88723/commits/d942ffcf9e51dd0a2e8ca068cd1c5abd81286483.

DaveCTurner:
> That means we're never actually running any data file uploads or restores that would fork wouldn't it?

Not sure about all such tests but at least here we're using the mock threadpool that `BlobStoreTestUtil#mockClusterService` creates which runs everything using a `SameThreadExecutorService`.

original-brownbear:
> Not sure about all such tests but at least here we're using the mock threadpool that BlobStoreTestUtil#mockClusterService creates which runs everything using a SameThreadExecutorService.

Ah right the test threadpool is now unused ... my bad ++ to asserting that things are completed right away then.

DaveCTurner (commented):
> Ah right the test threadpool is now unused

Good point yes `ThreadPool threadPool = new TestThreadPool(getClass().getSimpleName());` is no longer needed here.

DaveCTurner:
image

I'm also a big fan of putting the date in my branch names :) You might find it useful to use a fully numeric YYYY-MM-DD rather than the month name - it sorts better.

pxsalehi:
Yay! Even more code removed! :-)
Please see https://github.com/elastic/elasticsearch/pull/88723/commits/c112c3d3b09a3fc66a381c91fcd4bc0689c85905.

DaveCTurner (approved):
LGTM

pxsalehi:
Thanks David and Armin!

PR #88725, opened 2022-07-22
Increase `http.max_header_size` default to 16kb

### LINKED ISSUE #88501: Increase `http.max_header_size` default to 16KB
### Description

Our current default for the `http.max_header_size` setting is 8KB. This is lower than the current default for Kibana (16KB in 8.x, based on [the default for node 14 and up](https://www.educative.io/answers/what-is-the-node-httpmaxheadersize-property)), and our the ESS proxy (1MB based on the Go `http` library default). To align with the current convention of other Elastic components, we should increase the ES header size setting default to 16KB. Note that this limit is for all request headers combined, rather than per header.

This issue came out of an investigation around [JWT auth headers](https://www.elastic.co/guide/en/elasticsearch/reference/master/jwt-auth-realm.html) which can get quite sizeable.

I'm happy to open a PR since it's a small change.

n1v0lg:
Our current default for the `http.max_header_size` setting is 8kb. This
is lower than the current default for Kibana (16kb in 8.x), and the ESS
proxy (1mb based on the Go http library default). To align with the
current convention of other Elastic components, this PR increases the
ES header size setting default to 16kb.

Closes #88501


DaveCTurner (changes requested):
Looks good but the docs at `docs/reference/modules/http.asciidoc` still say the default is `8kb`.

n1v0lg:
Whoops, good catch!

DaveCTurner (approved):
LGTM

n1v0lg:

PR #88731, opened 2022-07-22
Add support for reading stable plugin descriptors

rjernst:
The new stable plugin api will have a slightly different descriptor file
format. This commit prepares for stable plugins by adding support for
reading those new files. The basic info for a plugin is the same like
name and version info. Other stuff like classname are not necessary. The
one additional property specific to the new plugins is "modular", which
indicates whether the jars of the plugin should be loaded as named
modules (this is akin to setting the module path when running java).


grcevski (approved):
LGTM! Looks great, I really like how all the tests simply run for both. It looks like there are two more tests that need to be adjusted for the new missing descriptors error message.

ChrisHegarty (approved):
LGTM

PR #88733, opened 2022-07-22
Script: Protected _source inside update scripts

stu-elastic:
CtxMap delegates all metadata keys to it's `Metadata` container and
all other keys to it's source map.  In most write contexts (update,
update by query, reindex), the source map should only contain one
key, `_source`, which has a `Map<String, Object>`.

This change adds validations to writes to the source map that rejects
insertion of invalid keys, illegal removal of the `_source` key, and
illegally overwriting the `_source` mapping with the wrong type.


server/src/main/java/org/elasticsearch/script/CtxMap.java
```diff
@@ -52,26 +51,21 @@ protected CtxMap(Map<String, Object> source, Metadata metadata) {
         }
     }
 
-    protected Map<String, Object> wrapSource(Map<String, Object> source) {
-        Map<String, Object> wrapper = Maps.newHashMapWithExpectedSize(1);
-        wrapper.put(SOURCE, source);
```
rjernst, line 57: Why do we need more than replacing this wrapper with `return Map.of(SOURCE, source);`? Then the map is immutable, no keys can be written to the map that _source is in.

stu-elastic: it is valid to replace `_source` with another map representing the new source.

rjernst: I don't think using an immutable map precludes that requirement. If a script has `ctx._source = ['foo': 'bar']`, the put still happens on CtxMap. It can then completely replace the immutable source map in this special case, it doesn't need to delegate.

stu-elastic: Puts also happen in the EntrySet's Map.Entry which is available through entrySet() and entrySet().iterator().

rjernst: But that is still on the CtxMap right? Ultimately that is what controls everything under `ctx`, so *replacing* _source by doing a put, wherever that is coming from, should be controllable within CtxMap, without needing any special source map wrapper.

jdconrad: I'm also not understanding the need for an additional indirection w/ `RestrictedMap` when this is already something `CtxMap` is specifically specialized for. Seems like `CtxMap` could easily handle the addition of the specialized key even if this needs to be done on a case by case per context with something like `IngestCtxMap`.

stu-elastic: That's wasn't too bad.  Any ideas on how we can ensure `source.put` is never called in `CtxMap`?

server/src/main/java/org/elasticsearch/script/CtxMap.java
```diff
@@ -52,20 +50,15 @@ protected CtxMap(Map<String, Object> source, Metadata metadata) {
...
-        return wrapper;
+    protected boolean directSourceAccess() {
```
rjernst, line 53: Can you please add a short javadoc describing what this controls (ie mentioning whether source is indirectly accessed through the "_source" key or at the root of ctx).

server/src/main/java/org/elasticsearch/script/Metadata.java
```diff
@@ -44,7 +44,7 @@ public class Metadata {
     protected static final String VERSION_TYPE = "_version_type";
     protected static final String VERSION = "_version";
     protected static final String TYPE = "_type"; // type is deprecated, so it's supported in the map but not available as a getter
-    protected static final String TIMESTAMP = "_now";
```
rjernst, line 47: I think this rename should be pulled out. It has nothing to do with protecting modification to ctx. If a dev is looking through git blame and sees this line was changed via this commit, it will be confusing what the two have to do with each other. Such a rename should be a simple and quick PR on its own.

jdconrad (approved):
Once you've met @rjernst's request to pull `timestamp` -> `now` into a separate PR, the rest of the change LGTM.

server/src/main/java/org/elasticsearch/script/CtxMap.java
stu-elastic: Done.

server/src/main/java/org/elasticsearch/script/Metadata.java
stu-elastic: I think it's useful to have the context of removing the invalid `_timestamp` test, which is why this rename happens in this PR.

rjernst: As far as I can see, there is not rename of `_timestamp` here. There is a test removal, related to this change, and then there is this rename of variables. The latter has nothing to do with protecting modification to ctx.

stu-elastic: `TIMESTAMP` is not a good name because there is an old metadata value `_timestamp`, which was indicated in the test.

The only reason to change it was due to the weirdness of removing a `_timestamp` test while having a constant of value `_now` named `TIMESTAMP`.

rjernst: To be clear: I am not objecting to the rename. It seems like a fine change. But that rename could happen right now, in the main branch. It would be a trivial change, on its own, which could quickly be reviewed. But the commit message would then indicate what it is: a rename, for the purpose of clarification. The commit message in this PR does not indicate anything about the rename; it's unrelated to the change here. Someone doing git blame later and seeing this rename, perhaps trying to understand how this class has changed over time, would see this unrelated change and would likely be confused on how the two relate.

stu-elastic: Thinking about this more, we also expose `getTimestamp()` so this change made less sense without also changing `getTimestamp()`.

That interface change is worth a discussion, so I've reverted that rename and we can discuss it later.

rjernst (approved):
LGTM, thanks for the iterations

stu-elastic:
Thanks for the reviews and the help.

PR #88735, opened 2022-07-22
Add source fallback for keyword fields using operation

jdconrad:
This is a follow up PR to (#88040) to contrast a parameter passed to fielddataBuilder versus separate methods of fielddataBuilder and scriptFielddataBuilder. The parameter is an enum named MappedFieldType.FielddataOperation where the operation specifies what data structures should be used to generate fielddata. Right now it only has SEARCH and SCRIPT, but in the future this could extended to AGGS, SORT, etc. to allow fine-grained tuning where doc values are pulled from.

Note that something like DOC_VALUES_ONLY or SOURCE_FALLBACK doesn't work for scripting here because with exceptions like text field scripting can't know if it wants doc values primarily or source primarily so it has to be determined via the MappedType per operation.


server/src/main/java/org/elasticsearch/index/query/SearchExecutionContext.java
```diff
@@ -91,7 +92,7 @@ public class SearchExecutionContext extends QueryRewriteContext {
     private final MappingLookup mappingLookup;
     private final SimilarityService similarityService;
     private final BitsetFilterCache bitsetFilterCache;
-    private final TriFunction<MappedFieldType, String, Supplier<SearchLookup>, IndexFieldData<?>> indexFieldDataService;
+    private final QuadFunction<MappedFieldType, String, Supplier<SearchLookup>, FielddataOperation, IndexFieldData<?>> indexFieldDataLookup;
```
romseygeek, line 95: I think by the time we have a generic four-input function we should probably just make this a functional interface called IndexFieldDataLookup? Alternatively we could try and combine the index name, lookup supplier and operation into a context object.

romseygeek: Thinking about this more, I like the idea of a context object, as it would mean we could also put the `sourcePaths()` method there rather than extending SearchLookup further.

romseygeek (commented):
I much prefer this version, thank you for being so patient and going through all these incremental changes!

What do you think about creating a context object that wraps up SearchLookup, index name, sourcefields and operation type?  Looking at where IndexFieldDataService.getForField() is called I think we can the context construction into SearchExecutionContext happily, and then have an empty one for building sort fields and running warming queries as those situations only support direct docvalues reading.

jdconrad:
@romseygeek is doing a bit of refactoring that should help clean up some of the plumbing of this PR. Pausing for the outcome of (https://github.com/elastic/elasticsearch/pull/88779).

server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -681,13 +682,44 @@ NamedAnalyzer normalizer() {
...
+                    CoreValuesSourceType.KEYWORD,
+                    (dv, n) -> new KeywordDocValuesField(FieldData.toString(dv), n)
+                );
+            }
+
+            if (operation == FielddataOperation.SCRIPT) {
+                return new SourceValueFetcherSortedBinaryIndexFieldData.Builder(
+                    name(),
+                    CoreValuesSourceType.KEYWORD,
+                    new SourceValueFetcher(searchLookup.get().sourcePaths(name()), nullValue) {
```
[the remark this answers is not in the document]
jdconrad: Sure. Do you have any suggestions for what you'd like to see as part of the API? My natural inclination would be to add a second valueFetcher or a new sourceValueFetcher method that takes sourcepaths and nullValue, but was wondering if you had something else in mind.

server/src/main/java/org/elasticsearch/index/query/SearchExecutionContext.java
jdconrad: I have integrated the changes from #88779 and it appears to have reduced the line count by around 400 or 25%!

server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java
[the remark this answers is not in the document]
jdconrad: I added private `sourceValueFetcher` methods for Keyword and Integer mapped types. I'm not completely sure about this is the way to go as it kind of evades the expected constructor and has to pull the source paths itself even when the search execution context is available. Any thoughts here?

jdconrad:
@romseygeek Would you please take another pass when you have the chance? Thank you!

server/src/main/java/org/elasticsearch/index/IndexService.java
```diff
@@ -206,7 +206,10 @@ public IndexService(
                 this.indexSortSupplier = () -> indexSettings.getIndexSortConfig()
                     .buildIndexSort(
                         mapperService::fieldType,
-                        (fieldType, searchLookup) -> indexFieldData.getForField(fieldType, FieldDataContext.noRuntimeFields("index sort"))
+                        (fieldType, searchLookup, fielddataOperation) -> indexFieldData.getForField(
```
romseygeek, line 209: I don't think we need to pass this extra fielddataOperation parameter here?

server/src/main/java/org/elasticsearch/index/fielddata/FieldDataContext.java
```diff
@@ -30,7 +40,9 @@ public record FieldDataContext(String fullyQualifiedIndexName, Supplier<SearchLo
     public static FieldDataContext noRuntimeFields(String reason) {
         return new FieldDataContext(
             "",
-            () -> { throw new UnsupportedOperationException("Runtime fields not supported for [" + reason + "]"); }
+            () -> { throw new UnsupportedOperationException("Runtime fields not supported for [" + reason + "]"); },
+            null,
```
romseygeek, line 44: It might be safer to have a default sourcePathsLookup of `s -> Set.of(s)` here?

server/src/test/java/org/elasticsearch/index/mapper/AbstractScriptFieldTypeTestCase.java
```diff
@@ -186,8 +186,8 @@ protected static SearchExecutionContext mockContext() {
         return mockContext(true);
     }
 
-    protected static FieldDataContext mockFielddataContext() {
-        return new FieldDataContext("test", mockContext()::lookup);
+    protected static FieldDataContext mockFielddataContext(MappedFieldType.FielddataOperation fielddataOperation) {
+        return new FieldDataContext("test", mockContext()::lookup, mockContext()::sourcePath, fielddataOperation);
     }
```
romseygeek, line 191: Given that these are script tests, I think we can just always use FielddataOperation.SCRIPT here?

romseygeek (commented):
This looks great! I left a few comments but I think we're nearly there.

server/src/main/java/org/elasticsearch/index/IndexService.java
jdconrad: Fixed. Also, for my understanding, is this because we never want sort to fallback so there's no reason to build a different operation type into an index sort?

romseygeek: Yes, exactly

server/src/main/java/org/elasticsearch/index/fielddata/FieldDataContext.java
jdconrad: Yes! Fixed.

server/src/test/java/org/elasticsearch/index/mapper/AbstractScriptFieldTypeTestCase.java
jdconrad: This is an interesting case for runtime fields. I don't think it's actually possible to hit source fallback for a runtime field at all, but they will likely be executed at a top level as a search operation, so I think for these tests it would make the most sense to switch it to SEARCH rather than SCRIPT.

romseygeek: > I don't think it's actually possible to hit source fallback for a runtime field at all

I'm confused, runtime fields explicitly use scripts so I think they're quite likely to use this path? Or am I misunderstanding what's going on here?

jdconrad:
@romseygeek Thank you for the continued review iterations. This is ready for another pass.

server/src/test/java/org/elasticsearch/index/mapper/AbstractScriptFieldTypeTestCase.java
jdconrad: My understanding is when a runtime field is first executed as a script its part of a search query where the runtime field looks like a standard field. At that first execution the runtime field should always be found as the doc values portion of a field that it's emulating, right? So I don't think the operation actually matters for that "top-level" runtime field. However, for fields referenced from that runtime field then the operation will be SCRIPT when we do the lookup from LeafDocLookup.

jdconrad: For the tests you're right, this should be SCRIPT. I'm thinking about this incorrectly.

jdconrad: Fixed.

romseygeek (approved):
LGTM!

jdconrad:
@romseygeek Thank you! Will commit as soon as CI passes.

server/src/main/java/org/elasticsearch/index/mapper/MappedFieldType.java
```diff
@@ -80,6 +80,15 @@ public MappedFieldType(
         this.meta = Objects.requireNonNull(meta);
     }
 
+    /**
+     * Operation to specify what data structures are used to retrieve
+     * field data from and generate a representation of doc values.
+     */
+    public enum FielddataOperation {
+        SEARCH,
+        SCRIPT
```
javanna, line 89: Should we add docs to clarify how these two options differ from each other?

server/src/main/java/org/elasticsearch/index/mapper/GeoPointFieldMapper.java
```diff
@@ -369,8 +371,30 @@ private boolean isPointGeometry(LatLonGeometry[] geometries) {
...
         public IndexFieldData.Builder fielddataBuilder(FieldDataContext fieldDataContext) {
-            failIfNoDocValues();
-            return new AbstractLatLonPointIndexFieldData.Builder(name(), CoreValuesSourceType.GEOPOINT, GeoPointDocValuesField::new);
+            FielddataOperation operation = fieldDataContext.fielddataOperation();
+
+            if (operation == FielddataOperation.SEARCH) {
+                failIfNoDocValues();
+            }
+
+            if ((operation == FielddataOperation.SEARCH || operation == FielddataOperation.SCRIPT) && hasDocValues()) {
```
javanna, line 380: is the first part of the if necessary? It's maybe a protection mechanism for potential new operations that may be added in the future? Isn't it more practical to just say if hasDocValues() load from docvalues?

server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -682,12 +684,34 @@ NamedAnalyzer normalizer() {
...
-                CoreValuesSourceType.KEYWORD,
-                (dv, n) -> new KeywordDocValuesField(FieldData.toString(dv), n)
-            );
+            FielddataOperation operation = fieldDataContext.fielddataOperation();
+
+            if (operation == FielddataOperation.SEARCH) {
+                failIfNoDocValues();
+            }
+
+            if ((operation == FielddataOperation.SEARCH || operation == FielddataOperation.SCRIPT) && hasDocValues()) {
```
javanna, line 693: same comment as above

server/src/main/java/org/elasticsearch/index/mapper/NumberFieldMapper.java
```diff
@@ -1403,8 +1425,24 @@ public Function<byte[], Number> pointReaderIfPossible() {
...
         public IndexFieldData.Builder fielddataBuilder(FieldDataContext fieldDataContext) {
-            failIfNoDocValues();
-            return type.getFieldDataBuilder(name());
+            FielddataOperation operation = fieldDataContext.fielddataOperation();
+
+            if (fieldDataContext.fielddataOperation() == FielddataOperation.SEARCH) {
+                failIfNoDocValues();
+            }
+
+            if ((operation == FielddataOperation.SEARCH || operation == FielddataOperation.SCRIPT) && hasDocValues()) {
```
javanna, line 1434: here too, same as above

server/src/main/java/org/elasticsearch/search/aggregations/support/AggregationContext.java
```diff
@@ -435,7 +435,7 @@ public Analyzer buildCustomAnalyzer(
 
         @Override
         protected IndexFieldData<?> buildFieldData(MappedFieldType ft) {
-            return context.getForField(ft);
+            return context.getForField(ft, MappedFieldType.FielddataOperation.SEARCH);
```
javanna, line 438: given all these related changes, too late now but have we considered having a separate getForField only for scripting and leaving the existing one alone, so that less changes are required in existing code?

javanna (commented):
I had a quick look as well, this change seemed important enough for me to look at even if it's already merged. I left some questions. An additional question I have is around the scope: the title mentions keyword fields, but I was under the impression that the fallback has been added also for numbers and geo_points?

PR #88736, opened 2022-07-22
Change HealthMetadata to ClusterState.Custom

gmarouli:
In #88175 we introduced the health metadata in the cluster state, so the master can share with the other nodes of the cluster the information needed to evaluate their health. We did this by adding the health metadata as a `Metadata.Custom`. `Metadata.Custom` persists across a full cluster restart which is something we do not need since the master will always reset the health metadata upon election. `ClusterState.Custom` doesn't have this property so it fits our use case better.

This PR changes the type of the `HealthMetadata` from `Metadata.Custom` to `ClusterState.Custom`


andreidan (approved):
LGTM, thanks for fixing this Mary

gmarouli:
@andreidan thanks for the quick review and @DaveCTurner for catching this.

gmarouli:
@elasticmachine update branch

PR #88739, opened 2022-07-22
Add Checkstyle rule for broken switch cases

pugnascotia:
We use `between(x, y)` calls with `switch` statements in tests to
randomize test behaviour. However, some usages define `case` statements
that can never execute, because the `case` value is outside the range
defined by the `between` call.

Write a rule that inspects the switches and the cases, and fails on the
broken cases.

The following issues cover the changes required to fix this problem.

   * [x] #88743
   * [x] #88744
   * [x] #88745
   * [x] #88746


mark-vieira:
@breskeby how hard would it be to add some benchmarks for this kind of thing or does the fact that our existing benchmark suite does `precommit` probably already account for this. The point being that complex checkstyle rules can be pretty expensive, so I want to ensure that we aren't adding to much overhead by adding rules for very specific edge case items like this.

pugnascotia:
@mark-vieira this implementation does at least limit itself to switch statements. It's not like the long-lines rules rule. But, I don't have any perf number for you, so 🤷‍♂️

mark-vieira:
> @mark-vieira this implementation does at least limit itself to switch statements. It's not like the long-lines rules rule. But, I don't have any perf number for you, so 🤷‍♂️

Yeah, I don't suspect it'll be a problem, but we have infrastructure for generating performance metrics on the effect of any given build change so I think it's worth leveraging that for stuff like this (or any additional static analysis checks) to give a number to the cost of such a thing.

DaveCTurner:
We also use `switch (randomIntBetween(...))` and `switch(randomInt(...))` which could suffer the same problems. Should we check these forms too? Or maybe we should drop these aliases and just use `between` everywhere?

breskeby:
@mark-vieira we do indirectly check the performance with the precommit checks we have. But given all the stuff we run there, I can imagine it would need to be really really significant for us to catch here. We could setup checkstyle specific performance checks to detect performance losses here early. I'm just not sure its worth it

pugnascotia:
@DaveCTurner I also handled "randomIntBetween" but I've added support for "randomInt" as well now.

mark-vieira:
> We could setup checkstyle specific performance checks to detect performance losses here early.

I think it might be worth adding a scenario for checkstyle specifically at some point. We are increasingly adding more complex checkstyle rules and we don't really have any idea of how costly they are. Checkstyle as a whole we know is pretty expensive but it's not clear if that is just the AST generation or the rules, or what.

pugnascotia:
@elasticmachine update branch

pugnascotia:
@joegallo @mark-vieira the dependent issues are now closed, can I get a review please?

joegallo (approved):
+100 on the net effect here, and the code looks reasonable to me. ❤️ for taking a look at this, @pugnascotia.

mark-vieira (approved):
LGTM

pugnascotia:
Backported to `7.17` in 50dc8ae8b9d (and fixed another error in 4c9f8b054bc).

PR #88740, opened 2022-07-22
Script: UpdateByQuery can read doc version if requested

### LINKED ISSUE #55745: Scripting: ctx[_version] in update by query returns -1 since 6.7
In [`TransportUpdateByQueryAction`](https://github.com/elastic/elasticsearch/blob/master/modules/reindex/src/main/java/org/elasticsearch/index/reindex/TransportUpdateByQueryAction.java#L86..L89) we always create `AsyncIndexBySearchAction` with `needsSourceDocumentVersions=false`, so the versions are never fetched.  This was on the switch to sequence numbers: https://github.com/elastic/elasticsearch/pull/37857.

The [documentation](https://www.elastic.co/guide/en/elasticsearch/painless/master/painless-update-by-query-context.html) still refers to `ctx['_version']` and we still populate that key in the context.

stu-elastic:
Allow UpdateByQuery to read the doc version if set in the request via
`version=true`.

If `version=true` is unset or false, the `ctx._version` is `-1`
indicating internal versioning via seq.

Fixes: #55745

### DIFF
--- a/docs/changelog/88740.yaml
+++ b/docs/changelog/88740.yaml
@@ -1,0 +1,5 @@
+pr: 88740
+summary: "Script: `UpdateByQuery` can read doc version if requested"
+area: "Distributed, Infra/Scripting"
+type: bug
+issues: []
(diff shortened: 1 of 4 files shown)


rjernst (approved):
LGTM

arteam (approved):
LGTM! Thank you

PR #88747, opened 2022-07-22
Fix switch in KeywordFieldMapperTests

### LINKED ISSUE #88746: `KeywordFieldMapperTests` has switch with impossible branches
The test class `KeywordFieldMapperTests` contains a switch statement with a branch that will never execute. The test uses a `between(x, y)` switch expression, but the `4` case is outside the randomised range. We need to fix this before #88739 can be merged. By "fix", I mean either remove the spurious case, or if we actually want it then extend the randomized range.

dnhatn:
Closes #88746


mayya-sharipova (approved):
@dnhatn Thanks Nhat.

dnhatn:
Thanks Mayya!

PR #88748, opened 2022-07-22
Fix erroneous switch/cases in ILM tests

### LINKED ISSUE #88743: ILM tests have switches with impossible branches
The following ILM tests all contain switch statements with branches that will never execute. The tests use a `between(x, y)` switch expression, but there are `case` values that are outside the randomised range. We need to fix these before #88739 can be merged. By "fix", I mean either remove the spurious cases, or if we actually want them, then extend the randomized range.

   * ActionConfigStatsTests
   * AllocateActionTests
   * CopyExecutionStateStepTests
   * RolloverActionTests
   * WaitForRolloverReadyStepTests

joegallo:
Closes #88743

/cc @pugnascotia


masseyke (approved):
LGTM

joegallo:

PR #88749, opened 2022-07-23
[7.17] Fix erroneous switch/cases in ILM tests

joegallo:
Related to https://github.com/elastic/elasticsearch/issues/88743, backport of #88748

This is a backport of the intent of #88748. The code is actually mostly different -- some of the issues that I fixed on #88748 just don't exist on 7.17, and some issues existed only on 7.17 and didn't need to be fixed on #88748.

/cc @pugnascotia


masseyke (approved):
LGTM

PR #88753, opened 2022-07-25
Fix racing when loading new JWKs from multiple threads

ywangd:
This PR ensures the mutation of JWKs is done in a single thread and
visible to all other threads, which in turn ensures validation to be
correctly performed concurrently.

Relates: #88023


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/jwt/JwtRealm.java
```diff
@@ -663,11 +662,31 @@ public void usageStats(final ActionListener<Map<String, Object>> listener) {
...
+         * they are different. The listener is called with false if the reloaded content is the same
+         * as the existing one or true if they are different.
+         */
+        void reload(final ActionListener<Boolean> listener) {
+            load(ActionListener.wrap(newContentAndJwksAlgs -> {
+                if (Arrays.equals(JwtRealm.this.contentAndJwksAlgsPkc.sha256, newContentAndJwksAlgs.sha256)) {
+                    // No change in JWKSet
+                    listener.onResponse(false);
+                } else {
+                    JwtRealm.this.contentAndJwksAlgsPkc = newContentAndJwksAlgs;
```
tvernum, line 684: As discussed, I really dislike nested classes breaking encapsulating like this to mutate the internal state of the enclosing class. I think we need to clarify the state responsibilities between the 2 classes.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/jwt/JwtRealm.java
```diff
@@ -677,7 +718,10 @@ private ListenableFuture<ContentAndJwksAlgs> getFuture() {
 
                 final ListenableFuture<ContentAndJwksAlgs> newFuture = new ListenableFuture<>();
                 if (this.reloadFutureRef.compareAndSet(null, newFuture)) {
-                    loadInternal(ActionListener.runAfter(newFuture, () -> this.reloadFutureRef.compareAndSet(newFuture, null)));
+                    loadInternal(ActionListener.runAfter(newFuture, () -> {
+                        final ListenableFuture<ContentAndJwksAlgs> oldValue = this.reloadFutureRef.getAndSet(null);
+                        assert oldValue == newFuture : "future reference changed unexpectedly";
```
ywangd, line 723: This change is unlikely to make real difference. But I think it is safer to always set to null regardless just in case compare ever not match and result into the reference holding a stale future.

tvernum (approved):
LGTM, thanks for iterating on it.

ywangd:
@elasticmachine update branch

ywangd:

ywangd:
@elasticmachine update branch

PR #88757, opened 2022-07-25
Fix incorrect switch cases in AggregationTests

### LINKED ISSUE #88745: Aggregation tests have switches with impossible branches
The following aggregation tests all contain switch statements with branches that will never execute. The tests use a `between(x, y)` switch expression, but there are `case` values that are outside the randomised range. We need to fix these before #88739 can be merged. By "fix", I mean either remove the spurious cases, or if we actually want them, then extend the randomized range.

   * InternalAutoDateHistogramTests
   * InternalAvgTests
   * InternalDerivativeTests
   * InternalGeoCentroidTests
   * InternalWeightedAvgTests
   * SignificantLongTermsTests
   * SignificantStringTermsTests

csoulios:
Fixes aggregation tests that contain switch statements with branches that will never execute. The tests use a between(x, y) switch expression, but there are case values that are outside the randomised range. 

Fixes #88745


iverase (approved):
LGTM

PR #88760, opened 2022-07-25
Add troubleshooting docs about data corruption

DaveCTurner:
Adds some docs giving more detailed background about what data
corruption really means and some suggestions about how to narrow down
the root cause.


docs/reference/troubleshooting/corruption-issues.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+indicator that your storage subsystem is not working correctly.
+
+To narrow down the source of the corruptions, systematically change components
+in your cluster until the corruptions stop. For instance, to rule out {es} as
+the source of data corruption, generate an I/O workload using something other
+than {es} and look for data integrity errors. On Linux the `fio` and
+`stress-ng` tools can both generate challenging I/O workloads and verify the
+integrity of the data they write. You can check that durable writes persist
+across power outages using a script such as
+[`diskchecker.pl`]https://gist.github.com/bradfitz/3172656. Try different
```
DaveCTurner, line 81: Unclear if we should mention these tools like this. Maybe we should give example invocations for `fio` and `stress-ng`? `diskchecker.pl` looks janky but it's pretty much the best way to find `fsync()` bugs (it's in the [PostgreSQL docs](https://www.postgresql.org/docs/current/wal-reliability.html)).

docs/reference/troubleshooting/corruption-issues.asciidoc
```diff
@@ -0,0 +1,82 @@
...
+proceses are causing corruption: they are examples of the rare times where
+reading a whole file is necessary. {es} takes the opportunity to verify the
+checksum at the same time, and this is when the corruption is detected and
+reported. It doesn't indicate the cause of the corruption or when it happened.
+Corruptions can remain undetected for many months.
+
+The files that make up a Lucene index are written sequentially from start to
+end and then never modified or overwritten. This access pattern means the
+checksum computation is very simple and can happen on-the-fly as the file is
+written, and also makes it very unlikely that an incorrect checksum is due to a
```
henningandersen, line 34: I think I agree to this. But we do not calculate it on every write, like when transferring the file over network or from snapshot. I'd also think the risk of a user space bug in those transfers is small and the argument as a whole stands.

The only change I would request here is to add "initially", i.e.:

```suggestion
checksum computation is very simple and can happen on-the-fly as the file is initially
written, and also makes it very unlikely that an incorrect checksum is due to a
```

docs/reference/troubleshooting/corruption-issues.asciidoc
```diff
@@ -0,0 +1,82 @@
...
+- faulty hardware (for instance the drive, RAID controller, RAM or CPU)
+
+Data corruption typically doesn't result in other evidence of problems apart
+from the checksum mismatch. Do not interpret this as an indication that your
+storage subsystem is working correctly and therefore that {es} itself caused
+the corruption. It is rare for faulty storage to show any evidence of problems
+apart from the data corruption, but data corruption itself is a very strong
+indicator that your storage subsystem is not working correctly.
+
+To narrow down the source of the corruptions, systematically change components
```
henningandersen, line 73: I think I'd like to reword this a bit to be more of a non-exhaustive list of suggestion of ways to help diagnosing corruptions.

henningandersen (commented):
This looks good, left a couple of mostly optional comments.

docs/reference/troubleshooting/corruption-issues.asciidoc
DaveCTurner: We do verify checksums when copying files over the network during a recovery. You're right that we don't when restoring from a snapshot, but I think we should (I opened https://github.com/elastic/elasticsearch/issues/88878). I'll commit this change.

docs/reference/troubleshooting/corruption-issues.asciidoc
DaveCTurner: :+1: see 37a0caa4d.

docs/reference/troubleshooting/corruption-issues.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+Typically these exceptions happen due to a checksum mismatch. Most of the data
+that {es} writes to disk is followed by a checksum using a simple algorithm
+known as CRC32 which is fast to compute and good at detecting the kinds of
+random corruption that may happen when using faulty storage. A CRC32 checksum
+mismatch definitely indicates that something is faulty, although of course a
+matching checksum doesn't prove the absence of corruption.
+
+Verifying a checksum is expensive since it involves reading every byte of the
+file which takes significant effort and might evict more useful data from the
+filesystem cache, so systems typically doesn't verify the checksum on a file
```
arteam, line 21: s/doesn't/don't ?

docs/reference/troubleshooting/corruption-issues.asciidoc
```diff
@@ -0,0 +1,92 @@
...
+checksum computation is very simple and can happen on-the-fly as the file is
+initially written, and also makes it very unlikely that an incorrect checksum
+is due to a userspace bug at the time the file was written. The routine that
+computes the checksum is straightforward, widely used, and very well-tested, so
+you can be very confident that a checksum mismatch really does indicate that
+the data read from disk is different from the data that {es} previously wrote.
+
+The files that make up a Lucene index are written in full before they are used.
+If a file is needed to recover an index after a restart then your storage
+system will previously have confirmed to {es} that this file was durably synced
```
arteam, line 42: Did you mean `had previously confirmed` here? `will previously have confirmed` sounds a bit strange to me.

arteam (approved):
This looks awesome! Thank you David!

docs/reference/troubleshooting/corruption-issues.asciidoc
DaveCTurner: This one is legitimate English - if you want the technical details, `will have confirmed` is the future perfect tense, and that's acceptable to use in the first conditional form as I'm doing here. But I think it's fair to say that it could be confusing too, so I simplified it in 85a8ee2.

henningandersen (approved):
LGTM.

PR #88761, opened 2022-07-25
[ML] Rebalance model assignments when outdated assignments exist

dimitris-athanasiou:
When a cluster has been updated from a version prior to the version
when distributed model allocation was introduced (8.4), there is
a possibility that after the full cluster has been upgraded there
still are assignments with routing entries that have `current_allocations`
and `target_allocations` equal to `0`. Those are outdated and we
should trigger a rebalance in order to apply the new distributed
model allocation.

This commit detects that and ensures a rebalance is done in such
scenario.

In addition, this fixes a bug where previous routing entries on
nodes where the planner does not assign any allocations for a model
would stick around.


droberts195 (approved):
LGTM

PR #88762, opened 2022-07-25
POC authz perf - only one pass over the global resources

albertzaharovits:
I hit a road-block here.
A general index expression can first include indices via wildcards pointing to aliases, but then exclude them via an exclusion wildcard for different aliases (which point to some of the same indices). In this scenario, we must first remember all the index names that have been included by a wildcard, before removing the ones that are covered by exclusion wildcards. It is not possible to evaluate the whole expression progressively, over every resource name in the cluster, and only include names, because an exclusion can at any moment remove included names via wildcards pointing to indices.


albertzaharovits:
I'm going to close this as I don't have the time to continue working on it.
Anyone please feel free to reuse anything you can scavange out of it.

PR #88772, opened 2022-07-25
Expanding switch statement to cover all cases in ConvertProcessorTests

### LINKED ISSUE #88744: `ConvertProcessorTests` tests have switches with impossible branches
The test `ConvertProcessorTests` contain switch statements with branches that will never execute. The tests use a `between(x, y)` switch expression, but there are `case` values that are outside the randomised range. We need to fix these before #88739 can be merged. By "fix", I mean either remove the spurious cases, or if we actually want them, then extend the randomized range.

masseyke:
ConvertProcessorTests.testConvertStringList had 5 test cases but would only randomly run 3 of them.
Closes #88744


joegallo (approved):
LGTM

masseyke:

PR #88773, opened 2022-07-25
Fix incorrect switch cases in `ConvertProcessorTests`

### LINKED ISSUE #88744: `ConvertProcessorTests` tests have switches with impossible branches

csoulios:
The test `ConvertProcessorTests` contain switch statements with branches that will never execute. The tests use a between(x, y) switch expression, but there are case values that are outside the randomised range.

This PR fixes the test by extending the randomized range.

Fixes https://github.com/elastic/elasticsearch/issues/88744

### DIFF
--- a/modules/ingest-common/src/test/java/org/elasticsearch/ingest/common/ConvertProcessorTests.java
+++ b/modules/ingest-common/src/test/java/org/elasticsearch/ingest/common/ConvertProcessorTests.java
@@ -416,7 +416,7 @@ public class ConvertProcessorTests extends ESTestCase {
         for (int j = 0; j < numItems; j++) {
             Object randomValue;
             String randomValueString;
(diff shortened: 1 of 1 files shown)


dakrone (approved):
LGTM, thanks for fixing this

masseyke:
Looks like we did the same thing: #88772

csoulios:
Closing  this, because it is a duplicate of https://github.com/elastic/elasticsearch/pull/88772

PR #88775, opened 2022-07-25
Stable Plugin API module and analysis interfaces

pgomulka:
This commit adds stable analysis plugin API with analysis components interfaces and annotations. 
It does not contain any usage of it yet. Separate changes to introduce example plugins or refactoring to existing ones will follow later.

It contains two gradle modules. One `plugin-api` with two annotations Nameable and NamedComponent, which can be reused for plugins other than analysis. 
And second `analysis-plugin-api` which contains analysis components (TokenFilterFactory, CharFilterFactory etc)

NamedComponent - used by plugin developer - indicates that a Nameable component will be registered under a given name. 
Nameable - for analysis plugins it is only used by the stable analysis api designers (ES) - indicates that component have a name and should be declared with NamedComponent

additional tasks that will follow: https://github.com/elastic/elasticsearch/issues/88980


libs/analysis-plugin-api/src/main/java/org/elasticsearch/analysis/plugin/api/Analyzer.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * An analysis component used to create Analyzers
+ */
+public interface Analyzer<T extends org.apache.lucene.analysis.Analyzer> extends Nameable {
+    /**
+     * Returns a lucene org.apache.lucene.analysis.Analyzer instance
+     */
+    T create();
+
```
romseygeek, line 21: I think we'll need `getAnalysisMode` on here as well.

romseygeek (commented):
The API looks sensible to me. Is the plan to add this to a branch and then start migrating our internal components to use these factories?

pgomulka:
> Is the plan to add this to a branch and then start migrating our internal components to use these factories?

@romseygeek we want merge this API and infrastructure to main branch. This will be followed by example plugins implementing this API and possibly some of analysis plugins we have in our repo (we did a PoC with nori)

libs/plugin-analysis-api/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+apply plugin: 'elasticsearch.publish'
+apply plugin: 'elasticsearch.build'
+
+
+tasks.named("loggerUsageCheck").configure {enabled = false }
+
+dependencies {
+  api "org.apache.lucene:lucene-core:${versions.lucene}"
+  api "org.apache.lucene:lucene-analysis-common:${versions.lucene}"
+  api "org.apache.lucene:lucene-analysis-common:${versions.lucene}"
```
ChrisHegarty, line 18: duplicate api dependency?

libs/plugin-analysis-api/src/main/java/module-info.java
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+module org.elasticsearch.analysis.plugin.api {
+    requires org.apache.lucene.core;
+    requires org.elasticsearch.plugin.api;
```
ChrisHegarty, line 11: I noticed that there is no requires on lucene analysis common (which is in the Gradle dependencies)?  Maybe it can be removed from the Gradle api dependency?

libs/plugin-analysis-api/src/main/java/org/elasticsearch/plugin/analysis/api/AnalysisMode.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Returns a mode that is compatible with both this mode and the other mode, that is:
+     * <ul>
+     * <li>ALL.merge(INDEX_TIME) == INDEX_TIME</li>
+     * <li>ALL.merge(SEARCH_TIME) == SEARCH_TIME</li>
+     * <li>INDEX_TIME.merge(SEARCH_TIME) throws an {@link IllegalStateException}</li>
+     * </ul>
+     */
```
ChrisHegarty, line 73: can an at throws ISE be added here, as well as a trivial at returns.

libs/plugin-analysis-api/src/main/java/org/elasticsearch/plugin/analysis/api/AnalysisMode.java
```diff
@@ -0,0 +1,75 @@
...
+     * AnalysisMode representing analysis components that can be used both at index and search time
+     */
+    ALL("all") {
+        @Override
+        public AnalysisMode merge(AnalysisMode other) {
+            return other;
+        }
+    };
+
+    private String readableName;
```
ChrisHegarty, line 52: this field can be final, right?

libs/plugin-analysis-api/src/main/java/org/elasticsearch/plugin/analysis/api/CharFilterFactory.java
```diff
@@ -0,0 +1,34 @@
...
+     * Wraps the given Reader with a CharFilter.
+     * @param reader reader to be wrapped
+     * @return a reader wrapped with CharFilter
+     */
+    Reader create(Reader reader);
+
+    /**
+     * Normalize a tokenStream for use in multi-term queries
+     * The default implementation is a no-op
+     */
```
ChrisHegarty, line 30: no-op, or simply returns the given reader.

libs/plugin-analysis-api/src/main/java/org/elasticsearch/plugin/analysis/api/CharFilterFactory.java
```diff
@@ -0,0 +1,34 @@
...
+public interface CharFilterFactory extends Nameable {
+    /**
+     * Wraps the given Reader with a CharFilter.
+     * @param reader reader to be wrapped
+     * @return a reader wrapped with CharFilter
+     */
+    Reader create(Reader reader);
+
+    /**
+     * Normalize a tokenStream for use in multi-term queries
```
ChrisHegarty, line 28: can we please end all sentences with a period/stop.

libs/plugin-api/src/main/java/org/elasticsearch/plugin/api/Nameable.java
```diff
@@ -0,0 +1,29 @@
...
+package org.elasticsearch.plugin.api;
+
+/**
+ * A named plugin component. Components with a name can be registered and fetched under a name given in
+ * <code>@NamedComponent</code>
+ * @see NamedComponent
+ */
+public interface Nameable {
+
+    /**
```
ChrisHegarty, line 18: Can we please add a minimal javadoc.

libs/plugin-analysis-api/src/main/java/module-info.java
pgomulka: good point, I am not sure why I needed this in the PoC branch. We can always add it back when needed

libs/plugin-api/src/main/java/org/elasticsearch/plugin/api/NamedComponent.java
```diff
@@ -0,0 +1,27 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * An annotation used on plugin components which will be registered under a given name by Elasticsearch server.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE })
+public @interface NamedComponent {
+    /**
+     * A name used for registration and lookup
```
ChrisHegarty, line 23: Trivially, can we change this to the definitive article, "A name" -> "The name"

libs/plugin-api/src/main/java/org/elasticsearch/plugin/api/package-info.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * An api used by plugin developers to implement custom Elasticsearch plugins.
+ * The package convention in plugin apis is as follows:
+ * <ul>
+ *      <li> The root package is org.elasticsearch.plugin</li>
+ *      <li> Specialised API jars have their name following the root package.
+ *           i.e. org.elasticsearch.analysis
+ *      </li>
+ *      <li> Interfaces and annotations used by plugin developers are in `api` package
+ *           i.e org.elasticsearch.analysis.api or org.elasticsearch.api
```
ChrisHegarty, line 18: o.e.plugin.analysis.api

ChrisHegarty (approved):

libs/plugin-analysis-api/src/main/java/org/elasticsearch/plugin/analysis/api/AnalyzerFactory.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package org.elasticsearch.plugin.analysis.api;
+
+import org.elasticsearch.plugin.api.Nameable;
+
+/**
+ * An analysis component used to create Analyzers.
+ */
+public interface AnalyzerFactory<T extends org.apache.lucene.analysis.Analyzer> extends Nameable {
```
romseygeek, line 16: Do we need the generic parameter here? The other factories are just returning a base type, which I think should be enough?

pgomulka: agree, I don't think we need a generic here

romseygeek (approved):
LGTM

libs/plugin-api/src/main/java/org/elasticsearch/plugin/api/package-info.java
pgomulka: somehow I missed those comments.. I will fix this in https://github.com/elastic/elasticsearch/pull/89772

PR #88776, opened 2022-07-25
Move before allocate

idegtiarenko:


server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceComputer.java
```diff
@@ -115,16 +115,9 @@ public DesiredBalance compute(
...
-                if (targetNodesIterator.hasNext()) {
-                    unassignedShardsToInitialize.computeIfAbsent(shardRouting, ignored -> new LinkedList<>())
-                        .add(targetNodesIterator.next());
-                } else {
-                    break;
-                }
-            }
 
+            // shards are moved to the desired locations before initializing unassigned to guarantee
+            // that there are no more started shards then desired locations for them
```
DaveCTurner, line 120:
```suggestion
            // Here existing shards are moved to desired locations before initializing unassigned shards because we prefer not to leave
            // immovable shards allocated to undesirable locations (e.g. a node that is shutting down). In contrast, reconciliation prefers
            // to initialize the unassigned shards first.
```

DaveCTurner (approved):
Suggested some changes to the comment but LGTM either way.

PR #88779, opened 2022-07-25
Add FieldDataContext

romseygeek:
MappedFieldType#fieldDataBuilder() currently takes two parameters, a fully qualified
index name and a supplier for a SearchLookup.  We expect to add more parameters here
as we add support for loading fielddata from source.  Rather than telescoping the
parameter list, this commit instead introduces a new FieldDataContext carrier object
which will allow us to add to these context parameters more easily.

### DIFF
--- a/modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/MatchOnlyTextFieldMapper.java
+++ b/modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/MatchOnlyTextFieldMapper.java
@@ -31,6 +31,7 @@ import org.elasticsearch.common.lucene.Lucene;
 import org.elasticsearch.common.unit.Fuzziness;
 import org.elasticsearch.index.analysis.IndexAnalyzers;
 import org.elasticsearch.index.analysis.NamedAnalyzer;
+import org.elasticsearch.index.fielddata.FieldDataContext;
 import org.elasticsearch.index.fielddata.IndexFieldData;
 import org.elasticsearch.index.mapper.DocumentParserContext;
 import org.elasticsearch.index.mapper.FieldMapper;
@@ -43,7 +44,6 @@ import org.elasticsearch.index.mapper.TextParams;
 import org.elasticsearch.index.mapper.TextSearchInfo;
 import org.elasticsearch.index.mapper.ValueFetcher;
 import org.elasticsearch.index.query.SearchExecutionContext;
-import org.elasticsearch.search.lookup.SearchLookup;
 import org.elasticsearch.search.lookup.SourceLookup;
 
 import java.io.IOException;
@@ -54,7 +54,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.function.Function;
-import java.util.function.Supplier;
 
 /**
  * A {@link FieldMapper} for full-text fields that only indexes
@@ -276,7 +275,7 @@ public class MatchOnlyTextFieldMapper extends FieldMapper {
         }
 
         @Override
-        public IndexFieldData.Builder fielddataBuilder(String fullyQualifiedIndexName, Supplier<SearchLookup> searchLookup) {
+        public IndexFieldData.Builder fielddataBuilder(FieldDataContext fieldDataContext) {
             throw new IllegalArgumentException(CONTENT_TYPE + " fields do not support sorting and aggregations");
         }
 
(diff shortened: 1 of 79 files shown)


server/src/main/java/org/elasticsearch/index/mapper/IndexFieldMapper.java
```diff
@@ -88,6 +87,11 @@ public List<Object> fetchValues(SourceLookup lookup, List<Object> ignoredValues)
                 }
             };
         }
+
+        @Override
+        public boolean eagerGlobalOrdinals() {
+            return false;
```
nik9000, line 93: Is this related?

test/framework/src/main/java/org/elasticsearch/index/mapper/MapperServiceTestCase.java
```diff
@@ -339,6 +340,8 @@ protected final XContentBuilder runtimeFieldMapping(CheckedConsumer<XContentBuil
         });
     }
 
+    protected static FieldDataContext EMPTY_CONTEXT = new FieldDataContext("test", null);
```
nik9000, line 343: Is it better to use `staticContext` here?

I wonder about the name `static`, btw. The name gets me mixed up with the `static` keyword.

nik9000 (commented):
I very much like it.

jdconrad (approved):
@romseygeek Thank you for doing this refactoring! It looks good to me, but I would be more comfortable if @nik9000 or someone else more familiar with this area of code approved as well prior to committing.

test/framework/src/main/java/org/elasticsearch/index/mapper/MapperServiceTestCase.java
jdconrad: Maybe `empty` is a better name than `static` for the method as well?

nik9000:
I approve.

server/src/main/java/org/elasticsearch/index/mapper/IndexFieldMapper.java
romseygeek: The parent value is false as well, so it's not really necessary, but it clarifies that this won't be called during index warming and so we don't need the fully qualified index name and can happily use an empty fielddatacontext at that point.

test/framework/src/main/java/org/elasticsearch/index/mapper/MapperServiceTestCase.java
romseygeek: I've renamed to `noScriptContext`, as the primary usecase seems to be places where we don't support runtime scripts in fielddata.

server/src/main/java/org/elasticsearch/index/mapper/IndexFieldMapper.java
nik9000: Maybe a comment and make it `final`?

nik9000 (approved):

server/src/main/java/org/elasticsearch/index/mapper/IndexFieldMapper.java
romseygeek: It's on a final class so checkstyle whinges if the method is marked as final 🤷

modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/MatchOnlyTextFieldMapper.java
```diff
@@ -276,7 +275,7 @@ public Query phrasePrefixQuery(TokenStream stream, int slop, int maxExpansions,
         }
 
         @Override
-        public IndexFieldData.Builder fielddataBuilder(String fullyQualifiedIndexName, Supplier<SearchLookup> searchLookup) {
+        public IndexFieldData.Builder fielddataBuilder(FieldDataContext fieldDataContext) {
```
javanna, line 278: bikeshedding: fieldData of fielddata?

server/src/main/java/org/elasticsearch/index/IndexService.java
```diff
@@ -205,7 +206,7 @@ public IndexService(
                 this.indexSortSupplier = () -> indexSettings.getIndexSortConfig()
                     .buildIndexSort(
                         mapperService::fieldType,
-                        (fieldType, searchLookup) -> indexFieldData.getForField(fieldType, indexFieldData.index().getName(), searchLookup)
+                        (fieldType, searchLookup) -> indexFieldData.getForField(fieldType, FieldDataContext.noRuntimeFields("index sort"))
```
javanna, line 209: IndexSortConfig throws an exception too by providing a supplier of search lookup that throws exception. Which exception comes first and do we need both?

javanna:
One question I had: could you add more context on what the plan we have in mind for loading fielddata from source? What additional parameters do we plan to add for instance that make this refactoring needed? Thanks!

nik9000:
Ah! Well then. A comment would be lovely. But otherwise it's cool.

PR #88788, opened 2022-07-25
Make org.elasticsearch.cluster.routing.RoutingNode#copyShards use Array

original-brownbear:
We used this in three spots, where it copies potentially huge arrays.
One of those spots doesn't need the `copyShards` call at all
and can use the normal iterator as there's no concurrent modfication.

The other two spots can at least just use an array, which will iterate
a little faster than a mutable list and also potentially saves another
round copying the array in the `ArrayList` constructor that the compiler
seems to not be able to eliminate in all cases.

relates #77466


DaveCTurner (approved):
LGTM

original-brownbear:
Thanks David!

PR #88789, opened 2022-07-25
TSDB: Build `_id` without reparsing

nik9000:
This replaces the code that build `_id` in tsid indices that used to
re-parse the entire json object with one that reuses the parsed values.
It speed up writes by about 4%. Here's the rally output:

```
|    Min Throughput |  8164.67 |  8547.24 | docs/s | +4.69% |
|   Mean Throughput |  8891.11 |  9256.75 | docs/s | +4.11% |
| Median Throughput |  8774.52 |  9134.15 | docs/s | +4.10% |
|    Max Throughput | 10246.7  | 10482.3  | docs/s | +2.30% |
```


server/src/main/java/org/elasticsearch/index/mapper/TsidExtractingIdFieldMapper.java
```diff
@@ -125,8 +125,15 @@ public static void createField(DocumentParserContext context, BytesRef tsid) {
         ByteUtils.writeLongBE(timestamp, suffix, 8);   // Big Ending shrinks the inverted index by ~37%
```
csoulios, line 125: I would love to understand why BE yields such savings. Also, why do we write the tsid hash in LE?

This is a question for my understanding, not a comment on the code. Thanks

server/src/main/java/org/elasticsearch/cluster/routing/IndexRouting.java
```diff
@@ -251,109 +252,123 @@ public void process(IndexRequest indexRequest) {}
         public int indexShard(String id, @Nullable String routing, XContentType sourceType, BytesReference source) {
             assert Transports.assertNotTransportThread("parsing the _source can get slow");
             checkNoRouting(routing);
-            return hashToShardId(hashSource(sourceType, source));
+            return hashToShardId(hashSource(sourceType, source).buildHash(IndexRouting.ExtractFromSource::defaultOnEmpty));
         }
 
         public String createId(XContentType sourceType, BytesReference source, byte[] suffix) {
```
csoulios, line 258: This is the method that parses the source to extract the `_id`? Is it still used?

csoulios (approved):
LGTM!

I left some comments that may be ignored. I learnt a lot by reviewing it.

server/src/main/java/org/elasticsearch/index/mapper/TsidExtractingIdFieldMapper.java
nik9000: It's like 30%. It's because in big endian the high bits change very little so the search index ends up with longer common prefixes. We use little endian for writing the hash because either direction is fine and we *expect* little endian the be faster because we *expect* it is the processor's native endianness.

server/src/main/java/org/elasticsearch/index/mapper/TimeSeriesIdFieldMapper.java
```diff
@@ -190,6 +192,15 @@ public static class TimeSeriesIdBuilder implements DocumentDimensions {
          * to build the _tsid field for the document.
          */
         private final SortedMap<BytesRef, BytesReference> dimensions = new TreeMap<>();
+        /**
+         * Builds the routing. Used for building {@code _id}. If null then skipped.
+         */
+        @Nullable
+        private final IndexRouting.ExtractFromSource.Builder routingBuilder;
+
+        public TimeSeriesIdBuilder(@Nullable IndexRouting.ExtractFromSource.Builder routingBuilder) {
```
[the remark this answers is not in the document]
nik9000: It's mostly that I want to collect the state as we go. I have no real love for doing it in any particular way, but I like that I can dump the state in as we bump into the fields.

server/src/main/java/org/elasticsearch/cluster/routing/IndexRouting.java
nik9000: Just in an assertion for now. I'm being a bit paranoid about it. I'd like to keep it for a little while, just because I think we might want to use it to rebuild the `_id` on the fly rather than loading it from a stored field.

salvatore-campagna (approved):
LGTM

PR #88790, opened 2022-07-25
TSDB: Rename some methods

nik9000:
This renames a couple of methods on `IndexMode` to be more clear and
adds some javadoc.

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/IndexMode.java
+++ b/server/src/main/java/org/elasticsearch/index/IndexMode.java
@@ -17,6 +17,7 @@ import org.elasticsearch.core.Nullable;
 import org.elasticsearch.index.mapper.DataStreamTimestampFieldMapper;
 import org.elasticsearch.index.mapper.DateFieldMapper;
 import org.elasticsearch.index.mapper.DocumentDimensions;
+import org.elasticsearch.index.mapper.FieldMapper;
 import org.elasticsearch.index.mapper.IdFieldMapper;
 import org.elasticsearch.index.mapper.MapperService;
 import org.elasticsearch.index.mapper.MappingLookup;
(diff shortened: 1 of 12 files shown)


nik9000:
@elasticmachine, retest this please

nik9000:
@elasticsearchmachine retest this please

nik9000:
@elasticmachine, retest this please

nik9000:
@elasticmachine, retest this please

server/src/main/java/org/elasticsearch/index/IndexMode.java
```diff
@@ -273,9 +273,16 @@ public String getName() {
...
+     * Build the {@link FieldMapper} for {@code _id}.
+     */
     public abstract IdFieldMapper buildIdFieldMapper(BooleanSupplier fieldDataEnabled);
 
-    public abstract IdFieldMapper buildNoFieldDataIdFieldMapper();
+    /**
+     * Get the singleton {@link FieldMapper} for {@code _id}. It can never support
+     * field data.
+     */
+    public abstract IdFieldMapper idFieldMapperWithoutFieldData();
```
romseygeek, line 285: This may be a dumb question, but is there a reason we can't just call `buildIdfieldMapper(() -> false)` here instead?

romseygeek (approved):
One question, LGTM otherwise

server/src/main/java/org/elasticsearch/index/IndexMode.java
nik9000: I kind of like having a version that never builds things - just returns singletons. And I like having the behavior explicit in the name. Basically - I like it this was better :)

salvatore-campagna (commented):
LGTM

PR #88792, opened 2022-07-26
User Profile - Enforce stricter validation for literal username

ywangd:
When a security domain is configured to use literal username for user
profiles. The username becomes part of the Profile UID. Because this UID
is then used in many places, it is desirable to have stricter validation
rule for the username. For example, comma should not be allowed because
it clashes with the common way to separate multiple strings.

This PR tighten the validation to only allow alphanumeric chars for 
usernames for literal username mode.

Note that the default auto-generation mode for profile UID is
recommended in almost all cases. So this change should not be a concern
for end-users. In fact, this chagne brings the allowed characters for
literal username close to what auto-generation (base64) allows.

Relates: #86952


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -486,10 +486,12 @@ void searchVersionedDocumentForSubject(Subject subject, ActionListener<Versioned
         });
     }
 
+    private static final Pattern VALID_LITERAL_USERNAME = Pattern.compile("^[a-zA-Z0-9][a-zA-Z0-9_-]{0,255}$");
```
tvernum, line 489: Do we want to support `_` here? It would be preferable if `_` was only ever used as a separator and not part of the username.

ywangd: It might be my misunderstanding. I thought this was your suggestion. I can drop both `_` and `-`. I don't see it's necessary to have `-` here once we drop `_`.

ywangd: Discussed offline: dash (`-`) is still needed because cloud file-realm has it in usernames.

tvernum (approved):
LGTM

ywangd:
@elasticmachine update branch

ywangd:

ywangd:
@elasticmachine update branch

PR #88794, opened 2022-07-26
Speedup BalanceUnbalancedClusterTests

### LINKED ISSUE #12629: Simplify BalanceUnbalancedClusterTest
This takes 10 seconds or more, while other allocation tests are almost instantaneous. Can we simplify this? It looks like it tries to do a basic allocation (5 shards, 1 replica) of a new index when a _ton_ of indexes already exist on just 4 nodes. Perhaps we could test similar circumstances without thousands of shards? Alternatively, we could just make this an integration test (leave the impl, but rename to IT). It doesn't really seem like a unit test as it is now.

Also, as a side note, this test is the only user of CatAllocationTestCase. Perhaps we can also eliminate this abstraction and just test directly (eliminating the zipped shard state)? @s1monw do you have any thoughts here?

idegtiarenko:
This commit speeds up the above tests (and probably many others) by changing how
 we assert the invariant. Previously invariant was checked by rebuilding
 internal collections from scratch and comparing them against ones already
 present in the object after every single modification twice. This commit
 verifies the invariant once after all bulk changes.

Closes: #12629

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
+++ b/server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
@@ -15,7 +15,6 @@ import org.elasticsearch.index.Index;
 import org.elasticsearch.index.shard.ShardId;
 
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -26,7 +25,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 
 /**
  * A {@link RoutingNode} represents a cluster node associated with a single {@link DiscoveryNode} including all shards
@@ -118,7 +116,14 @@ public class RoutingNode implements Iterable<ShardRouting> {
      * @param shard Shard to create on this Node
      */
     void add(ShardRouting shard) {
-        assert invariant();
+        addInternal(shard, true);
+    }
+
+    void addNoValidate(ShardRouting shard) {
+        addInternal(shard, false);
+    }
+
+    private void addInternal(ShardRouting shard, boolean validate) {
         final ShardRouting existing = shards.putIfAbsent(shard.shardId(), shard);
         if (existing != null) {
             final IllegalStateException e = new IllegalStateException(
@@ -142,11 +147,10 @@ public class RoutingNode implements Iterable<ShardRouting> {
             relocatingShards.add(shard);
         }
         shardsByIndex.computeIfAbsent(shard.index(), k -> new HashSet<>()).add(shard);
-        assert invariant();
+        assert validate == false || invariant();
     }
 
     void update(ShardRouting oldShard, ShardRouting newShard) {
-        assert invariant();
         if (shards.containsKey(oldShard.shardId()) == false) {
             // Shard was already removed by routing nodes iterator
             // TODO: change caller logic in RoutingNodes so that this check can go away
@@ -174,7 +178,6 @@ public class RoutingNode implements Iterable<ShardRouting> {
     }
 
     void remove(ShardRouting shard) {
-        assert invariant();
         ShardRouting previousValue = shards.remove(shard.shardId());
         assert previousValue == shard : "expected shard " + previousValue + " but was " + shard;
         if (shard.initializing()) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
```diff
@@ -142,11 +147,10 @@ void add(ShardRouting shard) {
             relocatingShards.add(shard);
         }
         shardsByIndex.computeIfAbsent(shard.index(), k -> new HashSet<>()).add(shard);
-        assert invariant();
+        assert validate == false || invariant();
     }
 
     void update(ShardRouting oldShard, ShardRouting newShard) {
-        assert invariant();
```
idegtiarenko, line 149: The invariant asserted **after** operation as well. I do not think it is required to do it twice.

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNodes.java
```diff
@@ -145,6 +145,9 @@ private RoutingNodes(RoutingTable routingTable, DiscoveryNodes discoveryNodes, b
                 }
             }
         }
+        for (var node : nodesToShards.values()) {
+            assert node.invariant();
+        }
```
idegtiarenko, line 150: Invariant checking is expensive in case of big collections so it is verified once in the end instead of after every operation.

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
```diff
@@ -342,20 +345,24 @@ public boolean isEmpty() {
...
-        assert relocatingShards.containsAll(shardRoutingsRelocating);
+        for (var shard : shards.values()) {
+            if (shard.initializing()) {
+                shardRoutingsInitializing.add(shard);
+            }
+            if (shard.relocating()) {
+                shardRoutingsRelocating.add(shard);
+            }
+            shardRoutingsByIndex.computeIfAbsent(shard.index(), k -> new HashSet<>(10)).add(shard);
+        }
```
idegtiarenko, line 362: This was rebuild for assertion twice after any shard addition to the routing node.
This was especially slow for a test with 4k shards.

DaveCTurner (commented):
I wonder if we could do something more stupid instead and just skip the expensive bit of the invariant check most of the time (perhaps 99% of the time if the cluster is "large" in some sense)?

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
```diff
@@ -342,20 +345,24 @@ public boolean isEmpty() {
...
-        assert relocatingShards.size() == shardRoutingsRelocating.size();
-        assert relocatingShards.containsAll(shardRoutingsRelocating);
+        for (var shard : shards.values()) {
+            if (shard.initializing()) {
+                shardRoutingsInitializing.add(shard);
+            }
+            if (shard.relocating()) {
+                shardRoutingsRelocating.add(shard);
+            }
+            shardRoutingsByIndex.computeIfAbsent(shard.index(), k -> new HashSet<>(10)).add(shard);
```
arteam, line 361: Does it makes sense to use `Sets.newHashSetWithExpectedSize` here to correctly pre0size the `HashSet`?

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
```diff
@@ -118,7 +116,14 @@ public int size() {
      * @param shard Shard to create on this Node
      */
     void add(ShardRouting shard) {
-        assert invariant();
+        addInternal(shard, true);
+    }
+
+    void addNoValidate(ShardRouting shard) {
```
arteam, line 122: How about `addWithoutValidation`? It sounds more natural for me.

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
arteam: Nice!

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNodes.java
arteam: That makes sense. One suggestion: I think we do the invariant check in one line with method references:

```
nodesToShards.values().forEach(RoutingNode::invariant)
```

arteam (commented):
Looking good! I've left a few comments, thank you for working on this Ievgen!

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
idegtiarenko: Unfortunately we do not know expected size here. I used some random number to avoid initial resize(s)

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNode.java
```diff
@@ -342,20 +345,24 @@ public boolean isEmpty() {
         return shards.isEmpty();
     }
 
-    private boolean invariant() {
-        // initializingShards must consistent with that in shards
-        Collection<ShardRouting> shardRoutingsInitializing = shards.values().stream().filter(ShardRouting::initializing).toList();
-        assert initializingShards.size() == shardRoutingsInitializing.size();
-        assert initializingShards.containsAll(shardRoutingsInitializing);
+    boolean invariant() {
+        var shardRoutingsInitializing = new ArrayList<ShardRouting>(shards.size());
```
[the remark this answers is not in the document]
idegtiarenko: I was thinking about it, but then decided to keep Lists to avoid calculating hash for each item.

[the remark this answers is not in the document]
arteam: `hashCode` is cached for `ShardRouting`, but I agree that it does seem redundant is this case, bceause these new collections are never queried.

arteam (approved):
LGTM! Thank Ievgen!

server/src/main/java/org/elasticsearch/cluster/routing/RoutingNodes.java
original-brownbear: Link https://github.com/elastic/elasticsearch/pull/88951 :)

PR #88805, opened 2022-07-26
Serialize ID last for `UpdateApiKeyRequest`

n1v0lg:
This PR changes the order in which fields are read and written for
`UpdateApiKeyRequest` to better accommodate generalization for bulk
updates. Fields other than `id` will be shared between single operation
and bulk update requests and may therefore be extracted into a base
class; unlike constructor parameters, the order in which we read/write
fields matters for inheritance.


ywangd (approved):
LGTM assuming the base class will be abstract.

PR #88807, opened 2022-07-26
[ML] Add inference cache hit count to inference node stats

droberts195:
The inference node stats for deployed PyTorch inference
models now contain two new fields: `inference_cache_hit_count`
and `inference_cache_hit_count_last_minute`.

These indicate how many inferences on that node were served
from the C++-side response cache that was added in
https://github.com/elastic/ml-cpp/pull/2305. Cache hits
occur when exactly the same inference request is sent to the
same node more than once.

The `average_inference_time_ms` and
`average_inference_time_ms_last_minute` fields now refer to
the time taken to do the cache lookup, plus, if necessary,
the time to do the inference. We would expect average inference
time to be vastly reduced in situations where the cache hit
rate is high.

Relates elastic/ml-cpp#2374

### DIFF
--- a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/assignment/AssignmentStats.java
+++ b/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/assignment/AssignmentStats.java
@@ -35,6 +35,7 @@ public class AssignmentStats implements ToXContentObject, Writeable {
         private final Instant lastAccess;
         private final Integer pendingCount;
         private final int errorCount;
+        private final Long cacheHitCount;
         private final int rejectedExecutionCount;
         private final int timeoutCount;
         private final RoutingStateAndReason routingState;
@@ -44,6 +45,7 @@ public class AssignmentStats implements ToXContentObject, Writeable {
         private final long peakThroughput;
         private final long throughputLastPeriod;
         private final Double avgInferenceTimeLastPeriod;
+        private final long cacheHitCountLastPeriod;
 
         public static AssignmentStats.NodeStats forStartedState(
             DiscoveryNode node,
@@ -51,6 +53,7 @@ public class AssignmentStats implements ToXContentObject, Writeable {
             Double avgInferenceTime,
             int pendingCount,
             int errorCount,
+            long cacheHitCount,
             int rejectedExecutionCount,
             int timeoutCount,
             Instant lastAccess,
@@ -59,7 +62,8 @@ public class AssignmentStats implements ToXContentObject, Writeable {
             Integer numberOfAllocations,
             long peakThroughput,
             long throughputLastPeriod,
-            Double avgInferenceTimeLastPeriod
+            Double avgInferenceTimeLastPeriod,
+            long cacheHitCountLastPeriod
         ) {
             return new AssignmentStats.NodeStats(
                 node,
@@ -68,6 +72,7 @@ public class AssignmentStats implements ToXContentObject, Writeable {
                 lastAccess,
                 pendingCount,
                 errorCount,
+                cacheHitCount,
                 rejectedExecutionCount,
                 timeoutCount,
                 new RoutingStateAndReason(RoutingState.STARTED, null),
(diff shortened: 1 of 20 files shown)


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/assignment/AssignmentStats.java
```diff
@@ -167,6 +179,13 @@ public NodeStats(StreamInput in) throws IOException {
                 this.throughputLastPeriod = 0;
                 this.avgInferenceTimeLastPeriod = null;
             }
+            if (in.getVersion().onOrAfter(Version.V_8_4_0)) {
+                this.cacheHitCount = in.readOptionalLong();
```
benwtrent, line 183:
```suggestion
                this.cacheHitCount = in.readOptionalVLong();
```

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/assignment/AssignmentStats.java
```diff
@@ -300,6 +331,10 @@ public void writeTo(StreamOutput out) throws IOException {
                 out.writeVLong(throughputLastPeriod);
                 out.writeOptionalDouble(avgInferenceTimeLastPeriod);
             }
+            if (out.getVersion().onOrAfter(Version.V_8_4_0)) {
+                out.writeOptionalLong(cacheHitCount);
```
benwtrent, line 335:
```suggestion
                out.writeOptionalVLong(cacheHitCount);
```

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/assignment/AssignmentStats.java
```diff
@@ -274,6 +304,7 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
             if (avgInferenceTimeLastPeriod != null) {
                 builder.field("average_inference_time_ms_last_minute", avgInferenceTimeLastPeriod);
             }
+            builder.field("inference_cache_hit_count_last_minute", cacheHitCountLastPeriod);
```
benwtrent, line 307: It seems weird to me to include this stat when `cacheHitCount == null`. It seems like either we include both when they are missing or neither.

droberts195: You're right it's weird, but I made it match `throughput_last_minute` which it logically goes with.  Higher up the file `inference_count` is omitted when nothing has happened, but immediately above `throughput_last_minute` isn't.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ml/inference/assignment/AssignmentStats.java
```diff
@@ -462,8 +503,7 @@ public InferenceStats getOverallInferenceStats() {
             nodeStats.stream().filter(n -> n.getInferenceCount().isPresent()).mapToLong(n -> n.getInferenceCount().get()).sum(),
             // This is for ALL failures, so sum the error counts, timeouts, and rejections
             nodeStats.stream().mapToLong(n -> n.getErrorCount() + n.getTimeoutCount() + n.getRejectedExecutionCount()).sum(),
-            // TODO Update when we actually have cache miss/hit values
-            0L,
+            nodeStats.stream().filter(n -> n.getCacheHitCount().isPresent()).mapToLong(n -> n.getCacheHitCount().get()).sum(),
```
dimitris-athanasiou, line 506: `InferenceStats` expects here the cache **miss** count, right? We shouldn't be passing the hit count.

droberts195: Yes, good point.  That relates to the clashing field name conversation we had earlier.  I thought I'd reverted that line but obviously not - lucky you spotted it.

dimitris-athanasiou (approved):
LGTM

benwtrent (approved):

Issue #73356, opened 2021-05-25
ILM & SLM use wall-clock time for tracking progress


DaveCTurner:
Today we use `System#currentTimeMillis` and friends to record the timestamp of ILM phase transitions and SLM activities. For instance, in the `wait-for-snapshot` step we wait for a snapshot with a timestamp after the corresponding ILM phase transition. This mostly works, but if the nodes in the cluster do not have synchronised (and monotonic) clocks then there's a risk that we will determine that events did not happen in the correct order.

I think it would work better to use a logical clock for this purpose to avoid the problems of potential clock skew. Since these activities all correspond with cluster state updates, and since the cluster state version is monotonically increasing and updated in a coordinated fashion, I believe it would work well to use the cluster state version for the logical clock.

DaveCTurner:
Gonna call this one closed by elastic/docs-content#7912 - it's only a problem if clocks aren't synchronized, and that causes so many other issues that we now document the need for clocks to be synchronized.

PR #88811, opened 2022-07-26
Fix NPE when checking if the last snapshot was success

andreidan:
This fixes a NPE when checking the recorded failure snapshot as the
start timestamp in this case is always null.
This also moves away from checking timestamps to determin if the latest
snapshot is a success as we currently use wall-clocks for the start/end
timestamps we record and they're subject to drifts ( https://github.com/elastic/elasticsearch/issues/73356 ) .


jbaiera (approved):
LGTM, nice!

PR #88815, opened 2022-07-26
Make Settings Diffable

original-brownbear:
Making settings diffable so that index metadata diffs
are smaller whenever the metadata changes without a setting change
as well as to make index setting updates over a wider number of
indices faster.

This saves about 3% of CPU time on master and about half that on data nodes
that is just burnt for writing setting strings (mainly a result of the extremely heavy default query fields list in the Beats template I think) when bootstrapping many shards benchmarks benchmarks to 50k indices.

relates #77466


server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java
```diff
@@ -402,6 +402,25 @@ public void writeTo(StreamOutput out) throws IOException {
         }
     }
 
+    private static <K, T> Map.Entry<K, T> mapEntry(K key, T newValue) {
```
original-brownbear, line 405: Just a quick fix here to enable `null` values as we can have them for settings. The default map entries don't allow null values.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetadata.java
```diff
@@ -1420,7 +1433,13 @@ public void writeTo(StreamOutput out) throws IOException {
                 out.writeVLong(aliasesVersion);
             }
             out.writeByte(state.id);
-            Settings.writeSettingsToStream(settings, out);
+            assert settings != null
```
original-brownbear, line 1436: Not the nicest way to go about BwC but I couldn't see a neater solution.
As we never write this over cross-cluster connections this should be safe as far as I can tell.

original-brownbear:
Adding WIP label here, as I just realized I forgot dedicated test for the diffing. Will add shortly, please hold off on reviewing.

server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java
rjernst, line 405: Why the special map entry? These will be recreated by any map impls right?

server/src/main/java/org/elasticsearch/common/settings/Settings.java
```diff
@@ -598,21 +602,53 @@ public static Settings readSettingsFromStream(StreamInput in) throws IOException
...
+        return new SettingsDiff(mapDiff);
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        // pull settings to exclude secure settings in size()
+        out.writeMap(settings, StreamOutput::writeString, Settings::writeSettingValue);
+    }
+
+    private static void writeSettingValue(StreamOutput streamOutput, Object value) throws IOException {
```
rjernst, line 640: I know this was already here, and I think it was from a previous PR, but why aren't we using writeGenericValue?I do notice that we use the fully generic read above. If it's for performance, can you add a comment so that me or anyone else that wonders this in the future remembers. :)

server/src/test/java/org/elasticsearch/common/settings/SettingsTests.java
```diff
@@ -441,6 +442,17 @@ public void testWriteSettingsToStream() throws IOException {
         assertEquals(Arrays.asList("1", "2"), settings.getAsList("test.key4.foo"));
     }
 
+    public void testDiff() throws IOException {
```
rjernst, line 445: Can we have some more tests? eg a noop test, a test with more than one setting, etc. Basically, can we have something to exercise each path in the setting specific diff code you added (maybe the noop is the only additional thing needed).

rjernst (approved):
Looks good, a couple minor comments.

server/src/main/java/org/elasticsearch/cluster/DiffableUtils.java
original-brownbear: Yea, we really didn't need `Map.Entry` for the sites that we use it to begin with, `Tuple` would've been just fine but I didn't want to mix more refactoring than necessary into this PR so I went with this since weirdly enough I couldn't find a built-in JDK way of creating a `Map.Entry` that allows me a `null` value ...

server/src/main/java/org/elasticsearch/common/settings/Settings.java
original-brownbear: > If it's for performance

just that's what it's for, the difference to the generic method is massive here. Will add a comment

original-brownbear:
Jenkins run elasticsearch-ci/packaging-tests-unix-sample

original-brownbear:
Thanks Ryan! Added a noop test and a list test case to cover all possible paths.
Merging despite the red unix-sample since it's just a random distribution failing on repeats because of Jenkins issues unrelated to this change.

PR #88818, opened 2022-07-26
Support multi valued metrics

salvatore-campagna:
With this PR we handle collection of multi valued metrics
The idea is just to iterate on each value and collect all of
them as if each value belongs to a different document.

### DIFF
--- a/x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
+++ b/x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
@@ -396,15 +396,15 @@ class RollupShardIndexer {
         }
 
         public void collect(final String field, int docValueCount, final Function<Integer, Object[]> fieldValues) {
-            final Object[] value = fieldValues.apply(docValueCount);
+            final Object[] values = fieldValues.apply(docValueCount);
             if (metricFieldProducers.containsKey(field)) {
                 // TODO: missing support for array metrics
-                collectMetric(field, value[0]);
+                collectMetric(field, values);
             } else if (labelFieldProducers.containsKey(field)) {
-                if (value.length == 1) {
-                    collectLabel(field, value[0]);
+                if (values.length == 1) {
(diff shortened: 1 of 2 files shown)


x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
```diff
@@ -423,13 +423,15 @@ private void collectLabel(final String field, final Object value) {
...
-        private void collectMetric(final String field, final Object value) {
-            if (value instanceof Number number) {
-                metricFieldProducers.get(field).collect(number);
-            } else {
-                throw new IllegalArgumentException(
-                    "Expected numeric value for field '" + field + "' but got non numeric value: '" + value + "'"
-                );
+        private void collectMetric(final String field, final Object[] values) {
+            for (var value : values) {
+                if (value instanceof Number number) {
```
csoulios, line 428: I wonder if you could add one `instanceof` check outside the `for` loop would work.
Something like this: `if (values instanceof Number[])`

On the other hand we should totally remove the `instanceof` checks.

csoulios (commented):
Change looks simple and clean.  And the code does the right thing when dealing with multi-values

I look forward to seeing how the code forms after leveraging the `MultiValueMode` enum.

salvatore-campagna:
I need an approval for this. I didn't manage to use `MultiValueMode` because the change is larger than expected. As a result, I would proceed to merge this and work on other issues.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/RollupShardIndexer.java
nik9000: Yeah, I think since we want to get the `instanceof` checks out entirely it's fine to do it either way.

nik9000 (approved):

nik9000:
I *think* I'm ok to approve this while @csoulios is out for a bit. We have a bit of time before it fully lands in production and we expect a fair bit of refactoring over here eventually.

Issue #61145, opened 2020-08-14
Support for quota-aware file systems


pugnascotia:
Some users run Elasticsearch on file systems that enforce per-user quotas. The details of how such quotas work and how to query them are specific to each filesystem implementation, and there is no provision within the JVM to read them. This is unfortunate, as it means that when Elasticsearch calculates e.g. the percentage of available disk space when applying watermarks, it does so based on the entire disk's free space, not the free space that is actually available to Elasticsearch.

Java allows the default file system implementation to be overridden using the property `java.nio.file.spi.DefaultFileSystemProvider`. Elasticsearch could install a custom implementation that can account for filesystem quotas. Since the details of quota provision are completely outside the scope of Elasticsearch, this custom implementation would only provide an interface to communicate the quota and current disk usage. It would not contain logic for actually fetching this information.

We intend to make these changes for 6.8.x, 7.x (some future 7.x minor release), and master (8.0).

PR #88822, opened 2022-07-26
Document hard linking limitation in Elasticsearch < 7.11

ppf2:
We are not currently documenting the limitation in Split Index API where hard linking does not work on quota aware file systems like `xfs` (e.g. Elastic Cloud) in earlier versions of Elasticsearch < 7.11.  This addresses the missed documentation.

Ref: https://github.com/elastic/elasticsearch/issues/61145

Please back port to all affected versions (6.1 is the first version in Elasticsearch with Split Index API).

### DIFF
--- a/docs/reference/indices/split-index.asciidoc
+++ b/docs/reference/indices/split-index.asciidoc
@@ -163,9 +163,10 @@ for efficient file copying.
 [IMPORTANT]
 =====================================
 
-Hard links are not supported in Elasticsearch versions < 7.11.  When Split Index API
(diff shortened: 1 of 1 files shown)


ywelsch:
I'm not 100% sure how far back the problem goes (e.g. when quota-aware FS was introduced), and would prefer to defer the review of this to the folks involved in https://github.com/elastic/elasticsearch/issues/61145, in particular @pugnascotia.

abdonpijpelink:
@pugnascotia I think this one may have fallen through the cracks. Would you mind taking a look?

pugnascotia (approved):

PR #88823, opened 2022-07-26
Add stable indicator for plugin descriptor

rjernst:
A forgotten piece of https://github.com/elastic/elasticsearch/pull/88731
is whether a plugin descriptor in memory came from a stable or internal
descriptor. This commit adds a flag for that. Note that this is implied
by the file that was loaded, so no new property is needed.


grcevski (approved):
LGTM!

PR #88737, opened 2022-07-22
[ML] refactor some nlp tokenizer internals

benwtrent:
This an internal refactor in preparation for additional tokenizers.

PR #88825, opened 2022-07-26
[ML] fix minor tokenization bug when using fill_mask tasks with roberta tokenizer

benwtrent:
Depending on where your `<mask>` resides when using byte-part encoding (e.g. roberta), it could be split out erroneously.

This commit fixes that bug.

NOTE, this bug existed before: https://github.com/elastic/elasticsearch/pull/88737

### DIFF
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/TokenizerUtils.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/TokenizerUtils.java
@@ -39,7 +39,11 @@ final class TokenizerUtils {
                 if (neverSplitSet.contains(maybeNeverSplit)) {
                     if (windowStart < neverSplitStart) {
                         bigTokens.add(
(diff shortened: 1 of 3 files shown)


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/TokenizerUtils.java
```diff
@@ -39,7 +39,11 @@ static LinkedList<DelimitedToken> splitOutNeverSplit(CharSequence input, CharTri
                 if (neverSplitSet.contains(maybeNeverSplit)) {
                     if (windowStart < neverSplitStart) {
                         bigTokens.add(
-                            new DelimitedToken(new CharSequenceRef(input, windowStart, neverSplitStart), windowStart, neverSplitStart)
+                            new DelimitedToken(
+                                new CharSequenceRef(input, windowStart, neverSplitStart - windowStart),
```
benwtrent, line 43: `CharSequenceRef` parameters are:

 - CharSequence
 - offset
 - LENGTH

Previously it pointed to the last index of the nextsplit, not the length 🤦

dimitris-athanasiou (approved):
LGTM

PR #88826, opened 2022-07-26
Synthetic source: Support ignore_above for keyword fields

nik9000:
Support `ignore_above` for `keyword` fields, but only when they don't
have sub-fields. If you send a field longer than `ignore_above` then
it'll *vanish*. It won't come back from the synthetic `_source` because
it wasn't saved anywhere.


nik9000:
I was thinking it'd be *lovely* if we could store the field in stored fields if it was over the limit and we could use it for synthesis. That's a bit trickier. It needs #87480 plus some more machinery.

romseygeek (approved):
LGTM.  Do you think we should add something to the docs to warn about this behaviour, as I think it's reasonable but not necessarily obvious at first thought?

nik9000:
> LGTM. Do you think we should add something to the docs to warn about this behaviour, as I think it's reasonable but not necessarily obvious at first thought?

Yeah. It's going to want docs. Sorry. I forgot.

nik9000:
I'm abandoning this in favor of a real implementation of `ignore_above` based on the work I've put together in #87480.

PR #88829, opened 2022-07-26
Consolidate logic for direct vs indirect access to source in CtxMap

rjernst:
Currently CtxMap handles direct access to source through subclassing
behavior for Ingest. This commit adds a flag to CtxMap which determines
in each map method whether direct access (ie ingest) should be used or
indirect access (ie _source indirection). As a result, for indirect
access, keys other than metadata keys or _source are blocked from
update.

### DIFF
--- a/server/src/main/java/org/elasticsearch/script/CtxMap.java
+++ b/server/src/main/java/org/elasticsearch/script/CtxMap.java
@@ -159,7 +159,10 @@ public class CtxMap extends AbstractMap<String, Object> {
     public boolean containsKey(Object key) {
         // uses map directly to avoid AbstractMaps linear time implementation using entrySet()
         if (key instanceof String str) {
-            return metadata.containsKey(str) || directSourceAccess() ? source.containsKey(key) : SOURCE.equals(key);
+            if (metadata.isAvailable(str)) {
+                return metadata.containsKey(str);
+            }
+            return directSourceAccess() ? source.containsKey(key) : SOURCE.equals(key);
         }
         return false;
     }
--- a/server/src/main/java/org/elasticsearch/script/CtxMap.java
+++ b/server/src/main/java/org/elasticsearch/script/CtxMap.java
@@ -104,7 +104,7 @@ public class CtxMap extends AbstractMap<String, Object> {
         return oldSource;
     }
 
-    @SuppressWarnings({"unchecked", "raw"})
+    @SuppressWarnings({ "unchecked", "raw" })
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/elasticsearch/script/CtxMap.java
```diff
@@ -121,28 +138,30 @@ public void clear() {
         for (String key : metadata.keySet()) {
             metadata.remove(key);
         }
+        // TODO: this is just bogus, there isn't any case where metadata won't trip a failure above?
         source.clear();
```
stu-elastic, line 142: In every context there is are some read-only metadata keys, so this will currently always trip.

server/src/main/java/org/elasticsearch/script/CtxMap.java
```diff
@@ -81,7 +74,8 @@ public Metadata getMetadata() {
     @Override
     public Set<Map.Entry<String, Object>> entrySet() {
         // Make a copy of the Metadata.keySet() to avoid a ConcurrentModificationException when removing a value from the iterator
-        return new EntrySet(source.entrySet(), new HashSet<>(metadata.keySet()));
+        Set<Map.Entry<String, Object>> sourceEntries = directSourceAccess() ? source.entrySet() : Set.of(new IndirectSourceEntry());
```
stu-elastic, line 77: So `entrySet().remove(IndirectSourceEntry instance)` will fail due to `Set.of(new IndirectSourceEntry())`.  Got it.

How does it handle `entrySet().iterator().remove()`?

server/src/main/java/org/elasticsearch/script/CtxMap.java
```diff
@@ -93,7 +87,26 @@ public Object put(String key, Object value) {
...
-        return source.put(key, value);
+        if (directSourceAccess()) {
+            return source.put(key, value);
+        } else if (SOURCE.equals(key)) {
+            return replaceSource(value);
+        }
+        throw new IllegalArgumentException("Cannot put key " + key + " into ctx");
+    }
+
+    private Object replaceSource(Object value) {
```
stu-elastic, line 98: I like that you've pulled out this method.

server/src/main/java/org/elasticsearch/script/CtxMap.java
rjernst: It will throw UOE because Set.of is immutable.

rjernst:
Closing since this has been integrated into https://github.com/elastic/elasticsearch/pull/88733

PR #88834, opened 2022-07-27
[Doc] Precise that shared cache is shared across shards, not nodes

tlrx:
The sentence "[shared] cache [...] is shared across nodes in the frozen tier." is wrong, this change tries to correct that.


henningandersen (approved):
LGTM.

tlrx:
Thanks Henning

PR #88848, opened 2022-07-27
Allocation commands

idegtiarenko:
Handle allocation commands


idegtiarenko:

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/ShardsAllocator.java
```diff
@@ -48,6 +50,28 @@ default void allocate(RoutingAllocation allocation, ActionListener<Void> listene
...
+     */
+    default RoutingExplanations execute(RoutingAllocation allocation, AllocationCommands commands, boolean explain, boolean retryFailed) {
+        var originalDebugMode = allocation.getDebugMode();
+        allocation.debugDecision(true);
+        // we ignore disable allocation, because commands are explicit
+        allocation.ignoreDisable(true);
+
+        if (retryFailed) {
+            allocation.resetFailedAllocationCounter();
+        }
```
idegtiarenko, line 64: This should be moved into try as well

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceComputer.java
```diff
@@ -167,6 +169,14 @@ public DesiredBalance compute(
             }
         }
 
+        for (MoveAllocationCommand command : pendingDesiredBalanceMoves) {
+            try {
+                command.execute(routingAllocation, false);
+            } catch (RuntimeException e) {
+                logger.debug("[{}] command failed during applying it to the desired balance", command);
```
idegtiarenko, line 176: Log exception as well

server/src/test/java/org/elasticsearch/action/admin/cluster/reroute/ClusterRerouteTests.java
```diff
@@ -67,92 +69,100 @@ public void testSerializeRequest() throws IOException {
         assertEquals(req.getCommands().commands().size(), deserializedReq.getCommands().commands().size());
     }
 
-    public void testClusterStateUpdateTask() {
+    public void testClusterStateUpdateTaskInDryRun() {
```
idegtiarenko, line 72: Check if similar could be done in main first

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -373,6 +374,30 @@ public RoutingAllocation mutableCloneForSimulation() {
         );
     }
 
+    public void resetFailedAllocationCounter() {
```
idegtiarenko, line 377: Could be done in main as well

idegtiarenko: Might be moved into `routingNodes` or its unassigned iterator

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceComputer.java
DaveCTurner, line 176: log the exception too here?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceReconciler.java
```diff
@@ -71,7 +71,6 @@ void run() {
         logger.trace("Reconciler#allocateUnassigned");
         allocateUnassigned();
         assert allocateUnassignedInvariant();
-
```
DaveCTurner, line 74: nit: revert?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
```diff
@@ -211,4 +235,19 @@ private static String diff(DesiredBalance old, DesiredBalance updated) {
         }
         return builder.append(newLine).toString();
     }
+
+    private static class SynchronizedListBuffer<T> {
```
DaveCTurner, line 239: could we replace this with a more standard queue?

server/src/test/java/org/elasticsearch/cluster/routing/allocation/AllocationCommandsTests.java
```diff
@@ -170,31 +173,59 @@ public void testAllocateCommand() {
 
         logger.info("--> allocating to non-existent node, should fail");
         try {
-            allocation.reroute(clusterState, new AllocationCommands(randomAllocateCommand(index, shardId.id(), "node42")), false, false);
+            allocation.reroute(
```
DaveCTurner, line 176: I wonder if (at least in this test suite) we should be using the desired balance allocator rather than just `BalancedShardsAllocator`?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/DesiredBalanceShardsAllocator.java
```diff
@@ -155,6 +169,17 @@ public void allocate(RoutingAllocation allocation, ActionListener<Void> listener
         }
     }
 
+    @Override
+    public RoutingExplanations execute(RoutingAllocation allocation, AllocationCommands commands, boolean explain, boolean retryFailed) {
+        var explanations = ShardsAllocator.super.execute(allocation, commands, explain, retryFailed);
+        for (AllocationCommand command : commands.commands()) {
+            if (command instanceof MoveAllocationCommand move) {
+                pendingDesiredBalanceMoves.add(move);
```
DaveCTurner, line 177: Is there the possibility of weird stuff happening because we don't add the whole batch of commands to the list at once? We might execute some of the commands, rebalance a bit, and then execute some more commands. Maybe it should be a `Queue<List<MoveAllocationCommand>>` instead?

server/src/test/java/org/elasticsearch/cluster/routing/allocation/AllocationCommandsTests.java
idegtiarenko: https://github.com/elastic/elasticsearch/pull/89428

I have tried the code from above pr here and this test has passed

idegtiarenko: Fixed in https://github.com/elastic/elasticsearch/pull/88848/commits/35a5b0e9e9a2477d4a0711715cfd13d4d8f2ffb6

DaveCTurner (approved):
LGTM

PR #82685, opened 2022-01-17
Discard intermediate results upon cancellation for stats endpoints

gmarouli:
Discard intermediate results when a stats request is cancelled.

Next tasks:

- [x] Discuss approach
- [x] Improve understanding of `_cluster/state` request
- [x] Check completeness, did we cover all the endpoints that cause problems
- [x] Implement tests
- [x] Pass tests

Resolves #82337

PR #88851, opened 2022-07-27
Discard intermediate node results when a request is cancelled

gmarouli:
Backport of  #82685

### DIFF
--- a/docs/changelog/82685.yaml
+++ b/docs/changelog/82685.yaml
@@ -1,0 +1,6 @@
+pr: 82685
+summary: Discard intermediate results upon cancellation for stats endpoints
+area: Stats
+type: bug
+issues:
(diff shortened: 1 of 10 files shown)


DaveCTurner (approved):
I'm not sure if we need to modify the changelog like this - we normally backport the changelog as-is. Still, LGTM, I'll raise the changelog question out-of-band.

gmarouli:
Thanks @DaveCTurner.

At first I also backported the changelog but then the CI created a new one, so I picked that one. I can check a few other backports to see what they have done, at least then I can be consistent. 

Please if you find something out let me know.

gmarouli:
I think this happened because I forgot to add the `backport` label, I will reinstate the old change log.

PR #88855, opened 2022-07-27
[ML] Previously assigned models should get at least one allocation

dimitris-athanasiou:
When for some reason ML nodes are replaced (cluster resize, upgrade, etc.),
it is possible that some models cannot be allocated at all. Then, while
the cluster is temporarily undersized, all cores are given for allocations
of the models that have survived. If those ML nodes return later, there may
be model deployments that were previously allocated that now do not get any
allocations. The reason is that our planner will try to preserve all current
allocations.

Operationally, this is not what serves best our users. Instead, as we are
already in a cluster that does not have enough resources to fully allocate
all model deployments, we should try to give at least one allocation to each
model that has previously been allocated.

In order to know a model has previously been allocated, this commit adds a field
to `TrainedModelAssignment` called `max_assigned_allocations` which records the
max number of allocations a deployment has received in its life. We can then use
this to establish whether a deployment has ever been allocated.

Finally, we modify the `AssignmentPlanner` so that after computing a plan we
check whether the plan gives at least one allocation to all previously allocated models.
If not, we then compute a plan that tries to give at least one allocation to each
previously allocated model. We can solve this just using bin-packing. Having that
plan we can invoke the planner one more time to optimize the rest of the allocations
whilst preserving the single allocations for previously allocated models.


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlanner.java
```diff
@@ -48,8 +55,41 @@ public AssignmentPlanner(List<Node> nodes, List<Model> models) {
...
+        } else {
+            bestPlan = planSatisfyingPreviousAssignments;
+        }
+
+        logger.debug(() -> "Best plan =\n" + bestPlan.prettyPrint());
+        logger.debug(() -> prettyPrintOverallStats(bestPlan));
+        return bestPlan;
+    }
+
+    private AssignmentPlan solveSatisfyingPreviousAssignments() {
```
benwtrent, line 92: I think this is for satisfying Current assignments right (figuring out an assignment where currently assigned models still have at least one allocation on a node assignment)? 

These function names are hard to keep together.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlanner.java
```diff
@@ -63,11 +103,56 @@ public AssignmentPlan computePlan() {
...
             bestPlan = planKeepingOneAllocationOnPreviousAssignments;
         }
-        logger.debug(() -> "Best plan =\n" + bestPlan.prettyPrint());
-        logger.debug(() -> prettyPrintOverallStats(bestPlan));
         return bestPlan;
     }
 
+    private AssignmentPlan solveAllocatingAtLeastOnceModelsThatWerePreviouslyAllocated() {
+        logger.debug(() -> "Attempting to solve assigning at least one allocations to previously assigned models");
+        List<Model> previouslyAssignedModelsOnly = models.stream()
```
benwtrent, line 111: This may be a dumb question, but just to confirm. 

In the scenario where we already have allocated models and we re-calculate, `previouslyAssignedModelsOnly` also includes models that are CURRENTLY assigned. 

Basically the same as `solveSatisfyingPreviousAssignments()`, but also adding the restriction of at least one assignment for all models that have ever been allocated.

dimitris-athanasiou: Correct. Names are tricky. I'll have another think to see if I can come up with clearer terms.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/assignment/planning/AssignmentPlanner.java
dimitris-athanasiou: Correct. I'll try to come up with better naming.

dimitris-athanasiou:
@benwtrent I have done some renaming which I hope makes things a bit clearer.

benwtrent (approved):
It would be good for @wwang500 to confirm this satisfies the bug fix.

dimitris-athanasiou:
@wwang500 I have added a temporary PR to this PR where I disable the validation added in https://github.com/elastic/elasticsearch/pull/88656 so you can repeat your test exactly as you did before. You can do so by using the docker image generated by the CI.

benwtrent:
@elasticmachine update branch

dimitris-athanasiou:
@elasticmachine update branch

dimitris-athanasiou:
@elasticmachine update branch

dimitris-athanasiou:
This has now been tested. I'll proceed to merge and backport.

Issue #88758, opened 2022-07-25
Support updates of API key attributes (bulk mode)


n1v0lg:
### Description

Currently, we support [updating the attributes of a single API key](https://www.elastic.co/guide/en/elasticsearch/reference/master/security-api-update-api-key.html).

To address scalability requirements, we should support a bulk mode for this operation. In particular, we should implement a route that applies the same update to a batch of API keys, `POST _security/api_key/_bulk_update`.

The route takes a list of API key IDs (`ids`), which to update, along with the same request parameters as the single operation route:

- `role_descriptors` - The list of role descriptors specified for the key. This is one of the two parts that determines an API key’s privileges.
- `metadata_flattened` - The searchable metadata associated to an API key

Analogously to the single operation route, a call to `_bulk_update` will automatically update the `limited_by_role_descriptors`, `creator`, and `version` fields for each API key.

This breaks down into the following tasks:

- [x] REST, transport, and service layer implementation
- [x] Audit logging
- [x] REST spec and YAML tests
- [x] [DOCS] API

amitkanfer:
@n1v0lg Thank you! Will this be available in 8.4 or 8.5?

n1v0lg:
@amitkanfer this will be out in 8.5 -- I think the label wasn't available yet when I created the issue. Fixed now!

PR #88856, opened 2022-07-27
Support bulk updates of API keys

n1v0lg:
This PR adds a new API route to support bulk updates of API keys: 

`POST _security/api_key/_bulk_update`

The route takes a list of IDs (`ids`) of API keys to update, along
with the same request parameters as the single operation route:

- `role_descriptors` - The list of role descriptors specified for the
key. This is one of the two parts that determines an API key’s
privileges. 
- `metadata_flattened` - The searchable metadata associated
to an API key

Analogously to the single operation route, a call to `_bulk_update`
automatically updates the `limited_by_role_descriptors`, `creator`, and
`version` fields for each API key.

The implementation ports the single API key update operation to use the
new bulk functionality under the hood, translating as necessary at the
transport layer.

Relates: #88758


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authz/privilege/ManageOwnApiKeyClusterPrivilege.java
```diff
@@ -62,7 +63,7 @@ private ManageOwnClusterPermissionCheck() {
         protected boolean extendedCheck(String action, TransportRequest request, Authentication authentication) {
             if (request instanceof CreateApiKeyRequest) {
                 return true;
-            } else if (request instanceof UpdateApiKeyRequest) {
+            } else if (request instanceof UpdateApiKeyRequest || request instanceof BulkUpdateApiKeyRequest) {
```
n1v0lg, line 66: I'm going for an explicit list of classes instead of using `BaseUpdateApiKeyRequest` here because it's less risky to omit a potential future sub-class than to accidentally include it. One is a functional failure, the other a harder to detect potential security issue.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -384,14 +387,105 @@ public void updateApiKey(
...
+        for (String id : requestedIds) {
+            if (foundIds.contains(id) == false) {
+                responseBuilder.error(id, new ResourceNotFoundException("no API key owned by requesting user found for ID [" + id + "]"));
+            }
+        }
+    }
+
     // package-private for testing
-    void validateCurrentApiKeyDocForUpdate(String apiKeyId, Authentication authentication, ApiKeyDoc apiKeyDoc) {
+    void validateForUpdate(final String apiKeyId, final Authentication authentication, final ApiKeyDoc apiKeyDoc) {
```
n1v0lg, line 488: Nit rename: the `currentApiKeyDoc` bit didn't seem to add info so removing it for a shorter method name

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -1070,10 +1162,40 @@ private void doUpdateApiKey(
         final Set<RoleDescriptor> userRoleDescriptors,
         final VersionedApiKeyDoc currentVersionedDoc,
         final ActionListener<UpdateApiKeyResponse> listener
+    ) throws IOException {
+        final IndexRequest indexRequest = maybeBuildIndexRequest(currentVersionedDoc, authentication, request, userRoleDescriptors);
```
n1v0lg, line 1166: My plan is to follow this PR up with a refactor to port the single update method over to use `bulkUpdateApiKey` to remove the redundant code below.

n1v0lg:

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/apikey/UpdateApiKeyRequest.java
```diff
@@ -7,98 +7,44 @@
...
-    public Map<String, Object> getMetadata() {
-        return metadata;
-    }
-
-    public List<RoleDescriptor> getRoleDescriptors() {
-        return roleDescriptors;
-    }
-
-    @Override
-    public boolean equals(Object o) {
```
n1v0lg, line 93: This was only used in a single test. To avoid the headache around equality and inheritance, removing it.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -384,14 +387,105 @@ public void updateApiKey(
...
+            );
+            return;
+        }
+
+        logger.debug("Bulk updating [{}] API keys", request.getIds().size());
+        findVersionedApiKeyDocsForSubject(
+            authentication,
+            request.getIds().toArray(new String[0]),
+            ActionListener.wrap(
+                versionedDocs -> bulkUpdateApiKeys(authentication, request, userRoleDescriptors, versionedDocs, listener),
```
n1v0lg, line 419: Didn't inline because the excessive nesting was irking me

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/rest/action/apikey/RestUpdateApiKeyActionTests.java
```diff
@@ -53,6 +53,9 @@ public void testAbsentRoleDescriptorsAndMetadataSetToNull() {
 
         dispatchRequest(builder.build());
 
-        assertEquals(new UpdateApiKeyRequest(apiKeyId, null, null), requestHolder.get());
+        final UpdateApiKeyRequest request = requestHolder.get();
```
n1v0lg, line 56: Removed `equals` from the class (see my other comment) so can't rely on it here anymore.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -1083,47 +1192,42 @@ private void doUpdateApiKey(
...
+        final BulkUpdateApiKeyResponse.Builder responseBuilder,
+        final Collection<VersionedApiKeyDoc> foundDocs,
+        final List<String> requestedIds
+    ) {
+        final Set<String> foundIds = foundDocs.stream().map(VersionedApiKeyDoc::id).collect(Collectors.toUnmodifiableSet());
+        for (String id : requestedIds) {
+            if (foundIds.contains(id) == false) {
+                responseBuilder.error(id, new ResourceNotFoundException("no API key owned by requesting user found for ID [" + id + "]"));
+            }
+        }
```
ywangd, line 1230: I think we can compare size first and short circuit if size is the same.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/apikey/BulkUpdateApiKeyRequest.java
```diff
@@ -0,0 +1,64 @@
...
+
+import static org.elasticsearch.action.ValidateActions.addValidationError;
+
+public final class BulkUpdateApiKeyRequest extends BaseUpdateApiKeyRequest {
+
+    public static BulkUpdateApiKeyRequest usingApiKeyIds(String... ids) {
+        return new BulkUpdateApiKeyRequest(Arrays.stream(ids).toList(), null, null);
+    }
+
+    private final List<String> ids;
```
ywangd, line 30: Any reason for using `List` instead of `Set` for this field given we are not goiong to check for duplication?

ywangd (commented):
I'd prefer us to unify single and bulk update operations at service layer and handle their differences at handler (transportAction) layer. Concretely, I am suggestion something like the follows:

```java
class ApiKeyService {
    void updateApiKeys(
        BulkUpdateApiKeyRequest request,
        Set<RoleDescriptor> userRoleDescriptors,
        ActionListener<BulkUpdateApiKeyResponse> listener
    ) { ... }
}

class TransportBulkUpdateApiKeyAction {
    void doUpdate(
        BulkUpdateApiKeyRequest request,
        Set<RoleDescriptor> roleDescriptors,
        ActionListener<BulkUpdateApiKeyResponse> listener
    ) {
        apiKeyService.updateApiKeys(authentication, request, roleDescriptors, listener);
    }
}

class TransportUpdateApiKeyAction {
    void doUpdate(
        UpdateApiKeyRequest request,
        Set<RoleDescriptor> roleDescriptors,
        ActionListener<UpdateApiKeyResponse> listener
    ) {
        apiKeyService.updateApiKeys(
            authentication,
            BulkUpdateApiKeyRequest.wrap(request),
            roleDescriptors,
            ActionListeners.wrap(bulkResponse -> listener.onResponse(toSingleResponse(bulkResponse)), listener::onFailure)
        );
    }
}
```

The differences of the two APIs are mostly just on the surface. I think we should keep it that way in implementation as well. This is similar to how single index request is translated to bulk action at handler layer without having a separate service layer implementation.

It is also possible to have complete separate models for the service layer so that we do not reuse request and response classes of BulkUpdate. That would be my preference if we are building something from scratch. But for the sake of simplicity and how things are already working in other parts, reusing these classes is fine.

n1v0lg:
> I'd prefer us to unify single and bulk update operations at service layer and handle their differences at handler (transportAction) layer.

Unifying the two methods is essentially what I wanted to do in the next PR. I alluded to it in my comment [here](https://github.com/elastic/elasticsearch/pull/88856#discussion_r932032567). Essentially, my plan was to have the service layer method `updateApiKey` call bulk and translate. I like the idea of doing this at the transport layer 👍 

I’m happy to fold this change into this PR. My reason for not doing it was that it's essentially two separate implementation steps: the addition of new functionality for bulk, and a follow up port of existing (single operation) code to that new functionality. 

Since I have all of Friday ahead, it makes sense to just include this change in the current PR instead.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/apikey/BulkUpdateApiKeyRequest.java
n1v0lg: I don't see a strong reason to use a set here.

Using a list to me signals that the ids provided will be processed in the order they are given, whereas a set signals no particular order. I have a slight preference for the former (order). Deterministic order can come when an end-user issues a request, checks for failures, makes some changes, and re-issues the request for the same ids. We could get deterministic order with sets too but the `Set` interface does not communicate that to someone simply looking at the class/API.

From a performance perspective, a set on the request model would only help in the case where there is a large number of duplicates, which is an unusual corner case that is probably not worth optimizing for. 

Likewise, for the response and response builder classes, we essentially just need a container to append to and have no requirement for fast membership checks or uniqueness (since the implementation is already robust to duplicates in other ways). 

It's a nice bonus that for the happy path where all ids get a successfully updated, the order of the ids in the request have the same order as those in the response, if we use an array/list.

In terms of consistency with other requests and responses, I predominantly see arrays and lists (e.g., `InvalidateApiKeyRequest` uses `String[]`, `GetProfilesResponse` uses `Profile[]`).

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
n1v0lg: Good call! I considered this originally, then got into over-thinking this when it comes to duplicates, then forgot about it 😅  Will add the short-circuit check.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/apikey/BulkUpdateApiKeyResponse.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    public Map<String, Exception> getErrorDetails() {
+        return errorDetails;
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+        builder.startObject().stringListField("updated", updated).stringListField("noops", noops);
+        builder.startObject("errors");
```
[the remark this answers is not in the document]
n1v0lg: I'm fine with that. I went with the structure from the proposal where we state:

> Detailed error messages. It will not present if `errors.count` is 0.

I interpreted this to mean always display the error count. No strong preference either way.

n1v0lg:

n1v0lg:
Failure is unrelated.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -1522,7 +1526,122 @@ public void testUpdateApiKey() throws ExecutionException, InterruptedException,
...
+        assertThat(responseWithNoExpectedUpdates.getNoops(), containsInAnyOrder(apiKeyIds.toArray()));
+        assertThat(responseWithNoExpectedUpdates.getErrorDetails().keySet(), containsInAnyOrder(notFoundIds.toArray()));
+        for (String apiKeyId : apiKeyIds) {
+            final Map<String, Object> doc = getApiKeyDocument(apiKeyId);
+            expectRoleDescriptorsForApiKey("role_descriptors", newRoleDescriptors, doc);
+            expectRoleDescriptorsForApiKey("limited_by_role_descriptors", expectedLimitedByRoleDescriptors, doc);
+            expectMetadataForApiKey(newMetadata, doc);
+        }
+
+        // Check that bulk update works when all updates result in errors
```
ywangd, line 1588: I think it would be helpful to test a scenario where the update request has actually updated some API keys along with some errors in addition to everything is an error.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -1083,47 +1135,47 @@ private void doUpdateApiKey(
         if (currentDocVersion.before(targetDocVersion)) {
```
ywangd, line 1135: With the bulk update function, this method can run many times in a loop. I think it is better if we check whether debug is enabled first, i.e.
```suggestion
        if (logger.isDebugEnabled() && currentDocVersion.before(targetDocVersion)) {
```
Same for the trace logging above.

ywangd (approved):
LGTM

Some minor points for your consideration.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
n1v0lg: This one we cover in `ApiKeyRestIT.testBulkUpdateApiKey` (though we don't randomize there) -- happy to add it here as well with randomization.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -2364,4 +2553,15 @@ private void assertErrorMessage(final ElasticsearchSecurityException ese, String
...
     }
+
+    @SuppressWarnings("unchecked")
+    private static <T extends Throwable> T expectThrowsWithUnwrappedExecutionException(Class<T> expectedType, ThrowingRunnable runnable) {
+        final var ex = expectThrowsAnyOf(List.of(expectedType, ExecutionException.class), runnable);
+        if (ex instanceof ExecutionException) {
+            assertThat(ex.getCause(), instanceOf(expectedType));
+            return (T) ex.getCause();
+        } else {
+            return (T) ex;
```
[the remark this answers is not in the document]
n1v0lg: nice, thanks!

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -1522,7 +1526,122 @@ public void testUpdateApiKey() throws ExecutionException, InterruptedException,
...
+            TEST_USER_NAME,
+            new BulkUpdateApiKeyRequest(idsWithDuplicates, newRoleDescriptors, newMetadata)
+        );
+
+        assertNotNull(response);
+        assertThat(response.getErrorDetails(), anEmptyMap());
+        final List<String> allIds = Stream.concat(response.getUpdated().stream(), response.getNoops().stream()).toList();
+        assertThat(allIds, containsInAnyOrder(apiKeyIds.toArray()));
+
+        // Check not found IDs reported once for duplicates
```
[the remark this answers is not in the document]
n1v0lg: Yup you're right. The comment is misleading but the test is still useful since it validates we get an entry for each error. Will fix the comment to accurately reflect the point of the test section.

x-pack/plugin/security/qa/security-trial/src/javaRestTest/java/org/elasticsearch/xpack/security/apikey/ApiKeyRestIT.java
```diff
@@ -241,8 +287,16 @@ public void testGrantTargetCanUpdateApiKey() throws IOException {
         assertThat(apiKeyEncoded, not(emptyString()));
 
         doTestUpdateApiKey(apiKeyName, apiKeyId, apiKeyEncoded, null);
+        // bulk update also allowed
+        final var bulkUpdateApiKeyRequest = new Request("POST", "_security/api_key/_bulk_update");
+        final Map<String, Object> updateApiKeyRequestBody = Map.of("ids", List.of(apiKeyId));
+        bulkUpdateApiKeyRequest.setJsonEntity(
+            XContentTestUtils.convertToXContent(updateApiKeyRequestBody, XContentType.JSON).utf8ToString()
+        );
+        assertOK(performRequestUsingRandomAuthMethod(bulkUpdateApiKeyRequest));
```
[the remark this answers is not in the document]
n1v0lg: I went with more explicit calls here because the test suite seems to lean more towards explicit calls and less randomization. It's a little easier to see at a glance what the test actually does that way. But I'm also good with randomizing inside `doTestUpdateApiKey`, or, to keep it more explicit and a `doTestBulkUpdateApiKey` and randomize the call here instead:

`randomBoolean() ? doTestUpdateApiKey() : doTestBulkUpdateApiKey()`

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/apikey/TransportUpdateApiKeyAction.java
```diff
@@ -37,31 +39,53 @@ public TransportUpdateApiKeyAction(
...
+
+    private UpdateApiKeyResponse toSingleResponse(final String apiKeyId, final BulkUpdateApiKeyResponse response) throws Exception {
+        if (response.getErrorDetails().isEmpty() == false) {
+            if (false == (response.getErrorDetails().size() == 1
+                && response.getErrorDetails().containsKey(apiKeyId)
+                && response.getUpdated().isEmpty()
+                && response.getNoops().isEmpty())) {
+                singleMatchingResponseRequiredException(apiKeyId);
+            }
+            throw response.getErrorDetails().values().iterator().next();
```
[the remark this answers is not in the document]
n1v0lg: Good point! I will add a `getTotalResponseCount()` method to `BulkUpdateApiKeyResponse` as a starter, then we can think about promoting it to an actual field in a follow up.

> Strictly speaking, a mismatched API key ID is not really a "single response error"

Agreed, I tried to pack to much into the semantics of that exception.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
n1v0lg: Tweaked the test to use a random subset so that we test some (or sometimes all) updates failing in one test section.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -1631,13 +1677,13 @@ private static VersionedApiKeyDoc convertSearchHitToVersionedApiKeyDoc(SearchHit
...
         ) {
-            return new VersionedApiKeyDoc(ApiKeyDoc.fromXContent(parser), hit.getSeqNo(), hit.getPrimaryTerm());
+            return new VersionedApiKeyDoc(ApiKeyDoc.fromXContent(parser), hit.getId(), hit.getSeqNo(), hit.getPrimaryTerm());
         } catch (IOException ex) {
             throw new UncheckedIOException(ex);
         }
     }
 
-    private record VersionedApiKeyDoc(ApiKeyDoc doc, long seqNo, long primaryTerm) {}
+    private record VersionedApiKeyDoc(ApiKeyDoc doc, String id, long seqNo, long primaryTerm) {}
```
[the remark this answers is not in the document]
n1v0lg: Yup, including the id on `ApiKeyDoc` makes sense. Will track in a GH issue.

PR #88858, opened 2022-07-27
[ML] add sentence-piece unigram tokenizer

benwtrent:
Add internal unigram tokenizer. 

This tokenizer is the same that XLM-Roberta utilizes, along with many other cross-lingual models and tasks. 

This does not fully integrate (adding configuration, integrating into nlp tasks, etc.). But instead is just the internal tokenization and some tests showing how it runs with a precompiled charsmap.


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/DelimitedToken.java
```diff
@@ -67,6 +68,23 @@ public String toString() {
...
+            if (tokens.size() == 1) {
+                return tokens.get(0);
+            }
+            int startOffSet = tokens.get(0).startOffset();
+            int endOffset = tokens.get(tokens.size() - 1).endOffset();
+            final int encoding = tokens.get(0).encoding;
+            List<CharSequence> sequences = new ArrayList<>(tokens.size());
+            for (var t : tokens) {
+                if (t.encoding != encoding) {
+                    throw new IllegalArgumentException("all merged tokens must have the same encoding");
```
droberts195, line 81: It would be good to include the two different encoding numbers in the error message - if this ever happens it will be a useful clue for debugging.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/UnigramTokenizer.java
```diff
@@ -0,0 +1,489 @@
...
+        if (dictionary.isEmpty()) {
+            throw new IllegalArgumentException("vocab empty");
+        }
+        if (unknownToken == null) {
+            throw new IllegalArgumentException("unknown token ID");
+        }
+        CharArraySet neverSplitSet = new CharArraySet(neverSplit, false);
+        CharTrie neverSplitTree = CharTrie.build(neverSplit);
+        if (dictionary.size() != scores.size()) {
+            throw new IllegalArgumentException("Yo! These should be equal!!!");
```
droberts195, line 61: Needs a proper error message, and also include both sizes to help debugging if this ever happens.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/UnigramTokenizer.java
```diff
@@ -0,0 +1,489 @@
...
+
+        void reset() {
+            bufferIndex = 0;
+            offset = 0;
+            dataLen = 0;
+            finalOffset = 0;
+            ioBuffer.reset();
+        }
+
+        Optional<DelimitedToken> next() throws IOException {
```
droberts195, line 442: It's not necessary to change it for this PR, but I think there could be a cost to using `Optional` in loops where we expect to create and destroy many `Optional` objects very quickly.  For example see https://stackoverflow.com/a/42179644.  As a followup it's probably worth checking if we can improve our tokenization performance by switching from `Optional` to `null` in the inner loops.

droberts195 (approved):
LGTM if you could just change a couple of error messages

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/nlp/tokenizers/UnigramTokenizer.java
benwtrent: Ah, yeah! My bad :D

ikuyamada:
Hi @benwtrent,
I am highly interested in using cross-lingual pretrained LLMs in Elasticsearch. Is it already possible to use them in the current version of Elasticsearch? Because the source code is somewhat complex to me mainly with NLP background, I cannot figure it out by myself.

benwtrent:
@ikuyamada ! Multi-lingual search is tricky to get right and thank you for your interest!

There are models that do OK, but always be sure to test relevance of the results in your own datasets before using in production. We hope to support a larger range of models in the future.

Here is one model (actually a bert model) that should work, or at least get you started https://huggingface.co/sentence-transformers/distiluse-base-multilingual-cased-v1

Here is how to deploy a model. https://www.elastic.co/guide/en/machine-learning/master/ml-nlp-deploy-models.html

The task you want is a text_embedding task. Then you will want to search the created vectors.

https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html

I hope this is enough to get you started! If you have lots of questions or run into bugs, check out

https://discuss.elastic.co/

Or our slack community

https://elasticstack.slack.com/

ikuyamada:
@benwtrent Thanks so much for your prompt reply! Actually, I have already tested that wordpiece-based multilingual models (e.g., BERT and DistilUSE) work OK in the current version. However, these models tend to treat every multibyte character (e.g., kanji) as a single token, thus its expressiveness is limited.
Therefore, I asked my original question since I was interested in using more sophisticated tokenization models such as sentencepiece. Do you already have a plan to integrate such multilingual models (e.g., XLM-RoBERTa and mLUKE) into Elasticsearch in the near future?

benwtrent:
Ah, @ikuyamada , adding support for sentence-piece unigram tokenization is indeed something we want to do. 

"Near future" is a loaded concept. I do not know how soon it can be provided. But know it is something we want to provide.

ikuyamada:
@benwtrent Thanks for your reply! I think it is highly useful for multilingual neural retrieval models and I am really looking forward to it!

PR #88862, opened 2022-07-27
[ML] Fix version substitution in put DFA docs

dimitris-athanasiou:
This fixes the version substitution in a couple of response
examples in the put DFA docs.


lcawl (approved):
LGTM, thanks!

PR #88870, opened 2022-07-27
Script: Rename TIMESTAMP constant NOW in Metadata

stu-elastic:
The value is `_now` and there was a previous metadata
value `_timestamp` (see test removal in #88733) so the
name is confusing.

Also renames the method `getTimestamp()` to `getNow()`
to reflect the change.

### DIFF
--- a/modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.txt
+++ b/modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.txt
@@ -35,7 +35,7 @@ class org.elasticsearch.script.Metadata {
     String getVersionType()
     void setVersionType(String)
 
-    ZonedDateTime getTimestamp()
+    ZonedDateTime getNow()
 }
 
 class org.elasticsearch.script.IngestScript {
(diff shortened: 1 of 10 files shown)


modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.txt
```diff
@@ -35,7 +35,7 @@ class org.elasticsearch.script.Metadata {
     String getVersionType()
     void setVersionType(String)
 
-    ZonedDateTime getTimestamp()
```
rjernst, line 38: We need to provide backcompat for existing scripts. It could be done through an augmentation so it is isolated to just scripts.

rjernst (commented):
The rename seems good, but I think we need to provide backcompat in scripts.

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.txt
stu-elastic: Is that the case for Metadata?  It's in 8.4 but not documented or supported.

rjernst: Well, 8.4 has just been branched. So I guess this could change, but it will need to go back to that branch (it should be fine, since it was literally just branched today).

rjernst (approved):
LGTM

PR #84940, opened 2022-03-14
Address CCR auto follow bug when auto following data streams

martijnvg:
WIP

Relates to #81751

PR #61993, opened 2020-09-04
Add data stream support to CCR

martijnvg:
This is a draft PR, so that the overall approach can be validated before adding more tests.

This commit adds support data stream support to CCR's auto following by making the following changes:
* When the auto follow coordinator iterates over the candidate indices to follow,
  the auto follow coordinator also checks whether the index is part of a data stream and
  if the name of data stream also matches with the auto follow pattern then the index
  will be auto followed.
* When following an index, the put follow api also checks whether that index is part
  of a data stream and if so then also replicates the data stream definition to the
  local cluster.
* In order for the follow index api to determine whether an index is part of a data
  stream, the cluster state api was modified to also fetch the data stream definition
  of the cluster state if only the state is queried for specific indices.

When a data stream is auto followed, only new backing indices are auto followed.
This is in line with how time based indices patterns are replicated today. This
means that the data stream isn't copied 1 to 1 into the local cluster. The local
cluster's data stream definition contains the same name, timestamp field and
generation, but the list of backing indices may be different (depending on when
a data stream was auto followed).

Closes #56259

PR #88875, opened 2022-07-27
Fix renaming data streams with CCR replication 

### LINKED ISSUE #81751: Auto follow patterns with 'follow_index_pattern' fail to replicate data streams
A auto follow patterns with a rename pattern fails to replicate a data stream:

```
{
     "pattern": {
        "remote_cluster": "leader_cluser",
        "leader_index_patterns": ["*"],
        "follow_index_pattern": "{{leader_index}}_copy"
    }
}
```

The follow error can then occur:

```
java.lang.IllegalArgumentException: a backing index name in the local and remote cluster must remain the same
```

The problem is that only the backing index being replicated is renamed, but not the data stream. This then fails internal validation that checks that name of backing indices contains the name of the data stream.

Maybe we can drop this validation (since there are cases when we already allow for this, for example when migrating an alias to data stream) or we should also rename the data stream when this happens (so that naming is what the validation logic expect it to be).

dakrone:
This commit fixes the situation where a user wants to use CCR to replicate indices that are part of
a data stream while renaming the data stream. For example, assume a user has an auto-follow request
that looks like this:

```
PUT /_ccr/auto_follow/my-auto-follow-pattern
{
  "remote_cluster" : "other-cluster",
  "leader_index_patterns" : ["logs-*"],
  "follow_index_pattern" : "{{leader_index}}_copy"
}
```

And then the data stream `logs-mysql-error` was created, creating the backing index
`.ds-logs-mysql-error-2022-07-29-000001`.

Prior to this commit, replicating this data stream means that the backing index would be renamed to
`.ds-logs-mysql-error-2022-07-29-000001_copy` and the data stream would *not* be renamed. This
caused a check to trip in `TransportPutLifecycleAction` asserting that a backing index was not
renamed for a data stream during following.

After this commit, there are a couple of changes:

First, the data stream will also be renamed. This means that the `logs-mysql-error` becomes
`logs-mysql-error_copy` when created on the follower cluster. Because of the way that CCR works,
this means we need to support renaming a data stream for a regular "create follower" request, so a
new parameter has been added: `data_stream_name`. It works like this:

```
PUT /mynewindex/_ccr/follow
{
  "remote_cluster": "other-cluster",
  "leader_index": "myotherindex",
  "data_stream_name": "new_ds"
}
```

Second, the backing index for a data stream must be renamed in a way that does not break the parsing
of a data stream backing pattern, whereas previously the index
`.ds-logs-mysql-error-2022-07-29-000001` would be renamed to
`.ds-logs-mysql-error-2022-07-29-000001_copy` (an illegal name since it doesn't end with the
rollover digit), after this commit it will be renamed to
`.ds-logs-mysql-error_copy-2022-07-29-000001` to match the renamed data stream. This means that for
the given `follow_index_pattern` of `{{leader_index}}_copy` the index changes look like:

| Leader Cluster | Follower Cluster |
|--------------|-----------|
| `logs-mysql-error` (data stream) | `logs-mysql-error_copy` (data stream) |
| `.ds-logs-mysql-error-2022-07-29-000001`      | `.ds-logs-mysql-error_copy-2022-07-29-000001` |

Which internally means the auto-follow request turned into the create follower request of:

```
PUT /.ds-logs-mysql-error_copy-2022-07-29-000001/_ccr/follow
{
  "remote_cluster": "other-cluster",
  "leader_index": ".ds-logs-mysql-error-2022-07-29-000001",
  "data_stream_name": "logs-mysql-error_copy"
}
```

Relates to https://github.com/elastic/elasticsearch/pull/84940 (cherry-picked the commit for a test)
Relates to https://github.com/elastic/elasticsearch/pull/61993 (where data stream support was first introduced for CCR)
Resolves https://github.com/elastic/elasticsearch/issues/81751


dakrone:

dakrone:

jbaiera (approved):
LGTM!

docs/reference/ccr/apis/follow/put-follow.asciidoc
```diff
@@ -76,6 +76,26 @@ referenced leader index. When this API returns, the follower index exists, and
   (Required, string) The <<remote-clusters,remote cluster>> containing
   the leader index.
 
+[[ccr-put-follow-request-body-data_stream_name]]`data_stream_name`::
```
gmarouli, line 79: Hm, if I understand correctly, CCR works on index level. With this API what we do is to follow a single leader index that belongs to a data stream and we provide also a data stream name. I do not see from this doc, any guidance on naming the follower index or the data stream of the follower index so it could be anything right?

Furthermore, I was wondering, if CCR is defined on an index level, this means that after the leader index will be rolled over the leader data stream would not be followed anymore. Right?

So, I was wondering. Is this API maybe too "free" and error prone, users might put invalid names and potentially create weird bugs? 

Should we instead, verify the name of the follower index that it follows the data stream pattern and extract the data stream name from it?

So, if a user tries to follow an data stream backing index, for example, `.ds-logs-mysql-default_copy-2022-01-01-000001` we will infer `logs-mysql-default_copy` as the data stream name. While in the case they try to create a follower index like `.ds-logs-mysql-default-2022-01-01-000001_copy`, we will fail and let them now that this is an invalid name.

x-pack/plugin/ccr/src/main/java/org/elasticsearch/xpack/ccr/action/TransportPutFollowAction.java
```diff
@@ -303,12 +315,23 @@ private void initiateFollowing(
         );
     }
 
-    static DataStream updateLocalDataStream(Index backingIndexToFollow, DataStream localDataStream, DataStream remoteDataStream) {
+    /**
+     * Given the backing index that the follower is going to follower, the local data stream (if it
```
gmarouli, line 319:
```suggestion
     * Given the backing index that the follower is going to follow, the local data stream (if it
```

x-pack/plugin/ccr/src/test/java/org/elasticsearch/xpack/ccr/action/AutoFollowCoordinatorTests.java
```diff
@@ -1001,6 +1003,279 @@ public void testGetFollowerIndexName() {
...
+                new IndexAbstraction.DataStream(
+                    new DataStream("logs-foo-bar", List.of(index), 1, Map.of(), false, false, false, true, IndexMode.STANDARD)
+                )
+            );
+
+            PutFollowAction.Request request = AutoFollower.generateRequest("remote", index, indexAbstraction, pattern);
+            assertThat(request.getRemoteCluster(), equalTo("remote"));
+            assertThat(request.getFollowerIndex(), equalTo("my-backing-index_copy"));
+            assertThat(request.getLeaderIndex(), equalTo("my-backing-index"));
+            assertThat(request.getDataStreamName(), equalTo("logs-foo-bar_copy"));
```
gmarouli, line 1189: This test is nicely representing the concern I raised earlier. I didn't know that it was possible and it's hard for me to assess the consequences.

gmarouli (approved):
LGTM, I had some ideas/concerns shared, but I see that I am probably my understanding of the data stream is not complete :) . I would be happy to discuss those but I do not want to block this PR for this.

docs/reference/ccr/apis/follow/put-follow.asciidoc
dakrone: > Should we instead, verify the name of the follower index that it follows the data stream pattern and extract the data stream name from it?

It's worse than that, a data stream can have *any* arbitrary backing index name, because we have the [migrate to data stream](https://www.elastic.co/guide/en/elasticsearch/reference/8.3/indices-migrate-to-data-stream.html) API that converts an alias to a data stream. So a data stream named `logs-mysql-info` could have backing indices named `myfirstindex` and `other-index` following no pattern. Rather than enforce that there is a pattern with the backing index names, I opted to go with a check for something that *looks* like our regular pattern and if that doesn't work, then we treat it as an "I don't know" pattern and just do the regular name replacement.

If for some reason we have something that *looks* like a good pattern name, but is not, then we throw an exception and stop the auto-following. I'll add a test for this specific scenario (in e3001d50afaa2537ad51d58cbd9761811ffcd4f7).

> Furthermore, I was wondering, if CCR is defined on an index level, this means that after the leader index will be rolled over the leader data stream would not be followed anymore. Right?

If the leader is rolled over then the auto-follow pattern would pick up the next index in the data stream (presumably `<thing>-000002`), the data stream is not really "followed" because CCR doesn't really follow data streams, but the next index would be replicated, only the data stream would already exist because the first index would have created it already.

Things are a little different in ILM, because ILM auto-injects some `unfollow` actions around rollover, but regardless the naming thing would be the same.

dakrone:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.4|[](https://github.com/elastic/elasticsearch/pull/88995)|

### Questions ?

PR #88882, opened 2022-07-28
Fix memory calculation on circuit breaker addMemory()

luigidellaquila:
The memory has to be accounted only after circuitBreaker.addEstimateBytesAndMaybeBreak() succeeds, otherwise the cleanup is performed with a wrong value.

Same as #88538

Unfortunately, unlike sequences, unit-testing the fix here is extremely complicated, due to the fact that mock result sets of intermediate results cannot be instantiated (private constructors in Server classes)


astefan (approved):
LGTM

PR #88885, opened 2022-07-28
Use tracing API in TaskManager

pugnascotia:
Split out from #88443. Part of #84369. Use the tracing API that was
added in #87921 in TaskManager. This won't actually do anything until we
provide a tracer with an actual implementation.


grcevski (approved):
LGTM! The failures look unrelated at a first glance.

pugnascotia:
> LGTM! The failures look unrelated at a first glance.

Plot twist! They are related.

PR #88899, opened 2022-07-28
[Stack Monitoring] Cleanup unused mappings properties

klacabane:
### Summary

Part of https://github.com/elastic/kibana/issues/135379

Remove unused `service.*`, `event.*` and `host.*` properties from the stack monitoring templates. This is a cleanup change with no functional impact.
Context on how the properties ended up being defined in the first place [here](https://github.com/elastic/kibana/issues/135379#issuecomment-1197284221). Now the approach to detect which properties were not used:
- we performed a diff between the monitoring ES mappings and the one defined in the integration packages and realized differences in these fields
- verified which properties were actually populated by looking at the metricbeat document references as well as the "Available fields" column in Kibana Discover app
- only kept the fields appearing in both places

### DIFF
--- a/x-pack/plugin/core/src/main/resources/monitoring-es-mb.json
+++ b/x-pack/plugin/core/src/main/resources/monitoring-es-mb.json
@@ -2118,126 +2118,6 @@
             "type": {
               "ignore_above": 1024,
               "type": "keyword"
-            },
-            "environment": {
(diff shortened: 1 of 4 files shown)


jbaiera (approved):
I don't have as much of a grasping of the specifics of the field changes here, but the code changes LGTM 😃

Issue #88324, opened 2022-07-06
Upgrade assistant should warn of incompatible system indices settings when migrating from 7 to 8


lucabelluccini:
### Elasticsearch Version

7.x, 8.x

### Java Version

_bundled_

### OS Version

N/A

### Problem Description

System indices do not accept several settings to be overridden in 8.x - only few are allowed at the time of writing:
```
index.blocks.read_only
index.blocks.read
index.blocks.write
index.blocks.metadata
index.blocks.read_only_allow_delete
```

When migrating from 7.x to 8.x, it can happen that for some reason a system index (e.g. `.security-7`) has some setting which is allowed in 7.x but the shards fail to be allocated while upgrading to 8.x

### Steps to Reproduce

1) Create a 7.17.5 cluster.

2) Perform:
```
PUT .security-7/_settings
{
  "index.search.slowlog.threshold.query.warn": "10s",
  "index.search.slowlog.threshold.query.info": "5s",
  "index.search.slowlog.threshold.query.debug": "2s",
  "index.search.slowlog.threshold.query.trace": "500ms",
  "index.search.slowlog.threshold.fetch.warn": "1s",
  "index.search.slowlog.threshold.fetch.info": "800ms",
  "index.search.slowlog.threshold.fetch.debug": "500ms",
  "index.search.slowlog.threshold.fetch.trace": "200ms",
  "index.search.slowlog.level": "info"
}
```

3) Go to the Upgrade Assistant - All good

4) Upgrade to 8.3.1

5) The upgrade will at a given point trigger a cluster `red` state.

6) The cluster allocation explain will be:
```
{
  "can_allocate": "yes",
  "index": ".security-7",
  "target_node": {
    "attributes": {
      "server_name": "instance-0000000001.018d26f51e90476dac2a56befc09ccfc",
      "availability_zone": "us-central1-b",
      "region": "unknown-region",
      "instance_configuration": "gcp.es.datahot.n2.68x10x45",
      "xpack.installed": "true",
      "logical_availability_zone": "zone-1",
      "data": "hot"
    },
    "transport_address": "10.42.4.59:19836",
    "id": "iy10uNC9QRCJVc9xcRkJsg",
    "name": "instance-0000000001"
  },
  "node_allocation_decisions": [
    {
      "node_decision": "yes",
      "transport_address": "10.42.6.132:19611",
      "node_name": "instance-0000000000",
      "node_id": "LiMkJ1k9QUWqa-j972KGmw",
[...]

PR #88903, opened 2022-07-28
Delete invalid settings for system indices

grcevski:
When upgrading from 7 to 8 we archive any index settings that are invalid or unknown, allowing the end users to review them and delete them after confirming their configuration is still correct after the update. However, this same logic applies to system indices, where the end-users cannot change anything.

Leaving archived settings behind on system indices renders the cluster unusable in some situations, without ability for end users to remove the invalid settings.

With this PR we delete unknown/invalid settings on system indices, while archiving as usual on regular indices.

Relates to #88324


server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetadataVerifier.java
```diff
@@ -206,26 +206,50 @@ public Set<Entry<String, NamedAnalyzer>> entrySet() {
      * Identify invalid or unknown index settings and archive them. This leniency allows Elasticsearch to load
```
williamrandolph, line 206: We should update this javadoc too.

server/src/main/java/org/elasticsearch/common/settings/AbstractScopedSettings.java
```diff
@@ -956,6 +956,51 @@ public Settings archiveUnknownOrInvalidSettings(
...
+     */
+    public Settings deleteUnknownOrInvalidSettings(
+        final Settings settings,
+        final Consumer<Map.Entry<String, String>> unknownConsumer,
+        final BiConsumer<Map.Entry<String, String>, IllegalArgumentException> invalidConsumer
+    ) {
+        Settings.Builder builder = Settings.builder();
+        boolean changed = false;
+        for (String key : settings.keySet()) {
+            try {
```
williamrandolph, line 979: It looks to me like line 982 is what we're relying on to throw the IllegalArgumentException. Would it make sense to scope this try/catch block more tightly, or to add a comment on that line that `setting.get(settings)` will throw an IllegalArgumentException if a setting has an invalid value?

server/src/test/java/org/elasticsearch/cluster/metadata/IndexMetadataVerifierTests.java
```diff
@@ -26,26 +26,61 @@ public class IndexMetadataVerifierTests extends ESTestCase {
...
         assertNotSame(indexMetadata, src);
         assertEquals("-1", indexMetadata.getSettings().get("archived.index.refresh.interval"));
 
         src = newIndexMeta("foo", indexMetadata.getSettings()); // double archive?
-        indexMetadata = service.archiveBrokenIndexSettings(src);
+        indexMetadata = service.archiveOrDeleteBrokenIndexSettings(src);
+        assertSame(indexMetadata, src);
+    }
+
+    public void testArchiveBrokenSystemIndexSettings() {
```
williamrandolph, line 52: nit: `testDeleteBrokenSystemIndexSettings()` ?

williamrandolph (approved):
A few nitpicks, but otherwise LGTM!

server/src/main/java/org/elasticsearch/common/settings/AbstractScopedSettings.java
grcevski: Yeah, I was following what was done for archive. I'll add a related comment.

Issue #86603, opened 2022-05-10
Synthetic Source


nik9000:
This shrinks the index by implementing a "synthetic" _source field. Instead of saving the field to disk we reconstruct it on the fly using our column store, doc values.

### Before removing the feature flag
* [x] Initial implementation #85649
* [x] Figure out how much performance we'd get from using synthetic source for recovery - thus removing the `_recovery_source field. https://github.com/elastic/elasticsearch/pull/85649#issuecomment-1122258873
* [x] Figure out final API to turn this on (@qhoxie)
* [x] If it stays `synthetic: true` then we'll have to fail mappings that contain `enabled: false, synthetic: true` #87270
* [x] Flip API from `synthetic: true` to `synthetic: strict` so we have room to add more later. We totally will. @romseygeek 
* [x] Rally track https://github.com/elastic/rally-tracks/pull/268  + https://github.com/elastic/rally-tracks/pull/270
* [x] Resolve remaining round trip tests #86760
* [x] Realtime GET should synthesize on load to return consistent results. (#87578)
* [x] Add an option to _search to simulate synthetic source (#87068)
* [x] Support the simulate option on GET and MGET (#87536 + #87574)
* [x] Make sure we throw an error if you try to enable or disable synthetic source on an index #87182
* [x] Support "subobjects" https://github.com/elastic/elasticsearch/pull/86166 (#87261)
* [x] Docs #87416
* [x] Figure out highlighting #87667

### Later
* [x] Make sure there is a nice error message when scripts try to access synethic source - it won't be there. They should use doc values or the fancy new fields API. @tmgordeeva #88334
* [x] Add support for more field types
  * [x] `aggregate_metric_double` field type #88909
  * [x] `constant_keyword` #88603 
  * [x] `dense_vector` #89840
  * [x] `histogram` #89833 
  * [x] `keyword` fields with `ignore_above` (#87480 + #89466)
  * [x] `match_only_text` #89516
  * [x] `version` #89706
  * [x] `_doc_count` #91465
* [x] Support `fields` in runtime fields scripts
  * [x] Numbers #89888
  * [x] `ip` #89888
  * [x] `text` (#89950 + more)
  * [x] `keyword` (#89950)
  * [x] `match_only_text` (#89950)
* [x] Support loading from stored fields (text would love it!) #87480
* [ ] Rally tests for *random* fetch
* [x] Look into the `enrich` processor (#89554)
* [ ] Improve performance of synthesis
  * [x] General #87882
[...]

PR #88909, opened 2022-07-28
Add `synthetic_source` support to `aggregate_metric_double` fields

csoulios:
This PR implements `synthetic_source` support to the `aggregate_metric_double` field type

Relates to #86603


x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
```diff
@@ -675,4 +686,115 @@ protected void parseCreateField(DocumentParserContext context) throws IOExceptio
...
+                    metricDocValues.put(m, dv);
+                }
+            }
+
+            if (metricDocValues.isEmpty()) {
+                return SourceLoader.SyntheticFieldLoader.NOTHING_LEAF;
+            }
+
+            return new AggregateMetricSyntheticFieldLoader.ImmediateLeaf(metricDocValues);
+        }
```
csoulios, line 726: I could not find an easy way to leverage the `NumericSyntheticFieldLoader` class, so I had to rewrite the same implementation so that it uses one  `SortedNumericDocValues` instance per metric. 

I skipped  the singleton optimization because having mutliple docs in leaf requires an array of values. This combined with multiple metrics per doc (a `Map<Metric, SortedNumericDocValues>` will make the code extremely complex. I can do it if we think this optimization is worth the complexity

x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
```diff
@@ -675,4 +686,115 @@ protected void parseCreateField(DocumentParserContext context) throws IOExceptio
...
+            String fieldName = subfieldName(name, metric);
+            SortedNumericDocValues dv = reader.getSortedNumericDocValues(fieldName);
+            if (dv != null) {
+                return dv;
+            }
+            NumericDocValues single = reader.getNumericDocValues(fieldName);
+            if (single != null) {
+                return DocValues.singleton(single);
+            }
+            return null;
```
csoulios, line 797: Since all metric subfields are stored using the `SortedNumericDocValuesField`, al doc values will be stored as `SortedNumericDocValues`. So, I am not sure  `reader.getNumericDocValues(fieldName)` will ever return non-null values

x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
```diff
@@ -675,4 +686,115 @@ protected void parseCreateField(DocumentParserContext context) throws IOExceptio
...
+                return hasValues;
+            }
+
+            @Override
+            public void write(XContentBuilder b) throws IOException {
+                if (false == hasValues) {
+                    return;
+                }
+                b.startObject(simpleName);
+                for (Map.Entry<Metric, SortedNumericDocValues> entry : metricDocValues.entrySet()) {
```
csoulios, line 766: `metricDocValues` is an `EnumMap` instance. This means that the order that sub fields will be returned in the source, will be the order in the `Metric` enum. Order is`min`,`max`, `sum`, `value_count`. It is not alphabetical, but it will always be consistent. Does it make sense to changes this so sub fields are retuned in an alphabetical order?

x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
```diff
@@ -292,6 +296,7 @@ private NumberFieldMapper.NumberFieldType delegateFieldType(Metric metric) {
 
         /**
          * Return a delegate field type for the default metric sub-field
+         *
```
nik9000, line 299: Does your formatter do this? I think it's more standard, but I don't tend to do it myself.

x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
nik9000: Try removing it! I believe if you store single values it'll flip to `NumericDocValues` anyway.

If you do have to keep loading from both of there it's probably worth making this method `static` and `public` in `NumericFieldMapper` or something like that.

x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
nik9000: It's cool. Consistency is lovely for reading these docs. Alphabetical just gets the consistency in an easy way. No need to do it here.

x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
```diff
@@ -675,4 +686,115 @@ protected void parseCreateField(DocumentParserContext context) throws IOExceptio
...
+                // we should check the doc value for each metric separately
+                dvHasValue = new EnumMap<>(Metric.class);
+                hasValues = false;
+                for (Map.Entry<Metric, SortedNumericDocValues> e : metricDocValues.entrySet()) {
+                    boolean b = e.getValue().advanceExact(docId);
+                    hasValues = hasValues || b;
+                    dvHasValue.put(e.getKey(), b);
+                }
+
+                return hasValues;
```
nik9000, line 757: I wonder if it'd be better to use the object mapper infrastructure and have it load all of the values? That'd get most of this stuff "for free". You certainly don't have to. What you have works fine.

nik9000 (commented):
I left a few questions. I think this works fine though! I'm just poking around the edges, suggesting some maybe silly ideas. Either way, I think it's right.

x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
csoulios: If I undertand the question well, this has nothing to do with formatting the field in the output. 

This has to do with the following: 

1. Querying the field: When running a `term` or a `range` query, the query is delegated to the subfield metric that has been marked as `default_metric` in the field mapping.
2. When retrieving the field using the fields api, the value of the `default_metric` is returned (this will change when #88534 is merged)

nik9000: Sorry, no, my question was about the extra line inserted into the javadoc.

nik9000: Sorry! I wasn't clear.

csoulios: Looks like it was me messing with IntelliJ formatting. Reverted it now.

csoulios:

csoulios:
@elasticmachine update branch

x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
csoulios: As discussed offline, refactoring  the ObjectMapper so that it is reusable will probably produce less simple and readable code. So, I think we should stick with this implementation.

x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
```diff
@@ -675,4 +676,97 @@ protected void parseCreateField(DocumentParserContext context) throws IOExceptio
...
+
+            private boolean hasValues;
+
+            ImmediateLeaf(Map<Metric, SortedNumericDocValues> metricDocValues) {
+                this.metricDocValues = metricDocValues;
+            }
+
+            @Override
+            public boolean empty() {
+                return metricDocValues.isEmpty();
```
nik9000, line 731: This is never empty, right? You could maybe just return `false` here. And In the ctor you could `assert metricDocValues.empty == false`

x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
```diff
@@ -675,4 +676,97 @@ protected void parseCreateField(DocumentParserContext context) throws IOExceptio
...
+            }
+
+            @Override
+            public boolean advanceToDoc(int docId) throws IOException {
+                // It is required that all defined metrics must exist. In this case
+                // it is enough to check for the first docValue. However, in the future
+                // we may relax the requirement of all metrics existing. In this case
+                // we should check the doc value for each metric separately
+                dvHasValue = new EnumMap<>(Metric.class);
```
nik9000, line 740: Could you use an `EnumSet` instead? And maybe `clear` it instead of recreate it?

x-pack/plugin/mapper-aggregate-metric/src/main/java/org/elasticsearch/xpack/aggregatemetric/mapper/AggregateDoubleMetricFieldMapper.java
```diff
@@ -675,4 +676,97 @@ protected void parseCreateField(DocumentParserContext context) throws IOExceptio
...
+                hasValues = false;
+                for (Map.Entry<Metric, SortedNumericDocValues> e : metricDocValues.entrySet()) {
+                    boolean leafHasValues = e.getValue().advanceExact(docId);
+                    hasValues = hasValues || leafHasValues;
+                    dvHasValue.put(e.getKey(), leafHasValues);
+                }
+
+                return hasValues;
```
nik9000, line 748: I think maybe `dvHasValue.isEmpty() == false` is enough here?

nik9000 (approved):
I left a tiny thing about switching types on the tracked map. Otherwise LGTM.

PR #88910, opened 2022-07-28
[Transform] improve error handling in state persistence

### LINKED ISSUE #88905: [Transform] one transform throwing 120k same error messages in a week
In one of 8.2.0 tests, one transform (`endpoint.metadata_united-default-8.2.0`) is throwing 120k messages in a week, is it a way to decrease the number of this log entries?

`
[instance] [endpoint.metadata_united-default-8.2.0] updating stats of transform failed. java.lang.RuntimeException: Failed to persist transform statistics for transform [endpoint.metadata_united-default-8.2.0] at org.elasticsearch.xpack.transform.persistence.IndexBasedTransformConfigManager.lambda$putOrUpdateTransformStoredDoc$21(IndexBasedTransformConfigManager.java:708) [transform-8.2.0.jar:8.2.0] at org.elasticsearch.action.ActionListener$2.onFailure(ActionListener.java:170) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.support.ContextPreservingActionListener.onFailure(ContextPreservingActionListener.java:38) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.client.internal.node.NodeClient$ActionResponseTaskListener.onFailure(NodeClient.java:185) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.tasks.TaskManager$1.onFailure(TaskManager.java:185) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.support.ContextPreservingActionListener.onFailure(ContextPreservingActionListener.java:38) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.ActionListener$Delegating.onFailure(ActionListener.java:92) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.bulk.TransportSingleItemBulkWriteAction.lambda$wrapBulkResponse$0(TransportSingleItemBulkWriteAction.java:58) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.ActionListener$2.onResponse(ActionListener.java:162) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.support.ContextPreservingActionListener.onResponse(ContextPreservingActionListener.java:31) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.xpack.security.action.filter.SecurityActionFilter.lambda$applyInternal$2(SecurityActionFilter.java:163) [x-pack-security-8.2.0.jar:8.2.0] at org.elasticsearch.action.ActionListener$DelegatingFailureActionListener.onResponse(ActionListener.java:245) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.ActionListener$RunBeforeActionListener.onResponse(ActionListener.java:415) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.bulk.TransportBulkAction$BulkOperation$1.finishHim(TransportBulkAction.java:612) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.bulk.TransportBulkAction$BulkOperation$1.onResponse(TransportBulkAction.java:593) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.bulk.TransportBulkAction$BulkOperation$1.onResponse(TransportBulkAction.java:582) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.client.internal.node.NodeClient$ActionResponseTaskListener.onResponse(NodeClient.java:175) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.tasks.TaskManager$1.onResponse(TaskManager.java:176) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.tasks.TaskManager$1.onResponse(TaskManager.java:170) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.support.ContextPreservingActionListener.onResponse(ContextPreservingActionListener.java:31) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.xpack.security.action.filter.SecurityActionFilter.lambda$applyInternal$2(SecurityActionFilter.java:163) [x-pack-security-8.2.0.jar:8.2.0] at org.elasticsearch.action.ActionListener$DelegatingFailureActionListener.onResponse(ActionListener.java:245) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.support.replication.TransportReplicationAction$ReroutePhase.finishOnSuccess(TransportReplicationAction.java:1069) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.support.replication.TransportReplicationAction$ReroutePhase$1.handleResponse(TransportReplicationAction.java:977) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.action.support.replication.TransportReplicationAction$ReroutePhase$1.handleResponse(TransportReplicationAction.java:968) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.transport.TransportService$ContextRestoreResponseHandler.handleResponse(TransportService.java:1340) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.transport.InboundHandler.doHandleResponse(InboundHandler.java:365) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.transport.InboundHandler.handleResponse(InboundHandler.java:352) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.transport.InboundHandler.messageReceived(InboundHandler.java:142) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.transport.InboundHandler.inboundMessage(InboundHandler.java:94) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.transport.TcpTransport.inboundMessage(TcpTransport.java:795) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.transport.InboundPipeline.forwardFragments(InboundPipeline.java:149) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.transport.InboundPipeline.doHandleBytes(InboundPipeline.java:121) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.transport.InboundPipeline.handleBytes(InboundPipeline.java:86) [elasticsearch-8.2.0.jar:8.2.0] at org.elasticsearch.transport.netty4.Netty4MessageInboundHandler.channelRead(Netty4MessageInboundHandler.java:63) [transport-netty4-8.2.0.jar:8.2.0] at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.handler.logging.LoggingHandler.channelRead(LoggingHandler.java:280) [netty-handler-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:103) [netty-codec-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.handler.ssl.SslHandler.unwrap(SslHandler.java:1372) [netty-handler-4.1.74.Final.jar:4.1.74.Final] at io.netty.handler.ssl.SslHandler.decodeJdkCompatible(SslHandler.java:1235) [netty-handler-4.1.74.Final.jar:4.1.74.Final] at io.netty.handler.ssl.SslHandler.decode(SslHandler.java:1284) [netty-handler-4.1.74.Final.jar:4.1.74.Final] at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:510) [netty-codec-4.1.74.Final.jar:4.1.74.Final] at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:449) [netty-codec-4.1.74.Final.jar:4.1.74.Final] at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:279) [netty-codec-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:919) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:166) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:722) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.nio.NioEventLoop.processSelectedKeysPlain(NioEventLoop.java:623) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:586) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:496) [netty-transport-4.1.74.Final.jar:4.1.74.Final] at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:986) [netty-common-4.1.74.Final.jar:4.1.74.Final] at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) [netty-common-4.1.74.Final.jar:4.1.74.Final] at java.lang.Thread.run(Thread.java:833) [?:?] Caused by: org.elasticsearch.index.engine.VersionConflictEngineException: [data_frame_transform_state_and_stats-endpoint.metadata_united-default-8.2.0]: version conflict, required seqNo [5879054], primary term [22]. current document has seqNo [5879057] and primary term [22]
`

hendrikmuhs:
transform persists the internal state of a transform (e.g. the data cursor) in state document.
This change improves the error handling and fixes the problem described in #88905. A transform
can now recover from this problem.

fixes #88905

side-effect: state persistence is now integrated into error handling and therefore a transform can
fail if state persistence remains broken. Previously the transform continued to operate. Note, if
state persistence is broken, failover does not work properly.


x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/ClientTransformIndexer.java
```diff
@@ -326,16 +332,50 @@ protected void persistState(TransformState state, ActionListener<Void> listener)
...
                     );
-                    assert false : "[" + getJobId() + "] updating stats of transform failed, unexpected version conflict of internal state";
+
+                    if (handleStatePersistenceFailure(statsExc) == false) {
+                        // get the current seqNo and primary term, however ignore the stored state
+                        transformsConfigManager.getTransformStoredDoc(
+                            transformConfig.getId(),
+                            false,
+                            ActionListener.wrap(storedDocAndSeqNoPrimaryTerm -> {
+                                updateSeqNoPrimaryTermAndIndex(seqNoPrimaryTermAndIndex, storedDocAndSeqNoPrimaryTerm.v2());
```
benwtrent, line 342: Why is this necessary if the previous storage succeeded?

We set the values when the previous call that caused THIS call to fail.

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/ClientTransformIndexer.java
```diff
@@ -314,7 +320,7 @@ protected void persistState(TransformState state, ActionListener<Void> listener)
                     // - this is not a catastrophic failure, if 2 state persistence calls run at the same time, 1 should succeed and update
                     // seqNoPrimaryTermAndIndex
                     // - for tests fail(assert), so we can debug the problem
-                    logger.error(
+                    logger.warn(
```
benwtrent, line 323: We are now treating this as a "catastrophic" failure in that we will fail the transform after 10 failures. 

Which, I am not sure will ever happen because the failure only occurs if one other stats storage succeeded before and incremented the doc seq_no and primary_term.

benwtrent (commented):
I don't think this improves anything. 

We already update our stored seq_no && primary_term on every storage update. 

So, a conflict can only happen if more than one updates occur with the same view of `seqNoPrimaryTermAndIndex`. Then all but one of those updates will fail. The one that succeeds updates `seqNoPrimaryTermAndIndex` then the error recovery for all the other failed ones doesn't do anything.

I am also not sure about failing the transform for a thing that is completely out of the users control. I don't think we should ever do that. Stats storage is an internal concern and the user cannot fix it.

This is contrary to other terminal errors like mapping conflicts, bad scripts, etc. which are user provided and controlled.

hendrikmuhs:
> I don't think this improves anything.
> 
> We already update our stored seq_no && primary_term on every storage update.
> 

In the case of a version conflict seq_no && primary_term aren't updated. That's why this causes an endless loop of failures. It _never_ recovers from this situation, because it tries to update the doc with the same seq_no/primary_term over and over. See #88905 which originates from cloud logs.

> So, a conflict can only happen if more than one updates occur with the same view of `seqNoPrimaryTermAndIndex`. Then all but one of those updates will fail. 

That's the theory and that was certainly the case in the past. #88905 teaches us, that there is still a race condition where state persistence goes nuts. If this originates from 2 calls in flight, the situation should recover, but this is not the case.

My concern about this fix is that we don't detect the situation - which should _not_ happen in the 1st place goes undetected, because I had to remove an `assert`. However for integration tests we set num_failures to 0, so the test would fail.

> The one that succeeds updates `seqNoPrimaryTermAndIndex` then the error recovery for all the other failed ones doesn't do anything.

This is definitely not the case as reported in #88905, the cluster loops with this error for ages. As a consequence state never gets persisted but the transform continues to work. This is a time bomb, as soon as the task stop or relocates - e.g. because of an update - the transform is not able to continue from the last position, but starts from a position way back. This will result in refetching and rewriting a potentially large historic backlog which might not even be available anymore.

> 
> I am also not sure about failing the transform for a thing that is completely out of the users control. I don't think we should ever do that. Stats storage is an internal concern and the user cannot fix it.

I agree this is an _internal error_, nevertheless it must be handled as any other internal error that can happen in software. The fix is usually a restart and that's the case here as well. The transform fails and must be started again. Therefore the user can fix it, because he can stop/start a transform and he can use indefinite retries. The problem should not re-occur after the stop/starting. I described the consequence of _doing nothing_ above, I think it is better to fail and let the user fix it by restarting the transform -iff the added self-recovery does not work- than to continue and pile up an historic backlog. Note that this can even cause data loss.

> 
> This is contrary to other terminal errors like mapping conflicts, bad scripts, etc. which are user provided and controlled.

We make no distinction between user controlled errors and other errors, a transform can also fail if search/bulk indexing breaks for any reason including a broken cluster.

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/ClientTransformIndexer.java
hendrikmuhs: The change _decreases_ the log level from `ERROR` to `WARN` not vice versa.

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/ClientTransformIndexer.java
hendrikmuhs: The storage failed and this gets the current seq_no/primary_term needed to hopefully succeeds on the next state persistence. Otherwise state persistence endlessly tries to update the doc with the same outdated seq_no forever.

In case of 2 concurrent state persistence calls - which is just a theory and actually should not happen anymore - it won't have an effect, because in order to update the reference it only updates if the old value matches. If this happens it gets logged as a warning. Even in that case state persistence should recover.

hendrikmuhs:
Note: I don't know the reason for the failure in #88905, but it obviously happens. It seems to be something else than 2 calls in flight, which I think is not possible anymore, but some other internal elasticsearch problem. This PR addresses this issue and lets the system recover.

The error scenario itself should not happen in the 1st place. The integration tests assert this behavior, for long running tests run by QA we should monitor the logs/audits for state persistence errors.

hendrikmuhs:

benwtrent:
> In the case of a version conflict seq_no && primary_term aren't updated. That's why this causes an endless loop of failures. It never recovers from this situation, because it tries to update the doc with the same seq_no/primary_term over and over. See https://github.com/elastic/elasticsearch/issues/88905 which originates from cloud logs.

Could we detect a version conflict failure and adjust seq_no and primary_term in the failure path instead of making a second call?

> That's the theory and that was certainly the case in the past. https://github.com/elastic/elasticsearch/issues/88905 teaches us, that there is still a race condition where state persistence goes nuts. If this originates from 2 calls in flight, the situation should recover, but this is not the case.

This signals a WAY bigger bug than what we are trying to fix. This means something out of band of the transform service is updating the document. 

In case #88905 do the logging messages all have the same seq_no and primary_term? Or are they continually changing in subsequent calls? 

If they are continually changing, then SOMETHING is changing the document. If they are never changing, then that means something out of band updated the stats doc, and to me that is pretty bad.

> I agree this is an internal error, nevertheless it must be handled as any other internal error that can happen in software. The fix is usually a restart and that's the case here as well. The transform fails and must be started again. 

I would argue that "any other internal error" is a bad description here if you are planning on failing the transform and requiring a reset from the previous checkpoint with user intervention. It is possible for many internal errors to be recovered from.

In short:

 - something updating the transforms stats document out of band is bad news. That seems like a way bigger bug to me. (unless its the user somehow doing it directly)
 - Can we get the seq_no && primary_term from the failure path without making a second `GET` call? 
 - I don't think this should ever fail the transform.

hendrikmuhs:
> Could we detect a version conflict failure and adjust seq_no and primary_term in the failure path instead of making a second call?

Not in a good way. The exception does not support getting  `seq_no` and `primary_term`, but the exception is just a string. There is a convenience constructor taking versions but it does not keep them. The constructor creates the message string and drops the versions. Another constructor takes an ordinary string, so the version aren't mandatory. I think we agree that string parsing is _not_ an option. The other option would be to make `seq_no` and `primary_term` members, that requires that they are mandatory. I remember discussing this issue once with the core team and we agreed that this is _not_ the way to do it. We agreed that the right way of using optimistic concurrency control is to re-get the doc when a conflict occurs, do the change again and push it again.

> This signals a WAY bigger bug than what we are trying to fix. This means something out of band of the transform service is updating the document.

Yes, the glitch is unexpected and this PR does _not_ fix how we got into that issue. It implements a way to recover and it fixes the log spam.

Note a change of `primary_term` is an expected result after a failure:

```
These primary terms are incremental and change when a primary is promoted.
``` 
(https://www.elastic.co/blog/elasticsearch-sequence-ids-6-0)

With other words: If the node that holds the transform internal index primary shard dies, a new primary gets promoted and `primary_term` is upped.

`seq_no` changes for every index operation, so the change we see in the issue is unexpected and it seems that it is coming from outside. But I am not sure, I don't know if there is a valid scenario where a seq number gets upped twice, e.g. maybe after a failure and translog replay? Or it might be a lost message that gets resend and than ends up twice (at least once queue model). I can find out.

Nevertheless, to protect against outside changes, this PR _doesn't_ take the state from the document, but only take the 2 version numbers. The state gets overwritten by the indexer on the next state persistence. For a reason I did _not_ implement an immediate retry, because if the cluster is in trouble it is not a good idea to retry immediately.

> In case #88905 do the logging messages all have the same seq_no and primary_term? Or are they continually changing in subsequent calls?

It remains the same. It looks like that there is one doc change that happens somehow externally or due to a bug. Afterwards the transform fails over and over with the same problem/message. The versions are never changed and it retries forever to push a doc with the outdated versions.

> 
> If they are continually changing, then SOMETHING is changing the document. If they are never changing, then that means something out of band updated the stats doc, and to me that is pretty bad.

No, for all the cases that I know - the linked issue is not the 1st one, but the most recent one - somehow the seq_no got updated and state persistence fails "forever" with the same error.

> I don't think this should ever fail the transform.

For the issue this PR is supposed to fix, this should not happen as the next state persistence should succeed. Only if there is really something that permanently changes the state doc the 10 retries will apply.

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/ClientTransformIndexer.java
benwtrent: I am meaning the comment above the message.

benwtrent (approved):
Thanks for clarifying my concerns :D. 

I agree with this change. I am still not convinced that we should even fail the transform if this thing fails 10 times in a row, but that will most likely never happen. 

Yeah, in the case of the primary changing, we would fail for forever, which is problematic. And since we cannot get the seq && primary from the failure, getting the doc as a recovery mechanism is good.

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/ClientTransformIndexer.java
hendrikmuhs: good point, I will update the documentation

PR #88912, opened 2022-07-28
Speed up operations on BlobStoreIndexShardSnapshots

original-brownbear:
This fixes a couple of slow points in `BlobStoreIndexShardSnapshots`,
which become performance critical when working with large repositories.

1. Fix `physicalFiles` containg the same `FileInfo` instances repeatedly for every
snapshot that holds the file. Without this fix the map can hold lists as long as the
number of snapshots for the shard for files common to all snapshots of the shard.
Also, only lazy build the map since it's only used during snapshotting and internalize
the logic into `BlobStoreIndexShardSnapshots` so we don't have to bother with wrapping
as unmodifiable.
2. Add efficient copy constructors for all 3 operations on the shard to avoid
expensive looping over all snapshots and their files in many cases.
3. Use list instead of redundant map in deserialization, we weren't using the map
for any deduplication anyways and are safe here thanks to Jackson's duplicate name
detection

relates #77466


server/src/main/java/org/elasticsearch/index/snapshots/blobstore/BlobStoreIndexShardSnapshots.java
```diff
@@ -33,54 +38,53 @@
...
+            updatedSnapshots.add(snapshot);
             for (FileInfo fileInfo : snapshot.indexFiles()) {
                 FileInfo oldFile = newFiles.put(fileInfo.name(), fileInfo);
                 assert oldFile == null || oldFile.isSame(fileInfo);
             }
-            // We are doing it in two loops here so we keep only one copy of the fileInfo per blob
-            // the first loop de-duplicates fileInfo objects that were loaded from different snapshots but refer to
-            // the same blob
-            for (FileInfo fileInfo : snapshot.indexFiles()) {
-                physicalFiles.computeIfAbsent(fileInfo.physicalName(), k -> new ArrayList<>()).add(newFiles.get(fileInfo.name()));
```
original-brownbear, line 59: This was the culprit of huge lists with duplicate `FileInfo` entries

original-brownbear:

original-brownbear:
@elasticmachine update branch

tlrx (approved):
LGTM

original-brownbear:
Thanks Tanguy!

PR #88917, opened 2022-07-28
Remove unused plugins dir var from server CLI

pugnascotia:
Split out of #88443. Remove the now-unused plugins directory variable from the
server CLI code.


grcevski (approved):
LGTM!

rjernst (approved):
LGTM too

PR #88920, opened 2022-07-28
Wrap code in new tracing contexts where required

pugnascotia:
Part of #84369. Split out from #88443. This PR wraps parts of the code
in a new tracing context. This is necessary so that a tracing
implementation can use the thread context to propagate tracing headers,
but without the code attempting to set the same key twice in the thread
context, which is illegal.

Note that in some places we actually clear the tracing context
completely. This is done where the operation to be performed should have
no association with the current trace context. For example, when
creating a new index via a REST request, the resulting background tasks
for the index should not be associated with the REST request in
perpetuity.


grcevski (approved):
LGTM!

pugnascotia:
I'll hold on merging this until @original-brownbear has had a change to review.

server/src/main/java/org/elasticsearch/transport/InboundHandler.java
```diff
@@ -294,14 +307,7 @@ public boolean isForceExecution() {
...
                                     @Override
-                                    public void onFailure(Exception e) {
-                                        sendErrorResponse(reg.getAction(), transportChannel, e);
-                                    }
-
-                                    @Override
-                                    public void onAfter() {
-                                        request.decRef();
-                                    }
+                                    public void onFailure(Exception e) {}
```
original-brownbear, line 310: This is not ok, the executor might directly call `onFailure` when it rejects the runnable because its queue is full. We have to go through the normal error handling mechanism here in this callback if we are using `AbstractRunnable`.

server/src/main/java/org/elasticsearch/transport/TransportService.java
```diff
@@ -966,19 +977,12 @@ public boolean isForceExecution() {
                         }
 
                         @Override
-                        public void onFailure(Exception e) {
-                            handleSendToLocalException(channel, e, action);
-                        }
+                        public void onFailure(Exception e) {}
```
original-brownbear, line 980: Same here, can't do this. We need to use the callback when using the exeuctor with an `AbstractRunnable`

server/src/main/java/org/elasticsearch/transport/TransportService.java
```diff
@@ -966,19 +977,12 @@ public boolean isForceExecution() {
...
 
                         @Override
                         public String toString() {
                             return "processing of [" + requestId + "][" + action + "]: " + request;
                         }
-
-                        @Override
-                        public void onAfter() {
-                            request.decRef();
```
original-brownbear, line 980: Where did this go, regardless of the fact that we need to do this differently, it seems we lost the `decRef` here?

original-brownbear (changes requested):
Commented inline, we have to do this a little differently.

server/src/main/java/org/elasticsearch/common/util/concurrent/AbstractRunnable.java
```diff
@@ -20,6 +20,13 @@ public boolean isForceExecution() {
         return false;
     }
 
+    /**
+     * Should the runnable start a new tracing context before it executes?
+     */
+    public boolean useNewTraceContext() {
```
original-brownbear, line 26: Maybe it makes more sense to just create a utility method on `ContextPreservingAbstractRunnable` that wraps an `AbstractRunnable` in a `ContextPreservingAbstractRunnable` with new trace context `true` flag set to `true` and use that?
That way we don't have to leak the high level tracing into low level `AbstractRunnable`. It's also probably less verbose than overriding to return `true` in multiple spots.

server/src/main/java/org/elasticsearch/cluster/InternalClusterInfoService.java
```diff
@@ -181,113 +181,117 @@ void execute() {
             nodesStatsRequest.clear();
             nodesStatsRequest.addMetric(NodesStatsRequest.Metric.FS.metricName());
             nodesStatsRequest.timeout(fetchTimeout);
-            client.admin().cluster().nodesStats(nodesStatsRequest, ActionListener.runAfter(new ActionListener<>() {
-                @Override
-                public void onResponse(NodesStatsResponse nodesStatsResponse) {
-                    logger.trace("received node stats response");
+            try (var ignored = threadPool.getThreadContext().clearTraceContext()) {
+                client.admin().cluster().nodesStats(nodesStatsRequest, ActionListener.runAfter(new ActionListener<>() {
```
original-brownbear, line 185: This change is fairly extreme in terms of how noisy it is line wise.
Maybe, also in the sense of future readability of this code, could we extract what's inside the `try-with-resources` into a new method here and in the other spots that got indented so extremely now? That way the change-set stays small because the code doesn't move in terms of indent level and things don't get much more complicated.

original-brownbear (commented):
I think I'd prefer less magic in terms of how we create the `ContextPreservingAbstractRunnable` and just do it outright as commented on inline. But other than that, this looks like it should work fine.

server/src/main/java/org/elasticsearch/common/util/concurrent/AbstractRunnable.java
pugnascotia: I’m struggling to see how this would work. `ContextPreservingAbstractRunnable` is private and only created in `ThreadContext.preserveContext(Runnable),` but I need to communicate the need for a new trace context from where the runnable is created in the first place, for example in `InboundHandler`.

I thought about creating a named sub-class of `AbstractRunnable` e.g. `TracingAbstractRunnable` or something like that, but I couldn't see how that would avoid the need to changing `AbstractRunnable`, at least not without duplicating much of `ContextPreservingAbstractRunnable`.

I don't think there's a simpler answer for where to insert this logic, because there's a lot of parts that are collaborating here. In the case of `InboundHandler`, we see:

   * `InboundHandler` creates an `AbstractRunnable` instance and schedules it
   * The scheduler asks the thread context to wrap the runnable
   * The thread context potentially wraps the runnable
   * The scheduler executes the runnable
   * `AbstractRunnable#run` orchestrates `doRun`, `onFailure` etc

We could change the visibility of `ContextPreservingAbstractRunnable` maybe, and sometimes create one explicitly? e.g. `ContextPreservingAbstractRunnable.preservingContext(AbstractRunnable)`?

original-brownbear: I think you could just add a method to `ThreadContext` similar to `org.elasticsearch.common.util.concurrent.ThreadContext#preserveContext`  that takes an `AbstractRunnable` and does the wrapping for you without even having to expose `ContextPreservingAbstractRunnable`? 

You have the thread context available via the `threadPool` and the threadpool will just pass through a `ContextPreservingAbstractRunnable` unchanged when executing => less code in the callsites and no need to change anything about `AbstractRunnable` right?

pugnascotia: @original-brownbear ah, you're right. I had another go, please take another look.

server/src/main/java/org/elasticsearch/cluster/InternalClusterInfoService.java
pugnascotia: I refactored most of these sites with the extra wrapping to extract methods.

original-brownbear (approved):
LGTM thanks for the extra iterations Rory!

PR #88924, opened 2022-07-29
Deprecate network plugins

rjernst:
Network plugins provide network implementations. In the past this has
been used for alternatives to netty based networking, using the JDK's
nio. However, nio has now been removed, and it is inadvisable for a
plugin to implement this low level part of the system.
Therefore, this commit marks the NetworkPlugin interface as deprecated.

### DIFF
--- a/docs/reference/migration/migrate_8_5.asciidoc
+++ b/docs/reference/migration/migrate_8_5.asciidoc
@@ -20,3 +20,36 @@ coming::[8.5.0]
 There are no breaking changes in {es} 8.5.
 // end::notable-breaking-changes[]
 
+[discrete]
+[[deprecated-8.5]]
+=== Deprecations
+
+The following functionality has been deprecated in {es} 8.3
+and will be removed in a future version.
+While this won't have an immediate impact on your applications,
+we strongly encourage you take the described steps to update your code
+after upgrading to 8.3.
+
+To find out if you are using any deprecated functionality,
+enable <<deprecation-logging, deprecation logging>>.
+
+
+[discrete]
+[[deprecations_85_network_plugins]]
+==== Plugin deprecations
+
+[[network_plugins_deprecated]]
+Plugins that extend the NetworkPlugin interface are deprecated.
+[%collapsible]
+====
+*Details* +
+Plugins may override funcionality that controls how nodes connect
+with other nodes over TCP/IP. These plugins extend the NetworkPlugin
+interface. In the next major release, these plugins will fail
+to install.
+
+*Impact* +
+Discontinue using any plugins which extend NetworkPlugin. You can
+see if any plugins use deprecated functionality by checking
+the Elasticsearch deprecation log.
+====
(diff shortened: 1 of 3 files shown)


rjernst:
Note: This PR doesn't actually add a deprecation log entry. That will come in a followup.

docs/reference/migration/migrate_8_5.asciidoc
```diff
@@ -20,3 +20,36 @@ coming::[8.5.0]
...
 
+[discrete]
+[[deprecated-8.5]]
+=== Deprecations
+
+The following functionality has been deprecated in {es} 8.3
+and will be removed in a future version.
+While this won't have an immediate impact on your applications,
+we strongly encourage you take the described steps to update your code
+after upgrading to 8.3.
```
tvernum, line 31:
```suggestion
The following functionality has been deprecated in {es} 8.5
and will be removed in a future version.
While this won't have an immediate impact on your applications,
we strongly encourage you take the described steps to update your code
after upgrading to 8.5.
```

tvernum (commented):
I left one comment on the docs.

rjernst:
@tvernum I fixed the version problem.

docs/reference/migration/migrate_8_5.asciidoc
```diff
@@ -20,3 +20,36 @@ coming::[8.5.0]
...
 // end::notable-breaking-changes[]
 
+[discrete]
+[[deprecated-8.5]]
+=== Deprecations
+
+The following functionality has been deprecated in {es} 8.5
+and will be removed in a future version.
+While this won't have an immediate impact on your applications,
+we strongly encourage you take the described steps to update your code
```
ChrisHegarty, line 30: typo? - "we strongly encourage you TO take the described steps to update your code"

docs/reference/migration/migrate_8_5.asciidoc
```diff
@@ -20,3 +20,36 @@ coming::[8.5.0]
...
+*Details* +
+Plugins may override funcionality that controls how nodes connect
+with other nodes over TCP/IP. These plugins extend the NetworkPlugin
+interface. In the next major release, these plugins will fail
+to install.
+
+*Impact* +
+Discontinue using any plugins which extend NetworkPlugin. You can
+see if any plugins use deprecated functionality by checking
+the Elasticsearch deprecation log.
```
ChrisHegarty, line 54: Additionally, and optionally, we could mention that a deprecated warning will be raised when re-compiling the plugin for 8.5 - assuming that the user compiles for 8.5 rather than simply bumping the version in the plugin descriptor.

rjernst: It's difficult to decide who these migration docs should be for, the plugin author, or the user installing such a plugin. They aren't necessarily the same person. Here I've focused on the user, so they know nothing of compiling the plugin.

docs/reference/migration/migrate_8_5.asciidoc
rjernst: Doh, copy/pasta error. Fixed.

rjernst:
@elasticsearchmachine run 
elasticsearch-ci/packaging-tests-unix-sample

qhoxie (approved):

PR #88933, opened 2022-07-29
Add generateStubReleaseNotes task

pugnascotia:
When we feature freeze Elasticsearch, we need to create stub
documentation for the next version. This turns out to be as simple as
running the usual `generateReleaseNotes` task without any inputs.

### DIFF
--- a/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/release/ReleaseToolsPlugin.java
+++ b/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/release/ReleaseToolsPlugin.java
@@ -71,9 +71,11 @@ public class ReleaseToolsPlugin implements Plugin<Project> {
 
         final Function<Boolean, Action<GenerateReleaseNotesTask>> configureGenerateTask = shouldConfigureYamlFiles -> task -> {
             task.setGroup("Documentation");
-            task.setDescription("Generates release notes from changelog files held in this checkout");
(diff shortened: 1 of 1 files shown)


build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/release/ReleaseToolsPlugin.java
```diff
@@ -67,10 +69,12 @@ public void apply(Project project) {
                 task.dependsOn(validateChangelogsAgainstYamlTask);
             });
 
-        project.getTasks().register("generateReleaseNotes", GenerateReleaseNotesTask.class).configure(task -> {
+        final Function<Boolean, Action<GenerateReleaseNotesTask>> configureGenerateTask = shouldConfigureYamlFiles -> task -> {
             task.setGroup("Documentation");
             task.setDescription("Generates release notes from changelog files held in this checkout");
```
breskeby, line 74: I would expect different descriptions for both task making clear what they differ in

breskeby (approved):
lgtm. one little remark

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/release/ReleaseToolsPlugin.java
pugnascotia: Ah, good catch, I've changed that.

PR #59278, opened 2020-07-09
Set specific keepalive options by default on supported platforms

ywelsch:
keepalives tell any intermediate devices that the connection remains alive, which helps with overzealous firewalls that are killing idle connections. keepalives are enabled by default in Elasticsearch, but use system defaults for their configuration, which often times do not have reasonable defaults (e.g. 7200s for TCP_KEEP_IDLE) in the context of distributed systems such as Elasticsearch.

This PR sets the socket-level keep_alive options for `transport.tcp.{keep_idle,keep_interval}` to 5 minutes on configurations that support it (`>= Java 11 & (MacOS || Linux)`) and where the system defaults are set to something higher than 5 minutes. This helps keep the connections alive while not interfering with system defaults or user-specified settings unless they are deemed to be set too high by providing better out-of-the-box defaults.

PR #88935, opened 2022-07-29
Ensure that the extended socket options TCP_KEEPXXX are available

### LINKED ISSUE #88897: Keep alive settings are not applied by default
In 8.3 the default TCP keep alive settings introduced in https://github.com/elastic/elasticsearch/pull/59278 are not applied by default, therefore it uses the OS default values. See:

8.3.0:
```
cat connections-8.3.0.txt  | grep -e '930[01]' | head
ESTAB  0       0         [::ffff:127.0.0.1]:46818     [::ffff:127.0.0.1]:9301    timer:(keepalive,119min,0) uid:1038 ino:6837419 sk:a2 <->
ESTAB  0       0         [::ffff:127.0.0.1]:37928     [::ffff:127.0.0.1]:9300    timer:(keepalive,119min,0) uid:1038 ino:6834869 sk:a3 <->
ESTAB  0       0         [::ffff:127.0.0.1]:9301      [::ffff:127.0.0.1]:46808   timer:(keepalive,119min,0) uid:1038 ino:6830957 sk:a4 <->
ESTAB  0       0         [::ffff:127.0.0.1]:37942     [::ffff:127.0.0.1]:9300    timer:(keepalive,119min,0) uid:1038 ino:6834871 sk:a5 <->
ESTAB  0       0         [::ffff:127.0.0.1]:9301      [::ffff:127.0.0.1]:46828   timer:(keepalive,119min,0) uid:1038 ino:6834882 sk:a6 <->
ESTAB  0       0         [::ffff:127.0.0.1]:37926     [::ffff:127.0.0.1]:9300    timer:(keepalive,119min,0) uid:1038 ino:6834868 sk:a7 <->
ESTAB  0       0         [::ffff:127.0.0.1]:9301      [::ffff:127.0.0.1]:46826   timer:(keepalive,119min,0) uid:1038 ino:6834881 sk:a8 <->
ESTAB  0       0         [::ffff:127.0.0.1]:46824     [::ffff:127.0.0.1]:9301    timer:(keepalive,119min,0) uid:1038 ino:6837425 sk:a9 <->
ESTAB  0       0         [::ffff:127.0.0.1]:37944     [::ffff:127.0.0.1]:9300    timer:(keepalive,119min,0) uid:1038 ino:6834879 sk:aa <->
ESTAB  0       0         [::ffff:127.0.0.1]:46828     [::ffff:127.0.0.1]:9301    timer:(keepalive,119min,0) uid:1038 ino:6837430 sk:ab <->
```

8.2.2:
```
cat connections-8.2.2.txt  | grep -e '930[01]' | head
ESTAB  0       0         [::ffff:127.0.0.1]:46752     [::ffff:127.0.0.1]:9301    timer:(keepalive,4min52sec,0) uid:1038 ino:6833637 sk:6e <->
ESTAB  0       0         [::ffff:127.0.0.1]:9301      [::ffff:127.0.0.1]:46746   timer:(keepalive,4min52sec,0) uid:1038 ino:6831684 sk:6f <->
ESTAB  0       0         [::ffff:127.0.0.1]:46732     [::ffff:127.0.0.1]:9301    timer:(keepalive,4min52sec,0) uid:1038 ino:6833629 sk:70 <->
ESTAB  0       0         [::ffff:127.0.0.1]:37876     [::ffff:127.0.0.1]:9300    timer:(keepalive,4min52sec,0) uid:1038 ino:6833618 sk:71 <->
ESTAB  0       0         [::ffff:127.0.0.1]:9301      [::ffff:127.0.0.1]:46730   timer:(keepalive,4min52sec,0) uid:1038 ino:6831676 sk:72 <->
ESTAB  0       0         [::ffff:127.0.0.1]:46748     [::ffff:127.0.0.1]:9301    timer:(keepalive,4min52sec,0) uid:1038 ino:6833640 sk:73 <->
ESTAB  0       0         [::ffff:127.0.0.1]:46736     [::ffff:127.0.0.1]:9301    timer:(keepalive,4min52sec,0) uid:1038 ino:6833631 sk:74 <->
ESTAB  0       0         [::ffff:127.0.0.1]:9301      [::ffff:127.0.0.1]:46732   timer:(keepalive,4min52sec,0) uid:1038 ino:6831677 sk:75 <->
ESTAB  0       0         [::ffff:127.0.0.1]:37860     [::ffff:127.0.0.1]:9300    timer:(keepalive,4min52sec,0) uid:1038 ino:6833611 sk:76 <->
ESTAB  0       0         [::ffff:127.0.0.1]:46734     [::ffff:127.0.0.1]:9301    timer:(keepalive,4min52sec,0) uid:1038 ino:6833630 sk:77 <->
```

It's possible that after moving `NetUtil` to the netty module https://github.com/elastic/elasticsearch/pull/85953 some of the internal details used are not available anymore.

ChrisHegarty:
A NIO socket channel supports setting and retrieving a set of socket
options. A socket channel supports a specified set of _standard_ socket
options, e.g. `SO_RCVBUF`, as well as optional implementation specific
_extended_ options. The network socket channels in the JDK support,
among others, the `TCP_KEEPIDLE`, `TCP_KEEPINTERVAL`, and
`TCP_KEEPCOUNT` _extended_ socket options (on certain platforms). These
extended socket options are only available when their defining module,
`jdk.net`, is present, and also when defined to the bootstrap class
loader - the socket channel implementation in `java.base` uses
its class loader (the bootstrap class loader) to indirectly initialize
`jdk.net.ExtendedSocketOptions`. Therefore, the `jdk.net` module must be
in the same class loader - the bootstrap class loader - for the extended
socket options to be available/supported **.

The transport-netty4 component sets a number of extended socket options
on its network channels. Specifically, the `TCP_KEEPIDLE`,
`TCP_KEEPINTERVAL`, and `TCP_KEEPCOUNT` socket options. The code
responsible for this uses reflection with a silent fallback if the
options are not available, see `o.e.transport.netty4.NetUtils`. The use
of reflection is because of the desire to compile and run on JDK 8 when
initially added. This is no longer a requirement, since the minimal
runtime version is now JDK 17 and the extended socket options are
available since JDK 11.

With the move to Java modules, only the explicitly required modules or
modules resolved during service binding, are resolved from the JDK. The
`jdk.net` module is not currently resolved, since the extended socket
options are not directly used by any code defined by the system class
loader. Therefore the extended socket options are not available /
supported by the NIO socket channels in the runtime.

The transport-netty4 component, and its dependencies, are defined as
modules by a module layer that is a child of the boot layer. The code in
the transport-netty4 component refers to the socket options and attempts
to set them on a socket channel. While this is all fine and as it should
be, it will not succeed unless the socket channel implementation
supports the extended socket options, which requires the `jdk.net`
module to be resolved in the boot module layer (and loaded by the
bootstrap class loader).

Finally, the solution is to simply resolve the `jdk.net` module in the
boot layer. We do this by adding the `--add-modules=jdk.net` option,
rather than an explicit `requires` directive in the server module-info,
since no code in server uses the extended socket options.  Additionally,
the transport-netty4 module-info has a requires directive added to
ensure that the `jdk.net` module is present. This directive doesn't
ensure that `jdk.net` is in the right module layer, but should be
sufficient for now, and will allow future refactorings in `NetUtils` to
replace the use of reflective access to the socket options with static
access.

Note: I'm still considering if or how best a test could be added to
assert this, but really once NetUtils is refactored that should be 
sufficient.

closes #88897

**
```
$ cat SupportedOptions.java

import java.net.SocketOption;
import java.nio.channels.SocketChannel;
public class SupportedOptions {
    public static void main(String... args) throws Exception {
        System.out.println(SocketChannel.open().supportedOptions().stream().map(SocketOption::toString).sorted().toList());
    }
}
```

```
$ java SupportedOptions
[IP_TOS, SO_KEEPALIVE, SO_LINGER, SO_OOBINLINE, SO_RCVBUF, SO_REUSEADDR, SO_REUSEPORT, SO_SNDBUF, TCP_KEEPCOUNT, TCP_KEEPIDLE, TCP_KEEPINTERVAL, TCP_NODELAY]
```

```
$ java --limit-modules=java.base SupportedOptions
[IP_TOS, SO_KEEPALIVE, SO_LINGER, SO_OOBINLINE, SO_RCVBUF, SO_REUSEADDR, SO_REUSEPORT, SO_SNDBUF, TCP_NODELAY]
```

### DIFF
--- a/distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ServerProcess.java
+++ b/distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ServerProcess.java
@@ -216,6 +216,7 @@ public class ServerProcess {
         command.addAll(jvmOptions);
         command.add("--module-path");
         command.add(esHome.resolve("lib").toString());
+        command.add("--add-modules=jdk.net");
         command.add("-m");
         command.add("org.elasticsearch.server/org.elasticsearch.bootstrap.Elasticsearch");
 
--- a/distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ServerProcess.java
+++ b/distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ServerProcess.java
@@ -216,7 +216,7 @@ public class ServerProcess {
         command.addAll(jvmOptions);
         command.add("--module-path");
         command.add(esHome.resolve("lib").toString());
-        command.add("--add-modules=jdk.net");
+        command.add("--add-modules=jdk.net"); // very special circumstance; explicit modules should typically not be added here
         command.add("-m");
         command.add("org.elasticsearch.server/org.elasticsearch.bootstrap.Elasticsearch");
 
--- a/modules/transport-netty4/src/main/java/org/elasticsearch/transport/netty4/NetUtils.java
+++ b/modules/transport-netty4/src/main/java/org/elasticsearch/transport/netty4/NetUtils.java
@@ -8,12 +8,15 @@
 
 package org.elasticsearch.transport.netty4;
 
+import jdk.net.ExtendedSocketOptions;
+
+import org.elasticsearch.core.SuppressForbidden;
+
 import java.io.IOException;
-import java.lang.reflect.Field;
 import java.net.SocketOption;
 import java.net.StandardSocketOptions;
 import java.nio.channels.NetworkChannel;
-import java.util.Arrays;
+import java.util.Objects;
 
 /**
  * Utilities for network-related methods.
@@ -22,37 +25,31 @@ public class NetUtils {
 
     private NetUtils() {}
 
+    // Accessors to the extended socket options reduce the proliferation of the non-portable
+    // ExtendedSocketOptions type.
+
     /**
-     * Returns the extended TCP_KEEPIDLE socket option, if available on this JDK
+     * Returns the extended TCP_KEEPIDLE socket option.
      */
-    public static SocketOption<Integer> getTcpKeepIdleSocketOptionOrNull() {
-        return getExtendedSocketOptionOrNull("TCP_KEEPIDLE");
+    @SuppressForbidden(reason = "access to non-portable socket option requied")
+    public static SocketOption<Integer> getTcpKeepIdleSocketOption() {
+        return ExtendedSocketOptions.TCP_KEEPIDLE;
     }
 
     /**
-     * Returns the extended TCP_KEEPINTERVAL socket option, if available on this JDK
+     * Returns the extended TCP_KEEPINTERVAL socket option.
      */
-    public static SocketOption<Integer> getTcpKeepIntervalSocketOptionOrNull() {
-        return getExtendedSocketOptionOrNull("TCP_KEEPINTERVAL");
+    @SuppressForbidden(reason = "access to non-portable socket option requied")
+    public static SocketOption<Integer> getTcpKeepIntervalSocketOption() {
+        return ExtendedSocketOptions.TCP_KEEPINTERVAL;
     }
 
     /**
-     * Returns the extended TCP_KEEPCOUNT socket option, if available on this JDK
+     * Returns the extended TCP_KEEPCOUNT socket option.
      */
-    public static SocketOption<Integer> getTcpKeepCountSocketOptionOrNull() {
-        return getExtendedSocketOptionOrNull("TCP_KEEPCOUNT");
-    }
-
-    @SuppressWarnings("unchecked")
-    private static <T> SocketOption<T> getExtendedSocketOptionOrNull(String fieldName) {
-        try {
-            final Class<?> extendedSocketOptionsClass = Class.forName("jdk.net.ExtendedSocketOptions");
-            final Field field = extendedSocketOptionsClass.getField(fieldName);
-            return (SocketOption<T>) field.get(null);
-        } catch (Exception t) {
-            // ignore
-            return null;
-        }
+    @SuppressForbidden(reason = "access to non-portable socket option requied")
+    public static SocketOption<Integer> getTcpKeepCountSocketOption() {
+        return ExtendedSocketOptions.TCP_KEEPCOUNT;
     }
 
     /**
@@ -67,13 +64,9 @@ public class NetUtils {
             if (socketChannel.supportedOptions().contains(StandardSocketOptions.SO_KEEPALIVE)) {
                 final Boolean keepalive = socketChannel.getOption(StandardSocketOptions.SO_KEEPALIVE);
                 assert keepalive != null;
-                if (keepalive.booleanValue()) {
-                    for (SocketOption<Integer> option : Arrays.asList(
-                        NetUtils.getTcpKeepIdleSocketOptionOrNull(),
-                        NetUtils.getTcpKeepIntervalSocketOptionOrNull()
-                    )) {
-                        setMinValueForSocketOption(socketChannel, option, 300);
-                    }
+                if (keepalive) {
+                    setMinValueForSocketOption(socketChannel, getTcpKeepIdleSocketOption(), 300);
+                    setMinValueForSocketOption(socketChannel, getTcpKeepIntervalSocketOption(), 300);
                 }
             }
         } catch (Exception e) {
@@ -84,7 +77,8 @@ public class NetUtils {
     }
 
     private static void setMinValueForSocketOption(NetworkChannel socketChannel, SocketOption<Integer> option, int minValue) {
-        if (option != null && socketChannel.supportedOptions().contains(option)) {
+        Objects.requireNonNull(option);
+        if (socketChannel.supportedOptions().contains(option)) {
             try {
                 final Integer currentIdleVal = socketChannel.getOption(option);
                 assert currentIdleVal != null;
--- a/modules/transport-netty4/src/main/java/org/elasticsearch/transport/netty4/NetUtils.java
+++ b/modules/transport-netty4/src/main/java/org/elasticsearch/transport/netty4/NetUtils.java
@@ -31,7 +31,7 @@ public class NetUtils {
     /**
      * Returns the extended TCP_KEEPIDLE socket option.
      */
-    @SuppressForbidden(reason = "access to non-portable socket option requied")
+    @SuppressForbidden(reason = "access to non-portable socket option required")
     public static SocketOption<Integer> getTcpKeepIdleSocketOption() {
         return ExtendedSocketOptions.TCP_KEEPIDLE;
     }
@@ -39,7 +39,7 @@ public class NetUtils {
     /**
      * Returns the extended TCP_KEEPINTERVAL socket option.
      */
-    @SuppressForbidden(reason = "access to non-portable socket option requied")
+    @SuppressForbidden(reason = "access to non-portable socket option required")
     public static SocketOption<Integer> getTcpKeepIntervalSocketOption() {
         return ExtendedSocketOptions.TCP_KEEPINTERVAL;
     }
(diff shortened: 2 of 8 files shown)


DaveCTurner:
Can we drop the use of reflection in `NetUtils` at the same time?

ChrisHegarty:
> Can we drop the use of reflection in `NetUtils` at the same time?

Yes. I wanted to keep the changes as small as possible, but let me see how the refactoring looks.

henningandersen:
> Can we drop the use of reflection in NetUtils at the same time?

I'd also prefer that. At a minimum we should change `getExtendedSocketOptionOrNull` to do:

```
            return (SocketOption<T>) jdk.net.ExtendedSocketOptions.class.getField(fieldName).get(null);
```

instead, since this was where the error originated. But we might as well go all in here. Happy to help out on this.

ChrisHegarty:
The use of reflection has been removed. There is a little indirection still, through accessors to the extended socket options, so as to narrow the suppression of the forbidden non-portable ExtendedSocketOptions type.

modules/transport-netty4/src/main/java/org/elasticsearch/transport/netty4/NetUtils.java
```diff
@@ -22,37 +25,31 @@ public class NetUtils {
...
+    // Accessors to the extended socket options reduce the proliferation of the non-portable
+    // ExtendedSocketOptions type.
+
     /**
-     * Returns the extended TCP_KEEPIDLE socket option, if available on this JDK
+     * Returns the extended TCP_KEEPIDLE socket option.
      */
-    public static SocketOption<Integer> getTcpKeepIdleSocketOptionOrNull() {
-        return getExtendedSocketOptionOrNull("TCP_KEEPIDLE");
+    @SuppressForbidden(reason = "access to non-portable socket option requied")
```
henningandersen: nit:

```suggestion
    @SuppressForbidden(reason = "access to non-portable socket option required")
```

(in all the `SuppressForbidden` lines)

modules/transport-netty4/src/test/java/org/elasticsearch/transport/netty4/NetUtilsTests.java
```diff
@@ -12,13 +12,18 @@
 import org.elasticsearch.core.IOUtils;
 import org.elasticsearch.test.ESTestCase;
 
+import java.nio.channels.SocketChannel;
+
 public class NetUtilsTests extends ESTestCase {
 
-    public void testExtendedSocketOptions() {
+    public void testExtendedSocketOptions() throws Exception {
         assumeTrue("JDK possibly not supported", Constants.JVM_NAME.contains("HotSpot") || Constants.JVM_NAME.contains("OpenJDK"));
```
henningandersen, line 20: Should we/can we remove this? As per your PR description, it should be supported on all JDKs?

henningandersen (approved):
LGTM, but I'd like David or Ryan to also review.

modules/transport-netty4/src/test/java/org/elasticsearch/transport/netty4/NetUtilsTests.java
```diff
@@ -12,13 +12,18 @@
...
         assumeTrue("Platform possibly not supported", IOUtils.LINUX || IOUtils.MAC_OS_X);
```
DaveCTurner, line 21: Likewise this one - if these options are not supported on Windows then we will need to do something more involved.

henningandersen: I think that one is OK though we could check that the options are not null on windows too. But IIUC, it is OK that they are not in `supportedOptions()` on windows, the original PR #59278 seemed to assume that windows does not have these options in the JDK.

DaveCTurner: Not really familiar with how this kind of unportable thing works in Java - it looks to me that we would try and access fields like `ExtendedSocketOptions.TCP_KEEPIDLE` whether this is claimed to be supported or not, and I expect that would be reasonably fatal if this field doesn't exist.

modules/transport-netty4/src/main/java/org/elasticsearch/transport/netty4/NetUtils.java
ChrisHegarty: Done.

modules/transport-netty4/src/test/java/org/elasticsearch/transport/netty4/NetUtilsTests.java
ChrisHegarty: The `ExtendedSocketOptions` type, and therefore its fields, are present in all OpenJDK distributions, once the `jdk.net` module is resolved. The way that platform optionality is supported is through the set of socket options returned by the `NetworkChannel::supportedOptions`. For example, on Windows set of options return by `SocketChannel::supportedOptions` will not contain `TCP_KEEPIDLE`.

ChrisHegarty: I reworked the assumptions in testExtendedSocketOptions, so that various aspects of the test only run when the assumption is valid ( invalid assumptions cause the test scenario to be skipped )

ChrisHegarty: Calling this out explicitly, since I now realise that it has been implicit so far.   

With the rework to remove reflection, this code now has an explicit dependency on a JDK-specific feature, namely the extended socket options provided by the `jdk.net` module. I don't think that this is a problem, and there are few (if any) distributions of the JDK these days that are not downstream distros of OpenJDK.

modules/transport-netty4/src/test/java/org/elasticsearch/transport/netty4/NetUtilsTests.java
ChrisHegarty: Yes, good point. Removed. It's a distro derived from OpenJDK or nothing!

rjernst (approved):
LGTM

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ServerProcess.java
```diff
@@ -216,6 +216,7 @@ private static Process createProcess(
         command.addAll(jvmOptions);
         command.add("--module-path");
         command.add(esHome.resolve("lib").toString());
+        command.add("--add-modules=jdk.net");
```
rjernst, line 219: maybe add a comment here that this is a very special circumstance, more explicit modules should not be added here

modules/transport-netty4/src/test/java/org/elasticsearch/transport/netty4/NetUtilsTests.java
DaveCTurner: Gotcha, thanks, this makes sense now.

DaveCTurner (approved):
LGTM3

ChrisHegarty:
Thanks for the reviews.   I'm going to merge only after a full clean CI run on all platforms. And will then proceed to backport to 8.4, and 8.3.x.

distribution/tools/server-cli/src/main/java/org/elasticsearch/server/cli/ServerProcess.java
ChrisHegarty: Done.

ChrisHegarty: Done.

ChrisHegarty:

ChrisHegarty:
@elasticmachine run elasticsearch-ci/part-1-windows

ChrisHegarty:

PR #88937, opened 2022-07-29
Wrap async search action logic in a new trace context

pugnascotia:
Part of #84369. Split out from #88443. This PR wraps parts logic in
`TransportSubmitAsyncSearchAction` in a new tracing context. This is
necessary so that a tracing implementation can use the thread context
to propagate tracing headers, but without the code attempting to set the
same key twice in the thread context, which is illegal.


pugnascotia:
Packaging test failure is unrelated.

romseygeek (approved):
LGTM

x-pack/plugin/async-search/src/main/java/org/elasticsearch/xpack/search/TransportSubmitAsyncSearchAction.java
```diff
@@ -83,65 +83,67 @@ public TransportSubmitAsyncSearchAction(
...
-                            public void onResponse(IndexResponse r) {
-                                if (searchResponse.isRunning()) {
-                                    try {
-                                        // store the final response on completion unless the submit is cancelled
-                                        searchTask.addCompletionListener(
-                                            finalResponse -> onFinalResponse(searchTask, finalResponse, () -> {})
-                                        );
-                                    } finally {
-                                        submitListener.onResponse(searchResponse);
+        try (var ignored = threadContext.newTraceContext()) {
```
javanna, line 86: Is this needed because we are registering a new task in the next line, which will stay alive once the submit action has returned a response? Is there a way to possibly do this automatically when needed? I can see how this is going to be forgotten in other places or possibly there are already places where the same change is needed which we are not aware of yet?

pugnascotia: Almost all uses of task registration happen via `registerAndExecute`, which handles the thread content stuff automatically. I've checked all the other places which manage tasks without `registerAndExecute` to make sure they're doing the right thing.

I haven't yet come up with an alternative approach unfortunately. Detecting these cases is straightforward at least, because if you don't start a new context, you get an exception!

PR #88944, opened 2022-07-29
Refactor authentication handling for grant actions

### LINKED ISSUE #88636: Refactor authentication handling for grant actions
### Description

`TransportGrantAction` currently provides an `executeWithGrantAuthentication` method which sub-classes can leverage. Instead of requiring sub-classes to invoke this method directly, we instead can move it into the `doExecute` method of `TransportGrantAction`, and allow inheritors to specify their own handler for grant authentication, something along the lines of: 

```
class TransportGrantAction {

    final void doExecute(...) {
        // executeWithGrantAuthentication can furthermore be inlined as part of this refactor
        executeWithGrantAuthentication(request, ActionListener.wrap(
            authentication -> { doSomethingWithAuthentication(...) }
        ));
    }

    abstract doSomethingWithAuthentication(...);
}
```

This reduces boilerplate and establishes a clearer contract for sub-classes of `TransportGrantAction`.

n1v0lg:
This PR establishes a cleaner contract between `TransportGrantAction`
and its sub-classes: it instruments checking grant authentication
instead of requiring this of sub-classes, and allows these to override
a handler method for successfully granted authentication. 

Closes #88636


n1v0lg:
@elasticmachine update branch

n1v0lg:
@elasticmachine update branch

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/TransportGrantAction.java
```diff
@@ -49,18 +50,19 @@ public TransportGrantAction(
         this.threadContext = threadContext;
     }
 
-    protected void executeWithGrantAuthentication(GrantRequest grantRequest, ActionListener<Authentication> listener) {
+    @Override
+    public void doExecute(Task task, Request request, ActionListener<Response> listener) {
```
ywangd, line 54: We should make this method `final`.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/TransportGrantAction.java
```diff
@@ -73,12 +75,15 @@ protected void executeWithGrantAuthentication(GrantRequest grantRequest, ActionL
...
                         // This can happen when the authentication itself is a run-as client-credentials token.
-                        assert runAsUsername != null || "access_token".equals(grantRequest.getGrant().getType());
+                        assert runAsUsername != null || "access_token".equals(request.getGrant().getType());
                         authorizationService.authorize(
                             authentication,
                             AuthenticateAction.NAME,
                             new AuthenticateRequest(effectiveUsername),
-                            ActionListener.wrap(ignore2 -> listener.onResponse(authentication), listener::onFailure)
+                            ActionListener.wrap(
+                                ignore2 -> doExecuteOnGrantAuthenticationSuccess(task, request, authentication, listener),
```
ywangd, line 84: Nit: I slightly prefer `doExecuteWithGrantAuthentication` because it also does a bit authorization here so it is not just "AuthenticationSuccess".

ywangd (approved):
LGTM

Thanks for picking this up.

PR #88945, opened 2022-07-29
[ML] address potential bug where trained models get stuck in starting after being allocated to node

benwtrent:
When a model is starting, it has been rarely observed that it will lock up while trying to restore the model objects to the native process. 

This would manifest as a trained model being stuck in "starting" while also being assigned to a node. So, there is a native process started and task available on the assigned nodes, but the model state never gets out of "starting".


benwtrent:
@elasticmachine update branch

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/inference/deployment/DeploymentManager.java
```diff
@@ -149,34 +150,46 @@ private void doStartDeployment(TrainedModelDeploymentTask task, ActionListener<T
...
-                // `startAndLoad` creates named pipes, blocking the calling thread, better to execute that in our utility
-                // executor.
-                executorServiceForDeployment.execute(() -> startAndLoad(processContext, modelConfig.getLocation(), modelLoadedListener));
-            }, listener::onFailure));
+                }, listener::onFailure));
+            } else {
+                listener.onFailure(
+                    new IllegalArgumentException(
+                        format(
+                            "[%s] must be an pytorch model found inference config of kind [%s]",
```
dimitris-athanasiou, line 186:
```suggestion
                            "[%s] must be a pytorch model; found inference config of kind [%s]",
```

dimitris-athanasiou (commented):
LGTM Just a suggestion for an error message.

benwtrent:
@elasticmachine update branch

dimitris-athanasiou (approved):

PR #88946, opened 2022-07-29
Graph: fix race condition in timeout

### LINKED ISSUE #55396: [CI] GraphTests.testTimedoutQueryCrawl fails on CI
The test GraphTests.testTimedoutQueryCrawl failed on CI today for the `7.7` branch. The error seems to be a negative `TimeValue` that is wrongly serialized and this is now better caught thanks to #53913:

```
Caused by: java.lang.IllegalArgumentException: duration cannot be negative, was given [-153]
```

It does not reproduce locally with
```
./gradlew ':x-pack:plugin:graph:test' --tests "org.elasticsearch.xpack.graph.test.GraphTests.testTimedoutQueryCrawl" -Dtests.seed=29C73B1311199D93 -Dtests.security.manager=true -Dtests.locale=ca-ES -Dtests.timezone=Asia/Omsk -Dcompiler.java=14
```

Build scan: https://gradle-enterprise.elastic.co/s/5e5mmdotpx6j4 

Other failures seen in build stats:
https://gradle-enterprise.elastic.co/s/v54asb7cqpazs (master)
https://gradle-enterprise.elastic.co/s/gqftqq3eiezcw (7.7)

nik9000:
Previously `graph` checked if the request timed out, then spent some
time doing work, then passed the timeout on to the next request. Over
and over again. It's quite possible that the response may not have timed
out for the first check but would have timed out for the second check.
This manifests as the timeout being sent to the next hop being a
negative number of milliseconds. We don't allow this sort of thing.

This fixes this by moving the timeout check to the same spot it is read
for setting the timeout on the next request - we just check if its `> 0`
to find the timeouts.

This does keep the request running slightly longer after it's officially
timed out - but it's just long enough to prepare the next layer of
request. Usually microseconds. Which should be fine.

Closes #55396

### DIFF
--- a/x-pack/plugin/graph/src/main/java/org/elasticsearch/xpack/graph/action/TransportGraphExploreAction.java
+++ b/x-pack/plugin/graph/src/main/java/org/elasticsearch/xpack/graph/action/TransportGraphExploreAction.java
@@ -6,6 +6,8 @@
  */
 package org.elasticsearch.xpack.graph.action;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.util.BytesRef;
 import org.apache.lucene.util.PriorityQueue;
@@ -61,13 +63,13 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Performs a series of elasticsearch queries and aggregations to explore
  * connected terms in a single index.
  */
 public class TransportGraphExploreAction extends HandledTransportAction<GraphExploreRequest, GraphExploreResponse> {
+    private static final Logger logger = LogManager.getLogger(TransportGraphExploreAction.class);
 
     private final ThreadPool threadPool;
     private final NodeClient client;
@@ -115,7 +117,6 @@ public class TransportGraphExploreAction extends HandledTransportAction<GraphExp
         private final ActionListener<GraphExploreResponse> listener;
 
         private final long startTime;
-        private final AtomicBoolean timedOut;
         private volatile ShardOperationFailedException[] shardFailures;
         private Map<VertexId, Vertex> vertices = new HashMap<>();
         private Map<ConnectionId, Connection> connections = new HashMap<>();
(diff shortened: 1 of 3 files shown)


x-pack/plugin/graph/src/main/java/org/elasticsearch/xpack/graph/action/TransportGraphExploreAction.java
```diff
@@ -173,16 +173,11 @@ private void removeVertex(Vertex vertex) {
...
-                return;
-            }
             Map<String, Set<Vertex>> lastHopFindings = hopFindings.get(currentHopNumber);
             if ((currentHopNumber >= (request.getHopNumbers() - 1)) || (lastHopFindings == null) || (lastHopFindings.size() == 0)) {
                 // Either we gathered no leads from the last hop or we have
                 // reached the final hop
-                listener.onResponse(buildResponse());
+                listener.onResponse(buildResponse(false));
                 return;
             }
```
nik9000, line 182: Here's the bit where we return the response even if we're over time. If we're *done* we may as well return anyway, I say.

x-pack/plugin/graph/src/main/java/org/elasticsearch/xpack/graph/action/TransportGraphExploreAction.java
```diff
@@ -318,16 +313,22 @@ synchronized void expand() {
...
+                source.timeout(TimeValue.timeValueMillis(timeRemainingMillis));
             }
             searchRequest.source(source);
 
-            // System.out.println(source);
             logger.trace("executing expansion graph search request");
             client.search(searchRequest, new ActionListener.Delegating<>(listener) {
                 @Override
                 public void onResponse(SearchResponse searchResponse) {
-                    // System.out.println(searchResponse);
```
nik9000, line 330: These just seemed like leftovers. I figure we'd dump these into the `trace` logs if we want them.

nik9000:

x-pack/plugin/graph/src/main/java/org/elasticsearch/xpack/graph/action/TransportGraphExploreAction.java
iverase: This is fine for me

iverase (approved):
LGTM

PR #88950, opened 2022-07-29
Bring back lost optimization to building metadata from a diff

original-brownbear:
This was lost again due to another merge conflict.

Same as https://github.com/elastic/elasticsearch/pull/88497 :)


original-brownbear:
merging without review since this was an obvious merge error

Issue #87776, opened 2022-06-17
Pre-validate removal of nodes


henningandersen:
### Description

When a cluster has RED health, a shard is not assigned to a node containing an in-sync copy of the data. If a node is to be removed from the cluster, there is a risk that the last remaining copy of such a shard is on that node. For that reason, orchestration systems typically do not proceed with changes involving node removals to the cluster while it is in RED state.

We propose to add an explicit API `_internal/pre_validate_node_remove` that should be used instead. It can be passed a list of node names/external-ids to remove from the cluster and will give a response per node on whether it is safe as well as a an overall response on the safety of removing all the nodes mentioned.

This check should at least disregard searchable snapshot shards when determining whether it is safe to remove a node. It could also try to reach out to the node to see if the node contains a copy of a RED shard on disk. If it can successfully conclude that it has no copies of RED shards on disk, it can conclude that it is safe to remove the node.

We can also check if any of the nodes not listed in the API has an in-sync copy of such shards. If a node to remove contains a copy but another node contains an in-sync copy for a RED shard, we should indicate it specifically in the response.

Such an API will allow orchestrators to continue with changes to the cluster in many more situations, requiring less manual intervention.

PR #88952, opened 2022-07-29
Prevalidate node removal API (pt. 1)

pxsalehi:
This PR adds the first part of the Prevalidate Node Removal API. This
API allows checking  whether attempting to remove some node(s) from the
cluster is likely to succeed or not. This check is useful when a node
needs to be removed from a RED cluster, without risking loosing the last
copy of some RED shards.

In this PR, we only check whether a RED index is a Searchable Snapshot
index or not, in which case the removal of any node is safe as the RED
index is backed by a snapshot.

Relates https://github.com/elastic/elasticsearch/issues/87776


pxsalehi:

pxsalehi:
API doc: https://elasticsearch_88952.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/prevalidate-node-removal-api.html

pxsalehi:
@henningandersen Please have a look at this draft PR. I have tried to find out all the places where the new API would "touch", and add a basic version of what is needed. As far as the prevalidation logic goes, it is for now as basic as: if red cluster then it might not be safe to remove any node, and otherwise any node is safe to remove (essentially what is currently implied by the cluster health status).

Please check if the naming/location of the API related files/doc, the doc itself, and whether there is anything missing from the structure/scaffolding of adding a new API. I'll try to add more logic/validation/test to the actual API implementation, and refine the doc.

server/src/internalClusterTest/java/org/elasticsearch/cluster/PrevalidateNodeRemovalIT.java
```diff
@@ -0,0 +1,34 @@
...
+        createIndex("test");
+        ensureGreen();
+        String nodeName = randomFrom(internalCluster().getNodeNames());
+        PrevalidateNodeRemovalRequest req = new PrevalidateNodeRemovalRequest(nodeName);
+        PrevalidateNodeRemovalResponse resp = client().execute(PrevalidateNodeRemovalAction.INSTANCE, req).get();
+
+        assertThat(resp.getPrevalidation().getOverallResult().isSafe(), equalTo(NodesRemovalPrevalidation.IsSafe.YES));
+        assertThat(resp.getPrevalidation().getPerNodeResult().size(), equalTo(1));
+        assertThat(resp.getPrevalidation().getPerNodeResult().containsKey(nodeName), equalTo(true));
+        assertThat(resp.getPrevalidation().getPerNodeResult().get(nodeName).isSafe(), equalTo(NodesRemovalPrevalidation.IsSafe.YES));
```
idegtiarenko, line 32: NIT: is it possible to replace it with a single assertion? This way in case of failure we would be able to see the entire map content without having to reproduce it.

```
assertThat(resp.getPrevalidation().getPerNodeResult(), equalTo(Map.of(nodeName, NodesRemovalPrevalidation.IsSafe.YES)));
```

I was hoping there is a hamcrest matches for it, but it looks like `hasEntry` permits other entries in map as well.

idegtiarenko: Oh, looks like the entry will contain a reason as well. Was it not asserted intentionally?

pxsalehi: Yes. For now those details are not clear. I just added that IT to make sure everything works together. There will be more tests and detailed assertions, ofc.

docs/reference/cluster/prevalidate-node-removal.asciidoc
```diff
@@ -0,0 +1,78 @@
+[[prevalidate-node-removal-api]]
+=== Prevalidate node removal API
+++++
+<titleabbrev>Prevalidate node removal</titleabbrev>
+++++
+
+This API is currently experimental for internal use by Elastic software only.
```
henningandersen, line 7: Not sure if this is meant to be removed before merge. We normally do the `{cloud-only}` warning only.

docs/reference/cluster/prevalidate-node-removal.asciidoc
```diff
@@ -0,0 +1,78 @@
...
+
+The response includes the overall safety of the removal of the specified nodes, and a detailed response for each node. The node-specific part of the response also includes more details on why removal of that node might not be safe.
+
+The result of the prevalidation is expressed as either `YES`, `NO`, or `UNKNOWN`.
+
+[[prevalidate-node-removal-api-path-params]]
+==== {api-path-parms-title}
+
+`<node_id>`::
+    (Required, string) Comma-separated list of node IDs.
```
henningandersen, line 36: I think we can support node-ids too, but we should support node name and node external-id (which is sort of the same).

This avoids orchestration systems having to lookup the node-id first. The lookup of node-id would need to also by-pass circuit breakers. And limiting the number of calls also increases the chance of success.

With two identification forms it might be easier to switch to a body for the input?

docs/reference/cluster/prevalidate-node-removal.asciidoc
```diff
@@ -0,0 +1,78 @@
...
+
+[[prevalidate-node-removal-api-response-body]]
+==== {api-response-body-title}
+
+`overall_result`::
+    (object) Prevalidation result for the removal of all the provided nodes.
+
+`per_node_result`::
+    (object) Prevalidation result for the removal of each of the provided nodes.
```
henningandersen, line 45: The naming here looks different from existing naming. Maybe you can look for a few examples to lean on?

I'd tentatively think that the outer level result needs no `overall` name, but could still be an object `result`.

`per_node_result` could be just `nodes` with a list of nodes containing (id, name, result) for each.

docs/reference/cluster/prevalidate-node-removal.asciidoc
```diff
@@ -0,0 +1,78 @@
...
+
+[[prevalidate-node-removal-api-example]]
+==== {api-examples-title}
+
+This example validates whether it is safe to remove the nodes `node1` and `node2`. The response indicates that it is safe to remove `node`, but it might not be safe to remove `node2`. Therefore, the overall prevalidation of the removal of the two nodes is `UNKNOWN`.
```
henningandersen, line 50:
```suggestion
This example validates whether it is safe to remove the nodes `node1` and `node2`. The response indicates that it is safe to remove `node1`, but it might not be safe to remove `node2`. Therefore, the overall prevalidation of the removal of the two nodes is `UNKNOWN`.
```

rest-api-spec/src/main/resources/rest-api-spec/api/_internal.prevalidate_node_removal.json
```diff
@@ -0,0 +1,29 @@
+{
+  "_internal.prevalidate_node_removal": {
+    "documentation":{
+      "url": null,
```
henningandersen, line 4: We should fill in the URL here.

rest-api-spec/src/main/resources/rest-api-spec/api/_internal.prevalidate_node_removal.json
```diff
@@ -0,0 +1,29 @@
...
+    "visibility":"private",
+    "headers":{
+      "accept": [ "application/json"]
+    },
+    "url":{
+      "paths":[
+        {
+          "path":"/_internal/prevalidate_node_removal/{node}",
+          "methods":[
+            "GET"
```
henningandersen, line 17: I think POST is slightly better here, since it mostly runs a function (prevalidate), rather than GETs state. We could support both at least.

server/src/main/java/org/elasticsearch/action/admin/cluster/node/remove/PrevalidateNodeRemovalAction.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package org.elasticsearch.action.admin.cluster.node.remove;
+
+import org.elasticsearch.action.ActionType;
+
+public class PrevalidateNodeRemovalAction extends ActionType<PrevalidateNodeRemovalResponse> {
+
+    public static final PrevalidateNodeRemovalAction INSTANCE = new PrevalidateNodeRemovalAction();
+    public static final String NAME = "cluster:admin/nodes/prevalidate_removal";
```
henningandersen, line 16: I am tempted to use:

```suggestion
    public static final String NAME = "cluster:admin/shutdown/prevalidate_removal";
```

This seems more shutdown related than "nodes" related.

server/src/main/java/org/elasticsearch/action/admin/cluster/node/remove/PrevalidateNodeRemovalResponse.java
```diff
@@ -0,0 +1,45 @@
...
+import org.elasticsearch.xcontent.ToXContentObject;
+import org.elasticsearch.xcontent.XContentBuilder;
+
+import java.io.IOException;
+
+public class PrevalidateNodeRemovalResponse extends ActionResponse implements ToXContentObject {
+    private final NodesRemovalPrevalidation prevalidation;
+
+    public PrevalidateNodeRemovalResponse(StreamInput in) {
+        throw new AssertionError("PrevalidateNodeRemovalAction should not be sent over the wire.");
```
henningandersen, line 23: Maybe this is just because it is draft? I see no reason this response could not be sent over the wire.

server/src/main/java/org/elasticsearch/action/admin/cluster/node/remove/TransportPrevalidateNodeRemovalAction.java
```diff
@@ -0,0 +1,82 @@
...
+import org.elasticsearch.transport.TransportService;
+
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import static org.elasticsearch.action.admin.cluster.node.remove.NodesRemovalPrevalidation.IsSafe;
+import static org.elasticsearch.action.admin.cluster.node.remove.NodesRemovalPrevalidation.Result;
+
+// TODO: should this instead extend TransportMasterNodeReadAction?
```
henningandersen, line 30: Yes.

rest-api-spec/src/main/resources/rest-api-spec/api/_internal.prevalidate_node_removal.json
```diff
@@ -0,0 +1,29 @@
...
+          ],
+          "parts":{
```
henningandersen, line 19: Since it is a master action, there are some timeouts to add here.

server/src/main/java/org/elasticsearch/action/admin/cluster/node/remove/TransportPrevalidateNodeRemovalAction.java
```diff
@@ -0,0 +1,82 @@
...
+    @Inject
+    public TransportPrevalidateNodeRemovalAction(ActionFilters actionFilters, TransportService transportService, NodeClient client) {
+        super(PrevalidateNodeRemovalAction.NAME, actionFilters, transportService.getTaskManager());
+        this.client = client;
+    }
+
+    @Override
+    protected void doExecute(Task task, PrevalidateNodeRemovalRequest request, ActionListener<PrevalidateNodeRemovalResponse> listener) {
+        // TODO: Need to set masterNodeTimeOut?
+        client.admin().cluster().health(new ClusterHealthRequest(), new ActionListener<>() {
```
henningandersen, line 46: `new ClusterStateHealth(clusterService.state()).getStatus()` is preferable here, avoids the async, avoids another action needing to have circuit breaking disabled and ensures we can use the same `ClusterState` for a full prevalidation as we add more features.

server/src/main/java/org/elasticsearch/action/admin/cluster/node/remove/TransportPrevalidateNodeRemovalAction.java
```diff
@@ -0,0 +1,82 @@
...
+// TODO: should this instead extend TransportMasterNodeReadAction?
+public class TransportPrevalidateNodeRemovalAction extends TransportAction<PrevalidateNodeRemovalRequest, PrevalidateNodeRemovalResponse> {
+
+    private static final Logger logger = LogManager.getLogger(TransportPrevalidateNodeRemovalAction.class);
+
+    private final NodeClient client;
+
+    @Inject
+    public TransportPrevalidateNodeRemovalAction(ActionFilters actionFilters, TransportService transportService, NodeClient client) {
+        super(PrevalidateNodeRemovalAction.NAME, actionFilters, transportService.getTaskManager());
```
henningandersen, line 39: We should utilize `canTripCircuitBreaker` set to false here (easiest to do after you switch to `TransportMasterNodeReadAction`.)

x-pack/plugin/security/qa/operator-privileges-tests/src/javaRestTest/java/org/elasticsearch/xpack/security/operator/Constants.java
```diff
@@ -43,6 +43,7 @@ public class Constants {
         "cluster:admin/migration/get_system_feature",
         "cluster:admin/migration/post_system_feature",
         "cluster:admin/nodes/reload_secure_settings",
+        "cluster:admin/nodes/prevalidate_removal",
```
henningandersen, line 46: This should be an operator only action instead, i.e., not added here.

henningandersen (commented):
Did a quick walk through of the code and added initial comments.

server/src/main/java/org/elasticsearch/action/admin/cluster/node/shutdown/TransportPrevalidateNodeRemovalAction.java
```diff
@@ -0,0 +1,172 @@
...
+                throw new IllegalArgumentException("more than one node matches [" + node + "]");
+            }
+            concreteNodes.add(matches.get(0));
+        }
+        if (concreteNodes.size() != nodes.size()) {
+            throw new IllegalArgumentException(
+                Strings.format("provided {} values for <nodes> which resolved to {} nodes", nodes.size(), concreteNodes.size())
+            );
+        }
+        ;
```
pxsalehi, line 106: I have tried to add some validations here to make sure prevalidation calls do not accidentally lead to a false safe removal response. By that I mean, since we do provide an overall result, if the provided nodes (name or id) do not each uniquely identify a node, I do not consider the request valid. I am assuming it is not safe to dismiss an id (perhaps with a typo) if there is no matching node for it. I am not sure whether this is strictly necessary, so I'd appreciate a second opinion.

pxsalehi (commented):
@henningandersen I've addressed your comments. Please have another quick look. I will cleanup a bit more and add more tests, but for now I have some questions about the API.

About the scope of the PR, do we need to address the second phase (reaching out to the nodes to see if they have a copy of an unassigned shard) in this PR? If possible I'd rather do that as a follow up, as just addressing the searchable snapshot indices for now seems to be also useful.

docs/reference/cluster/prevalidate-node-removal.asciidoc
```diff
@@ -0,0 +1,91 @@
...
+    (object) Prevalidation result for the removal of each of the provided nodes.
+
+[[prevalidate-node-removal-api-example]]
+==== {api-examples-title}
+
+This example validates whether it is safe to remove the nodes `node1` and `node2`. The response indicates that it is safe to remove `node1`, but it might not be safe to remove `node2`. Therefore, the overall prevalidation of the removal of the two nodes is `UNKNOWN`.
+
+[source,console]
+--------------------------------------------------
+POST /_internal/prevalidate_node_removal/node1,node2
```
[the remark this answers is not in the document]
henningandersen: I'd be happy with just POST.

docs/reference/cluster/prevalidate-node-removal.asciidoc
```diff
@@ -0,0 +1,91 @@
...
+
+The response includes the overall safety of the removal of the specified nodes, and a detailed response for each node. The node-specific part of the response also includes more details on why removal of that node might not be safe.
+
+The result of the prevalidation is expressed as either `YES`, `NO`, or `UNKNOWN`.
+
+[[prevalidate-node-removal-api-path-params]]
+==== {api-path-parms-title}
+
+`<node>`::
+    (Required, string) Comma-separated list of node IDs, names or external IDs.
```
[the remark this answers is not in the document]
henningandersen: I did not look at the code yet, but it would be nice if we know what field we expect here to disambiguate.

I think we should not support IP etc., those 3 forms you mention are all we need.

server/src/main/java/org/elasticsearch/action/admin/cluster/node/shutdown/TransportPrevalidateNodeRemovalAction.java
henningandersen: I think the ambiguous case (more than one node matches) is fine to throw on. I think I'd like to provide a response also if we are missing a node - but the response should indicate that it was missing. That way we can leave it up to the orchestration to make a decision - that might depend on the situation (vacate/system vs user initiated change).

henningandersen:
> About the scope of the PR, do we need to address the second phase (reaching out to the nodes to see if they have a copy of an unassigned shard) in this PR? If possible I'd rather do that as a follow up, as just addressing the searchable snapshot indices for now seems to be also useful.

I agree that we can omit that in this PR and do that as a follow-up as long as it does not result in false positives (i.e., us saying it is ok to shutdown the node when it is really not).

server/src/main/java/org/elasticsearch/action/admin/cluster/node/shutdown/PrevalidateNodeRemovalRequest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.elasticsearch.cluster.node.DiscoveryNode;
+import org.elasticsearch.common.io.stream.StreamInput;
+import org.elasticsearch.common.io.stream.StreamOutput;
+
+import java.io.IOException;
+import java.util.List;
+
+public class PrevalidateNodeRemovalRequest extends MasterNodeReadRequest<PrevalidateNodeRemovalRequest> {
+
+    private final String[] nodeIds; // Maybe named just nodes?
```
henningandersen, line 22: I'd like to see this split into ids, names and externalIds. Or a flag for the one type we expect (and reject if multiple types are used).

server/src/main/java/org/elasticsearch/action/admin/cluster/node/shutdown/PrevalidateNodeRemovalRequest.java
```diff
@@ -0,0 +1,60 @@
...
+        this.nodeIds = nodeIds;
+    }
+
+    public PrevalidateNodeRemovalRequest(final StreamInput in) throws IOException {
+        super(in);
+        nodeIds = in.readStringArray();
+    }
+
+    public void setConcreteNodes(DiscoveryNode... discoveryNodes) {
+        concreteNodes = discoveryNodes;
```
henningandersen, line 35: I wonder why we set this into the request, can we not keep that in a local state instead? I prefer the request to be purely the request and also that writeTo and reading results in the same object.

rest-api-spec/src/main/resources/rest-api-spec/api/_internal.prevalidate_node_removal.json
```diff
@@ -0,0 +1,35 @@
+{
+  "_internal.prevalidate_node_removal": {
+    "documentation":{
+      "url": "https://www.elastic.co/guide/en/elasticsearch/reference/master/prevalidate-node-removal-api.html",
+      "description":"Prevalidates node removal from the cluster"
+    },
+    "stability":"experimental",
```
henningandersen, line 7: I think we should call this stable before release.

server/src/test/java/org/elasticsearch/action/admin/cluster/node/shutdown/PrevalidateNodeRemovalResponseTests.java
```diff
@@ -0,0 +1,108 @@
...
+                        "is_safe" : "UNKNOWN",
+                        "reason" : "node hosts a red shard copy"
+                      }
+                    }
+                  ]
+                }""", Strings.toString(builder));
+        }
+    }
+
+    public void testSerialization() throws IOException {
```
henningandersen, line 78: Can we use `AbstractWireSerializingTestCase`?

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/PrevalidateNodeRemovalWithSearchableSnapshotTests.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.util.List;
+import java.util.Locale;
+
+import static org.elasticsearch.index.IndexSettings.INDEX_SOFT_DELETES_SETTING;
+import static org.elasticsearch.test.hamcrest.ElasticsearchAssertions.assertAcked;
+import static org.hamcrest.Matchers.equalTo;
+
+@ESIntegTestCase.ClusterScope(scope = ESIntegTestCase.Scope.TEST, numDataNodes = 0)
+public class PrevalidateNodeRemovalWithSearchableSnapshotTests extends BaseSearchableSnapshotsIntegTestCase {
```
henningandersen, line 28: Conventionally, these should be named either `IT` or `IntegTests`.

server/src/main/java/org/elasticsearch/action/admin/cluster/node/shutdown/PrevalidateNodeRemovalRequest.java
```diff
@@ -0,0 +1,60 @@
...
+public class PrevalidateNodeRemovalRequest extends MasterNodeReadRequest<PrevalidateNodeRemovalRequest> {
+
+    private final String[] nodeIds; // Maybe named just nodes?
+    private DiscoveryNode[] concreteNodes;
+
+    public PrevalidateNodeRemovalRequest(String... nodeIds) {
+        this.nodeIds = nodeIds;
+    }
+
+    public PrevalidateNodeRemovalRequest(final StreamInput in) throws IOException {
```
henningandersen, line 29: Can we add a `AbstractWireSerializingTestCase` for this class too?

rest-api-spec/src/main/resources/rest-api-spec/api/_internal.prevalidate_node_removal.json
pxsalehi: Why? What would be the requirement to call it stable? I thought we start with experimental and at some point call it stable. Do we have some guidelines about this?

docs/reference/cluster/prevalidate-node-removal.asciidoc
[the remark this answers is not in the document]
pxsalehi: If we are moving away from having the nodes in the path (I started with that since I thought it would be similar/consistent to some other APIs we have), should we have three query params and allow setting only one of them per request?

```
POST /_internal/prevalidate_node_removal?name=node1,node2
POST /_internal/prevalidate_node_removal?id=id1,id2
POST /_internal/prevalidate_node_removal?external_id=id1,id2
``` 

This avoids any kind of ambiguity and we can just resolve based on id, name or external ID.

I'm not entirely sure how problematic the length of the URL can be. I am assuming up to 2000 is ok and considering ID/names of length 40, and the fact that the number of id/names is in the order of a fraction of the cluster size, I think not using a body should be ok. WDYT?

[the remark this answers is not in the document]
henningandersen: Sounds fine to me.

rest-api-spec/src/main/resources/rest-api-spec/api/_internal.prevalidate_node_removal.json
henningandersen: Most features start as stable. In this case, the primary consumer is orchestration and as such we need to be confident enough in the feature to release it GA, i.e., fully support it and only make bwc changes to it in the future.

We can also hide this behind a feature flag until it is ready but I see little value in getting it out as experimental since we will be very unlikely to receive feedback affecting the design.

henningandersen (commented):
I did not see any code changes so I simply responded to questions in this review cycle.

pxsalehi:
> I did not see any code changes so I simply responded to questions in this review cycle.

@henningandersen Thanks for the answers. I haven't pushed any code, and the review request was by mistake I guess. :-)

docs/reference/cluster/prevalidate-node-removal.asciidoc
```diff
@@ -0,0 +1,91 @@
...
+
+[[prevalidate-node-removal-api-prereqs]]
+==== {api-prereq-title}
+
+* If the {es} {security-features} are enabled, you must have the `monitor` or `manage` <<privileges-list-cluster,cluster privilege>> to use this API.
+
+[[prevalidate-node-removal-api-desc]]
+==== {api-description-title}
+
+This API allows validating whether it is safe to remove the specified node(s) from the cluster. In case the cluster has a <<cluster-health,`red` cluster health status>>, this verifies that the removal of the node(s) would not risk removing the last remaining copy of an unassigned shard.
```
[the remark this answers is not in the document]
pxsalehi: I've made some changes. Please let me know if this is still not properly describing the API or not.

server/src/main/java/org/elasticsearch/action/admin/cluster/node/shutdown/TransportPrevalidateNodeRemovalAction.java
```diff
@@ -0,0 +1,171 @@
...
+                Result result;
+                List<NodeResult> nodes;
+                if (redNonSSIndices.isEmpty()) {
+                    result = new Result(IsSafe.YES, "all red indices are searchable snapshot indices");
+                    nodes = prevalidationRequest.getConcreteNodes()
+                        .stream()
+                        .map(dn -> new NodeResult(dn.getName(), dn.getId(), dn.getExternalId(), new Result(IsSafe.YES, "")))
+                        .toList();
+                } else {
+                    result = new Result(IsSafe.UNKNOWN, "cluster health is RED");
```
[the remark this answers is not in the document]
pxsalehi: I was going for later finding out the true NOs and separate them from the UNKNOWN. But I think it works both ways. I've changed to NO. So for now we'll have no UKNOWN answer.

rest-api-spec/src/main/resources/rest-api-spec/api/_internal.prevalidate_node_removal.json
pxsalehi: Thanks for the explanation. I've changed to stable. Frankly, I imagined we'd leave this experimental until at lease the second part of this API is merged, especially since this is only for internal uses I thought that would be possible w/o hindering adoption. Nonetheless, for now I cannot think of any non-BWC changes that the second part could introduce. It probably adds a timeout parameter and some of those NOs get refined into YES/NO/UNKNOWNs.

server/src/internalClusterTest/java/org/elasticsearch/cluster/PrevalidateNodeRemovalIT.java
```diff
@@ -0,0 +1,98 @@
...
+import org.elasticsearch.common.Priority;
+import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.test.ESIntegTestCase;
+
+import static org.hamcrest.Matchers.equalTo;
+
+@ESIntegTestCase.ClusterScope(scope = ESIntegTestCase.Scope.TEST, numDataNodes = 0)
+public class PrevalidateNodeRemovalIT extends ESIntegTestCase {
+
+    public void testNodeRemovalFromNonRedCluster() throws Exception {
```
pxsalehi, line 27: While revisiting this test, I thought of two corner cases which are not tested here, and I am not sure they make sense or not. So I'd like to ask for an opinion.

1. If the cluster is green/yellow, and I send a prevalidation request for all nodes in the cluster, the answer would be YES for the overall result and each node's result! Clearly, we do not mean one could remove all those nodes at once w/o causing any issue, even though such a test would pass! Should we explicitly mention this in the doc, or somehow in case of a non-RED cluster just return no responses. From what I understood this API is for clusters which are RED and could be preceded by a call to the cluster health API. I am more inclined towards clarifying what an overall yes means in the doc (specifically pointing out that it doesn't mean that all nodes could be removed at once).
2. Would it be necessary to validate that all the nodes that are specified are data nodes? Currently, we kind of assume that.

pxsalehi (commented):
@henningandersen I've addresses all the comments and have added more tests. I do not have any more changes to add. Please have a look.

server/src/internalClusterTest/java/org/elasticsearch/cluster/PrevalidateNodeRemovalIT.java
henningandersen: Thanks for those explicit questions:

1. Yes, we should clarify that in docs. It is not a replacement for using the shutdown API, it is a pre-validation, answering whether doing so would/should succeed. If there are cases here that are not handled by the shutdown API (might be), we should probably also bake them in (as a follow-up).
2. pre-validation should say yes to any non-data node regardless of shard status. We should still encourage to use the pre-validation API in case we have future requirements for other types of nodes.

pxsalehi: @henningandersen thanks for the answer. I've added some notes to the docs. The PR is ready for review. 

https://elasticsearch_88952.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/prevalidate-node-removal-api.html

pxsalehi:
While this is waiting for review. I'm gonna start with the second part.

pxsalehi:
As discussed with @henningandersen, I'll remove `UNKNOWN` from the possible answers that the API can return. Therefore, the `is_safe` node result will be a boolean. However, we'll add another field that helps refine the answer in a parsable way, and contains a certain "category" for the yes/no response. I propose renaming the current `reason` field to `message` and using `reason` as the name for the new field. As an example:

```json
result : {
  is_safe: "UNKNOWN",
  reason: "failed contacting the node with ID XXXX: exception ...."
}
```
will change to:
```json
result : {
  is_safe: "NO",
  reason: "no_response_from_node",
  message: "failed contacting the node with ID XXXX: exception ...."
}
```

pxsalehi:
@henningandersen I'm trying to make the change mentioned in my last comment. Since this needs to be changed in several places (parser, tests, etc), I want to make sure I'll get it right this time. I've decided to remove the 'result' top-level object since it is not directly useful there. So the API response would change from this:
```
{
  "result": {
    "is_safe": "UNKNOWN",  // --> or YES, or NO
    "reason": "cluster health is RED"  // --> Just a string
  },
  "nodes": [
    {
      "id": "node1-id",
      "name" : "node1",
      "external_id" : "node1-externalId",
      "result" : {  // -> Same type as the top-level result
        "is_safe": "YES",
        "reason": ""
      }
    }
  ]
}
```
to this:
```
{
  "is_safe": "false",  //--> boolean
  "message": "cluster health is RED"  // --> Just a string
  "nodes": [
    {
      "id": "node1-id",
      "name" : "node1",
      "external_id" : "node1-externalId",
      "result" : {  
        "is_safe": "true",  //--> boolean
        "reason": "no_red_shards",  //--> (just an example) an enum of fixed Strings (useful for parsing/automation)
        "message": "node does not host any red shards" //--> A String that could be anything with details specific to the call/node
      }
    }
  ]
}
```
What do you think about this?

henningandersen:
Thanks @pxsalehi that mostly makes sense, 1-2 comments:

>  ` "is_safe": "false",  //--> boolean`

should be 

>   "is_safe": false,  //--> boolean

Also, renaming reason to message makes sense here, but we could defer the introdution of the reason field to pt.2 (or later) to avoid too many changes in this PR (that is mostly reviewed).

pxsalehi:
@henningandersen Thanks for the feedback. I've made the changes. This is ready for review again. Please have a look.

pxsalehi:
@elasticmachine update branch

henningandersen (approved):
LGTM

Left a few comments but no need for another review round unless significant changes are made.

server/src/main/java/org/elasticsearch/action/admin/cluster/node/shutdown/PrevalidateNodeRemovalRequest.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        super.writeTo(out);
+        if (names == null) {
+            out.writeVInt(0);
+        } else {
+            out.writeStringArray(names);
+        }
```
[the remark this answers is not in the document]
pxsalehi: Yes. That's not needed anymore. I've removed them.

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/cluster.prevalidate_node_removal/10_basic.yml
```diff
@@ -0,0 +1,58 @@
...
+  - skip:
+      version: "- 8.5.99"
+      reason: "API added in 8.6.0"
+  - do:
+      catch: bad_request
+      _internal.prevalidate_node_removal: {}
+
+  - match: { status: 400 }
+  - match: { error.type: action_request_validation_exception }
+  - match: { error.reason: "Validation Failed: 1: request must contain node names, or IDs, or external IDs;" }
```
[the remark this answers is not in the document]
pxsalehi: These are fields from the `ValidationException` and already exist with those names.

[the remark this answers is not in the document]
henningandersen: Thanks and sorry for the confusion. I simply searched for reason and stumbled upon it, but clearly you are right.

pxsalehi:
@elasticmachine update branch

pxsalehi:
Thanks for the reviews Henning!

pxsalehi:

PR #88954, opened 2022-07-29
Support source fallback for byte, short, and long fields

jdconrad:
This change adds source fallback support for byte, short, and long fields. These use the already existing class `SourceValueFetcherSortedNumericIndexFieldData`. Also note the majority of the additional lines are testing, so the change itself is actually pretty small.


jdconrad:
@elasticsearchmachine run elasticsearch-ci/packaging-tests-unix-sample

romseygeek (approved):
LGTM!

PR #88957, opened 2022-07-29
Date histogram range edge case fix

tmgordeeva:
This fixes an illegal argument exception when using conflicting ranges.
Sometimes if the min of the range is high enough above the max, we get an error
in TimeUnitRounding.prepareOffsetOrJavaTimeRounding because our min time is
greater than our max. This appears to have started when we began using query
bounds to bound ranges.


tmgordeeva:
The REST test here does not actually reproduce the problem, despite using the same steps as in the original issue and those steps reproducing the problem using a locally built ES. Somehow these ranges are detected as disjoint in some cases but not others, which might be a bug in itself.

server/src/main/java/org/elasticsearch/search/aggregations/support/CoreValuesSourceType.java
```diff
@@ -362,11 +363,44 @@ public void visitLeaf(Query query) {
...
+                            @Override
+                            public double roundingSize(Rounding.DateTimeUnit timeUnit) {
+                                return 0;
+                            }
+
+                            @Override
+                            public long[] fixedRoundingPoints() {
+                                return null;
+                            }
+                        };
```
nik9000, line 401: I think maybe we should move this into `Rounding` and give it a couple of unit tests. Maybe just make it a constant?

tmgordeeva: I've convinced myself if we use the unknown rounding, the performance shouldn't matter since it should never get any actual values to use. Then we can avoid having this weird "empty" rounding class that I'm not sure fulfills whatever contracts Rounding is currently guaranteeing.

nik9000 (approved):

PR #55502, opened 2020-04-21
value type, unmapped field, scripting tests for date range agg

andyb-elastic:
Adds tests for unmapped fields, supported values source types, and
scripting. Moves some tests over from the integration test

In AggregatorTestCase, skips asserting on the bucket consumer when
collection did not occur, e.g. because of an unmapped field

PR #88958, opened 2022-07-29
Port tests for `date_range`

nik9000:
This ports some of the tests for `date_range` from a high overhead
`IntegTestCase` to a lower overhead `AggregatorTestCase`.
`AggregatorTestCase` is also lower level so it can find more fun things
like memory leaks. And it enables values source type testing for
`date_range`.

Inspired by #55502 by andy.bristol@elastic.co


not-napoleon (approved):
LGTM, thanks for picking this up

nik9000:

PR #88960, opened 2022-07-29
Ignore beats artifacts when resolving all artifact dependencies

mark-vieira:
After version bumps due to releases or feature freeze, snapshot versions of beats artifacts may not be published for some time. During that time attempts to run `resolveAllDependencies` to seed our CI worker images will fail, meaning they'll be out of date. We don't really need to seed these images with snapshot dependencies that are likely to change anyhow so let's just ignore them.


mark-vieira:

PR #88925, opened 2022-07-29
Deprecate overriding DiscoveryPlugin internals

rjernst:
DiscoveryPlugin allows extending getJoinValidator and
getElectionStrategies. These are implementation details of the system.
This commit deprecates these methods so that plugin authors
are discouraged from overriding them.

PR #88961, opened 2022-07-29
Add deprecation message for deprecated plugin APIs

rjernst:
Plugin APIs are defined by a set of interfaces from server. Many of
these APIs are actually implementation details of the system. As we move
these implementation details to use different hook mechanisms so that
internals are only implementable by builtin components, the existing
plugin APIs need to be deprecated. Java provides a means to indicate
deprecation - through the `@Deprecated` annotation. But that annotation
is only seen when compiling a plugin implementing deprecated hooks, and
only then if deprecation warnings are not disabled.

This commit adds an introspection step to plugin initialization which
inspects each loaded plugin and looks for any APIs marked with the
@Deprecated annotation which are overridden by the plugin. If any are
found, deprecation log messages are then emitted to the deprecation log.

### DIFF
--- a/server/src/main/java/org/elasticsearch/plugins/DiscoveryPlugin.java
+++ b/server/src/main/java/org/elasticsearch/plugins/DiscoveryPlugin.java
@@ -76,6 +76,7 @@ public interface DiscoveryPlugin {
      * join attempt but might be called multiple times during the lifetime of a node. Validators are expected to throw a
      * {@link IllegalStateException} if the node and the cluster-state are incompatible.
      */
+    @Deprecated
     default BiConsumer<DiscoveryNode, ClusterState> getJoinValidator() {
         return null;
     }
@@ -83,6 +84,7 @@ public interface DiscoveryPlugin {
     /**
      * Allows plugging in election strategies (see {@link ElectionStrategy}) that define a customized notion of an election quorum.
      */
+    @Deprecated
     default Map<String, ElectionStrategy> getElectionStrategies() {
         return Collections.emptyMap();
     }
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/elasticsearch/plugins/DiscoveryPlugin.java
```diff
@@ -76,13 +76,15 @@ default Map<String, Supplier<SeedHostsProvider>> getSeedHostProviders(
...
      */
+    @Deprecated
     default BiConsumer<DiscoveryNode, ClusterState> getJoinValidator() {
         return null;
     }
 
     /**
      * Allows plugging in election strategies (see {@link ElectionStrategy}) that define a customized notion of an election quorum.
      */
+    @Deprecated
```
rjernst, line 87: These deprecations will be added as part of https://github.com/elastic/elasticsearch/pull/88925

server/src/main/java/org/elasticsearch/plugins/PluginIntrospector.java
```diff
@@ -67,36 +79,66 @@ static PluginIntrospector getInstance() {
...
                 try {
                     Method m = pluginClass.getMethod(mt.name(), mt.parameterTypes());
-                    if (m.getDeclaringClass() == esPluginClass) {
+                    if (m.getDeclaringClass() == clazz) {
                         // it's not overridden
                     } else {
-                        assert esPluginClass.isAssignableFrom(m.getDeclaringClass());
-                        overriddenMethods.add(mt.name());
+                        assert clazz.isAssignableFrom(m.getDeclaringClass());
+                        overriddenMethods.put(mt.name(), clazz.getSimpleName());
```
grcevski: We should assert that overridenMethods.get(mt.name()) == null. Just in case someone mistakenly duplicates a name in two different interfaces.

grcevski (approved):
LGTM!

PR #88966, opened 2022-07-29
REST tests for cumulative pipeline aggs

nik9000:
Adds REST tests for the `cumulative_cardinality` and `cumulative_sum`
pipeline aggregations. This gives us forwards and backwards compatibility
tests for these aggs as well as mixed version cluster tests for these
aggs.

Relates to #26220


not-napoleon (approved):
LGTM

PR #88971, opened 2022-07-30
Fix PluginsServiceTests on Windows

ChrisHegarty:
This commit ensures that the PluginsServiceTests closes its
URLClassLoaders of loaded plugins, thus allowing the post-test cleanup
to succeed. (previously two of the test scenarios fail in post-test
cleanup on Windows)


rjernst (approved):
LGTM, sorry for that!

PR #88986, opened 2022-08-01
Avoid expensive loop in indicesDeletedFromClusterState() when possible

original-brownbear:
The loop over all indices here gets very expensive for large states, we
can avoid it often when metadata changes but not the indices maps.

relates #77466


idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

PR #88996, opened 2022-08-01
Extract least/most available paths to DiskUsage

gmarouli:
In the `InternalClusterInfoService` we collect the paths with the most and least available disk spaces. 

For the health API we are also interested in the path with the least available disk space, for this reason, we propose this refactoring. We extract the calculation of the least and most available disk space to separate methods and moved them to the `DiskUsage`.

Trade-off:
Obviously, the code now is less efficient because it needs two passes to calculate the least and most available disk space, if there are no objections, I think this is not too bad.

Relates to: #88390


andreidan (approved):
LGTM, thanks Mary

PR #89000, opened 2022-08-01
Refactor WildcardExpressionResolver to better track usages of indices lookup

albertzaharovits:
This is a pure refactoring of the `WildcardExpressionResolver`.
The objective is to restrict access to the indices lookup through the context parameter only.
Eventually, Security is going to plug into the context and only show a restricted view of the indices lookup, particular to the user context.


server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1245,22 +1237,23 @@ private static void validateAliasOrIndex(String expression) {
             }
         }
 
-        private static boolean aliasOrIndexExists(Context context, IndicesOptions options, Metadata metadata, String expression) {
-            IndexAbstraction indexAbstraction = metadata.getIndicesLookup().get(expression);
+        private static Optional<RuntimeException> aliasOrIndexExists(Context context, String expression) {
```
williamrandolph, line 1240: I think it's confusing to generate an exception and decide whether or not to throw it later in the code. On the other hand, I see the benefit in reducing the the checks and lookups that were previously duplicated.

I can't think of an obviously better way of doing this. The best I could come up with was:
```
        private static boolean aliasOrIndexExists(Context context, String expression, boolean throwExceptionIfAbsent) {
            final IndicesOptions options = context.getOptions();
            IndexAbstraction indexAbstraction = context.getState().getMetadata().getIndicesLookup().get(expression);
            if (indexAbstraction == null) {
                if (throwExceptionIfAbsent) {
                    throw indexNotFoundException(expression);
                }
                return false;
            }
...
```

What do you think?

williamrandolph (commented):
Looks mostly good to me. The one thing I found confusing was the change in the exception handling, though I don't object strongly enough to "request changes."

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1284,53 +1277,53 @@ private static IndexMetadata.State excludeState(IndicesOptions options) {
...
         }
 
         private static Map<String, IndexAbstraction> filterIndicesLookup(
-            Context context,
-            SortedMap<String, IndexAbstraction> indicesLookup,
+            Map<String, IndexAbstraction> indicesLookup,
             Predicate<? super Map.Entry<String, IndexAbstraction>> filter,
-            IndicesOptions options
+            boolean ignoreAliases,
+            boolean includeDataStreams
```
jakelandis, line 1314: minor: why not keep the Context here and pull the information (indicesLookup, ignoreAliases, includeDataStreams) from that ? It would be more inline with the other refactoring here.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
jakelandis: Agree this is odd pattern. It is also quite expensive to raise a exceptions especially if not used. I think some usage patterns would see a noticeable performance hit. I like William's suggestion.

jakelandis (changes requested):
Changes look good [1] and believe they are like for like for the prior code and reads a bit clean. 

[1] Except for the exception handling raised by @williamrandolph , I think this needs to addressed prior to merging (other comments are non blocking)

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
albertzaharovits: Thanks for th suggestion, I have pushed: https://github.com/elastic/elasticsearch/pull/89000/commits/21ca35cff0a0a5cf0462765c90e970601c581775 .

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1245,22 +1237,23 @@ private static void validateAliasOrIndex(String expression) {
...
 
             // treat aliases as unavailable indices when ignoreAliases is set to true (e.g. delete index and update aliases api)
             if (indexAbstraction.getType() == IndexAbstraction.Type.ALIAS && options.ignoreAliases()) {
-                return false;
+                return Optional.of(aliasesNotSupportedException(expression));
             }
 
-            if (indexAbstraction.isDataStreamRelated() && context.includeDataStreams() == false) {
-                return false;
+            if (indexAbstraction.isDataStreamRelated() && false == context.includeDataStreams()) {
```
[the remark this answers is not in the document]
albertzaharovits: I remember that `false ==` is preferable. I also got into the habit of using that, unless the context uses the other convention.   
I have pushed: https://github.com/elastic/elasticsearch/pull/89000/commits/98c6ea4574e4f85eed61e249f38d86442f5075a5 .

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
albertzaharovits: The `filterIndicesLookup` method doesn't always act on the indices lookup from the context, see `suffixWildcard`.
I have pushed https://github.com/elastic/elasticsearch/pull/89000/commits/[781a768](https://github.com/elastic/elasticsearch/pull/89000/commits/781a7685688842c0de9c5efb867016dcec14da41)5688842c0de9c5efb867016dcec14da41 to make suffixWildcard not use the context.

albertzaharovits:
Thanks for the quick review @williamrandolph and @jakelandis . I have addressed your points. Please take another look.

williamrandolph (approved):
LGTM!

jakelandis (approved):
LGTM

PR #89001, opened 2022-08-01
Eliminating initial delay of CoordinationDiagnosticsService#beginPollingClusterFormationInfo for integration tests

masseyke:
`CoordinationDiagnosticsService#beginPollingClusterFormationInfo` schedules a task to run after an initial delay of
10 seconds. `CoordinationDiagnosticsServiceIT` is asserting that we have results in 10 seconds or less (using the
default `assertBusy`). This commit exposes a way for integration tests to change that initial delay so that we don't
waste 10 seconds waiting for the task to be scheduled.


dakrone (approved):

masseyke:
Relates to #89001

PR #85821, opened 2022-04-12
Introduce unified entrypoint for CLI scripts

rjernst:
CLI scripts have a common infrastructure in that they call to the shared
elasticsearch-cli shell script which launches them with the appropriate
java command line. However, each underlying Java class must implement
its own main method.

This commit introduces a single main method to be shared by CLIs. The
new CliToolLauncher takes in system properties to determine which tool
is being run, and a new CliToolProvider SPI allows defining and finding
the named tools.

relates #85758

PR #89003, opened 2022-08-01
[DOCS] Fix plugins CLI doc CLI_JAVA_OPTS env var

ebuildy:
The commit https://github.com/elastic/elasticsearch/commit/1d4534f848feb396c00cee09fc1d0aef24a529a2 changes the env variable ``ES_JAVA_OPTS`` to ``CLI_JAVA_OPTS``. Doc must be updated as well.

Relate to https://github.com/elastic/elasticsearch/pull/85821 

### DIFF
--- a/docs/plugins/plugin-script.asciidoc
+++ b/docs/plugins/plugin-script.asciidoc
@@ -109,7 +109,7 @@ to a local Java truststore and pass the location to the script as follows:
 +
 [source,shell]
 -----------------------------------
-sudo ES_JAVA_OPTS="-Djavax.net.ssl.trustStore=/path/to/trustStore.jks" bin/elasticsearch-plugin install https://host/plugin.zip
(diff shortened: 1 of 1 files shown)


michaelbaamonde:
I believe this would be in Core/Infra's purview, so labeling as such.

williamrandolph:
Ryan is on leave, but, yes, `ES_JAVA_OPTS` should be changed to `CLI_JAVA_OPTS` in documentation. The rationale for the change is that these options are not just used for Elasticsearch, but for other tools like the keystore tool or the plugin CLI. It looks like the remaining uses of `ES_JAVA_OPTS` outside of documentation need to be cleaned up as well.

ebuildy:
I discovered this change because we build a Docker image to add elasticsearch plugins, using HTTP proxies. And our CI was broken from this commit https://github.com/elastic/elasticsearch/pull/85821.

lockewritesdocs (approved):
Thanks for finding and fixing this issue @ebuildy! I'll update this variable in the remainder of our docs.

PR #89005, opened 2022-08-01
Preemptively compute RoutingNodes and the indices lookup during publication

original-brownbear:
Computing routing nodes and the indices lookup takes considerable time
for large states. Both are needed during cluster state application and
Prior to this change would be computed on the applier thread in all cases.
By running the creation of both objects concurrently to publication, the
many shards benchmark sees a 10%+ reduction in the bootstrap time to
50k indices.

I chose the generic pool for this since there is some chance that a thread will just needlessly block a thread if the pool takes
too long to get to the computation and the cluster applier gets to work first which doesn't hurt much on the generic pool where
it most likely won't block other operations like it would on a smaller pool like management.

relates #77466

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/service/MasterService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/service/MasterService.java
@@ -330,6 +330,10 @@ public class MasterService extends AbstractLifecycleComponent {
                     }
 
                     logger.debug("publishing cluster state version [{}]", newClusterState.version());
+                    // initialize routing nodes and the indices lookup concurrently, we will need both of them for the cluster state
+                    // application and can compute them while we wait for the other nodes during publication
+                    threadPool.generic().execute(newClusterState::getRoutingNodes);
+                    threadPool.generic().execute(() -> newClusterState.metadata().getIndicesLookup());
                     publish(
                         clusterStatePublicationEvent,
                         new CompositeTaskAckListener(
--- a/server/src/main/java/org/elasticsearch/cluster/service/MasterService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/service/MasterService.java
@@ -333,7 +333,7 @@ public class MasterService extends AbstractLifecycleComponent {
                     // initialize routing nodes and the indices lookup concurrently, we will need both of them for the cluster state
                     // application and can compute them while we wait for the other nodes during publication
                     threadPool.generic().execute(newClusterState::getRoutingNodes);
-                    threadPool.generic().execute(() -> newClusterState.metadata().getIndicesLookup());
+                    threadPool.generic().execute(newClusterState.metadata()::getIndicesLookup);
                     publish(
                         clusterStatePublicationEvent,
                         new CompositeTaskAckListener(
--- a/docs/changelog/89005.yaml
+++ b/docs/changelog/89005.yaml
@@ -1,0 +1,5 @@
+pr: 89005
+summary: Preemptively compute `RoutingNodes` and the indices lookup during publication
+area: Cluster Coordination
+type: enhancement
+issues: []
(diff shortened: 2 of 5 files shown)


server/src/main/java/org/elasticsearch/cluster/service/MasterService.java
```diff
@@ -330,6 +330,10 @@ public String getDescription() {
                     }
 
                     logger.debug("publishing cluster state version [{}]", newClusterState.version());
+                    // initialize routing nodes and the indices lookup concurrently, we will need both of them for the cluster state
+                    // application and can compute them while we wait for the other nodes during publication
+                    threadPool.generic().execute(newClusterState::getRoutingNodes);
+                    threadPool.generic().execute(() -> newClusterState.metadata().getIndicesLookup());
```
original-brownbear, line 336: I benchmarked this without the locking and with profiling and even at close to 100k indices these would always run before the applier thread gets to needing either of these two objects, so while not saving any cycles this at least makes better use of master's cores.

original-brownbear:
Jenkins run elasticsearch-ci/part-1  (unrelated + known)

original-brownbear:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/cluster/service/MasterService.java
idegtiarenko, line 336:
```suggestion
                    threadPool.generic().execute(newClusterState.metadata()::getIndicesLookup);
```

idegtiarenko: Should we add some logging to find out when this starts to happen?

```
    private synchronized RoutingNodes buildRoutingNodes() {
        RoutingNodes r = routingNodes;
        if (r != null) {
            logger.warn("RoutingNodes object is already build.");
            return r;
        }
        r = RoutingNodes.immutable(routingTable, nodes);
        routingNodes = r;
        return r;
    }
```

original-brownbear: I don't think we can make this a warning, it will happen here and there for sure just by chance and that's ok IMO.
Even if this happens it's not so bad and still better than what we have today for heavy cases.
Either the applier thread enters the synchronized block before the generic threads (which is pretty unlikely) and the worst that happens is that we waste a generic thread blocked on nothing for a bit while it does the computation and performance is pretty much what we have right now.
Or the more likely case, the generic thread enters the block first happens, in which case even if it doesn't finish before the applier gets to the block, that's still better than the previous case of the applier having to do all the work itself, while now it merely has to wait for the already in-progress work on `GENERIC` to finish (though that seems to not be happening even at 200k indices in my benchmarks :)).

original-brownbear:
Jenkins run elasticsearch-ci/part-1  (good old network wires crossed it seems)

DaveCTurner (commented):
Looks fine to me as-is, but how about launching this work in `Coordinator#handlePublishRequest` so that all nodes get a headstart on these computations rather than just the master?

original-brownbear:
> Looks fine to me as-is, but how about launching this work in Coordinator#handlePublishRequest so that all nodes get a headstart on these computations rather than just the master?

But that will run a little later on master, after forking to the `cluster_coordination` pool I think?
Maybe we can do it there for non-master nodes and keep it here for the master? That said I'd rather keep that in a follow-up and run a quick benchmark on it in isolation to make sure nothing was missed. I can't see how that would cause trouble on data nodes but I'm a little paranoid here :)

DaveCTurner (approved):
I doubt it's _meaningfully_ later in most cases (we could do this before writing the state to disk) but sure we can investigate that in a follow up.

original-brownbear:
Thanks David + Ievgen!

PR #89010, opened 2022-08-01
Support source fallback for double, float, and half_float field types

jdconrad:
This change adds a `SourceValueFetcherSortedDoubleIndexFieldData` to support double doc values types for source fallback. This also adds support for double, float and half_float field types. Change is smaller than it looks as a good portion of the line count is testing.

### DIFF
--- a/server/src/main/java/org/elasticsearch/index/fielddata/SourceValueFetcherSortedDoubleIndexFieldData.java
+++ b/server/src/main/java/org/elasticsearch/index/fielddata/SourceValueFetcherSortedDoubleIndexFieldData.java
@@ -58,13 +58,13 @@ public class SourceValueFetcherSortedDoubleIndexFieldData extends SourceValueFet
     }
 
     @Override
-    public SourceValueFetcherSortedNumericLeafFieldData loadDirect(LeafReaderContext context) throws Exception {
-        return new SourceValueFetcherSortedNumericLeafFieldData(toScriptFieldFactory, context, valueFetcher, sourceLookup);
+    public SourceValueFetcherLeafFieldData<SortedNumericDoubleValues> loadDirect(LeafReaderContext context) throws Exception {
+        return new SourceValueFetcherSortedDoubleLeafFieldData(toScriptFieldFactory, context, valueFetcher, sourceLookup);
     }
 
-    public static class SourceValueFetcherSortedNumericLeafFieldData extends SourceValueFetcherLeafFieldData<SortedNumericDoubleValues> {
+    private static class SourceValueFetcherSortedDoubleLeafFieldData extends SourceValueFetcherLeafFieldData<SortedNumericDoubleValues> {
 
-        public SourceValueFetcherSortedNumericLeafFieldData(
+        private SourceValueFetcherSortedDoubleLeafFieldData(
             ToScriptFieldFactory<SortedNumericDoubleValues> toScriptFieldFactory,
             LeafReaderContext leafReaderContext,
             ValueFetcher valueFetcher,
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/elasticsearch/index/fielddata/SourceValueFetcherSortedDoubleIndexFieldData.java
```diff
@@ -0,0 +1,130 @@
...
+    ) {
+        super(fieldName, valuesSourceType, valueFetcher, sourceLookup, toScriptFieldFactory);
+    }
+
+    @Override
+    public SourceValueFetcherSortedNumericLeafFieldData loadDirect(LeafReaderContext context) throws Exception {
+        return new SourceValueFetcherSortedNumericLeafFieldData(toScriptFieldFactory, context, valueFetcher, sourceLookup);
+    }
+
+    public static class SourceValueFetcherSortedNumericLeafFieldData extends SourceValueFetcherLeafFieldData<SortedNumericDoubleValues> {
```
romseygeek, line 65: Do these need to be public?

server/src/main/java/org/elasticsearch/index/fielddata/SourceValueFetcherSortedDoubleIndexFieldData.java
```diff
@@ -0,0 +1,130 @@
...
+        @Override
+        public DocValuesScriptFieldFactory getScriptFieldFactory(String name) {
+            return toScriptFieldFactory.getScriptFieldFactory(
+                new SourceValueFetcherSortedNumericDoubleValues(leafReaderContext, valueFetcher, sourceLookup),
+                name
+            );
+        }
+    }
+
+    public static class SourceValueFetcherSortedNumericDoubleValues extends SortedNumericDoubleValues implements ValueFetcherDocValues {
```
romseygeek, line 85: Same here, can we make these private or at least package-protected?

server/src/main/java/org/elasticsearch/index/fielddata/SourceValueFetcherSortedDoubleIndexFieldData.java
```diff
@@ -0,0 +1,130 @@
...
+            this.sourceLookup = sourceLookup;
+        }
+
+        @Override
+        public boolean advanceExact(int doc) throws IOException {
+            sourceLookup.setSegmentAndDocument(leafReaderContext, doc);
+            values = new TreeSet<>();
+
+            for (Object value : valueFetcher.fetchValues(sourceLookup, Collections.emptyList())) {
+                values.add(((Number) value).doubleValue());
```
romseygeek, line 111: The cast makes me a bit worried as we don't have any kind of type checking on ValueFetcher at the moment.  Maybe add an `assert value instanceof Number` line above it for now?

romseygeek (approved):
A couple of nits but LGTM otherwise.  No need for another review.

server/src/main/java/org/elasticsearch/index/fielddata/SourceValueFetcherSortedDoubleIndexFieldData.java
jdconrad: Done. I added this to the numeric version as well. My understanding is the only way this could actually cause an assertion is if there's a bug during indexing because it should be caught as an invalid type at that time.

server/src/main/java/org/elasticsearch/index/fielddata/SourceValueFetcherSortedDoubleIndexFieldData.java
jdconrad: Changed to private.

server/src/main/java/org/elasticsearch/index/fielddata/SourceValueFetcherSortedDoubleIndexFieldData.java
jdconrad: Changed to private.

jdconrad:
@romseygeek Thank you for the review. I've addressed the feedback and will commit as soon as CI passes.

PR #89014, opened 2022-08-01
Polling for cluster diagnostics information

masseyke:
As shown in step 1.2.2.3 in the diagram at https://github.com/elastic/elasticsearch/issues/85624#issuecomment-1175421820, if there has not been a node elected master for more than 30 seconds, then when a user queries a non-master-eligible node it is supposed to reach out to a master-eligible node to gather information about what that node thinks of master stability. Since we don't want to block a user request while making those remote calls, it should be done speculatively so that we have the information when it is needed. If a non-master-eligible node gets a cluster state changed event that the master is null, then it will wait 10 seconds (in case it gets another cluster state changed event that the master has flipped to non-null quickly, which is the most likely scenario), then reach out to a random master-eligible node using a transport action (https://github.com/elastic/elasticsearch/pull/87984) to get its cluster diagnostics data. This data is stashed away in case a user request comes in. Since the user won't hit this path for 30 seconds after a master has become null (because we tell the user they have a stable master unless it's been missing for more than 30 seconds), we have plenty of time for this -- 10 seconds of delay in case we're notified that there is a new master plus 10 seconds to get the result back from the master-eligible node before timing out, plus 10 seconds to spare.
Note that the pattern used in this PR for doing asynchronous polling is very similar to what was done in https://github.com/elastic/elasticsearch/pull/88874. The main differences are that we're polling a single master node here instead of all of them, and we're using a different transport action.
Also note that there will be a follow-up PR that actually uses this information to make decisions about master stability. This PR only keeps it for later use.


server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -814,6 +840,198 @@ private Scheduler.Cancellable fetchClusterFormationInfo(
...
+        long startTime = System.nanoTime();
+        connectionListener.whenComplete(releasable -> {
+            if (masterEligibleNode == null) {
+                responseConsumer.accept(null);
+            } else {
+                /*
+                 * Since this block is not synchronized it is possible that this task is cancelled between the check above and when the
+                 * code below is run, but this is harmless and not worth the additional synchronization in the normal case. The result
+                 * will just be ignored.
+                 */
```
andreidan, line 956: This is a bit misleading as the check above is 
``` 
if (masterEligibleNode == null) {
```
I think we should rephrase this comment

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -814,6 +840,198 @@ private Scheduler.Cancellable fetchClusterFormationInfo(
...
+                        masterEligibleNode,
+                        masterEligibleNode.getVersion(),
+                        minSupportedVersion
+                    );
+                } else {
+                    /*
+                     * Since this block is not synchronized it is possible that this task is cancelled between the check above and when the
+                     * code below is run, but this is harmless and not worth the additional synchronization in the normal case. In that case
+                     * the connection will just be closed and the transport request will not be made.
+                     */
```
andreidan, line 1014: This is a bit misleading as the check above is either 
``` 
if (masterEligibleNode == null) {
```
or
```
if (masterEligibleNode.getVersion().onOrAfter(minSupportedVersion) == false) { 
```
I think we should rephrase this comment

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -814,6 +840,198 @@ private Scheduler.Cancellable fetchClusterFormationInfo(
...
+                        masterEligibleNode,
+                        ConnectionProfile.buildDefaultConnectionProfile(clusterService.getSettings()),
+                        connectionListener
+                    );
+                }
+            }
+        }, remoteRequestInitialDelay, ThreadPool.Names.SAME);
+    }
+
+    void cancelPollingRemoteStableMasterHealthIndicatorService() {
```
andreidan, line 1026: We should rename this as we're not polling `StableMasterHealthIndicatorService`

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -99,12 +101,24 @@ public class CoordinationDiagnosticsService implements ClusterStateListener {
     // Non-private for testing
     volatile ConcurrentMap<DiscoveryNode, ClusterFormationStateOrException> clusterFormationResponses = null;
 
+    /*
+     * This is a reference to the task that is periodically reaching out to a master eligible node to get its CoordinationDiagnosticsResult
+     * for diagnosis. It is null when no polling is occurring.
+     * The field is accessed (reads/writes) from multiple threads, and is also reassigned on multiple threads.
+     */
+    volatile AtomicReference<Scheduler.Cancellable> remoteStableMasterHealthIndicatorTask = null;
```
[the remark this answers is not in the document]
andreidan: Does this need to be an `AtomicReference` or a `volatile Cancellable` would suffice?

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -99,12 +101,24 @@ public class CoordinationDiagnosticsService implements ClusterStateListener {
...
+    /*
+     * This field holds the result of the task in the remoteStableMasterHealthIndicatorTask field above. The field is accessed
+     * (reads/writes) from multiple threads, and is also reassigned on multiple threads.
+     */
+    volatile AtomicReference<RemoteMasterHealthResult> remoteCoordinationDiagnosisResult = null;
```
andreidan, line 114: Does this need to be an `AtomicReference` or a `volatile RemoteMasterHealthResult` would suffice?

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -814,6 +840,198 @@ private Scheduler.Cancellable fetchClusterFormationInfo(
...
+    /**
+     * This method returns quickly, but in the background schedules to query the remote masterEligibleNode's cluster diagnostics in 10
+     * seconds unless cancel() is called on the Cancellable that this method returns.
+     * @param masterEligibleNode The masterEligibleNode to poll for cluster diagnostics. This masterEligibleNode can be null in the case
+     *                           when there are not yet any master-eligible nodes known to this masterEligibleNode's PeerFinder.
+     * @param responseConsumer The consumer of the cluster diagnostics for the masterEligibleNode, or the exception encountered while
+     *                         contacting it
+     * @return A Cancellable for the task that is scheduled to fetch cluster diagnostics
+     */
+    private Scheduler.Cancellable fetchCoordinationDiagnostics(
```
andreidan, line 941: This is very similar to `fetchClusterFormationInfo` ? Could we extract a common method to be used in both cases - this common method would additionaly have parameters for the action name, request and a response reader?

andreidan (commented):

Left a few comments

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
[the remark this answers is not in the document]
masseyke: We potentially assign this multiple times on the same polling "stack". So I think you're right that it doesn't necessarily have to be atomic, but it has to be some reference because we need to access it from the cancel method. That is, if I only had a Cancellable here, I couldn't reassign that same Cancellable reference to a new Cancellable and still have access to it in the cancel method. I can't just assign it to the object-level variable here because there might be more than one polling stack running at the same time.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: I'm not sure why I put that there. The comment in rescheduleDiagnosticsFetchConsumer covers this, so I'll just remove this one.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: What I was trying to say is better said in rescheduleDiagnosticsFetchConsumer, so removing this comment.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: For the same reason as the task AtomicReference, we really need this to be a reference. If I change it to just a RemoteMasterHealthResult, then each polling stack no longer gets its own copy so it is possible that a polling stack that we want to have been cancelled overwrites the result value of the current polling stack.

andreidan: Ah gotcha - we'd need an empty instance of `RemoteMasterHealthResult` to simulate the same thing. Thanks for the explanation !

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: Done.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -99,12 +105,24 @@ public class CoordinationDiagnosticsService implements ClusterStateListener {
...
+     */
+    volatile AtomicReference<RemoteMasterHealthResult> remoteCoordinationDiagnosisResult = null;
+
     /**
      * This is the amount of time that we wait before scheduling a remote request to gather diagnostic information. It is not
      * user-configurable, but is non-final so that integration tests don't have to waste 10 seconds.
      */
     // Non-private for testing
-    TimeValue remoteRequestInitialDelay = new TimeValue(10, TimeUnit.SECONDS);
+    public TimeValue remoteRequestInitialDelay = new TimeValue(10, TimeUnit.SECONDS);
```
andreidan, line 125: I think we can keep this `default` visible if we move `StableMasterDisruptionIT` to the `org.elasticsearch.cluster.coordination` package (it doesn't really need to be under `discovery` - and the `cluster.coordination` package is better suited given the indicator is under the same package too)

What do you think?

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -814,6 +840,198 @@ private Scheduler.Cancellable fetchClusterFormationInfo(
...
+                logger.trace("Not rescheduling request for cluster coordination info because this node is being shutdown", e);
+            } else {
+                throw e;
+            }
+        }
+    }
+
+    /**
+     * This wraps the responseConsumer in a Consumer that will run rescheduleDiagnosticsFetchConsumer() after responseConsumer has
+     * completed, adding the resulting Cancellable to cancellableConsumer.
```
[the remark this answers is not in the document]
andreidan:
```suggestion
     * completed, adding the resulting Cancellable to cancellableReference.
```

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -762,58 +792,228 @@ void cancelPollingClusterFormationInfo() {
...
+            Version.V_8_4_0
+        );
+    }
+
+    /**
+     * This method connects to masterEligibleNode and sends it a transport request for a remote response of type R. The remote response or
+     * exception are transformed into a common type T with responseToResultFunction or exceptionToResultFunction, and then consumed by
+     * responseConsumer.
+     *
+     * @param masterEligibleNode        The master eligible node to be queried, or null if we do not yet know of a master eligible node
```
andreidan, line 925: Shall we also mention that if null is passed that's the result the response consumer will get?

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -762,58 +792,228 @@ void cancelPollingClusterFormationInfo() {
...
+            CoordinationDiagnosticsAction.INSTANCE,
+            new CoordinationDiagnosticsAction.Request(true),
+            e -> new RemoteMasterHealthResult(masterEligibleNode, null, e),
+            response -> new RemoteMasterHealthResult(masterEligibleNode, response.getCoordinationDiagnosticsResult(), null),
+            Version.V_8_4_0
+        );
+    }
+
+    /**
+     * This method connects to masterEligibleNode and sends it a transport request for a remote response of type R. The remote response or
```
andreidan, line 921:
```suggestion
     * This method connects to masterEligibleNode and sends it a transport request for a response of type R. The  response or
```

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -762,58 +792,228 @@ void cancelPollingClusterFormationInfo() {
...
+     * exception are transformed into a common type T with responseToResultFunction or exceptionToResultFunction, and then consumed by
+     * responseConsumer.
+     *
+     * @param masterEligibleNode        The master eligible node to be queried, or null if we do not yet know of a master eligible node
+     * @param responseConsumer          The consumer of the transformed response
+     * @param transportActionType       The ActionType for the transport action
+     * @param transportActionRequest    The ActionRequest to be sent
+     * @param exceptionToResultFunction A function that converts a remote exception to the response type expected by the responseConsumer
+     * @param responseToResultFunction  A function that converts a remote response to the response type expected by the responseConsumer
```
andreidan, line 930: Could these 2 functions be replaced by a `BiFunction<R, Exception, T> responseTransformer` and then we can call this `fetchRemoteObject` with a method reference (`transformResponse`?) that takes either a response R or an Exception? 

ie. 
```
transformResponse(@Nullable R response, @Nullable Exception e) { 
 assert r != null || e != null : "a response or an exception must be provided";  
 if( response != null) { 
    return new ClusterFormationStateOrException(response.getClusterFormationState());
   } else {
   return new ClusterFormationStateOrException(e);
  }
```

What do you think?

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -762,58 +792,228 @@ void cancelPollingClusterFormationInfo() {
...
+     * @param minSupportedVersion       The minimum version that the transport action works with
+     * @return A Cancellable for the task that is scheduled to fetch the remote information
+     */
+    private <R extends ActionResponse, T> Scheduler.Cancellable fetchRemoteObject(
```
andreidan, line 934: Is `Object` a bit too generic here? (and possibly overlapping the `java.lang.Object` concept?) 

Should we call it something along the lines of `sendNodeRequest` (as the parameters do the rest in documenting what it's actually fetching) or `fetchRemoteDiagnosticInfo` ?

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
andreidan: ❤️

andreidan (commented):
Thanks for iterating on this Keith.

This looks great ! Left a few more minor suggestions

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: I'll take this out of this PR and do that in another one since it's unrelated to this work.

andreidan: Sounds good, thanks Keith

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: That seems like it'll be a little less clear to read, but I'll do it and we can see.

masseyke:

andreidan (approved):
Thanks for iterating on this Keith

LGTM 🚀

Issue #56578, opened 2020-05-12
Disk decider prevents allocation/fast recovery


henningandersen:
If a node is above low disk threshold when being restarted (rolling restart, network disruption or crash), the disk threshold decider prevents reusing the shard content on the restarted node.

This seems unfortunate, in particular in the good case where we can do a noop recovery or an operations based recovery with only few operations.

Notice that being above the low disk threshold on a node is not a bad state in itself. The cluster may have plenty space available, but an imbalance w.r.t. disk usage can make some nodes go above low threshold anyway. We only start moving shards off the node when high threshold is reached.

The test case [here](https://github.com/elastic/elasticsearch/compare/master...henningandersen:demonstrate_low_threshold_prevents_recovery?expand=1#diff-a4e9aaf71ee63b4e5a9a852f1561abb5R416) demonstrates this. It also demonstrates that even having nodes with space available for the shard is not enough due to delayed allocation.

This might partially repair itself (not demonstrated) in that when enough shards have been recovered elsewhere, the node could drop below the low threshold, making the rest of the shard contents available for faster recoveries.

henningandersen:
We discussed this at our weekly sync and while we did not reach a conclusion, we did cover following:

* There were agreement that this is a problem today.
* The watermarks may exist primarily as a hysteresis when moving shards off nodes.
* We may want to consider adding a test framework for clusters in high disk usage situations.

Issue #49972, opened 2019-12-08
Restarting data node should not cause local shards unassigned when disk size above low-watermark


wangkhc:
**Elasticsearch version** (`bin/elasticsearch --version`):
6.3.2

**Description of the problem including expected versus actual behavior**:

We have disk low-watermark threshold settings which will prevent allocating new shard when disk size above 85% by default. However on node simply rebooting scenario, this setting still prevents local existing shards to recover back, because master think of these shards are "new" to this node, which is not true since these shard data still located on this node and not being deleted.

**Steps to reproduce**:
 1. Setup 3 node cluster with all disk all above 85% filling by index shards
 2. reboot one node

![image](https://user-images.githubusercontent.com/32295978/70392135-7e138580-1a17-11ea-8672-e95aa32d6a69.png)

DaveCTurner:
Currently the master cannot determine that the data on the node can actually be re-used, so there is a risk that it allocates an apparently-reusable shard to the node and only then discovers that it needs to make a full copy.

However it does seem that there is room for improvement in the (common) case that the shard data can be re-used. For instance, maybe we could allocate the shard on a "ops-based-recovery-or-nothing" basis, never performing a file-based recovery. Marking this as "help wanted".

wangkhc:
> Currently the master cannot determine that the data on the node can actually be re-used, so there is a risk that it allocates an apparently-reusable shard to the node and only then discovers that it needs to make a full copy.
> 
> However it does seem that there is room for improvement in the (common) case that the shard data can be re-used. For instance, maybe we could allocate the shard on a "ops-based-recovery-or-nothing" basis, never performing a file-based recovery. Marking this as "help wanted".

Seems that "BaseGatewayShardAllocator" already done this validation check in function "allocateUnassigned", but finally rejected by DiskThresholdDecider which checks low-watermark settings. 

Can I start working on this issue?

DaveCTurner:
Sure, a PR would be welcome.

Note that the checks in the gateway shard allocators are not the only concern here - the `RecoverySourceHandler` may independently decide to perform a file-based recovery, and this should be forbidden if there isn't enough disk space.

wangkhc:
> Sure, a PR would be welcome.
> 
> Note that the checks in the gateway shard allocators are not the only concern here - the `RecoverySourceHandler` may independently decide to perform a file-based recovery, and this should be forbidden if there isn't enough disk space.

[...]

PR #89018, opened 2022-08-02
[DOCS] Add warning on restarting nodes exceeding low disk watermark

Leaf-Lin:
As per https://github.com/elastic/elasticsearch/issues/49972 and https://github.com/elastic/elasticsearch/issues/56578, if a node is above low disk threshold when being restarted (rolling restart, network disruption or crash), the disk threshold decider prevents reusing the shard content on the restarted node.

The consequence of the event is the node may take a long time to start.

### DIFF
--- a/docs/reference/setup/restart-cluster.asciidoc
+++ b/docs/reference/setup/restart-cluster.asciidoc
@@ -8,6 +8,9 @@ nodes in the cluster while in the case of
 <<restart-cluster-rolling,rolling restart>>, you shut down only one node at a
 time, so the service remains uninterrupted.
 
+[WARNING]
+Nodes exceed low watermark will be slow to restart. You may want to reduce disk
+usage below low watermark before proceeding to restart nodes.
 
 [discrete]
 [[restart-cluster-full]]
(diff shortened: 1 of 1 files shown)


docs/reference/setup/restart-cluster.asciidoc
```diff
@@ -8,6 +8,9 @@ nodes in the cluster while in the case of
 <<restart-cluster-rolling,rolling restart>>, you shut down only one node at a
 time, so the service remains uninterrupted.
 
+[WARNING]
+Nodes exceed low watermark will be slow to restart. You may want to reduce disk
+usage below low watermark before proceeding to restart nodes.
```
lockewritesdocs, line 13:
```suggestion
[WARNING]
====
Nodes exceeding the low watermark threshold will be slow to restart. Reduce the disk
usage below the <<cluster-routing-watermark-low,low watermark>> before to restarting nodes.
====
```

lockewritesdocs: @Leaf-Lin, this wording is stronger than your original text, which I think coincides with a `WARNING` admonition. Let me know if this is a change that users should definitely make, or if we want to say that "You may want to" make this change.

lockewritesdocs (commented):
Left a suggested change with a question about the wording. Let me know what you think @Leaf-Lin!

lockewritesdocs:

lockewritesdocs (approved):
LGTM 🦖

PR #89021, opened 2022-08-02
Wrap enrich execute action in new tracing context

pugnascotia:
Part of #84369. Split out from #88443. This PR wraps parts logic in
`InternalExecutePolicyAction` in a new tracing context. This is
necessary so that a tracing implementation can use the thread context
to propagate tracing headers, but without the code attempting to set the
same key twice in the thread context, which is illegal.


andreidan (approved):
LGTM, thanks Rory

PR #81910, opened 2021-12-20
User Profile: Add API for get profile by uid

ywangd:
This PR adds a REST endpoint to retrieve profile by uid. It also takes
an optional `data` parameter to return all or part of the application data
long the profile. The `data` parameter works the same way as the
`filter_path` parameter.

PR #89023, opened 2022-08-02
User Profile - GetProfile API nows supports multiple UIDs

ywangd:
This PR expands the existing GetProfile API to support getting multiple
profiles by IDs. As a result, the response format is also changed to
align with the latest version of API design guideline. Concretely, this
means moving the profiles as an array inside a top level "profiles"
field so that (1) does not mix dynamic fields (uid) with static fields
and (2) enforcing an order in the response which is desirable for
clients.

The change also reports any error encounter in the retrieving process
in a top level "errors" field.

Relates: #81910


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
```diff
@@ -363,35 +380,44 @@ private void getVersionedDocuments(Collection<String> uids, ActionListener<Tuple
...
+                        final ResultsAndErrors<VersionedDocument> resultsAndErrors = new ResultsAndErrors<>(retrievedDocs, errors);
+                        if (logger.isDebugEnabled() && false == resultsAndErrors.errors().isEmpty()) {
+                            Exception loggedException = null;
+                            final List<String> errorUids = errorUidsExcludingNotFound(resultsAndErrors.errors());
+                            for (String uid : errorUids) {
+                                loggedException = ExceptionsHelper.useOrSuppress(loggedException, resultsAndErrors.errors().get(uid));
+                            }
+                            if (loggedException != null) {
+                                logger.debug(() -> format("Failed to retrieve profiles %s", errorUids), loggedException);
+                            }
```
ywangd, line 418: I have a change of heart on how errors are handled in multi-get. I am now in favor of reporting all errors in detail and categorize not found uids as error as well. This errorUids extraction is added to maintain the existing behaviour. But if you are happy with the new approach, I can raise a follow-up PR to have errors reported in details for the hasPrivileges check.

albertzaharovits: I don't think the logging as a whole makes sense anymore since all the `errors` are returned in the response now.

albertzaharovits (approved):
LGTM

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/profile/ProfileService.java
ywangd: Agreed. I'll remove it in the follow-up PR for adding the `errors` field in the profileHasPrivileges response.

ywangd:
@azasypkin I will be merging this PR after 24 hours. Please let me know if this is sufficient to Kibana to sort out the CI.

Ping @elastic/clients-team for changes to request query parameter and response format.

azasypkin:
> @azasypkin I will be merging this PR after 24 hours. Please let me know if this is sufficient to Kibana to sort out the CI.

Thanks for the ping! Yes, I think it should be enough for Kibana, created PR to adapt to these changes: https://github.com/elastic/kibana/pull/138361

azasypkin:
```json
{
  "profiles": [],
  "errors": {
    "count": 1,
    "details": {
       "this_id_does_not_exist": {
         "type": "resource_not_found_exception",
         "reason": "profile document not found"
       }
    }
  }
}
```

@ywangd how would the `errors` field look like if there are multiple errors?

ywangd:
@azasypkin The `details` object is keyed by profile uids that counter errors. In the case where there are multiple errors, you would have something like
```json
{
  "profiles": [],
  "errors": {
    "count": 2,
    "details": {
       "this_id_does_not_exist": {
         "type": "resource_not_found_exception",
         "reason": "profile document not found"
       },
       "another_id_does_not_exist": {
         "type": "resource_not_found_exception",
         "reason": "profile document not found"
       }
    }
  }
}
```

ywangd:
Ah, is it the usage of `this_id_does_not_exist` somehow confusing since it looks like an actual error message instead of an ID? Maybe we should change the example to use a string that looks more like an actual ID.

azasypkin:
> Ah, is it the usage of this_id_does_not_exist somehow confusing since it looks like an error message instead of an actual ID? 

Yeah, I didn't realize it was an id, but I haven't dealt with any ES API that can return multiple errors yet, so maybe that's just me.

ywangd:
@elasticmachine update branch

PR #89024, opened 2022-08-02
Wrap ML model loading task in new tracing context

pugnascotia:
Part of #84369.

ML uses the task framework to register a tasks for each loaded model.
These tasks are not executed in the usual sense, and it does not make
sense to trace them using APM. Therefore, make it possible to register
a task without also starting tracing.


dimitris-athanasiou (approved):
LGTM

PR #89027, opened 2022-08-02
REST tests and spec for bulk update API keys

n1v0lg:
This PR adds REST API spec and YAML test files for the BulkUpdateApiKey
operation.


x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -1645,6 +1645,90 @@ public void testBulkUpdateApiKeysWithDuplicates() throws ExecutionException, Int
         assertThat(response.getNoops(), empty());
     }
 
+    public void testBulkUpdateApiKeysWithDifferentLimitedByRoleDescriptorsForSameUser() throws ExecutionException, InterruptedException,
```
n1v0lg, line 1648: Sneaking in a test I only thought of while working on the YAML tests. It's quite lengthy and specific so I think it's best placed as in the integration test suite.

rest-api-spec/src/main/resources/rest-api-spec/api/security.bulk_update_api_keys.json
```diff
@@ -0,0 +1,32 @@
+{
+  "security.bulk_update_api_keys": {
```
n1v0lg, line 2: @elastic/es-clients I was hoping to get your take on whether to use `bulk_update_api_keys` here or `update_api_keys`. 

I'm using `_bulk` to be consistent with the REST route path `.../_bulk_update` as well as better readability, to distinguish between `bulk_update_api_keys` and the single operation `update_api_key`. 

`update_api_key` and `update_api_keys` are quite close, and easy to misread. Since we want to be consistent between docs and the REST spec, this also extends to which API doc header we use, i.e. "Bulk update API keys" vs "Update API keys" and the same readability concern exists there.

The downside of `_bulk` is that we don't do this for other APIs, such as invalidate API keys.

I don't feel strongly either way, and was curious on your thoughts!

cc @ywangd

sethmlarson: I'm good with using "bulk" on the API endpoint name.

Btw, the `@elastic/es-clients` handle isn't used anymore. We use `@elastic/clients-team` now.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/30_update.yml
```diff
@@ -346,3 +346,508 @@ teardown:
...
+            }
+          }
+  - length: { "noops": 0 }
+  - length: { "updated": 2 }
+  - match: {
+    "updated.0": "$key_id_1"
+  }
+  - match: {
+    "updated.1": "$key_id_2"
+  }
```
ywangd, line 477: Nit: we can have a `is_false` for the `errors` field.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/30_update.yml
```diff
@@ -346,3 +346,508 @@ teardown:
...
+            "index": [
+              {
+                "names": ["index-a"],
+                "privileges": ["read"]
+              }
+            ]
+          }
+  - match: { "has_all_requested": true }
+
+  # Check that metadata was updated
```
ywangd, line 689: This is probably copy/paste error: Metadata didn't get updated in this test case.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/30_update.yml
```diff
@@ -346,3 +346,508 @@ teardown:
...
+              "role-a": {
+                "cluster": ["all"],
+                "index": [
+                  {
+                    "names": ["index-a"],
+                    "privileges": ["read"]
+                  }
+                ]
+              }
+            },
```
ywangd, line 579: I slightly prefer to have empty role_descriptors when creating the API keys. This test case is for auto-update limited_by_role_descriptors. Skipping the key's own descriptors here ensure everything works off the owner's permissions (and update).

rest-api-spec/src/main/resources/rest-api-spec/api/security.bulk_update_api_keys.json
ywangd: I think we need a consistent strategy on using `bulk` (or not) moving forward. For example, if we introduce a new API in the future that does not have a counterpart for single operation, do we still want to have "bulk" in its name? Also, if we decide to deprecate (and evetually remove) a single operation API (maybe because it is entirely replaceable by the bulk version), this would leave a bulk-named API without the corresponding version of single operation. Do we care? @sethmlarson

ywangd (approved):
LGTM

The discussion around when and where (or whether) to use the "bulk" in names is unresolved. I'd like to gather more thoughts about it. But it does not have to stop this PR since we can revisit it later given we are very early in the release cycle.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/30_update.yml
n1v0lg: Absolutely, good catch!

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/api_key/30_update.yml
n1v0lg: Part of the intention for this test is to show that the existing `role_descriptors` field doesn't get overwritten. I can change creation to not assign descriptors but the test in its current state captures a pretty common scenario: given an API key with previously assigned roles, and an update to its owner's roles, bring it up to date with the owner's roles _without_ touching the assigned descriptors. I could make this intention more clear via a comment or in the test name. Or I could drop role descriptors, I don't feel strongly here. WDYT?

n1v0lg: Hmm on second thought, since we don't really assert on the role descriptor values themselves (only on the effective privs of the API key), the test isn't really showing that they don't change. So I'm fully on board with your suggestion 👍

n1v0lg:
> The discussion around when and where (or whether) to use the "bulk" in names is unresolved. I'd like to gather more thoughts about it.

@ywangd agreed. I will start a slack thread on the topic soon, when I get to work on the docs PR.

rest-api-spec/src/main/resources/rest-api-spec/api/security.bulk_update_api_keys.json
n1v0lg: > Btw, the @elastic/es-clients handle isn't used anymore. We use @elastic/clients-team now.

Roger that! 

@ywangd thanks for raising this. I'll start a slack thread so we can collect more input.

PR #89032, opened 2022-08-02
Preemptively initialize routing nodes and indices lookup on all node types

original-brownbear:
Follow up to #89005 running the initialization as soon as possible on non-master
nodes as well.

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/coordination/Coordinator.java
+++ b/server/src/main/java/org/elasticsearch/cluster/coordination/Coordinator.java
@@ -414,8 +414,14 @@ public class Coordinator extends AbstractLifecycleComponent implements ClusterSt
         assert publishRequest.getAcceptedState().nodes().getLocalNode().equals(getLocalNode())
             : publishRequest.getAcceptedState().nodes().getLocalNode() + " != " + getLocalNode();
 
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/elasticsearch/cluster/coordination/Coordinator.java
```diff
@@ -414,8 +414,14 @@ PublishWithJoinResponse handlePublishRequest(PublishRequest publishRequest) {
         assert publishRequest.getAcceptedState().nodes().getLocalNode().equals(getLocalNode())
             : publishRequest.getAcceptedState().nodes().getLocalNode() + " != " + getLocalNode();
 
+        final ClusterState newClusterState = publishRequest.getAcceptedState();
+        if (newClusterState.nodes().isLocalNodeElectedMaster() == false) {
```
original-brownbear, line 418: I wasn't really able to benchmark whether this special handling of master is helpful or not admittedly (though I can't see how it would ever hurt). But given that this code is running on the single threaded coordinator pool, I found it easier to reason about master performance by leaving the special case in place here.

original-brownbear:
Jenkins run elasticsearch-ci/part-1  (known)

DaveCTurner (approved):
LGTM

original-brownbear:
Thanks David!

PR #75306, opened 2021-07-13
Improve efficiency of BoundedBreakIteratorScanner fragmentation algorithm (the 2nd)

thelink2012:
See #73569 for the issue this tries to solve. See PR #73785 for an explanation of the algorithm.

The previous PR had to be reverted due to it causing tests to fail in Kibana (see elastic/kibana#104466).

One of such failures was caused by `targetEndOffset` being a `int` but having a possibility of storing something bigger than that (more precisely for the Kibana tests, `if (targetEndOffset + 1 > textEndIndex)` overflowed). Commit e2635198d023fa0ef4b4c7132909d4da19e45574 in this PR fixes that.

There were still some failing tests after that. Needs investigation.

Issue #73569, opened 2021-05-31
Unified Highlighter way too slow for fragment_size > 0


thelink2012:
**Elasticsearch version** (`bin/elasticsearch --version`): 7.9.0 (also tested on latest: 7.13.0)

**Plugins installed**: []

**JVM version** (`java -version`): 14.0.1

**OS version** (`uname -a` if on a Unix-like system): Linux 5.8.0-53-generic 20.04.1-Ubuntu

**Description of the problem including expected versus actual behavior**:

While investigating performance in our workload, we identified the unified highlighter as taking up to 95% of profiling samples for certain queries. Further investigation reveals [this loop](https://github.com/elastic/elasticsearch/blob/a92a647b9f17d1bddf5c707490a19482c273eda3/server/src/main/java/org/apache/lucene/search/uhighlight/BoundedBreakIteratorScanner.java#L95-L101) in `BoundedBreakIteratorScanner` as the culprit. 

This can be confirmed by passing `fragment_size: 0` to the highlighter such that it uses `BreakIterator.getSentenceInstance()` unwrapped. Service time goes down to an acceptable level.

Below we include a reproduction sample that (in my developer machine) can go from ~500ms to ~30ms by changing `fragment_size: 300` to `fragment_size: 0` in the highlighting of a single document.

**Steps to reproduce**:

A relatively big document is required for reproduction. You can find one [here](https://drive.google.com/file/d/1Tc_DVm1sVzeUPsTQBt1_PA_MojWwmp82/view) as `doc.json`. The attached archive also contains an automated reproduction script. Nevertheless, as follow are the manual steps:

```bash
# 1) Create the index
curl -H "Content-Type: application/json" -XPUT "http://localhost:9200/sample-index" --data '
{
  "mappings": {
    "properties": {
      "content": {
        "type": "text",
        "index_options": "offsets"
      }
    }
  }
}'

# 2) Populate with certain documents
curl -H "Content-Type: application/json" -XPOST "http://localhost:9200/sample-index/_doc" --data @doc.json
curl -XPOST "http://localhost:39200/sample-index/_refresh"

# 3) Highlight with fragment_size: 0 (should be fast)
curl -H "Content-Type: application/json" -XGET "http://localhost:9200/sample-index/_search?request_cache=false&_source=false" -d '
{
  "query": {
    "query_string": {
      "query": "ter pelo no ser para"
    }
  },
  "highlight": {
    "fields": {
      "content": {
[...]

PR #89041, opened 2022-08-02
Improve efficiency of BoundedBreakIteratorScanner fragmentation algorithm

thelink2012:
This is PR #75306 (a fix for PR #73785) but in an user-repository such that _Allow edits by maintainers_ is possible. As follows is a description of the original PR. See those other PRs for previous discussions.

> As discussed in #73569 the current implementation is too slow in certain scenarios.
> 
> The [inefficient part](https://github.com/elastic/elasticsearch/blob/a92a647b9f17d1bddf5c707490a19482c273eda3/server/src/main/java/org/apache/lucene/search/uhighlight/BoundedBreakIteratorScanner.java#L92-L101) of the code can be stated as the following problem:
> 
>> Given a text (`getText()`) and a position in this text (`offset`), find the sentence boundary before and after the offset, in such a way that the after boundary is maximal but respects `end boundary - start boundary < fragment size`.
>>
>> In case it's impossible to produce an after boundary that respects the said condition, use the nearest boundary following offset.
> 
> The current approach begins by finding the nearest preceding and following boundaries, and expands the following boundary greedily while it respects the problem restriction. This is fine asymptotically, but `BreakIterator` which is used to find each boundary is sometimes expensive \[[Lucene noticed too](https://github.com/apache/lucene/blob/d5d6dc079395c47cd6d12dcce3bcfdd2c7d9dc63/lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/LengthGoalBreakIterator.java#L202)\].
> 
> The new approach maximizes the after boundary by scanning for the last boundary preceding the position that would cause the condition to be violated (i.e. knowing `start boundary` and `offset`, how many characters are left before resulting length is `fragment size`). If this scan finds the start boundary, it means it's impossible to satisfy the problem restriction, and we get the first boundary following offset instead (or better, since we already scanned `[offset, targetEndOffset]`, start from `targetEndOffset + 1`).
> 
> In theory, both approaches should produce exactly the same outputs given the same `(text, offset, fragment_size)` tuple. But it doesn't. As far as my investigation went, `BreakIterator` doesn't seem to be commutative. Previous method calls affects other calls. So I guess whatever change we make to this algorithm, may produce results that differ from each other. It is rare, but happens. Would love to be wrong on this though.
> 
> Highlighting related unit and integration tests pass. Though I'm not passing in tests that seem completely unrelated to this. See below.


romseygeek:
@elasticmachine generate changelog

romseygeek:
@elasticmachine update branch

pugnascotia:
@elasticmachine generate changelog

romseygeek (approved):
LGTM, thanks @thelink2012

thelink2012:
Hey @romseygeek. Thanks for the merge. Would it be possible for this patch to be backported to previous versions (at least 8.4)?

javanna:
heya @thelink2012 we only backport bug-fixes to patch releases, and this does not qualify as one.

PR #89044, opened 2022-08-02
Add test in vector tiles with runtime fields

iverase:
In https://github.com/elastic/elasticsearch/pull/85449, we fixed an issue with geo_point runtime fields that prevent user to use vector tiles with geo_point runtime fields. ThisPR adds a test to make sure the functionality is no broken.


craigtaverner (approved):
LGTM

PR #89046, opened 2022-08-02
[8.4] Docs: Remove paragraph that applies only before Elasticsearch 7.0 (#86209)

nik9000:
Backports the following commits to 8.4:
 - Docs: Remove paragraph that applies only before Elasticsearch 7.0 (#86209)


nik9000:
@spinscale was an employee when he wrote this. It's fine. Ignore it.

PR #77700, opened 2021-09-14
Change YAML test structure from list to object

philkra:
This change converts the range query from an array to object. 

```
range": {
  "number": [
    {
      "gte": 4
    }
  ]
}
```

to

```
range": {
  "number": {
    "gte": 4
  }
}
```

PR #89049, opened 2022-08-02
[7.15] Change YAML test structure from list to object (#77700)

philkra:
Backports the following commits to 7.15:
 - Change YAML test structure from list to object (#77700)


nik9000:
This branch is quite a bit out of date. Closing. See main issue for if we backport this.

PR #89052, opened 2022-08-02
Add support for source fallback with the boolean field type

jdconrad:
This change adds a SourceValueFetcherSortedBooleanIndexFieldData to support boolean doc values for source fallback. Change is smaller than it looks as a good portion of the line count is testing.


romseygeek (approved):
LGTM

PR #89053, opened 2022-08-02
Add support for source fallback with scaled float field type

jdconrad:
This change adds source fallback support for scaled float. This uses the already existing class SourceValueFetcherSortedDoubleIndexFieldData.


romseygeek (approved):
LGTM

PR #89056, opened 2022-08-02
Add delayed allocation diagnosis case to shards availability indicator

jbaiera:
This PR adds diagnosis logic to the shards availability health indicator that detects when a shard allocation is delayed. This usually happens when a node that a shard is allocated to disappears. It is often better to delay the recovery of a shard in case the node that hosts it comes back. Shards that are delayed in this manner have special flags set on their unassigned info that denote a delayed allocation.

This change adds a diagnosis to the indicator that identifies these delayed shards and provides guidance stating that they will eventually allocate on their own once the delay elapses, but if allocation is required immediately, an index setting can be updated to perform the allocation.

This PR also includes some light integration testing to ensure that more unassigned cases are covered by the indicator.


server/src/test/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorServiceTests.java
```diff
@@ -386,7 +387,7 @@ public void testShouldBeYellowWhenRestartingReplicasReachedAllocationDelay() {
                             List.of(ImpactArea.SEARCH)
                         )
                     ),
-                    List.of(new Diagnosis(ACTION_CHECK_ALLOCATION_EXPLAIN_API, List.of("restarting-index")))
+                    List.of(new Diagnosis(DIAGNOSIS_WAIT_FOR_OR_FIX_DELAYED_SHARDS, List.of("restarting-index")))
```
jbaiera, line 390: I'm a little hazy on how the shutdown delays work. From my understanding the DELAYED_ALLOCATION status can only be reached from node shutdown logic. Is this status something that requires intervention to clear or will it still be cleared like the other delayed allocation situations?

server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
```diff
@@ -147,6 +147,18 @@ public HealthIndicatorResult calculate(boolean explain) {
...
 
+    public static final String FIX_DELAYED_SHARDS_GUIDE = "http://ela.st/fix-delayed-shard-allocation";
+    public static final Diagnosis.Definition DIAGNOSIS_WAIT_FOR_OR_FIX_DELAYED_SHARDS = new Diagnosis.Definition(
+        "delayed_shard_allocations",
+        "Elasticsearch is not allocating some shards because they are marked for delayed allocation. Shards that are on nodes that leave "
+            + "the cluster are usually marked for delayed allocation because it is more efficient to see if the node returns to the "
+            + "cluster than to recover the shard immediately.",
+        "Restore the nodes that have left the cluster so that the shards can be restored. If this is not possible, Elasticsearch will "
+            + "allocate the shards when the delay has elapsed. If the shards must be made available immediately, configure the "
+            + "[index.unassigned.node_left.delayed_timeout] setting on the affected indices to 0 to force the allocation immediately.",
```
andreidan, line 158: Shall we recommend waiting for the configured `index.unassigned.node_left.delayed_timeout` timeout to lapse? (we can probably include the actual value of the configured timeout in the message)  - the allocation `AllocateUnassignedDecision` should contain this already

Cloud doesn't have the concept of nodes.

x-pack/plugin/core/src/internalClusterTest/java/org/elasticsearch/xpack/cluster/routing/allocation/DataTierShardAvailabilityHealthIndicatorIT.java
```diff
@@ -0,0 +1,197 @@
...
+import static org.elasticsearch.test.NodeRoles.onlyRole;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasItem;
+
+/**
+ * Contains all integration tests for the {@link org.elasticsearch.cluster.routing.allocation.ShardsAvailabilityHealthIndicatorService}
+ * that require the data tiers allocation decider logic.
+ */
+@ESIntegTestCase.ClusterScope(scope = ESIntegTestCase.Scope.TEST, numDataNodes = 0)
+public class DataTierShardAvailabilityHealthIndicatorIT extends ESIntegTestCase {
```
andreidan, line 48: ❤️

server/src/test/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorServiceTests.java
andreidan: Not sure I understand your question. What would are the other delayed allocation situations we would encounter?

andreidan (commented):
Thanks for working on this Jimmy

Left a few questions and suggestions.

I'm curious also if we can reliably diagnose the delayed allocation situation without having all of the allocation deciders initialised? As, in the not-known-node-restart situation the algorithm seems to be - "if no node can host the shard and the index has a configured delayed timeout", however, can we precisely decide "no node can host the shard" given the initialised subset of deciders?

server/src/test/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorServiceTests.java
jbaiera: I'm not sure about what other delayed allocation situations we would encounter. I couldn't find any others except for the node shutdown.

My main question is whether this status is something that shards can get stuck in and thus need intervention to fix, or if once the delay has elapsed will the shard transition to another stage of recovery which may or may not succeed for other reasons. Based on what I'm reading, it seems like the answer is the latter of the two options, which would be great if that is the case.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
jbaiera: > Cloud doesn't have the concept of nodes.

Good point, not sure how to describe the problem without nodes in this case, since this really only ever happens when nodes leave. Maybe the diagnosis message is just something like "The availability of these shards is temporarily disrupted, we are waiting to see if they come back"?

Agreed on just saying wait until the delayed timeout has elapsed. Small snag on displaying the timeout though is if multiple indices have different values in them. The message could get quite verbose if we list each of them. Maybe something worth putting into the details?

server/src/test/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorServiceTests.java
andreidan: My understanding is that the latter is the case. Could we add an integration test for this? (ie. lapse the delay and get a different diagnosis for that same shard?)

server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
andreidan: >"The availability of these shards is temporarily disrupted, we are waiting to see if they come back"

That sounds good. Should we mention nodes as a second sentence? I feel like the users who are aware of nodes would appreciate that? 

> Maybe something worth putting into the details?

Good shout about the possible complexity. I'd say the details is a good section for this.

jbaiera:
I put up some changes to the PR to clean up the default case to always diagnose shards. Also updated the diagnosis text to remove mention of nodes, and to simply refer to shards becoming unavailable. Also removed the advice to update any index settings, since the best course of action is always to wait out the timer.

server/src/test/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorServiceTests.java
jbaiera: I'm not sure how I would go about building that integration test out unfortunately. If the system is designed to automatically attempt allocation after the delay, then validating that behavior seems like something out of scope for the health indicator to attempt simulating.

jbaiera: Unless we have a specific and reproducible edge case for this, I think we'd be ok with relying on the underlying system's invariants for now.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
```diff
@@ -413,9 +425,21 @@ List<Diagnosis.Definition> diagnoseUnassignedShardRouting(ShardRouting shardRout
...
+                }
+                break;
             case DECIDERS_NO:
                 actions.addAll(explainAllocationsAndDiagnoseDeciders(shardRouting, state));
                 break;
+            case DELAYED_ALLOCATION:
+                actions.add(DIAGNOSIS_WAIT_FOR_OR_FIX_DELAYED_SHARDS);
+                break;
+            case DECIDERS_THROTTLED:
+            case FETCHING_SHARD_DATA:
```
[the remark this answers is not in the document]
jbaiera: I cleaned this up

server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
jbaiera: I just removed all mention of nodes. The message is a little simpler now. I also decided against adding the delay to the details section. I doubt many people are setting this property to wildly different values, and reporting that it is 1 minute by default is probably not worth the effort when we can touch on it in the troubleshooting guide.

server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
```diff
@@ -413,10 +423,21 @@ List<Diagnosis.Definition> diagnoseUnassignedShardRouting(ShardRouting shardRout
...
+                }
+                break;
             case DECIDERS_NO:
                 actions.addAll(explainAllocationsAndDiagnoseDeciders(shardRouting, state));
                 break;
+            case DELAYED_ALLOCATION:
+                actions.add(DIAGNOSIS_WAIT_FOR_OR_FIX_DELAYED_SHARDS);
+                break;
             default:
+                actions.addAll(explainAllocationsAndDiagnoseDeciders(shardRouting, state));
```
andreidan, line 440: Do we need to expand the integration tests for this change? (or pull it into a separate PR and do it as part of that?)

andreidan (commented):
Thanks for updating the message Jimmy

Left one comment to possibly cover the expansion of the diagnosis in ITs

server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
jbaiera: Alright so I really tried to get a test case where this default case is exercised but I can't manage it. 

The only remaining `lastAllocationStatus` values possible are `FETCHING_SHARD_DATA` and `DECIDERS_THROTTLED`. After some digging, it looks like these are only used for throttled recoveries and for searchable snapshots that have not finished reporting their stats, neither of which we currently provide debugging information for. Additionally, these states are almost always transient, and attempting to lodge them into the state for extended periods (like creating an index while setting concurrent recoveries to 0) has the health API skipping the diagnosis logic anyway (treating indices as new and thus not important to report failures for yet, timely shutdowns, etc). I don't think we even have logic to detect problems pertaining to these unassigned shard cases right now. Should we remove the default here?

jbaiera:

jbaiera:

server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
andreidan: Thanks for looking into this Jimmy.

I played around with this quite a bit today and I think `NO_ATTEMPT` covers some of the cases I had in mind.

e.g.

```
    public void testIndexDisabledAllocation() throws Exception {
        internalCluster().startMasterOnlyNodes(1);
        internalCluster().startNodes(1, onlyRole(DiscoveryNodeRole.DATA_HOT_NODE_ROLE));
        ElasticsearchAssertions.assertAcked(
            prepareCreate("test").setSettings(
                Settings.builder()
                    .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1)
                    .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 0)
                    .put(DataTier.TIER_PREFERENCE, DataTier.DATA_HOT)
            )
        );
        ensureGreen("test");

        Settings.Builder settingsBuilder = Settings.builder()
            .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 22)
            .put(EnableAllocationDecider.INDEX_ROUTING_ALLOCATION_ENABLE_SETTING.getKey(), "none");
        UpdateSettingsRequestBuilder settingsRequest = client().admin().indices().prepareUpdateSettings("test");
        settingsRequest.setSettings(settingsBuilder);
        settingsRequest.setTimeout("1s").setMasterNodeTimeout("1s");
        settingsRequest.execute().actionGet();

        ensureYellow("test");

        GetHealthAction.Response healthResponse = client().execute(
            GetHealthAction.INSTANCE,
            new GetHealthAction.Request(ShardsAvailabilityHealthIndicatorService.NAME, true)
        ).get();
        HealthIndicatorResult indicatorResult = healthResponse.findIndicator(ShardsAvailabilityHealthIndicatorService.NAME);
        assertThat(indicatorResult.status(), equalTo(HealthStatus.YELLOW));
        assertThat(
            indicatorResult.diagnosisList(),
            hasItem(
                new Diagnosis(
                    ShardsAvailabilityHealthIndicatorService.ACTION_ENABLE_INDEX_ROUTING_ALLOCATION,
                    List.of("test")
                )
            )
        );
    }
```

>Should we remove the default here?

Yeah let's do that.

Toying around with the integratino tests I stumbled a bit when grepping for the class. Should we rename `DataTierShardAvailabilityHealthIndicatorIT` to `ShardAvailabilityHealthIndicatorIT` ? 
Or do you envision tests that don't have to do with tier-ing (like the one I posted above) to be in a future `ShardAvailabilityHealthIndicatorIT` ?

andreidan (approved):
LGTM thanks for working on this Jimmy

server/src/main/java/org/elasticsearch/cluster/routing/allocation/ShardsAvailabilityHealthIndicatorService.java
jbaiera: I think `DataTierShardAvailabilityHealthIndicatorIT` is appropriate for now since that test requires the data tiers plugin to be installed to operate. I had originally called it just `ShardAvailabilityHealthIndicatorIT` when I started building it in the server project, but since almost all of the test cases of the health indicator depend on data tiers code I moved it and renamed it thusly.

jbaiera:
@elasticmachine update branch

PR #88987, opened 2022-08-01
Reject unknown request body fields in Mount API

tlrx:
The parser used to parse Mount API requests is configured to ignore unknown fields. I suspect we made it this way when it was created because we were expecting to change the request's body in the future, but that never happened.

This leniency confuses users (#75982) so we think it is better to simply reject requests with unknown fields starting v8.5.0.

Because the High Level REST Client has a bug (to be fixed in #79604 - I'll take that one to completion) that injects a wrong `ignored_index_settings` we decided to just ignore and not reject that one on purpose.

Closes #75982

Issue #75982, opened 2021-08-02
Mount searchable snapshot API doesn't return error for `ignored_index_settings` request body parameter


jrodewig:
**Elasticsearch version** (`bin/elasticsearch --version`):
Latest `elasticsearch-8.0.0-SNAPSHOT-darwin-x86_64.tar.gz`
Version: 8.0.0-SNAPSHOT, Build: default/tar/ec588cb74a008c9c6babd4aa96cf333ce55eb4bf/2021-08-02T17:03:08.099052Z, JVM: 16.0.1

**Plugins installed**: []

**JVM version** (`java -version`):
OpenJDK Runtime Environment AdoptOpenJDK-16.0.1+9 (build 16.0.1+9)
OpenJDK 64-Bit Server VM AdoptOpenJDK-16.0.1+9 (build 16.0.1+9, mixed mode, sharing)

**OS version** (`uname -a` if on a Unix-like system):
Darwin Kernel Version 20.5.0: Sat May  8 05:10:33 PDT 2021; root:xnu-7195.121.3~9/RELEASE_X86_64 x86_64

**Description of the problem including expected versus actual behavior**:

The mount searchable snapshot API supports an `ignore_index_settings` parameter. However, if a user accidentally passes an `ignored_index_settings` (**ignored**, not **ignore**) parameter, the API doesn't return an error.

It looks like `ignored_index_settings` is a no-op option. Users may think specified index settings are ignored when they aren't.

Relates to #75786

**Steps to reproduce**:

1. Create an fs snapshot repo.
```
PUT _snapshot/my_repo
{
  "type": "fs",
  "settings": {
    "location": "my/path"
  }
}
```

2. Create a `my_docs` index with a custom `index.refresh_interval` setting.
```
PUT my_docs
{
  "settings" : {
    "index.number_of_shards" : 1,
    "index.number_of_replicas" : 0,
    "index.refresh_interval": "30s"
  }
}
```

3. Create a snapshot containing `my_docs`.
```
PUT _snapshot/my_repo/my_snapshot?wait_for_completion=true
{
  "include_global_state": false,
  "indices": "my_docs"
}
```

4. Delete the `my_docs` index.
```
DELETE my_docs
```

5. Mount `my_docs` as a searchable snapshot. Pass `index.refresh_interval` in `ignored_index_setting` (Note the misspelling).
```
POST _snapshot/my_repo/my_snapshot/_mount?wait_for_completion=true
{
  "index": "my_docs", 
[...]

PR #89061, opened 2022-08-03
[7.17] Fix mount snapshot api ignoring unknown request parameter

tlrx:
The High Level REST Client provides a `MountSnapshotRequest#ignoredIndexSettings(String[])` method to define some index settings to ignore when mounting a snapshot as a searchable snapshot index.

Sadly the client generates a wrong request body field `ignored_index_settings` instead of `ignore_index_settings`. This wasn't caught until #75982 was reported because the parser of Mount API request ignores unknown fields in request body. 

We fixed the wrong leniency of the request parser on the Elasticsearch side (#88987) starting version 8.5.0, and we decided to continue to ignore the `ignored_index_settings` generated by the HLRC bug to avoid breaking HLRC client usages.

This pull request fixes the request body generated by the HLRC to pass the correct field name. This fix is for versions 7.17.6+ (we do not expect to release new versions of HLRC in 8.x). 

It also renames the HLRC methods to expose the change to client users.

### DIFF
--- a/client/rest-high-level/src/main/java/org/elasticsearch/client/searchable_snapshots/MountSnapshotRequest.java
+++ b/client/rest-high-level/src/main/java/org/elasticsearch/client/searchable_snapshots/MountSnapshotRequest.java
@@ -115,19 +115,19 @@ public class MountSnapshotRequest implements ToXContentObject, Validatable {
         return this;
     }
 
-    private String[] ignoredIndexSettings;
+    private String[] ignoreIndexSettings;
 
-    public String[] getIgnoredIndexSettings() {
-        return ignoredIndexSettings;
+    public String[] getIgnoreIndexSettings() {
+        return ignoreIndexSettings;
     }
 
-    public MountSnapshotRequest ignoredIndexSettings(final String[] ignoredIndexSettings) {
-        if (ignoredIndexSettings != null) {
-            for (final String ignoredIndexSetting : ignoredIndexSettings) {
-                Objects.requireNonNull(ignoredIndexSetting);
+    public MountSnapshotRequest ignoreIndexSettings(final String[] ignoreIndexSettings) {
+        if (ignoreIndexSettings != null) {
+            for (final String ignoreIndexSetting : ignoreIndexSettings) {
+                Objects.requireNonNull(ignoreIndexSetting);
             }
         }
-        this.ignoredIndexSettings = ignoredIndexSettings;
+        this.ignoreIndexSettings = ignoreIndexSettings;
         return this;
     }
 
@@ -146,8 +146,8 @@ public class MountSnapshotRequest implements ToXContentObject, Validatable {
                 }
                 builder.endObject();
             }
-            if (ignoredIndexSettings != null) {
-                builder.array("ignored_index_settings", ignoredIndexSettings);
+            if (ignoreIndexSettings != null) {
+                builder.array("ignore_index_settings", ignoreIndexSettings);
             }
         }
         builder.endObject();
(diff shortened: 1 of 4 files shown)


client/rest-high-level/src/main/java/org/elasticsearch/client/searchable_snapshots/MountSnapshotRequest.java
```diff
@@ -115,19 +115,19 @@ public MountSnapshotRequest indexSettings(final Settings indexSettings) {
         return this;
     }
 
-    private String[] ignoredIndexSettings;
+    private String[] ignoreIndexSettings;
 
-    public String[] getIgnoredIndexSettings() {
-        return ignoredIndexSettings;
+    public String[] getIgnoreIndexSettings() {
```
henningandersen, line 120: I am not sure of our breaking change guarantees for the java code of HLRC. In particular here, where it goes into a patch release. I prefer doing the rename that you have here, but I wonder if we should leave the old method in place, with `@Deprecated` and a proper javadoc warning on it to be bwc. Perhaps @elastic/es-clients has input to this.

henningandersen (commented):
This looks good to me though I wonder about bwc.

client/rest-high-level/src/main/java/org/elasticsearch/client/searchable_snapshots/MountSnapshotRequest.java
sethmlarson: Just to confirm behavior, users previously were able to set `ignored_index_settings` but the parameter would have no effect. With this change the parameter will start having an effect?

If so, I think this rename and removal of the previous method is correct. Upgrading users should get an error if they were erroneously using a parameter that did nothing. This would prompt them to choose if they want the parameter to have an effect or if they actually do want the "nothing" case.

(Also for future pings we use the `@elastic/clients-team` handle now)

tlrx: > With this change the parameter will start having an effect?

Yes

I don't think we have strong breaking change guarantees on Java APIs, do we? I see this change as a bugfix and renaming the method is a way to notify that the previous behavior was buggy and requires to take a decision.

Keeping the previous method as deprecated avoids any compilation/runtime error but it does not look like fair to me to keep the bug around... If the previous method was used we can expect the user to interested into a fix for it? Also, the hassle on the breaking change in Java is easy to fix so breaking the API looks like the right thing to do (specially on a deprecated client and on a not so widely used API).

If that's OK I'll continue with the removal of the old method.

swallez: FWIW, the breaking change policy in the context of clients is somewhat less strict than that in ES. Here's the one we wrote for [the new Java API client](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/current/breaking-changes-policy.html).

Long story short, causing a compilation error where the buggy method is used is fine. As for it happening in a patch releases, we have to consider that 7.17 is the LTS version of 7.x and patch releases happen at the same time as minor releases in 8.x.

henningandersen: Thanks for clarifying and for the link to the breaking changes policy. I am good with this change too then.

henningandersen (approved):
LGTM.

Just to ensure it is not missed, I'd still like the verification mentioned [here](https://github.com/elastic/elasticsearch/pull/89061#discussion_r936521399).

client/rest-high-level/src/test/java/org/elasticsearch/client/documentation/SearchableSnapshotsDocumentationIT.java
```diff
@@ -87,7 +87,7 @@ public void testMountSnapshot() throws IOException, InterruptedException {
             .put("index.number_of_replicas", 0)
             .build();
         request.indexSettings(indexSettings); // <8>
-        request.ignoredIndexSettings(
+        request.ignoreIndexSettings(
```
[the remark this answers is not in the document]
tlrx: Sure, I added ea0012af5c843b739ee0fe1664b61d6f85849dcc

PR #88943, opened 2022-07-29
[ML] Frequent Items: use a bitset for deduplication

hendrikmuhs:
By using bitsets instead of lists of longs item sets can be faster de-duplicated. A bit is set according to the order of top items (by count).

Screenshot_20220729_153822

Notes:
 - the bitset might be useful for transactions and can speedup the lookup to find out if a candidate set matches the transaction
 - bitsets reduce memory requirements(memory for remembering collected sets)

PR #89062, opened 2022-08-03
[ML] Optimize frequent items transaction lookup

hendrikmuhs:
represent transactions as bitsets for faster lookups when iterating over candidate sets. This PR implements a lookup table and
a subset check based on bits. It uses this lookup table to map transactions to items, this so-called horizontal representation is used to speedup the lookup that checks if a transaction contains the candidate item set

Screenshot_20220803_105849

Follow up:

The lookup table is kept generic to explore other usages. The next step is to explore if a vertical representation - a lookup table that maps items to transactions - executes faster (a vertical representation is used by fp-growth and other eclat implementations):

- requires some rewrite of the lookup code in `CountingItemSetTraverser`
  - might make the cache there unnecessary (the vertical representation is basically a cache already)
- requires further bit logic (intersection) in the lookup table
- explore usage of skipping via `nextClearBit`/`nextSetBit` to shortcut some loops

relates #88943

### DIFF
--- a/docs/changelog/89062.yaml
+++ b/docs/changelog/89062.yaml
@@ -1,0 +1,5 @@
+pr: 89062
+summary: Optimize frequent items transaction lookup
+area: Machine Learning
+type: enhancement
+issues: []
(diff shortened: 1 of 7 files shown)


droberts195 (approved):
LGTM

PR #89083, opened 2022-08-03
Improve SAML warning message when signing credentials cannot be found

slobodanadamovic:
This commit attempts to provide a better hint in case when the signature
cannot be validated because we can't find any signing key
in the configured metadata file.


slobodanadamovic:

slobodanadamovic:
@elasticmachine update branch

tvernum (approved):
LGTM

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/saml/SamlObjectHandler.java
```diff
@@ -228,25 +229,29 @@ protected void checkIdpSignature(CheckedFunction<Credential, Boolean, Exception>
...
+                "The XML Signature of this SAML message cannot be validated "
+                    + "because signing credentials could not be found in SAML metadata. "
+                    + "Please verify that the saml realm uses the correct and syntactically valid SAML "
+                    + "metadata file/URL for this Identity Provider."
+            );
+        } else {
+            logger.warn(
+                "The XML Signature of this SAML message cannot be validated. Please verify that the saml realm uses the correct SAML "
+                    + "metadata file/URL for this Identity Provider."
+            );
```
ywangd, line 243: I think we should maintain the existing text since it is still correct in this failure scenario and appending some extra message at the end for this new failure. Because we have a [doc page](https://www.elastic.co/guide/en/elasticsearch/reference/current/trb-security-saml.html) referring to this message (Symptom 8) and it would be useful to have them maintained. It may also be worthwhile to add another sub-bullet point to Symptom 8 for this new scenario.

slobodanadamovic: Good point. I wasn't aware of the docs. I agree that the message is still correct. The intention was to hint about another possible issue with metadata file. I can revert the change to simply append additional details about missing credentials and update docs for Symptom 8.

Generally, I'd prefer to have these symptoms distinguished in the code and to be able to provide a more specific error message so users do not have to lookup in the docs. But unfortunately, in this case, it's not possible.

Thanks for pointing docs out.

Issue #81223, opened 2021-12-01
Field data types docs - Consistency improvements 


mattkime:
- Most of the numeric types are found at https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html but `unsigned long` has its own page with some content not found on the number page - https://www.elastic.co/guide/en/elasticsearch/reference/current/unsigned-long.html
- `murmur3` is shown along side built in types in https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html but is available via a plugin. It should be called out in a separate section to make it more distinct. The same for `mapper-annotated-text`
- `completion` is shown on https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-types.html but doesn't appear in the sidebar list of types on the same page.

PR #89086, opened 2022-08-03
[DOCS] Move completion type asciidoc

abdonpijpelink:
Related to #81223. This PR extracts the `completion` field type discussion into its own file, in the same location as the other field type docs. To ensure the original page stays mostly unchanged, the new file is included in the same position as where the original discussion was.


lockewritesdocs (approved):
LGTM 🦖

lockewritesdocs:
I added the `8.3.3` label as well because that's our current docs version.

PR #89089, opened 2022-08-03
Use threadpool instead of mock in some Repository tests

pxsalehi:
Some test changes needed for https://github.com/elastic/elasticsearch/pull/88209. Just to keep the other PR light!

Using mocks in this tests causes NPEs on https://github.com/elastic/elasticsearch/pull/88209. Instead of more mocking, I'd rather use an actual threadpool.

### DIFF
--- a/server/src/test/java/org/elasticsearch/repositories/RepositoriesServiceTests.java
+++ b/server/src/test/java/org/elasticsearch/repositories/RepositoriesServiceTests.java
@@ -30,7 +30,6 @@ import org.elasticsearch.common.component.Lifecycle;
 import org.elasticsearch.common.component.LifecycleListener;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.common.util.MockBigArrays;
-import org.elasticsearch.common.util.concurrent.ThreadContext;
 import org.elasticsearch.index.shard.ShardId;
 import org.elasticsearch.index.snapshots.IndexShardSnapshotStatus;
 import org.elasticsearch.index.store.Store;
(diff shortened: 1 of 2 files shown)


kingherc:
Thanks for adding me :100: 

> Using mocks in this tests causes NPEs on https://github.com/elastic/elasticsearch/pull/88209.

Any idea why? Would improving the mocks behavior somehow solve it, rather than using actual thread pools (so it's closer to a unit testing rather than integration testing)?

But LGTM :+1:

kingherc (approved):

DaveCTurner (commented):
Seems reasonable, but this does make these tests multithreaded (potentially nondeterministic) now, and also a full `ThreadPool` is a pretty heavyweight thing.

 Does it work instead to create a `new DeterministicTaskQueue()` and then use its `getThreadPool()` method to get a deterministic single-threaded threadpool instead? That might require some test changes since you have to tell it to run the enqueued tasks at specific points.

DaveCTurner (changes requested):
Going to call my previous comment "request changes" for now.

kingherc:
Ah didn't think so far @DaveCTurner , nice!

DaveCTurner:
Alternatively you could make a non-forking non-deferring threadpool mock like in `BlobStoreTestUtil#mockClusterService`. Or even a real object that does the same thing, if you think it's useful.

pxsalehi:
The threadpool is not used at all in those tests, so there are no behaviour changes. But it definitely is too heavy just for that. What is happening in `BlobStoreTestUtil#mockClusterService` is anyway a bit more mocking. I'll do that in my PR as well. I'll close this PR then. Thanks!

PR #89091, opened 2022-08-03
[ML] Include start params in _stats for non-started model deployments

dimitris-athanasiou:
Adds the missing start parameters to the _stats API response
for non-started deployments.


droberts195 (approved):
LGTM

PR #89094, opened 2022-08-03
Deduplicate fetching doc-values fields

dnhatn:
If a docvalues field matches multiple field patterns, then ES will return the value of that doc-values field multiple times. Like fetching fields from source, we should deduplicate the matching doc-values fields.

This bug can reproduce with these two steps:

```js
POST /test_index/_doc/1
{
  "event.error.code": 404
}
```

```js
GET test_index/_search
{
  "docvalue_fields": [ "event.*", "*error*" ]
}
```

The value of `event.error.code` is returned twice because the field matches `event.*` and `*error*`.

```json
{
  "hits": [
    {
      "_index": "test_index",
      "_id": "1",
      "_score": 1,
      "_source": {
        "event.error.code": 404
      },
      "fields": {
        "event.error.code": [
          404,
          404
        ]
      }
    }
  ]
}
```

### DIFF
--- a/rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search/240_date_nanos.yml
+++ b/rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search/240_date_nanos.yml
@@ -91,8 +91,40 @@ setup:
           docvalue_fields:
             - field: date
               format: strict_date_optional_time
+          sort:
+            - date: desc
+
+  - match: { hits.total: 2 }
+  - length: { hits.hits: 2 }
+  - match: { hits.hits.0._id: "date_ns_1" }
+  - match: { hits.hits.1._id: "date_ms_1" }
+  - match: { hits.hits.0.fields.date: [ "2018-10-29T12:12:12.123Z"] }
+  - match: { hits.hits.1.fields.date: [ "2018-10-29T12:12:12.987Z"] }
+
+  - do:
+      search:
+        rest_total_hits_as_int: true
+        index: date*
+        body:
+          docvalue_fields:
             - field: date
               format: epoch_millis
+          sort:
+            - date: desc
+
+  - match: { hits.total: 2 }
+  - length: { hits.hits: 2 }
+  - match: { hits.hits.0._id: "date_ns_1" }
+  - match: { hits.hits.1._id: "date_ms_1" }
+  - match: { hits.hits.0.fields.date: [ "1540815132123.456789" ] }
+  - match: { hits.hits.1.fields.date: [ "1540815132987" ] }
+
+  - do:
+      search:
+        rest_total_hits_as_int: true
+        index: date*
+        body:
+          docvalue_fields:
             - field: date
               format: uuuu-MM-dd'T'HH:mm:ss.SSSSSSSSSX
           sort:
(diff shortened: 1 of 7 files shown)


rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search/240_date_nanos.yml
```diff
@@ -91,8 +91,40 @@ setup:
           docvalue_fields:
             - field: date
```
dnhatn, line 92: I consider fetching DV fields multiple times with different formats a misuse, and we don't need to keep BWC for this.

jtibshirani:
It's also confusing to me why we support this! I was wondering where you got the idea for the change -- was there a bug report, or some user found this confusing? The reason I ask is that it's a breaking behavior change. So I'm not sure we could do it in a minor or patch version. Since we are moving away from fetching using `docvalue_fields` (in favor of the `fields` option), it may be not be worth the disruption.

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search/240_date_nanos.yml
jtibshirani: 👍 makes sense to me too

server/src/main/java/org/elasticsearch/search/fetch/subphase/FetchDocValuesContext.java
```diff
@@ -23,20 +24,26 @@
...
     public FetchDocValuesContext(SearchExecutionContext searchExecutionContext, List<FieldAndFormat> fieldPatterns) {
+        // Use Linked HashMap to reserve the fetching order
+        final Map<String, FieldAndFormat> fieldToFormats = new LinkedHashMap<>();
         for (FieldAndFormat field : fieldPatterns) {
             Collection<String> fieldNames = searchExecutionContext.getMatchingFieldNames(field.field);
             for (String fieldName : fieldNames) {
-                fields.add(new FieldAndFormat(fieldName, field.format, field.includeUnmapped));
+                final FieldAndFormat format = new FieldAndFormat(fieldName, field.format, field.includeUnmapped);
+                final FieldAndFormat existing = fieldToFormats.put(fieldName, format);
+                if (existing != null && existing.equals(format) == false) {
```
jtibshirani, line 41: I wonder if it'd be better to just have the first definition take precedence, instead of throwing an error. Being flexible here seems less disruptive, and it matches what we do for the `fields` option. I also don't think this behavior is too "trappy" for users.

jtibshirani (commented):
Thanks @dnhatn, this looks good to me. I left one comment about whether we should throw an error.

For context: we discussed my question about whether it's too 'breaking' offline and agreed it should be treated as a bug. But to minimize disruption we'll only merge to 8.5, and not backport to any patch releases.

server/src/main/java/org/elasticsearch/search/fetch/subphase/FetchDocValuesContext.java
dnhatn: I pushed https://github.com/elastic/elasticsearch/pull/89094/commits/63a3f25be1822dfd68a648abc3fa4020dcdd18c9 to remove the restriction.

server/src/main/java/org/elasticsearch/search/fetch/subphase/FetchDocValuesContext.java
```diff
@@ -23,20 +24,23 @@
...
      */
     public FetchDocValuesContext(SearchExecutionContext searchExecutionContext, List<FieldAndFormat> fieldPatterns) {
+        // Use Linked HashMap to reserve the fetching order
+        final Map<String, FieldAndFormat> fieldToFormats = new LinkedHashMap<>();
         for (FieldAndFormat field : fieldPatterns) {
             Collection<String> fieldNames = searchExecutionContext.getMatchingFieldNames(field.field);
             for (String fieldName : fieldNames) {
-                fields.add(new FieldAndFormat(fieldName, field.format, field.includeUnmapped));
+                // the first matching field wins
+                fieldToFormats.putIfAbsent(fieldName, new FieldAndFormat(fieldName, field.format, field.includeUnmapped));
```
jtibshirani, line 40: I'm sorry, I remembered incorrectly what the `fields` option does. It always uses `put` on a `LinkedHashMap`, which means the last definition wins. I think it's worth doing the same for consistency/ simplicity.

dnhatn: ok, I pushed https://github.com/elastic/elasticsearch/pull/89094/commits/e1342509564f7499510980736627923b33e53c18

jtibshirani (approved):

dnhatn:
Thanks Julie.

PR #89050, opened 2022-08-02
Mention internal user `_async_search` in docs

justincr-elastic:
This PR will add internal user `_async_search` to the 7.x and 8.x docs.

Internal user `_async_search` was added in v7.7 by #52141.

Doc previews are available for a short time after a build
- https://elasticsearch_89050.docs-preview.app.elstc.co/diff

PR #89099, opened 2022-08-03
[8.3] Mention `_async_search` internal user which was added in 7.7 (#89050)

justincr-elastic:
 - Mention `_async_search` internal user which was added in 7.7 (#89050)


justincr-elastic:

PR #89106, opened 2022-08-04
SQL: Set track_total_hits to false when not needed

### LINKED ISSUE #88764: _sql should set track_total_hits=false when applicable
### Description

Tracking total hits sometimes has a non-negligible impact on performance. We should disable it when not strictly required.

luigidellaquila:
`track_total_hits` has some performance cost, that in many cases can be avoided.

This PR sets track_total_hits=false in all cases where the execution plan did not explicitly declare that track_total_hits is needed.

Fixes #88764


x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/execution/search/Querier.java
```diff
@@ -535,14 +536,14 @@ protected List<BucketExtractor> initBucketExtractors(SearchResponse response) {
             List<QueryContainer.FieldInfo> refs = query.fields();
 
             List<BucketExtractor> exts = new ArrayList<>(refs.size());
-            ConstantExtractor totalCount = new ConstantExtractor(response.getHits().getTotalHits().value);
+            Holder<ConstantExtractor> totalCountHolder = new Holder<>();
```
astefan, line 539: Can you, please, explain the Holder usage here?

astefan (commented):
It's good that we try to improve all the edge cases. Few comments though:
- Why using a Holder for `ConstantExtractor`? I don't see the benefit...
- Have you also looked at EQL queries regarding `track_total_hits`?

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/execution/search/Querier.java
luigidellaquila: The problem here is `response.getHits().getTotalHits().value`, that cannot be executed in case of `track_total_hits=false` (it would throw an error).
Using the holder I just postpone its execution until it's really needed, preserving the original logic (ie. instantiating only one `ConstantExtractor` and using it multiple times if needed). Though, it's really unlikely that we need more than one `ConstantExtractor`, and even in that case, having multiple instances should not be a problem.
So IMHO it's just a matter of preference, I can leave it as it is or I can remove the holder completely and instantiate a new `ConstantExtractor` every time it's needed.

luigidellaquila: About EQL, I didn't check it yet. Since the two codebases do not seem to overlap on this fix, I'll provide a separate PR if relevant.

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/execution/search/Querier.java
```diff
@@ -560,6 +562,9 @@ private BucketExtractor createExtractor(FieldExtraction ref, BucketExtractor tot
             }
 
             if (ref == GlobalCountRef.INSTANCE) {
+                if (totalCount == null) {
```
costin, line 565: Not sure this is safety check however right now it looks as if we're not sure if this change will not backfire.

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/execution/search/Querier.java
```diff
@@ -535,14 +536,15 @@ protected List<BucketExtractor> initBucketExtractors(SearchResponse response) {
             List<QueryContainer.FieldInfo> refs = query.fields();
 
             List<BucketExtractor> exts = new ArrayList<>(refs.size());
-            ConstantExtractor totalCount = new ConstantExtractor(response.getHits().getTotalHits().value);
+            TotalHits totalHits = response.getHits().getTotalHits();
+            ConstantExtractor totalCount = totalHits == null ? null : new ConstantExtractor(totalHits.value);
```
costin, line 540: This is trappy behavior since now the null can leak.
Just extend ConstantExtractor (or BucketExtractor) with a dedicated TotalHits implementation that handles the null case and just pas totalHits object to it and handle it there by getting the concrete value or -1.
This would have minimal changes to the code.

costin (changes requested):
The null handling is trappy - see my comment.
I'd be curious to see what impact this will have on the benchmarks; looking at the code we only enable total hit counts when needed so by default it should be disabled.
Meaning we wouldn't see much of a performance benefit however handling the totalHits count is a good idea.

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/execution/search/Querier.java
luigidellaquila: This is probably an excess of defensive programming, looking at the code this case should never happen.
Anyway, with the suggested changes this check can be removed.

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/execution/search/Querier.java
luigidellaquila: > looking at the code we only enable total hit counts when needed so by default it should be disabled.

The Java API is misleading here, `track_total_hits` is actually a number, by default it's 10.000; I'm also curious to see the impact on the benchmarks.

> Just extend ConstantExtractor (or BucketExtractor) with a dedicated TotalHits implementation that handles the null case and just pas totalHits object to it and handle it there by getting the concrete value or -1.

Do we really need to extend `ConstantExtractor` for this? If we just want to return a -1, we can just pass it to `ConstantExtractor` itself.

bpintea: My understanding is that this -1 should never be returned (and that's why Andrei's exception throwing proposal [above](https://github.com/elastic/elasticsearch/pull/89106#discussion_r941201104), I assume), since the `GlobalCountRef.INSTANCE` ref is only used when accurate counting is required.
Don't know if we should go this way, but if we do add a TotalHitsExtractor, it might make sense to keep throwing an exception the exception on requiring the count when not actually needed.
If we don't go that way, I'd suggest replacing the `-1` with `org.elasticsearch.search.internal.SearchContext.TRACK_TOTAL_HITS_DISABLED`.

luigidellaquila: Any value we pass here, will always be a wrong value. 
About `org.elasticsearch.search.internal.SearchContext.TRACK_TOTAL_HITS_DISABLED`, that's the config value to disable the total hits, while here we are returning a calculated value, so IMHO they do not really represent the same thing.
My initial intention was to keep failing if something is wrong in the planner (so the Holder, that delayed the calculation of the totalHits at the moment when it was actually required by the execution plan). This could save us from problems introduced by future changes; on the other hand, passing a default value (eg. -1) could hide the consequences of a wrong planning. Anyway, for current implementation I don't see a possible path where this could really happen.

I'd say let's just reach an agreement, the impact of this choice should not be dramatic.

x-pack/plugin/sql/src/main/java/org/elasticsearch/xpack/sql/execution/search/Querier.java
```diff
@@ -535,14 +536,15 @@ protected List<BucketExtractor> initBucketExtractors(SearchResponse response) {
             List<QueryContainer.FieldInfo> refs = query.fields();
 
             List<BucketExtractor> exts = new ArrayList<>(refs.size());
-            ConstantExtractor totalCount = new ConstantExtractor(response.getHits().getTotalHits().value);
+            TotalHits totalHits = response.getHits().getTotalHits();
+            ConstantExtractor totalCount = new ConstantExtractor(totalHits == null ? -1 : totalHits.value);
```
costin, line 540: Since -1 is not a valid value (despite showing up in the queryDSL), +1 from to encapsulate this into a dedicated ConstantExtractor-like that throws an exception if used.
This keeps the code clean and passes the circuitbreaker around which will trip instead of returning -1.

Since the topic has already been discussed, I'm approving the ticket from now before the changes are applied.

costin (approved):

luigidellaquila:
@elasticmachine update branch

x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/execution/search/extractor/TotalHitsExtractor.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public Object extract(SearchHit hit) {
+        return validate(super.extract(hit));
+    }
+
+    private Object validate(Object value) {
+        if (Number.class.isInstance(value) == false) {
+            throw new QlIllegalArgumentException("Inconsistent total hits count handling, expected a numeric value but found {}", value);
```
bpintea, line 39: If we'd somehow get a string representation of a number this message might not be very elucidating, wondering if we shouldn't output the class.

bpintea (approved):
Lgtm

x-pack/plugin/ql/src/main/java/org/elasticsearch/xpack/ql/execution/search/extractor/TotalHitsExtractor.java
luigidellaquila: Good point, maybe something like

```
            throw new QlIllegalArgumentException("Inconsistent total hits count handling, expected a numeric value but found a {}", value == null ? null : value.getClass().getSimpleName());

```

Luegg (approved):

astefan (approved):
LGTM

jpountz:
I added the release highlight label based on the speedup we observed in nightly benchmarks with this change: https://elasticsearch-benchmarks.elastic.co/index.html#tracks/sql/nightly/default/30d

Issue #41299, opened 2019-04-17
Validation of analysis settings


romseygeek:
We have a few open issues to do with analysis settings validation (eg #16451, #34604).  Adding some kind of general validation framework for analysis settings would be a huge improvement in terms of usability, and generally be in line with our efforts to be less lenient in what we accept.

We've discussed in the past whether or not we should remove analysis configuration from index settings and make it a part of the mapping definition.  This would get us XContent validation for free, but it's a substantial change that has implications for other parts of the system, such as index templates.

A simpler way forward would be to add an `acceptedSettings` parameter to the constructors of `AbstractTokenFilterFactory`, `AbstractCharFilterFactory` and `AbstractTokenizerFactory`, taking a list of settings that we expect to see, and throwing an error if the passed-in `Settings` object has any unrecognised keys.

cbuescher:
Adding another case for the kind of validation we should do because it leads to otherwise confusing behaviour comes from #50356. e.g. for

```
PUT test
{
  "settings": {
    "analysis": {
      "filter": {
        "min-length-token": {
          "type": "length",
          "min": 2
        }
      },
      "analyzer": {
        "my_analyzer": {
          "filter": [
            "min-length-token"
          ],
          "type" : "standard"
        }
      }
    }
  }
}

# a is returned despite being shorter than length of two
GET test/_analyze
{
  "analyzer": "my_analyzer",
  "text": "A"
}
```

The `filter` part of the `standard` analyzer section is ignored, leading to often undetected errors.

benwtrent:
The plan here seems to me:

 - Add this new ctor to these classes accepting the known settings
 - Deprecate the old ctor (for plugins, etc.)
 - Deprecate using any settings that are unknown in new indices
 - Set a timeline for removing support on NEW indices for these settings

We should NOT:
 - remove without deprecation path (can seriously break templates, plugins, etc.)
 - remove support on old indices (breaks restoration)

PR #89112, opened 2022-08-04
Custom analyzer to reject  unknown parameters

### LINKED ISSUE #85710: Create analyzer accepting any random param and returning that in response as well
### Elasticsearch Version

8.1.1

### Installed Plugins

None

### Java Version

_bundled_

### OS Version

macos

### Problem Description

Create a custom analyzer using below request, **Note `foo` as an param to analyzer with `bar` value.**

```
PUT customanlyzerindex
{
    "settings": {
        "analysis": {
            "analyzer": {
                "my_analyzer": {
                    "type": "custom",
                    "tokenizer": "standard",
                    "foo": "bar"
                }
            }
        }
    }
}
```

Index is created successfully, and `foo` param is returned in the _settings API response.

GET customanlyzerindex/_settings

```
{
    "customanz": {
        "settings": {
            "index": {
                "routing": {
                    "allocation": {
                        "include": {
                            "_tier_preference": "data_content"
                        }
                    }
                },
                "number_of_shards": "1",
                "provided_name": "customanz",
                "creation_date": "1649229792216",
                "analysis": {
                    "analyzer": {
                        "my_analyzer": {
                            "type": "custom",
                            "foo": "bar",
                            "tokenizer": "standard"
                        }
                    }
                },
                "number_of_replicas": "1",
                "uuid": "zo-99vRVQD6igTzbyFMNKA",
                "version": {
                    "created": "8010199"
                }
            }
        }
    }
}
```

### Steps to Reproduce

Same as problem Description 

### Logs (if relevant)

N/A

likzn:
# Main Change
Nowtime, when we create custom analyzer with some not required params("foo":"nar"), it will succeed. This PR will make check with these parms to ensure these were our need.

Close: #85710

```
PUT customanlyzerindex
{
    "settings": {
        "analysis": {
            "analyzer": {
                "my_analyzer": {
                    "type": "custom",
                    "tokenizer": "standard",
                    "foo": "bar"
                }
            }
        }
    }
}
```


likzn:

likzn:
@javanna @romseygeek @jtibshirani Hi, PTAL  if free~

likzn:
Hi, can someone take a look about it

likzn:
@javanna Hi, PTAL

server/src/main/java/org/elasticsearch/index/analysis/AnalyzerComponents.java
```diff
@@ -44,6 +44,12 @@ static AnalyzerComponents createComponents(
         final Map<String, CharFilterFactory> charFilters,
         final Map<String, TokenFilterFactory> tokenFilters
     ) {
+        for (String key : analyzerSettings.keySet()) {
+            switch (key) {
+                case "tokenizer", "char_filter", "filter", "type", "position_increment_gap" -> {}
+                default -> throw new IllegalArgumentException("Custom Analyzer not support [" + key + "] now");
```
javanna, line 50: could you rephrase the error to `"Custom analyzer [" + name + "] does not support [" + key + "]"` ?

modules/analysis-common/src/yamlRestTest/resources/rest-api-spec/test/search.query/10_match.yml
```diff
@@ -16,7 +16,6 @@
                   tokenizer: standard
                   filter: [lowercase]
                 search:
-                  rest_total_hits_as_int: true
```
javanna, line 19: why was this change necessary?

modules/analysis-common/src/yamlRestTest/resources/rest-api-spec/test/indices.analyze/10_analyze.yml
```diff
@@ -59,3 +59,22 @@
...
+            analysis:
+              analyzer:
+                my_analyzer:
+                  type: custom
+                  tokenizer: standard
+                  foo: bar
+
+  - match: { status: 400 }
+  - match: { error.type: illegal_argument_exception }
+  - match: { error.reason: "Custom Analyzer not support [foo] now" }
```
javanna, line 80: ++ thanks for adding this test!

server/src/main/java/org/elasticsearch/index/analysis/AnalyzerComponents.java
```diff
@@ -44,6 +44,12 @@ static AnalyzerComponents createComponents(
         final Map<String, CharFilterFactory> charFilters,
         final Map<String, TokenFilterFactory> tokenFilters
     ) {
+        for (String key : analyzerSettings.keySet()) {
+            switch (key) {
+                case "tokenizer", "char_filter", "filter", "type", "position_increment_gap" -> {}
```
javanna, line 49: another way of doing this, without deduplicating the expected analyzer keys, would be to duplicate the settings into a mutable map, which we'd remove from every time we read a certain setting. At the end, if any item is left in the map it means that some unsupported param has been provided.

server/src/test/java/org/elasticsearch/index/analysis/AnalysisTests.java
```diff
@@ -103,4 +116,15 @@ public void testParseWordList() throws IOException {
...
         assertEquals(Arrays.asList("hello", "world"), wordList);
     }
+
+    public void testCustomAnalyzerWithNotSupportKey() {
+        Settings analyzerSettings = Settings.builder().put("tokenizer", "standard").put("foo", "bar").put("type", "custom").build();
+
+        try {
+            createComponents("my_analyzer", analyzerSettings, testAnalysis.tokenizer, testAnalysis.charFilter, testAnalysis.tokenFilter);
+            fail("expected failure");
+        } catch (IllegalArgumentException e) {
```
javanna, line 126: you can use expectThrows instead .

server/src/test/java/org/elasticsearch/index/analysis/AnalysisTests.java
```diff
@@ -27,9 +31,18 @@
...
 import static org.hamcrest.Matchers.is;
 
 public class AnalysisTests extends ESTestCase {
+    private static TestAnalysis testAnalysis;
+    private static Settings settings = Settings.builder().put(IndexMetadata.SETTING_VERSION_CREATED, Version.CURRENT).build();
+
+    @BeforeClass
+    public static void setup() throws IOException {
+        testAnalysis = createTestAnalysis(new Index("test", "_na_"), settings);
+    }
```
javanna, line 44: do you need this as before class? It looks like it's only used in the new method that you introduced?

server/src/test/java/org/elasticsearch/index/analysis/AnalysisTests.java
```diff
@@ -103,4 +116,15 @@ public void testParseWordList() throws IOException {
         List<String> wordList = Analysis.getWordList(env, nodeSettings, "foo.bar");
         assertEquals(Arrays.asList("hello", "world"), wordList);
     }
+
+    public void testCustomAnalyzerWithNotSupportKey() {
```
javanna, line 120: testCustomAnalyzerWithUnsupportedKey ?

javanna (commented):
I left a couple of comments, the approach looks good to me. I am going to run tests and edit the title which is going to go in the changelog entry.

javanna:

docs/changelog/89112.yaml
```diff
@@ -0,0 +1,6 @@
+pr: 89112
+summary: Make param check when create custom analyzer
+area: Search/Analysis
+type: enhancement
+issues:
+ - 85710
```
javanna, line 6: the changelog generation is automatic in most situations, no need to create the entry yourself. And it gets updated as the labels and title of the PR get updated. Could you remove the changelog from your PR, then it should get automatically created with the expected area and all other fields. The current build errors are around the wrong area.

server/src/main/java/org/elasticsearch/index/analysis/AnalyzerComponents.java
```diff
@@ -44,6 +44,12 @@ static AnalyzerComponents createComponents(
         final Map<String, CharFilterFactory> charFilters,
         final Map<String, TokenFilterFactory> tokenFilters
     ) {
+        for (String key : analyzerSettings.keySet()) {
```
javanna, line 47: one additional suggestion: this is technically a breaking change, as existing indices will not be loaded if they have have an unknown param in the definition of a custom analyzer. We could mitigate this by making the new behaviour depend on the index created version, so that we'd throw error only for newly created indices. We'd need to have the index created version propagated from AnalysisRegistry though.

likzn: Sorry. I am late. But i find the `method of createComponents` is stateless. How can we propagate the index created version to it?

likzn: OK. I got it.

likzn:
@javanna Hi, i resolved all comments. PTAL~

javanna:
@elasticmachine update branch

javanna:
retest this please

benwtrent:
@javanna I am not sure we should actually make this change.

All analyzer builders accept unknown parameters without throwing, and it has been that way since the start. Seems like an expensive and broad ranging breaking change that might not be worth the squeeze.

javanna:
I agree that it's a breaking change. I agree it is rather late to fix it. It does feel like something we'll want to fix though at some point, in that incorrect configuration is being silently accepted, as opposed to how Elasticsearch works in many other places. Ideally we'd promptly fail instead and provide feedback to users.

benwtrent:
@javanna the real fix is to implement something like: 

https://github.com/elastic/elasticsearch/issues/41299#issuecomment-3179687244

And provide a deprecation warning on unknown parameters and deprecate the old ctor for custom analyzers that do not provide a "known settings" value.

I am closing this PR as it is very far away from a good path forward. 

New work can be reopened through the plan laid out in the linked issue.

javanna:
Sounds good to me @benwtrent thanks

PR #89113, opened 2022-08-04
Add assumptions for changed behaviour

idegtiarenko:
Assume that tests are executed using balanced shards allocator where it is
expected that shards are going to be reallocated to another node after failure.


DaveCTurner (changes requested):
Hmm I think I'd like these tests still to run in both cases, they're scenarios worth testing, we just need to change our expectations about what happens if we keep trying on the same node.

DaveCTurner (approved):
LGTM

PR #89114, opened 2022-08-04
Don't modify source map when parsing composite runtime field

romseygeek:
When calling `RuntimeField.parseRuntimeFields()` for fields defined in the
search request, we need to wrap the Map containing field definitions in another
Map that supports value removal, so that we don't inadvertently remove the
definitions from the root request.  CompositeRuntimeField was not doing this
extra wrapping, which meant that requests that went to multiple shards and
that therefore parsed the definitions multiple times would throw an error
complaining that the fields parameter was missing, because the root request
had been modified.


romseygeek:
OK this doesn't quite work because we need to distinguish between parsing at mapping time or at search time within the composite field as well. I'm digging further...

romseygeek:
Actually it looks as though it was a bug in serialization that wasn't caught before because the fields map was being emptied at parse time and so the unnecessary code to output it never actually got triggered.  This is ready for review now @javanna

server/src/main/java/org/elasticsearch/index/mapper/CompositeRuntimeField.java
```diff
@@ -118,11 +124,6 @@ public XContentBuilder toXContent(XContentBuilder builder, Params params) throws
         for (FieldMapper.Parameter<?> parameter : parameters) {
             parameter.toXContent(builder, includeDefaults);
         }
-        builder.startObject("fields");
-        for (RuntimeField subfield : subfields) {
-            subfield.toXContent(builder, params);
-        }
-        builder.endObject();
```
javanna, line 125: don't we need to print out the subfields? Shouldn't the toXContent output be the same as what users submit? Do we have tests for the toXContent output? I am a bit confused about what this causes etc. can you expand?

romseygeek: We previously needed to print these out separately because the `fields` parameter would get emptied out after parsing.  But now that we take a copy of it for parsing, having this extra bit of code in here results in the fields being emitted twice, which causes test failures all over the place.  There are existing tests for xcontent serialization, which is what caught this.

romseygeek:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/index/mapper/CompositeRuntimeField.java
romseygeek: The loop above prints the parameters, one of which is `fields`.  Before this commit it contained an empty map, so it didn't print anything.  Now the map doesn't get emptied and so things get printed.

javanna: I see!!!

javanna (approved):
LGTM

PR #89117, opened 2022-08-04
[ML] Confirm platinum license for experimental ML aggregations

droberts195:
We have 5 experimental aggregations in the ML plugin that
are not confirming a platinum license when they are used.
Before they are made generally available we should plug
this hole.

The 5 aggregations are:

1. change_point
2. bucket_correlation
3. bucket_count_ks_test
4. frequent_items
5. categorize_text

This PR also touches a sixth aggregation, namely inference.
This was already platinum licensed but is given its own
unique feature like the other five aggregations. In the
future this will allow us to tell how popular the different
aggregations are. (The principle of separate features per
licensed aggregation is taken from the spatial plugin.)


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/MachineLearning.java
```diff
@@ -498,6 +502,37 @@ public class MachineLearning extends Plugin
...
     );
 
+    private static final LicensedFeature.Momentary CATEGORIZE_TEXT_AGG_FEATURE = LicensedFeature.momentary(
+        MachineLearningField.ML_FEATURE_FAMILY,
+        "categorize-text-agg",
+        License.OperationMode.PLATINUM
+    );
+    private static final LicensedFeature.Momentary FREQUENT_ITEM_SETS_AGG_FEATURE = LicensedFeature.momentary(
+        MachineLearningField.ML_FEATURE_FAMILY,
+        "frequent-item-sets-agg",
```
hendrikmuhs, line 512: nit: the agg is called "frequent items"

droberts195: Oops, sorry, I copied the name here from the class names rather than the aggregation itself.

dimitris-athanasiou (approved):
LGTM

PR #89122, opened 2022-08-04
[DOCS] Improves frequent items aggregation docs

szabosteve:
## Overview

This PR makes changes to the frequent items aggregation docs:
* improves formatting,
* recommends using async search and adds link,
* changes the first example to use async search,
* adds response explanation.

### Preview

[Frequent items](https://elasticsearch_89122.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/search-aggregations-bucket-frequent-items-aggregation.html)

### DIFF
--- a/docs/reference/aggregations/bucket/frequent-items-aggregation.asciidoc
+++ b/docs/reference/aggregations/bucket/frequent-items-aggregation.asciidoc
@@ -7,8 +7,7 @@
 experimental::[]
 
 A bucket aggregation which finds frequent item sets. It is a form of association 
-rules mining that identifies items that often occur together. It also helps you 
-to discover relationships between different data points (items). Items that are 
+rules mining that identifies items that often occur together. Items that are 
 frequently purchased together or log events that tend to co-occur are examples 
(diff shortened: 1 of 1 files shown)


docs/reference/aggregations/bucket/frequent-items-aggregation.asciidoc
```diff
@@ -113,17 +119,20 @@ and (2.) from which cities they make those purchases. We are interested in sets
 with three or more items, and want to see the first three frequent item sets 
 with the highest support.
 
+Note that we use the <<async-search, async search>> endpoint in this first 
+example.
```
hendrikmuhs, line 123: doesn't this require changing this?

> The API returns a response similar to the following one:

Async returns an `id`, which than needs to be used to get the response.

szabosteve: I've adjusted the docs so they include the missing step. However, I didn't add the response to the POST request as it would be long and the instructions are clear enough without it. Let me know if you think it would be better to add the response though.

hendrikmuhs (approved):
LGTM

PR #89127, opened 2022-08-04
Add TransportHealthNodeAction

gmarouli:
Introduce a base transport action for health node requests.

This action ensures that:
- the target node is a health node


gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/node/action/NotHealthNodeException.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.elasticsearch.ElasticsearchException;
+import org.elasticsearch.common.io.stream.StreamInput;
+
+import java.io.IOException;
+
+/**
+ * Exception which indicates that an operation failed because the node stopped being the selected health node.
+ */
+public class NotHealthNodeException extends ElasticsearchException {
```
andreidan, line 18: I don't think we need this kind of exception ?

it doesn't seem used, is it?

server/src/main/java/org/elasticsearch/health/node/action/TransportHealthNodeAction.java
```diff
@@ -0,0 +1,141 @@
...
+    protected void doExecute(Task task, final Request request, ActionListener<Response> listener) {
+        ClusterState state = clusterService.state();
+        logger.trace("starting to process request [{}] with cluster state version [{}]", request, state.version());
+        if (task != null) {
+            request.setParentTask(clusterService.localNode().getId(), task.getId());
+        }
+        new AsyncSingleAction(task, request, listener).doStart(state);
+    }
+
+    class AsyncSingleAction {
```
andreidan, line 79: Do we need a class here? The logic it provides could be a method?

Maybe I'm missing something but seems to just wrap the arguments `doExecute` has?

server/src/main/java/org/elasticsearch/health/node/action/TransportHealthNodeAction.java
```diff
@@ -0,0 +1,141 @@
...
+                            } else {
+                                healthOperation(task, request, clusterState, listener);
+                            }
+                        } catch (Exception e) {
+                            listener.onFailure(e);
+                        }
+                    });
+                } else {
+                    logger.trace("forwarding request [{}] to health node [{}]", actionName, healthNode);
+                    transportService.sendRequest(
```
andreidan, line 115: wouldn't the client code lookup the health node themselves in the cluster state? 

and then use `transportService.sendRequest(healthNode, TransportHealthNodeAction.NAME, ...)` ?

andreidan (commented):
Thanks Mary, left a few comments

server/src/main/java/org/elasticsearch/health/node/action/NotHealthNodeException.java
gmarouli: You are right, I missed it, it's from the previous set-up :). I will remove it.

server/src/main/java/org/elasticsearch/health/node/action/TransportHealthNodeAction.java
gmarouli: That was my approach in the beginning, the local node would look up the health node and send the request directly. But then I didn't find many places in the code where we do this, so I thought to follow the way we "talk" to the master node. If you think it's not an anti-pattern, I think we should park this PR and try to use directly the `transportService` to notify the health node. If we see this is not enough we can come back. Right?

andreidan: >I think we should park this PR and try to use directly the transportService to notify the health node. 

I think you'd still need one transport action that the nodes in the cluster will call to send their health status to the health node (maybe a different PR if you prefer that)

masseyke: I can see the appeal of doing it this way -- only this base action needs to know how to find the health node, and writing and using these health node actions becomes simpler. Plus if we add more sophisticated logic to retry if the health node is null (like [TransportMasterNodeAction](https://github.com/elastic/elasticsearch/blob/main/server/src/main/java/org/elasticsearch/action/support/master/TransportMasterNodeAction.java) does, which I assume this was modeled after), then we don't have to update multiple actions in the future.
Although if we only have two transport actions (one that all nodes use to push data to the health node, and one that all nodes use to read the health node's cache) then it probably doesn't really matter either way. But it would probably still be nice to have this logic in one place instead of two.

gmarouli: @masseyke I was thinking the same and in the first iteration there was a retrying mechanism too. However, this is quite some boiler plate code to only facilitate sending a request to the health node (without the retry). I am considering to write something simpler that could be used by both actions, this way there will be one place with the code to forward requests.

Also we can always write this if we see it's necessary later, I prefer that than introducing before it's necessary.

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/node/action/TransportHealthNodeAction.java
gmarouli: I think after the latest trim the boiler plate code is not too much (excluding helper functions and test that we were going to need anyway), so I lean towards keeping it. A forwarder would not be much different because it also needs to choose if it's going to execute locally or not. The `TransportHealhNodeAction` will keep all thr routing logic and then the subclasses will contain only functional code and they will be easy to reason about. Any objections?

gmarouli:

server/src/main/java/org/elasticsearch/health/node/action/TransportHealthNodeAction.java
DaveCTurner: Retrying here seems kinda reasonable to me too, even if it's something that individual implementations might opt out of.

server/src/main/java/org/elasticsearch/health/node/action/TransportHealthNodeAction.java
```diff
@@ -0,0 +1,125 @@
...
+                        } else {
+                            healthOperation(task, request, clusterState, listener);
+                        }
+                    } catch (Exception e) {
+                        listener.onFailure(e);
+                    }
+                });
+            } else {
+                logger.trace("forwarding request [{}] to health node [{}]", actionName, healthNode);
+                transportService.sendRequest(
```
DaveCTurner, line 100: I think this should be `sendChildRequest` so that cancellation (and tracing etc) will propagate correctly.

server/src/main/java/org/elasticsearch/health/node/action/TransportHealthNodeAction.java
andreidan: Thanks for your thoughts @DaveCTurner, @masseyke, and @gmarouli 

Maybe I'm missing some points/context in how we'll use these actions. Here's how I think about the code that's going to use them.

1. Action that _sends_ data to the health node.

Every node in the cluster will occasionally send health updates to the health node. 
These nodes _might_ want to make use of a retry mechanism but I'm inclined to say this would be better as a client side implementation as we might want to stop retrying old updates when new ones come in (otherwise we might end up with ordering issues as we could have several retries in progress and the latest health state could arrive to the health node before an older state) 

Is there a better way to reason about the usage here?

2. Action that _gets_ data from the health node. 

This will be the disk health indicator. As we'll channel these requests on every health API request (or _might_ want to do it in the background?) we'll want to give an answer to the health API request ASAP (e.g. the disk indicator will not return the information from the health node). No retry needed in this case.


W.r.t. forwarding the request to the correct health node if the identity of the health node changes "mid-request".

For both types of actions we'll need to listen for changes in the health node's identity so we can send our health status to the new health node. We'll send the health data anyway, the forwarding is redundant IMO. 
Is this a wrong assumption on my part? 

If the identity of the health node changes mid-GET-request when requesting data from a health node, we'll get an error of sorts in which case the health API not report any information from the health node (we already need to handle this case at number 2 above isn't it? )

My main concerns with both these (nice :) ) features is to not be in control of what kind of traffic the health monitoring is generating in a cluster. I'm very happy to be corrected if I misunderstand some of the use-cases and the way we plan to use these APIs.

andreidan (approved):
LGTM thanks for iterating on this Mary 

(also note David's suggestion to use `sendChildRequest` - it'll remove the need to explicitly do `request.setParentTask(localNode.getId(), parentTask.getId());` - however  tracing and the likes are safe in the current implementation)

server/src/main/java/org/elasticsearch/health/node/action/TransportHealthNodeAction.java
gmarouli: Thanks for the thorough write up. Let me see if I reply equally well :).

1. Action that _sends_ data to the health node.

After the first round of reviewing I gave it some extra thought while I was drafting the next PR and I found a much better place for the retry mechanism. I suggest we keep the `lastSuccessfullyReported` health state in the `LocalHealthMonitor`. This will provide a simpler and more efficient retry mechanism since with every rerun we will check if the current state is different than the `lastSuccessfullyReported` and try to send it. If there is a failure and the state is still different than the last one reported then we will try to send it again at the next run. The only drawback I can see is that it depends on the polling interval of the local monitor. If someone changes this setting to a very large value then the retry will be really delayed, but I do not see this as an issue, since all reporting will be probably "delayed"

2. Action that _gets_ data from the health node.

Based on the requirement _"the health api needs to respond quickly"_ I agree there is not need for an elaborate retry mechanism. There is still a lot of useful information to return without the information from the health node.

I do think though that if a health node changed from, let's say, `node-1` to `node-2` mid request, it is not that costly to allow one extra hop, from `node-1` to `node-2` (assuming it's unlikely for the persistent task to be reassigned more than once for the duration of request_.

Based on the above, I think that the current version of the `TransportHealthNodeAction` extracts the "find the health node logic" in one place and allows us to have only functional logic in the sub-classes which will make them much easier to read and reason about. 

So that would be my proposal, keep the current base action as simple as possible in order to keep the sub-classes also as simple as possible. 

What do you think?

DaveCTurner: > otherwise we might end up with ordering issues as we could have several retries in progress and the latest health state could arrive to the health node before an older state

I think this is true even without retries so we would need to protect against it either way.

server/src/main/java/org/elasticsearch/health/node/action/TransportHealthNodeAction.java
DaveCTurner: Oh right as Andrei points out we are explicitly setting up the parent anyway. Yeah I'd prefer this to use `sendChildRequest` instead of doing that but at least it's not a bug either way.

gmarouli: TIL! @DaveCTurner thanks, I will use this if we have a task then, does it mean that the following code is redundant or does it serve a different purpose? It's one of the first things we do in `doExecute(..)`
```
        if (task != null) {
            request.setParentTask(clusterService.localNode().getId(), task.getId());
        }
```

gmarouli: Nvm, I saw @andreidan already confirmed that :)!

server/src/main/java/org/elasticsearch/health/node/action/TransportHealthNodeAction.java
gmarouli: Cleaned up.

PR #89132, opened 2022-08-04
Log when repository is marked as corrupted

idegtiarenko:
At the moment we do not log anything when repo is marked as corrupted.
Having such log entry might help to find a cause for corruption during RCA.

### DIFF
--- a/server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
+++ b/server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
@@ -2026,6 +2026,7 @@ public abstract class BlobStoreRepository extends AbstractLifecycleComponent imp
     private void markRepoCorrupted(long corruptedGeneration, Exception originalException, ActionListener<Void> listener) {
         assert corruptedGeneration != RepositoryData.UNKNOWN_REPO_GEN;
         assert bestEffortConsistency == false;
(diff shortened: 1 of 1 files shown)


original-brownbear (approved):

server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
```diff
@@ -2026,6 +2026,7 @@ private static String previousWriterMessage(@Nullable Tuple<Long, String> previo
     private void markRepoCorrupted(long corruptedGeneration, Exception originalException, ActionListener<Void> listener) {
         assert corruptedGeneration != RepositoryData.UNKNOWN_REPO_GEN;
         assert bestEffortConsistency == false;
+        logger.warn(() -> "Marking repository [" + metadata.name() + "] as corrupted", originalException);
```
henningandersen, line 2029: nit: the message-supplier form is unnecessary here:

```suggestion
        logger.warn("Marking repository [" + metadata.name() + "] as corrupted", originalException);
```

henningandersen (approved):
LGTM.

server/src/main/java/org/elasticsearch/repositories/blobstore/BlobStoreRepository.java
idegtiarenko: I was assuming we prefer it to avoid string concatenation (with repository name in this case). Is it because of warn?

PR #89137, opened 2022-08-04
Handling the master stability health case where there has never been an elected master node

masseyke:
If a master-eligible node comes up and has never seen an elected master node (and assuming that a quorum requires more than one node), then it ought to report that the master stability health is red because it cannot form a quorum.


jbaiera (approved):
LGTM! Thanks for the test 😃

PR #89138, opened 2022-08-04
Update SLM health diagnosis message to include unhealthy policy details

jbaiera:
This PR updates the SLM health indicator to output additional information about unhealthy SLM policies as part of the diagnosis.cause section of the response. An example of an unhealthy slm policy response would be as follows:

<pre><code>
{
  "status": "yellow",
  "symptom": "Encountered [1] unhealthy snapshot lifecycle management policies.",
  "details": {
    "slm_status": "RUNNING",
    "policies": 1,
    "unhealthy_policies": {
      "count": 1,
      "invocations_since_last_success": {
        "test-policy": 12
      }
    }
  },
  "impacts": [
    {
      "severity": 2,
      "description": "Some automated snapshots have not had a successful execution recently. Indices restored from affected snapshots may not contain recent changes.",
      "impact_areas": [
        "backup"
      ]
    }
  ],
  "diagnosis": [
    {
      <b>"cause": """Several automated snapshot policies are unhealthy:
- [test-policy] has not had [12] successful executions since [2022-08-04T21:24:06.817Z]
- [test-policy-2] has not had [12] successful executions since [2022-08-04T21:24:06.817Z]""",</b>
      <b>"action": """Check the snapshot lifecycle policies for detailed failure info:
- /_slm/policy/test-policy?human
- /_slm/policy/test-policy-2?human""",</b>
      "affected_resources": [
        "test-policy",
        "test-policy-2"
      ],
      "help_url": "https://ela.st/fix-recent-snapshot-failures"
    }
  ]
}
</code></pre>

One thing of note is that this breaks a former design decision to keep diagnosis text stable instead of generated, and to place any affected resources into the affected_resources section. I don't know if we want to continue following this rule, but it is something to keep in mind.


x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/slm/SlmHealthIndicatorService.java
```diff
@@ -54,13 +57,22 @@ public class SlmHealthIndicatorService implements HealthIndicatorService {
...
-        "check_recent_snapshot_failures",
-        "The following snapshot lifecycle policies have exceeded the warning threshold for repeat failures without a successful execution.",
-        "Check the snapshot lifecycle policies [/_slm/policy/<policy_name>?human] for detailed failure info.",
-        ACTION_CHECK_RECENTLY_FAILED_SNAPSHOTS_HELP_URL
-    );
+    private static final DateFormatter FORMATTER = DateFormatter.forPattern("iso8601").withZone(ZoneOffset.UTC);
+
+    public static final String DIAGNOSIS_CHECK_RECENTLY_FAILED_SNAPSHOTS_ID = "check_recent_snapshot_failures";
+    public static final String DIAGNOSIS_CHECK_RECENTLY_FAILED_SNAPSHOTS_ACTION = "Check the snapshot lifecycle policies "
+        + "[/_slm/policy/<policy_name>?human] for detailed failure info.";
```
andreidan, line 64: Since we're on the dynamic messaging here can we have the action provide the API call with the SLM policy in it? 

ie. `/_slm/policy/test-policy?human`

andreidan (commented):
Thanks for working on this Jimmy 

Left a minor suggestion. 

I think having dynamic diagnosis is cool  (we've done it before when recommending which tier to add to the system)

Can we work the messaging when multiple policies are experincing problems at the same time? 
Currently the diagnosis cause would repeat this message (on new lines)
```
"cause": "An automated snapshot [test-policy] has not had [12] successful executions since [2022-08-04T21:24:06.817Z]
An automated snapshot [test-policy1] has not had [6] successful executions since [2022-08-04T21:24:06.817Z]
An automated snapshot [test-policy2] has not had [55] successful executions since [2022-08-04T21:24:06.817Z]",
```

Are we ok with this or should we attempt to format it in a way that lists the policies and their corresponding details in a "bullet-point" manner? 

eg. 
```
"cause": "Several automated snapshot policies are unhealthy:
- [test-policy] has not had [12] successful executions since [2022-08-04T21:24:06.817Z]
- [test-policy1] has not had [6] successful executions since [2022-08-04T21:24:06.817Z]
- [test-policy2] has not had [55] successful executions since [2022-08-04T21:24:06.817Z]
```

(similar for the action if we parameterise the URL we include there

// cc @Leaf-Lin

Leaf-Lin:
+1 on bullet-point, to make it easier to read (both on the API, and later on integrated into the UI.

x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/slm/SlmHealthIndicatorService.java
jbaiera: Updated code and PR description with the new messages.

andreidan (approved):
LGTM, thanks Jimmy

Just a final thing I noticed on the last read, the messaging sounds a bit strange (to me) with the negation: `[test-policy] has not had [12] successful executions since [2022-08-04T21:24:06.817Z]`

Would it be easier to read to avoid the "has not had ... successful" negation ?
ie. `[test-policy] had [12] failed executions since [2022-08-04T21:24:06.817Z]`

jbaiera:
Agreed on the messaging - I think it's also important to include that it's not just a number of failures since X date, but rather a number of failures with a successful execution since X date:

`[test-policy] had [12] repeated failures without successful execution since [2022-08-04T21:24:06.817Z]`

I'll update the final wording with something like that.

andreidan:
@jbaiera Thanks for improving the messaging Jimmy, LGTM 🚀

PR #89140, opened 2022-08-04
Add advice on restoring ILM indices into unmanaged state

jbaiera:
This adds a blurb to the documentation of the ILM settings that explains how to take a snapshot of an index previously managed by ILM and restore it into an index that is unmanaged.

Unfortunately, IDEA would not allow me to save any edits without removing trailing spaces to a good number of lines...


dakrone (approved):
LGTM, thanks for adding this Jimmy

PR #89152, opened 2022-08-07
Add note on per-segment field name overhead

DaveCTurner:
We encountered a case where a substantial fraction of the heap usage was
due to per-segment-per-field `FieldInfo` objects, particularly
`FieldInfo#name`. This commit adds a note to the sizing docs about this
overhead.

Relates #77466

### DIFF
--- a/docs/reference/how-to/size-your-shards.asciidoc
+++ b/docs/reference/how-to/size-your-shards.asciidoc
@@ -76,8 +76,8 @@ Moreover every segment requires a small amount of heap memory for each mapped
 field. This per-segment-per-field heap overhead includes a copy of the field
 name, encoded using ISO-8859-1 if applicable or UTF-16 otherwise. Usually this
 is not noticeable, but you may need to account for this overhead if your shards
-have high segment counts and the corresponding mappings contain very long field
-names or high field counts.
+have high segment counts and the corresponding mappings contain high field
+counts and/or very long field names.
 
 [discrete]
 [[shard-auto-balance]]
--- a/docs/reference/how-to/size-your-shards.asciidoc
+++ b/docs/reference/how-to/size-your-shards.asciidoc
@@ -55,14 +55,14 @@ thread pool>>. This can result in low throughput and slow search speeds.
 
 [discrete]
 [[each-shard-has-overhead]]
-==== Each index, shard and field has overhead
+==== Each index, shard, segment and field has overhead
 
 Every index and every shard requires some memory and CPU resources. In most
 cases, a small set of large shards uses fewer resources than many small shards.
 
 Segments play a big role in a shard's resource usage. Most shards contain
-several segments, which store its index data. {es} keeps segment metadata in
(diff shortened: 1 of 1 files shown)


docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -72,6 +72,13 @@ space. By default {es} will automatically create a mapping for every field in
 every document it indexes, but you can switch off this behaviour to
 <<explicit-mapping,take control of your mappings>>.
 
+Moreover every segment requires a small amount of heap memory for each mapped
+field. This per-segment-per-field heap overhead includes a copy of the field
+name, encoded using UTF-8. Usually this is not noticeable, but you may need to
+account for this overhead if you have very long field names or high field
```
tlrx, line 78: I wonder if it should provide an indication of how many is a _high field count_?

DaveCTurner: I was kind of being deliberately vague here because I don't really want to work out how much of these overheads are covered by the other sizing guidance (particularly the 1kiB-per-field-per-index rule) particularly since the high segment count in this case was caused by a bug, and the duplication of field names in the heap is also a bug, and fixing either of these bugs would mean we can remove this paragraph.

tlrx: Got it, thanks.

docs/reference/how-to/size-your-shards.asciidoc
```diff
@@ -72,6 +72,13 @@ space. By default {es} will automatically create a mapping for every field in
...
+counts in any of your mappings, or if any of your shards have high segment
```
henningandersen, line 79: I think it is the combination of many segments with many fields (possibly long names) that makes this a problem. Can we mention that somehow, perhaps rephrasing to:

`... account for this overhead if you have very long field names or high field counts in combination with high segment counts
```

**Review by henningandersen — COMMENTED** [review]
A few details, otherwise this looks good.

File: docs/reference/how-to/size-your-shards.asciidoc [hunk 3]
```diff
@@ -72,6 +72,13 @@ space. By default {es} will automatically create a mapping for every field in
 every document it indexes, but you can switch off this behaviour to
 <<explicit-mapping,take control of your mappings>>.
 
+Moreover every segment requires a small amount of heap memory for each mapped
+field. This per-segment-per-field heap overhead includes a copy of the field
+name, encoded using UTF-8. Usually this is not noticeable, but you may need to
```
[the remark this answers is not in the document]
DaveCTurner: Ah oops yes. Fixed in 913e68b1748.

docs/reference/how-to/size-your-shards.asciidoc
DaveCTurner: In principle it could really be any combination of the above but yes it's only really not covered by the per-field overhead at high segment counts (or crazy field name lengths). Fixed in 913e68b1748.

henningandersen (approved):
LGTM.

PR #89153, opened 2022-08-07
Dry up custom immutable Map.Entry implementations

original-brownbear:
Follow-up to #88815.
No need to have two equivalent implementations here.


grcevski (approved):
LGTM!

original-brownbear:
Thanks Nikola!

Issue #89058, opened 2022-08-03
Support viewing API key privileges


ywangd:
Today the privileges of an API key cannot be viewed after its creation[^1]. The alternative is to check its privileges with the [HasPrivileges API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-has-privileges.html). This is useful for authorization purpose but not helpful in terms of management and audit etc. API keys are now updatable (since #88186). Since each update can potentially alter both the key's assigned descriptors (`role_descriptors`) and the captured owner user's permissions (`limited_by_role_descriptors`). It has become even harder to accurately know about an API key's privileges at any point of time. Hence the recommendation is to add support for viewing API key privileges with relevant APIs. The proposed work items are as the follows:

- [x] Return `role_descriptors` by default in [GetApiKey API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-get-api-key.html). (no change to the privilege model around it)
- [x] Optionally return `limited_by_role_descriptors` in GetApiKey API when (1) a `limited_by` query parameter is specified and (2) the authenticating subject is the owner of the API or has `manage_api_key` (or higher) privilege.
- [x] Update [GetUserPrivileges API](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-get-user-privileges.html) to return error 400 (instead of the current 500) when call with API keys and advise users to use the GetApiKeyAPI
- [x] Update [QueryApiKey API](https://www.elastic.co/guide/en/elasticsearch/reference/8.3/security-api-query-api-key.html) to have similar support for `role_descriptors` and `limited_by_role_descriptors`.
- [x] REST specs, YAML tests, Docs

[^1]: It is possible to retrieving the backing document of an API key by directly reading the security index. But directly accessing security index is generally discouraged.

PR #89166, opened 2022-08-08
Show assigned role descriptors in Get/QueryApiKey response

ywangd:
This PR adds a new `role_descriptors` field in the API key entity returned by both
GetApiKey and QueryApiKey APIs. The field value is the map of the role descriptors
that are assigned to an API key when creating or updating the key. If the key has no
assigned role descriptors, i.e. it inherits the owner user's privileges, an empty object
is returned in place.

Relates: #89058


ywangd:
This PR has changes for both Get and Query API key APIs because they are very closely related, i.e. they share the same underlying data model. So I think it is preferrable to have both of them updated. Doc and yaml changes are left out of this PR.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/apikey/ApiKey.java
```diff
@@ -200,6 +228,10 @@ public boolean equals(Object obj) {
         PARSER.declareString(constructorArg(), new ParseField("username"));
         PARSER.declareString(constructorArg(), new ParseField("realm"));
         PARSER.declareObject(optionalConstructorArg(), (p, c) -> p.map(), new ParseField("metadata"));
+        PARSER.declareNamedObjects(optionalConstructorArg(), (p, c, n) -> {
+            p.nextToken();
+            return RoleDescriptor.parse(n, p, false);
+        }, new ParseField("role_descriptors"));
```
ywangd, line 234: This PARSER is only used in tests. So it can be moved into a test class in future. For now, I keep it here to avoid unrelated changes.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -2226,6 +2289,56 @@ private void doTestUpdateApiKeysNotFound(final UpdateApiKeyRequest request) {
...
+        ASSIGNED_ROLE_DESCRIPTORS,
+        LIMITED_BY_ROLE_DESCRIPTORS
+    }
+
+    // Check attributes with both the raw document and the get api key response whenever possible
+    @SuppressWarnings("unchecked")
+    private void expectAttributesForApiKey(String apiKeyId, Map<ApiKeyAttribute, Object> attributes) throws IOException {
+        final Map<String, Object> apiKeyDocMap = getApiKeyDocument(apiKeyId);
+        final PlainActionFuture<GetApiKeyResponse> future = new PlainActionFuture<>();
+        client().execute(GetApiKeyAction.INSTANCE, GetApiKeyRequest.usingApiKeyId(apiKeyId, false), future);
```
ywangd, line 2304: This method replaces (and wraps) the existing methods like `expectRoleDescriptorsForApiKey` and it internally does the comparison use both the raw document and GetApiKey response. The majority of the changes in this class is due to this refactoring.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -2333,18 +2436,29 @@ private void verifyGetResponse(
...
+        verifyGetResponse(
+            ES_TEST_ROOT_USER,
+            expectedNumberOfApiKeys,
+            responses,
+            metadatas,
+            expectedRoleDescriptors,
+            response,
+            validApiKeyIds,
+            invalidatedApiKeyIds
+        );
```
ywangd, line 2453: This method now takes a list of role descriptors for comparison. This is the other major contributor to changes in this file.

n1v0lg (approved):
LGTM! 

Only small, optional nits throughout. 

From a high level standpoint, the only gotcha I see is that including `role_descriptors` in the query and get responses by default will result in potentially noticeably bigger responses than before (since `role_descriptors` can be relatively large compared to the other fields). 

This is only an observation, and not a strong reason to complicate the API by making this behavior toggle-able, esp. since the query API is paginated to begin with.

x-pack/plugin/security/qa/security-trial/src/javaRestTest/java/org/elasticsearch/xpack/security/apikey/ApiKeyRestIT.java
```diff
@@ -79,6 +81,115 @@ public void cleanUp() throws IOException {
         invalidateApiKeysForUser(MANAGE_OWN_API_KEY_USER);
     }
 
+    @SuppressWarnings("unchecked")
+    public void testGetApiKeyRoleDescriptors() throws IOException {
+        // First key without assigned role descriptors
+        final Request createApiKeyRequest1 = new Request("POST", "_security/api_key");
```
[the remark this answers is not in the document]
ywangd: Thanks for the suggestion. I added it as a randomisation for `createApiKeyRequest1` which now chooses between "omiting the role_descriptor field" and "explicitly configuring it to {}".

ywangd:
> From a high level standpoint, the only gotcha I see is that including `role_descriptors` in the query and get responses by default will result in potentially noticeably bigger responses than before (since `role_descriptors` can be relatively large compared to the other fields).

I had the same thought initially and proposed for it to be controlled by a query parameter (defaults to off). But Jake convinced me that showing it by default has better parity to its creation API counterpart. Also, the size is probably not too big of a concern. It may be in range of couple dozen MBs (for something 50K API keys), but some cluster state response can be couple hundred MBs.

PR #89167, opened 2022-08-08
[ML] Extract ML tasks into a context class for use in autoscaling dec…

dimitris-athanasiou:
…ider

This commit is a refactoring that encapsulates together the ML tasks
that are considered when taking autoscaling decisions.


droberts195 (approved):
LGTM

PR #89168, opened 2022-08-08
Remove redundant and slow null token check from KeywordFieldMapper

original-brownbear:
Just found this old fix I had lying around :) This gives about a 10% speedup on the KW field mapper for me in the `BeatsMapperBenchmark` (through making things inline nicer I suspect).

No need to check for the null token manually and parse `textOrNull`.
Either we can just use `text()` since we know we don't have to deal
with a null token or use `textOrNull` and check the return value for
`null`.
I chose the latter because it benchmarked slightly faster in `BeatsMapperBenchmark`
but both save the expensive call to `currentToken` on the heavily nested x-content
parser that we use here.


mayya-sharipova (approved):
Nice optimization!

original-brownbear:
Thanks Mayya!

PR #89174, opened 2022-08-08
Make sure listener is completed in a correct thread context

idegtiarenko:


server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/AllocationActionListener.java
```diff
@@ -29,10 +32,15 @@ public static ActionListener<Void> rerouteCompletionIsNotRequired() {
         return ActionListener.noop();
     }
 
+    // TODO replace with 2 arg call
```
idegtiarenko, line 35: I am working on remaining usages

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/AllocationActionListenerTests.java
```diff
@@ -59,4 +66,36 @@ public void testShouldDelegateFailureImmediately() {
...
+            listener.clusterStateUpdate().onResponse(AcknowledgedResponse.TRUE);
+        });
+        pool.generic().execute(() -> {
+            pool.getThreadContext().addResponseHeader("header", "reroute");
+            listener.reroute().onResponse(null);
+        });
+
+        queue.runAllTasks();
+        assertThat(completed.get(), equalTo("root"));
+    }
```
DaveCTurner, line 96: I wonder if we need the thread pool here? How about this?

```suggestion
    public void testShouldExecuteWithCorrectContext() {

        var threadContext = new ThreadContext(Settings.EMPTY);
        var completed = new AtomicReference<String>();
        AllocationActionListener<AcknowledgedResponse> listener;

        threadContext.putHeader("header", "root");
        listener = new AllocationActionListener<>(
            ActionListener.wrap(
                ignore -> completed.set(threadContext.getHeader("header")),
                exception -> {
                    throw new AssertionError("Should not fail in test");
                }
            ),
            threadContext
        );

        try (var ignored = threadContext.stashContext()) {
            threadContext.putHeader("header", "clusterStateUpdate");
            listener.clusterStateUpdate().onResponse(AcknowledgedResponse.TRUE);
        }

        try (var ignored = threadContext.stashContext()) {
            threadContext.putHeader("header", "reroute");
            listener.reroute().onResponse(null);
        }

        assertThat(completed.get(), equalTo("root"));
    }
```

NB using request headers in this test - response headers are not cleared in a `ContextPreservingActionListener`, and in fact you get all three response headers in the result so the following listener will fail.

```
            ActionListener.wrap(
                ignore -> {
                    final var testHeaders = pool.getThreadContext().getResponseHeaders().get("header");
                    assertThat(testHeaders, hasSize(1));
                    completed.set(testHeaders.get(0));
                },
                exception -> { throw new AssertionError("Should not fail in test"); }
            ),
```

Alternatively we could assert that we do want this response-header-preserving behaviour here.

DaveCTurner (approved):
LGTM

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/AllocationActionListenerTests.java
```diff
@@ -59,4 +67,84 @@ public void testShouldDelegateFailureImmediately() {
...
+
+        var listener = new AllocationActionListener<>(ActionListener.wrap(ignore -> {
+            assertThat(context.getResponseHeaders().get("header"), containsInAnyOrder("1", "3", "4"));
+            assertThat(context.getHeader("header"), equalTo("root"));
+        }, exception -> { throw new AssertionError("Should not fail in test"); }), context);
+
+        // this header should be ignored as it is added after context is captured
+        context.addResponseHeader("header", "2");
+
+        // reroute is executed for multiple changes so its headers should be ignored
```
DaveCTurner, line 87: nit: duplicated comment (should this one be about the CS update listener instead?)

```suggestion
        // headers for clusterStateUpdate listener are captured
```

server/src/test/java/org/elasticsearch/cluster/routing/allocation/allocator/AllocationActionListenerTests.java
```diff
@@ -59,4 +67,84 @@ public void testShouldDelegateFailureImmediately() {
...
+
+        var context = new ThreadContext(Settings.EMPTY);
+
+        // should not be changed after the listener is created
+        context.putHeader("header", "root");// ensure this is visible in the listener
+        context.addResponseHeader("header", "1");
+
+        var listener = new AllocationActionListener<>(ActionListener.wrap(ignore -> {
+            assertThat(context.getResponseHeaders().get("header"), containsInAnyOrder("1", "3", "4"));
+            assertThat(context.getHeader("header"), equalTo("root"));
```
DaveCTurner, line 81: I'd prefer to have some verification that this listener is actually completed (even though a bug of that nature would make other tests fail too)

DaveCTurner (commented):
Looks good, sorry just one more comment on the test.

DaveCTurner (approved):
LGTM

PR #89176, opened 2022-08-08
Fix CheckTargetShardsCountStep #(48460)

### LINKED ISSUE #78460: [CI] ShrinkActionIT testAutomaticRetryFailedShrinkAction failing
**Build scan:**
https://gradle-enterprise.elastic.co/s/yicchv3754amg/tests/:x-pack:plugin:ilm:qa:multi-node:javaRestTest/org.elasticsearch.xpack.ilm.actions.ShrinkActionIT/testAutomaticRetryFailedShrinkAction

**Reproduction line:**
`./gradlew ':x-pack:plugin:ilm:qa:multi-node:javaRestTest' --tests "org.elasticsearch.xpack.ilm.actions.ShrinkActionIT.testAutomaticRetryFailedShrinkAction" -Dtests.seed=D966D2304D0930DC -Dtests.locale=pl-PL -Dtests.timezone=Europe/Tallinn -Druntime.java=11`

**Applicable branches:**
master

**Reproduces locally?:**
Most of the time we can achieve 1 or 2 failures if we run it with `-Dtests.iters=100`.

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.xpack.ilm.actions.ShrinkActionIT&tests.test=testAutomaticRetryFailedShrinkAction

**Failure excerpt:**
```
java.lang.AssertionError: (No message provided)

  at __randomizedtesting.SeedInfo.seed([D966D2304D0930DC:1F9215989061A05]:0)
  at org.junit.Assert.fail(Assert.java:86)
  at org.junit.Assert.assertTrue(Assert.java:41)
  at org.junit.Assert.assertTrue(Assert.java:52)
  at org.elasticsearch.xpack.ilm.actions.ShrinkActionIT.lambda$testAutomaticRetryFailedShrinkAction$17(ShrinkActionIT.java:295)
  at org.elasticsearch.test.ESTestCase.assertBusy(ESTestCase.java:985)
  at org.elasticsearch.xpack.ilm.actions.ShrinkActionIT.testAutomaticRetryFailedShrinkAction(ShrinkActionIT.java:295)
  at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(NativeMethodAccessorImpl.java:-2)
  at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
  at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.lang.reflect.Method.invoke(Method.java:566)
  at com.carrotsearch.randomizedtesting.RandomizedRunner.invoke(RandomizedRunner.java:1758)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$8.evaluate(RandomizedRunner.java:946)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$9.evaluate(RandomizedRunner.java:982)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$10.evaluate(RandomizedRunner.java:996)
  at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
  at org.apache.lucene.util.TestRuleSetupTeardownChained$1.evaluate(TestRuleSetupTeardownChained.java:44)
  at org.apache.lucene.util.AbstractBeforeAfterRule$1.evaluate(AbstractBeforeAfterRule.java:43)
  at org.apache.lucene.util.TestRuleThreadAndTestName$1.evaluate(TestRuleThreadAndTestName.java:45)
  at org.apache.lucene.util.TestRuleIgnoreAfterMaxFailures$1.evaluate(TestRuleIgnoreAfterMaxFailures.java:60)
  at org.apache.lucene.util.TestRuleMarkFailure$1.evaluate(TestRuleMarkFailure.java:44)
  at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
  at com.carrotsearch.randomizedtesting.ThreadLeakControl$StatementRunner.run(ThreadLeakControl.java:375)
  at com.carrotsearch.randomizedtesting.ThreadLeakControl.forkTimeoutingTask(ThreadLeakControl.java:824)
  at com.carrotsearch.randomizedtesting.ThreadLeakControl$3.evaluate(ThreadLeakControl.java:475)
  at com.carrotsearch.randomizedtesting.RandomizedRunner.runSingleTest(RandomizedRunner.java:955)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$5.evaluate(RandomizedRunner.java:840)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$6.evaluate(RandomizedRunner.java:891)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$7.evaluate(RandomizedRunner.java:902)
  at org.apache.lucene.util.AbstractBeforeAfterRule$1.evaluate(AbstractBeforeAfterRule.java:43)
  at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
  at org.apache.lucene.util.TestRuleStoreClassName$1.evaluate(TestRuleStoreClassName.java:38)
  at com.carrotsearch.randomizedtesting.rules.NoShadowingOrOverridesOnMethodsRule$1.evaluate(NoShadowingOrOverridesOnMethodsRule.java:40)
  at com.carrotsearch.randomizedtesting.rules.NoShadowingOrOverridesOnMethodsRule$1.evaluate(NoShadowingOrOverridesOnMethodsRule.java:40)
  at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
  at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
  at org.apache.lucene.util.TestRuleAssertionsRequired$1.evaluate(TestRuleAssertionsRequired.java:53)
  at org.apache.lucene.util.AbstractBeforeAfterRule$1.evaluate(AbstractBeforeAfterRule.java:43)
  at org.apache.lucene.util.TestRuleMarkFailure$1.evaluate(TestRuleMarkFailure.java:44)
  at org.apache.lucene.util.TestRuleIgnoreAfterMaxFailures$1.evaluate(TestRuleIgnoreAfterMaxFailures.java:60)
  at org.apache.lucene.util.TestRuleIgnoreTestSuites$1.evaluate(TestRuleIgnoreTestSuites.java:47)
  at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
  at com.carrotsearch.randomizedtesting.ThreadLeakControl$StatementRunner.run(ThreadLeakControl.java:375)
  at com.carrotsearch.randomizedtesting.ThreadLeakControl.lambda$forkTimeoutingTask$0(ThreadLeakControl.java:831)
  at java.lang.Thread.run(Thread.java:829)

```

gmarouli:
**The issue**
The flaky test simulates the following:
- Create a shrink policy with an invalid target shard count
- Then change the policy to have a valid target shard count
- Expectation: the `check-target-shards-count` will return true and the shrink operation will be successful.

What was happening in some cases in the background:
- Create the shrink policy with an invalid target shard count
- The `check-target-shards-count` gets created and queued to be executed with the invalid target shards count.
- The task doesn't get enough priority to be executed
- We change the policy to have a valid target shards count
- We execute the queued task which still has the outdated target shard count.

**Proof**
We enriched the code with some extra logging to verify that the above scenario is actually correct:
```
## Adding the check target shards to the executingTasks

[2022-08-08T18:02:52,824][INFO ][o.e.x.i.IndexLifecycleRunner] [javaRestTest-0] #78460: Adding task to queue check if I can shrink to numberOfShards = 5
[2022-08-08T18:02:52,825][TRACE][o.e.x.i.h.ILMHistoryStore] [javaRestTest-0] queueing ILM history item for indexing [ilm-history-5]: [{"index":"index-zmmrkzfhht","policy":"policy-bEmKF","@timestamp":1659970972825,"index_age":12608,"success":true,"state":{"phase":"warm","phase_definition":"{\"policy\":\"policy-bEmKF\",\"phase_definition\":{\"min_age\":\"0ms\",\"actions\":{\"shrink\":{\"number_of_shards\":5}}},\"version\":1,\"modified_date_in_millis\":1659970962968}","action_time":"1659970968847","phase_time":"1659970966014","action":"shrink","step":"check-target-shards-count","creation_date":"1659970960217","step_time":"1659970972076"}}]

## We change the policy before even the condition is never even evaluated.

[2022-08-08T18:02:52,825][INFO ][o.e.x.i.a.TransportPutLifecycleAction] [javaRestTest-0] updating index lifecycle policy [policy-bEmKF]
[2022-08-08T18:02:52,826][DEBUG][o.e.x.c.i.PhaseCacheManagement] [javaRestTest-0] [index-zmmrkzfhht] updated policy [policy-bEmKF] contains the same phase step keys and can be refreshed
[2022-08-08T18:02:52,826][TRACE][o.e.x.c.i.PhaseCacheManagement] [javaRestTest-0] [index-zmmrkzfhht] updating cached phase definition for policy [policy-bEmKF]
[2022-08-08T18:02:52,826][DEBUG][o.e.x.c.i.PhaseCacheManagement] [javaRestTest-0] refreshed policy [policy-bEmKF] phase definition for [1] indices

## We check the condition for the first time but the target shard count is already outdated

[2022-08-08T18:02:53,406][ERROR][o.e.x.c.i.CheckTargetShardsCountStep] [javaRestTest-0] #78460: Policy has different target number of shards in cluster state 2 vs what will be executed 5.
[2022-08-08T18:02:53,441][DEBUG][o.e.x.c.i.CheckTargetShardsCountStep] [javaRestTest-0] lifecycle action of policy [policy-bEmKF] for index [index-zmmrkzfhht] cannot make progress because the target shards count [5] must be a factor of the source index's shards count [4]
```

**Impact**
We do not think that the impact is that big for production clusters because there are many more cluster state updates. However, it might cause some inconvenience to the users who fixed a policy and do not see the effect as soon as they could have.

**The fix**
Our proposed fix is to not provide the target shard count upon task creation but to retrieve from the cluster state. This way we ensure it will have the newest value.

**Future work**
Currently for every cluster state we go through all the indices and we check if any step needs to be executed. This doesn't scale well. We would like to try to switch to a more efficient model potentially with a cluster state observer. Issue will be created soon.

Resolves #78460


gmarouli:
@elasticmachine update branch

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/WithTargetNumberOfShards.java
```diff
@@ -0,0 +1,17 @@
...
+ * 2.0.
+ */
+
+package org.elasticsearch.xpack.core.ilm;
+
+/**
+ * The CheckTargetShardsCountStep requires for an action to provide the target number of shards. Any action
+ * implementing this interface can have the CheckTargetShardsCountStep step.
+ */
+public interface WithTargetNumberOfShards {
```
original-brownbear, line 14: Why do we need this interface? It seems we only have `ShrinkAction` implement it? Why not do the `instance of` check on `ShrinkAction`?

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/CheckTargetShardsCountStep.java
```diff
@@ -69,4 +63,22 @@ public Result isConditionMet(Index index, ClusterState clusterState) {
 
         return new Result(true, null);
     }
+
+    private Integer getTargetNumberOfShards(String policyName, ClusterState clusterState) {
+        IndexLifecycleMetadata indexLifecycleMetadata = clusterState.metadata().custom(IndexLifecycleMetadata.TYPE);
+        LifecycleAction lifecycleAction = indexLifecycleMetadata.getPolicyMetadatas()
```
original-brownbear, line 69: More of a NIT: Should we be a little more careful here to avoid NPEs if the policy got concurrently modified/deleted?

original-brownbear (commented):
Looks just fine to me in principle, just one point on the seemingly redundant interface that was added here.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/WithTargetNumberOfShards.java
gmarouli: Hm, I tried to generalize it because from the name of the step there is not association with the Shrink action, but we take the actual implementation of the step, I think it is actually pretty specific to the Shrink action. I will remove it.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/CheckTargetShardsCountStep.java
gmarouli: Hm, good point. We should, but then we need to see how to handle it. I do not have a clear picture yet how we handle the deletion or if the step got removed. Do I remember correctly that keep a cached version of the step being executed in the index metadata? Should we fallback there?

original-brownbear: I think throwing an appropriate exception that explains what happened is just fine here. No need to add additional logic, just figured it'd be nice to avoid a possible NPE and replace it with an easy to interpret exception.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/WithTargetNumberOfShards.java
gmarouli: Removed

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/ilm/CheckTargetShardsCountStepTests.java
```diff
@@ -88,4 +118,47 @@ public void testStepIncompleteIfTargetShardsCountNotValid() {
...
+            .numberOfShards(10)
+            .numberOfReplicas(randomIntBetween(0, 5))
+            .build();
+
+        Phase phase = new Phase(
+            checkTargetShardsCountStep.getKey().getPhase(),
+            TimeValue.ZERO,
+            singletonMap(checkTargetShardsCountStep.getKey().getAction(), new ReadOnlyAction())
+        );
+        LifecyclePolicy lifecyclePolicy = new LifecyclePolicy(policyName, singletonMap(phase.getName(), phase));
```
original-brownbear, line 138: NIT: why not use `Map.of(` throughout (here and in the other usage of the singleton map below) like you also used below to keep things consistent? :)

original-brownbear (approved):
LGTM, but maybe wait for Andrei's review as well :)
Also, I still wouldn't mind a nicer exception instead of an NPE for a missing policy below but that's optional too from my end

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/CheckTargetShardsCountStep.java
gmarouli: I see, I am still doubting if that is sufficient though. What concerns me in this approach is that it will throw errors and get stuck there permanently. That will also change the code behavior compared to what it does now. 
If this would have happened in the current set-up, the check would yield either true or false and it would continue. That's what I am trying to achieve by looking for a graceful fallback. Don't you think that's worth it?

original-brownbear: That's a fair point if we get stuck in error permanently. I thought other functionality would take care of clearing things up the policy disappears? (maybe that's something to fix separately if not?)
That said, that's a different issue from what we're working on here IMO. Here I was just aiming at getting an easy to understand exception.

> the check would yield either true or false and it would continue. 

Right but then we'd throw on the next step anyway (at least we should) wouldn't we so likely not really a change from the user's perspective?

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/ilm/CheckTargetShardsCountStepTests.java
gmarouli: I will, sloppy copy paste ;)

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/CheckTargetShardsCountStep.java
gmarouli: True, I need to check what happens now and how can we nicely deal with it, if it's an error in this step or the next then definitely an error is sufficient. If there is a way of handling the deleted/changed policy during a step transition, then we need to not fail and move to that transition. @andreidan can you shed some light?

andreidan: We don't allow deleting a policy whilst it's used by any index. 
Equally, steps execute based on the cached phase. The cached phase is refreshed if it's "safe" (safe currently means the number of steps in the new phase == number of steps in the cached phase - so the action cannot disappear from underneath our feet https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/PhaseCacheManagement.java#L243)

x-pack/plugin/core/src/test/java/org/elasticsearch/xpack/core/ilm/CheckTargetShardsCountStepTests.java
gmarouli: Fixed

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/CheckTargetShardsCountStep.java
andreidan: Thanks for looking into this failure Mary

I think this will potentially break our contract towards the phase we cache for execution.

For e.g. let's say an index is in the `warm` phase that currently has actions `shrink to 3 shards and forcemerge`. Our index is in the `shrink` action. 
A user updates the `warm` phase to `shrink to 1 shard` (also removing `forcemerge`). Since we removed an action, all the indices currently in the `warm` phase MUST execute the **previously defined** (and cached) phase - to honour the started flow https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/PhaseCacheManagement.java#L243
So an index in the `warm` phase will continue to execute `shrink to 3 and forcemerge`.

With _this_ change, the user will pick up a `shrink` action to a different target (1 shard) and the `forcemerge` will also be executed (as we didn't update the cached phase). 

I don't think we should make this change, but rather rework the test.

andreidan (changes requested):

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/CheckTargetShardsCountStep.java
gmarouli: @andreidan if I understand correctly then the safest option that fixes also the outdated step count issue it to read it from the cached phase, right?
Assuming that if you update only the target shard count then the cached phase will be updated because the number of  steps remained the same.

andreidan: ++ that's a safe option

gmarouli: What if we combine the two approaches to be a bit more efficient. What if we compare if the number of shards in the cluster state and in the step are the same (that is probably the most common case). If they are different then we try to parse the cached phase in an effort to improve user experience and we accept the performance hit since this will probably will not be a very common situation?

andreidan: @gmarouli That's an interesting idea. 

Do you think this is worth doing though? The `ClusterStateWaitStep` will eventually get the correct value and this affects concurrent updates only. 

IMO we should fix the test and trigger a cluster state update if the test fails just to make sure it's not run into this race condition.

gmarouli: You are probably right..... It's probably not worth the trouble... I wanted to explore all the options before I change the test instead for the race condition.

gmarouli:
@elasticmachine update branch

gmarouli:
@elasticmachine update branch

x-pack/plugin/ilm/qa/multi-node/src/javaRestTest/java/org/elasticsearch/xpack/TimeSeriesRestDriver.java
```diff
@@ -396,6 +398,18 @@ public static String getSnapshotState(RestClient client, String snapshot) throws
         return (String) snapResponse.get("state");
     }
 
+    @Nullable
+    public static String waitAndGetShrinkIndexNameWithExtraClusterStateChange(RestClient client, String originalIndex)
```
andreidan, line 402: Can you please document `the why` behind this? (the cluster state batching and such)

andreidan (approved):
LGTM thanks for fixing this

x-pack/plugin/ilm/qa/multi-node/src/javaRestTest/java/org/elasticsearch/xpack/TimeSeriesRestDriver.java
gmarouli: Ah very good point! On it.

gmarouli:
@elasticmachine update branch

PR #89177, opened 2022-08-08
Downsampling: copy`index.hidden` setting from source

csoulios:
Rollup indices are initially created as hidden (`index.hidden: true`). 

At the end of the rollup process, we must set this setting to the value the source index has


x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
```diff
@@ -267,6 +267,10 @@ protected void masterOperation(
                             if (sourceIndexMetadata.getNumberOfReplicas() > 0) {
                                 settings.put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, sourceIndexMetadata.getNumberOfReplicas());
                             }
+                            // Setting index.hidden has been initially set to true. We revert this to the value of the source index
+                            if (sourceIndexMetadata.isHidden() == false) {
+                                settings.put(IndexMetadata.SETTING_INDEX_HIDDEN, false);
+                            }
```
csoulios, line 273: There's a small catch here. If the setting `index.hidden` was set at all in the source index, the default value is `false`. Here we set the value explicitly to `false`. Hopefully, this does not cause any problems.

x-pack/plugin/rollup/src/test/java/org/elasticsearch/xpack/rollup/v2/RollupActionSingleNodeTests.java
```diff
@@ -805,6 +805,9 @@ private void assertRollupIndexSettings(String sourceIndex, String rollupIndex, G
             indexSettingsResp.getSetting(rollupIndex, IndexMetadata.SETTING_NUMBER_OF_REPLICAS)
         );
         assertEquals("true", indexSettingsResp.getSetting(rollupIndex, "index.blocks.write"));
+
+        boolean isSourceHidden = indexSettingsResp.getSetting(sourceIndex, IndexMetadata.SETTING_INDEX_HIDDEN) == "true";
```
salvatore-campagna, line 809: Maybe

```
if ("true".equals(indexSettingsResp.getSetting(sourceIndex, IndexMetadata.SETTING_INDEX_HIDDEN))) {
...
}
```

salvatore-campagna:
Just a minor comment. For the rest LGTM.

x-pack/plugin/rollup/src/test/java/org/elasticsearch/xpack/rollup/v2/RollupActionSingleNodeTests.java
csoulios: Fixed in 625f86acce7de12945efeaa359ed0696554881bf

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
dakrone: Could we do something like 

```java
if (sourceIndexMetadata.isHidden() == false && sourceIndexMetadata.getSettings().keySet().contains(IndexMetadata.SETTING_INDEX_HIDDEN)) {
    settings.put(IndexMetadata.SETTING_INDEX_HIDDEN, false);
}
```

To only set the setting if it does indeed exist on the source as explicitly set?

dakrone (approved):
LGTM, I left one suggestion though, about how we might be able to do it without the explicit setting.

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
csoulios: When the rollup index is created setting `index.hidden` is set to true. So, after rollup ends we want to revert this value.

I changed the code so that it removes setting (calling `setNull()`) if it is not set at all in the source index.

salvatore-campagna (approved):
LGTM

x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java
dakrone: Sounds good to me :+1:

PR #89179, opened 2022-08-08
Remove needless optimization ShardRouting.asList

original-brownbear:
This iterator is never used in hot code (there seems to only be a single production code usage for it),
no need to cache a list here just for it.


idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

PR #89182, opened 2022-08-08
Fixing internal action names

masseyke:
The names of the internal actions used by CoordinationDiagnosticsService did not begin with "internal:", so they could not be used in the system context.


andreidan (approved):
LGTM, thanks for fixing this Keith

PR #89190, opened 2022-08-09
Deduplicate ShardRouting instances when building ClusterInfo

original-brownbear:
The equality checks on these in `DiskThresholdDecider` become very expensive
during reroute in a large cluster. Deduplicating these when building the `ClusterInfo`
saves more than 2% CPU time during many-shards benchmark bootstrapping because
the lookup of the shard data path by shard-routing mostly hit instance equality.
Also, this saves a little memory.

This PR also moves the callback for building `ClusterInfo` from the stats response to
the management pool as it is now more expensive (though the overall CPU use from it is trivial
relative to the cost savings during reroute) and was questionable to run on
a transport thread in a large cluster to begin with.

relates #77466

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/InternalClusterInfoService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/InternalClusterInfoService.java
@@ -18,7 +18,6 @@ import org.elasticsearch.action.admin.cluster.node.stats.NodesStatsResponse;
 import org.elasticsearch.action.admin.indices.stats.IndicesStatsRequest;
 import org.elasticsearch.action.admin.indices.stats.IndicesStatsResponse;
 import org.elasticsearch.action.admin.indices.stats.ShardStats;
-import org.elasticsearch.action.support.DefaultShardOperationFailedException;
 import org.elasticsearch.action.support.IndicesOptions;
 import org.elasticsearch.action.support.PlainActionFuture;
 import org.elasticsearch.action.support.ThreadedActionListener;
--- a/server/src/main/java/org/elasticsearch/cluster/InternalClusterInfoService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/InternalClusterInfoService.java
@@ -211,8 +211,7 @@ public class InternalClusterInfoService implements ClusterInfoService, ClusterSt
 
                                 if (indicesStatsResponse.getShardFailures().length > 0) {
                                     final Set<String> failedNodeIds = new HashSet<>();
-                                    for (final DefaultShardOperationFailedException shardFailure : indicesStatsResponse
-                                        .getShardFailures()) {
+                                    for (final var shardFailure : indicesStatsResponse.getShardFailures()) {
                                         if (shardFailure.getCause()instanceof final FailedNodeException failedNodeException) {
                                             if (failedNodeIds.add(failedNodeException.nodeId())) {
                                                 logger.warn(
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/elasticsearch/cluster/routing/allocation/decider/DiskThresholdDecider.java
```diff
@@ -605,14 +605,6 @@ String getPath() {
             return diskUsage.getPath();
         }
 
-        String getNodeId() {
```
original-brownbear, line 608: unused stuff

server/src/test/java/org/elasticsearch/cluster/InternalClusterInfoServiceSchedulingTests.java
```diff
@@ -105,6 +105,7 @@ protected PrioritizedEsThreadPoolExecutor createThreadPoolExecutor() {
             setFlagOnSuccess(becameMaster2)
         );
         runUntilFlag(deterministicTaskQueue, becameMaster2);
+        deterministicTaskQueue.runAllRunnableTasks();
```
original-brownbear, line 108: We now sometimes had a runnable (but failing on a block) task for the internal cluster info service ready to execute here that would mess up the count assertions in the loop below, so draining the executor here.

server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java
```diff
@@ -671,7 +671,7 @@ public boolean isRelocationSourceOf(ShardRouting other) {
     /** returns true if the current routing is identical to the other routing in all but meta fields, i.e., unassigned info */
     public boolean equalsIgnoringMetadata(ShardRouting other) {
         return primary == other.primary
-            && Objects.equals(shardId, other.shardId)
+            && shardId.equals(other.shardId)
```
original-brownbear, line 674: This is never `null` so no need for the extra indirection that (weirdly enough) doesn't inline here.

server/src/main/java/org/elasticsearch/cluster/InternalClusterInfoService.java
```diff
@@ -191,76 +196,93 @@ private void fetchIndicesStats() {
...
-                                if (failedNodeIds.add(failedNodeException.nodeId())) {
-                                    logger.warn(
-                                        () -> format("failed to retrieve shard stats from node [%s]", failedNodeException.nodeId()),
-                                        failedNodeException.getCause()
-                                    );
+            client.admin()
+                .indices()
+                .stats(
+                    indicesStatsRequest,
+                    new ThreadedActionListener<>(
```
original-brownbear, line 203: Couldn't really find a nice way of doing this without a big changeset from the indents, no changes to the code here other than the deduplication though.

DaveCTurner: ... and the fact that we now fork to `MANAGEMENT`? Presumably that's because the deduplication is too expensive to run on the transport thread right?

DaveCTurner (approved):
LGTM

server/src/main/java/org/elasticsearch/cluster/InternalClusterInfoService.java
original-brownbear: Yea, it's not a massive slowdown here from this change (maybe like 30%) but even before this change running this on a transport thread wasn't ideal for a 50k shards cluster so I didn't want to make that situation worse and changed it in this PR.

original-brownbear:
Thanks David!

PR #89197, opened 2022-08-09
[ML] Extract downscale to zero logic into its own method

dimitris-athanasiou:
This commit extracts the logic that downscales to zero when
the ML autoscaling decider detects there are no ML tasks at all.


droberts195 (approved):
LGTM

PR #89198, opened 2022-08-09
Weaken language about "low-latency" networks

DaveCTurner:
Today we say that voting-only nodes require a "low-latency" network.
This term has a specific meaning in some operating environments which is
different from our intended meaning. To avoid this confusion this commit
removes the absolute term "low-latency" in favour of describing the
requirements relative to the user's own performance goals.


henningandersen (approved):
LGTM, thanks David.

PR #89199, opened 2022-08-09
geo_line aggregation returns a geojson point when the resulting line has only one point

### LINKED ISSUE #85748: GeoLine aggregation can output corrupt geojson: a line with just 1 point
Reproduction (from the docs, but removed 2 of the 3 docs):

```
PUT test
{
    "mappings": {
        "dynamic": "strict",
        "_source": {
            "enabled": false
        },
        "properties": {
            "my_location": {
                "type": "geo_point"
            },
            "group": {
                "type": "keyword"
            },
            "@timestamp": {
                "type": "date"
            }
        }
    }
}

POST /test/_bulk?refresh
{"index": {}}
{"my_location": {"lat":37.3450570, "lon": -122.0499820}, "@timestamp": "2013-09-06T16:00:36"}

POST /test/_search?filter_path=aggregations
{
  "aggs": {
    "line": {
      "geo_line": {
        "point": {"field": "my_location"},
        "sort": {"field": "@timestamp"}
      }
    }
  }
}
```

This reports:

```
      "geometry" : {
        "coordinates" : [
          [
            -122.049982,
            37.345057
          ]
        ],
        "type" : "LineString"
      },
```

which is invalid as a line should have at least 2 points, e.g. elasticsearch errors if you try to index a line with just 1 point.

If you use the geo-line aggregation in a transform, transform fails if it stumbles upon such a document.

It seems to me that the geo-line aggregation should not report these points or it should report it as points.

iverase:
geo_line aggregation currently returns a geojson line containing one point if there is only one point in the result of the aggregation. That result is invalid geo_json and parsers will throw an error when parsing it. This PR changes the geojson output to return a point instead in the cases a geo_line aggregation only contains one point.

fixes https://github.com/elastic/elasticsearch/issues/85748

### DIFF
--- a/docs/changelog/89199.yaml
+++ b/docs/changelog/89199.yaml
@@ -1,0 +1,7 @@
+pr: 89199
+summary: Geo_line aggregation returns a geojson point when the resulting line has
+  only one point
+area: Geo
+type: bug
+issues:
+ - 85748
--- a/x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/InternalGeoLine.java
+++ b/x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/InternalGeoLine.java
@@ -227,8 +227,13 @@ public class InternalGeoLine extends InternalAggregation implements GeoShapeMetr
             );
         }
         final Map<String, Object> geoJSON = new HashMap<>();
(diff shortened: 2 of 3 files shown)


iverase:
I open https://github.com/elastic/elasticsearch/issues/89206

hendrikmuhs (approved):
LGTM, thanks for fixing.

PR #89200, opened 2022-08-09
Correct ctx.op value to valid 'noop'

adammichalik:
The documentation mentions the `noop` value ("Otherwise it does nothing (`noop`)"), however, the value used in the example script is `none`. This change corrects the value in the example script to `noop`.


stu-elastic (changes requested):
Hi @adammichalik, thanks for the contribution.  PRs should target `main` unless they are a version specific bug-fix.

In fact, before 8.4, the correct op for skipping _was_ `none` but long ago we added leniency for unknown operations, which is why `noop` also works.

In 8.3 and previous, we emitted an warning log for `noop`:

> Used upsert operation [noop] for script [if (ctx._source.tags.contains(params["tag"])) { ctx.op = "delete" } else { ctx.op = "noop" }], doing nothing...

As of 8.4, we _do_ accept `noop` so this change is fine for `main` but we would not want to backport it.

adammichalik:
@stu-elastic Thanks for the explanation. I tried editing this PR to change the target branch, but only made a mess ;) I will close this one and open a new one, targeting the `main` branch.

PR #89207, opened 2022-08-09
Verify auto follower recover after full leader cluster restart

### LINKED ISSUE #86672: Verify auto follower recover after full leader cluster restart
### Description

Currently we have a test that verifies auto following behavior regarding leader and follower cluster restart.

However remote cluster is disabled before the leader restart and configured again after:

https://github.com/elastic/elasticsearch/blob/33b3ccaabc6960f02c005b6bc14a1c527a24875b/x-pack/plugin/ccr/src/internalClusterTest/java/org/elasticsearch/xpack/ccr/RestartIndexFollowingIT.java#L92-L96

This happens as leader cluster changes the port during the restart and otherwise would become not accessible.

idegtiarenko:
Re-setting remote connection is no longer required as cluster port is not
changing after restart.

Closes: #86672


tlrx (approved):
LGTM

tlrx:
(Thanks Ievgen)

PR #89210, opened 2022-08-09
Check for polygon self-intersections in ShapeFieldMapper

iverase:
In LUCENE-9538:, it was added the possibility to detect polygon self-intersection in the Lucene tessellator to improve the usability and provide feedback to the clients. This ability has been already enable for the geo case but we forgot adding for the cartesian case. This PR set the sef-intersection flag to true so we can give better error message to the users.

I am adding as well a test for the geo case as it seems to be missing.


iverase:
Looks like a transient error:

iverase:

craigtaverner (approved):
LGTM

PR #89212, opened 2022-08-09
[Transform] add an unattended mode setting to transform

hendrikmuhs:
Add an unattended mode setting. This will change how transform reacts on certain error types and lets transform run without failing.


x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/TransformFailureHandler.java
```diff
@@ -0,0 +1,276 @@
...
+ *    parts should be read from the context object instead
+ */
+class TransformFailureHandler {
+    private static final Logger logger = LogManager.getLogger(TransformFailureHandler.class);
+    public static final int LOG_FAILURE_EVERY = 10;
+    private final TransformAuditor auditor;
+    private final String transformId;
+    private final TransformContext context;
+
+    // protected for unit tests
```
droberts195, line 41: This comment implies that it's _only_ the unit tests that construct this class directly, when in fact another production class does too.

The constructor is package private and not commenting on that at all causes least confusion in my opinion.

droberts195 (approved):
LGTM

x-pack/plugin/transform/src/main/java/org/elasticsearch/xpack/transform/transforms/TransformFailureHandler.java
hendrikmuhs: good spot. I think I did not comment the constructor, but some member which I deleted later. Re-formatting deleted the blankline and somehow this ended up here. 

It makes no sense, of course I use this.

PR #89214, opened 2022-08-09
Fixing remote master stability request when there has never been an elected master

masseyke:
This fixes an edge case in the master stability polling code from #89014. If there has not been an elected master node for the entire life of a non-master-eligible node, then `clusterChanged()` will have never been called on that node, so `beginPollingRemoteMasterStabilityDiagnostic()` will have never been called. And even though the node might know of some master-eligible nodes, it will never have requested diagnostic information from them. This PR adds a call to `beginPollingRemoteMasterStabilityDiagnostic` in `CoordinationDiagnosticsService`'s constructor to cover this edge case. In almost all cases, `clusterChanged()` will be called within 10 seconds so the polling will never occur. However if there is no master node then there will be no cluster changed events, and `clusterChanged()` will not be called, and the results of the polling will likely be useful.
This PR has several possibly controversial pieces of code. I'm listing them here with some discussion:

1. Because there is now a call to `beginPollingRemoteMasterStabilityDiagnostic()` in the ~~constructor~~ object's initialization code, `beginPollingRemoteMasterStabilityDiagnostic()` is no longer solely called from the cluster change thread. However, this call happens before the object is registered as a cluster service listener, so there is no new thread safety concern.
2. Because there is now a call to `beginPollingRemoteMasterStabilityDiagnostic()` in the ~~constructor~~ object's initialization code, we have to explicitly switch to the system context so that the various transport requests work in secure mode.
3. ~~When we're in the constructor, we don't actually know yet whether we're a master eligible node or not, so we kick off `beginPollingRemoteMasterStabilityDiagnostic()` for all node types, including master-eligible nodes. This will be fairly harmless for master eligible nodes though. In the worst case, they'll retrieve some information that they'll never use. This explains why `clusterChanged()` now cancels polling even if we are on a master eligible node.~~
4. ~~It is now possible that we use `clusterService.state()` before it is ready when we're trying to get the list of master-eligible peers. In production mode this method returns null, so we can check that before using it. If assertions are enabled in the JVM, just calling that method throws an `AssertionError`. I'm currently catching that with the assumption that it is harmless because there does not seem to be a way around it (without even further complicating code).~~
5. ~~It is now possible that we call `transportService.sendRequest()` before the transport service is ready. This happens if the server is initializing unusually slowly (i.e. it takes more than 10 seconds to complete the `Node` constructor) and if assertions are enabled. I don't see a way around this without further complicating the code, so I'm catching `AssertionError` and moving on, with the assumption that it will work 10 seconds later when it runs again. I'm also catching and storing `Exception`, which I think I should have been doing before anyway.~~

Note: Points 3, 4, and 5 are no longer relevant because I moved the call to `beginPollingRemoteMasterStabilityDiagnostic()` out of the constructor, and am now calling it after the transport service and cluster state have been initialized.

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
@@ -844,7 +844,6 @@ public class CoordinationDiagnosticsService implements ClusterStateListener {
 
     void beginPollingRemoteMasterStabilityDiagnostic() {
         assert ThreadPool.assertInSystemContext(transportService.getThreadPool());
-        // Note that this method must be called from the system context because it calls internal transport actions
         AtomicReference<Scheduler.Cancellable> cancellableReference = new AtomicReference<>();
         AtomicReference<RemoteMasterHealthResult> resultReference = new AtomicReference<>();
         remoteCoordinationDiagnosisTask = cancellableReference;
--- a/server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
@@ -171,6 +171,13 @@ public class CoordinationDiagnosticsService implements ClusterStateListener {
         this.nodeHasMasterLookupTimeframe = NODE_HAS_MASTER_LOOKUP_TIMEFRAME_SETTING.get(clusterService.getSettings());
         this.unacceptableNullTransitions = NO_MASTER_TRANSITIONS_THRESHOLD_SETTING.get(clusterService.getSettings());
         this.unacceptableIdentityChanges = IDENTITY_CHANGES_THRESHOLD_SETTING.get(clusterService.getSettings());
+    }
+
+    /**
+     * This method completes the initialization of the CoordinationDiagnosticsService. It kicks off polling for remote master stability
+     * results on non-master-eligible nodes, and registers the service as a cluster service listener on all nodes.
+     */
+    public void start() {
         /*
          * This is called here to cover an edge case -- when there are master-eligible nodes in the cluster but none of them has been
          * elected master. In the most common case this node will receive a ClusterChangedEvent that results in this polling being
@@ -179,10 +186,12 @@ public class CoordinationDiagnosticsService implements ClusterStateListener {
          * results will always be harmlessly ignored. Note that beginPollingRemoteMasterStabilityDiagnostic results in several internal
          * transport actions being called, so it must run in the system context.
          */
-        final ThreadContext threadContext = transportService.getThreadPool().getThreadContext();
-        try (ThreadContext.StoredContext ignored = threadContext.stashContext()) {
-            threadContext.markAsSystemContext();
-            beginPollingRemoteMasterStabilityDiagnostic();
+        if (clusterService.localNode().isMasterNode() == false) {
+            final ThreadContext threadContext = transportService.getThreadPool().getThreadContext();
+            try (ThreadContext.StoredContext ignored = threadContext.stashContext()) {
+                threadContext.markAsSystemContext();
+                beginPollingRemoteMasterStabilityDiagnostic();
+            }
         }
         clusterService.addListener(this);
     }
@@ -621,17 +630,8 @@ public class CoordinationDiagnosticsService implements ClusterStateListener {
             }
         });
         // Coordinator does not report the local node, so add it:
-        try {
-            if (clusterService.state() != null && clusterService.localNode() != null && clusterService.localNode().isMasterNode()) {
-                masterEligibleNodes.add(clusterService.localNode());
-            }
-        } catch (AssertionError e) {
-            /*
-             * Unfortunately there does not seem to be a way to tell if clusterService.state() will throw an AssertionError if assertions
-             * are enabled. In production with assertions disabled clusterService.state() will be null rather than throwing an
-             * AssertionError.
-             */
-            logger.trace("Attempted to use cluster state before it existed, and assertions are enabled in the JVM", e);
+        if (clusterService.localNode().isMasterNode()) {
+            masterEligibleNodes.add(clusterService.localNode());
         }
         return masterEligibleNodes;
     }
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -170,6 +171,19 @@ public CoordinationDiagnosticsService(
...
+         * cancelled almost immediately. If that does not happen, then we do in fact need to be polling. Unfortunately there is no way to
+         * tell at this point whether this node is master-eligible or not, so we kick this off regardless. On master-eligible nodes the
+         * results will always be harmlessly ignored. Note that beginPollingRemoteMasterStabilityDiagnostic results in several internal
+         * transport actions being called, so it must run in the system context.
+         */
+        final ThreadContext threadContext = transportService.getThreadPool().getThreadContext();
+        try (ThreadContext.StoredContext ignored = threadContext.stashContext()) {
+            threadContext.markAsSystemContext();
+            beginPollingRemoteMasterStabilityDiagnostic();
+        }
```
andreidan, line 186: Doing business logic in the constructor is an anti-pattern as you're racing against the initializer thread (e.g. an uncomplete`this` could escape in the async code)

I think  the `clusterService.addListener(this);` calls should be done outside the constructor too (as it publishes `this`)

Should we have an `init` or `start` method where we do these things and call this new method from the outside after we called the constructor?

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -654,12 +677,10 @@ public void clusterChanged(ClusterChangedEvent event) {
         } else {
             cancelPollingClusterFormationInfo();
         }
-        if (clusterService.localNode().isMasterNode() == false) {
-            if (currentMaster == null) {
-                beginPollingRemoteMasterStabilityDiagnostic();
-            } else {
-                cancelPollingRemoteMasterStabilityDiagnostic();
-            }
+        if (currentMaster == null && clusterService.localNode().isMasterNode() == false) {
```
andreidan, line 680: So both `cancel...` methods are called every time there's a non-null master right? 

Can we combine and simplify the if statemets to reflect this? 
ie.
```
   if master == null { 
       if isMasterEligible { pollABC } else { pollEFG }
    } else {
       cancelABC
       cancelEFG
    }
```

What do you think?

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -819,7 +840,7 @@ private Scheduler.Cancellable fetchClusterFormationInfo(
     }
 
     void beginPollingRemoteMasterStabilityDiagnostic() {
-        assert ThreadPool.assertCurrentThreadPool(ClusterApplierService.CLUSTER_UPDATE_THREAD_NAME);
+        // Note that this method must be called from the system context because it calls internal transport actions
```
andreidan, line 843: could this be coded as an assertion on the thread context? ie. `ThreadContext#isSystemContext`

andreidan (commented):
Thanks for the great description Keith and for working on fixing these tricky cases 

This generally looks good, left a few suggestions

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: It's a little more complicated than that because our test code doesn't run in the system context. I'll add a method to ThreadPool to work with unit tests like we do for assertCurrentThreadPool.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: Yeah I can do that.

masseyke: Actually by doing that I can call init() much later in Node's startup, and avoid several of the other problems in this PR.

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
masseyke: Since I moved the call to `beginPollingRemoteMasterStabilityDiagnostic()` out of the constructor, I was able to check whether a node was master-eligible before calling `beginPollingRemoteMasterStabilityDiagnostic()`. So now we only need to cancel it for non-master-eligible nodes. So I put it back the way it used to be.

server/src/main/java/org/elasticsearch/node/Node.java
```diff
@@ -1323,6 +1323,7 @@ public Node start() throws NodeValidationException {
         assert clusterService.localNode().equals(localNodeFactory.getNode())
             : "clusterService has a different local node than the factory provided";
         transportService.acceptIncomingRequests();
+        injector.getInstance(CoordinationDiagnosticsService.class).start();
```
andreidan, line 1326: Shall we add a comment why it's important for this service to be started here?

andreidan (approved):
LGTM thanks for fixing this Keith

server/src/main/java/org/elasticsearch/cluster/coordination/CoordinationDiagnosticsService.java
```diff
@@ -967,16 +991,20 @@ private <R extends ActionResponse, T> Scheduler.Cancellable sendTransportRequest
...
+                        transportActionRequest,
+                        TransportRequestOptions.timeout(transportTimeout),
+                        new ActionListenerResponseHandler<>(
+                            ActionListener.runBefore(fetchRemoteResultListener, () -> Releasables.close(releasable)),
+                            transportActionType.getResponseReader()
+                        )
+                    );
+                } catch (Exception e) {
+                    responseConsumer.accept(responseTransformationFunction.apply(null, e));
+                }
```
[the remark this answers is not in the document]
DaveCTurner: We only catch a `NodeNotConnectedException` thrown from `ConnectionManager#getConnection` so in theory something else could escape from here, although in practice that's the only exception that any implementations will throw. We should probably just `catch Exception` here to make sure, rather than putting the onus on callers. Would you open an issue for the distrib team to fix that?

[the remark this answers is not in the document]
andreidan: Thanks David. Opened https://github.com/elastic/elasticsearch/issues/89274

PR #89216, opened 2022-08-09
Centroid aggregation for cartesian points and shapes

### LINKED ISSUE #90156: Centroid aggregation for Cartesian points and shapes
### Description

There is absolutely no support for cartesian data in aggregations. There is a meta-task for adding [Cartesian support for aggregations](https://github.com/elastic/elasticsearch/issues/87804), and this issue is the first aggregation to add, the `centroid` aggregation.

It is modelled on the existing [`geo_centroid`](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-geocentroid-aggregation.html) and as such could be called either:

* `cartesian_centroid`
* or just `centroid`

Sub-tasks:

* https://github.com/elastic/elasticsearch/pull/89388
* https://github.com/elastic/elasticsearch/pull/89216
* Documentation

craigtaverner:
Based on the groundwork to [Refactor GeoPoint and GeoShape with generics](https://github.com/elastic/elasticsearch/pull/89388) this PR provides support for centroid aggregations over `Cartesian` `point` and `shape` fields.

Closes [https://github.com/elastic/elasticsearch/issues/90156](https://github.com/elastic/elasticsearch/issues/90156)


x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/support/ShapeValuesSourceType.java
```diff
@@ -0,0 +1,37 @@
...
+import org.elasticsearch.search.aggregations.support.FieldContext;
+import org.elasticsearch.search.aggregations.support.ValueType;
+import org.elasticsearch.search.aggregations.support.ValuesSource;
+import org.elasticsearch.search.aggregations.support.ValuesSourceType;
+import org.elasticsearch.xcontent.ToXContentFragment;
+
+import java.io.IOException;
+
+public abstract class ShapeValuesSourceType<T extends ToXContentFragment> implements Writeable, ValuesSourceType {
+
```
iverase, line 23: Is this generic really needed? T seems unused here.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/support/ShapeValuesProvider.java
```diff
@@ -0,0 +1,38 @@
...
+import org.elasticsearch.index.fielddata.FieldData;
+import org.elasticsearch.index.fielddata.SortedBinaryDocValues;
+import org.elasticsearch.search.aggregations.support.ValuesSource;
+import org.elasticsearch.xcontent.ToXContentFragment;
+import org.elasticsearch.xpack.spatial.index.fielddata.ShapeValues;
+
+import java.io.IOException;
+
+public abstract class ShapeValuesProvider<T extends ToXContentFragment> extends ValuesSource {
+    public abstract ShapeValues<T> shapeValues(LeafReaderContext context);
```
iverase, line 21: I wonder if the generic here should be ShapeValues<?> instead

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/support/ShapeValuesSourceType.java
craigtaverner: Looks like a failed attempt to make generic code for the two source types. But now with the ElasticPoint interface, most of my generics are possible to delete, so I've been doing a big cleanup. It's getting simpler and simpler, which feels like the right direction for sure.

iverase: Looking into the ElasticPoint interface, maybe a better name would we Point2d?

craigtaverner: Indeed. I'm happy with all suggestions, as `ElasticPoint` was a really wild guess. Interestingly, as I've used it for a few days now, I've started to get used to the name, and started to think of it is as precursor to something involving CRS. So not just 2D, but 2 or 3D, and geo and cartesian, and in fact then any CRS. For this we would simply add a crs attribute to the interface, and either add an option z coordinate, or change the getX and getY into `double[] getCoordinates()`, or even move in a JTS direction with a `Coordinate` class. Then the Point class is simply tying together a Coordinate class to a CRS class.

For me, right now, simply having a single interface (named anything sensible) makes a _HUGE_ difference to the complexity of bring cartesian up to geo.

craigtaverner: Both concerns were addressed in the previous PR which is now merged. https://github.com/elastic/elasticsearch/pull/89388

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/support/ShapeValuesProvider.java
craigtaverner: Addressed in the previous PR, now merged: https://github.com/elastic/elasticsearch/pull/89388

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/support/CartesianShapeValuesSourceType.java
```diff
@@ -0,0 +1,114 @@
...
+        DocValueFormat docValueFormat,
+        AggregationContext context
+    ) {
+        // TODO: This code is almost entirely in common between Geo and Cartesian
+        CartesianShapeValuesSource shapeValuesSource = (CartesianShapeValuesSource) valuesSource;
+        final ShapeValues.ShapeValue missing = CartesianShapeValues.EMPTY.missing(rawMissing.toString());
+        return new CartesianShapeValuesSource() {
+            @Override
+            public CartesianShapeValues shapeValues(LeafReaderContext context) {
+                ShapeValues values = shapeValuesSource.shapeValues(context);
```
iverase, line 73: I would really like that CartesianShapeValuesSource returns a CartesianShapeValues. There are methods in ShapeValues that are geo specific and should only be present in GeoShapeValues. There is something that feels not right at the moment.

craigtaverner: The previous PR was supposed to make sure that ShapeValues had no geo-specific code, but I can already see that I left at least `CoordinateEncoder.Geo` there, so I'll review this and fix it. I suspect there will be more cases where the previous PR did not quite complete the generalization of the code.

craigtaverner: I notice also the use of `LatLonGeometry` and `LatLonGeometryRelationVisitor`. However, despite their names, those two classes are not geo-specific and look like they work perfectly well for cartesian data too. Perhaps they should be renamed?

iverase: The biggest problem is not in ShapeValues but ShapeValue. For example the method # relate(LatLonGeometry latLonGeometry) is for geo. 

I think GeoshapeValue and Cartisianshapevalue are meant to support different things and should not be represented with the same ShapeValue.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/script/CartesianPointFieldScript.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.Map;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * Script producing cartesian points.
+ */
+public abstract class CartesianPointFieldScript extends AbstractPointFieldScript<CartesianPoint> {
+    public static final ScriptContext<Factory> CONTEXT = newContext("point_field", Factory.class);
+
```
iverase, line 28: I see this is not used anywhere, maybe we should remove it until we decide to have runtime support

server/src/main/java/org/elasticsearch/script/AbstractFieldScript.java
```diff
@@ -31,7 +31,7 @@ public abstract class AbstractFieldScript extends DocBasedScript {
      */
     public static final int MAX_VALUES = 100;
 
-    static <F> ScriptContext<F> newContext(String name, Class<F> factoryClass) {
+    protected static <F> ScriptContext<F> newContext(String name, Class<F> factoryClass) {
```
iverase, line 34: This come from an unused class, let's remove everything? we can revisit later.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/script/CartesianPointFieldScript.java
craigtaverner: Agreed. I also remember we discussed with the painless team the option of building a completely different fields API for script for this, so none of the prototype code here is of value. I created an issue to discuss that at https://github.com/elastic/elasticsearch/issues/90182

server/src/main/java/org/elasticsearch/script/AbstractFieldScript.java
craigtaverner: Agreed. I also remember we discussed with the painless team the option of building a completely different fields API for script for this, so none of the prototype code here is of value. I created an issue to discuss that at https://github.com/elastic/elasticsearch/issues/90182

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/support/CartesianShapeValuesSourceType.java
craigtaverner: But that is what I meant above. The method `relate(LatLonGeometry)` does not have anything geo-specific in it. All that code, and the code it calls is specific to the Lucene `Component2D` mechanism and is identical between geo and cartesian. The only mistake here is that the classes are incorrectly named `LatLon...`, while they have nothing to do with geo, and are used for both geo and cartesian data. They all work directly on the `Geometry` and `Component2D` interfaces, both of which are used for both geo and cartesian data.

So I think I can _fix_ this by renaming the classes prefixed with `LatLon` something else. Then it will be more clear.

server/src/main/java/org/elasticsearch/script/AbstractFieldScript.java
craigtaverner: Fixed in https://github.com/elastic/elasticsearch/pull/89216/commits/3fa56032243e566da4f8f4c70beb7f964531a1ad

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/script/CartesianPointFieldScript.java
craigtaverner: Fixed in https://github.com/elastic/elasticsearch/pull/89216/commits/3fa56032243e566da4f8f4c70beb7f964531a1ad

server/src/main/java/org/elasticsearch/index/fielddata/GeoPointScriptDocValues.java
```diff
@@ -13,6 +13,7 @@
 import org.elasticsearch.script.GeoPointFieldScript;
 
 public final class GeoPointScriptDocValues extends AbstractSortedNumericDocValues {
+    // private final AbstractPointFieldScript<GeoPoint> script;
```
iverase, line 16: can we remove this?

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/index/fielddata/CartesianShapeValues.java
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Creates a new {@link CartesianShapeValues} instance
+     */
+    protected CartesianShapeValues() {
+        super(CoordinateEncoder.CARTESIAN, CartesianShapeValues.CartesianShapeValue::new, new CartesianShapeIndexer("missing"));
+    }
+
+    /** thin wrapper around a {@link GeometryDocValueReader} which encodes / decodes values using
+     * the Geo decoder */
+    public static class CartesianShapeValue extends ShapeValues.ShapeValue {
```
iverase, line 65: There is something that is not right here because ShapeValues.ShapeValue contains geo specific code. For example this means that CartesianShapeValue supports calling #relate(LatLonGeometry) which is not right. I think we ought to move the geo specific methods out of ShapeValue and only leave the generic ones?

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/support/CartesianShapeValuesSourceType.java
craigtaverner: I looked at this again and discovered that the generalization I did in the previous PR actually completed the necessary generalization. There is no need for instances of `CartesianShapeValues` or `GeoShapeValues` at all. The geo and cartesian behaviour had already been extracted and provided as parameters directly to the `ShapeValues` constructor. This meant I could simply remove those classes, although I kept them as factory classes with static methods. And I could move most of the common code directly into `ShapeValues`. Three blocks of duplicated code have been moved reducing code duplication quite a lot:

* Empty shape values
* Shape values based on BinaryDocValues
* Shape values based on wrapping other ShapeValues and replacing missing values

This improvement is pushed in https://github.com/elastic/elasticsearch/pull/89216/commits/678c3472602b3092a86399f4c828390d24ec65e7

craigtaverner: I have still not fixed the incorrectly named `LatLon*` classes that have nothing to do with Geo, and are used by both geo and cartesian. That could be done anytime. It is tech debt that precedes this work.

server/src/main/java/org/elasticsearch/index/fielddata/GeoPointScriptDocValues.java
craigtaverner: I believe I did that in the same commit that removed the script classes.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/index/fielddata/CartesianShapeValues.java
craigtaverner: There is no geo-specific code in `ShapeValues` or in `ShapeValues.ShapeValue`. I double checked and the only mistake is in class naming. For some reason classes are named `LatLon*` when they have no geo-specific behaviour. I commented on this above, and also pushed a new commit that further reduces the code duplication in `ShapeValues`.

If you can find any geo-specific method in `ShapeValues` or in `ShapeValues.ShapeValue`, please let me know, because I cannot find them. It was made generic in the previous PR, there is nothing geo-specific left.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/support/CartesianShapeValuesSourceType.java
iverase: I am getting confused, I think you are not working on top of the latest main, still exposing ShapeValues on geo / cartesian classes

craigtaverner: Yes, indeed. But I've fixed that now. Reverted the commit that moved back to GeoShapeValues.

craigtaverner: That commit was based on the assumption that ShapeValues has geo-specific code in it. And that is still not true. I remember when I approved that before I said that I wanted to re-evaluate my assumptions and reconsider that work once we did the Cartesian PR. Now this is the Cartesian PR and it looks to me like the generalization is in fact a very good idea. We make a whole stack of code common between Cartesian and Geo. And this is all enabled by the Lucene Component2D interface which is what actually makes this stack common.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/support/ShapeValuesSource.java
```diff
@@ -16,8 +16,8 @@
 
 import java.io.IOException;
 
-public abstract class ShapeValuesSource<T extends ShapeValues<?>> extends ValuesSource {
-    public abstract T shapeValues(LeafReaderContext context);
+public abstract class ShapeValuesSource extends ValuesSource {
+    public abstract ShapeValues shapeValues(LeafReaderContext context);
```
iverase, line 20: I disagree here, please do not remove the generic. GeoShapeValuesSource should return GeoShapeValues.

iverase (changes requested):
Sorry, you just reverted the changes we did before. I don't want ShapeValue to be used indistinctly between geo and cartesian, they are different things.

x-pack/plugin/spatial/src/main/java/org/elasticsearch/xpack/spatial/search/aggregations/support/CartesianShapeValuesSourceType.java
iverase: I think you assumptions are wrong, please don't expose ShapeValues as being the same for geo and cartesian, they are not.

x-pack/plugin/src/yamlRestTest/resources/rest-api-spec/test/spatial/20_geo_centroid.yml
```diff
@@ -52,3 +52,58 @@
   - match: { aggregations.centroid.location.lat: 51.00982965203002 }
   - match: { aggregations.centroid.location.lon: 3.9662131341174245 }
   - match: { aggregations.centroid.count: 6 }
+
+---
+"Test cartesian_centroid aggregation on shape field":
```
iverase, line 57: Maybe we can rename this file to 10_centroid so it can contain geo and cartesian implementations?

x-pack/plugin/spatial/src/test/java/org/elasticsearch/xpack/spatial/util/GeoTestUtils.java
```diff
@@ -55,6 +56,14 @@ public static BinaryShapeDocValuesField binaryGeoShapeDocValuesField(String name
         return field;
     }
 
+    // TODO: Should this be in a cartesian specific test class, and if so, which?
+    public static BinaryShapeDocValuesField binaryCartesianShapeDocValuesField(String name, Geometry geometry) {
```
iverase, line 60: Agree, this method sees weird here, I am ok to address it in a follow up.

iverase (approved):
LGTM

PR #89218, opened 2022-08-09
Remove side-effects in streams in PrimaryShardAllocator

arteam:
`map` should be a side-effect free function, because it's a non-terminal operation.
If we want to have side effect, we should use `forEach` which is terminal.


arteam:
@elasticmachine update branch

original-brownbear (approved):

arteam:
Thanks Armin!

PR #89223, opened 2022-08-09
Ensure APM module is always installed in release test clusters

mark-vieira:
Our release builds have been failing because on [startup ES now checks](https://gradle-enterprise.elastic.co/s/yndbpn74kho24/console-log?task=:rest-api-spec:yamlRestTest#L7) to ensure the APM module is present in non-snapshot builds. To remedy this we simply need to ensure that the APM module is explicitly included in test distribution when performing release builds.

### DIFF
--- a/distribution/build.gradle
+++ b/distribution/build.gradle
@@ -177,7 +177,7 @@ project.rootProject.subprojects.findAll { it.parent.path == ':modules' }.each { 
   }
 
   distro.copyModule(processDefaultOutputsTaskProvider, module)
-  if (module.name.startsWith('transport-')) {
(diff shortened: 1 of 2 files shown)


mark-vieira:
@elasticmachine update branch

breskeby (approved):
would it make sense to always include the apm package and not just in release builds? would that maybe things overall simpler?

mark-vieira:
> would it make sense to always include the apm package and not just in release builds? would that maybe things overall simpler?

Simply, yeah, but in generally I think we should aim towards as light an integ-test distribution as possible, since in theory we'll get better cacheability of tests the smaller the footprint of the code we are testing.

PR #89233, opened 2022-08-10
Check DLS feature before accessing indices access control

slobodanadamovic:
This PR changes the order in which document level security feature is checked.
The indention is to avoid loading indices access control as much as possible.
In case the DLS feature is not enabled there is no need to loop and check DLS at all.


slobodanadamovic:
@elasticmachine update branch

PR #89234, opened 2022-08-10
Add the ability to run Rest integration tests with 1 allocated processor

fcofdez:
This commit adds a new system property `tests.configure_test_clusters_with_one_processor`
that configures Rest integration tests to run with 1 processor, this should help finding possible
deadlocks (if any) when the Elasticsearch nodes have `node.processors` set to 1.


fcofdez:
I haven't set it up for Rest tests as the nodes are configured in gradle and we don't have access to the test random seed there.

DaveCTurner:
> configured in gradle and we don't have access to the test random seed  

Could you do something like this?

https://github.com/elastic/elasticsearch/blob/main/x-pack/qa/multi-cluster-search-security/build.gradle#L25

DaveCTurner (approved):
LGTM

fcofdez:
I discussed this with @mark-vieira offline and we decided to define a new CI job that would run all the tests using 1 processor a few times a day, the arguments in favour of that approach were:
- Introducing this change based on randomness could make difficult to identify the root cause (is this test failure cased by a deadlock? or is it something else?) while if we get the same test failure in the specific CI job it would be easier to triage.
- There were some concerns about the slowness of tests running with 1 processor configured, but for most tests that shouldn't be problematic.

As cons:
- We would run the tests with 1 processor less often, meaning that possible problems would take longer to surface.

The idea is to change this PR and let the test code accept a new system property that would set `node.processors` to `1` and @mark-vieira will create the CI job to run the tests with that flag.  

Are there any objections to this @henningandersen @DaveCTurner?

DaveCTurner:
> Are there any objections to this?

No, that also sounds good to me.

henningandersen:
> Are there any objections to this

I'd like to keep the randomness in `InternalTestCluster` (since we always had it), but do not mind for the REST tests (though it seems a bit overkill for something we expect not to result in more failures, nor a slowdown).

test/framework/src/main/java/org/elasticsearch/snapshots/AbstractSnapshotIntegTestCase.java
```diff
@@ -470,14 +470,24 @@ protected void createIndexWithRandomDocs(String indexName, int docCount) throws
     }
 
     protected void indexRandomDocs(String index, int numdocs) throws InterruptedException {
+        indexRandomDocs(index, numdocs, true);
```
mark-vieira, line 473: Are these simply test fixes related to this change?

test/framework/src/main/java/org/elasticsearch/test/InternalTestCluster.java
```diff
@@ -524,7 +524,7 @@ private static Settings getRandomNodeSettings(long seed) {
 
         builder.put(
             EsExecutors.NODE_PROCESSORS_SETTING.getKey(),
-            1 + random.nextInt(Math.min(4, Runtime.getRuntime().availableProcessors()))
+            RandomNumbers.randomIntBetween(random, 1, Runtime.getRuntime().availableProcessors())
```
mark-vieira, line 527: Ok, so looks like we were _already_ doing this sort of thing for internal cluster tests.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/InternalTestClustersPlugin.java
```diff
@@ -36,6 +38,16 @@ public void apply(Project project) {
...
+        if (shouldConfigureTestClustersWithOneProcessor()) {
+            NamedDomainObjectContainer<ElasticsearchCluster> testClusters = (NamedDomainObjectContainer<ElasticsearchCluster>) project
+                .getExtensions()
+                .getByName(TestClustersPlugin.EXTENSION_NAME);
+            testClusters.configureEach(elasticsearchCluster -> elasticsearchCluster.setting("node.processors", "1"));
+        }
     }
 
+    private boolean shouldConfigureTestClustersWithOneProcessor() {
+        return Boolean.parseBoolean(System.getProperty("tests.configure_test_clusters_with_one_processor", "false"));
```
mark-vieira, line 51: @breskeby Can you confirm it's no longer necessary to use provider factory to get system properties to ensure things are config cache compliant?

mark-vieira (approved):
One comment that I'd like @breskeby to provide feedback on. Otherwise LGTM.

build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/InternalTestClustersPlugin.java
breskeby: @mark-vieira yes its not required anymore with newer gradle versions

breskeby (approved):

test/framework/src/main/java/org/elasticsearch/snapshots/AbstractSnapshotIntegTestCase.java
fcofdez: Yes, that's correct

fcofdez:
@elasticmachine update branch

fcofdez:
Thanks all!

Issue #89171, opened 2022-08-08
[CI] BWC issue unknown setting [index.time_series.start_time]


hendrikmuhs:
### CI Link

https://gradle-enterprise.elastic.co/s/uapcsmavbl7fg

### Repro line

./gradlew ':qa:mixed-cluster:v8.0.1#mixedClusterTest' -Dtests.class="org.elasticsearch.backwards.MixedClusterClientYamlTestSuiteIT" -Dtests.method="test {p0=field_caps/40_time_series/Get time series field caps with conflicts}" -Dtests.seed=124BAB58491BC3E1 -Dtests.bwc=true -Dtests.locale=zh-Hant-TW -Dtests.timezone=Pacific/Niue -Druntime.java=17

### Does it reproduce?

Yes

### Applicable branches

main

### Failure excerpt

```
[...]

PR #89236, opened 2022-08-10
TSDB: fix time series field caps bwc yaml test

weizijun:
This PR fix the issue https://github.com/elastic/elasticsearch/issues/89171
When add index.mode = time_series, the minimum version is 8.1.0.  So I change the skip version config.


droberts195 (approved):
LGTM

Thanks @weizijun.  I'll merge this as a way to mute the failing test, although I think there might still be more for @elastic/es-analytics-geo to consider which I'll comment on in #89171.

csoulios:
Thanks for taking care of this @droberts195

PR #89238, opened 2022-08-10
ILM: Get policy support wildcard name

weizijun:
ILM Get policy support wildcard name is a useful feature. 
We can use `Get  /_ilm/policy/policy*` to get the wildcard result.


weizijun:
@dakrone can you help to reivew this PR?

dakrone:
@weizijun supporting both `,` and wildcards leads to a lot of very difficult edge cases around response codes. I think for now it would be better to stick with *either* comma-separated specific values *or* a wildcard. (So either `GET /_ilm/policy/mypolicy1,mypolicy2` *OR* `GET /_ilm/policy/mypolicy*`.) Would you be able to make that change to this PR?

weizijun:
> @weizijun supporting both `,` and wildcards leads to a lot of very difficult edge cases around response codes. I think for now it would be better to stick with _either_ comma-separated specific values _or_ a wildcard. (So either `GET /_ilm/policy/mypolicy1,mypolicy2` _OR_ `GET /_ilm/policy/mypolicy*`.) Would you be able to make that change to this PR?

Okay, I will change as you said. @dakrone ,do you mean that `GET /_ilm/policy/mypolicy1*,mypolicy2*` is failed? Or don't use wildcard to handle the two values?

I see that the old template action, `TransportGetIndexTemplatesAction` support multi names, and each name support wildcard. The new template action supports only one name. Which one is reasonable?

dakrone:
> do you mean that `GET /_ilm/policy/mypolicy1*,mypolicy2*` is failed? Or don't use wildcard to handle the two values?

It should throw a helpful exception that says to use *either* comma-separated values *or* wildcards.

> I see that the old template action, `TransportGetIndexTemplatesAction` support multi names, and each name support wildcard. The new template action supports only one name. Which one is reasonable?

The new one is more reasonable. The old one supports it just because we cannot make the change without breaking backwards compatibility.

weizijun:
hi, @dakrone , I have finished the change, can you help to review it?

dakrone:
@elasticmachine ok to test

x-pack/plugin/ilm/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/ilm/10_basic.yml
```diff
@@ -58,6 +58,96 @@ setup:
...
+
+  - do:
+      ilm.get_lifecycle:
+        policy: "my_timeseries*"
+  - match: { my_timeseries_lifecycle_1.version: 1 }
+  - match: { my_timeseries_lifecycle_2.version: 1 }
+
+  - do:
+      ilm.get_lifecycle:
+        policy: "my_timeseries_lifecycle_1, my_timeseries_lifecycle_2"
```
dakrone, line 132: I believe this extra space is causing test failures:
```suggestion
        policy: "my_timeseries_lifecycle_1,my_timeseries_lifecycle_2"
```

dakrone (commented):
@weizijun could you add a changelog file to this PR? I left a couple of other comments also

dakrone:
@elasticmachine update branch

dakrone:
The CI failures for this are transient and due to the recent 8.5.0 version bump, so I expect them to go away as those get sorted out.

weizijun:
@elasticmachine update branch

dakrone (approved):
LGTM, thanks @weizijun!

weizijun:
> LGTM, thanks @weizijun!

Thanks @dakrone !

PR #89239, opened 2022-08-10
Improve rejection of ambiguous voting config name

DaveCTurner:
Today if there are multiple nodes with the same name then
`POST /_cluster/voting_config_exclusions?node_names=ambiguous-name` will
return a `500 Internal Server Error` and a mysterious message. This
commit changes the behaviour to throw an `IllegalArgumentException`
(i.e. `400 Bad Request`) along with a more useful message describing the
problem.

### DIFF
--- a/docs/changelog/89239.yaml
+++ b/docs/changelog/89239.yaml
@@ -1,0 +1,5 @@
+pr: 89239
+summary: Improve rejection of ambiguous voting config name
+area: Cluster Coordination
+type: bug
+issues: []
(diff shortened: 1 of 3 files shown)


DaveCTurner:
For the sake of future searches, the mysterious message we get today looks like this: `Duplicate key instance-0000000001 (attempted merging values {instance-0000000001}{...} and {instance-0000000001}{...})`.

idegtiarenko (approved):

PR #89240, opened 2022-08-10
[ML] Move method to compute current memory scale into NativeMemoryCap…

dimitris-athanasiou:
…acity

This commit moves method `currentScale` from `MlAutoscalingDeciderService`
to `NativeMemoryCapacity` as it allows for easier reuse without coupling
to the autoscaling service.


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/NativeMemoryCapacity.java
```diff
@@ -27,14 +31,6 @@ public class NativeMemoryCapacity {
 
     private static final Logger logger = LogManager.getLogger(NativeMemoryCapacity.class);
 
-    static NativeMemoryCapacity from(NativeMemoryCapacity capacity) {
```
dimitris-athanasiou, line 30: This method was not used anywhere

droberts195 (approved):
LGTM

PR #89251, opened 2022-08-10
Include runtime fields in total fields count

### LINKED ISSUE #88265: Include runtime fields in total fields count
We have a check that enforces the total number of fields needs to be below a certain (configurable) threshold. Runtime fields currently do not contribute to the count. There have been misunderstandings around this, caused by runtime fields not having a corresponding lucene field, but the reason to have the total fields limit is to limit the cluster state size as well as the memory footprint of mappings, which runtime fields certainly contribute to. For this reason we should fix this and count runtime fields as well.

mayya-sharipova:
We have a check that enforces the total number of fields needs to be below a
certain (configurable) threshold. Before runtime fields did not contribute
to the count. This patch makes all runtime fields contribute to the
count, runtime fields:
- that were explicitly defined in mapping by a user
- as well as runtime fields that were dynamically created by dynamic
 mappings

Closes #88265

### DIFF
--- a/server/src/internalClusterTest/java/org/elasticsearch/index/mapper/DynamicMappingIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/index/mapper/DynamicMappingIT.java
@@ -214,7 +214,7 @@ public class DynamicMappingIT extends ESIntegTestCase {
         }
     }
 
-    public void testTotalFieldsLimitWithRunTimefields() {
+    public void testTotalFieldsLimitWithRuntimeFields() {
         Settings indexSettings = Settings.builder()
             .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1)
             .put(INDEX_MAPPING_TOTAL_FIELDS_LIMIT_SETTING.getKey(), 4)
@@ -243,10 +243,10 @@ public class DynamicMappingIT extends ESIntegTestCase {
         ensureGreen("index1");
 
         {
-            // introduction of a new object with 2 sub-fields fails
+            // introduction of a new object with 2 new sub-fields fails
             final IndexRequestBuilder indexRequestBuilder = client().prepareIndex("index1")
                 .setId("1")
-                .setSource("my_object2", Map.of("field1", "value1", "field2", "value2"));
+                .setSource("field3", "value3", "my_object2", Map.of("new_field1", "value1", "new_field2", "value2"));
             Exception exc = expectThrows(MapperParsingException.class, () -> indexRequestBuilder.get(TimeValue.timeValueSeconds(10)));
             assertThat(exc.getMessage(), Matchers.containsString("failed to parse"));
             assertThat(exc.getCause(), instanceOf(IllegalArgumentException.class));
@@ -257,8 +257,8 @@ public class DynamicMappingIT extends ESIntegTestCase {
         }
 
         {
-            // introduction of a single field succeeds
-            client().prepareIndex("index1").setId("2").setSource("field4", 100).get();
+            // introduction of a new single field succeeds
+            client().prepareIndex("index1").setId("2").setSource("field3", "value3", "new_field4", 100).get();
         }
 
         {
@@ -272,8 +272,10 @@ public class DynamicMappingIT extends ESIntegTestCase {
                     }
                 """, XContentType.JSON));
 
-            // introduction of a new object with 2 sub-fields succeeds
-            client().prepareIndex("index1").setId("2").setSource("my_object2", Map.of("field1", "value1", "field2", "value2")).get();
+            // introduction of a new object with 2 new sub-fields succeeds
+            client().prepareIndex("index1")
+                .setId("1")
+                .setSource("field3", "value3", "my_object2", Map.of("new_field1", "value1", "new_field2", "value2"));
         }
     }
 
--- a/server/src/internalClusterTest/java/org/elasticsearch/index/mapper/DynamicMappingIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/index/mapper/DynamicMappingIT.java
@@ -18,6 +18,7 @@ import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.action.support.WriteRequest;
 import org.elasticsearch.cluster.ClusterState;
 import org.elasticsearch.cluster.ClusterStateUpdateTask;
+import org.elasticsearch.cluster.metadata.IndexMetadata;
 import org.elasticsearch.cluster.metadata.MappingMetadata;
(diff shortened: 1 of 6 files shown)


server/src/internalClusterTest/java/org/elasticsearch/index/mapper/DynamicMappingIT.java
```diff
@@ -213,6 +214,69 @@ public void onFailure(Exception e) {
...
+            """;
+
+        client().admin().indices().prepareCreate("index1").setSettings(indexSettings).setMapping(mapping).get();
+        ensureGreen("index1");
+
+        {
+            // introduction of a new object with 2 sub-fields fails
+            final IndexRequestBuilder indexRequestBuilder = client().prepareIndex("index1")
+                .setId("1")
+                .setSource("my_object2", Map.of("field1", "value1", "field2", "value2"));
```
javanna, line 249: the doc here holds only new fields? Should we add some already known fields and some new fields?

server/src/internalClusterTest/java/org/elasticsearch/index/mapper/DynamicMappingIT.java
```diff
@@ -213,6 +214,69 @@ public void onFailure(Exception e) {
         }
     }
 
+    public void testTotalFieldsLimitWithRunTimefields() {
```
javanna, line 217: s/RunTime/Runtime ?

With this test you are verifying the behaviour for dynamic mapping updates, I think you also need a test for the scenario where runtime fields are mapped through put mappings API. There are already similar tests in MapperServiceTests.

javanna (changes requested):
I left a couple of comments. I also wonder if we should make the error message clearer and return the count of new fields, and even say how many objects, how many mappers and how many runtime fields are being added. That can be a followup though

server/src/internalClusterTest/java/org/elasticsearch/index/mapper/DynamicMappingIT.java
mayya-sharipova: Great suggestion, addressed in 1ee42705cb6ad4c4c5b6083a33c042a1b8269912

server/src/internalClusterTest/java/org/elasticsearch/index/mapper/DynamicMappingIT.java
mayya-sharipova: Great suggestion, addressed in 1ee42705cb6ad4c4c5b6083a33c042a1b8269912

server/src/main/java/org/elasticsearch/index/mapper/DocumentParserContext.java
```diff
@@ -254,6 +247,17 @@ public final void addDynamicMapper(Mapper mapper) {
         dynamicMappers.add(mapper);
     }
 
+    private void checkFieldLimit(String fieldName) {
+        // eagerly check field name limit here to avoid OOM errors
+        // only check fields that are not already mapped or tracked in order to avoid hitting field limit too early via double-counting
+        // note that existing fields can also receive dynamic mapping updates (e.g. constant_keyword to fix the value)
+        if (mappingLookup.getMapper(fieldName) == null
+            && mappingLookup.objectMappers().containsKey(fieldName) == false
+            && newFieldsSeen.add(fieldName)) {
```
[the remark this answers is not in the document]
mayya-sharipova: Right, based on the comments, since existing mappers can receive mapping updates, we needed this check.

> When we are adding a dynamic runtime field these these two checks are redundant, because they will always return false? Do we want to have two different codepaths then for the two scenarios?

Thanks for explaining this, I did not know this. I've made the tow different codepaths in 1ee42705cb6ad4c4c5b6083a33c042a1b8269912

mayya-sharipova:
@javanna Thanks for the initial review. I've tried to address in 1ee42705cb6ad4c4c5b6083a33c042a1b8269912

> I also wonder if we should make the error message clearer and return the count of new fields, and even say how many objects, how many mappers and how many runtime fields are being added. That can be a followup though

I am happy to add extra info in the error message as a part of this PR.  I wonder if this constitutes a breaking change, or we are fine to change to expand error messages?

server/src/main/java/org/elasticsearch/index/mapper/DocumentParserContext.java
```diff
@@ -279,6 +279,9 @@ final ObjectMapper getDynamicObjectMapper(String name) {
      * Add a new runtime field dynamically created while parsing.
      */
     public final void addDynamicRuntimeField(RuntimeField runtimeField) {
+        if (newFieldsSeen.add(runtimeField.name())) {
```
javanna, line 282: A runtime field can have the same name as an indexed field, in which case it will shadow the indexed field at search time. I wondered if it's ok to have the same set for both runtime fields and indexed fields and it does look ok because when it comes to dynamic mappings, when we encounter a new field, either we map it as a runtime field or as an indexed field, and never both, so there are never going to be collisions. Shall we add a comment to clarify that?

server/src/main/java/org/elasticsearch/index/mapper/DocumentParserContext.java
```diff
@@ -279,6 +279,9 @@ final ObjectMapper getDynamicObjectMapper(String name) {
      * Add a new runtime field dynamically created while parsing.
      */
     public final void addDynamicRuntimeField(RuntimeField runtimeField) {
```
javanna, line 281: I looked into enforcing that this is called only for new runtime fields. That is true right now but we don't enforce it anywhere. Enforcing it though looks overkill, as it adds potential overhead and I am not even sure how worthwhile it is. Should we start by making this method package protected?

javanna (approved):
left two nits, looks great though, thanks a lot for working on this! Let's iterate on the error message as a follow-up.

server/src/main/java/org/elasticsearch/index/mapper/DocumentParserContext.java
mayya-sharipova: @javanna Good suggestion,  addressed in dbf20223bce7a5340f803eb59188a9d7d5aa5418

Indeed for dynamic updates, a field can be mapped either as runtime field or as indexed field, so no problem here.

For PUT _mapping updates, a user can introduce a new  runtime field with the same name as existing indexed field. In this case, this new runtime field will be added to the total count, and can cause an exception for total field counts.

server/src/main/java/org/elasticsearch/index/mapper/DocumentParserContext.java
mayya-sharipova: addressed in dbf2022

PR #89252, opened 2022-08-10
Mark shard failures caused by unsupported aggregations or queries against rolled up data so Kibana can identify them

salvatore-campagna:
When an unsupported aggregation is executed on a
`aggregate_double_metric` field an error with a specific
`type` field is generated. This allows clients like Kibana
to identify these errors and handle them properly.

Also we would like to fail on queries using a date histogram
aggregation with `calendar_interval` on rollup indices.
Date histogram aggregations are executed on rollup indices
only if using `fixed_interval` and a `UTC` timezone.


server/src/main/java/org/elasticsearch/ElasticsearchException.java
```diff
@@ -1578,6 +1578,12 @@ private enum ElasticsearchExceptionHandle {
             HealthNodeNotDiscoveredException::new,
             166,
             Version.V_8_5_0
+        ),
+        UNSUPPORTED_AGGREGATION_ON_DOWNSAMPLED_FIELD_EXCEPTION(
+            org.elasticsearch.search.aggregations.UnsupportedAggregationOnDownsampledField.class,
+            org.elasticsearch.search.aggregations.UnsupportedAggregationOnDownsampledField::new,
+            167,
+            Version.V_8_4_0
```
salvatore-campagna, line 1586: Will change this to 8.5.0

x-pack/plugin/rollup/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/rollup/20_unsupported_aggs.yml
```diff
@@ -0,0 +1,164 @@
+setup:
+  - skip:
+      version: " - 8.3.99"
```
salvatore-campagna, line 3: Will change this to 8.4.99 and comment with 8.5.0

server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java
```diff
@@ -417,9 +420,28 @@ protected ValuesSourceAggregatorFactory innerBuild(
...
+                    + this.getName()
+                    + "] with interval type ["
+                    + dateHistogramIntervalType.getPreferredName()
+                    + "]"
+            );
+        }
 
         final ZoneId tz = timeZone();
+        if (IndexMode.TIME_SERIES.getName().equals(indexMode) && ZoneId.of("UTC").equals(tz) == false) {
+            throw new UnsupportedAggregationOnDownsampledField(
```
salvatore-campagna, line 439: Should we use a different exception here?

salvatore-campagna:

server/src/main/java/org/elasticsearch/search/aggregations/support/ValuesSourceRegistry.java
```diff
@@ -169,6 +171,13 @@ public <T> T getAggregator(RegistryKey<T> registryKey, ValuesSourceConfig values
             @SuppressWarnings("unchecked")
             T supplier = (T) aggregatorRegistry.get(registryKey).get(valuesSourceConfig.valueSourceType());
             if (supplier == null) {
+                if (valuesSourceConfig.fieldType() != null
+                    && valuesSourceConfig.fieldType().typeName() != null
+                    && "aggregate_metric_double".equals(valuesSourceConfig.fieldType().typeName().toLowerCase(Locale.ROOT))) {
+                    throw new UnsupportedAggregationOnDownsampledField(
+                        valuesSourceConfig.getDescription() + " is not supported for aggregation [" + registryKey.getName() + "]"
+                    );
+                }
```
nik9000, line 180: I wonder if there's a better place to put this. It feels like kind of a janky spot. @not-napoleon, do you have an opinion on this? Basically we want field types to be able to throw a specific exception if you ask or an unsupported aggregation. Well, really, we just want *this* field to throw it, but I'll bet there is a cleaner way.

salvatore-campagna: Maybe I can do this at the moment the aggregator is created.

salvatore-campagna: This needs to be done for all aggregations. So the only place I see to do it (excluding doing it in every aggregation) is some base class where the construction of the aggregator takes place. Any better idea?

server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java
```diff
@@ -417,9 +420,28 @@ protected ValuesSourceAggregatorFactory innerBuild(
         AggregatorFactory parent,
         AggregatorFactories.Builder subFactoriesBuilder
     ) throws IOException {
-        DateHistogramAggregationSupplier aggregatorSupplier = context.getValuesSourceRegistry().getAggregator(REGISTRY_KEY, config);
+        final String indexMode = context.getIndexSettings().getSettings().get(IndexSettings.MODE.getKey());
```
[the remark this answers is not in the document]
salvatore-campagna: I don't see the method `validateCalendarIntervalType`. Is it something you introduced recently?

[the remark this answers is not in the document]
salvatore-campagna: Oh...maybe you are suggesting to add it...

server/src/main/java/org/elasticsearch/search/aggregations/support/ValuesSourceRegistry.java
salvatore-campagna: Notice, anyway, that if I remove this piece of code, it fails anyway with 

`throw new IllegalArgumentException(
                    valuesSourceConfig.getDescription() + " is not supported for aggregation [" + registryKey.getName() + "]"
);`

so at the end that is just a way to replace one exception with another.

An option I see is to extract the responsibility of checking for "no value in the registry" with another class that is injected into the ValuesSourceRegistry. A kind of "no such element in the registry" handler.

not-napoleon: I think we already have the handler injected - the `ValuesSourceType` itself can do this.  What you're asking for here is a type-specific behavior.  In aggregations, we don't (generally) work directly with field types, we instead work with the `ValuesSourceType` abstraction, so unless we have a Real Good Reason :TM: that should be the lowest granularity of type switching (real good reasons generally involve "we can optimize this specific index structure, but only if we know about it", for reference).

So, I'd suggest we add a method to the `ValuesSourceType` interface `getUnmappedException(String aggregationName)` or something like that (the VST already has a `typeName()` method it can use for the other string component).  We can add a default implementation that returns the current exception, and override it in `AggergateMetricValuesSourceType.AGGREGATE_METRIC` to throw your new exception.  This also has the advantage of locating the special messaging close to the associated type logic, which seems more natural than hard coding it in the registry.

Let me know if that doesn't work for some reason.

salvatore-campagna: That sound better. Thanks.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java
```diff
@@ -417,9 +419,27 @@ protected ValuesSourceAggregatorFactory innerBuild(
         AggregatorFactory parent,
         AggregatorFactories.Builder subFactoriesBuilder
     ) throws IOException {
-        DateHistogramAggregationSupplier aggregatorSupplier = context.getValuesSourceRegistry().getAggregator(REGISTRY_KEY, config);
+        IndexMode indexMode = context.getIndexSettings().getMode();
+        final DateIntervalWrapper.IntervalTypeEnum dateHistogramIntervalType = dateHistogramInterval.getIntervalType();
+        if (context.getIndexSettings().getMode().validateCalendarIntervalType(dateHistogramIntervalType)) {
```
nik9000, line 424: Can this method be `void` and throw for invalid values? Could you do the same for the time zone below?

server/src/main/java/org/elasticsearch/search/aggregations/support/ValuesSourceRegistry.java
salvatore-campagna: I had to add the `ValueSourceConfig` to the parameters because otherwise composing the exact error message was not possible. The `getDescription` has some logic to build it.

nik9000: I wouldn't bother to check that the exception is non-null here. Returning a `null` supplier isn't really a thing we want to support. I think it's an error is the type doesn't return an exception.

server/src/main/java/org/elasticsearch/search/aggregations/support/ValuesSourceType.java
```diff
@@ -92,4 +92,8 @@ default DocValueFormat getFormatter(String format, ZoneId tz) {
      * @return the name of the Values Source Type
      */
     String typeName();
+
+    default RuntimeException getUnmappedException(ValuesSourceConfig config, String name) {
```
salvatore-campagna, line 96: Will add the Javadoc.

server/src/main/java/org/elasticsearch/search/aggregations/support/ValuesSourceRegistry.java
not-napoleon: this needs an else-throw.  If we didn't find a supplier in the registry, we *MUST* throw from this method, or something very weird will break down the road.  Really, it should be impossible for `getUnmappedException` to return `null`, but I'm all for redundant checks.

server/src/main/java/org/elasticsearch/search/aggregations/support/ValuesSourceType.java
not-napoleon, line 104: I don't love passing a config in here.  I get why, but I'd almost rather we passed in a string and let the caller resolve this.  This is just weirdly recursive, since the config is (generally) the thing that holds the VST.

Maybe this takes two strings, and we add a method to ValuesSourceConfig that calls it with the description? IDK.

Also, I know I suggested the name `getUnmappedException` earlier, but seeing it out of context on `ValuesSourceType` I realized that we use "unmapped" to mean something else generally, and this might be confusing.  How do you feel about `getUnregisteredException` instead?  And either way, this method needs some javadoc.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java
salvatore-campagna: Yes I was doing that...returning true/false is confusing...is it returning false if validation is ok or not?

server/src/main/java/org/elasticsearch/search/aggregations/support/ValuesSourceRegistry.java
salvatore-campagna: I see...I thought about that but my idea was "there is the default method". Anyway that is not enough. A subclass overriding the parent method might still return null and then the flow would continue without an exception thrown.

salvatore-campagna: Does it make sense to add a warning message saying something like "returned unexpected null exception, please fix this!"?

server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java
nik9000: I'm not sure! I'd go with the same exception for now and see if the kibana folk want to change it.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java
nik9000: I'd just validate and throw.

server/src/main/java/org/elasticsearch/search/aggregations/support/ValuesSourceRegistry.java
nik9000: You'd asked about double checking that this isn't null rather than just throwing blindly. I'm ok either way. It's a "you built this wrong" sort of error. I don't expect folks to hit it much, but it'd be kind to tell them. I'd use an `assert` for it or not build it at all. But you can do whatever you like here. no big deal to me.

not-napoleon: yeah, thinking about it more, the assert is probably the right choice.  We want to catch this in testing, not give the user an error they can't do anything with, and the assert communicates that more clearly.

salvatore-campagna:
After discussing with @flash1293 it looks like that the ability to distinguish two types of exceptions would be nice to have. That allows Kibana to have a better handling of errors, telling the user to fix their config if an unsupported aggregation is used on a field, versus telling the user to open a bug because one of the supported aggregations is using an incorrect parameter. This last case applies, at the moment, just to the usage of `calendar_interval` and `non utc timezone` when a date histogram aggregation is executed. For the second type of error the idea is to use an `IllegalArgumentException`.

nik9000 (approved):
LGTM. @not-napoleon, how about you?

server/src/main/java/org/elasticsearch/index/IndexMode.java
```diff
@@ -194,6 +202,25 @@ public DocumentDimensions buildDocumentDimensions() {
         public boolean shouldValidateTimestamp() {
             return true;
         }
+
+        /**
+         * For a time series index we just support 'fixed_interval' aggregations.
```
nik9000, line 207: I think I might have missed something. I thought we only wanted to limit to fixed interval for the rolled up indices. Time series indices are supposed to support all the things. I think. I certainly could be forgetting things, but that was my memory.

flash1293: good catch, this was also my understanding

server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java
```diff
@@ -417,9 +417,28 @@ protected ValuesSourceAggregatorFactory innerBuild(
         AggregatorFactory parent,
         AggregatorFactories.Builder subFactoriesBuilder
     ) throws IOException {
-        DateHistogramAggregationSupplier aggregatorSupplier = context.getValuesSourceRegistry().getAggregator(REGISTRY_KEY, config);
+        final DateIntervalWrapper.IntervalTypeEnum dateHistogramIntervalType = dateHistogramInterval.getIntervalType();
+        context.getIndexSettings()
+            .getMode()
+            .validateCalendarIntervalType(
+                dateHistogramIntervalType,
+                config.getDescription()
```
not-napoleon, line 425: I don't love this pattern of passing the message in and throwing from the validator.  Why not just have the validator return a boolean and throw from here if we want to?  I don't feel terribly strongly about it though.

server/src/test/java/org/elasticsearch/search/aggregations/support/ValuesSourceRegistryTests.java
```diff
@@ -56,4 +58,61 @@ public void testAggregatorNotFoundException() {
         expectThrows(IllegalArgumentException.class, () -> registry.getAggregator(key, fieldOnly));
         expectThrows(IllegalArgumentException.class, () -> registry.getAggregator(key, scriptOnly));
     }
+
+    public void testValueSourceTypeNullUnregisteredExceptionHandler() {
```
not-napoleon, line 62: I feel weird saying this, but I think this test is unnecessary to the point of being distracting.  You are literally just testing that the assert works, which isn't going to run in production anyway.  So this test doesn't describe the production behavior, which is a little troubling to me.  I think the assert by itself is enough.

not-napoleon (approved):
Couple of nits, but I think the interaction with the `ValuesSourceRegistry` is okay.

nik9000 (changes requested):
Hey! officially "requesting changes" while we figure out what's up with that last validation.

server/src/test/java/org/elasticsearch/search/aggregations/support/ValuesSourceRegistryTests.java
salvatore-campagna: I agree it is pretty much useless by itself, but I wanted to have a test just to stress the point that a non-null exception is required. The other option was to have some Javadoc...but if I have to choose between documentation and a test I usually prefer a test for two reasons:
1. a test fails if something is wrong
2. a test is also a piece of documentation
3. it is very easy to forget updating documentation while you will have a test failing and you have to fix it if something changes

salvatore-campagna: Anyway I can get rid of it.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java
nik9000: :rofl: I advocated for having the method do the throwing. I think it's *generally* a cleaner way to do it, especially because the method has to return a very "related" type of exception. So! Maybe it won't matter. This is the thing we're talking about maybe not applying correctly anyway - I think we only want to throw if we're operating on a rolled up index. Not a timeseries one. And *that* will work differently anyway.

salvatore-campagna: I think I did something like this before and I changed it. I find it weird to return a boolean from a validation method.
Should the method return true or false if validation succeeds?

server/src/main/java/org/elasticsearch/index/IndexMode.java
salvatore-campagna: What about the UTC/non-UTC check?

flash1293: Same for UTC/non-UTC by my understanding. Not sure whether it's right or not, but in Kibana we currently check this via the `meta.fixed_interval` property on the field: https://github.com/elastic/kibana/blob/9799dbba27c5baf594357eae0bbfc79b4e7da77c/src/plugins/data_views/server/fetcher/lib/field_capabilities/field_caps_response.ts#L145

server/src/main/java/org/elasticsearch/search/aggregations/bucket/histogram/DateHistogramAggregationBuilder.java
not-napoleon: Maybe I'm splitting hairs, but if the validation method is going to throw, it should own building the message.  That's all I'm saying.  Having the message in one place and the decision to throw in a different place is weird to me.

server/src/test/java/org/elasticsearch/search/aggregations/support/ValuesSourceRegistryTests.java
not-napoleon: Point 2 is exactly my concern - to me, this test documents that if you return a null exception, we still throw, which is what we want.  That seems to say "it's okay to have a null here".

Since our tests run with asserts enabled, a test will fail if we return a null from `getUnregisteredException()`; that's the whole point of the assert being there, and that's the behavior we want to enforce.

server/src/main/java/org/elasticsearch/index/IndexMode.java
csoulios: > I don't get what is wrong here...do you mean this should fail only if the date histogram aggregation runs on a rolled-up index? And not on every time series index? If that is what I missed then I have no idea how do we know if an index is a result of a rollup operation or not. Maybe we have some meta-data? I will check.

It should fail only if the `date_histogram` agg is executed on a rolled up index. "Raw" time-series indices support any `date_histogram` interval. 
Rollup indices contain the `IndexMetadata.INDEX_ROLLUP_SOURCE_NAME` setting in their index settings. And the `@timestamp` field should contain the `fixed_interval` property in the field mapping `meta`. See https://github.com/elastic/elasticsearch/blob/03f3c8119ac1e7bb7034be1f7c4f3355fec4b300/x-pack/plugin/rollup/src/main/java/org/elasticsearch/xpack/rollup/v2/TransportRollupAction.java#L440-L453

server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetadata.java
```diff
@@ -127,6 +128,15 @@ public class IndexMetadata implements Diffable<IndexMetadata>, ToXContentFragmen
         EnumSet.of(ClusterBlockLevel.WRITE)
     );
 
+    public boolean isRollupIndex() {
```
salvatore-campagna, line 131: I have the feeling this is not the best I can do here. I am working on a better way to do this at the moment but I wanted to push this change because it might be good enough. If that feeling is shared but this is considered "good enough" we could proceed merging this to unblock Kibana folks experimenting with catching exceptions. Then I can create another issue to refactor this. N.B. I am on vacation next week and Christos is on vacation too and I would like to avoid Kibana folks being stuck waiting for this.

x-pack/plugin/rollup/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/rollup/20_unsupported_aggs.yml
```diff
@@ -0,0 +1,164 @@
...
+              percentiles:
+                field: total_memory_used
+                percents: [90, 95, 99]
+
+  - match: { status: 400 }
+  - match: { error.root_cause.0.type: unsupported_aggregation_on_downsampled_field }
+  - match: { error.root_cause.0.reason: "Field [total_memory_used] of type [aggregate_metric_double] is not supported for aggregation [percentiles]" }
+
+---
+"Top-metrics aggregation on aggregate_metric_double field":
```
flash1293, line 141: Not sure whether out of scope but what about a test for the mixed case (as this will be extremely common in practice)? Querying both indices at the same time, resulting in one of them succeeding and the other producing a shard failure so there's partial data in the response

flash1293 (commented):
Looks mostly good to me, left a nit about another test we could add

x-pack/plugin/rollup/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/rollup/20_unsupported_aggs.yml
salvatore-campagna: Oh that is a good test I think...I am not sure what the result is in this case. I guess we hava a shard failure for the rollup index and partial results coming from the time series index...?

flash1293: Exactly

x-pack/plugin/rollup/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/rollup/30_date_histogram.yml
```diff
@@ -219,3 +219,30 @@ setup:
...
+        body:
+          size: 0
+          aggs:
+            date_histogram:
+              date_histogram:
+                field: "@timestamp"
+                calendar_interval: hour
+                min_doc_count: 1
+
+  - match: { _shards.total: 2 }
```
salvatore-campagna, line 236: NOTE: here we rely on the fact that the rollup index is created with the same number of shards of the original index, which is set to 1 in the setup block.

x-pack/plugin/rollup/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/rollup/20_unsupported_aggs.yml
salvatore-campagna: This test is not necessary for the errors we throw if the aggregation is not supported (aggregation on field of type `agrgegate_metric_double`). In that case we don't check if the index is a time series index or not...just the field type.

x-pack/plugin/rollup/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/rollup/30_date_histogram.yml
```diff
@@ -0,0 +1,250 @@
+setup:
```
salvatore-campagna, line 1: I had to change this test adding the logic to create the rollup index from the source index because some settings and metadata are applied by the rollup operation and it is not possible to set them from the yaml test. The two settings I use are the source index name and the rollup status which I check to verify if the index is a rollup index or not.

server/src/test/java/org/elasticsearch/search/aggregations/support/ValuesSourceRegistryTests.java
salvatore-campagna: Yes right...but if the assert is removed as a result of a code change than something unpredictable can happen. While,  if we have a test checking for the assertion error we would, at least have, a test failing because of the missing assertion. Am I wrong?

salvatore-campagna: Anyway if the assert is removed for whatever reason and `unmappedException` is null then we have:

```
Unexpected exception type, expected AssertionError but got java.lang.NullPointerException: Cannot throw exception because "unmappedException" is null
```

salvatore-campagna: Of course when a production build is created the assertion goes away but at least the CI would have the test checking that the assertion is still there.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetadata.java
nik9000: I think it's cool as a temporary thing. Maybe a `TODO` in the code so we know we thought it was temporary when we added it.

server/src/main/java/org/elasticsearch/index/IndexMode.java
```diff
@@ -196,8 +216,55 @@ public DocumentDimensions buildDocumentDimensions(IndexSettings settings) {
...
+         * @param intervalType the type of interval used by the date histogram
+         */
+        @Override
+        public void validateCalendarIntervalType(
+            IndexSettings indexSettings,
+            DateIntervalWrapper.IntervalTypeEnum intervalType,
+            String valuesSourceDescription,
+            String aggregationName
+        ) {
+            if (indexSettings.getIndexMetadata().isRollupIndex() && DateIntervalWrapper.IntervalTypeEnum.CALENDAR.equals(intervalType)) {
```
nik9000, line 232: I *think* we want this test for all time rolled up indices. Not that there are any non-time series ones, but I think the addition of `isRollupIndex` means we don't need these new methods in `IndexMode` any more.

server/src/main/java/org/elasticsearch/index/IndexMode.java
```diff
@@ -196,8 +216,55 @@ public DocumentDimensions buildDocumentDimensions(IndexSettings settings) {
...
+        }
     };
 
+    private static boolean isIndexRollup(final IndexSettings indexSettings) {
+        final String sourceIndex = indexSettings.getIndexMetadata().getSettings().get(IndexMetadata.INDEX_ROLLUP_SOURCE_NAME_KEY);
+        final String indexRollupStatus = indexSettings.getIndexMetadata().getSettings().get(IndexMetadata.INDEX_ROLLUP_STATUS_KEY);
+        final boolean rollupSuccess = IndexMetadata.RollupTaskStatus.SUCCESS.name()
+            .toLowerCase(Locale.ROOT)
+            .equals(indexRollupStatus != null ? indexRollupStatus.toLowerCase(Locale.ROOT) : IndexMetadata.RollupTaskStatus.UNKNOWN);
+        return Strings.isNullOrEmpty(sourceIndex) == false && rollupSuccess;
```
nik9000, line 265: I think this is a duplicate.

server/src/main/java/org/elasticsearch/search/aggregations/UnsupportedAggregationOnDownsampledField.java
```diff
@@ -0,0 +1,35 @@
...
+import org.elasticsearch.rest.RestStatus;
+
+import java.io.IOException;
+
+/**
+ * Thrown when executing an aggregation on a time series index field whose type is not supported.
+ * Downsampling uses specific types while aggregating some fields (like 'aggregate_metric_double').
+ * Such field types do not support some aggregations.
+ */
+public class UnsupportedAggregationOnDownsampledField extends AggregationExecutionException {
```
nik9000, line 21: It's probably worth leaving a note that the *name* of this class is part of a contract with Kibana so we shouldn't change it.

nik9000 (changes requested):

x-pack/plugin/rollup/qa/rest/src/yamlRestTest/resources/rest-api-spec/test/rollup/20_unsupported_aggs.yml
```diff
@@ -0,0 +1,164 @@
...
+                    from: 0.0
+                    to: 200.0
+                  -
+                    from: 201.0
+                    to: 400.0
+                  -
+                    from: 401.0
+
+  - match: { status: 400 }
+  - match: { error.root_cause.0.type: unsupported_aggregation_on_downsampled_field }
```
nik9000, line 94: It's probably worth leaving a comment here that this `type` name is part of a contract with kibana. Just extra paranoia so folks don't change it.

nik9000 (approved):

csoulios (commented):
I had a very quick look at the code. I only left a small comment about naming the exception, because this is something we cannot change in the future.

nik9000 (approved):

nik9000:
I've merged this so @flash1293 should be able to grab it sooner rather than later.

PR #89253, opened 2022-08-10
[ML] Extract timing of autoscaling into its own class

dimitris-athanasiou:
Extracts the time keeping necessary for the ML autoscaling
decider into its own class.


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/ScaleTimer.java
```diff
@@ -0,0 +1,71 @@
...
+        this.timeSupplier = Objects.requireNonNull(timeSupplier);
+        this.scaleDownDetected = NO_SCALE_DOWN_POSSIBLE;
+    }
+
+    void markScale() {
+        previousScaleTimeMs = lastScaleTimeMs;
+        lastScaleTimeMs = timeSupplier.getAsLong();
+    }
+
+    Optional<Long> lastScaleToScaleIntervalMillis() {
```
droberts195, line 43: Could use `OptionalLong` here

droberts195 (approved):
LGTM

PR #78727, opened 2021-10-05
Improve Node Shutdown Observability

AthenaEryma:
This PR does two things:
1) Logs a message when the Get Shutdown Status API detects there's a shard causing a node shutdown to stall. The log message isn't new, just bumped from `trace` to `info`.
2) Adds Allocation Explain output to the shard migration status portion of the node shutdown response if it's causing a shard to stall.

Example of the new Get Shutdown Status response:
```json
{
    "nodes": [
        {
            "node_id": "GszZYZ2bSZiLi5dBinWf8g",
            "type": "REMOVE",
            "reason": "idk my bff jill",
            "shutdown_startedmillis": 1633475221483,
            "status": "STALLED",
            "shard_migration": {
                "status": "STALLED",
                "shard_migrations_remaining": 1,
                "explanation": "shard [0] [primary] of index [my-index] cannot move",
                "node_allocation_decisions": {
                    "can_remain_on_current_node": "no",
                    "can_remain_decisions": [
                        {
                            "decider": "node_shutdown",
                            "decision": "NO",
                            "explanation": "node [GszZYZ2bSZiLi5dBinWf8g] is preparing to be removed from the cluster"
                        },
                        {
                            "decider": "filter",
                            "decision": "NO",
                            "explanation": "node does not match index setting [index.routing.allocation.require] filters [include._id:\"GszZYZ2bSZiLi5dBinWf8g\"]"
                        }
                    ],
                    "can_move_to_other_node": "no",
                    "move_explanation": "cannot move shard to another node, even though it is not allowed to remain on its current node",
                    "node_allocation_decisions": [
                        {
                            "node_id": "T6AnBS1mT2iXlD-McXSteg",
                            "node_name": "node-0",
                            "transport_address": "127.0.0.1:9301",
                            "node_attributes": {
                                "ml.machine_memory": "34359738368",
                                "ml.max_open_jobs": "512",
                                "xpack.installed": "true",
                                "ml.max_jvm_size": "17179869184"
                            },
                            "node_decision": "no",
                            "weight_ranking": 1,
                            "deciders": [
                                {
                                    "decider": "filter",
                                    "decision": "NO",
                                    "explanation": "node does not match index setting [index.routing.allocation.require] filters [include._id:\"GszZYZ2bSZiLi5dBinWf8g\"]"
                                }
                            ]
                        },
                        {
                            "node_id": "81mJIQ_XRrG2zK1ojhjLMg",
                            "node_name": "node-2",
[...]

PR #89254, opened 2022-08-10
Fix message for stalled shutdown

DaveCTurner:
Today if a node shutdown is stalled due to unmoveable shards then we say
to use the allocation explain API to find details. In fact, since #78727
we include the allocation explanation in the response already so we
should tell users just to look at that instead. This commit adjusts the
message to address this.


grcevski (approved):
LGTM!

PR #89256, opened 2022-08-11
Add sortable param to allow named filters agg return buckets as an array of objects

### LINKED ISSUE #83957: Add ability to override keyed parameter in filters aggregation
See https://github.com/elastic/elasticsearch/issues/83957#issuecomment-1061038677 for the reproduction and the description of the issue. 

Original bug report: 
----

### Elasticsearch Version

6.4.3

### Installed Plugins

/

### Java Version

_bundled_

### OS Version

linux

### Problem Description

using the bucket_sort in a filters aggregation results in no sorting at all

### Steps to Reproduce

Example query:

````
{
  "aggs": {
    "Category": {
      "filters": {
        "other_bucket": false,
        "other_bucket_key": "No Category",
        "filters": {
          "165422": {
            "term": {
              "product.categories.id": 165422
            }
          },
          "171229": {
            "term": {
              "product.categories.id": 171229
            }
          },
          "165523": {
            "term": {
              "product.categories.id": 165523
            }
          },
          "165530": {
            "term": {
              "product.categories.id": 165530
            }
          }
        }
      },
      "aggs": {
        "cancelation_count": {
          "sum": {
            "script": {
              "inline": "if(doc['order_cancellation_id'].value > 0){return 1} return 0;",
              "lang": "painless"
            }
          }
        },
        "sort_cancelation_count": {
          "bucket_sort": {
            "sort": [ "cancelation_count" ],
            "size": 4
          }
        }
      }
    }
  },
  "size": 0
}
````

### Logs (if relevant)

Example result (filters aggregations are not sorted via the sum:cancelation_count but instead via the bucket key):
```
{
  "took": 620,
  "timed_out": false,
  "_shards": {
    "total": 5,
    "successful": 5,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": 9580162,
    "max_score": 0,
    "hits": []
  },
  "aggregations": {
    "Category": {
      "buckets": {
        "165422": {
          "doc_count": 5835212,
          "cancelation_count": {
            "value": 145488
          }
        },
        "165523": {
          "doc_count": 1974874,
          "cancelation_count": {
            "value": 43586
          }
        },
        "165530": {
          "doc_count": 1231949,
          "cancelation_count": {
            "value": 33821
          }
        },
        "171229": {
          "doc_count": 193478,
          "cancelation_count": {
            "value": 5939
          }
        }
      }
    }
  }
}
```

### PR DESCRIPTION
**Pull request by amooncake**
Named filters with bucket sort currently return object, which may not be
reliably supported by JSON since it doesn't guarantee the order of
elements in the object. In this case, add a sortable parameter to the
filters agg, which will be ignored by the anonymous filters, the buckets
will be returned in an array of objects when it is true.

Closes #83957

### CODE REVIEW

File: server/src/main/java/org/elasticsearch/search/aggregations/pipeline/BucketSortPipelineAggregator.java [hunk 1]
```diff
@@ -78,7 +79,13 @@ public InternalAggregation reduce(InternalAggregation aggregation, AggregationRe
...
             newBuckets.add(ordered.get(i).internalBucket);
         }
-        return originalAgg.create(newBuckets);
+
+        InternalAggregation internalAggregation = originalAgg.create(newBuckets);
+        if (internalAggregation instanceof InternalFilters internalFilters) {
+            return new InternalFilters(internalFilters.getName(), internalFilters.getBuckets(), false, internalFilters.getMetadata());
+        }
+
+        return internalAggregation;
```
nik9000, line 88: I'm not super comfortable having one aggregation cause another aggregation to change it's rendering like this. I get that it's a much more sensible way to do it, but it feels weird. I'd prefer to have this be something you configure on the agg up front. Over [here](https://github.com/elastic/elasticsearch/issues/83957#issuecomment-1083317711) we talked about letting folks configure the `keyed` parameter - but I'm not really sure that makes sense....

Both this solution and overriding the `keyed` parameter kind of have the opposite problem - if you don't name the filters and then you sort then you can't figure out which one is which.

It all sort of makes me wonder - why point the bucket_sort pipeline at the filters agg at all? It sort of breaks it.

Finally, I'm a bit worried about changing this json output without some kind of extra parameter because it's *technically* breaking. I mean, the way we're outputting these now is kind of garbage in combination with the `bucket_sort` pipeline agg. But so many folks use Elasticsearch right now, any change in json format is likely to break someone.

So I wonder if the "right" way to fix this is to add a new parameter to the `filters` agg that causes it to output the results in this way. Maybe a `ordered_buckets: true` or  something. I don't know. I'm not good at names. And, yeah, it's lame to need a new parameter. But it's paranoid. And in this case I think that paranoia is good.

amooncake: I agreed, and then I added a keyed_bucket_in_array param as you suggested to avoid judging this in bucket sort.

Now the user can directly modify the return type regardless of whether there is a bucket sort.

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.aggregation/220_filters_bucket.yml
```diff
@@ -137,6 +137,40 @@ setup:
...
+"named filters with bucket sort test":
+  - do:
+      search:
+        rest_total_hits_as_int: true
+        body:
+          size: 0
+          aggs:
+            the_filter:
+              filters:
+                keyed_bucket_in_array: true
```
nik9000, line 150: I don't particularly like the name of this parameter. But I don't really know of a better name. @not-napoleon, do you have an opinion here?

server/src/main/java/org/elasticsearch/search/aggregations/bucket/filter/FiltersAggregationBuilder.java
```diff
@@ -123,6 +126,7 @@ public FiltersAggregationBuilder(StreamInput in) throws IOException {
         }
         otherBucket = in.readBoolean();
         otherBucketKey = in.readString();
+        keyedBucketInArray = in.readBoolean();
```
nik9000, line 129: This'll probably cause something like:
```
./gradlew -p qa/mixed-cluster/ -Dtests.method='*filter*'
```

to fail. The fix it to change it to something like:
```
keyedBucketInArray = in.getVersion().before(Version.V_8_5_0) ? false : in.readBoolean();
```

And to skip the write on the other side if the version is too old.

server/src/main/java/org/elasticsearch/search/aggregations/bucket/filter/InternalFilters.java
```diff
@@ -148,6 +157,7 @@ public InternalFilters(StreamInput in) throws IOException {
     @Override
     protected void doWriteTo(StreamOutput out) throws IOException {
         out.writeBoolean(keyed);
+        out.writeBoolean(keyedBucketInArray);
```
nik9000, line 160: This needs the same guard as I mentioned above, sadly.

nik9000:
@elasticsearchmachine, ok to test.

amooncake:

nik9000:

nik9000:
> @elasticsearchmachine update branch

I think it needs employees.

nik9000:
@elasticsearchmachine, update branch please

amooncake:
What a coincidence, I'll update manually.

nik9000:
> What a coincidence, I'll update manually.

:+1:

nik9000:
Could you rename the new parameter to `sortable`?

nik9000:
It's not a great name. But maybe a little better than other one...

nik9000 (approved):
Looks good to me. Let's get this merged. Are you interested in putting together some docs for this? That's it's own process. If not I can put something together in a bit.

amooncake:
Preview link: https://elasticsearch_89256.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/search-aggregations-bucket-filters-aggregation.html#sortable-buckets

nik9000:
> Preview link: https://elasticsearch_89256.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/search-aggregations-bucket-filters-aggregation.html#sortable-buckets

Nice!

I'm trying to get some folks from our clients team to have a look at this to make sure it doesn't break them.

nik9000:
Hey! I talked this over with the clients team and they felt like it'd be quite a lot of overhead for them to support the different "shape" here. Even though it's better, it may not be worth doing just because of their overhead. Especially when folks can sort on the client side. I'm going to invoke @qhoxie and @not-napoleon to have another look here and try and figure out if this is worth the overhead on their side. Sorry for the delay here. And for the uncertainty.

sethmlarson:
@amooncake Apologies for the delay, I've provided a summary of my thoughts on how to best accomplish this change here: https://github.com/elastic/elasticsearch/issues/83957#issuecomment-1304115505

If you're still interested in fixing this issue I believe the changes described above aren't far off what you have now, only need a little tweaking.

amooncake:
Since the `keyed` parameter is already used in the source code to indicate the keyed filters, the newly added `keyed` parameter in the DSL will be parsed as `keyedBucket` in the source code.
https://github.com/elastic/elasticsearch/blob/91a2d5ddf881575e0e33cb1c3d2a81cd4942f610/server/src/main/java/org/elasticsearch/search/aggregations/bucket/filter/FiltersAggregationBuilder.java#L42-L51
https://github.com/elastic/elasticsearch/blob/91a2d5ddf881575e0e33cb1c3d2a81cd4942f610/server/src/main/java/org/elasticsearch/search/aggregations/bucket/filter/FiltersAggregationBuilder.java#L290-L291
The current implementation feels a bit weird and confusing. But I don't know of a better name or implementation. @nik9000 @sethmlarson Could you help me with this?

nik9000 (approved):
OK! Sorry for the huge delay here. It looks right except for the versions are stale. I'll update it.

nik9000:
It is weird and confusing, yeah. But there are worse things, I guess.

nik9000:
@not-napoleon could you review this as well? I feel like it's blurred a bit for me and would appreciate another set of eyes on it.

not-napoleon (approved):
LGTM

nik9000 (commented):
There seems to be a backwards compatibility issue. Checking.

nik9000:
Thanks for all of your patience @amooncake.

amooncake:
Thanks @nik9000 !

PR #89264, opened 2022-08-11
Drop transport client from ping_schedule docs

DaveCTurner:
The docs for `transport.ping_schedule` note that the transport client
defaults to a 5s ping schedule, but this is no longer relevant. This
commit drops this from the docs, and also moves the docs for this
setting further down the page to reflect its relative unimportance.


henningandersen (approved):
LGTM.

PR #89266, opened 2022-08-11
Autoscaling requirement for empty tier

henningandersen:
Autoscaling would require a node-level size for an empty tier,
which seems wrong and caused ESS to scale the empty tier into
existence. Fixed to only report a node level size if the tier must exist.

Marked non-issue since this has not been released.


fcofdez (approved):
LGTM, good catch 👍

henningandersen:
Thanks for the quick review, Francisco!

PR #89275, opened 2022-08-11
Health info overview

gmarouli:
In this PR we are introducing the last piece of the push interaction model in which every node monitors their health and sends it to the health node in the following situations:
- when it changes from the last reported state
- when there is a new health node

**Implementation details**

To synchronize the code we use an atomic reference to the last successfully reported health info. This variable gets updated in the following ways:
- it gets updated if the health info was successfully received by the health node and last reported value hasn't changed since this method read from it.
- it gets reset (set to null without any preconditions) when there is a change in the health node. Effectively, when the health node changes, the health node doesn't have any health data so it makes sense for the last successfully reported health info to be `null` too.


gmarouli:
@elasticmachine elasticsearch-ci/part-2

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -103,51 +125,80 @@ void setEnabled(boolean enabled) {
...
+                client.execute(
+                    UpdateHealthInfoCacheAction.INSTANCE,
+                    new UpdateHealthInfoCacheAction.Request(nodeId, currentHealth),
+                    ActionListener.runAfter(listener, () -> {
+                        inProgress.set(false);
+                        // Scheduling happens after the flag inProgress is false, this ensures that
+                        // if the feature is enabled after the following schedule statement, the setEnabled
+                        // method will be able to schedule the next run, and it will not be a no-op.
+                        // We prefer to err towards an extra scheduling than miss the enabling of this feature alltogether.
+                        maybeScheduleNextRun(monitorInterval);
```
masseyke, line 193: We only schedule a next run if currentHealth is not equal to previousHealth. I assume we want to schedule it regardless right?

masseyke: And related, we'll want to set inProgress to false even if we don't run `UpdateHealthInfoCacheAction`.

gmarouli: Definitely, I completely missed it!

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/HealthStatus.java
```diff
@@ -34,6 +41,15 @@ public void writeTo(StreamOutput out) throws IOException {
         out.writeByte(value);
```
andreidan, line 41: we can use `writeEnum` and avoid the `byte` value we put in the enum 
(the enum ordinal serves this purpose already)

server/src/main/java/org/elasticsearch/health/HealthStatus.java
```diff
@@ -34,6 +41,15 @@ public void writeTo(StreamOutput out) throws IOException {
...
     }
 
+    public static HealthStatus from(StreamInput in) throws IOException {
+        byte b = in.readByte();
```
andreidan, line 45: we can use `readEnum` and remove the `values` map

server/src/main/java/org/elasticsearch/health/node/DiskHealthInfo.java
```diff
@@ -8,20 +8,66 @@
...
+
+        private final byte value;
+
+        Cause(byte value) {
+            this.value = value;
+        }
+
+        @Override
+        public void writeTo(StreamOutput out) throws IOException {
+            out.writeByte(value);
```
andreidan, line 57: `writeEnum`

server/src/main/java/org/elasticsearch/health/node/DiskHealthInfo.java
```diff
@@ -8,20 +8,66 @@
...
+        }
+
+        public byte value() {
+            return value;
+        }
+
+        public static DiskHealthInfo.Cause from(StreamInput in) throws IOException {
+            byte b = in.readByte();
```
andreidan, line 65: `readEnum`

server/src/main/java/org/elasticsearch/health/node/DiskHealthInfo.java
```diff
@@ -8,20 +8,66 @@
...
+        out.writeOptionalWriteable(cause);
+    }
+
+    public enum Cause implements Writeable {
+        NODE_OVER_HIGH_THRESHOLD((byte) 0),
+        NODE_OVER_THE_FLOOD_STAGE_THRESHOLD((byte) 1),
+        FROZEN_NODE_OVER_FLOOD_STAGE_THRESHOLD((byte) 2),
+        NODE_HAS_NO_DISK_STATS((byte) 3);
+
+        private static final Map<Byte, Cause> values = Arrays.stream(Cause.values())
```
andreidan, line 46: the enum  has the `values()` already. there's no need for this I don't think

server/src/main/java/org/elasticsearch/health/node/HealthInfoCache.java
```diff
@@ -0,0 +1,62 @@
...
+            // Resetting the cache is not synchronized for efficiency and simplicity.
+            // Processing a delayed update after the cache has been emptied because
+            // the node is not the health node anymore has small impact since it will
+            // be reset in the next round again.
+        } else if (diskInfoByNode.isEmpty() == false) {
+            diskInfoByNode = new ConcurrentHashMap<>();
+        }
+    }
+
+    // A shallow copy is enough because the inner data is immutable.
```
andreidan, line 58: What do we mean by "immutable" here?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -112,42 +132,77 @@ private void maybeScheduleNextRun(TimeValue time) {
...
+            }
+            DiskHealthInfo previousHealth = this.lastReportedDiskHealthInfo.get();
             DiskHealthInfo currentHealth = diskCheck.getHealth(healthMetadata, clusterState);
             if (currentHealth.equals(previousHealth) == false) {
-                logger.debug("Health status changed from {} to {}", previousHealth, currentHealth);
-                this.lastReportedDiskHealthInfo = currentHealth;
+                String nodeId = clusterService.localNode().getId();
+                ActionListener<AcknowledgedResponse> listener = ActionListener.wrap(response -> {
+                    // Update the last reported value if there was no change, that would mean that there
+                    // is a new health node that might have not received the value, so we should retry.
```
andreidan, line 172: `Update the last reported value if there was no change,` - no change in health status? We shouldn't in that case right? But we're in the `if (currentHealth.equals(previousHealth) == false) {` branch so we know for sure there was a health status change.

Apologies, I dont' really follow what  this comment is meant to convey?

server/src/main/java/org/elasticsearch/health/node/UpdateHealthInfoCacheAction.java
```diff
@@ -0,0 +1,109 @@
...
+import org.elasticsearch.common.io.stream.StreamInput;
+import org.elasticsearch.common.io.stream.StreamOutput;
+import org.elasticsearch.health.node.action.TransportHealthNodeAction;
+import org.elasticsearch.tasks.Task;
+import org.elasticsearch.threadpool.ThreadPool;
+import org.elasticsearch.transport.TransportService;
+
+import java.io.IOException;
+
+public class UpdateHealthInfoCacheAction extends ActionType<AcknowledgedResponse> {
```
andreidan, line 29: Shall we document this ?

server/src/main/java/org/elasticsearch/health/node/UpdateHealthInfoCacheAction.java
```diff
@@ -0,0 +1,109 @@
...
+        @Override
+        public ActionRequestValidationException validate() {
+            return null;
+        }
+
+        @Override
+        public void writeTo(StreamOutput out) throws IOException {
+            super.writeTo(out);
+            out.writeString(nodeId);
+            out.writeOptionalWriteable(diskHealthInfo);
```
andreidan, line 63: Is this optional?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -112,42 +132,77 @@ private void maybeScheduleNextRun(TimeValue time) {
...
+                    logger.debug(
+                        "Health info [{}] successfully send, last reported value: {}.",
+                        currentHealth,
+                        lastReportedDiskHealthInfo.get()
+                    );
+                }, e -> logger.error(() -> format("Failed to send health info [%s] to health node, will try again.", currentHealth), e));
+                client.execute(
+                    UpdateHealthInfoCacheAction.INSTANCE,
+                    new UpdateHealthInfoCacheAction.Request(nodeId, currentHealth),
+                    ActionListener.runAfter(listener, () -> {
```
andreidan, line 183: I know we're FWind the request, but we know who the health node is. 

Should we not direct the request to the specific node here? (avoid - most of the times - some network hops)

andreidan (commented):
Thanks for working on this Mary.

This generally looks good, left a few rather minor suggestions

server/src/main/java/org/elasticsearch/health/HealthStatus.java
gmarouli: I will adjust it in all places.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: Is this a network hop? I thought it would resolve if it's local or send it to the health node directly? If I understand you correctly, that assumption is wrong?

server/src/main/java/org/elasticsearch/health/node/UpdateHealthInfoCacheAction.java
gmarouli: Hm, no, on a second thought if it's missing we would use the unknown value never null.

server/src/main/java/org/elasticsearch/health/node/DiskHealthInfo.java
gmarouli: I did this only for retrieval, which is not necessary if we are using the `readEnum`. I removed it.

server/src/main/java/org/elasticsearch/health/node/DiskHealthInfo.java
gmarouli: removed

server/src/main/java/org/elasticsearch/health/node/DiskHealthInfo.java
gmarouli: Removed

server/src/main/java/org/elasticsearch/health/HealthStatus.java
gmarouli: done

server/src/main/java/org/elasticsearch/health/node/UpdateHealthInfoCacheAction.java
gmarouli: Yes, I overlooked it.

server/src/main/java/org/elasticsearch/health/node/HealthInfoCache.java
gmarouli: I mean that the values in the map are records so we do not need to worry if a value gets changed concurrently, for example it's not possible to change the `HealthStatus` in an instance of `DiskHealthInfo`. Do you think the comment is not necessary?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
andreidan: Ah you're correct it does execute locally (and `TransportHealthNodeAction` does the FW if necessary).

Nevermind :) Sorry for the noise (I was under the impression a random node is chosen)

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: No worries, I see it's not very clear. This comment is referring to this line: `lastReportedDiskHealthInfo.compareAndSet(previousHealth, currentHealth);`. What I intend to do with this is to ensure that we only set the `lastReportedDiskHealthInfo` if we know that the current health node received it. 

- A health node change will set `lastReportedDiskHealthInfo` to `null`, capturing that the current health node hasn't received any info.
- If the health node changed happened right after the info was sent to the previous health node and before the `lastReportedDiskHealthInfo` was updated, then `lastReportedDiskHealthInfo.compareAndSet(previousHealth, currentHealth)` will not go through. At the next monitoring run it will update the current health node.

While writing this though, I see there is a blind spot in this. If the `previousHealth` is `null`, then this will not catch it.

I will think of another solution :/.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: No worries, it's good to verify these things ;).

server/src/main/java/org/elasticsearch/health/node/UpdateHealthInfoCacheAction.java
```diff
@@ -0,0 +1,109 @@
...
+import org.elasticsearch.threadpool.ThreadPool;
+import org.elasticsearch.transport.TransportService;
+
+import java.io.IOException;
+
+public class UpdateHealthInfoCacheAction extends ActionType<AcknowledgedResponse> {
+
+    public static class Request extends ActionRequest {
+        private final String nodeId;
+        private final DiskHealthInfo diskHealthInfo;
```
masseyke, line 33: @gmarouli and I have discussed this outside of this PR, but it's probably worth mentioning the tradeoffs of putting a disk-specific field here (and I don't think this discussion is worth blocking this PR for, just starting it here). If we do this then that means that whenever we add a new indicator that uses the health node we'll need to modify this transport action and the one that fetches the data (which doesn't exist yet). I think it also means we'll need to modify HealthService for each one so that it knows which data to send to which indicator.
An alternative is to use a generic list or map and use reflection when serializing/deserializing the data. But that has some unpleasant tradeoffs as well.
So I think for now we're proposing that we'll have:

- Concrete fields in UpdateHealthInfoCacheAction and the upcoming FetchHealthInfoCacheAction for each health node data type
- Concrete fields in LocalHealthMonitor for each health node data type
- HealthService will keep health node indicators separate, and explicitly send the correct data to each health node indicator (or something like that).

And if we wind up with a large number of indicators that use the health node we'll revisit this because at that point the downsides of reflection might be worth it.

masseyke:
> For the request that fetches data from the health node (ie. `FetchHealthInfoCacheAction` ) I think the tradeoffs for the request might be different as usually we'll want all the data in one request (?)

@andreidan the working idea is that `LocalHealthMonitor` will use `UpdateHealthInfoCacheAction` to write data for all indicators for one node to the health node, where it will be lumped together with the data for all the other nodes. Then `FetchHealthInfoCacheAction` will retrieve the data for all indicators for all nodes. Then `HealthService` will send the data from all nodes to each indicator that is of the type that indicator needs. The discussion I started was around whether we'd explicitly enumerate this data in each place (`FetchHealthInfoCacheAction`, `UpdateHealthInfoCacheAction`, `FetchHealthInfoCacheAction`, and `HealthService`), or keep it in some kind of collection so that we only have to update `FetchHealthInfoCacheAction` when we add a new indicator.

server/src/main/java/org/elasticsearch/health/node/UpdateHealthInfoCacheAction.java
gmarouli: > HealthService will keep health node indicators separate, and explicitly send the correct data to each health node indicator (or something like that).

We could also pass all the health info to the indicators and each indicator will pick the information it needs, that might help with keeping the health service as clean as possible.

One more option that we could do to keep things generic but without introducing reflection, is that we keep a generic class `HealthInfo` that will contain the `HealthStatus` and the `Cause` and we use this for all indicators. Then we have a map as shown follows:
```
Map.of(
    "disk", new HealthInfo(HealthStatus.YELLOW, Cause.ABOVE_HIGH_WATERMARK)
    "threadpool", new HealthInfo(HealthStatus.GREEN, null)
    );
```

I would like to avoid reflection at all cost... But I am open to using strings as labels :).

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/node/UpdateHealthInfoCacheAction.java
gmarouli: Added

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: I changed the code to check the `lastSeenHealthNode` before it updates the last reported health info. This way we ensure that that when we update `lastReportedDiskHealthInfo` we are certain that the current health node received it. If not it will try again on the next try. 

What do you think?

gmarouli:
@elasticmachine update branch

server/src/internalClusterTest/java/org/elasticsearch/health/UpdateHealthInfoCacheIT.java
```diff
@@ -0,0 +1,116 @@
...
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.notNullValue;
+
+@ESIntegTestCase.ClusterScope(scope = ESIntegTestCase.Scope.TEST, numDataNodes = 4)
+public class UpdateHealthInfoCacheIT extends ESIntegTestCase {
```
andreidan, line 29: Thanks for adding the integration tests ❤️

server/src/internalClusterTest/java/org/elasticsearch/health/UpdateHealthInfoCacheIT.java
```diff
@@ -0,0 +1,116 @@
...
+
+    private static final DiskHealthInfo GREEN = new DiskHealthInfo(HealthStatus.GREEN, null);
+
+    public void testHappyFlow() throws Exception {
```
andreidan, line 33: Shall we use a more descriptive term? 
`testNodesReportHealth` or something along those lines?

server/src/internalClusterTest/java/org/elasticsearch/health/UpdateHealthInfoCacheIT.java
```diff
@@ -0,0 +1,116 @@
...
+                for (String nodeId : nodeIds) {
+                    assertThat(healthInfoCache.get(nodeId), equalTo(GREEN));
+                }
+            });
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to close internal cluster: " + e.getMessage(), e);
+        }
+    }
+
+    private static DiscoveryNode waitAndGetHealthNode(InternalTestCluster internalCluster) throws InterruptedException {
```
andreidan, line 98: Shall we mark this with @Nullable? And the usages can possibly `assertNonNull` before using the health node?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -112,42 +135,80 @@ private void maybeScheduleNextRun(TimeValue time) {
...
         maybeScheduleNextRun(TimeValue.ZERO);
     }
 
     @Override
     public void clusterChanged(ClusterChangedEvent event) {
-        if (prerequisitesFulfilled == false) {
-            prerequisitesFulfilled = event.state().nodesIfRecovered().getMinNodeVersion().onOrAfter(Version.V_8_5_0)
-                && HealthMetadata.getFromClusterState(event.state()) != null;
-            maybeScheduleNow();
+        DiscoveryNode current = HealthNode.findHealthNode(event.state());
```
andreidan, line 145: nit: shall we name this `currentHealthNode`

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -112,42 +135,80 @@ private void maybeScheduleNextRun(TimeValue time) {
...
+        prerequisitesFulfilled = event.state().nodesIfRecovered().getMinNodeVersion().onOrAfter(Version.V_8_5_0)
+            && HealthMetadata.getFromClusterState(event.state()) != null
+            && current != null;
+        if (prerequisitesFulfilled) {
+            DiscoveryNode previous = HealthNode.findHealthNode(event.previousState());
```
andreidan, line 150: nit: maybe on the naming bit `previousHealthNode`? 

IMO it'd make the following lines a bit more readable where we "talk" about previous/current

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -112,42 +135,80 @@ private void maybeScheduleNextRun(TimeValue time) {
...
-                logger.debug("Health status changed from {} to {}", previousHealth, currentHealth);
-                this.lastReportedDiskHealthInfo = currentHealth;
+                String nodeId = clusterService.localNode().getId();
+                String healthNodeId = lastSeenHealthNode.get();
+                ActionListener<AcknowledgedResponse> listener = ActionListener.wrap(response -> {
+                    // Update the last reported value only if the health node hasn't changed.
+                    if (Objects.equals(healthNodeId, lastSeenHealthNode.get())
+                        && lastReportedDiskHealthInfo.compareAndSet(previousHealth, currentHealth)) {
+                        logger.debug(
+                            "Health info [{}] successfully send, last reported value: {}.",
```
andreidan, line 180:
```suggestion
                            "Health info [{}] successfully sent, last reported value: {}.",
```

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -112,42 +135,80 @@ private void maybeScheduleNextRun(TimeValue time) {
...
-            inProgress.set(false);
-            // Scheduling happens after the flag inProgress is false, this ensures that
-            // if the feature is enabled after the following schedule statement, the setEnabled
-            // method will be able to schedule the next run, and it will not be a no-op.
-            // We prefer to err towards an extra scheduling than miss the enabling of this feature alltogether.
-            maybeScheduleNextRun(monitorInterval);
         }
     }
 
     DiskHealthInfo getLastReportedDiskHealthInfo() {
```
andreidan, line 205: I think we should mark this as `Nullable`

server/src/main/java/org/elasticsearch/health/node/HealthInfoCache.java
```diff
@@ -0,0 +1,62 @@
...
+        return healthInfoCache;
+    }
+
+    public void updateNodeHealth(String nodeId, DiskHealthInfo diskHealthInfo) {
+        diskInfoByNode.put(nodeId, diskHealthInfo);
+    }
+
+    @Override
+    public void clusterChanged(ClusterChangedEvent event) {
+        DiscoveryNode currentHealthNode = HealthNode.findHealthNode(event.state());
```
andreidan, line 41: Shall we shortcircuit this method a bit with a check for null on `currentHealthNode` and do an empty `return` in case it's `null`? 

I know we are not at risk of NPE because of the way we call `equals` below, but I'd rather have the 
nullable explicitly visible

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
andreidan: As before, shall we shortcircuit this branch with a null check for `previous` ? (for better visibility that it _can_ be null)

andreidan (commented):
Thanks for iterating on this Mary

Left a few more suggestions (all rather minor 🚀 )

server/src/internalClusterTest/java/org/elasticsearch/health/UpdateHealthInfoCacheIT.java
gmarouli: Thanks for helping find a way to do it ;)

server/src/internalClusterTest/java/org/elasticsearch/health/UpdateHealthInfoCacheIT.java
gmarouli: Hm, yes, it is cleaner this way.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: Yes, no objection.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: Done.

server/src/internalClusterTest/java/org/elasticsearch/health/UpdateHealthInfoCacheIT.java
gmarouli: okay!

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -112,42 +135,80 @@ private void maybeScheduleNextRun(TimeValue time) {
...
-            prerequisitesFulfilled = event.state().nodesIfRecovered().getMinNodeVersion().onOrAfter(Version.V_8_5_0)
-                && HealthMetadata.getFromClusterState(event.state()) != null;
-            maybeScheduleNow();
+        DiscoveryNode current = HealthNode.findHealthNode(event.state());
+        prerequisitesFulfilled = event.state().nodesIfRecovered().getMinNodeVersion().onOrAfter(Version.V_8_5_0)
+            && HealthMetadata.getFromClusterState(event.state()) != null
+            && current != null;
+        if (prerequisitesFulfilled) {
+            DiscoveryNode previous = HealthNode.findHealthNode(event.previousState());
+            if (Objects.equals(previous, current) == false) {
```
masseyke, line 151: We've discussed this in another channel, but this check doesn't work because at least sometimes `HealthMetadata.getFromClusterState(event.state())` becomes non-null in a ClusterChangedEvent after the ClusterChangedEvent where the health node changes. In that case `previous` will always be equal to `current`, and so the two variables below don't get reset, so the health updates are never sent to the new health node.

gmarouli: Nice find @masseyke , I was planning to rework this code to be more readable and now I will include your suggestion to compare the `currentHealthNode` with the `lastSeenHealthNode`. That should improve things considerably!

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/node/HealthInfoCache.java
gmarouli: I wanted to take the fact that the `currentHealthNode` becoming `null` as an event to trigger the reset of the values. I think it makes sense to reset it there. For this reason I do not want to do an empty return. But I am reworking the code to make it a bit cleaner. So we can check if the readability has improved on the reworked code.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: I do not see what we could shortcircuit here.... Can you be more specific?

gmarouli:
After the weekend gap I was able to find quite a few bugs with the code that was ensuring the single execution at any point in time.

1. The condition `if (inProgress.compareAndSet(false, true) && prerequisitesFulfilled)` was wrong because it would set the `inProgress` flag and then it wouldn't continue because of the `prerequisitesFulfilled` could be `false`.
2. If there was an exception in the `monitorHealth` before the call to the `runAfter` was scheduled we might end up with a `inProgress` stuck in `true`.

I believe these were difficult to find while coding and during the review because the concurrency code was mingled with the functional code. For this reason I tried to separate them so we can argue better about their behavior. 

Let me know what you think.

gmarouli:
@masseyke pointed out offline that it is possible to have multiple schedules running at the same time:
- schedule-1: running monitoring tasks at seconds: 1, 10, 20
- schedule-2: running monitoring tasks at seconds: 3, 13, 23
etc.

This is not desirable and it's a waste of resources. I will revert back from using `threadPool.scheduleUnlessShuttingDown` to using `threadPool.schedule` which gives a cancellable schedule.

I mark this PR as draft until I have finished the change.

gmarouli:
**Progress update**

I see that the scheduling logic is error prone. Bugs keep popping up which makes me think this is not well implemented. I think it's important for this code to be robust. If any node's update goes missing it will be hard to diagnose and even if it's easy to fix, it's disruptive. My next steps:
- Look into existing scheduling code and see if I reuse it, existing code might give more confidence since it's been in use longer.
- Otherwise separate the scheduling logic, so it can be also tested separately.

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -91,63 +116,226 @@ private void registerListeners() {
...
+        public boolean isCancelled() {
+            return run == false;
+        }
+
+        /**
+         * This method evaluates the health info of this node and if there is a change it sends an update request to the health node
+         */
+        @Override
+        public void run() {
+            boolean nextRunScheduled = false;
```
masseyke, line 270: We probably ought to check the value of the `run` variable here right? Since it could be that this was scheduled to run in 30 seconds in the future in `scheduleNextRunIfNecessary()` and was cancelled at some point in that 30 seconds.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -91,63 +116,226 @@ private void registerListeners() {
...
+                HealthMetadata healthMetadata = HealthMetadata.getFromClusterState(clusterState);
+                if (healthMetadata != null) {
+                    DiskHealthInfo previousHealth = this.lastReportedDiskHealthInfo.get();
+                    DiskHealthInfo currentHealth = diskCheck.getHealth(healthMetadata, clusterState);
+                    if (currentHealth.equals(previousHealth) == false) {
+                        String nodeId = clusterService.localNode().getId();
+                        String healthNodeId = lastSeenHealthNode.get();
+                        ActionListener<AcknowledgedResponse> listener = ActionListener.wrap(response -> {
+                            // Update the last reported value only if the health node hasn't changed.
+                            if (Objects.equals(healthNodeId, lastSeenHealthNode.get())
```
masseyke, line 282: We probably ought to check the value of `run` here at the beginning of the `if` block right? Since there could theoretically be a lot of time between when `client.execute()` was called and when this listener is called (it's possible that in the meantime this run could have been cancelled and there could have been other requests made I think).

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -91,63 +115,224 @@ private void registerListeners() {
...
+        }
+    }
+
+    private void maybeStartNewSchedule() {
         if (prerequisitesFulfilled && enabled) {
-            threadPool.scheduleUnlessShuttingDown(time, ThreadPool.Names.MANAGEMENT, this::monitorHealth);
+            // We synchronize on this block to ensure that we will not override an existing
+            // schedule that is not cancelled
+            // TODO check in which thread are the setting updates executed
+            synchronized (this) {
```
masseyke, line 151: I don't _think_ this needs to be synchronized, does it? This can only be called from the object initialization code or the cluster change event thread (which only begins calling this object at the very end of object initialization).

gmarouli: You are right. `synchronized` removed!

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: Yes indeed. Good catch!

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: Yes indeed. Good catch!

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: Hm, interesting thought. So you suggest to check if it was cancelled for two reasons:
- _Because it might have changed at a different thread_
In this case, I think we are covered since we only update these fields if `lastReportedDiskHealthInfo` hasn't changed and the health node is the same. Right?

- _Because, the monitoring task was cancelled._
I completely agree with you that we should check this at the beginning, but I wouldn't want to check it in multiple places in the code because then there will be arbitrary places that the execution will behave like it was interrupted if it was cancelled, I feel this would make it more difficult to argue about the code. Furthermore, I don't think that it's bad that these fields get updated after the task was cancelled since the execution was already in progress and it wasn't cancelled. In the end, the health info was sent to the current health node, and since this is what the current health node just received we should update them. What do you think?

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/node/HealthInfoCache.java
gmarouli: Ignore my previous answer, I thought this comment referred to different file. 

If the `currentHealthNode` is null we still want to empty the cache, so we cannot return immediately, but I can add an explicit check that `currentHealthNode` is not `null`.

andreidan (commented):
Thanks for iterating on this Mary

Left a round of comments (still testing this) but this looks great !

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
masseyke: The reason for the 2nd check was b/c it's potentially running much later since it's a listener. If `previousHealth` were null for both this one would still incorrectly get set. Maybe as you say though that doesn't matter since it came from the correct health node? It's probably the same result.

masseyke (approved):
LGTM

server/src/main/java/org/elasticsearch/health/node/DiskHealthInfo.java
```diff
@@ -8,20 +8,45 @@
...
+        out.writeOptionalEnum(cause);
+    }
+
+    public enum Cause {
+        NODE_OVER_HIGH_THRESHOLD((byte) 0),
+        NODE_OVER_THE_FLOOD_STAGE_THRESHOLD((byte) 1),
+        FROZEN_NODE_OVER_FLOOD_STAGE_THRESHOLD((byte) 2),
+        NODE_HAS_NO_DISK_STATS((byte) 3);
+
+        private final byte value;
```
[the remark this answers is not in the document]
gmarouli: Hm, in the beginning I thought adding specific values and implementing our own deserialization would make it less fragile to break during a maintenance. Somebody wouldn't accidentally change the order or something and break backwards compatibility. However, since we are using the built-in writable implementation, the `value` is not used at all anymore. I will remove it.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -91,63 +118,249 @@ private void registerListeners() {
...
+                    threadPool,
+                    lastReportedDiskHealthInfo,
+                    lastSeenHealthNode,
+                    diskCheck,
+                    clusterService,
+                    client
+                );
+                logger.debug("Monitoring task started {}", scheduledMonitoringTask);
+            } else {
+                logger.debug("Monitoring task already started {}, skipping", scheduledMonitoringTask);
```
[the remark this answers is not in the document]
gmarouli: Ah, you are right. Yes, I will change it.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -91,63 +118,249 @@ private void registerListeners() {
...
-        return lastReportedDiskHealthInfo;
+        return lastReportedDiskHealthInfo.get();
+    }
+
+    /**
+     * This class is responsible for running the health monitoring in the configured intervals. The first execution happens
+     * upon initialization. If there is an exception, it will log it and continue to schedule the next execution unless it
+     * was a EsRejectedExecutionException which will cancel the task.
+     */
+    static class MonitoringTask implements Runnable, Scheduler.Cancellable {
```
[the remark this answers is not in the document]
gmarouli: No, I think this is a valid concern and it can cause confusion in the future. I will try to come up with a better name.

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/node/DiskHealthInfo.java
```diff
@@ -8,20 +8,45 @@
...
+import org.elasticsearch.common.io.stream.Writeable;
 import org.elasticsearch.health.HealthStatus;
 
+import java.io.IOException;
+
 /**
  * The health status of the disk space of this node along with the cause.
  */
-record DiskHealthInfo(HealthStatus healthStatus, Cause cause) {
+public record DiskHealthInfo(HealthStatus healthStatus, Cause cause) implements Writeable {
```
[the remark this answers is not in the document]
gmarouli: Done!

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -91,63 +118,249 @@ private void registerListeners() {
...
+                                    currentHealth,
+                                    lastReportedDiskHealthInfo.get()
+                                );
+                            }
+                        }, e -> {
+                            if (e.getCause() instanceof NodeNotConnectedException
+                                || e.getCause() instanceof HealthNodeNotDiscoveredException) {
+                                logger.warn("Failed to connect to the health node [{}], will try again.", e.getCause().getMessage());
+                            } else {
+                                logger.error(
```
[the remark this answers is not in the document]
gmarouli: I see, yes, I do not have any strong objections. It might be difficult to catch it if something is going wrong here. But I can put it in debug and we will see if that is good enough in the longrun.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -91,63 +118,249 @@ private void registerListeners() {
...
+         * will not be interrupted but the task will not be scheduled anymore.
+         *
+         * @return false, if the task was already cancelled; {@code true}
+         * otherwise.
+         */
+        @Override
+        public boolean cancel() {
+            final boolean alreadyCancelled = cancelled;
+            cancelled = true;
+            if (alreadyCancelled == false) {
```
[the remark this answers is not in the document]
gmarouli: `scheduledTask.cancel()` but that refers to only one run. This cancels both the current run but also all subsequent runs, I think that's what makes it necessary. What do you think?

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -91,63 +118,249 @@ private void registerListeners() {
...
+        DiscoveryNode currentMasterNode = event.state().nodes().getMasterNode();
+        boolean healthNodeChanged = hasHealthNodeChanged(currentHealthNode, event);
+        boolean masterNodeChanged = hasMasterNodeChanged(currentMasterNode, event);
+        if (healthNodeChanged || masterNodeChanged) {
+            // On health node or on master node changes, the health node might be reset so the reported
+            // health info gets reset to null, to ensure it will be resent.
+            lastSeenHealthNode.set(currentHealthNode == null ? null : currentHealthNode.getId());
+            lastReportedDiskHealthInfo.set(null);
+            logger.debug(
+                "Change detected, resetting the health monitoring [masterNodeChanged={},healthNodeChanged={},newHealthNode={}]",
```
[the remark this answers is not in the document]
gmarouli: 😄 yes, sure

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -91,63 +118,249 @@ private void registerListeners() {
...
+        /**
+         * This method evaluates the health info of this node and if there is a change it sends an update request to the health node
+         */
+        @Override
+        public void run() {
+            if (cancelled) {
+                return;
+            }
+            boolean nextRunScheduled = false;
+            Runnable runOnceScheduleNextRunIfNecessary = new RunOnce(this::scheduleNextRunIfNecessary);
```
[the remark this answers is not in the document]
gmarouli: Yes, makes sense

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -91,63 +118,249 @@ private void registerListeners() {
...
+            }
+        }
+
+        private void scheduleNextRunIfNecessary() {
+            if (cancelled) {
+                return;
+            }
+            try {
+                scheduledTask = scheduler.schedule(this, interval, executor);
+            } catch (final EsRejectedExecutionException e) {
```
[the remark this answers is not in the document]
gmarouli: No strong feelings. I will change it, it feels a bit cleaner because then the schedule only gets cancelled from the user of this class.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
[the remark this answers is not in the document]
andreidan: the subsequent runs are cancelled by changing `cancelled` to true right? 
Which should still remain.

I was wondering if the `alreadyCancelled` dance is necessary since `scheduledTask.cancel` is idempotent. 

Could this cancel method be?
```
public boolean cancel() {
        cancelled = true;
       scheduledTask.cancel();
}
```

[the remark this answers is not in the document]
gmarouli: This return value of this is not the same. Because `scheduledTask.cancel()` might return false, if the run went through but for us, it should return that it got cancelled. We can simplify in the following way though:
```
public boolean cancel() {
            if (cancelled) {
                return false;
            }
            scheduledTask.cancel();
            return cancel = true;
}
```
or something similar to make it a bit more clear. What do you think?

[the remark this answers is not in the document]
gmarouli: After a short offline discussion with @andreidan we decided to keep an implementation that matches what Java does in `BaseFuture`:

> if future is NOT running return false // nothing to cancel
else cancel and return true

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -91,63 +118,257 @@ private void registerListeners() {
...
+        ) {
+            this.interval = interval;
+            this.executor = ThreadPool.Names.MANAGEMENT;
+            this.scheduler = scheduler;
+            this.lastReportedDiskHealthInfo = lastReportedDiskHealthInfo;
+            this.lastSeenHealthNode = lastSeenHealthNode;
+            this.clusterService = clusterService;
+            this.diskCheck = diskCheck;
+            this.client = client;
+            scheduledRun = scheduler.schedule(this, TimeValue.ZERO, executor);
```
andreidan, line 270: `this` is incomplete as it's in the constructor.
Shall we have a start/schedule/init method that kicks things off?

andreidan (approved):
LGTM, thanks for iterating on this Mary 🚀

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
```diff
@@ -91,63 +118,257 @@ private void registerListeners() {
...
+                    reason = "the master node and the health node";
+                } else if (healthNodeChanged) {
+                    reason = "the health node";
+                } else {
+                    reason = "the master node";
+                }
+                logger.debug(
+                    "Resetting the health monitoring because {} changed, current health node is {}.",
+                    reason,
+                    currentHealthNode == null ? null : format("{{%s}{%s}}", currentHealthNode.getName(), currentHealthNode.getId())
```
andreidan, line 195:
```suggestion
                    currentHealthNode == null ? null : format("[%s][%s]", currentHealthNode.getName(), currentHealthNode.getId())
```

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: Yes, same as we do with the cluster state listeners. On it.

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: In most logging lines I have encountered we use this format `{{%s}{%s}}` is there a reason you want to switch to the `[%s][%s]` for this log line?

gmarouli:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/health/node/LocalHealthMonitor.java
gmarouli: Nvm, it's looks cleaner with the square brackets :)

gmarouli:

PR #89276, opened 2022-08-11
[DOCS] Adds note about escaping backslashes in regex

abdonpijpelink:
The backslash character is an escape character in both JSON strings and regular expressions. This PR adds a note with an example about the need to escape both.

You can preview the change here: https://elasticsearch_89276.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/regexp-syntax.html#regexp-reserved-characters

### DIFF
--- a/docs/reference/query-dsl/regexp-syntax.asciidoc
+++ b/docs/reference/query-dsl/regexp-syntax.asciidoc
@@ -41,26 +41,10 @@ backslash or surround it with double quotes. For example:
 [NOTE]
 ====
 
-The backslash character is an escape character in both JSON strings and regular
+The backslash is an escape character in both JSON strings and regular
 expressions. You need to escape both backslashes in a query, unless you use a
-language client, which takes care of this. For example, the keyword string `a\b`
-needs to be indexed as `"a\\b"`:
-
-////
-[source,console]
-----
-PUT my-index-000001
-{
-  "mappings": {
-    "properties": {
-      "my_field": {
-        "type": "keyword"
-      }
-    }
-  }
-}
-----
-////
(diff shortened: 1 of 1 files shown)


immavalls (approved):
Thanks, @abdonpijpelink, for the fast response. It looks good to me, @crisdarocha might have more feedback as he's been involved in the investigation that led to this

crisdarocha (approved):
I like the change. Example is clear.

Would love to hear what Joe and Tim think about the topic. If this is enough, or if we need to be even more explicit.

docs/reference/query-dsl/regexp-syntax.asciidoc
```diff
@@ -37,7 +37,55 @@ backslash or surround it with double quotes. For example:
...
+
+The backslash character is an escape character in both JSON strings and regular
+expressions. You need to escape both backslashes in a query, unless you use a
+language client, which takes care of this. For example, the keyword string `a\b`
+needs to be indexed as `"a\\b"`:
+
+////
+[source,console]
+----
+PUT my-index-000001
```
joegallo, line 52: I'm not sure the mapping here should be hidden -- if you try to reproduce this without using the same mapping (i.e. with `my_field` either implicitly or explicitly as `"type": "text"`) then the example doesn't quite work.

I defer to others, though, this is outside my wheelhouse.

abdonpijpelink: @joegallo I like your suggestion to simplify the example by removing the mapping and querying `my_field.keyword` instead. I've updated the example accordingly.

debadair (approved):
LGTM. Thanks for adding this!

PR #89279, opened 2022-08-11
[DOCS] Double quote the env variable in curl command

ywangd:
The env variable can contain special chacacters. Without quoting it
could mis-behave when used in the curl command of the setup docker
container.


lockewritesdocs (approved):
LGTM 🦖 Thanks for making this change @ywangd!

PR #89281, opened 2022-08-11
Add support for floating point node.processors setting

fcofdez:
This commit adds support for floating point node.processors setting.
This is useful when the nodes run in an environment where the CPU
time assigned to the ES node process is limited (i.e. using cgroups).
With this change, the system would be able to size the thread pools
accordingly, in this case it would round up the provided setting
to the closest integer.


server/src/main/java/org/elasticsearch/common/util/concurrent/EsExecutors.java
```diff
@@ -37,13 +37,31 @@ public class EsExecutors {
     /**
      * Setting to manually control the number of allocated processors. This setting is used to adjust thread pool sizes per node. The
      * default value is {@link Runtime#availableProcessors()} but should be manually controlled if not all processors on the machine are
-     * available to Elasticsearch (e.g., because of CPU limits).
+     * available to Elasticsearch (e.g., because of CPU limits). Note that this setting accepts floating point processors, so it should not
+     * be used directly, use always {@link EsExecutors#allocatedProcessors(Settings)} to obtain a rounded number of processors.
      */
-    public static final Setting<Integer> NODE_PROCESSORS_SETTING = Setting.intSetting(
+    public static final Setting<Double> NODE_PROCESSORS_SETTING = new Setting<>(
```
henningandersen, line 43: I think there is a documentation change that we need to make as well.

henningandersen (commented):
I think this looks good, though I'll leave it unapproved to ensure others have a change to chime in about this change.

docs/reference/modules/threadpool.asciidoc
```diff
@@ -177,7 +177,9 @@ thread_pool:
 The number of processors is automatically detected, and the thread pool settings
 are automatically set based on it. In some cases it can be useful to override
 the number of detected processors. This can be done by explicitly setting the
-`node.processors` setting.
+`node.processors` setting. This setting accepts floating point numbers, this
+can be useful in environments where the Elasticsearch nodes are configured
+to run with CPU limits, such as `Cgroups`.
```
henningandersen, line 182:
```suggestion
to run with CPU limits, such as cpu shares or quota under `Cgroups`.
```

server/src/main/java/org/elasticsearch/common/util/concurrent/EsExecutors.java
```diff
@@ -37,13 +37,31 @@ public class EsExecutors {
     /**
      * Setting to manually control the number of allocated processors. This setting is used to adjust thread pool sizes per node. The
      * default value is {@link Runtime#availableProcessors()} but should be manually controlled if not all processors on the machine are
-     * available to Elasticsearch (e.g., because of CPU limits).
+     * available to Elasticsearch (e.g., because of CPU limits). Note that this setting accepts floating point processors, so it should not
+     * be used directly, use always {@link EsExecutors#allocatedProcessors(Settings)} to obtain a rounded number of processors.
```
henningandersen, line 41:
```suggestion
     * available to Elasticsearch (e.g., because of CPU limits). Note that this setting accepts floating point processors. If a rounded number is
     * needed, always use {@link EsExecutors#allocatedProcessors(Settings)}.
```

server/src/main/java/org/elasticsearch/common/util/concurrent/EsExecutors.java
```diff
@@ -37,13 +37,31 @@ public class EsExecutors {
...
-        Runtime.getRuntime().availableProcessors(),
+        Double.toString(Runtime.getRuntime().availableProcessors()),
+        textValue -> {
+            double numberOfProcessors = Double.parseDouble(textValue);
+            if (Double.isNaN(numberOfProcessors) || Double.isInfinite(numberOfProcessors)) {
+                String err = "Failed to parse value [" + textValue + "] for setting [node.processors]";
+                throw new IllegalArgumentException(err);
+            }
+
+            if (numberOfProcessors < Double.MIN_VALUE) {
```
henningandersen, line 53: Just curious, but is this different from `numberOfProcessors <= 0.0d`?

server/src/test/java/org/elasticsearch/common/util/concurrent/EsExecutorsTests.java
```diff
@@ -437,7 +438,7 @@ public void testGetTasks() throws InterruptedException {
     }
 
     public void testNodeProcessorsBound() {
-        final Setting<Integer> processorsSetting = EsExecutors.NODE_PROCESSORS_SETTING;
+        final Setting<Double> processorsSetting = EsExecutors.NODE_PROCESSORS_SETTING;
         final int available = Runtime.getRuntime().availableProcessors();
         final int processors = randomIntBetween(available + 1, Integer.MAX_VALUE);
```
henningandersen, line 443: Can we change this to a float here and rather than `+ 1`, add something potentially small (`Math.ulp` maybe)?

server/src/test/java/org/elasticsearch/common/util/concurrent/EsExecutorsTests.java
```diff
@@ -452,4 +453,50 @@ public void testNodeProcessorsBound() {
...
+            is(equalTo(1))
+        );
+
+        assertThat(
+            EsExecutors.allocatedProcessors(Settings.builder().put(EsExecutors.NODE_PROCESSORS_SETTING.getKey(), 1.2).build()),
+            is(equalTo(2))
+        );
+
+        assertThat(
+            EsExecutors.allocatedProcessors(Settings.builder().put(EsExecutors.NODE_PROCESSORS_SETTING.getKey(), 3).build()),
```
henningandersen, line 473: I think this requires CI to run on a node with at least 3 CPUs. Is that always the case?

henningandersen (commented):
A few minor comments, the only worrying one is the one about CI.

server/src/main/java/org/elasticsearch/common/util/concurrent/EsExecutors.java
fcofdez: Not really, I can change it to `numberOfProcessors <= 0.0d` if that's clearer

server/src/test/java/org/elasticsearch/common/util/concurrent/EsExecutorsTests.java
fcofdez: I think it should be fine, but I'll change to some value within the available processors to avoid issues.

henningandersen (approved):
LGTM.

fcofdez:
Thanks Henning!

PR #89292, opened 2022-08-12
move log-related logic into log block in IndexLifecycleRunner

mushao999:
This PR moved some logic that is only used by log in IndexLifecycleRunner#isReadyToRansitionToThisPhase(...) into trace-log- block which may save cpu and memory if trace log was not enabled.


dakrone:
@elasticmachine ok to test

docs/changelog/89292.yaml
```diff
@@ -0,0 +1,5 @@
+pr: 89292
+summary: move log-related logic into log block in IndexLifecycleRunner
+area: ILM
```
dakrone, line 3: This needs to conform to a specific enum to pass CI, so can you change this to:

```suggestion
area: ILM+SLM
```

dakrone (approved):
LGTM, thanks!

dakrone:
@elasticmachine update branch

Issue #89285, opened 2022-08-11
[CI] JdkDownloadPluginFuncTest transforms of type UnzipTransform are kept across builds failing


iverase:
**Build scan:**
https://gradle-enterprise.elastic.co/s/ernsozqztszym/tests/:integTest/org.elasticsearch.gradle.internal.JdkDownloadPluginFuncTest/transforms%20of%20type%20UnzipTransform%20are%20kept%20across%20builds

**Reproduction line:**
`null`

**Applicable branches:**
main

**Reproduces locally?:**
Didn't try

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.gradle.internal.JdkDownloadPluginFuncTest&tests.test=transforms%20of%20type%20UnzipTransform%20are%20kept%20across%20builds

**Failure excerpt:**
```
java.io.IOException: Failed to copy full contents from '/tmp/junit1699766051303869009/junit4162579748942594183/caches/journal-1/journal-1.lock' to 'build/test-debug/junit1699766051303869009/junit4162579748942594183/caches/journal-1/journal-1.lock'

  at org.apache.commons.io.FileUtils.doCopyFile(FileUtils.java:1087)
  at org.apache.commons.io.FileUtils.doCopyDirectory(FileUtils.java:1358)
  at org.apache.commons.io.FileUtils.doCopyDirectory(FileUtils.java:1356)
  at org.apache.commons.io.FileUtils.doCopyDirectory(FileUtils.java:1356)
  at org.apache.commons.io.FileUtils.doCopyDirectory(FileUtils.java:1356)
  at org.apache.commons.io.FileUtils.copyDirectory(FileUtils.java:1319)
  at org.apache.commons.io.FileUtils.copyDirectory(FileUtils.java:1191)
  at org.apache.commons.io.FileUtils.copyDirectory(FileUtils.java:1160)
  at org.elasticsearch.gradle.fixtures.AbstractGradleFuncTest.cleanup(AbstractGradleFuncTest.groovy:51)

```

mark-vieira:
@breskeby did we want to keep [the bits here](https://github.com/elastic/elasticsearch/blob/48d30637910b1cb3d85690606b4a8e036203f28f/build-tools/src/testFixtures/groovy/org/elasticsearch/gradle/fixtures/AbstractGradleFuncTest.groovy#L50) commented out? Or perhaps could we be smarter here and only keep the temporary directory contents when the test failed?

[...]

PR #89304, opened 2022-08-12
Unify handling of custom Gradle User home in build tool tests

breskeby:
- keep guh separated from test project dir
- unify folder handling

as a side effect when keeping the project dir for debugging failed tests we do not copy the whole GUH which usually isn't providing any additional help for debugging those


breskeby:
Should reduce risk of running into https://github.com/elastic/elasticsearch/issues/89285

Issue #88536, opened 2022-07-14
[ML] Add test to reproduce rare feature importance issue


pheyos:
In the Kibana functional tests, we have two data frame analytics jobs, that sometimes (rarely) don't produce any features although they usually do. In order to stay focus on UI behavior, we'll adjust the Kibana tests to not fail in that case, but it would be good to cover the same scenario in Elasticsearch integration tests, so if it fails again it's easier to debug.

Here is what the Kibana tests are doing (left away the Kibana specific data view creation and space sync operation):
- Load the `ihp_outlier` data archive (loads data into the `ft_ihp_outlier` index
  - this is a downsampled version of our Iowa house prices dataset, which is mainly used for outlier detection (hence the name), but also works for classification
  - loading is done via Kibana's `esArchiver` tool, which basically just creates the index with the provided mappings file and then loads the data file (new line separated JSON blobs, not NDJSON)
  - Mappings file: [mappings.tar.gz](https://github.com/elastic/elasticsearch/files/9110845/mappings.tar.gz)
  - Data file: [data.json.gz](https://github.com/elastic/elasticsearch/files/9110840/data.json.gz)
- Create classification jobs using the ES API (job configurations are listed below)
- Run the jobs (first wait for training record count to be positive, so we know it actually started, then wait for state to be stopped so we know it's done)
- Look at the results

<details>
<summary>Job configuration 1</summary>

```
{
    id: 'ihp_fi_binary',
    description: 'binary classification job',
    source: {
        index: ['ft_ihp_outlier'],
        query: {match_all: {}},
    },
    dest: {
        index: `user-ihp_fi_binary`,
        results_field: 'ml_central_air',
    },
    analyzed_fields: {
        includes: [
            'CentralAir',
            'GarageArea',
            'GarageCars',
            'YearBuilt',
            'Electrical',
            'Neighborhood',
            'Heating',
            '1stFlrSF',
        ],
    },
    analysis: {
        classification: {
            dependent_variable: 'CentralAir',
            num_top_feature_importance_values: 5,
            training_percent: 35,
            prediction_field_name: 'CentralAir_prediction',
            num_top_classes: -1,
            max_trees: 10,
        },
    },
    model_memory_limit: '60mb',
    allow_lazy_start: false,
}
```
</details>

<details>
<summary>Job configuration 2</summary>

```
{
    id: 'ihp_fi_multi',
    description: 'multi class classification job',
    source: {
        index: ['ft_ihp_outlier'],
[...]

PR #89307, opened 2022-08-12
[ML] Feature importance test for house pricing data

valeriy42:
This PR implements a test to capture missing feature importance values for the house pricing dataset described in #88536.

Since the failure occurs very irregularly, we want to capture the random seed to be able to dig into a specific case.

It doesn't affect the functionality; hence I mark this PR as a non-issue.


dimitris-athanasiou (approved):
LGTM

PR #89311, opened 2022-08-12
Refactor WildcardExpressionResolver for reusability

albertzaharovits:
This is a pure refactoring that shouldn't exhibit any behavior change.
The goal here is to specify the interface of `WildcardExpressionResolver#innerResolve`,
i.e. what are the properties of its return value.
The ultimate goal is to facilitate the replacement of `IndicesAndAliasesResolver#resolveIndicesAndAliases`.


server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -664,7 +664,12 @@ public Set<String> resolveExpressions(ClusterState state, String... expressions)
             getSystemIndexAccessPredicate(),
             getNetNewSystemIndexPredicate()
         );
-        return Set.copyOf(resolveExpressions(Arrays.asList(expressions), context));
+        Collection<String> resolved = resolveExpressions(Arrays.asList(expressions), context);
+        if (resolved instanceof Set<String>) {
```
grcevski, line 668: I think this change slightly changes the semantics. It appears that the underlying resolve methods can return a mutable Set, while previously with Set.copyOf we would always return unmodifiable set. Since this is a public method, I think we should retain the immutable semantics.

Slight improvement to the code, with the instanceof expressions now we can avoid the cast on the return:

```
if (resolved instanceof Set<String> resolvedSet) {
   return Set.copyOf(resolvedSet);
}

File: server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java [hunk 2]
```diff
@@ -1200,17 +1216,26 @@ private static Set<String> innerResolve(Context context, List<String> expression
...
+                expandMatches(context, matches.values(), expression.startsWith("."), expanded -> {
+                    if (add) {
+                        finalResult.add(expanded);
+                    } else {
+                        finalResult.remove(expanded);
+                    }
+                });
+            }
+            if (result == null) {
+                result = expressions;
```
grcevski, line 1233: Is there a potential issue with us returning the original list here instead of a null? Is it safer to just return an empty Set, or do we need the original expressions? I think there's a code path where a non-empty expressions list can still lead to a null result.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
albertzaharovits: This has the potential to be a long list; there are benchmarks where this is 50k. I've preferred to avoid the copying.
If we want to maintain the unmodifiable semantics, should we instead use `Collections.unmodifiableSet`, as that doesn't copy the Set?

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
albertzaharovits: I think this is safe, and returning an empty Set would not work.

This PR is **very** hard to follow as a whole, but I've tried to make every commit self-contained, and I think it would be easier to review the changes incrementally commit by commit, eg https://github.com/elastic/elasticsearch/pull/89311/commits/9b90714bd4c63ec2d1961c499f680c81a2dd1d79 for this change.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1124,48 +1129,59 @@ private WildcardExpressionResolver() {
...
-        private static Set<String> innerResolve(Context context, List<String> expressions) {
-            Set<String> result = null;
+        /**
+         * Returns all the existing resource (index, alias and datastream) names that the {@param expressions} resolve to.
+         * The passed-in {@param expressions} might contain wildcards and exclusions, as well as plain resource names.
+         * The return is a {@code Collection} (usually a {@code Set} but can also be a {@code List}, for performance reasons) of plain
+         * resource names only. All the returned resources are "accessible", in the given context, i.e. the resources exist
+         * and are not an alias or a datastream if the context does not permit it.
+         * Wildcard expressions, depending on the context:
+         *   - might not cover hidden or system resources (but plain names can refer to hidden or system resources)
```
jakelandis, line 1178: nit: might not "cover" .. cover == resolve, right ? if so I think it would be a bit more clear to use "resolve" as above.

jakelandis (commented):
Overall is looking.  I went commit by commit to try identify any changes in behavior, and haven't been able to spot any behavior changes. 

2 areas were a bit difficult to parse 
1) returning null vs. list of expressions (as mentioned by @grcevski )
2) changes to the implicit hidden

In both cases I think I just need a bit more time to wrap my head around the functionality code and not just evaluating like for like at the Java code level. Any hints / explanations in these two areas are much appreciated.  

Also huge thanks for breaking up this by commits as well as the javadocs.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
grcevski: Yes, thanks using an unmodifiableSet would work great.

albertzaharovits: I have pushed https://github.com/elastic/elasticsearch/pull/89311/commits/da5565a2aa07c73ba9aeec23f7a15110f1f5da75  . I think ideally the return result of `resolveExpressions` should be unmodifiable, ie without making it so at this level, but maybe I'll tackle that later.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
albertzaharovits: To be clear, the [original code](https://github.com/elastic/elasticsearch/blob/1403ab318b14ad39030c6824237e1477a9392dd8/server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java#L1167) returned `null` when all the passed-in expression items existed, in which case the caller would detect the `null` and itself return the passed-in expression. It is safe to move the "if null return expression argument" condition inside this method because the expression argument cannot cannot be empty, see also: https://github.com/elastic/elasticsearch/commit/9b90714bd4c63ec2d1961c499f680c81a2dd1d79 .

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1200,17 +1216,26 @@ private static Set<String> innerResolve(Context context, List<String> expression
...
                     throw indexNotFoundException(expression);
                 }
-                Set<String> expand = expand(context, matches, expression);
-                if (add) {
-                    result.addAll(expand);
-                } else {
-                    result.removeAll(expand);
-                }
+                Collection<String> finalResult = result;
+                expandMatches(context, matches.values(), expression.startsWith("."), expanded -> {
```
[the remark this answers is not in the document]
albertzaharovits: > Can we call expressions that start with dots implicitly hidden ? 

Not quite. The logic is that wildcard expressions that start with "." will cover hidden indices that also start with ".", even if the request options set `expandWildcardsHidden` to `false`.

> For example Regex.isSimpleMatchPattern(expression) is not checked here as part of the implicitHidden check and hard to tell if the result is like for like.

`Regex.isSimpleMatchPattern(expression)` is assumed `true` at this stage, in the `expandMatches` method (expansion happens only for wildcards).
The intention is to eliminate the `expression` variable from the expansion method. The `expression` is used in matches, but then expanding the matches shouldn't need the `expression` anymore.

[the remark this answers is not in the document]
albertzaharovits: Tangentially:

Depending on the context, individual wildcards that expand to the empty list will throw exceptions.

But this is not `true` in the case that the wildcard expands to hidden or system indices which are then disregarded (so it's as if the wildcard expanded to nothing, but no error is thrown).
I would consider this a bug, and solving this would actually simplify the code here because the `expansion` method would not be concerned about whether the expression starts with a "." or not.

If you agree this is a bug I would LOVE to fix it.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1335,47 +1360,39 @@ private static Map<String, IndexAbstraction> filterIndicesLookup(
...
-                        if (context.isPreserveDataStreams() && indexAbstraction.getType() == IndexAbstraction.Type.DATA_STREAM) {
-                            expand.add(indexAbstraction.getName());
+                if (indexAbstraction.isHidden()
+                    && context.getOptions().expandWildcardsHidden() == false
+                    && (indexAbstraction.getName().startsWith(".") && implicitlyIncludeHidden) == false) {
+                    continue;
+                }
+                if (context.isPreserveAliases() && indexAbstraction.getType() == IndexAbstraction.Type.ALIAS) {
+                    expandConsumer.accept(indexAbstraction.getName());
+                } else if (context.isPreserveDataStreams() && indexAbstraction.getType() == Type.DATA_STREAM) {
```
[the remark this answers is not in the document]
albertzaharovits: If the `context.isPreserveDataStreams() && indexAbstraction.getType() == Type.DATA_STREAM` condition is `false`, in the original code, only the `for` block below is executed. But if it is `true`, the `for` AND this `elseif` blocks are both executed, in the original code.
But there are only 2 places in the code where `ontext.isPreserveDataStreams()` is `true`: https://github.com/elastic/elasticsearch/blob/1403ab318b14ad39030c6824237e1477a9392dd8/server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java#L208 and https://github.com/elastic/elasticsearch/blob/1403ab318b14ad39030c6824237e1477a9392dd8/server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java#L208  and in both places the `for` is either not run or its output is discarded.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1124,48 +1129,59 @@ private WildcardExpressionResolver() {
...
+                final List<IndexAbstraction> dataStreamsAbstractions = context.getState()
+                    .metadata()
+                    .getIndicesLookup()
+                    .values()
+                    .stream()
+                    .filter(indexAbstraction -> indexAbstraction.getType() == IndexAbstraction.Type.DATA_STREAM)
+                    .collect(Collectors.toList());
+                // dedup backing indices if expand hidden indices option is true
+                Set<String> resolvedIncludingDataStreams = new HashSet<>(resolvedExpressions);
+                expandMatches(context, dataStreamsAbstractions, false, resolvedIncludingDataStreams::add);
```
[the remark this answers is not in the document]
albertzaharovits: Side note, I think the `expandMatches` here should eventually be removed, but that's a conversation to have another time.

Wildcard expressions that start with "." must cover hidden indices that start with "." . In this case, the expression is "*" or "_all" so it's not a wildcard that starts with "." so it will not cover hidden indices that with ".", which is the meaning of the `false` flag param.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
albertzaharovits: I have pushed https://github.com/elastic/elasticsearch/pull/89311/commits/6b3f9684e4a775a046c41a68258562b26a99fcff

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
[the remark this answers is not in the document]
albertzaharovits: In Security, a wildcard that expands only to hidden/system resources will indeed throw the `indexNotFound` exception, so I would even more lean in treating this behavior as a bug (ie that a wildcard that expands to only hidden/system resources does NOT throw the `indexNotFound` exception).

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1335,47 +1362,39 @@ private static Map<String, IndexAbstraction> filterIndicesLookup(
             }
         }
 
-        private static Set<String> expand(Context context, Map<String, IndexAbstraction> matches, String expression) {
+        private static void expandMatches(
+            Context context,
+            Collection<IndexAbstraction> matches,
+            final boolean implicitlyIncludeHidden,
```
jakelandis, line 1368: Can you change this variable name to expressionStartsWithDot ?

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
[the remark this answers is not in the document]
albertzaharovits: @jakelandis Above I'm proposing that wildcards that ultimately expand to nothing because they only expand to hidden or system indices, that are ultimately not resolved, throw the index not found exception (this is what happens when Security is enabled or when the wildcard expands to no index at all). In a chat, you've asked what happens to the hidden backing indices of data streams.

Because said indices start with `.ds-` a wildcard that accesses them must also start with `.` (or be the `_all` or `*`, which is a special case outside of this discussion), for eg `.ds*`. All wildcards that start with `.`  will match indices that start with `.` even if the indices are hidden. So wildcards for backing indices of datastreams are not concerned about the change.

So, for example, this is what I'm proposing to fix:
 * say there is only a single `index` in the cluster, which is created as a hidden index. `index*/_search?allow_no_indices=false` returns empty result (only when Security is disabled), but `test*/_search?allow_no_indices=false` throws the index_not_found 400 exception.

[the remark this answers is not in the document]
jakelandis: I am pretty sure that is bug.  Per our doc: https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#multi-hidden the request should error. 

> allow_no_indices
    (Optional, Boolean) If false, the request returns an error if any wildcard expression, [index alias](https://www.elastic.co/guide/en/elasticsearch/reference/current/aliases.html), or _all value targets only missing or closed indices. This behavior applies even if the request targets other open indices. For example, a request targeting foo*,bar* returns an error if an index starts with foo but no index starts with bar.

In your example, if this was the request
`index*/_search?allow_no_indices=false&expand_wildcards=hidden` then it should return results (and not an error).

jakelandis (approved):
LGTM. As far as I can tell this is like for like behavior with the elder code.

grcevski (approved):
LGTM! Thanks for all the iterations!

albertzaharovits:
@elasticsearchmachine update this please

albertzaharovits:
@elasticmachine update branch

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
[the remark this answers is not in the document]
albertzaharovits: > In your example, if this was the request
index*/_search?allow_no_indices=false&expand_wildcards=hidden then it should return results (and not an error).

Correct.

I'm going to follow-up with a quick PR to fix this.

PR #89313, opened 2022-08-12
Fix LocalAllocateDangledIndices reroute usage

idegtiarenko:


server/src/main/java/org/elasticsearch/gateway/LocalAllocateDangledIndices.java
```diff
@@ -52,24 +53,24 @@ public class LocalAllocateDangledIndices {
...
         AllocationService allocationService,
-        IndexMetadataVerifier indexMetadataVerifier
+        IndexMetadataVerifier indexMetadataVerifier,
+        ThreadPool threadPool
     ) {
         this.transportService = transportService;
         this.clusterService = clusterService;
         this.allocationService = allocationService;
         this.indexMetadataVerifier = indexMetadataVerifier;
+        this.threadPool = threadPool;
```
DaveCTurner, line 73: I think we could just use `transportService.getThreadPool()` (or `clusterService.getClusterApplierService().threadPool()`), no need to capture this separately.

idegtiarenko: I did this initially but then decided to wire own thread pool instead of "borrowing" it from somewhere else. I am concerned about relying on implementation details of other services

DaveCTurner: If we can't rely on `transportService.getThreadPool()` then we have bigger concerns :)

server/src/main/java/org/elasticsearch/gateway/LocalAllocateDangledIndices.java
```diff
@@ -105,6 +103,23 @@ public void messageReceived(final AllocateDangledRequest request, final Transpor
                 indexNames[i] = request.indices[i].getIndex().getName();
             }
             final String source = "allocation dangled indices " + Arrays.toString(indexNames);
+
+            var listener = new AllocationActionListener<>(ActionListener.<AllocateDangledResponse>wrap(response -> {
+                try {
+                    channel.sendResponse(new AllocateDangledResponse());
```
DaveCTurner, line 109: I think we already have an `AllocateDangledResponse` to hand, no need to make another one?

```suggestion
                    channel.sendResponse(response);
```

server/src/main/java/org/elasticsearch/gateway/LocalAllocateDangledIndices.java
```diff
@@ -105,6 +103,23 @@ public void messageReceived(final AllocateDangledRequest request, final Transpor
...
+                }
+            }, exception -> {
+                logger.error(() -> "unexpected failure during [" + source + "]", exception);
+                try {
+                    channel.sendResponse(exception);
+                } catch (Exception inner) {
+                    inner.addSuppressed(exception);
+                    logger.warn("failed send response for allocating dangled", inner);
+                }
+            }), transportService.getThreadPool().getThreadContext());
```
DaveCTurner, line 121: In fact, can we just do this? I don't think the logging has much value here now that this action is triggered by a REST call and therefore we can send errors back to the client.

```suggestion
            var listener = new AllocationActionListener<>(
                new ChannelActionListener<AllocateDangledResponse, AllocateDangledRequest>(channel, task.getAction(), request),
                transportService.getThreadPool().getThreadContext()
            );
```

DaveCTurner (approved):
LGTM but I left a couple of optional suggestions for simplifications.

PR #89315, opened 2022-08-12
[DOCS] Fixes broken example in ingest pipeline tutorial

### LINKED ISSUE #89213: [Docs] Two back slashes do not work in the "parse logs in the common log format" example
image

The correct one should be a single back slash:
image

abdonpijpelink:
Fixes https://github.com/elastic/elasticsearch/issues/89213


liu-xiao-guo:
I think it is OK. Thanks

liu-xiao-guo (commented):
It looks good to me!

liu-xiao-guo (approved):

PR #89316, opened 2022-08-12
[ML] Improve reason when autoscaling capacity cannot be computed

dimitris-athanasiou:
When we cannot compute autoscaling capacity we call
`MlAutoscalingDeciderService.buildDecisionAndRequestRefresh`.
There are 4 callers. 3 of them set their own specific reason before
the call. However, the method overwrites the reason with the generic
`MEMORY_STALE` msg. This commit addresses this issue which should
bubble up more detailed reasons for those edge cases.


droberts195 (approved):
LGTM

droberts195:
If this problem exists in 7.17 and the code hasn't diverged too much to make a backport hard then it's probably worth backporting to 7.17.6 too.

dimitris-athanasiou:
@droberts195 Code looks different in `7.17` (https://github.com/elastic/elasticsearch/blob/7.17/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlAutoscalingDeciderService.java).

PR #89318, opened 2022-08-12
Check circuit breaker before sending join request

### LINKED ISSUE #85003: Join validation has a tight retry loop if the joining node hits parent circuit breaker
### Elasticsearch Version

8.1 and many more

### Installed Plugins

N/A

### Java Version

_bundled_

### OS Version

N/A

### Problem Description

If a joining node has heap usage in excess of the parent circuit breaker then it will still request to join the cluster. The master will send it a copy of the cluster state for validation which it'll immediately reject with a `CircuitBreakingException`, failing the join, and triggering a pretty-much-immediate retry.

### Steps to Reproduce

Increase sensitivity of parent circuit breaker on a node and observe that it tries and fails to join the cluster in a tight loop.

### Logs (if relevant)

_No response_

DaveCTurner:
Adds a simple preflight check to `JoinHelper#sendJoinRequest` to avoid
sending a join request if it looks like the `inflight_requests` circuit
breaker is going to trip on the join validation message.

Closes #85003

### DIFF
--- a/server/src/internalClusterTest/java/org/elasticsearch/discovery/DiscoveryDisruptionIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/discovery/DiscoveryDisruptionIT.java
@@ -16,7 +16,10 @@ import org.elasticsearch.cluster.coordination.PublicationTransportHandler;
 import org.elasticsearch.cluster.node.DiscoveryNode;
 import org.elasticsearch.cluster.node.DiscoveryNodes;
 import org.elasticsearch.cluster.service.ClusterService;
+import org.elasticsearch.common.breaker.CircuitBreaker;
+import org.elasticsearch.common.breaker.CircuitBreakingException;
 import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.indices.breaker.CircuitBreakerService;
 import org.elasticsearch.test.ESIntegTestCase;
 import org.elasticsearch.test.disruption.NetworkDisruption;
 import org.elasticsearch.test.disruption.ServiceDisruptionScheme;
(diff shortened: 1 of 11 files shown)


server/src/internalClusterTest/java/org/elasticsearch/discovery/DiscoveryDisruptionIT.java
```diff
@@ -257,4 +260,60 @@ public void testJoinWaitsForClusterApplier() throws Exception {
...
+            assertNotEquals(action, JoinHelper.JOIN_ACTION_NAME);
+            connection.sendRequest(requestId, action, request, options);
+        });
+
+        // fix the network disruption
+        logger.info("--> removing network disruption");
+        masterTransportService.clearAllRules();
+        ensureStableCluster(2, masterName);
+
+        Thread.sleep(20000);
```
idegtiarenko, line 304: Is it possible to remove or at least reduce this sleep?

server/src/internalClusterTest/java/org/elasticsearch/discovery/DiscoveryDisruptionIT.java
```diff
@@ -257,4 +260,60 @@ public void testJoinWaitsForClusterApplier() throws Exception {
...
+        final var victimTransportService = (MockTransportService) internalCluster().getInstance(TransportService.class, victimName);
+        victimTransportService.addSendBehavior((connection, requestId, action, request, options) -> {
+            assertNotEquals(action, JoinHelper.JOIN_ACTION_NAME);
+            connection.sendRequest(requestId, action, request, options);
+        });
+
+        // fix the network disruption
+        logger.info("--> removing network disruption");
+        masterTransportService.clearAllRules();
+        ensureStableCluster(2, masterName);
```
idegtiarenko, line 302: I am surprised this is still 2 nodes. Should this permit to connect the node again?

server/src/internalClusterTest/java/org/elasticsearch/discovery/DiscoveryDisruptionIT.java
DaveCTurner: Ah oops that was only to make sure the test really did what it said :) Gone now.

server/src/internalClusterTest/java/org/elasticsearch/discovery/DiscoveryDisruptionIT.java
DaveCTurner: No, the victim node still can't join the cluster here because it has a tripping circuit breaker (that's the point!). We have to release the resources tracked by the CB to get it to join again.

idegtiarenko (approved):

Issue #89183, opened 2022-08-08
File Based Settings for Elasticsearch


grcevski:
### Description

# Introduction

Watches a directory under Elasticsearch config for changes to a file called settings.json. This is a JSON file describing the settings that the operator wants to set that will be immutable through the REST API. Any changes to these settings will only be allowed through modifying the file. As configured by default we watch for a path operator/settings.json under the config directory. The name of the directory operator can be changed via settings.

The File Settings Service will notice updates to the file and perform an update. The operator directory doesn't have to exist, the file setting service will notice when it's created. Recommended way to change operator file based settings is through modifying the settings.json file in a brand new location and then performing a symlink under the Elasticsearch config directory.

Initial support for File Based Settings was delivered in 8.4 through this PR: https://github.com/elastic/elasticsearch/pull/88329

# Remaining work

- [x] Restore file settings state after snapshot restore (#89321)
- [x] /_cluster/settings (v8.4.0)
- [x] /_ilm/policy/<policy_id> (v8.4.0)
- [x] ​​/_security/role_mapping (https://github.com/elastic/elasticsearch/pull/89667)
- [x] /_autoscaling/policy (https://github.com/elastic/elasticsearch/pull/89708)
- [x] /_snapshot/cloud-snapshots (https://github.com/elastic/elasticsearch/pull/89601)
- [x] /_slm/policy (https://github.com/elastic/elasticsearch/pull/89567)
- [x] /_ingest/pipeline/<pipeline> (https://github.com/elastic/elasticsearch/pull/89735)
- [x] /_index_template/<index-template> (https://github.com/elastic/elasticsearch/pull/90143)
- [x] /_component_template/<component-template>  (https://github.com/elastic/elasticsearch/pull/90143)

grcevski:
Complete.

PR #89321, opened 2022-08-12
Handle snapshot restore in file settings

grcevski:
# Background information

File based settings create a reserved cluster state, preventing REST mutation of any settings/entities created in this mode. The only way these settings can be modified is through a change in the operator/settings.json file inside our config directory.

When we do snapshot restore we can also bring back the global cluster state, effectively rewriting the cluster state and with it any reserved state. This introduces few issues related to file based settings:

- If there is an operator/settings.json file, it effectively won't be the source of truth anymore, until an update is made to this file.
- If the restore is from another cluster, the metadata version of the reserved state can be higher than the current cluster version accounting, meaning settings updates might fail even if the file is updated.
- If the cluster we are restoring into doesn't use file settings, but the snapshot we are restoring had some reserved state, we would be permanently locking from updates those settings that were reserved.

# Adding support for snapshots in file based settings

To support snapshots in file based settings we take the following approach:

1. If the snapshot doesn't restore the global state, we don't do anything.
2. In case we are restoring the global state, we now call the file settings service to:
    - If there's an operator/settings.json file: reset the reserved state metadata version to  0 and touch the file so that the watcher service will apply the settings.
    - If there's no operator/settings.json, we remove the reserved cluster state effectively unlocking those settings that were reserved.


server/src/main/java/org/elasticsearch/reservedstate/ReservedClusterStateHandler.java
```diff
@@ -28,8 +28,6 @@
  * </p>
  */
 public interface ReservedClusterStateHandler<T> {
-    String CONTENT = "content";
```
grcevski, line 31: This was just some dead code, left from earlier refactoring.

grcevski:

server/src/main/java/org/elasticsearch/cluster/metadata/ReservedStateMetadata.java
```diff
@@ -214,6 +214,20 @@ public Builder(String namespace) {
...
         }
 
+        /**
+         * Creates an reserved state metadata builder
+         *
+         * @param metadata  the previous metadata
+         */
+        public Builder(ReservedStateMetadata metadata) {
+            this(metadata.namespace);
+            if (metadata != null) {
```
ChrisHegarty, line 224: Is this null check redundant? On the previous line, `metadata.namespace` will throw a NPE if metadata is null

ChrisHegarty (approved):
LGTM.

PR #89322, opened 2022-08-12
Fixing a version check for master stability functionality

masseyke:
The version check for fetching remote master. history was backwards.

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/coordination/MasterHistoryService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/coordination/MasterHistoryService.java
@@ -115,7 +115,7 @@ public class MasterHistoryService {
      */
     public void refreshRemoteMasterHistory(DiscoveryNode node) {
         Version minSupportedVersion = Version.V_8_4_0;
-        if (node.getVersion().onOrAfter(minSupportedVersion) == false) { // This was introduced in 8.3.0 and modified in 8.4.0
+        if (node.getVersion().before(minSupportedVersion)) { // This was introduced in 8.3.0 and modified in 8.4.0
             logger.trace(
                 "Cannot get master history for {} because it is at version {} and {} is required",
                 node,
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/elasticsearch/cluster/coordination/MasterHistoryService.java
```diff
@@ -115,7 +115,7 @@ public List<DiscoveryNode> getRemoteMasterHistory() throws Exception {
      */
     public void refreshRemoteMasterHistory(DiscoveryNode node) {
         Version minSupportedVersion = Version.V_8_4_0;
-        if (node.getVersion().onOrAfter(minSupportedVersion)) { // This was introduced in 8.3.0
+        if (node.getVersion().onOrAfter(minSupportedVersion) == false) { // This was introduced in 8.3.0 and modified in 8.4.0
```
joegallo, line 118: Is `.onOrAfter(minSupportedVersion) == false` better as `.before(minSupportedVersion)`?

gmarouli: +1 

One more think, this confuses me a bit, should it be `node.getVersion().onOrAfter(Version.V_8_3_0)`? 

Because the way it's now, I understand that it means that this will only work if it's older than `8.4.0`? So, for `8.5.0` it won't work?

masseyke: The is the path that just returns (doesn't do the request). We want it to just return on anything earlier than 8.4.0.

gmarouli: Ah clear!

masseyke:
@elasticsearchmachine run elasticsearch-ci/docs-check

joegallo (approved):

PR #89330, opened 2022-08-14
Simplified invalidating ssl session

Hakky54:
This is a small improvement on the part of invalidating ssl sessions. I discovered this part when I was just reading the source code. I refactored to use Java Stream API to make it more readable. It does not change the functionality.

### DIFF
--- a/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ssl/SSLService.java
+++ b/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ssl/SSLService.java
@@ -48,7 +48,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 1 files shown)


DaveCTurner:
Hi @Hakky54, thanks for the suggestion. I think we shouldn't make this change for a few reasons.

1. The [contributing guidelines](https://github.com/elastic/elasticsearch/blob/main/CONTRIBUTING.md#reviewing-and-accepting-your-contribution) say we typically _reject isolated PRs that only perform simple refactorings or otherwise "tidy up" certain aspects of the code_. That looks to apply here, unless this change is a prequel to a bigger and more meaningful change of which I'm currently unaware. And even then ...

2. ... the performance effects of this change need more analysis than I think we're willing to do for those (debatable) readability benefits. From a quick glance it seems we're no longer running in O(1) memory and O(1) allocations - we allocate a whole extra list to make a copy of the session IDs before iterating over them, and the list is created by dynamically growing an `ArrayList` so there's O(log N) hidden allocations here too. Quite possibly this doesn't matter in this context, but that's where the review time investment outweighs the benefits.

3. Finally it loses the comment. Comments like the one you're deleting are useful and important.

Therefore I'm closing this.

Issue #78273, opened 2021-09-23
Validate HTTP links in the source code


jakelandis:
The deprecation info API is used during upgrades to help users better understand the changes that are necessary before they upgrade. To help present this information each entry exposes an HTTP link for additional documentation.  These HTTP links are embedded in the source code and like all remote links have the potential to be broken. 

We recently switched to using [shortened links](https://github.com/elastic/elasticsearch/pull/78026) which will allow us to better mitigate any broken links. However, we should also include some CI validation that these links are still good. 

Ideally, with both shortened URLs and validation we could also consider opening the door to allow exceptions, deprecation logs, server logs, etc. to also embed HTTP links if it makes sense.  

I assume we would need to :

1) externalize all of the HTTP links to a resource file then 
2) create a gradle task that can validate the response code and possibly an expected regex the
3) wire that check into CI.

PR #89333, opened 2022-08-15
Return 400 error for GetUserPrivileges call with API keys

ywangd:
The GetUserPrivileges API returns a 500 error when it is called with an
API key that has assigned role descriptors. This is because the
underlying LimitedRole class that represents the API key's effective
privileges does not support building a simple view of the privileges.

This PR changes the code to return 400 error instead of 500 along with a
better error message that suggests the GetApiKey API as an alternative.

Relates: #89058


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportGetUserPrivilegesAction.java
```diff
@@ -42,12 +45,25 @@ public TransportGetUserPrivilegesAction(
...
         }
 
-        authorizationService.retrieveUserPrivileges(subject, securityContext.getAuthorizationInfoFromContext(), listener);
+        try {
+            authorizationService.retrieveUserPrivileges(subject, securityContext.getAuthorizationInfoFromContext(), listener);
+        } catch (UnsupportedOperationException e) {
+            if (authentication.isApiKey()) {
+                throw new ElasticsearchSecurityException(
+                    "Cannot retrieve privileges for API keys with assigned role descriptors. "
+                        + "Please use the Get API key information API https://ela.st/es-api-get-api-key",
```
ywangd, line 61: @jakelandis I am not entirely clear about the current recommendation of adding HTTP links into error messages (or more general HTTP responses). My previous understanding is that we don't want to do it because the links may break. But I have noticed that links are being used today, e.g. in the health API. Could you please adivse whether it is OK to use such a short url here? If not, what would be our choices? Thanks!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportGetUserPrivilegesAction.java
```diff
@@ -42,12 +45,25 @@ public TransportGetUserPrivilegesAction(
...
+        final Subject subject = authentication.getEffectiveSubject();
         if (subject.getUser().principal().equals(request.username()) == false) {
             listener.onFailure(new IllegalArgumentException("users may only list the privileges of their own account"));
             return;
         }
 
-        authorizationService.retrieveUserPrivileges(subject, securityContext.getAuthorizationInfoFromContext(), listener);
+        try {
+            authorizationService.retrieveUserPrivileges(subject, securityContext.getAuthorizationInfoFromContext(), listener);
+        } catch (UnsupportedOperationException e) {
```
ywangd, line 57: I decided to use `try-catch` instead of pro-actively checking the Role class because:
1. Whether LimitedRole can throw exception on building privileges response is an implementation detail. The LimitedRole class is in a separate module (`core`) and in theory does not even need to be exposed to the security module. In fact, security module currently has no direct reference to it.
2. Checking the Role class does not really work with Custom Authorization engine by checking because it is theoretically possible for a custom engine to implement the `retrieveUserPrivileges` call in a different way.

n1v0lg: My only reservation here is that we are leaking this implementation detail all the way up to the transport layer in an unstructured way: we are assuming that any `UnsupportedOperationException` thrown by `retrieveUserPrivileges` for an API key is because our implementation of `retrieveUserPrivileges` doesn't support this. `UnsupportedOperationException` is generic though so this could lead to unexpected behavior/incorrect error messages for custom implementations (or if we throw `UnsupportedOperationException` somewhere else throughout `retrieveUserPrivileges`).

I'm wondering if we should specialize the exception thrown in `LimitedRole` to a sublass of `UnsupportedOperationException`. This way we still leak the implementation but in a more robust, structured way. 

Alternatively, we could throw the 400 API-key specific exception inside `RBACEngine` which would keep it closer to the root cause and not force our interpretation on exceptions thrown by custom implementations.

WDYT?

n1v0lg (commented):
A question around where and how we throw, before I review the rest of the PR. 

Question inline, in response to your [comment here](https://github.com/elastic/elasticsearch/pull/89333/files#r945432922).

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportGetUserPrivilegesAction.java
ywangd: I agree it's awkward. The problem is not every API key should fail. Only those with assigned role descriptors will fail. Underlying those API keys have their roles represented as `LimitedRole`. But I don't want to expose `LimiteRole` to the security module because it should just rely on the `Role` interface. I pondered on this and didn't find a clear better solution:

1. Throwing specific subclass of `UnsupportedOperationException` (or just a specific exception) is slightly better. But it still leaks implementation. I also didn't want to add another class just for this error message. My thought was that the exception + checking authentication being an API key (`authentication.isApiKey()` inside the catch block) is sufficient to rule out accidental unexpected behaviour in future. 
2. By "throwing 400 API key specific exception inside RBACEngine", I assume you mean actively checking whether the `Role` is a `LimitedRole` and throw error accordingly? This would require `RBACEngine` be aware of `LimitedRole` which I rather dislike as stated above.
3. If you mean use try-catch for `UnsupportedOperationException` in RBACEngine, this means putting the error message inside RBACEngine. I initially dislike this because I felt this error message, especially with the HTTP url should be a handle level thing. But I now noticed that similar messages are prepared at other service classes (mostly implementations of the new `HealthIndicatorService`). So maybe this is not too bad.

So overall, I think option 3 is my current preference. More concretely, catching `UnsupportedOperationException` in `RBACEngine#getUserPrivileges` and retrow it as a `IllegalArgumentException` with proper error message. Using `IllegalArgumentException` avoids having to set an explicit 400 status code which is something I also didn't like at service layer. Please let me know if this works for you.

n1v0lg: Exactly, I meant option 3 👍

n1v0lg (approved):
LGTM!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportGetUserPrivilegesAction.java
n1v0lg: I don't have a strong opinion on this but I'm generally in favor of the link in the error message. It's helpful and potentially helps us cut down on lengthy failure messages.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/RBACEngine.java
```diff
@@ -658,7 +658,17 @@ public void getUserPrivileges(AuthorizationInfo authorizationInfo, ActionListene
             );
         } else {
             final Role role = ((RBACAuthorizationInfo) authorizationInfo).getRole();
-            listener.onResponse(buildUserPrivilegesResponseObject(role));
+            final GetUserPrivilegesResponse getUserPrivilegesResponse;
+            try {
+                getUserPrivilegesResponse = buildUserPrivilegesResponseObject(role);
+            } catch (UnsupportedOperationException e) {
+                throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
ywangd: I was trying to be consistent with the existing behaviour, i.e. `buildUserPrivilegesResponseObject` just throws exceptions. But consistent with the caller method is also good. I'll make the change

ywangd:
Test failure is being fixed by https://github.com/elastic/elasticsearch/pull/89360/files#r946240748

ywangd:
@elasticmachine update branch

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportGetUserPrivilegesAction.java
jakelandis: Both deprecation's and new health API use shortened URL links embedded in the code. The primary reason is that they will be viewed inside Kibana. I am a fan of using URL links in error messages (especially shortened links) but I don't think we have widely adopted them quite yet. While the shortening service itself is solid, some of the tooling and management of the shortened URLs are lacking. For example: https://github.com/elastic/elasticsearch/issues/78273 and creating/updating these is a hodgepodge of custom scripts and manual efforts. We have also had some conversations around version specific links, but settled on just pointing to current version and we can figure that part out later. 

I don't think we have an official stance, but IMO if it is only additive information to the error message I don't see any harm. If the docs move, they generally leave behind a re-direct or page to manually redirect and having the link is better than not. I imagine existing usages and increased usage will force a better eco-system around managing these. cc @masseyke (since he worked with these recently)

ywangd: Thanks for the context, Jake. The link you shared is super helpful! I like the proposal it puts forward. It's good to know that we are thinking ways to make things more robust.

For this particular usage here, my take is that it should be fine. Once the broader framework is ready, we just need to remember migrating it to the new system. The `https://ela.st` part is pretty easy to search upon. So I am not worried at all.\

ywangd:
@elasticmachine update branch

PR #89335, opened 2022-08-15
[DOCS] Adds unattended setting to transforms API docs

szabosteve:
## Overview

Related to https://github.com/elastic/elasticsearch/pull/89212.

This PR documents a new transform setting called `unattended`. The setting is added to the create, preview, and update transform API docs.

### Preview

[PUT transform](https://elasticsearch_89335.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/put-transform.html#put-transform-request-body)


hendrikmuhs (approved):
LGTM

I noticed that the settings aren't ordered. I think it makes sense to order them as we now have a longer list of options.

szabosteve:
@hendrikmuhs Yup, you are right. It makes sense to have the settings in order. I added it to my to do list and open a separate PR for that soon.

PR #89336, opened 2022-08-15
Report better error for GCS credentials load failure

DaveCTurner:
Today if the GCS credentials file setting is invalid we report some kind
of JSON parsing error but it's not clear what JSON is being parsed so
the error is hard to track down. This commit adds the problematic
setting name to the exception message.


fcofdez (approved):
LGTM, I was writing a similar PR. You were faster than me 😄.

DaveCTurner:
:) sorry I had a branch that did this from a while back, not sure why I never merged it.

fcofdez:
No problem! 👍

PR #78876, opened 2021-10-08
Deprecate lenient parsing of bulk actions

arteam:
Make sure there are no arbitrary fields after an action declaration and that it gets properly closed by a curly bracket.

Resolves #43774

PR #89339, opened 2022-08-15
Force rejection of unsupported bulk actions in v9

DaveCTurner:
Adds a mention of `Version.V_7_17_0` so that we don't forget to remove
this deprecated behaviour in the next major version.

Relates #78876

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/bulk/BulkRequestParser.java
+++ b/server/src/main/java/org/elasticsearch/action/bulk/BulkRequestParser.java
@@ -8,6 +8,7 @@
 
 package org.elasticsearch.action.bulk;
 
+import org.elasticsearch.Version;
 import org.elasticsearch.action.DocWriteRequest;
 import org.elasticsearch.action.delete.DeleteRequest;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/elasticsearch/action/bulk/BulkRequestParser.java
```diff
@@ -184,6 +185,7 @@ public void parse(
                 }
                 String action = parser.currentName();
                 if (SUPPORTED_ACTIONS.contains(action) == false) {
+                    assert Version.CURRENT.major == Version.V_7_17_0.major + 1;
```
idegtiarenko, line 188: Why not just `assert Version.CURRENT.major == Version.V_8_0_0.major;` ?

DaveCTurner: The point is to mention a symbol that will be removed in v9 so that this is a compile error (the assertion failure is kinda not the point)

idegtiarenko (approved):

DaveCTurner:
Thanks Ievgen :)

PR #89347, opened 2022-08-15
[ML] Skip renormalization after node shutdown API called

droberts195:
A node can be informed that it is about to be shut down
using the node shutdown API. When this happens we
gracefully stop the jobs on the node and they persist
state. This state persistence includes latest quantiles,
and usually receipt of new quantiles triggers a
renormalization. However, in the case of an impending
node shutdown we do not want to be kicking off new
processes that may delay the shutdown.

This PR changes the anomaly detection job results
processor so that it will not trigger a renormalization
based on quantiles received after a node shutdown message
is received.

### DIFF
--- a/docs/changelog/89347.yaml
+++ b/docs/changelog/89347.yaml
@@ -1,0 +1,5 @@
+pr: 89347
+summary: Skip renormalization after node shutdown API called
+area: Machine Learning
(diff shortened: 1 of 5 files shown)


edsavage (approved):
LGTM

PR #89352, opened 2022-08-15
Fix duplication bug for source fallback in numeric types

jdconrad:
Currently, source fallback numeric types do not match doc values numeric types. Source fallback numeric types de-duplicate numeric values in multi-valued fields. This change removes the de-duplication for source fallback values for numeric types using value fetchers. This also adds test cases for all the supported source fallback types to ensure they continue to match their doc values counterparts exactly.

Note the majority of the change is additional test cases.


romseygeek (commented):
This looks great.  I have the same question as I did on the unsigned_long PR - can we use a single ArrayList and clear/repopulate it on each doc rather than allocating a new one each time?

jdconrad:
@romseygeek Thanks for this review as well. I updated based on the feedback.

romseygeek (approved):
LGTM

PR #89358, opened 2022-08-15
Prepare synthetic source docs for tech-preview

nik9000:
Now that we're releasing synthetic _source as a tech preview feature, we
no longer want to remove the docs from the non-release builds. And we
want to mark all of the headings describing synthetic `_source` as a
preview.


romseygeek (approved):
LGTM

kilfoyle (approved):
LGTM! 🚗

nik9000:
Backporting with #89392

PR #89359, opened 2022-08-15
Document kNN with aggregations

jtibshirani:
This commit adds a short note to the 'search your data' docs around kNN search
to explain how approximate kNN works with aggregations:
* Make section on 'hybrid retrieval' more general and include aggregations info
* Remove an example response from the previous section on filtering, since this
  page was getting long

### DIFF
--- a/docs/reference/search/search-your-data/knn-search.asciidoc
+++ b/docs/reference/search/search-your-data/knn-search.asciidoc
@@ -279,41 +279,8 @@ POST image-index/_search
 ----
 // TEST[continued]
 
-[source,console-result]
-----
-{
-  "took": 5,
-  "timed_out": false,
-  "_shards": {
-    "total": 1,
-    "successful": 1,
-    "skipped": 0,
-    "failed": 0
-  },
-  "hits": {
-    "total": {
-      "value": 1,
-      "relation": "eq"
-    },
-    "max_score": 0.003144654,
-    "hits": [
-      {
-        "_index": "image-index",
-        "_id": "2",
-        "_score": 0.003144654,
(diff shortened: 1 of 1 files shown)


docs/reference/search/search-your-data/knn-search.asciidoc
```diff
@@ -354,6 +321,11 @@ each score in the sum. In the example above, the scores will be calculated as
 score = 0.9 * match_score + 0.1 * knn_score
 ```
 
+The `knn` option can also be used with <<search-aggregations, `aggregations`>>. In general, {es} computes aggregations
+over all documents that match the search. So for approximate kNN search, aggregations are calculated on the top `k`
+nearest documents. If the search also includes a `query`, then aggregations are calculated on the combined set of `knn`
+and `query` matches.
+
```
mayya-sharipova, line 328: I guess it is obvious here, that if a `query` is not specifically provided aggregations are calculated only on the top `k` from knn query. Or do you think we should all add this clarification, as this is different how aggs would work without knn section?

mayya-sharipova (approved):
@jtibshirani Thanks Julie for this clarification!

docs/reference/search/search-your-data/knn-search.asciidoc
jtibshirani: I tried to explain that in this sentence: "So for approximate kNN search, aggregations are calculated on the top `k` nearest documents." Is there something we could add to clarify this?

jtibshirani:
Thanks @mayya-sharipova, I'm going to merge as I find the current text to be clear.

PR #89365, opened 2022-08-16
Stable Plugin API: Add Ubermodule ClassLoader for non-modularized stable plugins

williamrandolph:
This PR's UberModuleClassLoader takes a group of jars and loads them as a single synthetic module.

The motivation for this work is the Stable Plugin API project, but the classloader is general enough for other use cases. In the case of stable plugins, we don't want to require that plugin developers modularize their plugins. (If a plugin developer chooses to modularize their plugin, then we can load it using JDK-provided utilities.)

The advantage of a synthetic module is that we can programmatically use module layers to control what the code from the synthetic module has access to. A non-modularized plugin will be provided in the form of one or more jars. When this classloader loads a class from one of those jars, that class will be bound to the synthetic module. We can use a deny-list to prevent the class from accessing packages from particular modules.

In the world of modules, each module owns the packages it contains, and it is forbidden to split packages across module boundaries. This means that when we load a class, we can check its package name. If our synthetic module owns the package, then we don't need to search for the class in parent classloaders. When fetching jar resources, however, we simply delegate to the internal URLClassLoader.


server/src/main/java/org/elasticsearch/plugins/UberModuleClassLoader.java
```diff
@@ -0,0 +1,281 @@
...
+        Configuration cf,
+        ModuleLayer mparent,
+        Set<String> moduleDenyList,
+        Set<String> packageNames
+    ) {
+        super(parent);
+
+        this.internalLoader = new URLClassLoader(jarURLs);
+        // code source is always the first jar on the list
+        this.codeSource = new CodeSource(jarURLs[0], (CodeSigner[]) null);
```
ChrisHegarty, line 127: we'll need to revisit this, but it is fine for the initial implementation.

server/src/main/java/org/elasticsearch/plugins/UberModuleClassLoader.java
```diff
@@ -0,0 +1,281 @@
...
+        this.moduleController = ModuleLayer.defineModules(cf, List.of(mparent), s -> this);
+        this.module = this.moduleController.layer().findModule(moduleName).orElseThrow();
+
+        // Every module reads java.base by default, but we can add all other modules
+        // that are not in the deny list so that plugins can use, for example, java.management
+        mparent.modules()
+            .stream()
+            .filter(Module::isNamed)
+            .filter(m -> "java.base".equals(m.getName()) == false)
+            .filter(m -> moduleDenyList.contains(m.getName()) == false)
```
ChrisHegarty, line 138: deny-list is great, but at some point it might be more straightforward to have an allow-list, say to allow readability to all the JDK modules and a subset of the ES modules.

server/src/main/java/org/elasticsearch/plugins/UberModuleClassLoader.java
```diff
@@ -0,0 +1,281 @@
...
+
+        try (InputStream in = internalLoader.getResourceAsStream(rn)) {
+            if (in == null) {
+                return null;
+            }
+            byte[] bytes = in.readAllBytes();
+            return defineClass(name, bytes, 0, bytes.length, codeSource);
+        } catch (IOException e) {
+            // TODO
+            throw new IllegalStateException(e);
```
ChrisHegarty, line 180: we should probably log here or something, right?

server/src/main/java/org/elasticsearch/plugins/UberModuleClassLoader.java
```diff
@@ -0,0 +1,281 @@
...
+    protected URL findResource(String moduleName, String name) {
+        if (Objects.isNull(moduleName) || this.module.getName().equals(moduleName) == false) {
+            return null;
+        }
+        return findResource(name);
+    }
+
+    /**
+     * This classloader does not restrict access to resources in its jars. Users should
+     * expect the same behavior as that provided by {@link URLClassLoader}.
```
ChrisHegarty, line 203: this seems pragmatic.

server/src/main/java/org/elasticsearch/plugins/UberModuleClassLoader.java
```diff
@@ -0,0 +1,281 @@
...
+    protected Enumeration<URL> findResources(String name) throws IOException {
+        return internalLoader.findResources(name);
+    }
+
+    @Override
+    protected Class<?> loadClass(String cn, boolean resolve) throws ClassNotFoundException {
+        // if cn's package is in this ubermodule, look here only (or just first?)
+        String packageName = packageName(cn);
+
+        // TODO: do we need to check the security manager here?
```
ChrisHegarty, line 231: let's revisit this later, I wanna spend a bit more time thinking on it.

server/src/main/java/org/elasticsearch/plugins/UberModuleClassLoader.java
williamrandolph: I switched it to `UncheckedIOException`, since that's what we're throwing in `EmbeddedImplClassLoader`.

server/src/test/java/org/elasticsearch/plugins/UberModuleClassLoaderTests.java
```diff
@@ -0,0 +1,428 @@
...
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0 and the Server Side Public License, v 1; you may not use this file except
+ * in compliance with, at your election, the Elastic License 2.0 or the Server
+ * Side Public License, v 1.
+ */
+
+package org.elasticsearch.plugins;
+
+import joptsimple.internal.Strings;
```
ChrisHegarty, line 11: Might be more straightforward to use `org.elasticsearch.common.Strings`?

server/src/main/java/org/elasticsearch/plugins/UberModuleClassLoader.java
```diff
@@ -0,0 +1,281 @@
...
+                resolveClass(c);
+            }
+
+            return c;
+        }
+    }
+
+    /**
+     * Returns the package name for the given class name
+     * (from jdk.internal.loader.Loader)
```
ChrisHegarty, line 261: we could remove this comment (referring to jdk.internal), as the implementation is trivial.

server/src/test/java/org/elasticsearch/plugins/UberModuleClassLoaderTests.java
williamrandolph: This was a mistake. Thanks for catching it!

ChrisHegarty (approved):
LGTM.

server/src/main/java/org/elasticsearch/plugins/UberModuleClassLoader.java
```diff
@@ -0,0 +1,280 @@
...
+        Set<String> moduleDenyList,
+        Set<String> packageNames
+    ) {
+        super(parent);
+
+        this.internalLoader = new URLClassLoader(jarURLs);
+        // code source is always the first jar on the list
+        this.codeSource = new CodeSource(jarURLs[0], (CodeSigner[]) null);
+        // we need a module layer to bind our module to this classloader
+        this.moduleController = ModuleLayer.defineModules(cf, List.of(mparent), s -> this);
```
pgomulka, line 130: I wonder if it would be helpful to have a more detailed comment how this binding works. I think it might not be obvious at first and I feel like this is the crucial bit?

williamrandolph: Sounds good. I'll add that comment before merging.

williamrandolph:

williamrandolph:
@elasticmachine update branch

PR #89370, opened 2022-08-16
Add YAML spec docs about matching errors

DaveCTurner:
It's not obvious that a YAML test with a `catch` stanza also permits
`match` blocks to assert things about the structure of the error
response, but this structure may be an important part of the API spec.
This commit adds this info to the docs about YAML tests.


rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/README.asciidoc
```diff
@@ -320,8 +320,13 @@ be caught and tested. For instance:
         catch:        missing
         get:
             index:    test
-            type:    test
+            type:     test
             id:        1
+
```
nik9000, line 325: Maybe,
```
# And, optionally, you can assert on the contents of the precise contents of the error message.
```

nik9000 (approved):

PR #89376, opened 2022-08-16
[ML] frequent_items should not be a release highlight for 8.4

droberts195:
The frequent_items aggregation is experimental in 8.4, and we
want to gain confidence in it for internal use cases before
publicising more widely, so we'll just give it a single bullet
point release note rather than a whole section in the release
highlights.


dimitris-athanasiou (approved):
LGTM

PR #89078, opened 2022-08-03
[8.4][DOCS] Adds ML-related release highlights

szabosteve:
## Overview

This PR adds the `highlight` property with title and some description to the YAML files in the changelog that contain ML-related changes that need to be highlighted.

It also adds the item about "Early stopping for data frame analytics" to the YAML file of the "aggregations in datafeeds are GA". As the early stopping functionality has been added to the stack via a PR in the ML-CPP repo, there is no corresponding YAML file in the elasticsearch repo. This situation and the nature of the highlight generation process necessitate this less elegant way to add the highlight item to the "What's new" section.

PR #89377, opened 2022-08-16
[ML] Sync changelogs between 8.4 and main

droberts195:
Copies a few changelog edits from the 8.4 branch into main so that
people looking at the main branch see the same docs as people
looking at 8.4.

Relates #89078
Relates #89286 
Relates #89376


dimitris-athanasiou (approved):
LGTM

PR #89378, opened 2022-08-16
Small cleanups to Allocation Performance

original-brownbear:
Two fixes:
1. Looking up `Custom` values over and over for every shard incurs a measurable cost.
This removes that cost for desired nodes and node shutdown metadata.
2. Node shutdown metadata logic wasn't inlining nicely because of the wrapped map.
No need to be as complicated as we were in many spots, use a simple immutable map
for all operations and remove a bunch of branching.

relates #77466

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
+++ b/server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
@@ -1024,9 +1024,7 @@ public class Metadata extends AbstractCollection<IndexMetadata> implements Diffa
     }
 
     public Map<String, SingleNodeShutdownMetadata> nodeShutdowns() {
-        return Optional.ofNullable((NodesShutdownMetadata) this.custom(NodesShutdownMetadata.TYPE))
-            .map(NodesShutdownMetadata::getAllNodeMetadataMap)
-            .orElse(Collections.emptyMap());
+        return this.custom(NodesShutdownMetadata.TYPE, NodesShutdownMetadata.EMPTY).getAllNodeMetadataMap();
     }
 
     public Map<String, Custom> customs() {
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/elasticsearch/cluster/metadata/Metadata.java
```diff
@@ -1024,9 +1024,7 @@ public Map<String, DataStreamAlias> dataStreamAliases() {
     }
 
     public Map<String, SingleNodeShutdownMetadata> nodeShutdowns() {
-        return Optional.ofNullable((NodesShutdownMetadata) this.custom(NodesShutdownMetadata.TYPE))
-            .map(NodesShutdownMetadata::getAllNodeMetadataMap)
-            .orElse(Collections.emptyMap());
+        return this.custom(NodesShutdownMetadata.TYPE, NodesShutdownMetadata.EMPTY).getAllNodeMetadataMap();
```
idegtiarenko, line 1027: This is way simpler :+1:

server/src/main/java/org/elasticsearch/cluster/routing/allocation/RoutingAllocation.java
```diff
@@ -73,6 +72,11 @@ public class RoutingAllocation {
 
     private final Map<String, SingleNodeShutdownMetadata> nodeReplacementTargets;
 
+    private final Map<String, SingleNodeShutdownMetadata> nodeShutdowns;
+
+    @Nullable
+    private final DesiredNodes desiredNodes;
```
idegtiarenko, line 78: Do we need above to make it inline?

original-brownbear: This isn't really about inlining. This is just faster outright because we don't have to lookup from the customs map for every shard.

idegtiarenko (approved):

original-brownbear:
Thanks Ievgen!

PR #89380, opened 2022-08-16
Remove deprecated code from `ApiKeyService`

n1v0lg:
Small refactor to remove use of deprecated functionality in
`ApiKeyService`.


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -1344,16 +1344,10 @@ private <T> void findApiKeysForUserRealmApiKeyIdAndNameCombination(
     /**
      * Performs the actual invalidation of a collection of api keys
      *
-     * @param apiKeyIds       the api keys to invalidate
-     * @param listener        the listener to notify upon completion
-     * @param previousResult  if this not the initial attempt for invalidation, it contains the result of invalidating
```
n1v0lg, line 1349: `previousResult` was always `null` so simplifying the code here

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -1744,23 +1733,34 @@ AtomicLong getLastEvictionCheckedAt() {
      */
     public static String getCreatorRealmName(final Authentication authentication) {
         if (authentication.isApiKey()) {
-            return (String) authentication.getMetadata().get(AuthenticationField.API_KEY_CREATOR_REALM_NAME);
+            return (String) authentication.getAuthenticatingSubject().getMetadata().get(AuthenticationField.API_KEY_CREATOR_REALM_NAME);
         } else {
             return authentication.getSourceRealm().getName();
```
n1v0lg, line 1738: I wanted to replace this as well, but `getCreatorRealmName` gets used in audit logging and working out the ramifications of either throwing/returning `null` got a little hairy so I'm leaving this for another day.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
ywangd: I think this pattern was copied from TokenService. But it is not useful here.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -1744,23 +1733,34 @@ AtomicLong getLastEvictionCheckedAt() {
      */
     public static String getCreatorRealmName(final Authentication authentication) {
         if (authentication.isApiKey()) {
-            return (String) authentication.getMetadata().get(AuthenticationField.API_KEY_CREATOR_REALM_NAME);
+            return (String) authentication.getAuthenticatingSubject().getMetadata().get(AuthenticationField.API_KEY_CREATOR_REALM_NAME);
```
ywangd, line 1736: I think we should consistently use `getEffectiveSubject()` here as well. In practice, both `getEffectiveSubject` and `getAuthenticatingSubject` return the same object because API key cannot be the target of run-as. But using `getAuthenticatingSubject` is assuming that knowledge here which isn't great and can in theory become incorrect if we start supporting run-as an API Key. In general, code at service layer should mostly not have to worry about authenticatingSubject unless there is a strong reason for it.

I am aware that using `getAuthenticatingSubject` is more faithful to what the existing code does. But in this case, I think the existing code isn't quite right either.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
ywangd: Perhaps useful to leave a comment here unless you plan to work on it right away.

ywangd (approved):
LGTM

I'd appreciate if you could address the comment about `getAuthenticatingSubject` vs `getEffectiveSubject`. I commented it on one place. But it applies to a few other places as well.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
n1v0lg: I'm good with using `getEffectiveSubject` here. I went with `getAuthenticatingSubject` because it was suggested by the deprecation Javadoc for [`getMetadata`](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/authc/Authentication.java#L226).

n1v0lg:

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ApiKeyService.java
```diff
@@ -338,14 +337,20 @@ private void createApiKeyAndIndexIt(
...
+                                listener.onFailure(indexResponse.getFailure().getCause());
+                            } else {
+                                assert request.getId().equals(indexResponse.getResponse().getId());
+                                assert indexResponse.getResponse().getResult() == DocWriteResponse.Result.CREATED;
+                                final ListenableFuture<CachedApiKeyHashResult> listenableFuture = new ListenableFuture<>();
+                                listenableFuture.onResponse(new CachedApiKeyHashResult(true, apiKey));
+                                apiKeyAuthCache.put(request.getId(), listenableFuture);
+                                listener.onResponse(new CreateApiKeyResponse(request.getName(), request.getId(), apiKey, expiration));
+                            }
+                        }, listener::onFailure)
```
ywangd, line 353: I meant to comment on this but somehow forgot. It is for future refactoring: This `wrapBulkResponse` method is used in a few other places. It might be helpful to move this method to a different non-deprecated class (TransportBulkAction?) so that we can keep reusing it instead of having to repeat the same logic in other palces?

n1v0lg: Absolutely, was thinking this too. I'll open a separate PR with a util method. Is this best reviewed by the distributed team or core or another team I'm not thinking of?

ywangd: I think technically it should be the distributed team.

PR #89389, opened 2022-08-16
Assign the right path to objects merged when parsing mappings

### LINKED ISSUE #88573: mapper_parsing_exception (Field is defined more than once) in 8.3.2
### Elasticsearch Version

8.3.2

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

Ubuntu 20.04.4 LTS

### Problem Description

When we have two legacy index templates and one of them defines the field name with dots (.), we may get mapper_parsing_exception.

### Steps to Reproduce

1. create two legacy index templates
```
PUT /_template/template1
{
  "order": 100,
  "index_patterns": [
    "haha*"
  ],
  "settings": {},
  "mappings": {
    "properties": {
      "kwm": {
        "properties": {
          "source": {
            "properties": {
              "geo": {
                "properties": {
                  "location": {
                    "type": "geo_point"
                  }
                }
              }
            }
          }
        }
      },
      "source": {
        "properties": {
          "geo": {
            "properties": {
              "location": {
                "type": "geo_point"
              }
            }
          }
        }
      }
    }
  }
}

PUT /_template/template2
{
  "order": 10,
  "index_patterns": [
    "haha*"
  ],
  "settings": {},
  "mappings": {
    "properties": {
      "kwm.source.geo": {
        "properties": {
          "location": {
            "type": "geo_point"
          }
        }
      }
    }
  }
}
```
3. try to create an index
```
POST haha/_doc
{
}
```
it produces `mapper_parsing_exception` in `8.3.2` as below
```
{
  "error": {
    "root_cause": [
      {
        "type": "mapper_parsing_exception",
        "reason": "Field [source.geo.location] is defined more than once"
      }
    ],
    "type": "mapper_parsing_exception",
    "reason": "Field [source.geo.location] is defined more than once"
  },
  "status": 400
}
```
but no error in `8.2.3`

4. if we change the second index template `template2` as below:
```
PUT /_template/template2
{
  "order": 10,
  "index_patterns": [
    "haha*"
  ],
  "settings": {},
  "mappings": {
    "properties": {
      "kwm": {
        "properties": {
          "source": {
            "properties": {
              "geo": {
                "properties": {
                  "location": {
                    "type": "geo_point"
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```
we can create the `haha` index without errors in both `8.3.2` and `8.2.3`

I think we should not get `mapper_parsing_exception` in both `8.3.2` and `8.2.3` in the step 3 because `kwm.source.geo.location` field definition of two index templates are identical and `source.geo.location` is defined only in the `template1` index template.
I think this may be related with https://github.com/elastic/elasticsearch/pull/86166

### Logs (if relevant)

_No response_

javanna:
When parsing mappings, we may find a field with same name specified twice, either
because JSON duplicate keys are allowed, or because a mix of object notation and dot
notation is used when providing mappings. The same can happen when applying dynamic
mappings as part of parsing an incoming document, as well as when merging separate
index templates that may contain the definition for the same field using a
mix of object notation and dot notation.

While we propagate the MapperBuilderContext across merge calls thanks to #86946, we do not
propagate the right context when we call merge on objects as part of parsing/building
mappings. This causes a situation in which the leaf fields that result from the merge
have the wrong path, which misses the first portion e.g. sub.field instead of obj.sub.field.

This commit applies the correct mapper builder context when building the object mapper builder
and two objects with same name are found.

The scenario in which this bug was triggered is limited: the same document or mappings need to have the same field specified twice, using the object notation as well as the dot notation. In this case, we would end up indexing the leaf fields that were merged (the leaves that both object held) in the wrong lucene fields, which can only be fixed through a reindex. Closing and reopening the index triggers a reparsing of the mappings that addressed the problem though.

Relates to #86946

Closes #88573

### DIFF
--- a/docs/changelog/89389.yaml
+++ b/docs/changelog/89389.yaml
@@ -1,0 +1,6 @@
+pr: 89389
+summary: Assign the right path to objects merged when parsing mappings
+area: Mapping
+type: bug
+issues:
+ - 88573
--- a/server/src/internalClusterTest/java/org/elasticsearch/indices/template/SimpleIndexTemplateIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/indices/template/SimpleIndexTemplateIT.java
@@ -14,6 +14,8 @@ import org.elasticsearch.action.admin.indices.settings.get.GetSettingsResponse;
 import org.elasticsearch.action.admin.indices.template.get.GetIndexTemplatesResponse;
 import org.elasticsearch.action.admin.indices.template.put.PutIndexTemplateRequestBuilder;
 import org.elasticsearch.action.bulk.BulkResponse;
+import org.elasticsearch.action.fieldcaps.FieldCapabilities;
+import org.elasticsearch.action.fieldcaps.FieldCapabilitiesResponse;
 import org.elasticsearch.action.index.IndexRequest;
 import org.elasticsearch.action.search.SearchResponse;
 import org.elasticsearch.cluster.ClusterState;
@@ -39,6 +41,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import static org.elasticsearch.action.support.WriteRequest.RefreshPolicy.IMMEDIATE;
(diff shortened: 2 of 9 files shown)


romseygeek (approved):
LGTM

PR #89390, opened 2022-08-16
Remove root argument from buildMappers method

javanna:
The callers of buildMappers can provide the right context, instead of passing a boolean
argument that controls what context is used.


javanna:
run elasticsearch-ci/part-1

romseygeek (approved):
LGTM, nice cleanup

PR #89391, opened 2022-08-16
Correct ctx.op value to valid 'noop'

adammichalik:
The documentation mentions the `noop` value ("Otherwise it does nothing (`noop`)"), however, the value used in the example script is `none`. This change corrects the value in the example script to `noop`.


stu-elastic (approved):

stu-elastic:
Thanks for the contribution, @adammichalik!

PR #89393, opened 2022-08-16
[DOCS] Update add node section

### LINKED ISSUE #86334: [Docs] "Add and remove nodes" docs should link to security "Enroll additional nodes" docs
### Description

In the **Set up Elasticsearch** section of the docs, there is a page for [Add and remove nodes in your cluster](https://www.elastic.co/guide/en/elasticsearch/reference/current/add-elasticsearch-nodes.html).

It has these instructions:
> 1. Set up a new Elasticsearch instance.
> 2. Specify the name of the cluster with the cluster.name setting in elasticsearch.yml. For example, to add a node to the logging-prod cluster, add the line cluster.name: "logging-prod" to elasticsearch.yml.
> 3. Start Elasticsearch. The node automatically discovers and joins the specified cluster.

However, if security (including TLS between nodes) is enabled (which it is by default) those instructions will not work, because each node will generate its own set of certificates.

Those steps should include some sort of link to the relevant security docs to [Enroll additional nodes in your cluster](https://www.elastic.co/guide/en/elasticsearch/reference/current/configuring-stack-security.html#stack-enroll-nodes)

lockewritesdocs:
Updates the verbiage in _Add and remove nodes in your cluster_ to reference the section for enrolling additional nodes in your cluster.

Preview link: https://elasticsearch_89393.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/add-elasticsearch-nodes.html

Closes #86334

### DIFF
--- a/docs/reference/setup/add-nodes.asciidoc
+++ b/docs/reference/setup/add-nodes.asciidoc
@@ -25,16 +25,21 @@ green.
 
 image::setup/images/elas_0204.png["A cluster with three nodes"]
 
-You can
-{ref}/configuring-stack-security.html#stack-enroll-nodes[enroll additional nodes]
-on your local
-machine to experiment with how an {es} cluster with multiple nodes behaves. 
+[discrete]
+=== Enroll nodes in an existing cluster
+You can enroll additional nodes on your local machine to experiment with how an
+{es} cluster with multiple nodes behaves. 
 
-To add a node to a cluster running on multiple machines, you must also
-<<unicast.hosts,set `discovery.seed_hosts`>> so that the new node can discover
+[NOTE]
+====
+To add a node to a cluster running on multiple machines, you must also set
+<<unicast.hosts,`discovery.seed_hosts`>> so that the new node can discover
 the rest of its cluster.
+====
 
-For more information about discovery and shard allocation, see
+include::../../../x-pack/docs/en/security/enroll-nodes.asciidoc[]
+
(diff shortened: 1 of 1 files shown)


ywangd (commented):
I wonder whether it would be better to embed the content of the "enroll-nodes" page instead of linking to it. This page is already titled as "Add and remove nodes in your cluster". It feels awkward that it has to link to another page that has almost the same (sub) title. 

Related, but not directly for this PR, is that I noticed the same "enroll-nodes" content is embedded twice in the [Starting Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/starting-elasticsearch.html#starting-elasticsearch) page ([first](https://www.elastic.co/guide/en/elasticsearch/reference/current/starting-elasticsearch.html#_enroll_nodes_in_an_existing_cluster_3) and [second](https://www.elastic.co/guide/en/elasticsearch/reference/current/starting-elasticsearch.html#_enroll_nodes_in_an_existing_cluster_4)). I think linking is actually a better option here since (1) the page is mostly about starting up not additional nodes and (2) avoid repetition.

lockewritesdocs:
> I wonder whether it would be better to embed the content of the "enroll-nodes" page instead of linking to it. This page is already titled as "Add and remove nodes in your cluster".

I updated the page to include this content directly, rather than linking to it.

ywangd (approved):
LGTM

PR #89394, opened 2022-08-16
Issue in the code - geo_point field with GeoJSON type wouldn't work!

madhusudhankonda:

### DIFF
--- a/docs/reference/mapping/types/geo-point.asciidoc
+++ b/docs/reference/mapping/types/geo-point.asciidoc
@@ -32,6 +32,8 @@ PUT my-index-000001
   }
 }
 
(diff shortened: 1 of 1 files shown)


jtibshirani:
@madhusudhankonda thanks for the feedback. It sounds like you've found a bug where GeoJSON is not support for `geo_point` fields. If this is the case, could you please file a GitHub issue with a simple reproduction (including the Elasticsearch version, etc.)?

craigtaverner:
With no answer to @jtibshirani questions to the author, we will close this PR. My understanding is that GeoJSON support for geo_point was added in 8.2.0, so it could be that the author tested an older version.

PR #89395, opened 2022-08-16
Limit concurrent msearch with preference in expand fetch phase

dnhatn:
Today, a multi-search request from the expand-fetch-phase of a search request with `preference` can take over the search thread pool of some data nodes. This happens because we route all expanded requests to the same set of nodes (due to the inherited `preference` parameter), and the multi_search_action doesn't limit the number of concurrent search requests properly. I see two ways to alleviate this issue:
- Ensure that the number of shard requests doesn't exceed the max allowed shard requests on each node. This PR implements this strategy.
- Do not pass the preference to the expanded requests meaning routing them to as many nodes as possible.

### DIFF
--- a/server/src/main/java/org/elasticsearch/action/search/ExpandSearchPhase.java
+++ b/server/src/main/java/org/elasticsearch/action/search/ExpandSearchPhase.java
@@ -10,18 +10,23 @@ package org.elasticsearch.action.search;
 
 import org.elasticsearch.action.ActionListener;
 import org.elasticsearch.common.util.Maps;
+import org.elasticsearch.common.util.concurrent.AtomicArray;
 import org.elasticsearch.index.query.BoolQueryBuilder;
 import org.elasticsearch.index.query.InnerHitBuilder;
 import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.index.query.QueryBuilders;
 import org.elasticsearch.search.SearchHit;
 import org.elasticsearch.search.SearchHits;
+import org.elasticsearch.search.SearchPhaseResult;
 import org.elasticsearch.search.builder.SearchSourceBuilder;
 import org.elasticsearch.search.collapse.CollapseBuilder;
 import org.elasticsearch.search.internal.InternalSearchResponse;
+import org.elasticsearch.threadpool.ThreadPool;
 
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/elasticsearch/action/search/ExpandSearchPhase.java
```diff
@@ -60,6 +72,9 @@ public void run() {
             MultiSearchRequest multiRequest = new MultiSearchRequest();
             if (collapseBuilder.getMaxConcurrentGroupRequests() > 0) {
                 multiRequest.maxConcurrentSearchRequests(collapseBuilder.getMaxConcurrentGroupRequests());
+            } else if (context.getRequest().preference() != null) {
```
dnhatn, line 75: Do we need to preserve the preference parameter for the expanded search requests? Can we discard it and route the expanded requests to all replicas?

ywelsch, line 75: I'm not sure how much "preference" contributes to being the problem. If you have the typical setup with number_of_replicas:1, the same data-node level burst of requests is likely to occur even if you are not using preference here.

I'm wondering if we should always bound this in some way when doing internal multi-search requests, or whether we should seek at addressing this at the multi-search request handling level (e.g. some sensible default for maxConcurrentSearchRequests).

dnhatn: > I'm not sure how much "preference" contributes to being the problem.

The customer reported this issue has 14 replicas for their target indies.

> I'm wondering if we should always bound this in some way when doing internal multi-search requests

+1

> whether we should seek at addressing this at the multi-search request handling level (e.g. some sensible default for maxConcurrentSearchRequests).

I think the default maxConcurrentSearchRequests in TransportMultiSearchAction is too large https://github.com/elastic/elasticsearch/blob/cce5ad2e4fdfe20fdce0972a1e20a9c4a722a75a/server/src/main/java/org/elasticsearch/action/search/TransportMultiSearchAction.java#L103-L113.

dnhatn:
I am closing this PR as we prefer having a small and fixed number of concurrent msearch in the expand fetch phase.

Issue #81246, opened 2021-12-02
Scripts to expose whole values for fields of the `text` family


jpountz:
One issue I keep hearing about is that it's too hard to define a runtime field that extracts some information from a `message` field with Painless. Something like extracting the HTTP status code from a log line of an Apache access log.

I think that this issue has been put into the general meta issue of "doing simple things with Painless should be simpler" but in my opinion this particular issue has more to do with mappings than with Painless. Historically, fielddata on analyzed `string` fields would uninvert the inverted index in memory and Elasticsearch would consider that the value of a field is the set of analyzed terms that it contains. This would require lots of memory, and over time we've increasingly discouraged users from doing it.

These semantics don't work well with runtime extraction of data. If you try to extract data using a regular expression that applies to `doc['message']`, you'll get an exception that fielddata is disabled by default on `text` fields. And even if Elasticsearch returned values, you'd get individual terms, which you cannot leverage to properly extract data from the message.

I suggest that we change the semantics of fielddata on fields of the `text` family (including `text` and `match_only_text`) so that it returns whole values instead. This will enable us to give a more intuitive experience with scripts, where `doc` could read data from `_source` on `text` fields (#80504).

Note that this brings a downside: in order to make it easy to slice and dice the data, Elasticsearch allows users to use terms produce by `terms` aggregations in `term` filters, in order to dig further data that falls within a given bucket. This would not work on `text` fields. I don't think it's the end of the world, since `terms` aggregations do not work on `text` fields today anyway given that we disallow fielddata, but I wanted to highlight it since it would create an exception to a rule that is otherwise honored by `keyword`, `ip` or numeric fields.

romseygeek:
+1. Combined with an 'exact' query, this would also allow us to do away with the keyword + text multifield we generate by default for text input, replacing it with a fielddata-enabled text field.

markharwood:
+1

>Note that this brings a downside ... Elasticsearch allows users to use terms produce by terms aggregations in term filters, in order to dig further data that falls within a given bucket. This would not work on text fields.

The pattern for "discover then drill-down" on text fields is to use `significant_text` agg and then a `term` query.
I use this daily to see trending terms/shingles in news headlines and then click to see details:


<img width="390" alt="Kibana" src="https://user-images.githubusercontent.com/170925/144406177-41040bbc-ebf7-4469-b85f-03e92889624c.png">

The change proposed on this issue shouldn't effect that technique.

jdconrad:
[...]

PR #89396, opened 2022-08-16
Add text field support in the Painless scripting fields API

jdconrad:
This change adds access to mapped text fields via the Painless scripting fields API. The values returned from a text field via the scripting fields API always use source as described by (https://github.com/elastic/elasticsearch/issues/81246). Access via the old-style through doc will still depend on field data, so there is no change and avoids bwc issues.

### DIFF
--- a/docs/changelog/89396.yaml
+++ b/docs/changelog/89396.yaml
@@ -1,0 +1,5 @@
+pr: 89396
+summary: Add text field support in the Painless scripting fields API
+area: Mapping
+type: enhancement
+issues: []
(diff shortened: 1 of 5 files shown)


jdconrad:

javanna (approved):
LGTM I think that this change closes effectively the issue you linked. Thanks for working on this, it will make accessing text fields from a script much easier!

romseygeek (approved):
LGTM too!

jdconrad:
@javanna @romseygeek Thank you both for the review!

PR #89397, opened 2022-08-16
Preserve `system_index_migration` cluster state when loading metadata from disk

### LINKED ISSUE #88983: Named XContent `system_index_migration` is not registered
I see several clusters (versions 7.15-8.3) reporting `Skipping unknown custom object with type system_index_migration` which indicates they encountered a `Metadata.Custom` called `system_index_migration` while loading the metadata from disk, but this name is not a registered `NamedXContent` so it is ignored[^1]. If we don't want this custom to persist across restarts then it should be a `ClusterState.Custom` instead, and if we do want it to persist across restarts then it must be registered.

[^1] The lenience in this area is a bug, see #21471, but we can't make it stricter yet because of this issue.

williamrandolph:
We used a Metadata.Custom object to track the results of feature migration, but neglected to register it in the NamedXContentRegistry. This violates the expectations for `Metadata.Custom` and causes this metadata to be ignored when loading from disk.

Fixes #88983

### DIFF
--- a/docs/changelog/89397.yaml
+++ b/docs/changelog/89397.yaml
@@ -1,0 +1,7 @@
+pr: 89397
+summary: Preserve `system_index_migration` cluster state when loading metadata from
+  disk
+area: Infra/Core
+type: bug
+issues:
+ - 88983
--- a/server/src/main/java/org/elasticsearch/upgrades/SystemIndexMigrationExecutor.java
+++ b/server/src/main/java/org/elasticsearch/upgrades/SystemIndexMigrationExecutor.java
@@ -122,6 +122,11 @@ public class SystemIndexMigrationExecutor extends PersistentTasksExecutor<System
                 PersistentTaskState.class,
                 new ParseField(SystemIndexMigrationTaskParams.SYSTEM_INDEX_UPGRADE_TASK_NAME),
                 SystemIndexMigrationTaskState::fromXContent
+            ),
+            new NamedXContentRegistry.Entry(
+                FeatureMigrationResults.class,
+                new ParseField(FeatureMigrationResults.TYPE),
+                FeatureMigrationResults::fromXContent
             )
         );
     }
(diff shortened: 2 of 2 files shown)


grcevski (commented):
Looks good! Thanks for fixing this. Can we please add some serialization tests to prove we can read the information correctly?

DaveCTurner:
+1 to this deserving a test, although I'd prefer to see one derived from `ESIntegTestCase` that verifies this metadata really does persist across a full cluster restart.

williamrandolph:
@grcevski We already have a serialization test (`FeatureMigrationResultsTests`), so I think @DaveCTurner 's suggestion is what we need for this PR to be merge-able.

williamrandolph:
Closing this dangling PR until I can figure out an `ESIntegTestCase` that will check it.

DaveCTurner:
I would have guessed that something like this would be enough:

```diff
diff --git a/modules/reindex/src/internalClusterTest/java/org/elasticsearch/migration/FeatureMigrationIT.java b/modules/reindex/src/internalClusterTest/java/org/elasticsearch/migration/FeatureMigrationIT.java
index 56db671fec5..17eebeb5aae 100644
--- a/modules/reindex/src/internalClusterTest/java/org/elasticsearch/migration/FeatureMigrationIT.java
+++ b/modules/reindex/src/internalClusterTest/java/org/elasticsearch/migration/FeatureMigrationIT.java
@@ -301,6 +301,10 @@ public class FeatureMigrationIT extends AbstractFeatureMigrationIntegTest {
             fail("cluster state update failed, see log for details");
         }

+        if (randomBoolean()) {
+            internalCluster().fullRestart();
+        }
+
         PostFeatureUpgradeRequest migrationRequest = new PostFeatureUpgradeRequest();
         PostFeatureUpgradeResponse migrationResponse = client().execute(PostFeatureUpgradeAction.INSTANCE, migrationRequest).get();
         // Make sure we actually started the migration
```

However this entire test suite seems to be skipped at the moment because of an `assumeTrue` in the `setup()`.

williamrandolph:
@DaveCTurner I didn't know about that `fullRestart()` method.

I added that `assumeTrue` method #93666 because we wanted to make the system feature migration endpoint a no-op, and did this by lowering the "migration needed" threshold from 8.0 to 7.0. Once we start developing and testing 9.x migrations, this suite should start firing again.

Maybe I can rework the way we do this in such a way that the tests still run. I'll see if I can come up with something.

PR #89401, opened 2022-08-17
Introduce TriangleTreeReader.DecodedVisitor

iverase:
This PR introduces a DecodedVisitor that removes duplicate code for visitors that need to decode coordinates.


craigtaverner (approved):
LGTM. I see the reduced duplication between Component2DVisitor and LabelPositionVisitor. I would probably have used different method names than `do*`, perhaps `visit*Decoded`, or even keep the names and rely on parameter types for method uniqueness. But `do*` is fine too.

PR #89415, opened 2022-08-17
GeoShapeValue can determine the spatial relationship with a LatLonGeometry

iverase:
This PR adds a new method to GeoShape value called `GeoRelation relate(LatLonGeometry latLonGeometry)` that replaces `boolean intersects(Geometry geometry)`.  The new method returns the spatial relationship and it can be used to navigate grids that are composed of geometries different to bounding boxes.

A new visitor is added called LatLonGeometryRelationVisitor which works in a similar way to Tile2DVisitor.

### DIFF
--- a/x-pack/plugin/spatial/src/test/java/org/elasticsearch/xpack/spatial/index/fielddata/LatLonGeometryRelationVisitorTests.java
+++ b/x-pack/plugin/spatial/src/test/java/org/elasticsearch/xpack/spatial/index/fielddata/LatLonGeometryRelationVisitorTests.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License
+ * 2.0; you may not use this file except in compliance with the Elastic License
+ * 2.0.
+ */
+
+package org.elasticsearch.xpack.spatial.index.fielddata;
+
+import org.apache.lucene.document.ShapeField;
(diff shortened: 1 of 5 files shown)


x-pack/plugin/spatial/src/test/java/org/elasticsearch/xpack/spatial/index/fielddata/LatLonGeometryRelationVisitorTests.java
```diff
@@ -0,0 +1,78 @@
...
+        doTestShapes(GeoTestUtil::nextPolygon);
+    }
+
+    private void doTestShapes(Supplier<LatLonGeometry> supplier) throws Exception {
+        Geometry geometry = GeometryTestUtils.randomGeometryWithoutCircle(0, false);
+        GeoShapeValues.GeoShapeValue geoShapeValue = GeoTestUtils.geoShapeValue(geometry);
+        GeometryDocValueReader reader = GeoTestUtils.geometryDocValueReader(geometry, CoordinateEncoder.GEO);
+        for (int i = 0; i < 1000; i++) {
+            LatLonGeometry latLonGeometry = supplier.get();
+            GeoRelation relation = geoShapeValue.relate(latLonGeometry);
```
craigtaverner, line 43: It took me a moment to realize that you are testing the new relates visitor against the existing specific visitors. Seems like a valid approach to testing this.

craigtaverner (approved):
LGTM. I will deal with any conflicts with cartesian work later.

PR #89386, opened 2022-08-16
[ML] Get categories endpoint to use ECS Grok patterns

edsavage:
Change the Grok pattern creator for _ml/anomaly_detectors/<job_id>/results/categories to always use ECS Grok patterns

relates #77065

Issue #77065, opened 2021-08-31
[ML] Switch to ECS Grok patterns in text structure finder and categorization


droberts195:
#76885 introduced the possibility of using ECS Grok patterns instead of the legacy ones.

We should switch to using these in the text structure plugin and for the Grok patterns we add to categorization results.

It looks like some new date formats exist in the latest set of Grok patterns - certainly for newer versions of Tomcat and Catalina, possibly others - we should add those to the timestamp format finder too.

- [x] Add `ecs_compatibility` option to `_text_structure/find_structure` endpoint, default `disabled`, and change that endpoint to use ECS Grok patterns if it's set to `v1`. This _may_ also necessitate making the timestamp format finder aware of two different Grok patterns per timestamp format, and then having it use the appropriate one depending on whether ECS Grok patterns are in use (investigation required).
- [x] Change UI to set `ecs_compatibility` to `v1` when calling `_text_structure/find_structure`. https://github.com/elastic/kibana/issues/138428
- [x] Have a look through the ECS Grok patterns that were added in #76885 and see if there are any new timestamp formats that didn't exist in the original Grok patterns. Maybe Tomcat and Catalina have some new ones, maybe others. If any are found add configs for them to the timestamp format finder in `_text_structure/find_structure`.
- [x] Change the Grok pattern creator for `_ml/anomaly_detectors/<job_id>/results/categories` to always use ECS Grok patterns - this change can be made unconditionally without keeping a BWC option for the old Grok patterns, as the functionality is experimental. #89386

droberts195:
All tasks complete now - closing

PR #89424, opened 2022-08-17
[ML] Performance improvements related to ECS Grok pattern usage

edsavage:
Swap out `TOMCAT_DATESTAMP` for `TOMCATLEGACY_DATESTAMP` when ECS compatibility is set to `v1`

See comments on #89386

Relates to #77065


droberts195 (approved):
LGTM

droberts195:
I think this can be a non-issue as the original ECS Grok patterns PR isn’t released yet.

PR #89427, opened 2022-08-17
Upgrade to lucene-9.4.0-snapshot-923a9f800ae

mayya-sharipova:
Upgrade ES to a new Lucene snapshot:
Changes of interest:
- LUCENE-10592 Build HNSW Graph on indexing
- LUCENE-10678: Fix potential overflow when computing the partition point
- LUCENE-10633: Dynamic pruning for sorting on SORTED(_SET) fields


mayya-sharipova:
@jpountz I wanted to clarify with you what is our plan for integrating LUCENE-10633 (Dynamic pruning for sorting on SORTED(_SET) fields) to ES? 
Several points to consider:
- I've disabled sort optimization in `BytesRefFieldComparatorSource.java` for custom sort fields.
- But for standard sort options optimization will be enabled (hence I had to update `AbstractStringFieldDataTestCase.java` to ask to collect all hits (otherwise `SortedSetDVStringFieldDataTests` tests were failing). Is this what we wanted for ES users?
- I've noticed that for `SortedSetSortField` there is no way to completely disable sort optimization (as there is for `SortField` with `setOptimizeSortWithPoints(false)`? May be we should introduce in Lucene an option for this?

jpountz:
To be honest, I had not thought much of how the integration would work, I had - probably a bit naively - assumed that it would work in a similar way to how we integrated dynamic pruning for numeric/date sorts.

Disabling pruning first as part of the Lucene upgrade and enabling it as part of a separate pull request sounds like a good approach to me. Changes to `AbstractStringFieldDataTestCase` make sense to me, I guess another approach would have been to remove the assertion that verifies that the total number of collected hits is equal to the number of docs in the index, which I guess is the assertion that failed?

> May be we should introduce in Lucene an option for this as well?

Oh, good point I had missed this. +1 to add the option to disable pruning on `SortedSetSortField` too.

mayya-sharipova:

mayya-sharipova:
@jpountz Thanks for the initial review.

> Changes to AbstractStringFieldDataTestCase make sense to me, I guess another approach would have been to remove the assertion that verifies that the total number of collected hits is equal to the number of docs in the index, which I guess is the assertion that failed?

Addressed in 2bb83bbe3dad2d2c5c7d5bf6d39a4b941e8a1317

Please continue your review whenever you have time.

---

Meanwhile I have also opened a [Lucene PR](https://github.com/apache/lucene/pull/3125) to add ability for SortedSetSortField to disable sort optimization completely.

gradle/verification-metadata.xml
```diff
@@ -2436,121 +2436,241 @@
...
+      <component group="org.apache.lucene" name="lucene-analysis-common" version="9.4.0-snapshot-923a9f800ae">
+         <artifact name="lucene-analysis-common-9.4.0-snapshot-923a9f800ae.jar">
+            <sha256 value="163c5002f4a4115af1d83c785dac61f71d2173becaf0208fd98b110222a003d5" origin="Generated by Gradle"/>
+         </artifact>
+      </component>
       <component group="org.apache.lucene" name="lucene-analysis-icu" version="9.3.0">
          <artifact name="lucene-analysis-icu-9.3.0.jar">
             <sha256 value="e3af1f650877bcc438e15ea2b12dfac4d3ac12c27105cd4b000d0916730f6a74" origin="Generated by Gradle"/>
          </artifact>
       </component>
```
jpountz, line 2448: I'm not familiar with this file, is it expected that it still lists 9.3.0 artifacts?

jpountz (approved):

gradle/verification-metadata.xml
mayya-sharipova: @jpountz Thanks a good comment. I don't know. I will ping ES delivery team to confirm.
These changes are autogenerated by the command `./gradlew --write-verification-metadata sha256 precommit` that we are using instead of `./gradlew updateShas` now.

mayya-sharipova: @elastic/es-delivery We found it surprising that `verification-metadata.xml` file contains both artifacts for lucene-9.3.0 and the upgrade we are working on lucene-9.4.0.  These changes are autogenerated by the command `./gradlew --write-verification-metadata sha256 precommit` that we are asked to used.

We want to confirm with you that they are expected?

mark-vieira: I believe that's because `--wirte-verification-metadata` simply _appends_ to the existing file. It does not `cull` older entries. We may decide to cleanup older dependency entries at a later time.

mayya-sharipova: @mark-vieira Thanks for confirmation. For now we will proceed with this PR as it is. 
I am not sure who is responsible for cleaning older dependencies. Please let us know if our team should do at a later time.

PR #89430, opened 2022-08-17
Inactive shard flush should wait for ongoing one

### LINKED ISSUE #87888: [CI] FlushIT testFlushOnInactive failing
Actually failed on a PR branch but the PR didn't touch anything in this area I think.

**Build scan:**
https://gradle-enterprise.elastic.co/s/oof57yigaovuk/tests/:server:internalClusterTest/org.elasticsearch.indices.flush.FlushIT/testFlushOnInactive

**Reproduction line:**
`./gradlew ':server:internalClusterTest' --tests "org.elasticsearch.indices.flush.FlushIT.testFlushOnInactive" -Dtests.seed=38FC56033AE2DC88 -Dtests.locale=en-IN -Dtests.timezone=Europe/Zaporozhye -Druntime.java=17`

**Applicable branches:**
master

**Reproduces locally?:**
Didn't try

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.indices.flush.FlushIT&tests.test=testFlushOnInactive

**Failure excerpt:**
```
java.lang.AssertionError: 
Expected: <0>
     but: was <2>

  at __randomizedtesting.SeedInfo.seed([38FC56033AE2DC88:DF99C84AD94DDF44]:0)
  at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:18)
  at org.junit.Assert.assertThat(Assert.java:956)
  at org.junit.Assert.assertThat(Assert.java:923)
  at org.elasticsearch.indices.flush.FlushIT.lambda$testFlushOnInactive$1(FlushIT.java:145)
  at org.elasticsearch.test.ESTestCase.assertBusy(ESTestCase.java:1098)
  at org.elasticsearch.indices.flush.FlushIT.testFlushOnInactive(FlushIT.java:143)
  at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(NativeMethodAccessorImpl.java:-2)
  at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
  at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.lang.reflect.Method.invoke(Method.java:568)
  at com.carrotsearch.randomizedtesting.RandomizedRunner.invoke(RandomizedRunner.java:1758)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$8.evaluate(RandomizedRunner.java:946)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$9.evaluate(RandomizedRunner.java:982)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$10.evaluate(RandomizedRunner.java:996)
  at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
  at org.apache.lucene.tests.util.TestRuleSetupTeardownChained$1.evaluate(TestRuleSetupTeardownChained.java:44)
  at org.apache.lucene.tests.util.AbstractBeforeAfterRule$1.evaluate(AbstractBeforeAfterRule.java:43)
  at org.apache.lucene.tests.util.TestRuleThreadAndTestName$1.evaluate(TestRuleThreadAndTestName.java:45)
  at org.apache.lucene.tests.util.TestRuleIgnoreAfterMaxFailures$1.evaluate(TestRuleIgnoreAfterMaxFailures.java:60)
  at org.apache.lucene.tests.util.TestRuleMarkFailure$1.evaluate(TestRuleMarkFailure.java:44)
  at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
  at com.carrotsearch.randomizedtesting.ThreadLeakControl$StatementRunner.run(ThreadLeakControl.java:375)
  at com.carrotsearch.randomizedtesting.ThreadLeakControl.forkTimeoutingTask(ThreadLeakControl.java:824)
  at com.carrotsearch.randomizedtesting.ThreadLeakControl$3.evaluate(ThreadLeakControl.java:475)
  at com.carrotsearch.randomizedtesting.RandomizedRunner.runSingleTest(RandomizedRunner.java:955)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$5.evaluate(RandomizedRunner.java:840)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$6.evaluate(RandomizedRunner.java:891)
  at com.carrotsearch.randomizedtesting.RandomizedRunner$7.evaluate(RandomizedRunner.java:902)
  at org.apache.lucene.tests.util.AbstractBeforeAfterRule$1.evaluate(AbstractBeforeAfterRule.java:43)
  at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
  at org.apache.lucene.tests.util.TestRuleStoreClassName$1.evaluate(TestRuleStoreClassName.java:38)
  at com.carrotsearch.randomizedtesting.rules.NoShadowingOrOverridesOnMethodsRule$1.evaluate(NoShadowingOrOverridesOnMethodsRule.java:40)
  at com.carrotsearch.randomizedtesting.rules.NoShadowingOrOverridesOnMethodsRule$1.evaluate(NoShadowingOrOverridesOnMethodsRule.java:40)
  at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
  at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
  at org.apache.lucene.tests.util.TestRuleAssertionsRequired$1.evaluate(TestRuleAssertionsRequired.java:53)
  at org.apache.lucene.tests.util.AbstractBeforeAfterRule$1.evaluate(AbstractBeforeAfterRule.java:43)
  at org.apache.lucene.tests.util.TestRuleMarkFailure$1.evaluate(TestRuleMarkFailure.java:44)
  at org.apache.lucene.tests.util.TestRuleIgnoreAfterMaxFailures$1.evaluate(TestRuleIgnoreAfterMaxFailures.java:60)
  at org.apache.lucene.tests.util.TestRuleIgnoreTestSuites$1.evaluate(TestRuleIgnoreTestSuites.java:47)
  at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
  at com.carrotsearch.randomizedtesting.ThreadLeakControl$StatementRunner.run(ThreadLeakControl.java:375)
  at com.carrotsearch.randomizedtesting.ThreadLeakControl.lambda$forkTimeoutingTask$0(ThreadLeakControl.java:831)
  at java.lang.Thread.run(Thread.java:833)

```

kingherc:
org.elasticsearch.indices.flush.FlushIT#testFlushOnInactive would
sometimes fail in the following case:
* SHARD_MEMORY_INTERVAL_TIME_SETTING is set very low, e.g., 10ms
* The regularly scheduled multiple flushes proceed to
  org.elasticsearch.index.shard.IndexShard#flushOnIdle
* There, the first flush will handle e.g., the first document
  that was indexed. The second flush will arrive shortly after,
  before the first flush finishes.
* The second flush will find that wasActive = true (due to the
  indexing of the remaining documents), and will set it to false.
* However, the second flush will not be executed because
  waitIfOngoing = false, and there is the ongoing first flush.
* No other flush is scheduled (since any next regularly scheduled
  flush will find wasActive = false), which creates the problem.

Fixes #87888


kingherc:
Hi @henningandersen and @DaveCTurner . After investigating the test failure, we (with the help of @fcofdez ) figured out the situation, which is presented in the PR description. It can happen in some rare situations when the SHARD_MEMORY_INTERVAL_TIME_SETTING is set too low and/or if a flush takes too long.

There are 3 solutions we discussed so far:
1. The current PR's solution, to change the waitIfOngoing to true. Drawback: one of the flush threadpool's threads may wait on the flushLock. But for normal values of SHARD_MEMORY_INTERVAL_TIME_SETTING (e.g. the default is 5 seconds), this drawback would appear in some very weirdly long-running flush that was being executed already. This is why the PR has this solution.
2. To avoid the drawback of the first solution, we could "chain" a second flush request to the first ongoing request. And make the first request, when finished, execute any chained requests. That way, we do not have a thread waiting on the flushLock. Drawback: complexity vs the other approaches.
3. We remove the wasActive smartness. Drawback: every SHARD_MEMORY_INTERVAL_TIME_SETTING ms, there will be a possible no-op flush request scheduled.

Feel free to tell us your opinion on the selected approach and any other thoughts you may have.

henningandersen:
It does look like the old synced-flush would use `waitIfOngoing=true`, see [here](https://github.com/elastic/elasticsearch/blob/bcfbf000748b538884087623113682165abcb92d/server/src/main/java/org/elasticsearch/indices/flush/SyncedFlushService.java#L589). However, I wonder if we could let `flush` return a boolean if it attempted the flush or not - and if not, set `active` back to true?

A more straightforward solution could also be to move the setting of `active` to be after indexing is done rather than before? This also affects `scheduledRefresh`, but I think it is fine to not consider an indexing request that is ongoing active there too. I think I'd prefer that option, unless I am overseeing something.

fcofdez:
> A more straightforward solution could also be to move the setting of active to be after indexing is done rather than before? This also affects scheduledRefresh, but I think it is fine to not consider an indexing request that is ongoing active there too. I think I'd prefer that option, unless I am overseeing something.

I'm not sure if this solves the issue? Theoretically we could still miss the last flush if the first flush is still running after the remaining documents have been indexed. But maybe I'm missing something here.

henningandersen:
> I'm not sure if this solves the issue? Theoretically we could still miss the last flush if the first flush is still running after the remaining documents have been indexed. But maybe I'm missing something here.

You might be right :slightly_smiling_face: . The idea would be that by marking active after the indexing has occurred, the next round of `flushOnIdle` would know that it needs to flush?

kingherc:
Hi! Thanks for the awesome conversation. I think I agree with @fcofdez . If we follow that approach @henningandersen , I so see a rare situation where:
* There's an ongoing flush of docs [1,N]
* A new doc N+1 is indexed and sets active = true
* The next flushOnIdle() is called, which sets active = false, and tries another flush request, which fails because of the ongoing flush.
* The ongoing flush finishes, but has missed N+1 document.
* Any next scheduled flushOnIdle() will not try flush requests since active = false.
Right?

> However, I wonder if we could let flush return a boolean if it attempted the flush or not - and if not, set active back to true?

I think this is also a nice solution. I can try that if you agree.

fcofdez:
Just to add more context here, this only affects in cases where we stop indexing after the latest flush is skipped (a rare edge case)

> I think this is also a nice solution. I can try that if you agree.

👍 let's try that.

henningandersen:
Makes sense, the `active` flag would need to be cleared differently too (inside flush) if were to pursue the solution where we set active after indexing.

kingherc:
Hi @fcofdez , @henningandersen . Thanks for the conversation. The method where if flush does not wait for the ongoing flush, it returns false, and we set the active flag back to true, works. I did that -- feel free to review the PR.

fcofdez:
It looks like there are some related test failures, additionally we usually avoid force-pushing since Github gets confused and it hides/removes some review comments.

kingherc:
> It looks like there are some related test failures, additionally we usually avoid force-pushing since Github gets confused and it hides/removes some review comments.

Fixed the test.

Oh, about the force pushes, I will avoid from now on. Either way the commits get squashed when merging the PR, so indeed I do not see a reason now why I did it :) 
Did you miss adding a comment? if yes, please mention it again please.

server/src/main/java/org/elasticsearch/index/shard/IndexShard.java
```diff
@@ -1355,8 +1356,11 @@ public void flush(FlushRequest request) {
          */
         verifyNotClosed();
         final long time = System.nanoTime();
-        getEngine().flush(force, waitIfOngoing);
-        flushMetric.inc(System.nanoTime() - time);
+        boolean flushHappened = getEngine().flush(force, waitIfOngoing);
+        if (flushHappened) {
```
henningandersen, line 1360: I am in doubt about this. The [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-stats.html#cluster-nodes-stats-api-response-body) says this is "Number of flush operations". So far it has been the same as API calls, now it is only those that do anything. It is a subtle change and hardly breaking.

But this change puts the flush-metric between API calls and actual flushes. It is not actual flushes, because if there is nothing to flush, we will still increase it here.

So I'd prefer to inc always here and then we can always make it more accurate.

server/src/main/java/org/elasticsearch/index/shard/IndexShard.java
```diff
@@ -2188,8 +2192,12 @@ public void onFailure(Exception e) {
 
                     @Override
                     protected void doRun() {
-                        flush(new FlushRequest().waitIfOngoing(false).force(false));
-                        periodicFlushMetric.inc();
+                        if (flush(new FlushRequest().waitIfOngoing(false).force(false))) {
+                            periodicFlushMetric.inc();
```
henningandersen, line 2196: I think the same argument as above holds here too.

henningandersen (commented):
This directions looks good to me. I have a few comments that I'd like to see addressed though.

server/src/main/java/org/elasticsearch/index/engine/ReadOnlyEngine.java
```diff
@@ -437,8 +437,8 @@ public boolean shouldPeriodicallyFlush() {
     }
 
     @Override
-    public void flush(boolean force, boolean waitIfOngoing) throws EngineException {
-        // noop
+    public boolean flush(boolean force, boolean waitIfOngoing) throws EngineException {
+        return true; // noop
```
fcofdez, line 441: I think this should return false? This implementation won't flush.

fcofdez (commented):
I left a small comment, the direction looks good.

server/src/main/java/org/elasticsearch/index/engine/ReadOnlyEngine.java
henningandersen: But that is not really the meaning of the return value as it stands in `InternalEngine`. `InternalEngine` may also return true without flushing. What the return value currently mean is

false: if `waitIfOngoing` detected a collision
true: otherwise.

Here true seems most correct since no collision was detected.

fcofdez: Yes, I was wrong, please dismiss my comment.

server/src/main/java/org/elasticsearch/index/engine/Engine.java
```diff
@@ -1015,8 +1015,9 @@ public boolean refreshNeeded() {
      * @param force         if <code>true</code> a lucene commit is executed even if no changes need to be committed.
      * @param waitIfOngoing if <code>true</code> this call will block until all currently running flushes have finished.
      *                      Otherwise this call will return without blocking.
+     * @return <code>true</code> if the flush happened, else <code>false</code> (e.g., if it did not wait for an ongoing request)
```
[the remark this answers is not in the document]
kingherc: Done

server/src/main/java/org/elasticsearch/index/shard/IndexShard.java
kingherc: Done

server/src/main/java/org/elasticsearch/index/engine/InternalEngine.java
```diff
@@ -1908,7 +1908,8 @@ public void flush(boolean force, boolean waitIfOngoing) throws EngineException {
             if (flushLock.tryLock() == false) {
                 // if we can't get the lock right away we block if needed otherwise barf
                 if (waitIfOngoing == false) {
-                    return;
+                    logger.trace("returning as there is an in-flight flush that we do not need to wait for");
```
henningandersen, line 1911: nit: I'd like the log message to be less about the "return" statement and more about the situation:

```suggestion
                    logger.trace("detected an in-flight flush, not blocking to wait for it's completion");
```

server/src/main/java/org/elasticsearch/index/shard/IndexShard.java
```diff
@@ -1343,8 +1343,9 @@ public BulkStats bulkStats() {
      * Executes the given flush request against the engine.
      *
      * @param request the flush request
+     * @return <code>false</code> if the flush did not wait for an ongoing request and returned, else <code>true</code>
```
henningandersen, line 1346:
```suggestion
     * @return <code>false</code> if waitIfOngoing==true and an ongoing request is detected, else <code>true</code>. If false is returned, no flush happened.
```

server/src/main/java/org/elasticsearch/index/shard/IndexShard.java
```diff
@@ -2188,7 +2190,10 @@ public void onFailure(Exception e) {
 
                     @Override
                     protected void doRun() {
-                        flush(new FlushRequest().waitIfOngoing(false).force(false));
+                        if (flush(new FlushRequest().waitIfOngoing(false).force(false)) == false) {
+                            // In case the flush did not happen, revert active flag so that a next flushOnIdle request can happen (#87888)
```
henningandersen, line 2194:
```suggestion
                            // In case an ongoing flush was detected, revert active flag so that a next flushOnIdle request will retry (#87888)
```

server/src/test/java/org/elasticsearch/index/shard/IndexShardTests.java
```diff
@@ -3897,6 +3898,63 @@ public void testFlushOnIdle() throws Exception {
...
+                new MockLogAppender.SeenEventExpectation(
+                    "should see first flush getting lock immediately",
+                    Engine.class.getCanonicalName(),
+                    Level.TRACE,
+                    "acquired flush lock immediately"
+                )
+            );
+            assertBusy(mockLogAppender::assertAllExpectationsMatched);
+
+            // While the previous flushOnIdle request is happening, immediately issue a second flush request with waitIfOngoing=false
```
henningandersen, line 3928: I think we need to block the original flush in order to be sure this happens?

henningandersen (commented):
I think there is a problem with the new test, otherwise this looks good.

server/src/test/java/org/elasticsearch/index/shard/IndexShardTests.java
kingherc: Hi @henningandersen , indeed, that would be best. For the moment, I've indexed a large-enough amount to ensure the flush is still ongoing. Would you have a recommendation about how to inject a block of the original flush just for the test (rather than in reality)?

kingherc: After some tips from @fcofdez , I found a nice way to do it. Feel free to review :)

henningandersen (approved):
LGTM.

fcofdez (approved):
LGTM 👍

server/src/test/java/org/elasticsearch/index/shard/IndexShardTests.java
```diff
@@ -3897,6 +3899,78 @@ public void testFlushOnIdle() throws Exception {
...
+            mockLogAppender.addExpectation(
+                new MockLogAppender.SeenEventExpectation(
+                    "should see first flush releasing lock",
+                    Engine.class.getCanonicalName(),
+                    Level.TRACE,
+                    "released flush lock"
+                )
+            );
+            assertBusy(mockLogAppender::assertAllExpectationsMatched);
+
```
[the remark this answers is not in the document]
kingherc: Hi @henningandersen , could not do it immediately because I called flush() instead of a flushOnIdle request. Modified the test slightly to also do another flushOnIdle request to test that the active flag turns true. Thanks for the suggestion!

kingherc:
@elasticmachine run elasticsearch-ci/part-1 please

server/src/test/java/org/elasticsearch/index/shard/IndexShardTests.java
```diff
@@ -3897,6 +3899,89 @@ public void testFlushOnIdle() throws Exception {
...
+            // The second flush (that did not happen) should have returned false and turned the active flag to true
```
henningandersen, line 3972: I am not sure I understand this comment, is `active` not only set/reset in `flushOnIdle` and not in `flush`?

kingherc: Hi @henningandersen , `active` is only toggled during the `flushOnIdle()` function (which later calls `flush()`) and **not** during `flush()` itself.

henningandersen: Perhaps change comment to be about `flushOnIdle`? The second `flush` is easy to mistake for the flush a few lines up...

kingherc: Oh I see, sure! I did that and improved a bit the comments.

PR #89442, opened 2022-08-17
Better support for multi cluster for run task

jakelandis:
This commit introduces a `./gradlew run-ccs` task with the following goals:
* mirror the ease of use of `./gradlew run` for manual cross cluster search development
  * same credentials
  * same well known ports
  * uses ccs specific naming
  * enable debugging across both clusters

This is admittedly kinda hacky. Test clusters have support multi-cluster and are in use for 
 for automated testing. There are some nuances that make that setup (and this setup) 
a bit cumbersome..specifically needing to read one cluster's config to configure another cluster. 

The run task adds a bit more config (well defined ports, etc.) than the tests need to 
so that also complicates this abit more. I found that without the additions here I was unable to get both 
sharing of cluster configuration (like in the [tests](https://github.com/elastic/elasticsearch/blob/main/qa/ccs-common-rest/build.gradle#L55)) 
and the run task's hard coded config to work well together. Hopefully the additions to the run task are not
too hacky as I could not find any other way. 

===

Also FWIW, this is in support of a new security model for CCS and CCR so there will likely
be more PR's with multi-cluster configs to help testing that effort.

### DIFF
--- a/build-tools-internal/src/main/groovy/elasticsearch.run-ccs.gradle
+++ b/build-tools-internal/src/main/groovy/elasticsearch.run-ccs.gradle
@@ -44,7 +44,7 @@ tasks.register('initfulfillingCluster', RunTask) {
   useCluster testClusters.named("fulfilling-cluster")
   initOnly = true //only initialize the testCluster, don't start it
   portOffset = 1 //when only initializing, instruct to use one above the normal ports to avoid collisions when other cluster also initializes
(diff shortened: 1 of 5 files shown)


mark-vieira (approved):
Multi-cluster configuration with the run task in indeed quite awkward. We could improve this but this is probably fine for now given it's pretty self-contained.

jakelandis:
Thanks Mark !

PR #89444, opened 2022-08-18
[Test] Speed up ApiKeyIntegTests

ywangd:
When creating multiple API keys, instead of waiting for the security
index to refresh on each API key, do not wait for refresh and force
refresh immediately on the last key creation. This helps bringing the
executime time of the test suite from 4m10s to 1m32s.

### DIFF
--- a/x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
+++ b/x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
@@ -301,6 +301,7 @@ public class ApiKeyIntegTests extends SecurityIntegTestCase {
                 .setExpiration(null)
                 .setRoleDescriptors(Collections.singletonList(descriptor))
                 .setMetadata(ApiKeyTests.randomMetadata())
+                .setRefreshPolicy(WriteRequest.RefreshPolicy.NONE)
                 .get();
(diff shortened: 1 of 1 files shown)


x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
```diff
@@ -301,6 +301,7 @@ public void testMultipleApiKeysCanHaveSameName() {
                 .setExpiration(null)
                 .setRoleDescriptors(Collections.singletonList(descriptor))
                 .setMetadata(ApiKeyTests.randomMetadata())
+                .setRefreshPolicy(WriteRequest.RefreshPolicy.NONE)
```
n1v0lg, line 304: Sanity check: we don't need `i == noOfApiKeys - 1 ? WriteRequest.RefreshPolicy.IMMEDIATE : WriteRequest.RefreshPolicy.NONE` here because we are not interested in searching afterwards, correct?

n1v0lg (approved):
LGTM

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/xpack/security/authc/ApiKeyIntegTests.java
ywangd: 👍

PR #89449, opened 2022-08-18
Run potentially expensive checks in SystemIndexMetadataUpgraderService off the applier thread

original-brownbear:
Checking all indices for changes during all cluster state updates can become very expensive at scales of O(10k) indices or more.
It takes more than 1% of the total CPU time when bootstrapping the many-shards benchmark cluster and quite a bit more than
that in terms of wall-time by holding up CS application. As far as I can see we can safely move this check to the management pool.

In order to simplify the mechanics of this move, this change brings the check in-line with how we do similar checks in e.g. the snapshot
service. We don't have to hold on to the last inspected map of indices here. We can simply check all indices when becoming master
and then keep checking against the last applied state. This makes the logic in the CS update task that might be triggered easier
to reason about and more stable as well. For the master update task this comes at the cost of making it slightly more expensive but as it should effectively never run, its performance is irrelevant here.

relates #77466

### DIFF
--- a/server/src/main/java/org/elasticsearch/cluster/metadata/SystemIndexMetadataUpgradeService.java
+++ b/server/src/main/java/org/elasticsearch/cluster/metadata/SystemIndexMetadataUpgradeService.java
@@ -15,10 +15,11 @@ import org.elasticsearch.cluster.ClusterState;
 import org.elasticsearch.cluster.ClusterStateListener;
 import org.elasticsearch.cluster.ClusterStateUpdateTask;
 import org.elasticsearch.cluster.service.ClusterService;
-import org.elasticsearch.common.collect.ImmutableOpenMap;
 import org.elasticsearch.common.settings.Settings;
+import org.elasticsearch.common.util.concurrent.AbstractRunnable;
 import org.elasticsearch.core.SuppressForbidden;
 import org.elasticsearch.indices.SystemIndices;
+import org.elasticsearch.threadpool.ThreadPool;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/elasticsearch/cluster/metadata/SystemIndexMetadataUpgradeService.java
```diff
@@ -36,41 +37,55 @@ public class SystemIndexMetadataUpgradeService implements ClusterStateListener {
...
-                            break;
+            final var previousIndices = event.previousState().metadata().indices();
+            final long triggerV = event.state().version();
+            triggeredVersion = triggerV;
+            // Fork to the management pool to avoid blocking the cluster applier thread unnecessarily for very large index counts
+            // TODO: we should have a more efficient way of getting just the changed indices so that we don't have to fork here
+            clusterService.threadPool().executor(ThreadPool.Names.MANAGEMENT).execute(new AbstractRunnable() {
+                @Override
+                protected void doRun() {
+                    if (triggeredVersion != triggerV) {
```
williamrandolph, line 64: I just want to make sure I understand what is happening here. Example scenario:

1. Update A happens, and index metadata is changed. We set the volatile `triggeredVersion` field to this change event's version, and submit a new task to the management thread pool.
2. Update B happens, and index metadata is changed again. We set the volatile `triggeredVersion` field to B's version, and submit a new task to the management thread pool.
3. The management thread pool begins processing the task from update A, but `triggeredVersion` has changed, so we return without running expensive updates.
4. The management thread pool begins processing the task from update B, and finds that `triggeredVersion` has not changed, so we submit our tasks to update system index metadata.

Since updating index metadata is really a matter of taking current index metadata and making it consistent with system expectations, we don't need to worry about dropping these intermediate updates. As long as we have processed the most recent update, our metadata should end up in a good state.

Am I getting anything wrong?

williamrandolph (commented):
The idea here makes really good sense to me. Will we be able to verify the performance improvement for the many shards benchmark after we merge?

I am almost ready to approve, but I want to make sure my understanding of the change is correct.

server/src/main/java/org/elasticsearch/cluster/metadata/SystemIndexMetadataUpgradeService.java
original-brownbear: No, you're spot on :)

original-brownbear:
> Will we be able to verify the performance improvement for the many shards benchmark after we merge?

Yes, I found this in profiling the many-shards benchmarks and this will shave off a couple percent from the single-threaded cluster state update path for large clusters in a fairly straight-forward manner.

williamrandolph (approved):
LGTM! Thanks for fixing this.

original-brownbear:
Thanks William!

PR #89458, opened 2022-08-18
Fork to MANAGEMENT for ILM cache cleanup operations

original-brownbear:
No need to run this synchronously. The loop setting up the deleted indices list
takes quite a while for large clusters (1%+ of CPU time and more than that in wall time
during many-shards benchmark bootstrapping)
during many shards benchmark bootstrap. There's no value in running the cleanup in-sync
on the applier thread, so we can just fork off and save the time.

relates #77466

### DIFF
--- a/x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/IndexLifecycleService.java
+++ b/x-pack/plugin/ilm/src/main/java/org/elasticsearch/xpack/ilm/IndexLifecycleService.java
@@ -308,8 +308,14 @@ public class IndexLifecycleService
 
         // if we're the master, then process deleted indices and trigger policies
         if (this.isMaster) {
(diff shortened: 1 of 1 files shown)


dakrone:
@original-brownbear should the `indicesDeleted()` method on the `ClusterChangedEvent` be memoized so that its computation is not expensive more than once? (The helpfulness would obviously depend on how many times we use it, so we'd need to determine whether it's worth the cost). Are there other event methods that could be memoized?

original-brownbear:
@dakrone currently this is the only spot where we use this list by the master role so it wouldn't help here. Likewise, for the data role we only have one usage as well.
So memoizing would only help if we have both roles, but in that case we probably don't have enough indices for it to matter.

dakrone (approved):
Makes sense, thanks for the info.

original-brownbear:
Np + Thanks Lee

PR #89466, opened 2022-08-18
Synthetic _source: support ignore_above

nik9000:
This allows you to use `ignore_above` with `keyword` fields in synthetic
source. Ignored values are stored in a "backup" stored field and added
to the end of the list of results. This makes `ignore_above` work pretty
much the same way as it does when you don't have synthetic source. The
only difference is the order of the results. But synthetic source
changes the order of results anyway. That should be fine.


rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search/400_synthetic_source.yml
```diff
@@ -29,6 +29,7 @@ keyword:
           query:
             ids:
               values: [1]
+  - is_false: hits.hits.0.fields
```
nik9000, line 32: There's a different bug here that I'm fixing along side `ignore_above` but I only noticed it when I was working on this. Previously we were adding all stored fields we loaded to the fetched field list. Now we only add if you ask for them.

server/src/main/java/org/elasticsearch/search/fetch/FetchPhase.java
```diff
@@ -257,7 +257,9 @@ private static FieldsVisitor createStoredFieldsVisitor(
             boolean loadSource = sourceRequired(context);
             if (loadSource) {
                 if (false == sourceLoader.requiredStoredFields().isEmpty()) {
-                    return new CustomFieldsVisitor(sourceLoader.requiredStoredFields(), true);
+                    // add the stored fields needed to load the source mapping to an empty set so they aren't returned
+                    sourceLoader.requiredStoredFields().forEach(fieldName -> storedToRequestedFields.putIfAbsent(fieldName, Set.of()));
+                    return new CustomFieldsVisitor(storedToRequestedFields.keySet(), true);
```
nik9000, line 262: This change fixes that leaking `fields` in the search response. It's tricky!

server/src/main/java/org/elasticsearch/index/fieldvisitor/FieldsVisitor.java
```diff
@@ -84,6 +84,9 @@ public Set<String> getFieldNames() {
     public final void postProcess(Function<String, MappedFieldType> fieldTypeLookup) {
         for (Map.Entry<String, List<Object>> entry : fields().entrySet()) {
             MappedFieldType fieldType = fieldTypeLookup.apply(entry.getKey());
+            if (fieldType == null) {
+                continue; // TODO this is lame
+            }
```
nik9000, line 89: I don't like adding such looseness. I'm sure there is a better way than this, but I couldn't think of something off the bat and figured we'd be refactoring here before too long anyway. So I wanted to see what other folks think.

I don't particularly like that this `valueForDisplay` thing is mixed up into the Lucene loading. That feels like it should be something we take care of externally but I have no idea how mixed up it is with other things.

nik9000:
Note for those following along at home - this doesn't really effect the performance of loading synthetic source any - even if there are many, many fields, all of which have `ignore_above`. It's in the microseconds to build the extra HashMap entries. If you are worrying about microseconds you want to keep the mapping small.

server/src/main/java/org/elasticsearch/index/fieldvisitor/FieldsVisitor.java
romseygeek: Yeah this really should be pulled out elsewhere, but I think I'm fine with it staying in here for the time being.

server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -945,6 +959,10 @@ private void indexValue(DocumentParserContext context, String value) {
 
         if (value.length() > fieldType().ignoreAbove()) {
             context.addIgnoredField(name());
+            if (context.isSyntheticSource()) {
+                // Save a copy of the field so synthetic source can load it
+                context.doc().add(new Field(originalName(), value, ORIGINAL_FIELD_TYPE));
```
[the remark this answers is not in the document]
romseygeek: If `stored=true` is set on here then we store it twice, right? Is it worth detecting that case and deferring to the 'normal' stored field rather than always using the hidden field?

server/src/main/java/org/elasticsearch/search/fetch/FetchPhase.java
romseygeek: Sneaky.  Nice catch.

romseygeek (commented):
Thanks, I think this makes sense.  I have a question around what to do if the field is already being stored.

server/src/main/java/org/elasticsearch/index/mapper/KeywordFieldMapper.java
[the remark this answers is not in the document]
nik9000: If I did that then fields longer than the limit would start being loadable via stored fields. That feels like something folks could accidentally rely on.

server/src/main/java/org/elasticsearch/index/fieldvisitor/FieldsVisitor.java
nik9000: That's what I was thinking. I don't want us to start relying on this null-ok behavior in other places. I don't particularly want to build some fancy abstraction for this either. Any ideas on something a little more defensive than this but not super overkill if we're going to rip it out?

nik9000:
Ready for you again @romseygeek !

romseygeek (approved):
LGTM

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/get/100_synthetic_source.yml
```diff
@@ -33,6 +33,7 @@ keyword:
   - match:
       _source:
         kwd: foo
+#  - is_false: fields  TODO fix me
```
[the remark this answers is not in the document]
nik9000: It's the same problem as we have on fetch! I just noticed it and didn't want to forget it. I thought maybe I'd fix it like I did the fetch one. But it was a little more complex so I didn't. But I'll grab it soon!

PR #89467, opened 2022-08-18
Remove unexpected meta param in agg's response

### LINKED ISSUE #89455: Unexpected meta param found in filters agg's response
### Elasticsearch Version

8.2.2

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

Linux 249a534da7fc 3.10.0-1160.6.1.el7.x86_64 #1 SMP Tue Nov 17 13:59:11 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

### Problem Description

Unexpected meta param found in filters agg's response.

Perhaps when agg is rewritten, an empty meta will be returned.

https://github.com/elastic/elasticsearch/blob/8.3/server/src/main/java/org/elasticsearch/search/aggregations/AggregationBuilder.java#L127
```java
    @Override
    public final AggregationBuilder rewrite(QueryRewriteContext context) throws IOException {
        AggregationBuilder rewritten = doRewrite(context);
        AggregatorFactories.Builder rewrittenSubAggs = factoriesBuilder.rewrite(context);
        if (rewritten != this) {
            return rewritten.setMetadata(getMetadata()).subAggregations(rewrittenSubAggs);
        } else if (rewrittenSubAggs != factoriesBuilder) {
            return shallowCopy(rewrittenSubAggs, getMetadata());
        } else {
            return this;
        }
    }
```

https://github.com/elastic/elasticsearch/blob/8.3/server/src/main/java/org/elasticsearch/search/aggregations/AbstractAggregationBuilder.java#L117
```java
    public Map<String, Object> getMetadata() {
        return metadata == null ? Collections.emptyMap() : Collections.unmodifiableMap(metadata);
    }
```

### Steps to Reproduce

```
DELETE test_1

PUT test_1
{
  "mappings": {
    "properties": {
      "int_field":{
        "type": "integer"
      }
    }
  }
}

PUT test_1/_bulk?refresh
{"index":{"_id":"1"}}
{"int_field":1}
{"index":{"_id":"2"}}
{"int_field":51}
{"index":{"_id":"3"}}
{"int_field":101}
{"index":{"_id":"4"}}
{"int_field":151}

GET test_1/_search
{
  "size": 0,
  "aggs": {
    "the_filters": {
      "filters": {
        "filters": [
          {
            "match": {
              "int_field": 101
            }
          },
          {
            "match": {
              "int_field": 151
            }
          }
        ]
      }
    }
  }
}

response:
{
    "took": 3,
    "timed_out": false,
    "_shards": {
        "total": 1,
        "successful": 1,
        "skipped": 0,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 4,
            "relation": "eq"
        },
        "max_score": null,
        "hits": []
    },
    "aggregations": {
        "the_filters": {
            "meta": {},
            "buckets": [
                {
                    "doc_count": 1
                },
                {
                    "doc_count": 1
                }
            ]
        }
    }
}
```

### Logs (if relevant)

_No response_

amooncake:
Closes #89455


nik9000 (commented):
I wonder if we're better off keeping the emptyMap stuff and not rendering the object if it's empty.

We probably also want some kind of test for this. Since it's about the HTTP response it's probably right to stick it in the `220_filters_bucket.yml` file. I believe `is_false` will fail if we send an empty map but succeed for missing stuff.

nik9000 (commented):
I think it'd be nicer to fix this on the other side - if the `meta` is empty then don't render it. I know we were using `null` for that. But:
1. I don't like using `null` for that. It feels icky.
2. It makes it easy to make these kinds of problems.

nik9000:
That point is totally up for debate, btw, but I feel like "never render empty" is a simpler rule.

nik9000:
@elasticmachine, ok to test

amooncake:

nik9000 (approved):

nik9000:
@amooncake I've merged this one! Sorry it took me so long. I was thinking it'd be better to only render if the thing wasn't empty, but that's *technically* breaking and it isn't worth the risk of breaking folks.

nik9000:
So! Thanks again for doing it!

amooncake:
Thanks @nik9000 !

PR #77864, opened 2021-09-16
Deprecate ignore_throttled parameter

martijnvg:
Backporting #77479 to 7.x branch.

Frozen indices are deprecated, which will make ignore_throttled obsolete.

Some changes to HLRC  HLRC and related tests were required in order to deprecate ignore_throttled parameter:
* Change HLRC request classes to set indices options initial value to null.
    This avoids sending the ignore_throttled parameter (and other indices options parameters)
    and therefore avoids the depracation warning header in the response. All these
    request classes had the indices options set to what is the defaukt in the corresponding
    action request class. So sending these indices options params was a noop, which is the same
    as not sending indices options parameters, which is what this change does.
* For transport action request classes that are reused in the HLRC as request class,
    change the corresponding hlrc request converter to only send indices options params
    if the indices options are different than what the default is.
* For tests that use a non default indices options, allow 'ignore_throttled is deprecated'
    warning.

Other changes:
* A number of tests have been changed to allowed warning header for deprecated ignore_throttled parameter.
* Some rest v7 compat yaml tests are temporary muted, but will be unmuted once this
    change has been backported.

Relates to #70192

PR #89469, opened 2022-08-18
Add note about inadvertent breakage for HLRC in 7.16 regarding indices options

dakrone:
This adds a note about an unintended break in behavior for the HLRC with regard to the indices
options that are sent by default with requests.

Relates to #77864


swallez (approved):
LGTM

dakrone:
@elasticmachine update branch

PR #89470, opened 2022-08-18
[ML] Refactor memory autoscaling decider into its own class

dimitris-athanasiou:
This commit factors out the code that decides how much memory
is required during an autoscale request into its own decider
class. This prepares the ML autoscaling decider service to
accommodate more deciders for processors and storage in the
future.


edsavage (approved):
LGTM

PR #89473, opened 2022-08-18
Add source fallback support for match_only_text mapped type

jdconrad:
This change adds access to mapped match_only_text fields via the Painless scripting fields API. The values returned from a match_only_text field via the scripting fields API always use source as described by (https://github.com/elastic/elasticsearch/issues/81246). These are not available via doc values so there are no bwc issues.

### DIFF
--- a/docs/changelog/89473.yaml
+++ b/docs/changelog/89473.yaml
@@ -1,0 +1,5 @@
+pr: 89473
+summary: Add source fallback support for `match_only_text` mapped type
+area: Mapping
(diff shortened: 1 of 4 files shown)


javanna (approved):
LGTM thanks @jdconrad !

javanna:
This one should close #81246 for real this time? :)

jdconrad:
@javanna Thanks for the review! I do believe it should indeed close #81246 for real this time :)

PR #89491, opened 2022-08-22
Rollup performance improvement: mv Set.union to outer method

weizijun:
Rollup performance improvement: mv Set.union to outer method. To avoid the heavy cost in every doc collect.
In some case, the Set.union cost 18% cpu performance.
image


csoulios:
@elasticmachine ok to test

csoulios (approved):
LGTM!

csoulios:
@elasticmachine update branch

PR #89492, opened 2022-08-22
More accurate error message for LDAP user modes

ywangd:
When a ldap relam is configured with settings for both user search and
user dn modes, ES node will fail to start. This PR improves the error
mesesage so it lists the exact conflicting user-search settings instead
of always printing out base_dn.

### DIFF
--- a/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ldap/LdapUserSearchSessionFactory.java
+++ b/x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ldap/LdapUserSearchSessionFactory.java
@@ -17,6 +17,7 @@ import com.unboundid.ldap.sdk.SimpleBindRequest;
 import org.elasticsearch.action.ActionListener;
 import org.elasticsearch.action.ActionRunnable;
 import org.elasticsearch.common.settings.SecureString;
+import org.elasticsearch.common.settings.Setting;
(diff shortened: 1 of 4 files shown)


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/ldap/LdapUserSearchSessionFactory.java
```diff
@@ -66,14 +67,14 @@ class LdapUserSearchSessionFactory extends PoolingSessionFactory {
...
-    static boolean hasUserSearchSettings(RealmConfig config) {
+    static List<? extends Setting.AffixSetting<?>> configuredUserSearchSettings(RealmConfig config) {
         return Stream.of(
             LdapUserSearchSessionFactorySettings.SEARCH_BASE_DN,
             LdapUserSearchSessionFactorySettings.SEARCH_ATTRIBUTE,
             LdapUserSearchSessionFactorySettings.SEARCH_SCOPE,
             LdapUserSearchSessionFactorySettings.SEARCH_FILTER,
             LdapUserSearchSessionFactorySettings.POOL_ENABLED
-        ).anyMatch(config::hasSetting);
+        ).filter(config::hasSetting).toList();
```
ywangd, line 77: Theorectically all settings under `user_search.` prefix should be considered. But that would be a breaking change. So keep this logic as is with these 5 settings.

tvernum (approved):
LGTM

PR #89494, opened 2022-08-22
Fix wrong stepkey for nextKey in StepTestCase#mutateInstance(...)

mushao999:
This pr fix incorrect stepKey for nextKey in StepTestCase#mutateInstance(...).


docs/changelog/89494.yaml
```diff
@@ -0,0 +1,5 @@
+pr: 89494
```
gmarouli, line 1: I do not think we need a change log for this since this is a test change. Do you mind removing it?

gmarouli (changes requested):
@mushao999 thank you for your contribution and your thoroughness catching all the instances of this bug. Just a request to remove the change log and then I will merge!

gmarouli:
@elasticmachine ok to test

docs/changelog/89494.yaml
mushao999: Thanks for your reply, I've removed the changelog

gmarouli: Great, thanks!

gmarouli (approved):
@mushao999 , LGTM, once again thank you for your contribution and the quick reaction to the comments! I will move forward with merging this.

PR #89495, opened 2022-08-22
Rollup performance improvement:  index bucket replace map with smile builder

weizijun:
index bucket replace map with smile builder.
It can greatly improve the build source and index api performance.


csoulios:
@elasticmachine ok to test

csoulios (commented):
This change theoretically improves indexing performance.
Have you benchmarked its impact? Can you share any numbers?

Also, can you please merge from upstream and resolve conflicts?

Thanks

weizijun:
before:


after:


In my benchmark, the downsampling cost from 2.4 minute to 2.2 minute.

csoulios:
@weizijun thanks for posting the graphs and results. 

It is worth noting the downsampling interval as well as the ratio of source_index_doc_count/rollup_index_doc_count.

We deem the downsampling operation spends more time in reading the source index than writing to the rollup index. So, improving the write part of the operation will yield small improvements compared to improving the read part.

However, this change makes perfect sense and we plan to merge it.

csoulios (approved):
LGTM!

weizijun:
> @weizijun thanks for posting the graphs and results.
> 
> It is worth noting the downsampling interval as well as the ratio of source_index_doc_count/rollup_index_doc_count.
> 
> We deem the downsampling operation spends more time in reading the source index than writing to the rollup index. So, improving the write part of the operation will yield small improvements compared to improving the read part.
> 
> However, this change makes perfect sense and we plan to merge it.

Thanks @csoulios !

PR #89496, opened 2022-08-22
Do not propagate exception in case of index corruption

idegtiarenko:
When node performs listStoreMetadata it bubbles the corrupt index exception to
the caller. This is currently treated as retryable exception however it is not
likely that index is going to be fixed. Returning empty instead.


DaveCTurner (commented):
Production change looks good but this deserves a test :)

server/src/test/java/org/elasticsearch/index/store/StoreTests.java
```diff
@@ -1187,4 +1187,14 @@ public void testVersionIsIncludedInBootstrapCommit() throws IOException {
             assertThat(segmentInfos.getUserData().get(Engine.ES_VERSION), is(equalTo(org.elasticsearch.Version.CURRENT.toString())));
         }
     }
+
+    public void testReadMetadataSnapshotReturnsEmptyWhenIndexIsCorrupted() throws IOException {
+        var shardId = new ShardId("index", "_na_", 1);
+        var dir = createTempDir();
+        try (Store store = new Store(shardId, INDEX_SETTINGS, new NIOFSDirectory(dir), new DummyShardLock(shardId))) {
+            store.markStoreCorrupted(new IOException("test exception"));
```
DaveCTurner, line 1195: Suggest this otherwise the store is empty anyway:

```suggestion
            store.createEmpty();
            store.markStoreCorrupted(new IOException("test exception"));
```

server/src/main/java/org/elasticsearch/index/store/Store.java
```diff
@@ -451,6 +451,8 @@ public static MetadataSnapshot readMetadataSnapshot(
             return MetadataSnapshot.loadFromIndexCommit(null, dir, logger);
         } catch (IndexNotFoundException ex) {
             // that's fine - happens all the time no need to log
+        } catch (CorruptIndexException e) {
+            logger.info(() -> shardId + "is corrupted", e);
```
DaveCTurner, line 455: Suggest harmonising with the `ShardLockObtainFailedException` case below (also `"is corrupted"` needs a leading space):

```suggestion
        } catch (CorruptIndexException ex) {
            logger.info(() -> format("%s: corrupted", shardId), ex);
```

DaveCTurner (approved):
LGTM with 2 nits

PR #89497, opened 2022-08-22
Rollup performance improvement: remove value validator

weizijun:
The value validator will cost 4% cpu to check the value. The validator is mainly used in the histogram. But when histogram remove, I think the validator can also be moved. 
Because the value fetcher will validator the value itself. and the format method will also check the value.
image


csoulios:
@elasticmachine update branch

csoulios:
@elasticmachine ok to test

csoulios (approved):

csoulios:
@elasticmachine update branch

PR #89498, opened 2022-08-22
[Docs] Indicate that CCR does not auto-follow existing indices, only newly created ones

tlrx:


docs/reference/ccr/apis/auto-follow/put-auto-follow-pattern.asciidoc
```diff
@@ -58,7 +58,10 @@ This API creates a new named collection of
 <<ccr-auto-follow,auto-follow patterns>> against the remote cluster
 specified in the request body. Newly created indices on the remote cluster
 matching any of the specified patterns will be automatically configured as follower
-indices. Additionally, this API can be used to update existing
+indices. Indices on the remote cluster that have been created before the auto-follow
+ pattern is created won't be auto-followed even if they match the pattern.
```
DaveCTurner, line 62: Small fix:

```suggestion
indices. Indices on the remote cluster that were created before the auto-follow
pattern is created won't be auto-followed even if they match the pattern.
```

or perhaps this?

```suggestion
indices. Indices on the remote cluster that exist when the auto-follow
pattern is created won't be auto-followed even if they match the pattern.
```

DaveCTurner (approved):
LGTM, one wording nit

docs/reference/ccr/apis/auto-follow/put-auto-follow-pattern.asciidoc
tlrx: I borrowed the first suggestion, thanks David!

PR #89505, opened 2022-08-22
Compare shards allocators IT

idegtiarenko:
This is not intended to be merged.

This test sets up 9 node cluster with 5k shards and then measures the time to create 100 more indices.
This is the results from my dev workstation:
```
Allocator [desired_balance]. avg: 243.13, max: 379, p99: 379, p95: 322
Allocator [balanced].        avg: 188.17, max: 308, p99: 308, p95: 276
```


idegtiarenko:
No longer required

PR #89513, opened 2022-08-22
Stop using BlobContainer to build paths

original-brownbear:
We were using/abusing the containers just to build the paths here.
This is a problem for the `FsRepository` which will run a `mkdir -p`
to create its own path in the constructor which becomes quite slow for
large repositories. Obvious fix by adding dedicated methods for the
paths.

relates #77466


tlrx (approved):
LGTM

original-brownbear:
Thanks Tanguy!

fcofdez:
LGTM 2 👍

PR #89516, opened 2022-08-22
Synthetic _source: support match_only_text

nik9000:
This adds support for synthetic `_source` to the `match_only_text` field
type. When synthetic `_source` is enabled `match_only_text` fields
create a hidden stored field to contain their text. This should have
similar or better search performance for this specific field type,
though it will have slightly worse indexing performance because
synthetic `_source` is still writing `_recovery_source`, which means
we're writing the bits for this field twice.


modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/MatchOnlyTextFieldMapper.java
```diff
@@ -168,6 +185,35 @@ private Function<LeafReaderContext, CheckedIntFunction<List<Object>, IOException
...
+                            public void stringField(FieldInfo fieldInfo, String value) {
+                                assert fieldInfo.name.equals(name);
+                                values.add(value);
+                            }
+                        });
+                        return values;
+                    } catch (IOException e) {
+                        throw new UncheckedIOException(e);
+                    }
+                };
```
nik9000, line 211: There's a stored field lookup thing in `searchExecutionContext.lookup()` but it can't be convinced to load the hidden stored field. If we feel strongly about it I can try and integrate into it, but I'm not super sure how at the moment.

romseygeek: Yeah that's strictly for scripts and is integrated poorly with other stored field lookup stuff, so I don't think it's worth trying to re-use it for the moment.  I *do* think that the document lookup API I'm playing with at the moment will improve this though.

modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/MatchOnlyTextFieldMapper.java
```diff
@@ -326,6 +371,10 @@ protected void parseCreateField(DocumentParserContext context) throws IOExceptio
         Field field = new Field(fieldType().name(), value, fieldType);
         context.doc().add(field);
         context.addToFieldNames(fieldType().name());
+
+        if (context.isSyntheticSource()) {
+            context.doc().add(new Field(fieldType().originalFieldName(), value, ORIGINAL_FIELD_TYPE));
```
romseygeek, line 376: Use `StoredField` rather than creating a new field type

modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/MatchOnlyTextFieldMapper.java
```diff
@@ -279,6 +321,9 @@ public IndexFieldData.Builder fielddataBuilder(FieldDataContext fieldDataContext
             throw new IllegalArgumentException(CONTENT_TYPE + " fields do not support sorting and aggregations");
         }
 
+        private String originalFieldName() {
```
romseygeek, line 324: Maybe call this `storedFieldName` to make it a bit clearer what it's used for?

modules/mapper-extras/src/test/java/org/elasticsearch/index/mapper/extras/MatchOnlyTextFieldMapperTests.java
```diff
@@ -65,6 +91,13 @@ protected void registerParameters(ParameterChecker checker) throws IOException {
         );
     }
 
+    private XContentBuilder testMapping() throws IOException {
+        if (randomBoolean()) {
+            return syntheticSourceMapping(b -> b.startObject("field").field("type", "match_only_text").endObject());
+        }
+        return fieldMapping(b -> b.field("type", "match_only_text"));
+    }
+
```
romseygeek, line 100: I think I'd be happier with coverage here if we explicitly run both tests for the synthetic and non-synthetic case, this feels at the moment like we're only testing 50% of the functionality on each run and I think that will come back to bite us.

romseygeek (commented):
This looks great! I think it may be worth waiting for a few days to see if I can get document loaders working though as it will tidy up the impl a fair amount.

modules/mapper-extras/src/test/java/org/elasticsearch/index/mapper/extras/MatchOnlyTextFieldMapperTests.java
nik9000: It has a lovely caching behavior that I kind of miss here. Like, if two queries need this then it'll cache.

modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/MatchOnlyTextFieldMapper.java
nik9000: searchExecutionContext.lookup()

nik9000: It has a lovely caching mechanism that I think could be quite nice. If multiple queries need to recheck the source it'll load once while this won't.

nik9000:
> This looks great! I think it may be worth waiting for a few days to see if I can get document loaders working though as it will tidy up the impl a fair amount.

:+1:

modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/MatchOnlyTextFieldMapper.java
nik9000: :+1:

nik9000: :+1:

modules/mapper-extras/src/main/java/org/elasticsearch/index/mapper/extras/MatchOnlyTextFieldMapper.java
romseygeek: So, somewhat annoyingly, I don't think we will be able to re-use a stored field loader from the SearchLookup here because the underlying API expects a LeafReaderContext, not a LeafSearchLookup.  But you should be able to use a LeafStoredFieldLoader rather than a FieldVisitor here which will at least be more readable.

nik9000:
@romseygeek this should be ready for you too!

romseygeek (approved):
LGTM

nik9000:
I almost clicked the merge button! It wouldn't have compiled! sneaky sneaky.

PR #89524, opened 2022-08-22
Simplify and improve changelog YAML file validation

mark-vieira:
This pull request makes some improvements as well as simplifications to our changelog validation:

1. We've removed the Java-based changelog validation task. We now use JSON Schema for _all_ validation, including conditionally required fields and checking for triple-backtick code blocks.
2. Exception messages thrown by schema validation task now comply with Gradle Enterprise [verification failure conventions](https://docs.gradle.com/enterprise/failure-classification/) meaning that these types of failures will be correctly categorized as "verification" failures.


breskeby (approved):
looks good to me. That list of verification compliant task error messages are odd. even compilation errors are considered a verification failure. 🤔

mark-vieira:
> looks good to me. That list of verification compliant task error messages are odd. even compilation errors are considered a verification failure. 🤔

That's by design. Compiling is "validation" of sorts. A failure to compile bad code is an "expected" error. Compare that with "unexpected" errors, like failing to download a dependency, write a file, out of memory, etc.

Basically the intention is to separate failures due to bad inputs vs failures due to bad build code or environment. From our end we look at non-verification errors to see what is wrong with the _build_. Verification errors indicate what is wrong with the _code_.

Issue #71378, opened 2021-04-06
Replace existing identity provider test fixtures with ones that support aarch64


mark-vieira:
The `:x-pack:test:idp-fixture` project relies on a couple of Docker images that only support amd64 architecture, and therefore cannot be run on ARM. Specifically the `unicon/shibboleth-idp` and `c2id/c2id-server` images. We should look into finding alternatives here that support aarch64 so we aren't losing coverage on that platform.

jakelandis:
I reached out to unicorn about updating their shibboleth-idp image and pretty confident they will not be updating their docker image. The good news is that the docker configuration is open source and there are a few forks, so we don't need to start from scratch to build an arm64 compatible image. The bad news is none of the images work out of the box so additional effort is needed for us to create an image, support arm64, and then update to the latest version. 


Some issues:

The shibboleth nexus repo seems to have issue, but this link seems to have similar (same ?) downloads
https://build.shibboleth.net/maven/releases

https://shibboleth.net/downloads/identity-provider/$idp_version appears to be missing, but it is availble at https://shibboleth.net/downloads/identity-provider/archive/archive/$idp_version

`http://central.maven.org` is no more and replaced with `https://repo.maven.apache.org`

Simply bumping up jetty versions does not seem to work, but https://github.com/winstonhong/Shibboleth-SAML-IdP-and-SP/blob/master/shibboleth-idp-dockerized/Dockerfile is an example of a newer jetty version

Simply bumping the idp version in the Dockerfile does not seem to work but https://github.com/CSCfi/shibboleth-idp-dockerized/blob/master/Dockerfile has an example of a newer (4.x) idp

https://github.com/winstonhong/Shibboleth-SAML-IdP-and-SP/blob/master/shibboleth-idp-dockerized/ is the closest to working out of the box, but our tests (./gradlew :x-pack:qa:saml-idp-tests:javaRestTest --info) fail during the SSL handshake.

We can build from source located here : git clone https://git.shibboleth.net/git/java-identity-provider.git with Java11 and newer version of maven. However, we shouldn't need to due to https://shibboleth.net/downloads/identity-provider/archive/archive/$idp_version

mark-vieira:
> The bad news is none of the images work out of the box so additional effort is needed for us to create an image, support arm64

[...]

PR #89526, opened 2022-08-23
Update OIDC test fixture

### LINKED ISSUE #89477: OpenID Connect integration tests failing due to missing fixture Docker image
It seems the `c2id/c2id-server` docker image has dissappeared from DockerHub so our idp-fixture that depends on it can no longer start. This is currently only required by our `oidc-op-tests` project so for now I've disabled these tests. We'll need to find an alternative fixture here. I tried simply replacing it with the `c2id-server-demo` image but it wasn't a direct replacement.

https://gradle-enterprise.elastic.co/s/p4vwhzkudvutw

jakelandis:
This commit updates the OIDC test fixture to use the latest demo docker image
from https://connect2id.com.  As part of this change some of the docker image demo 
login has changed from prior versions and the test code is updated accordingly. 

fixes #89477
reverts #89478 / e949dff.

---
To figure out the required changes I setup a local proxy to watch traffic and manually executed some of the workflows and adjusted the test setup accordingly. 
When run as a suite of tests depending on order some tests would fail... I found that setting op.authz.alwaysPromptForConsent=true and  op.authz.alwaysPromptForAuth=true allows for the tests to pass consistently. 

Buildscan with only these tests: https://gradle-enterprise.elastic.co/s/xnysxq5b2m6za


jkakavas (approved):
LGTM

mark-vieira:
Thanks, Jake! It's worth noting that [this issue](https://github.com/elastic/elasticsearch/issues/71378) still exists since we still aren't using a docker image that provides an `aarch64` build.

PR #89529, opened 2022-08-23
Move bulk action single item wrapper util

n1v0lg:
This PR moves the wrapper method for translating from a bulk response
to a single item response from the deprecated
`TransportSingleItemBulkWriteAction` class to a non-deprecated class.
This method is still useful and lets us avoid boiler plate code.


server/src/main/java/org/elasticsearch/action/bulk/TransportBulkAction.java
```diff
@@ -169,6 +173,32 @@ public static IndexRequest getIndexWriteRequest(DocWriteRequest<?> docWriteReque
         return indexRequest;
     }
 
+    public static BulkRequest wrapSingleRequest(final ReplicatedWriteRequest<?> request) {
```
n1v0lg, line 176: Also good with moving it to `BulkRequest` -- leaving it here to keep it together with `toSingleResponse`

server/src/main/java/org/elasticsearch/action/bulk/BulkRequest.java
```diff
@@ -92,6 +94,20 @@ public BulkRequest(@Nullable String globalIndex) {
         this.globalIndex = globalIndex;
     }
 
+    /**
+     * See {@link TransportBulkAction#wrapBulkAsSingleItemResponse(ActionListener)} for wrapping translation of a bulk response
+     * to a single item response.
+     */
+    public static BulkRequest fromSingleRequest(final ReplicatedWriteRequest<?> request) {
```
n1v0lg, line 101: [`wrapBulkAsSingleItemResponse`](https://github.com/elastic/elasticsearch/pull/89529/files#diff-572344b1e2a22818057c05a76eecc4530d7231b347d6ff1b5f3b1686222ce5d8R178) is primarily the util method I want to port in this PR but `fromSingleRequest` (previously [`toSingleItemBulkRequest`](https://github.com/elastic/elasticsearch/blob/be7c7415627377a1b795400fb8dfcc6cbdf0e322/server/src/main/java/org/elasticsearch/action/bulk/TransportSingleItemBulkWriteAction.java#L63)) also seems useful.

I'm not sure if the intention of deprecating `TransportSingleItemBulkWriteAction` was to also deprecate `toSingleItemBulkRequest`. If that's the case I'm happy to leave it out of this PR. 

If we do want to keep it, I can also add support for `UpdateRequest` since `fromSingleRequest` currently only supports Index and Delete.

ywangd: I don't have strong opinion. Using BulkRequestBuilder is pretty easy to create bulk request with a single item, including update request. So I don't it's really necessary to add support for `UpdateRequest`. Less code to maintain is better.

server/src/main/java/org/elasticsearch/action/update/TransportUpdateAction.java
```diff
@@ -191,21 +191,59 @@ protected void shardOperation(final UpdateRequest request, final ActionListener<
...
-                        response.getResult()
-                    );
-                    if (request.fetchSource() != null && request.fetchSource().fetchSource()) {
-                        Tuple<XContentType, Map<String, Object>> sourceAndContent = XContentHelper.convertToMap(
-                            upsertSourceBytes,
-                            true,
-                            upsertRequest.getContentType()
+                client.bulk(
+                    BulkRequest.fromSingleRequest(upsertRequest),
+                    TransportBulkAction.wrapBulkAsSingleItemResponse(ActionListener.<IndexResponse>wrap(response -> {
```
ywangd, line 196: Can we have these methods as unqualified static imports like the existing code does? I don't have strong preference for this style. But in this particular case, it should cause less cascading style changes and be easier for review.

ywangd (commented):
I believe its a pure refactor. But would appreciate if you could address the unqualified method imports and I'll just need another quick view afterwards. Thanks!

server/src/main/java/org/elasticsearch/action/update/TransportUpdateAction.java
n1v0lg: Yup I can do this. I went with this format because `fromSingleRequest` without the class is pretty awkward but don't feel too strongly.

server/src/main/java/org/elasticsearch/action/bulk/BulkRequest.java
n1v0lg: In the interest of having one way of doing things (i.e., use the request builder), I'll skip porting this method. It's got that side-effect on the passed-in request too which I'm not a fan of.

server/src/main/java/org/elasticsearch/action/update/TransportUpdateAction.java
```diff
@@ -191,46 +191,49 @@ protected void shardOperation(final UpdateRequest request, final ActionListener<
...
-                                request.concreteIndex(),
-                                response.getSeqNo(),
-                                response.getPrimaryTerm(),
-                                response.getVersion(),
-                                sourceAndContent.v2(),
-                                sourceAndContent.v1(),
-                                upsertSourceBytes
-                            )
+                client.bulk(
+                    toSingleItemBulkRequest(upsertRequest),
```
n1v0lg, line 195: This is just reformatting since the name got slightly longer. Now that the method sits with `TransportBulkAction`, I think the extra quantifier `asSingle` is warranted.

n1v0lg:

server/src/main/java/org/elasticsearch/action/update/TransportUpdateAction.java
n1v0lg: 😬 sorry, lines of code length is not on our side so there's still a bit of cascading formatting diff that happens

ywangd: No worries. Thanks for going through the trouble of changing the imports.

ywangd (approved):
LGTM

server/src/main/java/org/elasticsearch/action/bulk/TransportBulkAction.java
```diff
@@ -169,6 +171,22 @@ public static IndexRequest getIndexWriteRequest(DocWriteRequest<?> docWriteReque
         return indexRequest;
     }
 
+    public static <Response extends ReplicationResponse & WriteResponse> ActionListener<BulkResponse> wrapBulkResponseAsSingle(
```
DaveCTurner, line 174: Naming nit: this is more like unwrapping than wrapping? How about something like this?

```suggestion
    public static <Response extends ReplicationResponse & WriteResponse> ActionListener<BulkResponse> unwrappingSingleItemBulkResponse(
```

DaveCTurner (approved):
LGTM, one naming nit

server/src/main/java/org/elasticsearch/action/bulk/TransportBulkAction.java
n1v0lg: I'm good with unwrapping. I went with wrap because under the hood we call `ActionListener.wrap(...)` so it's wrapping in a different sense. 

Is `unwrapAs` instead of `unwrapping` okay with you?

PR #89534, opened 2022-08-23
Fix incorrect failed shards count in APIs for current snapshots

original-brownbear:
The failed shards count approach is wrong, if we have an in-progress snapshot
the difference in counts does not work out to the failed shards count.
Since every failed shard has an entry in the failure list anyway we can just
return its size instead for the correct answer.

### DIFF
--- a/server/src/internalClusterTest/java/org/elasticsearch/snapshots/GetSnapshotsIT.java
+++ b/server/src/internalClusterTest/java/org/elasticsearch/snapshots/GetSnapshotsIT.java
@@ -191,6 +191,13 @@ public class GetSnapshotsIT extends AbstractSnapshotIntegTestCase {
         assertStablePagination(repos, allSnapshotNames, GetSnapshotsRequest.SortBy.START_TIME);
         assertStablePagination(repos, allSnapshotNames, GetSnapshotsRequest.SortBy.NAME);
         assertStablePagination(repos, allSnapshotNames, GetSnapshotsRequest.SortBy.INDICES);
+        final List<SnapshotInfo> currentSnapshots = clusterAdmin().prepareGetSnapshots(matchAllPattern())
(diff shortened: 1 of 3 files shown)


tlrx (approved):
LGTM, good catch

fcofdez (approved):
LGTM 👍

server/src/internalClusterTest/java/org/elasticsearch/snapshots/GetSnapshotsIT.java
```diff
@@ -191,6 +191,13 @@ public void testSortAndPaginateWithInProgress() throws Exception {
         assertStablePagination(repos, allSnapshotNames, GetSnapshotsRequest.SortBy.START_TIME);
         assertStablePagination(repos, allSnapshotNames, GetSnapshotsRequest.SortBy.NAME);
         assertStablePagination(repos, allSnapshotNames, GetSnapshotsRequest.SortBy.INDICES);
+        final List<SnapshotInfo> currentSnapshots = clusterAdmin().prepareGetSnapshots(matchAllPattern())
+            .setSnapshots(GetSnapshotsRequest.CURRENT_SNAPSHOT)
+            .get()
+            .getSnapshots();
+        for (SnapshotInfo currentSnapshot : currentSnapshots) {
+            assertThat(currentSnapshot.toString(), currentSnapshot.failedShards(), is(0));
```
idegtiarenko, line 199: Should the marcher be `equalTo(0)`?

idegtiarenko (approved):

server/src/internalClusterTest/java/org/elasticsearch/snapshots/GetSnapshotsIT.java
original-brownbear: We used `is(` for all other tests here so I guess we have to go with this for consistency

original-brownbear:
Thanks all!

PR #89535, opened 2022-08-23
Update openjdk to 18.0.2.1

### LINKED ISSUE #89531: Upgrade OpenJDK to 18.0.2.1+0
OpenJDK 18.0.2.1+0 is now available. We need to upgrade the version that is bundled with Elasticsearch to the newer version. We currently bundle 18.0.2+9.

breskeby:
closes #89531


build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/JdkDownloadPlugin.java
```diff
@@ -108,7 +108,7 @@ private void setupRepository(Project project, Jdk jdk) {
                     + "/[module]/[classifier]/jdk/hotspot/normal/adoptium";
             }
         } else if (jdk.getVendor().equals(VENDOR_OPENJDK)) {
-            repoUrl = "https://download.oracle.com";
```
breskeby, line 111: How come we used download.oracle.com? this seems not to work (anymore) I switched this to download.java.net

mark-vieira (approved):
LGTM. Is the Oracle link now working for all variants we need?

mark-vieira:
@breskeby let's make sure this gets backported to `8.4` as well.

breskeby:
@mark-vieira 👍

PR #89554, opened 2022-08-23
More tests for enrich processor

nik9000:
Adds more tests for the enrich processor around different index types.
Right now they all work fine (yay!) but this feels like a good amount of
paranoia.


dakrone (approved):
LGTM, thanks for adding these Nik! I left one comment about adding a description string for the test

nik9000:
> These are all missing a human-readable "name" for the test (the `"Test that X does Y"` sort of string under the initial `---`), can we add those so people can see at a glance what the test is testing?

I don't tend to include the words "test that" but I have pushed some more descriptive language. `run` was terrible.

Issue #89283, opened 2022-08-11
Better rollover handling of small and empty indices


joegallo:
This meta issue tracks some enhancements and other changes around handling rollovers (especially ILM rollovers) better in the cases of small or empty indices.

- Enhancements
  - [x] https://github.com/elastic/elasticsearch/pull/83345
  - [x] https://github.com/elastic/elasticsearch/issues/86203

- Related changes
  - [ ] https://github.com/elastic/kibana/issues/137464
  - [ ] https://github.com/elastic/elasticsearch/issues/89284
  - [ ] Evaluate adding a min_* condition by default to stack-included ILM policies (e.g. `ilm-history-ilm-policy`, `logs`, as well as the policies installed by APM or Kibana, etc.)
  - [ ] https://github.com/elastic/elasticsearch/issues/83039 and just a little more broadly, documenting the idea of how to retire an index alias or datastream around an ILM policy

- Tech Debt
  - [ ] https://github.com/elastic/elasticsearch/issues/88018
  - [ ] https://github.com/elastic/elasticsearch/issues/89282

joegallo:
re: https://github.com/elastic/elasticsearch/issues/83039 (that we don't have an especially great way to 'retire' an alias or datastream that's not going to be written to anymore):

At the moment we have some error conditional handling around situations where the index in question is unexpectedly marked with `index.lifecycle.indexing_complete` set to `true`:

https://github.com/elastic/elasticsearch/blob/1a3032beb6790805b2bba01180f9b875ac2fa971/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/WaitForRolloverReadyStep.java#L130-L168

A potentially interesting implementation of alias/datastream 'retirement' would be to revise that logic a bit so that we could merely mark an index as finished by setting `index.lifecycle.indexing_complete` to `true` and then having ILM continue past `check-rollover-ready` and continue on through the rest of the lifecycle. In order to make this work, though, we'd need to juggle `is_write_index` when we see this (since the rollover itself wouldn't handle that anymore). Taking that a step further, for datastreams that would open up the possibility that a datastream would have no current write index, so a nearby or parallel effort to this would be having the ability to make a datastream read only.

joegallo:
#100439 seems pretty closely related to this.

joegallo:
re: https://github.com/elastic/elasticsearch/issues/89283#issuecomment-1277738200:

[...]

PR #89557, opened 2022-08-23
ILM don't rollover empty indices

### LINKED ISSUE #86203: Prevent ILM from spuriously rolling over (many) empty indices
### Description

This has been brought up a few times in different forms, see https://github.com/elastic/elasticsearch/issues/46161, https://github.com/elastic/elasticsearch/issues/73349, https://github.com/elastic/elasticsearch/issues/83039, https://github.com/elastic/elasticsearch/issues/85054.

Any ILM policy with a `max_age` associated with the `rollover` action could trigger this scenario, but in order to talk about something concrete, I'll use metricbeat as an example (double emphasizing, though, this isn't unique to metricbeat, it's just the nature of the way `rollover` currently works with a `max_age`).

With a test 8.1.3 Elasticsearch cluster, I ran `metricbeat-8.1.2` for a few seconds and then stopped it, and then `metricbeat-8.1.3` for a bit longer. The default `metricbeat` policy has rollover with `"max_age" : "30d"` (30 days) but in order to illustrate this problem better, I've set that to `"1m"` (1 minute) instead:

```
PUT /_cluster/settings
{
  "persistent": {
    "indices.lifecycle.poll_interval": "5s"
  }
}

PUT _ilm/policy/metricbeat
{
  "policy" : {
    "phases" : {
      "hot" : {
        "min_age" : "0ms",
        "actions" : {
          "rollover" : {
            "max_size" : "50gb",
            "max_age" : "1m"
          }
        }
      }
    }
  }
}
```

After a few minutes, my cluster looks like this:

```
GET _cat/indices/.ds-metricbeat-*?s=index
yellow open .ds-metricbeat-8.1.2-2022.04.26-000001 GBqDAprYSl2NmFzi81n9Ug 1 1 1134 0 652.2kb 652.2kb
yellow open .ds-metricbeat-8.1.2-2022.04.26-000002 Ybd4SCiWT0-7W0v9zKPR4A 1 1    0 0    225b    225b
yellow open .ds-metricbeat-8.1.2-2022.04.26-000003 3_9OmFkOSKaEfF_J-_D9TA 1 1    0 0    225b    225b
yellow open .ds-metricbeat-8.1.2-2022.04.26-000004 4olQItwcTtCOWrBotcqoLw 1 1    0 0    225b    225b
yellow open .ds-metricbeat-8.1.2-2022.04.26-000005 N9_gYkcORWSVfacUwnDegw 1 1    0 0    225b    225b
yellow open .ds-metricbeat-8.1.3-2022.04.26-000001 kWW-N_bfRbO0vMR4z3F72g 1 1  862 0 639.2kb 639.2kb
yellow open .ds-metricbeat-8.1.3-2022.04.26-000002 qzu-L-zZQqqm-6GQAZMtgA 1 1  235 0 431.2kb 431.2kb
yellow open .ds-metricbeat-8.1.3-2022.04.26-000003 iW68NzFyTv-CCAg3Rsfj4A 1 1  265 0   494kb   494kb
yellow open .ds-metricbeat-8.1.3-2022.04.26-000004 NaIa-gUjShKpEHzNcAxA3w 1 1  234 0 451.7kb 451.7kb
yellow open .ds-metricbeat-8.1.3-2022.04.26-000005 lDhzxqtdR8miPnqvsf7HDQ 1 1  271 0 595.8kb 595.8kb
```

That is, for a little while, the first writer (metricbeat version 8.1.2) wrote documents, and then it stopped and was upgraded and replaced by the second writer (metricbeat version 8.1.3). Each of those writers uses a versioned datastream (`metricbeat-8.1.2` and `metricbeat-8.1.3` respectively).

The problem is easy to see -- notice that we're getting a new empty (0 document) `.ds-metricbeat-8.1.2-[...]` index every minute, and that we'll keep accumulating them forever. ILM doesn't have any special logic around empty indices like this, i.e. empty indices are treated the same as non-empty indices as far as ILM is concerned.

In this simple scenario, we know that the `metricbeat-8.1.2` datastream is done now, and can be retired. However, there's no particular point in time where Elasticsearch itself or some individual metricbeat process could know that. I'm using just one metricbeat writer, but I could be running one on each of N hosts. No one writer process in this scenario knows that it is special and should "turn off the lights when it's done".

To further complicate matters, maybe I have a weekly batch process which will run on Sunday evening and write some logs after a long quiet period (and its logs are still being monitored by metricbeat version 8.1.2)-- when it does so we could end up with more data flowing into the current `metricbeat-8.1.2` write index. Let's call that the "sporadic writer" case. In that case, we'd end up with periods of no data flowing in and the accumulation of empty indices, followed by one or more non-empty indices, and then back to accumulating empty indices again.

ILM doesn't know whether there's a sporadic writer out there or not, and ignorant of whether more documents will be coming one day, it dutifully executes the policy, rolling over the now defunct `metricbeat-8.1.2` datastream every minute and leaving a trail of empty `.ds-metricbeat-8.1.2-[...]` indices in its wake.

An additional note: my illustration here is datastream specific, but in the broad strokes this issue could also exist in a pre-datastream indexing strategy built around aliases. It would be most excellent if we were able to solve both the datastream and alias -based versions of this empty index problem (but reserving a degree of freedom, I don't think the solution must necessarily be precisely the same in both cases).

joegallo:
Closes #86203, related to #89283

ILM will no longer roll over empty indices by default. This is accomplished by injecting a `min_docs: 1` condition into the rollover dry run request in `WaitForRolloverReadyStep`. 

Such a condition will only be injected if there's not already a `min_docs` (or `min_primary_shard_docs`) condition associated with the rollover, so a policy can opt out of the new behavior by adding an explicit `min_docs: 0` condition to the rollover action.

This can also be disabled on a cluster-wide basis by setting `indices.lifecycle.rollover.only_if_has_documents` to `false`.


x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/WaitForRolloverReadyStep.java
```diff
@@ -196,6 +196,24 @@ public void evaluateCondition(Metadata metadata, Index index, Listener listener,
...
+            .rolloverIndex(
+                rolloverRequest,
+                ActionListener.wrap(
+                    response -> listener.onResponse(rolloverRequest.areConditionsMet(response.getConditionStatus()), EmptyInfo.INSTANCE),
+                    listener::onFailure
+                )
+            );
+    }
+
+    // visible for testing
```
dakrone, line 215: I'd be even better if this had javadoc, even though it's essentially private :)

x-pack/plugin/ilm/qa/multi-node/src/javaRestTest/java/org/elasticsearch/xpack/ilm/actions/RolloverActionIT.java
```diff
@@ -193,14 +195,105 @@ public void testRolloverActionWithMaxPrimaryDocsSize() throws Exception {
...
+        }
+
+        // because the index is empty, it doesn't roll over
+        assertBusy(() -> {
+            assertThat(getStepKeyForIndex(client(), originalIndex).getName(), is(WaitForRolloverReadyStep.NAME));
+            assertFalse(indexExists(secondIndex));
+            assertTrue(indexExists(originalIndex));
+        }, 30, TimeUnit.SECONDS);
+
+        switch (between(0, 2)) {
```
dakrone, line 233: I think rather than randomize between the three potential ways to force a rollover, I'd rather explicitly do each of them. This would allow us to catch issues better in the future that might only break one of the branches since we don't test each method every CI run.

dakrone (approved):
LGTM, I left two very minor reviews, thanks for adding this, I think it'll help a lot of people!

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/WaitForRolloverReadyStep.java
joegallo: Done, b4f780acd3d9aff427412e136e3cfdf3f54137fc

x-pack/plugin/ilm/qa/multi-node/src/javaRestTest/java/org/elasticsearch/xpack/ilm/actions/RolloverActionIT.java
joegallo: How does 1ea80c1bbcadf016801a9d09a5cb9b50f49b1e12 work for you, @dakrone?

dakrone: Works for me, 🚢

PR #89563, opened 2022-08-23
Better support for multiple nodes via ./gradlew run 

jakelandis:
This commit makes minor changes to support multiple nodes
via ./gradlew run. The specific desired use case, where there 2 nodes 
in the cluster where one is a coordinating only node is used as 
the example configuration.


build-tools/src/main/java/org/elasticsearch/gradle/testclusters/ElasticsearchCluster.java
```diff
@@ -120,7 +120,7 @@ public void setNumberOfNodes(int numberOfNodes) {
             throw new IllegalArgumentException("Number of nodes should be >= 1 but was " + numberOfNodes + " for " + this);
         }
 
-        if (numberOfNodes <= nodes.size()) {
+        if ((numberOfNodes != 1 && nodes.size() != 1) && numberOfNodes <= nodes.size()) {
```
jakelandis, line 123: this is here to allow advertising `numberOfNodes = 1` in the default configuration

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/ElasticsearchNode.java
```diff
@@ -113,7 +113,8 @@ public class ElasticsearchNode implements TestClusterConfiguration {
     private static final List<String> OVERRIDABLE_SETTINGS = Arrays.asList(
         "path.repo",
         "discovery.seed_providers",
-        "cluster.deprecation_indexing.enabled"
+        "cluster.deprecation_indexing.enabled",
+        "cluster.initial_master_nodes"
```
jakelandis, line 117: this is here since if we change the master node configuration we need to update the cluster.initial_master_nodes

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/RunTask.java
```diff
@@ -139,12 +140,12 @@ public void beforeStart() {
...
-            cluster.getFirstNode().setHttpPort(String.valueOf(httpPort));
-            httpPort++;
-            cluster.getFirstNode().setTransportPort(String.valueOf(transportPort));
-            transportPort++;
             cluster.setPreserveDataDir(preserveData);
             for (ElasticsearchNode node : cluster.getNodes()) {
+                node.setHttpPort(String.valueOf(httpPort));
+                httpPort++;
+                node.setTransportPort(String.valueOf(transportPort));
+                transportPort++;
```
jakelandis, line 148: this is here to ensure consistent port numbering (it worked before but picked a random port for the subsequent nodes)

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/RunTask.java
```diff
@@ -165,6 +165,7 @@ public void runAndWait() throws IOException {
         if (initOnly) {
             return;
         }
+        getClusters().forEach(ElasticsearchCluster::waitForAllConditions);
```
jakelandis, line 168: this is the actual fix to support multi-node via run. This causes the ports files to be written out to disk that is needed to form the cluster.

TESTING.asciidoc
```diff
@@ -142,6 +142,47 @@ Now the http.p12 file will be placed in the config directory of the running clus
...
+Finally, we need to "restart" the cluster to pick up the changes. However, it does not actually start/stop/start
+the cluster since we have not started the cluster yet and this instructs the plugin to update the configuration.
+
+-------------------------------------
+tasks.register("run", RunTask) {
+  useCluster testClusters.named("runTask")
+  description = 'Runs elasticsearch in the foreground'
+  group = 'Verification'
+
+  doFirst {
```
mark-vieira, line 170: Why would we do this here rather than just configure the test cluster this way to begin with up above?

TESTING.asciidoc
```diff
@@ -142,6 +142,47 @@ Now the http.p12 file will be placed in the config directory of the running clus
...
+    def cluster = testClusters.named("runTask").get()
+    cluster.getNodes().each { node ->
+      node.setting('cluster.initial_master_nodes', cluster.getLastNode().getName())
+      node.setting('node.roles', '[master,data_hot,data_content]')
+    }
+    cluster.getFirstNode().setting('node.roles', '[]')
+    cluster.restart()
+  }
+  impliesSubProjects = true
```
mark-vieira, line 179: This setting is not required. This is just something to that running `./gradlew run` works from the CLI w/o specifyign the whole task path.

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/ElasticsearchCluster.java
mark-vieira: I don't follow what is meant by "advertising" here. Isn't a cluster of 1 node the default. Why would it even be necessary to explicitly specify a size of `1`, and even so, what would be the problem with that w/o this change?

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/ElasticsearchNode.java
mark-vieira: Is this something we intend folks to do? How is it that none of our existing tests require such a setting? What scenario requires setting this setting that is not required by any of our existing test suites?

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/RunTask.java
mark-vieira: We can simplify this to just `node.setHttpPort(String.valueOf(httpPort++));`. Also, I wonder if it's worth detecting whether we are running multiple nodes, and then printing this information to the user. Otherwise it's quite hard to know what the ports for each of these cluster/nodes actually are.

mark-vieira (commented):
I'm confused why we don't use the same approach as for CCS, where you simply configure a test cluster with however many nodes and with whatever config you want, then just instruct the runtask to use that cluster?

jakelandis:
> I'm confused why we don't use the same approach as for CCS

Indeed. I noticed that too when working on this and am testing a change to make run-ccs look like this and will likely submit a PR to use this approach for run-css. The missing piece was the cluster.restart() to update the configuration before start.

TESTING.asciidoc
jakelandis: I wondered about that setting , will remove it from the example (and replace the other non-relevant parts with `...` like in the other example.

TESTING.asciidoc
jakelandis: If i understand correctly the testCluster is not realized yet  (i.e. only a provider and only registered) and wasn't sure how to get a reference to each of the nodes in the above task. By placing it here and I am realizing the cluster in this task and re-configuring it a bit before but before it is started.

mark-vieira: You would configure it in exactly the same way you set the number of nodes. Any and all configuration can be done there, it's simply deferred until the cluster is actually required.

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/ElasticsearchCluster.java
jakelandis: > Isn't a cluster of 1 node the default.
yes
> Why would it even be necessary to explicitly specify a size of 1
It is not necessary, but allows me to define `numberOfNodes = 1` in the task definition to "advertise" you can have a more than one node by changing this value.  It is purely cosmetic and can remove this and the explicit definition of  `numberOfNodes = 1`

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/ElasticsearchNode.java
jakelandis: > Is this something we intend folks to do?

For this specific use case of manually debugging - yes. It allows you to _not_ have the master node role on every node. 

> How is it that none of our existing tests require such a setting?

I believe most tests that require fine control over node roles (specifically which node is master) is handled via other types of tests. 

> What scenario requires setting this setting that is not required by any of our existing test suites?

Starting a new cluster with multiple nodes where one of the nodes is not a master node.

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/RunTask.java
jakelandis: I can simplify the increment.  The information is printing in the server logs and is pretty guessable, 9200/9300 then 9201/9301 ...

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/ElasticsearchCluster.java
mark-vieira: I think we could just fix this by changing the conditional to `numberOfNodes < nodes.size()`. I'm not sure why we would fail when we set the number of nodes to the existing size anyway. We just want to protect against folks inadvertently trying to shrink a cluster.

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/RunTask.java
mark-vieira: Fair enough. I just figured since it's getting spammed all together it might be hard to grok, we can add that later if folks find this cumbersome.

TESTING.asciidoc
jakelandis: maybe I am not understanding but putting this config as-is just below numberOfNodes in the above task registration results in `Could not create domain object 'runTask' (ElasticsearchCluster)` https://gradle-enterprise.elastic.co/s/xjmtjfgceprli similar results wrapped in a doLast. 

I did find some other examples and the following would work inside the testCluster.register("runTask")
```
      nodes."runTask-0".setting('node.roles', '[]')
      nodes."runTask-1".setting('node.roles', '[master,data_hot,data_content]')
      nodes."runTask-0".setting('cluster.initial_master_nodes', 'runTask-1')
      nodes."runTask-1".setting('cluster.initial_master_nodes', 'runTask-1')
```
I do slightly prefer the example in the PR, especially one i redo the run-ccs to work like that.

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/ElasticsearchNode.java
jakelandis: ...actually i did find one example, where I think it should require this (but didn't)  --> x-pack/qa/mixed-tier-cluster/build.gradle . I will try to track down how that is working, maybe i am missing something.

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/ElasticsearchCluster.java
jakelandis: done [`dbda577` (#89563)](https://github.com/elastic/elasticsearch/pull/89563/commits/dbda577de2b43a4feb465b5397bcebe19cb103e5)

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/RunTask.java
jakelandis: incrementor simplified: [`e4649e5` (#89563)](https://github.com/elastic/elasticsearch/pull/89563/commits/e4649e50746bc28ceaecaba0f5cc71938f3e7157)

TESTING.asciidoc
jakelandis: removed from example: [`e4649e5` (#89563)](https://github.com/elastic/elasticsearch/pull/89563/commits/e4649e50746bc28ceaecaba0f5cc71938f3e7157)

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/ElasticsearchNode.java
jakelandis: OK.. I think I figured out why that case worked (./gradlew :x-pack:qa:mixed-tier-cluster:bwcTest)  and my manual testing hung when trying to start.  In my example, I have 2 nodes (1 master) but in the example there are 3 nodes (2 master). So with 2/3 masters it can form a quorum, but not with 1/2. If I manually test with 2/3 masters i do not need to override `cluster.initial_master_nodes`.  When it hangs it is pretty hard to understand why and killing the gradle task leaves the ES processes running which then causes more hard to diagnose failures since the process occupies the ports.  I would prefer to leave this as-is if possible and keep it's usage in the example to avoid having to troubleshoot this scenario.

TESTING.asciidoc
jakelandis: For context [here is my WIP](https://github.com/elastic/elasticsearch/compare/main...jakelandis%3Are_do_run_ccs?body=&expand=1) for re-doing run-css.

jakelandis:
@mark-vieira - ready for another round of review/discussion. thanks !

jakelandis:

TESTING.asciidoc
mark-vieira: That error is because you are trying to call `restart()` on a cluster that hasn't started. Since we are configuring everything before the cluster has started, there's no need for a restart at all in this scenario so you should just be able to remote that line.

jakelandis: Thanks !  I like that much better. I updated the doc via [`9b031d4` (#89563)](https://github.com/elastic/elasticsearch/pull/89563/commits/9b031d42536eb624380f28315e282ccb9ae12960)

jakelandis:

jakelandis:
@elasticmachine update branch

build-tools/src/main/java/org/elasticsearch/gradle/testclusters/RunTask.java
mark-vieira: This might be problematic if something explodes in startup, because this call blocks, we won't get any log output until the cluster reports green, which might be never. Should we move this down below?

jakelandis: good catch.  All I really need is that call to update the ports information. I removed this blocking line and just called the method to write out the ports info. [`23b4be1` (#89563)](https://github.com/elastic/elasticsearch/pull/89563/commits/23b4be1305be16f7c5b4f1db7b24e2b47f0ad4a0)

mark-vieira (approved):

PR #89565, opened 2022-08-23
Don't try to disable ML on incompatible versions

mark-vieira:
Follow up to #89517 to ensure we don't mistakenly pass the `xpack.ml.enabled` flag on versions of ES that have no such setting.


mark-vieira:

mark-vieira:

PR #89570, opened 2022-08-24
Option to return profile uid in GetUser response

ywangd:
This PR adds a new query parameter to the GetUser API to optionally return
profile uid for each user if there is an associated profile.

PS: Doc change will be a separate PR.


ywangd:
Note to reviewers:

1. GetUser API works even when the realm is disabled. As such, profile uid retrieval also works when the realm is disabled. The challenge here is that a disabled realm is _not_ part of the `activeRealms`. It is not even tracked anywhere (other than settings themselves). So it is difficult to build a `RealmRef` for it. To tackle this problem, this PR updates the `Realms` class to track all realmRefs (explicitly or implicitly configured regardless whether it is disabled). 
2. The `User` class is left untouched, i.e. it does not contain a reference to profile uid. The association is done at response level. I believe this is the right choice. `User` is really a security concept while profile is not. It is not feasible to always populate profile uid for a `User` object. Vaving a field that is not always populated is not great. Addtionally, the meaning of a `null` profile_uid can be either "not requested" or "not exists". Chaging the user class also has lots of cascading changes. Overall, it's better to not have it for `User`.
3. This PR choses to resolve profile uid at API call time. One alternative is to save this uid along with the user document on profile activation. It may seem to be more efficient for the GetUser API. But it has major downsides: This uid becomes 2nd source of truth (in addition to resolve it dynamically) and may not always be up to date. For example, if native realm is removed from a domain, the profile uid may become incorrect because the native user might get a new profile. 
4. The Anonymous user can also have profile. As such the profile uid will be returned for it as well.
5. The profile uid will be returned even when the profile is disabled. I think this makes sense because the admin user can still perform management work for disabled profile, including re-enable them (we don't support it today, but it seems reasonable to expect).

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportGetUsersAction.java
```diff
@@ -97,4 +131,35 @@ protected void doExecute(Task task, final GetUsersRequest request, final ActionL
...
+                    .collect(Collectors.toUnmodifiableMap(Tuple::v1, Tuple::v2));
+                listener.onResponse(profileUidLookup);
+            } else {
+                final ElasticsearchStatusException exception = new ElasticsearchStatusException(
+                    "failed to retrieve profile for users. please retry without fetching profile uid",
+                    RestStatus.BAD_REQUEST
+                );
+                resultsAndErrors.errors().values().forEach(exception::addSuppressed);
+                listener.onFailure(exception);
+            }
```
ywangd, line 162: The API call throws error entirely even when profile uid retrieval errors for a single user. We can definitely be more subtle on error handling. But I feel being strict is a safe start.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/Realms.java
```diff
@@ -100,8 +104,11 @@ public Realms(
             .distinct()
             .collect(Collectors.toMap(DomainConfig::name, Function.identity()));
         final List<RealmConfig> realmConfigs = buildRealmConfigs();
+
+        // initRealms will add default file and native realm config if they are not explicitly configured
         final List<Realm> initialRealms = initRealms(realmConfigs);
-        configureRealmRef(initialRealms, realmConfigs, realmToDomainConfig);
+        realmRefs = calculateRealmRefs(realmConfigs, realmToDomainConfig);
+        initialRealms.forEach(realm -> realm.initRealmRef(realmRefs.get(new RealmConfig.RealmIdentifier(realm.type(), realm.name()))));
```
albertzaharovits, line 111: nit: `initialRealms.forEach(realm -> realm.initRealmRef(realmRefs))` , I prefer it because the current code is verbose and I feel it adds to the confusion.

albertzaharovits (approved):
LGTM
Like we discussed, I think the overall strategy to keep `User` and `Profile` separate and combine them in the "get user" response, is the best.
But, implementation wise, I think this does too much in the transport action handler, particularly translating the `User` to `Subject` is a low level thing that I would've burried at lower abstraction levels (maybe in the `ProfileService`).
I'm glad to take another look if you find better ways to hide it, but this is a minor design gripe so I don't think it's required for me to do so.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authc/Realms.java
ywangd: Updated as suggested. Thanks!

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportGetUsersAction.java
ywangd: 500 makes sense. Updated as suggested.

ywangd:
@albertzaharovits Thanks for the review!

> implementation wise, I think this does too much in the transport action handler, particularly translating the User to Subject is a low level thing that I would've burried at lower abstraction levels (maybe in the ProfileService).

I considered this but decided to leave it in the transport handler. The main issue is that the conversation from `User` to `Subject` is rather _not_ generic and it is more error prone if we move it elsewhere. 

Take `profileService#searchProfilesForInternalUsers(User, Listener)` as you suggested for an example. The `User` object has no inherent concept about being "internal". It is impossible to look at just the `User` object and tell which `realm` it is from and convert it to a `Subject`. Therefore, even though our intention is that only reserved and native user objects should be called with this method, there is no way we can enforce that which feels risky. For the current PR, the transport handler is where both user and realm information can be enforced because the handler only deals with reserved/native users and can tell them apart and convert them to Subject reliably. Therefore, I'll leave it with the transport handler. I hope this is OK with you.

PS: In long term, I suspect most methods should be interfaced with `Subject` instead of `User`. The `User` object should mostly be used within a class but not cross class because it is not sufficient to make many decisions. As such, I suspect we probably do not want to have ProfileService to deal with `User` directly.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportGetUsersAction.java
```diff
@@ -97,4 +131,35 @@ protected void doExecute(Task task, final GetUsersRequest request, final ActionL
...
+    private void resolveProfileUids(List<User> users, ActionListener<Map<String, String>> listener) {
+        final List<Subject> subjects = users.stream().map(user -> {
+            if (user instanceof AnonymousUser) {
+                return new Subject(user, Authentication.RealmRef.newAnonymousRealmRef(Node.NODE_NAME_SETTING.get(settings)));
+            } else if (ClientReservedRealm.isReserved(user.principal(), settings)) {
+                return new Subject(user, reservedRealm.realmRef());
+            } else {
+                return new Subject(user, nativeRealmRef);
+            }
+        }).toList();
```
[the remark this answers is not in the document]
ywangd: I address this comment [here](https://github.com/elastic/elasticsearch/pull/89570#issuecomment-1231043754) since it is related to the other comment.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/user/TransportGetUsersAction.java
```diff
@@ -6,46 +6,69 @@
...
         super(GetUsersAction.NAME, transportService, actionFilters, GetUsersRequest::new);
         this.settings = settings;
         this.usersStore = usersStore;
         this.reservedRealm = reservedRealm;
+        this.nativeRealmRef = realms.getRealmRefs()
+            .values()
+            .stream()
+            .filter(realmRef -> NativeRealmSettings.TYPE.equals(realmRef.getType()))
+            .findFirst()
+            .orElseThrow(() -> new IllegalStateException("native realm realm ref not found"));
```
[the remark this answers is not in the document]
ywangd: Same as above.

albertzaharovits:
> Therefore, I'll leave it with the transport handler. I hope this is OK with you.

It's OK with me, :shipit:

PR #89578, opened 2022-08-24
Small cleanup in ShardGetService

romseygeek:
We can use try-with-resources here, and remove a null check because `IndexShard.get` 
never returns `null`.


romseygeek:
I was in here anyway and thought I'd pull this tiny refactoring out into its own PR.

nik9000 (approved):

PR #89580, opened 2022-08-24
Discuss: Script Write Field API with basic path resolution.

stu-elastic:
This implements most of the field CRUD API with basic path resolution.

Structural changes, such as path move, are unimplemented along with handling nested documents.


stu-elastic:
@jdconrad @grcevski Can you take a look at these changes and comment on the high level approach?  This PR isn't intended to be merged, I'll put a new one up once we have consensus on approach.

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.txt
```diff
@@ -40,4 +40,25 @@ class org.elasticsearch.script.Metadata {
 
 class org.elasticsearch.script.IngestScript {
     Metadata metadata()
+    WriteField field(String)
+}
+
+class org.elasticsearch.script.field.WriteField {
+    String getName()
+    boolean isExists()
```
grcevski, line 48: Could this simply be named `exists()`?

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.txt
```diff
@@ -40,4 +40,25 @@ class org.elasticsearch.script.Metadata {
...
+    WriteField set(def)
+    WriteField append(def)
+    boolean isEmpty()
+    int size()
+    Iterator iterator()
+    def get(def)
+    def get(int, def)
+    boolean hasValue(Predicate)
+    WriteField transform(Function)
+    WriteField dedupe()
```
grcevski, line 61: Since 'dupe' can have other meanings, I'd name this API as `deduplicate` or if we desire shorter name `uniq` like the Unix command.

grcevski (approved):
This matches the design we discussed, I just had two minor comments around the naming of the APIs.

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.txt
stu-elastic: The reason it's called `isExists()` is because painless shortcuts `getX`, `setX` and/or `isX` to `x`, so a script can go `field('foo').exists` rather than `field('foo').exists()`.

modules/lang-painless/src/main/resources/org/elasticsearch/painless/org.elasticsearch.script.ingest.txt
stu-elastic: Will change.

server/src/main/java/org/elasticsearch/script/field/WriteField.java
```diff
@@ -0,0 +1,386 @@
...
+     * Get the path represented by this Field
+     */
+    public String getName() {
+        return path;
+    }
+
+    /**
+     * Does the path exist?
+     */
+    public boolean isExists() {
```
jdconrad, line 50: I agree with @grcevski on the naming of this. I don't think users need the shortcut here as we don't really have the `exists` shortcut in other data structures. `exists()` seems fine to me.

server/src/main/java/org/elasticsearch/script/field/WriteField.java
```diff
@@ -0,0 +1,386 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+public class WriteField implements Field<Object> {
+    protected String path;
+    protected final Supplier<Map<String, Object>> rootSupplier;
```
jdconrad, line 25: What is the advantage of having this as a supplier?

server/src/main/java/org/elasticsearch/script/field/WriteField.java
```diff
@@ -0,0 +1,386 @@
...
+
+    protected Map<String, Object> container;
+    protected String leaf;
```
jdconrad, line 28: Since we only update container and leaf, does this mean that two write fields operating on the same path may not necessarily reflect the changes of each other?

jdconrad (approved):
Left a few comments/questions; otherwise, this looks good.

server/src/main/java/org/elasticsearch/script/field/WriteField.java
stu-elastic: The root may change because it was completely replaced by the user.

server/src/main/java/org/elasticsearch/script/field/WriteField.java
stu-elastic: Yes.  If there is a structural change then two write fields may not see the changes with each other.  I think that's fine for an initial implementation.

The ways to solve the problem are to re-resolve on every operation or use a backing data structure for coordination.

server/src/main/java/org/elasticsearch/script/field/WriteField.java
stu-elastic: Will change.

server/src/main/java/org/elasticsearch/script/field/WriteField.java
jdconrad: Do you foresee any issues w/ bwc if we do make the fields reflect changes to other fields? I think long term it's probably most correct to have changes reflected, so I worry if we release this will be a difficult thing to change. I also worry about users just outright considering this a bug, but I do admit for most use cases it'll probably not come up.

server/src/main/java/org/elasticsearch/script/field/WriteField.java
jdconrad: Makes sense, thanks.

server/src/main/java/org/elasticsearch/script/field/WriteField.java
stu-elastic: > Do you foresee any issues w/ bwc if we do make the fields reflect changes to other fields?

Adding functionality is backwards compatible, which is why I want to start with this and then work on adding functionality.

> I also worry about users just outright considering this a bug

Yes, I could see that.  I would say it's a missing feature.  It would be like moving a working directory after a user has `cd`ed into it, it resolved once, at `cd` time and didn't re-resolve.

stu-elastic:
Thanks @jdconrad and @grcevski https://github.com/elastic/elasticsearch/pull/89738 is the PR for code review

PR #89582, opened 2022-08-24
[8.4] Don't try to disable ML on incompatible versions (#89565)

mark-vieira:
Backports the following commits to 8.4:
 - Don't try to disable ML on incompatible versions (#89565)


mark-vieira:
@elasticmachine update branch

PR #89583, opened 2022-08-24
[DOCS] Add troubleshooting searches guide

### LINKED ISSUE #73221: [DOC] Troubleshooting Invalid Search Issues
Hiya! Will you correct, finish, & post this type of doc content related to troubleshooting Elasticsearch search issues? ([blog about this](https://www.elastic.co/blog/found-beginner-troubleshooting) from 2014.) Recommended investigative steps:

## 1. index exists

looking to [see index exists w/docs.count>0](https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-indices.html)

```
GET _cat/indices?v=true&s=index&h=index,docs.count,status,health
```

IF NOT: obvs nothing's going to work past this step

### using index patterns

Sometimes, users will search the Kibana [index pattern](https://www.elastic.co/guide/en/kibana/current/index-patterns.html) not understanding it's different. That's ok, because it has to format match w/wildards. Usually, you'll look this up via the UI, but if you feel inclined via the [Kibana API](https://www.elastic.co/guide/en/kibana/current/saved-objects-api-find.html):

```
curl -i -X GET "https://USERNAME:PASSWORD@KIBANA_URL:9243/api/saved_objects/_find?fields=title,type&per_page=10000&type=index-pattern" -H 'kbn-xsrf: true'
```

### using aliases

Sometimes, users will be searching an alias, so it looks like things don't exist/reference, but they do. [Correlate alias to indices](https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-alias.html)

```
GET _cat/aliases?v=true&h=index,alias,is_write_index&s=index
```

## 2. example doc

looking to [see column exists & has values](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html)

```
GET INSERT_INDEX_NAME/_search
{"size":1, "sort": [{"@timestamp": {"order": "desc"}}]}
```

IF NOT: pivot to ingest/datasource

## 3. index mapping

looking to see column data type searchable in [mapping](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-mapping.html) in method attempted (🤔 not fully sure if this is valid as a whole, but know searching keywords wrong causes no responses when there should be)

```
GET INSERT_INDEX_NAME/_mapping
```

IF NOT: change search method or [reindex](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-reindex.html) w/different mapping column value

## 4. index settings

check [index settings](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-get-settings.html), not sure about any other potential issues, but aware that global searches halted if column not in `index.query.default_field`

```
GET INSERT_INDEX_NAME/_settings?flat_settings=true&include_defaults=true
```

IF NOT: [update existing index settings](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html) & look back at [pipelines](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest.html#add-pipeline-to-indexing-request) or [template](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html) inducing this

```
# 🤔 I'm not fully sure how the connection references, so pulling all
GET /_index_template
```

## 5. search profiler

I'm outta ideas so time to use the [search profiler](https://www.elastic.co/guide/en/kibana/current/xpack-profiler.html) on the attempted search. If you're coming from the [Kibana Discover](https://www.elastic.co/guide/en/kibana/current/discover.html) or [Kibana Visualize](https://www.elastic.co/guide/en/kibana/6.8/vis-inspector.html) use Inspect / View:Requests / Request.

abdonpijpelink:
Adds a troubleshooting guide for searches. Closes #73221 

Preview link: https://elasticsearch_89583.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/troubleshooting-searches.html


abdonpijpelink:
@elasticmachine update branch

stefnestor (approved):
🎉 thank you

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
```diff
@@ -0,0 +1,240 @@
...
+----
+
+Use the <<cat-indices,cat indices API>> to list all indices:
+
+[source,console]
+----
+GET _cat/indices?v=true&s=index&h=index,docs.count,status,health
+----
+
+Use the <<cat-alias,cat aliases API>> to list all aliases:
```
javanna, line 40: having to check two cat API, one for indices and one for aliases is not great from a users perspective. Should we rather point to an API that returns each index together with the aliases they are associated with?

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
```diff
@@ -0,0 +1,240 @@
...
+correct data view and a valid time range. Also, ensure the data view has been
+configured with the correct time field.
+
+[discrete]
+[[troubleshooting-searches-field-exists]]
+=== Ensure the field exists and has data
+
+Querying a field that does not exist or has no data will not return any results.
+Use the <<query-dsl-exists-query,`exists` query>> to check whether there are
+documents that contain an indexed value for a field. Check that `count` in the
```
javanna, line 81: it could also be a runtime field, in which case the field does not necessarily need to be indexed.

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
```diff
@@ -0,0 +1,240 @@
...
+----
+//TEST[continued]
+
+NOTE: When getting no search results in {kib}, check that you have selected the
+correct data view and a valid time range. Also, ensure the data view has been
+configured with the correct time field.
+
+[discrete]
+[[troubleshooting-searches-field-exists]]
+=== Ensure the field exists and has data
```
javanna, line 77: I would give a more prominent role to the field_caps API: that can totally answer the "is the field there" question. I wonder if the exists query step is even needed then. I would consider showing the output for one field, for clarify.

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
```diff
@@ -0,0 +1,240 @@
...
+  "size": 0,
+  "aggs": {
+    "top_values": {
+      "terms": {
+        "field": "my-field",
+        "size": 10
+      }
+    }
+  }
+}
```
javanna, line 114: would this apply to numeric fields too? Or would you rather use a different stats agg for numerics?

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
```diff
@@ -0,0 +1,240 @@
...
+A field's capabilities are determined by its mapping. To retrieve the mapping,
+use the <<indices-get-mapping,get mapping API>>:
+
+[source,console]
+----
+GET /my-index-000001/_mappings
+----
+//TEST[continued]
+
+For `text` fields, pay attention to the analyzer that may have been configured.
```
javanna, line 147: If you are querying a text field?

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
```diff
@@ -0,0 +1,240 @@
...
+GET /my-index-000001/_analyze
+{
+  "field" : "my-field",
+  "text" : "this is a test"
+}
+----
+//TEST[continued]
+
+<<updating-field-mappings,Changing the mapping of a field>> generally requires
+creating a new index with the correct mappings. Next, reindex the data into that
```
javanna, line 162: not for runtime fields :) they can be removed at all times. But there isn't yet support for a text runtime field, so what you wrote suits well this specific scenario.

javanna (changes requested):
This is a great effort! I left some comments

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
abdonpijpelink: That's a good idea. How about `GET _all?filter_path=*.aliases`?

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
abdonpijpelink: I think the exists query answers the "has data" question. I don't think field_caps answers that question? I think it makes sense to first check whether this is a field that returns any values, and if not, then move to field_caps as the next step?

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
abdonpijpelink: I wasn't sure whether to include an agg in this guide. At this point we don't even know if the field is aggregatable. Maybe we add a separate section "Check for common values" after the field caps section and include a terms and a stats agg there?

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
abdonpijpelink: I'll simplify this paragraph to simply point to the "Change the mapping of an existing field" documentation

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
abdonpijpelink: How about simply "documents that have a value for a field"?

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
javanna: I would think the other way around. Field caps lets you know if the field is there and whether it is searchable or not. Then you may want to know if the field has values etc.

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
javanna: documents that hold or produce a value for a field? it is subtle though... I don't like it that much :)

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
javanna: can you post the output? :)

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
javanna: field_caps would tell you if the field is aggregatable. I am ok with your plan

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
abdonpijpelink: It outputs a list of indices, with the aliases for each index:

```
{
  ".ds-logs-enterprise_search.api-default-2022.08.29-000001": {
    "aliases": {}
  },
  ".ds-logs-enterprise_search.audit-default-2022.08.29-000001": {
    "aliases": {}
  },
  "image-index": {
    "aliases": {}
  },
  "kibana_sample_data_ecommerce": {
    "aliases": {}
  },
  "kibana_sample_data_flights": {
    "aliases": {}
  },
  "kibana_sample_data_logs": {
    "aliases": {}
  },
  "my-index-000001": {
    "aliases": {
      "my-alias": {}
    }
  },
  "product-index": {
    "aliases": {}
  }
}
```

javanna: looks neat, thanks!

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
abdonpijpelink: How about "Use the exists query to check whether there are documents that return a value for a field."?

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
abdonpijpelink: I've pushed a change to reorder the steps so field caps comes earlier:

1. Check that the data stream, index or alias exists
2. Use _count to ensure the data stream, index or alias contain data
3. Use _field_caps to check that the field exists and its capabilities
4. Check mapping and text analysis
5. Check for values with the exists query and aggs
6. Validate, explain and profile
7. Check index settings

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
```diff
@@ -0,0 +1,291 @@
...
+or results in an unexpected order. This guide describes how to troubleshoot
+searches.
+
+[discrete]
+[[troubleshooting-searches-exists]]
+=== Ensure the data stream, index, or alias exists
+
+Elasticsearch returns an `index_not_found_exception` when the data stream, index
+or alias you try to query does not exist. This can happen when you misspell the
+name or when the data has been indexed to a different data stream or index.
```
javanna, line 15: would it make sense to mention the indices options `ignore_unavailable` etc. or does it complicate things?

javanna (approved):
left a small comment, but it looks great already. thanks

docs/reference/troubleshooting/troubleshooting-searches.asciidoc
abdonpijpelink: I've added a note about retrieving partial search results by setting `ignore_unavailable` to`true`.

javanna:
Thanks for all the iterations @abdonpijpelink !

Issue #89574, opened 2022-08-24
Update .fleet-agents mappings for integration health status


joshdover:
Related to https://github.com/elastic/elastic-agent/issues/100 we need to update https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/src/main/resources/fleet-agents.json to add mappings for the new fields.

Here is a sample of the new fields added:
```
{
    "status": "online",
    "events": null,
    "message": "Running",
    "components": [
        {
            "id": "winlog-default",
            "type": "winlog",
            "status": "starting",
            "message": "Starting: spawned pid '41297'",
            "units": [
                {
                    "id": "winlog-default-winlog-system-066f58e7-8cb9-4ff1-9324-26361cc929d0",
                    "type": "input",
                    "status": "starting",
                    "message": "Starting: spawned pid '41297'"
                },
                {
                    "id": "winlog-default",
                    "type": "output",
                    "status": "starting",
                    "message": "Starting: spawned pid '41297'"
                }
            ]
        },
        {
            "id": "system/metrics-default",
            "type": "system/metrics",
            "status": "starting",
            "message": "Starting: spawned pid '41298'",
            "units": [
                {
                    "id": "system/metrics-default",
                    "type": "output",
                    "status": "starting",
                    "message": "Starting: spawned pid '41298'"
                },
                {
                    "id": "system/metrics-default-system/metrics-system-066f58e7-8cb9-4ff1-9324-26361cc929d0",
                    "type": "input",
                    "status": "starting",
                    "message": "Starting: spawned pid '41298'"
                }
            ]
        },
        {
            "id": "log-default",
            "type": "log",
            "status": "starting",
            "message": "Starting: spawned pid '41296'",
            "units": [
                {
                    "id": "log-default-logfile-system-066f58e7-8cb9-4ff1-9324-26361cc929d0",
                    "type": "input",
                    "status": "starting",
                    "message": "Starting: spawned pid '41296'"
                },
                {
                    "id": "log-default",
                    "type": "output",
                    "status": "starting",
                    "message": "Starting: spawned pid '41296'"
                }
            ]
        }
    ]
}
```

aleksmaus:
The actual .fleet-agents document example:

```
{
    "access_api_key_id": "Di0iooIBTRpRIDsNZNU3",
    "action_seq_no": [
        -1
    ],
    "active": true,
    "agent": {
[...]

PR #89599, opened 2022-08-24
Fleet: Add new mappings for .fleet-agents last_checkin_message and components fields

aleksmaus:
## What does this PR do?

Adds new ```last_checkin_message``` mapping for the agent V2.
Adds new ```components``` mapping that doesn't need to be indexed.

This change is prerequisite to the fleet server feature:
https://github.com/elastic/fleet-server/pull/884

## Related issues
- Relates: https://github.com/elastic/elasticsearch/issues/89574

## Screenshots

Verified ```last_checkin_message``` and ```components``` mapping :
Screen Shot 2022-08-24 at 2 33 30 PM

Verified the document is stored with the new properties:
Screen Shot 2022-08-24 at 2 33 47 PM


aleksmaus:
@elasticmachine update branch

docs/changelog/89599.yaml
```diff
@@ -0,0 +1,6 @@
+pr: 89599
+summary: "Fleet: Add new mappings for .fleet-agents `last_checkin_message` and components\
+  \ fields"
```
AndersonQ, line 3: Is this line break intended? Or should it be all in one line?
```suggestion
summary: "Fleet: Add new mappings for .fleet-agents `last_checkin_message` and components fields"
```

AndersonQ (changes requested):

docs/changelog/89599.yaml
aleksmaus: this file was automagically creating by the existing tooling, so I assume it knows what it's doing

x-pack/plugin/core/src/main/resources/fleet-agents.json
```diff
@@ -48,6 +48,19 @@
...
         },
+        "last_checkin_message": {
+          "type": "text",
+          "fields": {
+            "keyword": {
+              "type": "keyword",
+              "ignore_above": 1024
+            }
+          }
+        },
```
javanna, line 59: May I ask how last_checking_message is going to be consumed? Is it going to be queried, hence it needs to be tokenized, but also sorted/aggregated on, hence it needs to have a keyword sub-field?

docs/changelog/89599.yaml
AndersonQ: good point, I'd assume the same

AndersonQ (approved):

x-pack/plugin/core/src/main/resources/fleet-agents.json
aleksmaus: I could image both use cases, where it would need to be searched to figure out what or how many agents have particular check-in message. That's why ended up with what seems to be fairly standard auto-mapping for the tex/string content.

joeypoon (approved):

x-pack/plugin/core/src/main/resources/fleet-agents.json
javanna: Cool, it is rather standard but sometimes it is also misused and ends up creating two lucene fields. If the field needs to be tokenized, searched as well as aggregated/sorted, this is the way to go.

PR #89567, opened 2022-08-24
Operator/slm policy

grcevski:
This PR adds support for /_slm/policy in file based settings.

Relates to https://github.com/elastic/elasticsearch/issues/89183

PR #89601, opened 2022-08-24
Add reserved snapshot/repo action

grcevski:
This PR adds support for /_snapshot/repo file based settings.

Pre-requisite for: https://github.com/elastic/elasticsearch/pull/89567


server/src/main/java/org/elasticsearch/action/admin/cluster/repositories/delete/TransportDeleteRepositoryAction.java
```diff
@@ -69,4 +73,14 @@ protected void masterOperation(
             listener.map(unregisterRepositoryResponse -> AcknowledgedResponse.of(unregisterRepositoryResponse.isAcknowledged()))
         );
     }
+
+    @Override
```
grcevski, line 77: SLM and ILM might depend on this snapshot repo configuration, but not always. This extends the handler interface to allow optional dependencies, they are here simply for ordering purposes so we can create the repo before the SLM policy.

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
```diff
@@ -193,52 +184,7 @@ public void registerRepository(final PutRepositoryRequest request, final ActionL
...
-                if (found == false) {
-                    repositoriesMetadata.add(new RepositoryMetadata(request.name(), request.type(), request.settings()));
-                }
-                repositories = new RepositoriesMetadata(repositoriesMetadata);
-                mdBuilder.putCustom(RepositoriesMetadata.TYPE, repositories);
-                changed = true;
-                return ClusterState.builder(currentState).metadata(mdBuilder).build();
-            }
-
+        submitUnbatchedTask("put_repository [" + request.name() + "]", new RegisterRepositoryTask(this, request, acknowledgementStep) {
```
grcevski, line 187: This refactoring simply moves the code so we can reuse the task execute method for the reserved state handler.

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
```diff
@@ -142,16 +142,7 @@ public RepositoriesService(
...
-        validate(request.name());
-
-        // Trying to create the new repository on master to make sure it works
-        try {
-            closeRepository(createRepository(newRepositoryMetadata, typesRegistry, RepositoriesService::throwRepositoryTypeDoesNotExists));
-        } catch (Exception e) {
-            listener.onFailure(e);
-            return;
-        }
+        validateRepository(request, listener);
```
grcevski, line 145: The validation logic was extracted so we can use it in the reserved state handler.

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
```diff
@@ -350,40 +390,7 @@ public void clusterStateProcessed(ClusterState oldState, ClusterState newState)
...
-                        return ClusterState.builder(currentState).metadata(mdBuilder).build();
-                    }
-                }
-                if (Regex.isMatchAllPattern(request.name())) { // we use a wildcard so we don't barf if it's not present.
-                    return currentState;
-                }
-                throw new RepositoryMissingException(request.name());
-            }
-
+        submitUnbatchedTask("delete_repository [" + request.name() + "]", new UnregisterRepositoryTask(request, listener) {
```
grcevski, line 393: Similar extraction of the execute logic in a standalone task to be reused.

server/src/main/java/org/elasticsearch/reservedstate/ReservedClusterStateHandler.java
```diff
@@ -77,6 +77,20 @@ default Collection<String> dependencies() {
...
+     * List of optional dependent handler names for this handler.
+     *
+     * <p>
+     * These are dependent handlers which may or may not exist for this handler to be
+     * processed. If the optional dependency exists, then they are simply ordered to be
+     * merged into the cluster state before this handler.
+     *
+     * @return a collection of optional reserved state handler names
+     */
+    default Collection<String> optionalDependencies() {
```
grcevski, line 90: SLM and ILM might depend on this snapshot repo configuration, but not always. This extends the handler interface to allow optional dependencies, they are here simply for ordering purposes so we can create the repo before the SLM policy.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -82,13 +82,11 @@ public FileSettingsService(ClusterService clusterService, ReservedClusterStateSe
         this.operatorSettingsDir = environment.configFile().toAbsolutePath().resolve(OPERATOR_DIRECTORY);
     }
 
-    // package private for testing
-    Path operatorSettingsDir() {
+    public Path operatorSettingsDir() {
```
grcevski, line 85: Making these public so that we can use them in tests outside of the package.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedClusterStateService.java
```diff
@@ -267,7 +267,21 @@ private void addStateHandler(String key, Set<String> keys, LinkedHashSet<String>
...
+            if (keys.contains(dependency)) {
+                addStateHandler(dependency, keys, ordered, visited);
+            }
+        }
+
         visited.remove(key);
         ordered.add(key);
     }
+
+    /**
```
grcevski, line 280: The state handlers are initialized on creation of the service, which is hung off the ActionModule. The Node creation code has many service interdependencies and it's impossible to create all reserved state handlers ahead of time. This API allows us the flexibility to add other state handlers as we build the modules in Node.

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
original-brownbear: This seems broken, we should return here on any validation exception?
Maybe better to leave this a non-async method and do the try catch here and let the exception bubble up in the new spot this is used in where the listener is somewhat artificial anyway?

original-brownbear (commented):
Gave it a quick read and looks reasonable. Will do a deeper review soon, but I think found one issue that needs addressing until then.

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
grcevski: So while fixing this I noticed something, the name validation is a sync failure, but creating the repository uses the listener. I'm not sure there's a reason for this difference, but I changed the code to mimic the original behaviour now.

server/src/main/java/org/elasticsearch/action/admin/cluster/repositories/reservedstate/ReservedRepositoryAction.java
```diff
@@ -0,0 +1,112 @@
...
+
+    }
+
+    @Override
+    public List<PutRepositoryRequest> fromXContent(XContentParser parser) throws IOException {
+        List<PutRepositoryRequest> result = new ArrayList<>();
+
+        Map<String, ?> source = parser.map();
+
+        for (String name : source.keySet()) {
```
original-brownbear, line 100: Not too important but this is a bit of a strange loop, why not loop EntrySet and avoid the `source.get(name)`?

original-brownbear (commented):
Thanks Nikola, this looks alright to me now in terms of the cluster state updates and changes to the `RepositoriesService`. 
But just generally speaking, shouldn't this have some end-to-end test in the form of an internal cluster test or REST test?

grcevski:
> Thanks Nikola, this looks alright to me now in terms of the cluster state updates and changes to the `RepositoriesService`.
> But just generally speaking, shouldn't this have some end-to-end test in the form of an internal cluster test or REST test?

Thanks Armin! Great point on the testing, I had a test when this work was together with the SLM policies, I forgot to bring the part of the integration test. I'll follow-up with and update now.

grcevski:
There's still a race condition. I need to think a bit more how to prevent it.

server/src/internalClusterTest/java/org/elasticsearch/reservedstate/service/RepositoriesFileSettingsIT.java
```diff
@@ -0,0 +1,250 @@
...
+        assertEquals(
+            "[err-repo] missing",
+            expectThrows(
+                RepositoryMissingException.class,
+                () -> client().execute(GetRepositoriesAction.INSTANCE, new GetRepositoriesRequest(new String[] { "err-repo" })).actionGet()
+            ).getMessage()
+        );
+
+        // This should succeed, nothing was reserved
+        client().execute(PutRepositoryAction.INSTANCE, sampleRestRequest("err-repo")).actionGet();
```
original-brownbear, line 217: Can we use just plain `get()`, `actionGet()` hides the exact point where the exception was thrown by unwrapping the execution exception and unless it's needed for something like `expectThrows` above I'd rather avoid it because it makes debugging test failures a pain :)

server/src/internalClusterTest/java/org/elasticsearch/reservedstate/service/RepositoriesFileSettingsIT.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    private void assertClusterStateSaveOK(CountDownLatch savedClusterState, AtomicLong metadataVersion) throws Exception {
+        boolean awaitSuccessful = savedClusterState.await(20, TimeUnit.SECONDS);
+        assertTrue(awaitSuccessful);
+
+        final var reposResponse = client().execute(
+            GetRepositoriesAction.INSTANCE,
+            new GetRepositoriesRequest(new String[] { "repo", "repo1" })
+        ).actionGet();
```
original-brownbear, line 144: Lets use `get` here

server/src/internalClusterTest/java/org/elasticsearch/reservedstate/service/RepositoriesFileSettingsIT.java
```diff
@@ -0,0 +1,250 @@
...
+                          "location": "my_backup_location"
+                       }
+                    }
+                 }
+             }
+        }""";
+
+    private void assertMasterNode(Client client, String node) {
+        assertThat(
+            client.admin().cluster().prepareState().execute().actionGet().getState().nodes().getMasterNode().getName(),
```
original-brownbear, line 98: `get` :)

original-brownbear (approved):
LGTM just one thing to maybe fix in the tests if you have a sec :) Thanks for the iterations!

grcevski:
Thanks Armin!

Issue #89585, opened 2022-08-24
[ML] Return 408 instead of 500 when open/start requests time out


dolaru:
### Description

This applies to the following APIs:
- [ ] open anomaly detection
- [ ] start datafeed
- [ ] start data frame analytics job
- [x] start trained model deployment

Currently, if the operation takes longer than the configured `timeout`, the response status code is `500`.

With Kibana in mind, a `4xx` response would be more suitable as it will lead to showing more useful information to the user about why the request failed.

Looking at [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-408-request-timeout), it seems the most appropriate `4xx` code would be `408 Request Timeout`, even though the RFC describes a different situation. 

The alternative, `504 Gateway Timeout`, would be more appropriate for this situation but it wouldn't improve the user experience in Kibana.

PR #89612, opened 2022-08-25
[ML] Return 408 when start deployment api times out

dimitris-athanasiou:
This commit changes the status code returned when the start
trained model deployment api times out from `500` to `408`.
In addition, we add validation that the timeout must be positive.

Relates #89585


dimitris-athanasiou:
@elasticmachine update branch

edsavage (approved):
LGTM

PR #89614, opened 2022-08-25
Cleanup dead branch and duplication in ThreadContext

original-brownbear:
The context stashing keeps showing up in profiling on the network thread. It's not a big deal but easy to simplify and speed up a little:

We never create a stored context that has transient headers to clear while at the same
time preserves response headers. Not supporting this path allows for splitting up
the logic into 3 distinct methods that are much easier to follow logically.

Also, this commit removes a few redundant threadlocal.get and set calls.


grcevski (approved):
LGTM!

original-brownbear:
Thanks Nikola!

PR #89619, opened 2022-08-25
Make SnapshotsInProgress Diffable

### LINKED ISSUE #88732: Make `SnapshotsInProgress` diffable
`SnapshotsInProgress` can be pretty large and today it's an `AbstractNamedDiffable` so any changes to any snapshot status result in sending the whole thing over the wire to all nodes. Yet, it undergoes a lot of small changes as snapshots proceed towards completion. We should make it properly diffable so that small changes have smaller representations on the wire and unchanged parts can be shared between instances.

Relates #77466.

original-brownbear:
This is very important for #77466. Profiling showed that serializing snapshots-in-progress when there's a few snapshots with high shard count running takes a significant amount of CPU and heap for sending the full data structure over an over.
This PR adds diffing in the simplest way I could think of on top of the existing data structure.

closes #88732


server/src/test/java/org/elasticsearch/snapshots/SnapshotsInProgressSerializationTests.java
```diff
@@ -187,44 +187,8 @@ protected Custom mutateInstance(Custom instance) {
     }
 
     private Entry mutateEntry(Entry entry) {
-        switch (randomInt(8)) {
```
original-brownbear, line 190: These examples were not legal state transitions for a snapshot entry so I didn't implement diffing for them and remove the test cases.

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
```diff
@@ -1310,5 +1357,326 @@ public void writeTo(StreamOutput out) throws IOException {
...
+        }
+
+        /**
+         * @param diffBySnapshotUUID diff of a map of snapshot UUID to snapshot entry
+         * @param positionDiff diff of a map with snapshot UUID keys and positions in {@link ByRepo#entries} as values. Used to efficiently
+         *                     diff an entry moving to another index in the list
+         */
+        private record ByRepoDiff(
+            DiffableUtils.MapDiff<String, Entry, Map<String, Entry>> diffBySnapshotUUID,
+            DiffableUtils.MapDiff<String, Integer, Map<String, Integer>> positionDiff
```
original-brownbear, line 1663: Not the nicest solution ever to diffing a list that can see both entries change and move about in the list but the best I could come up with without refactoring things in this class.

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
```diff
@@ -780,6 +791,42 @@ private Entry(
             assert assertShardsConsistent(this.source, this.state, this.indices, this.shards, this.shardStatusByRepoShardId);
         }
 
+        private Entry(
```
original-brownbear, line 794: This could use another follow-up to make this the only primary constructor since we now have two. I wanted this for efficiencies sake but didn't want to do that refactoring here since it requires turning all other constructors into static methods and adds a bunch of LoC changed (since there's so much logic in the other primary constructor).

server/src/test/java/org/elasticsearch/snapshots/SnapshotsInProgressSerializationTests.java
DaveCTurner: Hmm but we also use `mutateEntry` to test the non-diff serialization, and equals/hashcode, and in that case we do want to verify these mutations.

original-brownbear: Ah right it's also one of those tests. I pushed https://github.com/elastic/elasticsearch/pull/89619/commits/852cf9d818db9be81c7e13fe01095f8bd4973371to split out the "legal" mutations so I can just use them for the diffing and brought the other ones back.

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
```diff
@@ -721,7 +732,7 @@ public Entry(
                 failure,
                 userMetadata,
                 version,
-                null,
+                (SnapshotId) null,
```
DaveCTurner, line 735: Distinguishing overloads by the type of a single nullable parameter? Bleurgh.

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
DaveCTurner: Maybe do that first?

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
DaveCTurner: Likewise, could we do that prep work first and avoid having to do this? If not, could we at least sprinkle a bunch of assertions around here to validate our assumptions about the diff (no gaps, no duplicates, etc)?

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
```diff
@@ -1310,5 +1357,326 @@ public void writeTo(StreamOutput out) throws IOException {
...
+        }
+
+        @Override
+        public SnapshotsInProgress apply(Custom part) {
+            return new SnapshotsInProgress(mapDiff.apply(((SnapshotsInProgress) part).entries));
+        }
+
+        @Override
+        public Version getMinimalSupportedVersion() {
+            return Version.CURRENT.minimumCompatibilityVersion();
```
idegtiarenko, line 1577: Should it be 8_5?

DaveCTurner (commented):
Tests look ok now, I left a couple of other ideas.

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
original-brownbear: Sure how about https://github.com/elastic/elasticsearch/pull/89813 :)

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
original-brownbear: I know :) lets do https://github.com/elastic/elasticsearch/pull/89813 first then ;)

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
original-brownbear: I think a big refactoring that avoids the current structure of a map of lists will be hard to pull off in the short-term. It's a massive change-set that would be required here and I'd much rather do it in increments.
The beauty of the code here is that at least for the positions we need no such assertions since `List.of(...)` will be unforgiving to gaps. So if we take the size of the array from the uuid->snaphot map and then use the positions for the iteration we can't really run into trouble ever can we?

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
original-brownbear: I don't think so because we were always allegedly "diffable" but used a non-diff. That's why I had to add the BwC `new SimpleDiffable.CompleteDiff<>(after).writeTo(out);` below. So in a sense this is just a wire format change from the perspective of the checks on this I think.

original-brownbear:
@DaveCTurner ping in case you have a second, this would be quite nice to have in for our snapshot benchmarking. It's a surprisingly large speedup :)

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
```diff
@@ -1312,5 +1323,338 @@ public void writeTo(StreamOutput out) throws IOException {
...
+                i -> DiffableUtils.readJdkMapDiff(
+                    i,
+                    REPO_SHARD_ID_KEY_SERIALIZER,
+                    (DiffableUtils.ValueSerializer<RepositoryShardId, ShardSnapshotStatus>) SHARD_SNAPSHOT_STATUS_VALUE_SERIALIZER
+                )
+            );
+        }
+
+        @SuppressWarnings("unchecked")
+        EntryDiff(Entry before, Entry after) {
```
DaveCTurner, line 1428: I think we should be explicit about the fields that must not change between `before` and `after` here, at least asserting that they're the same but maybe also protecting against that kind of bug in production too.

server/src/test/java/org/elasticsearch/snapshots/SnapshotsInProgressSerializationTests.java
```diff
@@ -252,84 +258,84 @@ private Entry mutateEntry(Entry entry) {
...
-                IndexId indexId = new IndexId(randomAlphaOfLength(10), randomAlphaOfLength(10));
-                indices.put(indexId.getName(), indexId);
-                Map<ShardId, SnapshotsInProgress.ShardSnapshotStatus> shards = new HashMap<>(entry.shards());
-                Index index = new Index(indexId.getName(), randomAlphaOfLength(10));
-                int shardsCount = randomIntBetween(1, 10);
-                for (int j = 0; j < shardsCount; j++) {
-                    shards.put(new ShardId(index, j), randomShardSnapshotStatus(randomAlphaOfLength(10)));
-                }
+            case 1 -> {
+                long repositoryStateId = randomValueOtherThan(entry.startTime(), ESTestCase::randomLong);
```
DaveCTurner, line 320: Think we mean this?

```suggestion
                long repositoryStateId = randomValueOtherThan(entry.repositoryStateId(), ESTestCase::randomLong);
```

server/src/test/java/org/elasticsearch/snapshots/SnapshotsInProgressSerializationTests.java
```diff
@@ -147,7 +147,7 @@ protected Custom makeTestChanges(Custom testInstance) {
                 for (int i = 0; i < entries.size(); i++) {
                     if (randomBoolean()) {
                         final Entry entry = entries.get(i);
-                        entries.set(i, mutateEntry(entry));
+                        entries.set(i, mutateEntryWithLegalChange(entry));
                     }
                 }
                 updatedInstance = updatedInstance.withUpdatedEntriesForRepo(perRepoEntries.get(0).repository(), entries);
```
DaveCTurner, line 153: Can we also reorder the entries for one or more repositories? Really just to give the `ByRepoDiff` stuff a proper workout.

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
DaveCTurner: Ok - I'll suggest a couple of extra assertions in some following comments.

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
```diff
@@ -1312,5 +1323,338 @@ public void writeTo(StreamOutput out) throws IOException {
...
+            for (int i = 0; i < part.entries.size(); i++) {
+                res.put(part.entries.get(i).snapshot().getSnapshotId().getUUID(), i);
+            }
+            return res;
+        }
+
+        public static Map<String, Entry> toMapByUUID(ByRepo part) {
+            final Map<String, Entry> res = Maps.newMapWithExpectedSize(part.entries.size());
+            for (Entry entry : part.entries) {
+                res.put(entry.snapshot().getSnapshotId().getUUID(), entry);
```
DaveCTurner, line 1629:
```suggestion
                final var snapshotUUID = entry.snapshot().getSnapshotId().getUUID();
                assert res.containsKey(snapshotUUID) == false;
                res.put(snapshotUUID, entry);
```

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
```diff
@@ -1312,5 +1323,338 @@ public void writeTo(StreamOutput out) throws IOException {
...
+                    DiffableUtils.getStringKeySerializer(),
+                    INT_DIFF_VALUE_SERIALIZER
+                )
+            );
+        }
+
+        public static Map<String, Integer> toPositionMap(ByRepo part) {
+            final Map<String, Integer> res = Maps.newMapWithExpectedSize(part.entries.size());
+            for (int i = 0; i < part.entries.size(); i++) {
+                res.put(part.entries.get(i).snapshot().getSnapshotId().getUUID(), i);
```
DaveCTurner, line 1621:
```suggestion
                final var snapshotUUID = part.entries.get(i).snapshot().getSnapshotId().getUUID();
                assert res.containsKey(snapshotUUID) == false;
                res.put(snapshotUUID, i);
```

DaveCTurner (commented):
A few more small comments

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
```diff
@@ -72,13 +72,14 @@ private static Map<String, List<Entry>> collectByRepo(StreamInput in) throws IOE
             final Entry entry = Entry.readFrom(in);
             entriesByRepo.computeIfAbsent(entry.repository(), repo -> new ArrayList<>()).add(entry);
         }
+        final Map<String, ByRepo> res = Maps.newMapWithExpectedSize(entriesByRepo.size());
         for (Map.Entry<String, List<Entry>> entryForRepo : entriesByRepo.entrySet()) {
-            entryForRepo.setValue(List.copyOf(entryForRepo.getValue()));
+            res.put(entryForRepo.getKey(), new ByRepo(entryForRepo.getValue()));
```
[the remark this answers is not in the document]
original-brownbear: The constructor of `ByRepo` does the immutable copy.

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
```diff
@@ -87,26 +88,26 @@ public SnapshotsInProgress withUpdatedEntriesForRepo(String repository, List<Ent
...
-        final Map<String, List<Entry>> copy = new HashMap<>(this.entries);
+        final Map<String, ByRepo> copy = new HashMap<>(this.entries);
         if (updatedEntries.isEmpty()) {
             copy.remove(repository);
             if (copy.isEmpty()) {
                 return EMPTY;
             }
         } else {
-            copy.put(repository, List.copyOf(updatedEntries));
+            copy.put(repository, new ByRepo(updatedEntries));
```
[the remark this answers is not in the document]
original-brownbear: The constructor of `ByRepo` does the immutable copy.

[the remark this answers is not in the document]
DaveCTurner: Ah oops sorry I forgot records can do stuff like that.

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
original-brownbear: Makes sense, I added actual exception throwing in production. If we ever run into a bug here, it's still better to stall everything and (maybe) for a full cluster restart to fix things than to corrupt the repo IMO :)

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
original-brownbear: ++ added

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
original-brownbear: ++

server/src/test/java/org/elasticsearch/snapshots/SnapshotsInProgressSerializationTests.java
original-brownbear: Jup added shuffling the list. Had to change the way we generate the repo name for that. It was a random string per snapshot and we'd effectively never see collisions there. Now we actually have multiple snapshots per repo here and the position reordering is covered by tests.

server/src/test/java/org/elasticsearch/snapshots/SnapshotsInProgressSerializationTests.java
original-brownbear: I think so yes, changed.

server/src/main/java/org/elasticsearch/cluster/SnapshotsInProgress.java
```diff
@@ -87,26 +88,26 @@ public SnapshotsInProgress withUpdatedEntriesForRepo(String repository, List<Ent
...
         }
         return new SnapshotsInProgress(copy);
     }
 
     public SnapshotsInProgress withAddedEntry(Entry entry) {
-        final List<Entry> forRepo = new ArrayList<>(entries.getOrDefault(entry.repository(), List.of()));
+        final List<Entry> forRepo = new ArrayList<>(entries.getOrDefault(entry.repository(), ByRepo.EMPTY).entries);
```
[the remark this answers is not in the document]
original-brownbear: ++ done

original-brownbear:
Thanks @DaveCTurner all points addressed I think :)

server/src/test/java/org/elasticsearch/snapshots/SnapshotsInProgressSerializationTests.java
DaveCTurner: :+1: sounds good

DaveCTurner (approved):
LGTM

original-brownbear:
Thanks David!

Issue #54550, opened 2020-04-01
Unable to create filtered alias with terms lookup


junjun-zhang:
**Elasticsearch version** (`bin/elasticsearch --version`):
OpenJDK 64-Bit Server VM warning: Option UseConcMarkSweepGC was deprecated in version 9.0 and will likely be removed in a future release.
Version: 7.6.1, Build: default/deb/aa751e09be0a5072e8570670309b1f12348f023b/2020-02-29T00:15:25.529771Z, JVM: 13.0.2

**Plugins installed**: []

**JVM version** (`java -version`):
openjdk version "11.0.1" 2018-10-16 LTS
OpenJDK Runtime Environment Zulu11.2+3 (build 11.0.1+13-LTS)
OpenJDK 64-Bit Server VM Zulu11.2+3 (build 11.0.1+13-LTS, mixed mode)

**OS version** (`uname -a` if on a Unix-like system):
`Linux  5.3.0-42-generic #34-Ubuntu SMP Fri Feb 28 05:49:40 UTC 2020 x86_64 GNU/Linux`

**Description of the problem including expected versus actual behavior**:

**Steps to reproduce**:
 1. Index some documents:
```
POST _bulk
{"index":{"_index":"colors","_id":"1"}}
{"color":"blue"}
{"index":{"_index":"colors","_id":"2"}}
{"color":["red","green"]}
{"index":{"_index":"products","_id":"1"}}
{"id":"a","color":"red"}
{"index":{"_index":"products","_id":"2"}}
{"id":"b","color":"blue"}
```

 2. Create a filtered alias with terms lookup:
```
POST /products/_aliases/color2_products
{
  "filter": {
    "terms": {
      "color": {
        "index": "colors",
        "id": "2",
        "path": "color"
      }
    }
  }
}
```

 3. Get an error:
```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "illegal_argument_exception",
        "reason" : "failed to parse filter for alias [color2_products]"
      }
    ],
    "type" : "illegal_argument_exception",
    "reason" : "failed to parse filter for alias [color2_products]",
    "caused_by" : {
      "type" : "illegal_state_exception",
      "reason" : "async actions are left after rewrite"
    }
  },
  "status" : 400
}
```

junjun-zhang:
not sure whether it's related to #54315

joegallo:
[...]

PR #89621, opened 2022-08-25
Fix bug for kNN with filtered aliases

### LINKED ISSUE #89561: Add support for kNN with filtered aliases
### Description

Currently we can’t run an approximate kNN search on a [filtered alias](https://www.elastic.co/guide/en/elasticsearch/reference/master/aliases.html#filter-alias). Running a kNN search against a filtered alias may incorrectly result in fewer than k hits.

Reproduction:

```js
PUT index1
{
    "settings": {
        "number_of_shards": 1,
        "number_of_replicas": 0
    },
    "mappings": {
        "dynamic": "false",
        "properties": {
            "my_vector": {
                "type": "dense_vector",
                "dims": 4,
                "index" : true,
                "similarity" : "l2_norm"
            },
            "k": {
                "type": "keyword",
                "store": true 
            }
        }
    }
}
```

```js
POST index1/_bulk?refresh=true
{ "index" : {"_id": 1} }
{"my_vector" : [230.0, 300.33, -34.8988, 15.555], "k" : "v1"}
{ "index" : {"_id": 2} }
{"my_vector" : [0.5, 0.5, 0.5, -1], "k" : "v1"}
{ "index" : {"_id": 3} }
{"my_vector" : [0.5, 0.5, 0.5, 0.5], "k" : "v2"}
{ "index" : {"_id": 4} }
{"my_vector" : [0.5, 0.5, 0.5, 0], "k" : "v2"}
```

```js
POST _aliases
{
    "actions": [
        {
            "add": {
                "index": "index1",
                "alias": "my-alias",
                "filter": {
                    "term": {
                        "k": "v1"
                    }
                }
            }
        }
    ]
}
````

// correctly returns 2 results as request by "k"
```js
GET index1/_search
{
    "knn": {
        "field": "my_vector",
        "query_vector": [
            230.0,
            300.33,
            -34.8988,
            15.555
        ],
        "k": 2,
        "num_candidates": 100,
        "filter": {
            "term": {
                "k": "v1"
            }
        }
    },
    "_source": [
        "k"
    ]
}
```

// incorrectly returns 1 result even "k = 2"
```js
GET my-alias/_search
{
    "knn": {
        "field": "my_vector",
        "query_vector": [
            230.0,
            300.33,
            -34.8988,
            15.555
        ],
        "k": 2,
        "num_candidates": 100
    },
    "_source": "k"
}
```

jdconrad:
This change adds the filter query for a filtered alias to the knn query during the dfs phase on the shard. This ensures the correct number of k results are returned instead of removing results as a post filter.

Fixes: https://github.com/elastic/elasticsearch/issues/89561


rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.vectors/60_knn_search_filter_alias.yml
```diff
@@ -0,0 +1,107 @@
...
+            filter:
+              term:
+                name: v1
+
+  - match: { hits.hits.0._id: "1" }
+  - match: { hits.hits.0.fields.name.0: v1 }
+
+  - match: { hits.hits.1._id: "2" }
+  - match: { hits.hits.1.fields.name.0: v1 }
+
```
mayya-sharipova, line 91: may be worth providing a small comment explaining what's happening,  or may be making vector values more obvious.
Something like: if there was no proper pre-filter implemented for vector search, the closest docs returned from vector search are with ids of  `1` and `3`, and the post-filter on them will return only a doc with id:`1`. With proper pre-filter, we ensure that filter is applied during knn search thus returning 1 result.

mayya-sharipova (approved):
@jdconrad Thanks Jack, this LGTM! Great job!

Let's also as a part of this PR, update documentation on knn search to remove a warning that knn search can't work with filtered alias.

jdconrad:
Thank you, @mayya-sharipova for the review. I will make sure to update this PR with your feedback.

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.vectors/60_knn_search_filter_alias.yml
jdconrad: @mayya-sharipova I added the additional test you suggested where the first search w/o the filtered alias shows the results where vectors 1 and 3 are nearest where 1 has the term v1 and 3 has the term v2. I then added the filtered alias and the results change to vectors 1 and 2 where those are the two nearest with the term v1. Is this the correct behavior since k = 2 or should the filter only return a single result since 1 and 2 are not technically the two nearest vectors without it?

mayya-sharipova: @jdconrad Thanks, that's correct behaviour; we need to return the nearest vectors that satisfy the filter (so it should  ids: 1 and 2 )

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.vectors/60_knn_search_filter_alias.yml
```diff
@@ -105,3 +105,37 @@ setup:
 
   - match: { hits.hits.1._id: "2" }
   - match: { hits.hits.1.fields.name.0: v1 }
+
+  - do:
```
mayya-sharipova, line 109: Great idea adding this test, now it is much more obvious.

mayya-sharipova:
@jdconrad Thanks Jack, new changes LGTM!

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.vectors/60_knn_search_filter_alias.yml
jdconrad: Credit to you for the idea :)

jdconrad:
@elasticsearchmachine run elasticsearch-ci/docs

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.vectors/60_knn_search_filter_alias.yml
```diff
@@ -0,0 +1,282 @@
...
+      index:
+        index: test
+        id: "5"
+        body:
+          name: v2
+          test_vector: [0.5, 0.5, 0.5, 0.4]
+
+  - do:
+      indices.put_alias:
```
jtibshirani, line 66: Small comment, we could fold this into the "create index" call to make it more obvious and concise.

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.vectors/60_knn_search_filter_alias.yml
```diff
@@ -0,0 +1,282 @@
+setup:
```
jtibshirani, line 1: It'd be good to add another test where the alias filter requires a query rewrite. That way we can check the alias filter is rewritten before being used in kNN search. You can see `KnnSearchSingleNodeTests#testRewrite` for an example using terms lookup queries.

jtibshirani (commented):
Thanks @jdconrad, this looks good to me overall.

One question I had: say we run a kNN-only search on a filtered alias. During the DFS phase, the kNN search will take the filter into account. Then during the query phase we'll also take a boolean AND between the kNN search results and filter. This is redundant, since the results already obey the filter. Is there a clean way to avoid this?

mayya-sharipova:
@jtibshirani 

> One question I had: say we run a kNN-only search on a filtered alias. During the DFS phase, the kNN search will take the filter into account. Then during the query phase we'll also take a boolean AND between the kNN search results and filter. This is redundant, since the results already obey the filter. Is there a clean way to avoid this?

I also thought about it, and could not come up with a good idea. May be @jdconrad  can come with a good idea.
May be one thing we can do if a search request contains only knn part, we can remove filtered alias query from a search request.

jdconrad:
@jtibshirani Thank you for the feedback. I will work on addressing it.

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.vectors/60_knn_search_filter_alias.yml
jdconrad: Updated as requested.

rest-api-spec/src/yamlRestTest/resources/rest-api-spec/test/search.vectors/60_knn_search_filter_alias.yml
jdconrad: As we discussed I attempted to use a terms lookup query as a filtered alias, but ran into #54550. Is there another type of filter that would guarantee a rewrite fails if not rewritten? I did add a single node test and traced through to ensure that we attempt to rewrite a filtered alias always, but obviously having a test is better.

jdconrad:
I explored removing the filtered alias from knn search during the query phase. I don't think it would be impossible, but it does seem difficult as the query phase really isn't built around knn. We may be able to add something to rewrite as part of the search context to detect when knn is used and not apply that filter, but I'm not sure the additional complexity is worth it.

jtibshirani (approved):

PR #89622, opened 2022-08-25
Fix allow_no_indices request option in special cases

albertzaharovits:
The `allow_no_indices` request option, when set to `false` (it is `true` by default for all APIs),
should fail indices requests that contain any wildcard expression that resolves to no resources.
This PR addresses some cases where a wildcard can expand to no resources, and the request
is nevertheless successful.
The fixed cases are when the wildcard resolves only to hidden and/or system resources and
the given request context prohibits such resources. Another case is when the wildcard resolves
to only open or closed indices and again the request context prohibits such resources.

Note that the fix only applies when Security is disabled (when enabled the behavior is  already
correct).

This is another step towards reusing the Core's `WildcardExpressionResolver#innerResolve`
in Security, following https://github.com/elastic/elasticsearch/pull/89311 .


server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1304,97 +1316,89 @@ private static IndexMetadata.State excludeState(IndicesOptions options) {
...
-                return suffixWildcard(indicesLookup, expression, context.getOptions().ignoreAliases(), context.includeDataStreams());
+        /**
+         * Given a single wildcard {@param expression}, return the {@code Stream} that contains all the resources (i.e. indices, aliases,
+         * and datastreams), that exist in the cluster at this moment in time, and that the wildcard "resolves" to (i.e. the resource's
+         * name matches the {@param expression} wildcard).
+         * The {@param context} provides the current time-snapshot view of all the resources, as well as conditions
+         * on whether to consider alias, datastream, system, and hidden resources.
+         * It does NOT consider the open or closed status of index resources.
+         */
+        private static Stream<IndexAbstraction> matchResourcesToWildcard(Context context, String wildcardExpression) {
```
albertzaharovits, line 1327: This method concentrates most (besides open/closed) resource wildcard matching and filtering in a stream-wise fashion.
Before, it was scattered accross two methods, separated by the `allow_no_indices` check.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1218,28 +1230,28 @@ private static Collection<String> innerResolve(Context context, List<String> exp
...
+                if (context.getOptions().allowNoIndices() == false) {
+                    emptyWildcardExpansion.set(true);
+                    matchingOpenClosedNames = matchingOpenClosedNames.peek(x -> emptyWildcardExpansion.set(false));
+                }
+                if (add) {
+                    matchingOpenClosedNames.forEachOrdered(result::add);
+                } else {
+                    matchingOpenClosedNames.forEachOrdered(result::remove);
+                }
+                if (emptyWildcardExpansion.get()) {
```
albertzaharovits, line 1245: Only checks the `allow_no_indices` condition at the very last when all filtering is complete.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
```diff
@@ -1304,97 +1316,89 @@ private static IndexMetadata.State excludeState(IndicesOptions options) {
...
-                    e -> Regex.simpleMatch(expression, e.getKey()),
-                    context.getOptions().ignoreAliases(),
-                    context.includeDataStreams()
-                );
+                matchesStream = indicesLookup.values().stream();
+            }
+            if (context.getOptions().ignoreAliases()) {
+                matchesStream = matchesStream.filter(indexAbstraction -> indexAbstraction.getType() != Type.ALIAS);
+            }
+            if (Regex.isMatchAllPattern(wildcardExpression) == false && Regex.isSuffixMatchPattern(wildcardExpression) == false) {
```
grcevski, line 1340: Is there any reason to filter by type before we filter by the expression? I think the code might be clearer if we moved this conditional in the else statement of `if (Regex.isSuffixMatch...`

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
grcevski: What's the reason we check for empty expansion after the `add` if/else? We are going to abandon the result anyway by throwing the exception.

grcevski (approved):
LGTM! With some comments, but this looks good to me and thanks a lot for breaking up the changes into meaningful commits, it made the PR a lot easier to follow.

I have one final remark about backwards compatibility. As we are making things less lenient, are we concerned at all about breaking changes?

slobodanadamovic (approved):
LGTM

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
albertzaharovits: > If I understand this correctly, it is required to call forEachOrdered before checking emptyWildcardExpansion flag because intermediate peek method is only executed when terminal operation is called (except for count and findFirst).

Yes, that's right.

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
albertzaharovits: I also think it would be clearer. I kept the original order, though.
Any thoughts @jakelandis ?

jakelandis (approved):
LGTM.

For posterity...  the example provided on https://github.com/elastic/elasticsearch/pull/89311#discussion_r951656374 isn't quite accurate. To exercise the bug fixed here the `allow_no_indices` description on https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html has a good example and with this change the code now now matches the documented behavior. 

> If false, the request returns an error if any wildcard expression, [index alias](https://www.elastic.co/guide/en/elasticsearch/reference/current/aliases.html), or _all value targets only missing or closed indices. This behavior applies even if the request targets other open indices. For example, a request targeting foo*,bar* returns an error if an index starts with foo but no index starts with bar 

So create an index foo as a hidden index, and bar as a normal index, then (without security) call ES with `foo*,bar*/_search?allow_no_indices=false` before this change it allowed the request to go through since it found the bar index. This was incorrect behavior since foo is hidden and we didn't request to see hidden indices and documented behavior is that if any wildcard fails to resolve the request should error. In this case foo* fails to resolve and it should error. With the change here, the query will now (correctly) error. Both before and after this change no error is raised (correctly) when explicitly asking for hidden i.e. `foo*,bar*/_search?allow_no_indices=false&expand_wildcards=hidden,open`. In this case since we requested hidden indices foo* does resolve and no error is needed.

albertzaharovits:
Thank you for the write-up in https://github.com/elastic/elasticsearch/pull/89622#pullrequestreview-1089220881 .
I realized later that `foo*/_search?allow_no_indices=false` was not enough to repro the bug, it needed to be `foo*,bar/_search?allow_no_indices=false` or `foo*,bar*/_search?allow_no_indices=false` .
Thank you for double checking!

server/src/main/java/org/elasticsearch/cluster/metadata/IndexNameExpressionResolver.java
albertzaharovits: I've pushed [1b065c8](https://github.com/elastic/elasticsearch/pull/89622/commits/1b065c8350657f7fb0a91bf8f1e657e4f7aa5fe3) . The order can theoretically influence performance, because the alias filter is cheaper from the regex, but it's hard to quantify.

PR #89624, opened 2022-08-25
[TEST] Fix SimpleThreadPoolIT file watcher thread name

grcevski:
The JDK implementation of the Watcher service is platform dependent and the two base implementations: PollingWatchService and AbstractPoller differ in the internal thread name they use: FileSystemWatchService and FileSystemWatcher, respectively. 

The current SimpleThreadPoolIT test matches the JDK thread name for Linux and Windows, but you may get failures when running locally on MacOS (which uses the polling implementation).


joegallo (approved):

grcevski:
Thanks Joe!

PR #89626, opened 2022-08-25
[DOCS] Explain dynamic behavior for unmapped copy_to fields

### LINKED ISSUE #83282: copy_to does not complain if target field does not exist
### Elasticsearch Version

7.16

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

Docker image

### Problem Description

Based on documentation, I would expect that `copy_to` field has to exist, but this does not seem to be so. See steps to reproduce to see that index is created just fine when `copy_to` is set to nonexistent `foobarX` instead of `foobar`. But it is not possible really to know what is done internally (there seems to be no way to list fields which have really been made), does this make 3 fields? 2? Does this mean that `foobar2` is simply not copied to `foobar` silently?

### Steps to Reproduce

```
curl -X PUT "localhost:9200/test?pretty" -H 'Content-Type: application/json' -d'
{
  "mappings": {
    "dynamic": false,
    "properties": {
      "foobar": {
        "type": "keyword"
      },
      "foobar2": {
        "type": "keyword",
        "copy_to": "foobarX"
      }
    }
  }
}
'
```

### Logs (if relevant)

_No response_

abdonpijpelink:
Adds a short explanation of the dynamic mapping behavior of unmapped copy_to fields. Closes #83282

### DIFF
--- a/docs/reference/mapping/params/copy-to.asciidoc
+++ b/docs/reference/mapping/params/copy-to.asciidoc
@@ -67,6 +67,13 @@ Some important points:
 * You cannot copy recursively via intermediary fields such as a `copy_to` on 
 `field_1` to `field_2` and `copy_to` on `field_2` to `field_3` expecting 
 indexing into `field_1` will eventuate in `field_3`, instead use copy_to 
-directly to multiple fields from the originating field. 
+directly to multiple fields from the originating field.
+* If the target field does not exist in the mapping, the usual
+<<dynamic-mapping,dynamic mapping>> behavior applies. By default, with
+<<dynamic,`dynamic`>> set to `true`, a non-existent target field will be
+dynamically added to the document and the mapping. If `dynamic` is set to
+`false` the target field will not be added and the value not copied. If
+`dynamic` is set to `strict`, copying to a non-existent field will result in an
+error.
 
-NOTE: `copy-to` is _not_ supported for field types where values take the form of objects, e.g. `date_range`
+NOTE: `copy_to` is _not_ supported for field types where values take the form of objects, e.g. `date_range`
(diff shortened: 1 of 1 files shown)


docs/reference/mapping/params/copy-to.asciidoc
```diff
@@ -67,6 +67,13 @@ Some important points:
 * You cannot copy recursively via intermediary fields such as a `copy_to` on 
 `field_1` to `field_2` and `copy_to` on `field_2` to `field_3` expecting 
 indexing into `field_1` will eventuate in `field_3`, instead use copy_to 
-directly to multiple fields from the originating field. 
+directly to multiple fields from the originating field.
+* If the target field does not exist in the mapping, the usual
+<<dynamic-mapping,dynamic mapping>> behavior applies. By default, with
+<<dynamic,`dynamic`>> set to `true`, a non-existent target field will be
+dynamically added to the document and the mapping. If `dynamic` is set to
```
javanna, line 74: to the mapping. I think that if you say "to the document" people may understand that we modify `_source`. But you are right, it is added to the lucene document.

docs/reference/mapping/params/copy-to.asciidoc
```diff
@@ -67,6 +67,13 @@ Some important points:
...
+`false` the target field will not be added and the value not copied. If
```
javanna, line 75: will not be added to the index mappings  ?

javanna (commented):
left two nits, LGTM otherwise

docs/reference/mapping/params/copy-to.asciidoc
abdonpijpelink: I've changed it into "a non-existent target field will be dynamically added to the index mappings"

docs/reference/mapping/params/copy-to.asciidoc
abdonpijpelink: I have changed it into "the target field will not be added to the index mappings, and the value will not be copied."

javanna (approved):
LGTM

PR #89629, opened 2022-08-25
REST tests for normalize agg

nik9000:
This adds a REST test for the normalize pipeline agg so we have
backwards compatibility tests for it.


not-napoleon (approved):
LGTM

PR #89630, opened 2022-08-25
Fix FileSettingsService hang on error update

grcevski:
While working on new tests for https://github.com/elastic/elasticsearch/pull/89567 I discovered a bug where on duplicate error update we don't unlock the file settings watcher thread properly.

It's a race condition where the settings file might be updated in succession quicker than the error state is updated. We check for duplicate error state and avoid the new cluster change update event, however we didn't call the error listener will null to unlock the watcher.

This is intermittent and should've been caught by our existing error state save integration tests, but just happened to be hit with the SLM specific test I was writing.

### DIFF
--- a/server/src/main/java/org/elasticsearch/reservedstate/service/ReservedClusterStateService.java
+++ b/server/src/main/java/org/elasticsearch/reservedstate/service/ReservedClusterStateService.java
@@ -102,7 +102,7 @@ public class ReservedClusterStateService {
             stateChunk = stateChunkParser.apply(parser, null);
         } catch (Exception e) {
             ErrorState errorState = new ErrorState(namespace, -1L, e, ReservedStateErrorMetadata.ErrorKind.PARSING);
-            saveErrorState(errorState);
+            saveErrorState(clusterService.state(), errorState);
             logger.debug("error processing state change request for [{}] with the following errors [{}]", namespace, errorState);
 
             errorListener.accept(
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/elasticsearch/reservedstate/service/ReservedClusterStateService.java
```diff
@@ -186,8 +188,7 @@ static boolean isNewError(ReservedStateMetadata existingMetadata, Long newStateV
             || existingMetadata.errorMetadata().version() < newStateVersion);
     }
 
-    private void saveErrorState(ErrorState errorState) {
-        ClusterState clusterState = clusterService.state();
+    private void saveErrorState(ClusterState clusterState, ErrorState errorState) {
```
grcevski, line 191: This change to pass in the most recent cluster state from the cluster state update task is sort of tangential, it just ensures we hit the issue more reliably. ClusterService.state() might not be the latest state, but the one we get in the update task will be.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedClusterStateService.java
```diff
@@ -170,6 +170,8 @@ public void onFailure(Exception e) {
                         if (isNewError(existingMetadata, reservedStateVersion.version())) {
                             logger.debug("Failed to apply reserved cluster state", e);
                             errorListener.accept(e);
+                        } else {
+                            errorListener.accept(null);
```
grcevski, line 174: This is the actual bug, we never called the callback to ensure the async watcher is released.

ChrisHegarty (approved):
LGTM

grcevski:
Thanks Chris!

PR #89632, opened 2022-08-25
Fix pipeline id not present in ingest metadata inside on_failure block

dakrone:
This commit fixes an issue where inside of an `on_failure` block the `_ingest.on_failure_pipeline`
metadata was not being propagated. The root cause seems to be an off-by-one error where we were
checking for at least two pipelines, when we should be checking for only a minimum of a single
pipeline that has been executed before adding the metadata.


dakrone:

seanstory (approved):
Thanks for the super quick investigation and fix!

jbaiera (approved):
LGTM

dakrone:
@elasticmachine update branch

dakrone:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|7.17|[](https://github.com/elastic/elasticsearch/pull/89740)|

### Questions ?

PR #89634, opened 2022-08-26
[Doc] Correct time units for JWT datetime claims

ywangd:
All JWT datetime claims for the JWT realm must be in epoch seconds
instead of milliseconds.


lockewritesdocs (approved):
LGTM 🦖 Thanks for making this change @ywangd!

ywangd:
@elasticmachine update branch

justincr-elastic (approved):
LGTM!

PR #89645, opened 2022-08-26
[ML] Add processor autoscaling decider

dimitris-athanasiou:
This adds a processor decider to the ML autoscaling decider service.
This first implementation is simple and naive. It simply computes
the required processor capacity to be the max trained model deployment
`threads_per_allocation` for the node, and the sum of all processors
required by trained model deployments for the tier.

### DIFF
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlMemoryAutoscalingDecider.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlMemoryAutoscalingDecider.java
@@ -822,7 +822,8 @@ class MlMemoryAutoscalingDecider {
         int totalMlProcessors = mlNodes.stream().mapToInt(node -> {
             String allocatedProcessorsString = node.getAttributes().get(MachineLearning.ALLOCATED_PROCESSORS_NODE_ATTR);
             try {
-                return Processors.of(Double.parseDouble(allocatedProcessorsString)).roundUp();
+                double allocatedProcessorsAsDouble = Double.parseDouble(allocatedProcessorsString);
+                return allocatedProcessorsAsDouble > 0 ? Processors.of(allocatedProcessorsAsDouble).roundUp() : 0;
             } catch (NumberFormatException e) {
                 assert e == null
                     : MachineLearning.ALLOCATED_PROCESSORS_NODE_ATTR
--- a/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlMemoryAutoscalingDecider.java
+++ b/x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlMemoryAutoscalingDecider.java
@@ -185,8 +185,6 @@ class MlMemoryAutoscalingDecider {
 
         final List<String> partiallyAllocatedModels = mlContext.findPartiallyAllocatedModels();
 
-        // TODO for autoscaling by memory, we only care about if the model is allocated to at least one node (see above)
-        // We should do this check in our autoscaling by processor count service, which will be a separate decider for readability's sake
         if (mlContext.waitingAnalyticsJobs.isEmpty() == false
             || mlContext.waitingSnapshotUpgrades.isEmpty() == false
             || mlContext.waitingAnomalyJobs.isEmpty() == false
@@ -257,7 +255,6 @@ class MlMemoryAutoscalingDecider {
                 if (capacity == null) {
                     return null;
                 }
-                // TODO we should remove this when we can auto-scale (down and up) via a new CPU auto-scaling decider
                 if (modelAssignmentsRequireMoreThanHalfCpu(mlContext.modelAssignments.values(), mlContext.mlNodes)) {
                     logger.debug("not down-scaling; model assignments require more than half of the ML tier's allocated processors");
                     return null;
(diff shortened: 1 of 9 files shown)


x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlMemoryAutoscalingDecider.java
```diff
@@ -185,8 +185,6 @@ public MlMemoryAutoscalingCapacity scale(Settings configuration, AutoscalingDeci
 
         final List<String> partiallyAllocatedModels = mlContext.findPartiallyAllocatedModels();
 
-        // TODO for autoscaling by memory, we only care about if the model is allocated to at least one node (see above)
```
dimitris-athanasiou, line 188: I'm thinking we should keep this check in so that we don't downscale memory anyway if there are partially allocated model deployments. What do you think?

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlProcessorAutoscalingDecider.java
```diff
@@ -0,0 +1,160 @@
...
+    MlProcessorAutoscalingCapacity computeCurrentCapacity(List<DiscoveryNode> mlNodes) {
+        int maxNodeProcessors = 0;
+        int tierProcessors = 0;
+        for (DiscoveryNode node : mlNodes) {
+            int nodeProcessors = getProcessors(node);
+            maxNodeProcessors = Math.max(maxNodeProcessors, nodeProcessors);
+            tierProcessors += nodeProcessors;
+        }
+        return MlProcessorAutoscalingCapacity.builder(maxNodeProcessors, tierProcessors).build();
+    }
```
droberts195, line 142: I think there could be a nasty interaction with #89281 here.  #89281 allows `node.processors` to be a `double`, but then rounds the number up when passing it on as an `int` to consumers that expected an `int` before.

But this could lead to spurious scale ups in Cloud.  Suppose we have 3 AZs in the ML tier, and 3 ML nodes, one per AZ, and a CPU quota of 2.5 per node.  So the whole ML tier has a total CPU quota of 7.5.  But with the rounding of 2.5 to 3 for each node we will ask for 9 CPUs for the tier when we mean "no change required".  This will lead to an unnecessary scale up.  Even at the larger tier size the same thing could happen again if `node.processors` is still fractional.  We'd keep asking for scale ups until `node.processors` was set to an integer.

I think that in places where `node.processors` is used in autoscaling code we'll need to get and use the floating point value rather than the rounded up value, so we'll need a new method `EsExecutors.unroundedAllocatedProcessors()` (or some similar name) and use that within autoscaling code.

Do you have any thoughts on this @fcofdez or @henningandersen?

droberts195 (commented):
I had a first look at this but stopped looking at the details after I realised there might be a problem with float vs rounded CPU quota.  I think we need to make sure that's round-tripped consistently for Cloud to Elasticsearch via `node.processors` and then Elasticsearch to Cloud via autoscaling responses to avoid the risk of needless scale ups.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlProcessorAutoscalingDecider.java
henningandersen: #89662 should address this I think. Hopefully we can get that one in soon and unblock this.

However, this does make your code dependent on receiving the right input from the orchestration system.

dimitris-athanasiou:
@elasticmachine  update branch

dimitris-athanasiou:
@droberts195 I have updated the PR to make use of the newly added `Processors` class. We now store the ML node attribute as a double and thus we can return the exact value when we return current capacity.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlProcessorAutoscalingDecider.java
```diff
@@ -0,0 +1,167 @@
...
+            return computeRequiredCapacity(trainedModelAssignmentMetadata).setReason(
+                "requesting scale up as there are unsatisfied deployments"
+            ).build();
+        }
+
+        final MlProcessorAutoscalingCapacity currentCapacity = computeCurrentCapacity(mlContext.mlNodes);
+
+        final MlProcessorAutoscalingCapacity requiredCapacity = computeRequiredCapacity(trainedModelAssignmentMetadata).build();
+
+        if (requiredCapacity.tierProcessors().roundUp() == currentCapacity.tierProcessors().roundUp()) {
```
droberts195, line 60: This line means that if we have 2.5 processors and need 3 then we won't scale up.

Even if it's not fixed in this PR I think it's worth adding a comment to call that out.  Maybe we can fix it in a followup for 8.6.

droberts195 (approved):
LGTM, although I think there is one thing that could be improved in the future.  Please add a comment about that.

x-pack/plugin/ml/src/main/java/org/elasticsearch/xpack/ml/autoscaling/MlProcessorAutoscalingDecider.java
dimitris-athanasiou: If we get here we are definitely not in a scenario where we want to scale up. We know all our model deployments are satisfied. The reason I added that check is to provide a nicer message why we don't downscale nor upscale if we're right where we need to be.

droberts195: > We know all our model deployments are satisfied.

But the question is whether 2.5 processors should "satisfy" a model that wanted 3.  In `TrainedModelAssignmentRebalancer` we call `getNodeAllocatedProcessors(discoveryNode).roundUp()`, so it will happily allocate a model that wants 3 processors to a node that has 2.5.  I think it's a followup for the future to consider whether we should scale up in that scenario.  Maybe the code should be changed in `TrainedModelAssignmentRebalancer` rather than here though.

Anyway, I'm happy to leave it as-is for the first version.

dimitris-athanasiou: Right, I see the point now. OK, I'll merge as-is and revisit later.

PR #89647, opened 2022-08-26
Remove needless flush call from Netty4 write handlers

original-brownbear:
These inline flush calls will never make the channel writable since
they will just get queued on the executor and run a redundant flush
execution later.
We can simplify the code here and save some cycles by simply giving up
and forwarding the flush once the channel stops being writable and
then have the logic pick up again on the next writability changed
call.


henningandersen (approved):
LGTM.

original-brownbear:
Thanks Henning!

PR #89650, opened 2022-08-26
Copy isHidden during ILM alias swap (#89604)

### LINKED ISSUE #89604: ILM swap aliases step doesn't copy the is_hidden value
### Elasticsearch Version

8.x

### Installed Plugins

_No response_

### Java Version

_bundled_

### OS Version

Any

### Problem Description

ILM [SwapAliasesAndDeleteSourceIndexStep](https://github.com/elastic/elasticsearch/blob/ffcf0eac010df3b5ba7989c5dd87bf362dd0d311/x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/ilm/SwapAliasesAndDeleteSourceIndexStep.java) doesn't play nicely with with hidden aliases. 

When we copy the aliases from the source index in the move step, we do not copy the is_hidden value.

This can lead to permanently bad ILM state with an error like:

```
policy [kibana-event-log-policy] for index [.kibana-event-log-7.11.1-000044] failed on step [{"phase":"frozen","action":"searchable_snapshot","name":"swap-aliases"}]. Moving to ERROR step java.lang.IllegalStateException: alias [.kibana-event-log-7.11.1] has is_hidden set to true on indices [...]; alias must have the same is_hidden setting on all indices at org.elasticsearch.cluster.metadata.Metadata$Builder.validateAlias(Metadata.java:2089) 
```

### Steps to Reproduce

Combination of ILM and hidden aliases.

### Logs (if relevant)

_No response_

gmarouli:
Resolves #89604

### DIFF
--- a/docs/changelog/89650.yaml
+++ b/docs/changelog/89650.yaml
@@ -1,0 +1,6 @@
+pr: 89650
+summary: Copy `isHidden` during ILM alias swap
+area: ILM+SLM
+type: bug
+issues:
+ - 89604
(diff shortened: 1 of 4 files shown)


dakrone (approved):
LGTM, thanks Mary!

gmarouli:
@dakrone nice Friday afternoon fix :)

Issue #89644, opened 2022-08-26
[CI] ResetJobIT testReset failing


slobodanadamovic:
```

[2022-07-13T09:28:01,484][ERROR][o.e.x.m.d.DataFrameAnalyticsTask] [javaRestTest-2] [test_model_memory_limit_lower_than_estimated_memory_usage] Setting task to failed org.elasticsearch.ElasticsearchException: [test_model_memory_limit_lower_than_estimated_memory_usage] analytics/12724 process stopped unexpectedly: Input error: memory limit [1 mb] has been exceeded. Please force stop the job, increase to new estimated limit [3 mb] and restart. |  
-- | --
  | »  	at org.elasticsearch.xcore@8.4.0-SNAPSHOT/org.elasticsearch.xpack.core.ml.utils.ExceptionsHelper.serverError(ExceptionsHelper.java:61) |  
  | »  	at org.elasticsearch.xpack.ml.dataframe.process.AnalyticsProcessManager.processData(AnalyticsProcessManager.java:232) |  
  | »  	at org.elasticsearch.xpack.ml.dataframe.process.AnalyticsProcessManager.lambda$runJob$1(AnalyticsProcessManager.java:153) |  
  | »  	at org.elasticsearch.server@8.4.0-SNAPSHOT/org.elasticsearch.common.util.concurrent.ThreadContext$ContextPreservingRunnable.run(ThreadContext.java:710) |  
  | »  	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) |  
  | »  	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) |  
  | »  	at java.base/java.lang.Thread.run(Thread.java:833)
```

**Build scan:**
https://gradle-enterprise.elastic.co/s/jpakm3keu7zsq/tests/:x-pack:plugin:ml:qa:native-multi-node-tests:javaRestTest/org.elasticsearch.xpack.ml.integration.ResetJobIT/testReset

**Reproduction line:**
`gradlew ':x-pack:plugin:ml:qa:native-multi-node-tests:javaRestTest' --tests "org.elasticsearch.xpack.ml.integration.ResetJobIT.testReset" -Dtests.seed=219F87BCE4E842F2 -Dtests.locale=sr-Latn-BA -Dtests.timezone=Africa/Mogadishu -Druntime.java=18`

**Applicable branches:**
main

[...]

PR #89653, opened 2022-08-26
[ML] Improve assertion error msg in ResetJobIT

dimitris-athanasiou:
This adds an error message that reveals the audit messages
if the assertion that the last msg is about resetting the job
fails.

Relates #89644


edsavage (approved):
LGTM

PR #89657, opened 2022-08-26
Add a stored fields loader

romseygeek:
To load stored fields we currently directly use lucene's FieldVisitor API.  There are
a number of nice shortcuts and some dodgy hacks that get better performance here,
which need to be reproduced everywhere the code uses this API.  It is also confusingly
stateful, and not especially easy to use.

This commit adds a new StoredFieldLoader abstraction which exposes a per-leaf stored
field loader as an iterator, and adapts ShardGetService and FetchPhase to use this new
abstraction in place of field visitors.  Further work to integrate source loaders will be
done in followup commits.


server/src/main/java/org/elasticsearch/index/fieldvisitor/LeafStoredFieldLoader.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * @return the routing path for the current document
+     */
+    String routing();
+
+    /**
+     * @return stored fields for the current document
+     */
+    Map<String, List<Object>> storedFields();
```
nik9000, line 48: I'd almost prefer callbacks on day. OTOH, I think having these is nicer for caching.

But callbacks don't need the wrapper objects....

server/src/main/java/org/elasticsearch/index/fieldvisitor/StoredFieldLoader.java
```diff
@@ -0,0 +1,173 @@
...
+    private static List<String> fieldsToLoad(boolean loadSource, Set<String> fields) {
+        Set<String> fieldsToLoad = new HashSet<>();
+        fieldsToLoad.add("_id");
+        fieldsToLoad.add("_routing");
+        if (loadSource) {
+            fieldsToLoad.add("_source");
+        }
+        fieldsToLoad.addAll(fields);
+        return fieldsToLoad.stream().sorted().toList();
+    }
```
nik9000, line 52: Maybe shove this at the end of the file so readers see the public interface first.

nik9000 (commented):
Looks right to me. i left like one thing. seems lovely, really.

server/src/main/java/org/elasticsearch/index/fieldvisitor/LeafStoredFieldLoader.java
nik9000: Let's leave this callbacks things for later. It can keep.

nik9000 (approved):

PR #89662, opened 2022-08-26
Centralize the concept of processors configuration

fcofdez:
This commit centralizes the processor count concept into the `Processors` class.
With this change now all the places using a processor count rely on this new class,
such as desired nodes, `node.processors` setting and autoscaling deciders.

- Processor counts are rounded to up to 5 decimal places
- Processors can be represented as doubles

Desired nodes processors were stored as floats, this poses some challenges during
upgrades as once the value is casted to a double, the precision increases and therefore
the number is not the same. In order to allow idempotent desired nodes updates after
upgrades, this commit introduces `DesiredNode#equalsWithProcessorsCloseTo(DesiredNode that, double maxError)`
which allows comparing two desired nodes that differ up to `maxError` in their processor
specification as floats.


fcofdez:

server/src/main/java/org/elasticsearch/cluster/desirednodes/DesiredNodesSettingsValidator.java
```diff
@@ -92,7 +92,8 @@ private void validate(DesiredNode node) {
             int minProcessors = node.roundedDownMinProcessors();
             Integer roundedUpMaxProcessors = node.roundedUpMaxProcessors();
             int maxProcessors = roundedUpMaxProcessors == null ? minProcessors : roundedUpMaxProcessors;
-            Setting.intSetting(NODE_PROCESSORS_SETTING.getKey(), minProcessors, 1, maxProcessors, Setting.Property.NodeScope).get(settings);
+            Setting.doubleSetting(NODE_PROCESSORS_SETTING.getKey(), minProcessors, 1, maxProcessors, Setting.Property.NodeScope)
```
henningandersen, line 95: Should we lower the `minValue` to `Double.MIN_VALUE` or `Processors.MIN_REPRESENTABLE_PROCESSORS`?

server/src/main/java/org/elasticsearch/cluster/metadata/DesiredNode.java
```diff
@@ -354,42 +382,42 @@ public String toString() {
...
-    }
-
-    private static int roundUp(float value) {
-        return (int) Math.ceil(value);
-    }
-
-    private static int roundDown(float value) {
-        return Math.max(1, (int) Math.floor(value));
+    private static boolean invalidNumberOfProcessors(Processors processors) {
+        return processors != null && processors.count() <= 0;
```
henningandersen, line 386: I think this can only be true for `processors.count() == 0` now, but is that even a legal case for `Processors`? Given that we now validate `Processors` separately perhaps this method is no longer needed?

server/src/main/java/org/elasticsearch/common/unit/Processors.java
```diff
@@ -0,0 +1,175 @@
...
+    public int roundUp() {
+        return (int) Math.ceil(count);
+    }
+
+    public int roundDown() {
+        return Math.max(1, (int) Math.floor(count));
+    }
+
+    private static boolean validNumberOfProcessors(double processors) {
+        return Double.isFinite(processors) && processors >= 0.0;
```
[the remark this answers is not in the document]
henningandersen: Hmm, now seeing this is indeed needed for autoscaling. Sad, I wish we could find another way there, but that is for another day.

But can we ensure that only the `ZERO` constant can create a `0.0` value by having a special private constructor?

server/src/main/java/org/elasticsearch/common/unit/Processors.java
```diff
@@ -0,0 +1,175 @@
...
+        } else {
+            this.count = Math.max(
+                MIN_REPRESENTABLE_PROCESSORS,
+                new BigDecimal(count).setScale(NUMBER_OF_DECIMAL_PLACES, RoundingMode.HALF_UP).doubleValue()
+            );
+        }
+    }
+
+    @Nullable
+    public static Processors of(Double count) {
```
henningandersen, line 56: I wonder if we should move all `Processors` construction to use an `of` method, just to simplify and avoid mistakes? That `of` method could hold the validation too then, helping on the `ZERO` part I commented on in another comment (though a minor point).

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/existence/FrozenExistenceDeciderService.java
```diff
@@ -56,7 +56,7 @@ public AutoscalingDeciderResult scale(Settings configuration, AutoscalingDecider
             builder.total(MINIMUM_FROZEN_STORAGE, MINIMUM_FROZEN_MEMORY, null);
             builder.node(MINIMUM_FROZEN_STORAGE, MINIMUM_FROZEN_MEMORY, null);
         } else {
-            builder.total(0L, 0L, 0f);
+            builder.total(ByteSizeValue.ZERO, ByteSizeValue.ZERO, null);
```
henningandersen, line 59: Should we use `Processors.ZERO` instead of `null`?

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/AutoscalingTestCase.java
```diff
@@ -153,6 +154,10 @@ public static AutoscalingPolicy randomAutoscalingPolicyOfName(final String name)
         return new AutoscalingPolicy(name, randomRoles(), randomAutoscalingDeciders());
     }
 
+    public static Processors randomProcessors() {
+        return Processors.of(randomInt(128) + randomDouble());
```
henningandersen, line 158: nit: can we use `randomDoubleBetween`?

henningandersen (commented):
I went through the production code and left comments.

server/src/main/java/org/elasticsearch/monitor/os/OsInfo.java
```diff
@@ -8,19 +8,22 @@
...
 
 import java.io.IOException;
 
 public class OsInfo implements ReportingService.Info {
+    private static final Version DOUBLE_PRECISION_ALLOCATED_PROCESSORS_SUPPORT = Version.V_8_5_0;
 
     private final long refreshInterval;
     private final int availableProcessors;
-    private final int allocatedProcessors;
+    private final Processors allocatedProcessors;
```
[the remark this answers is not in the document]
fcofdez: Good point, I'll serialize the rounded allocated processors in x-content. I wonder if we want to serialize the fractional value in an additional field.

server/src/main/java/org/elasticsearch/action/admin/cluster/desirednodes/TransportUpdateDesiredNodesAction.java
```diff
@@ -39,6 +39,7 @@
 
 public class TransportUpdateDesiredNodesAction extends TransportMasterNodeAction<UpdateDesiredNodesRequest, UpdateDesiredNodesResponse> {
     private static final Logger logger = LogManager.getLogger(TransportUpdateDesiredNodesAction.class);
+    private static final double MAX_DELTA_PROCESSORS = 7E-5;
```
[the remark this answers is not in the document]
fcofdez: I run quite a few experiments in `DesiredNodeTests#testFloatProcessorsConvertedToDoubleAreCloseToEqual` starting from `1E-5` but that failed quite a few times. I increased it by `1E-5` at a time until the test stopped failing. I agree that's not super scientific method.

server/src/main/java/org/elasticsearch/common/unit/Processors.java
```diff
@@ -0,0 +1,175 @@
...
+            throw new IllegalArgumentException(
+                format(Locale.ROOT, "Only a positive number of [%s] are allowed and [%f] was provided", parser.currentName(), count)
+            );
+        }
+        return new Processors(count);
+    }
+
+    @Override
+    public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+        return builder.value(count);
```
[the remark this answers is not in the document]
fcofdez: `count` is already truncated to 5 decimal places in the constructor. Additionally, there's a test (`ProcessorsTests#testTruncatesAfterFiveDecimalPlaces`) that checks that the string representation has up to 5 decimal places.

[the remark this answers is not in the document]
henningandersen: I tried to read `Double.toString`. IIUC, we would get scientific notation for numbers less than 1e-3? Maybe not a problem though, I suppose it is json compatible.

I also worry that there is some number, where even when rounded to 5 decimal places we need more decimals to represent it uniquely. I do not have an example, but did you run the test thousands of times to check?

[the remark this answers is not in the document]
fcofdez: > I tried to read Double.toString. IIUC, we would get scientific notation for numbers less than 1e-3? Maybe not a problem though, I suppose it is json compatible.


According to [RFC-8259](https://www.rfc-editor.org/rfc/rfc8259) the scientific notation should be supported by any standard parser.

> I also worry that there is some number, where even when rounded to 5 decimal places we need more decimals to represent it uniquely. I do not have an example, but did you run the test thousands of times to check?

I improved the test and I've run it 10K times.

server/src/main/java/org/elasticsearch/common/unit/Processors.java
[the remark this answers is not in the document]
henningandersen: Can we remove `0.0` as a valid number now?

```suggestion
        return Double.isFinite(processors) && processors > 0.0;
```

server/src/test/java/org/elasticsearch/cluster/metadata/DesiredNodeTests.java
```diff
@@ -225,16 +226,63 @@ public void testDesiredNodeIsCompatible() {
...
+            desiredNode1 = new DesiredNode(settings, desiredNodes1ProcessorsRange, memory, storage, Version.CURRENT);
+            desiredNode2 = new DesiredNode(settings, desiredNodes2ProcessorsRange, memory, storage, Version.CURRENT);
         }
 
-        return result;
+        assertThat(desiredNode1.equalsWithProcessorsCloseTo(desiredNode2), is(isEqualOrCloseTo));
+    }
+
+    private double randomNumberOfProcessors() {
+        return randomDoubleBetween(Math.ulp(0.0), 512.99999999, true);
```
henningandersen, line 281: Is `Math.ulp(0.0)` not the same as `Double.MIN_VALUE`? I'd prefer to use the latter then.

server/src/test/java/org/elasticsearch/cluster/metadata/DesiredNodesTestCase.java
```diff
@@ -184,4 +184,8 @@ public static DiscoveryNode newDiscoveryNode(String nodeName) {
             Version.CURRENT
         );
     }
+
+    public static double randomNumberOfProcessors() {
+        return randomDoubleBetween(Math.ulp(0.0), 512.99999999, true);
```
henningandersen, line 189: Also use `Double.MIN_VALUE` here?

server/src/test/java/org/elasticsearch/common/unit/ProcessorsSerializationTests.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class ProcessorsSerializationTests extends AbstractWireSerializingTestCase<Processors> {
+    @Override
+    protected Writeable.Reader<Processors> instanceReader() {
+        return Processors::readFrom;
+    }
+
+    @Override
+    protected Processors createTestInstance() {
+        return Processors.of(randomDoubleBetween(Math.ulp(0.0), 512.99999999, true));
```
henningandersen, line 24: Also use `Double.MIN_VALUE` here?

server/src/test/java/org/elasticsearch/common/unit/ProcessorsTests.java
```diff
@@ -0,0 +1,103 @@
...
+            assertThat(exponent, is(lessThan(0)));
+            decimalPlaces = Math.abs(exponent);
+        } else {
+            decimalPlaces = processorsString.length() - processorsString.indexOf(".") - 1;
+        }
+        assertThat(decimalPlaces, is(lessThanOrEqualTo(Processors.NUMBER_OF_DECIMAL_PLACES)));
+    }
+
+    public void testRounding() {
+        {
```
henningandersen, line 37: Can we add 1-2 more cases:

1. A number higher than 1.something, for instance 10.1.
2. A number that is already an integer, for instance 10.

server/src/test/java/org/elasticsearch/common/unit/ProcessorsTests.java
```diff
@@ -0,0 +1,103 @@
...
+import org.elasticsearch.test.ESTestCase;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.greaterThan;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.lessThan;
+import static org.hamcrest.Matchers.lessThanOrEqualTo;
+
+public class ProcessorsTests extends ESTestCase {
+    public void testTruncatesAfterFiveDecimalPlaces() {
```
henningandersen, line 20: For completeness, can we add a test that `Processors.of(value).count() == value`?

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/capacity/AutoscalingCapacityWireSerializationTests.java
```diff
@@ -57,7 +59,7 @@ protected AutoscalingCapacity mutateInstance(AutoscalingCapacity instance) {
                     instance.total().memory(),
                     hasAllMetrics && (instance.node() == null || instance.node().processors() == null) && randomBoolean()
                         ? null
-                        : randomIntBetween(1, 64) + Optional.ofNullable(instance.total().processors()).orElse(0f)
+                        : Optional.ofNullable(instance.total().processors()).map(p -> p.plus(randomProcessors())).orElse(randomProcessors())
```
henningandersen, line 62: I wonder if `p.plus` ensures a different value here? If `randomProcessors` becomes `Double.MIN_VALUE` and `p` is already a value with `Math.ulp(value) > Double.MIN_VALUE`, the resulting `Processors` object could be identical, breaking the test?

Similar comment for other uses of `plus` in this class.

henningandersen (approved):
LGTM.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/capacity/AutoscalingCapacityWireSerializationTests.java
fcofdez: `Processors#plus` adds the rounded up values. In the case you present we would be rounding up both processors to the minimal representable processor `1E-5` and the result should be `2E-5`.

henningandersen: Makes sense, thanks.

fcofdez:
Thanks Henning!

PR #89667, opened 2022-08-26
Operator/role mapping

grcevski:
This PR adds support for /_security/role_mapping for file based settings. We create similar reserved cluster state information about the role mappings, even though they are not part of the cluster state.

Since the role_mappings are not saved in the cluster state, we needed to create separate reserved state action wrapper for any non-cluster state updates.


server/src/main/java/org/elasticsearch/action/support/ReservedStateAwareHandledTransportAction.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.Set;
+
+import static org.elasticsearch.reservedstate.service.ReservedClusterStateService.validateForReservedState;
+
+/**
+ * An extension of the {@link HandledTransportAction} class, which wraps the doExecute call with a check for clashes
+ * with the reserved cluster state.
+ */
+public abstract class ReservedStateAwareHandledTransportAction<Request extends ActionRequest, Response extends ActionResponse> extends
```
grcevski, line 28: Since the role mappings don't use TransportMasterNodeAction (which is cluster state specific) we add this small wrapper class to ensure the reserved state information is checked before allowing REST actions to create/update/delete role mappings.

server/src/main/java/org/elasticsearch/action/support/master/TransportMasterNodeAction.java
```diff
@@ -177,37 +175,23 @@ protected Set<String> modifiedKeys(Request request) {
...
-                errors.add(format("[%s] set as read-only by [%s]", String.join(", ", conflicts), metadata.namespace()));
-            }
-        }
-
-        if (errors.isEmpty() == false) {
-            throw new IllegalArgumentException(
-                format("Failed to process request [%s] with errors: [%s]", request, String.join(", ", errors))
-            );
-        }
+        ReservedClusterStateService.validateForReservedState(state, handlerName.get(), modifiedKeys(request), request.toString());
```
grcevski, line 182: Since this check is common now, I moved it to ReservedClusterStateService to be reused.

server/src/main/java/org/elasticsearch/action/support/master/TransportMasterNodeAction.java
```diff
@@ -177,37 +175,23 @@ protected Set<String> modifiedKeys(Request request) {
...
     }
 
     // package private for testing
-    boolean supportsImmutableState() {
+    boolean supportsReservedState() {
```
grcevski, line 186: These are simple renames, which were somehow missed before. We did rename from immutable to reserved.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedClusterStateService.java
```diff
@@ -270,4 +271,28 @@ private void addStateHandler(String key, Set<String> keys, LinkedHashSet<String>
...
     }
+
+    /**
+     * Helper method that verifies for key clashes on reserved state updates
+     * @param state the current cluster state
+     * @param handlerName the name of the reserved state handler related to this implementation
+     * @param modified the set of modified keys by the related request
+     * @param request a string representation of the request for error reporting purposes
+     */
+    public static void validateForReservedState(ClusterState state, String handlerName, Set<String> modified, String request) {
```
grcevski, line 282: Moved here from TransportMasterNodeAction for reuse.

x-pack/plugin/core/src/main/java/org/elasticsearch/xpack/core/security/action/rolemapping/PutRoleMappingRequest.java
```diff
@@ -162,4 +162,16 @@ public void writeTo(StreamOutput out) throws IOException {
     public ExpressionRoleMapping getMapping() {
         return new ExpressionRoleMapping(name, rules, roles, roleTemplates, metadata, enabled);
     }
+
+    public static PutRoleMappingRequest fromMapping(ExpressionRoleMapping mapping) {
```
grcevski, line 166: Helper method used by the reserved handler and test to create the request from the parsed mappings.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/rolemapping/ReservedRoleMappingAction.java
```diff
@@ -0,0 +1,130 @@
...
+        this.roleMappingStore = roleMappingStore;
+    }
+
+    @Override
+    public String name() {
+        return NAME;
+    }
+
+    @SuppressWarnings("unchecked")
+    public Collection<PutRoleMappingRequest> prepare(Object input) {
```
grcevski, line 58: We convert from expression role mappings to a request to satisfy the native store API contract.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/LocalReservedSecurityStateHandlerProvider.java
```diff
@@ -0,0 +1,40 @@
...
+import org.elasticsearch.reservedstate.ReservedClusterStateHandler;
+import org.elasticsearch.reservedstate.ReservedClusterStateHandlerProvider;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * Mock Security Provider implementation for the {@link ReservedClusterStateHandlerProvider} service interface
+ */
+public class LocalReservedSecurityStateHandlerProvider implements ReservedClusterStateHandlerProvider {
```
grcevski, line 20: Mock version of the SPI handler provider so I could test properly with integration tests.

grcevski:

ywangd (commented):
I haven't followed closely on the work of File based settings[^1]. I still need to get speed up with prior changes. So please bear with me if I have any misunderstanding of the current changes.

I browsed through the changes to the main code. Besides a few comments below on specific items, I do have a more broad question: Have we considered actually storing these reserved role mappings in cluster state instead of the security index? 

This PR basically adds a file interface to the native role mapping store (which is the security index). As such, it is not *truly* reserved because users can still modify them by directly manipulating the index or snapshot/restore. Working with the index also makes it hard to achieve atomicity for clean abort and rollback. Are these concerns at all?

The role mappings are already backed by different storages, file and index. It is not too wild to have a 3rd storage type of cluster state which is likely to solve the above two issues, "truly" reserved and atomicity. The index storage mostly make search easier. But we don't really leverage that for role mappings. Role mappings stored in index are also gettable via APIs. We can make it read from the cluster state as well. But do we really need reserved role mappings to be gettable via the same API? ESS's usage for reserved role mappings is likely an implementation detail and should not be concerned with users. Expanding from this, it is even arguable that we don't have to prevent users from creating role mappings that have the same names as the reserved ones. Because they are backed by different storages, they are naturally separate and we don't rely on names for perform the actual role mapping operation. 

[^1]: Based on what it does, shouldn't it be called File based "state" rather than "settings"?

grcevski:
> I browsed through the changes to the main code. Besides a few comments below on specific items, I do have a more broad question: Have we considered actually storing these reserved role mappings in cluster state instead of the security index?

Yes, this particular question did come up during the offline design discussions and we decided to keep the existing approach for now, while adding the file based state manipulation. As you say, perhaps we should be storing these in the cluster state instead, but we decided to keep it as is and we can move to cluster storage in the future.

> This PR basically adds a file interface to the native role mapping store (which is the security index). As such, it is not _truly_ reserved because users can still modify them by directly manipulating the index or snapshot/restore. Working with the index also makes it hard to achieve atomicity for clean abort and rollback. Are these concerns at all?

Snapshot restore should be fine, we allow the restores to happen and if there is file based state we call the service to re-apply everything it has currently. Direct index manipulation isn't something I thought of here, can users directly manipulate the security index? I know they can delete it, which would mean we have to somehow re-process the file based state. Do you have any suggestions on how to make this more solid, if we were to keep storing the mappings in the security index?

The lack of atomicity was also brought up in the design discussion and we agreed that it was fine for the purpose we have here, cluster state update shouldn't fail, but your point is valid. The way I have done this can be improved. I can let all of the handlers to get processed without any updates to the native store and only then modify the role mappings, when we are sure all handlers can proceed. 

> ## Footnotes
> 1. Based on what it does, shouldn't it be called File based "state" rather than "settings"? [↩](#user-content-fnref-1-4022792b6f091e78d7037c43b54fcfd9)

Agree, I'll try to rename where possible. Settings is sort of a left over from when this feature was originally brought up as a way to update the cluster settings.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/ReservedSecurityStateHandlerProvider.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Security Provider implementation for the {@link ReservedClusterStateHandlerProvider} service interface
+ */
+public class ReservedSecurityStateHandlerProvider implements ReservedClusterStateHandlerProvider {
+    private final Security plugin;
+
+    public ReservedSecurityStateHandlerProvider() {
+        throw new IllegalStateException("Provider must be constructed using PluginsService");
+    }
```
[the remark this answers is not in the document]
grcevski: Since security is a module, Java forces us to declare the no-args constructor when SPI is used. It's not ideal, but we should resolve all these when the new plugin infrastructure is finished.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/rolemapping/ReservedRoleMappingAction.java
```diff
@@ -0,0 +1,130 @@
...
+        List<ExpressionRoleMapping> roleMappings = (List<ExpressionRoleMapping>) input;
+        List<PutRoleMappingRequest> requests = roleMappings.stream().map(rm -> PutRoleMappingRequest.fromMapping(rm)).toList();
+
+        for (var request : requests) {
+            var exception = request.validate();
+            if (exception != null) {
+                throw exception;
+            }
+        }
```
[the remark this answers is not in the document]
grcevski: Good point, I'll collect all errors and throw at the end if any.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/rolemapping/ReservedRoleMappingAction.java
```diff
@@ -0,0 +1,130 @@
...
+
+        return requests;
+    }
+
+    @Override
+    public TransformState transform(Object source, TransformState prevState) throws Exception {
+        var requests = prepare(source);
+
+        ClusterState state = prevState.state();
```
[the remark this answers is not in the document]
ywangd: I overlooked the `keys`. Thanks for the explanation. For my education purpose: I assume these keys will at the end be stored as part of cluster state? Likley all the keys from different handlers are updated in a single batch once all `transform` is completed?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/rolemapping/ReservedRoleMappingAction.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    public Collection<PutRoleMappingRequest> prepare(Object input) {
+        List<ExpressionRoleMapping> roleMappings = (List<ExpressionRoleMapping>) input;
+        List<PutRoleMappingRequest> requests = roleMappings.stream().map(rm -> PutRoleMappingRequest.fromMapping(rm)).toList();
+
+        var exceptions = new ArrayList<String>();
+        for (var request : requests) {
+            var exception = request.validate();
```
ywangd, line 65: It is possible that we will use some reserved metadata to indicate that these role mappings are special, e.g. `{"_reserved": true}` or `{"_managed": true}`. But this will fail to pass the `validate` call because it does not allow metadata keys with leading underscore. This is mainly to prevent defining such metadata using the REST API and should not be applicable here. We might want to have a parameterized overload of the `validate` method so it can forgo the check for metadata keys.

server/src/main/java/org/elasticsearch/reservedstate/ReservedClusterStateHandler.java
```diff
@@ -60,6 +62,25 @@
...
+     *
+     * We loop through the postTransform method calls after all transform methods are called.
+     *
+     * @param source The parsed information specific to this handler from the combined cluster state content
+     * @param clusterState The final cluster state from all handler calls to transform
+     * @param previousKeys The previous keys saved for this handler
+     * @return The current keys set by this handler
+     * @throws Exception
+     */
+    default Set<String> postTransform(Object source, ClusterState clusterState, Set<String> previousKeys) throws Exception {
```
ywangd, line 80: I think the signature of this method be simplifed to
```
Set<String> postTransform(Object source, TransformState prevState)
```
since ClusterState + Keys = TransformState?

The contract of this method seems a bit odd to me. I almost feel it should be returned as a runnable/callable for calling `transform`. The main reason is what needs to be done with `postTransform` (update, delete) is already known at `transform` time. The content is also known. `postTransform` only needs to perform the actions that are outside of clusterstate. So we don't really need to pass it the clusterstate and keys again. In fact, passing them again here raises the question about whether they should be same or different compared to those passed into `transform`. The name `previousKeys` is also a bit fuzzy because it can be current or previous?

That said, the current code works for now and I haven't involved in the design of file based state feature. So you don't have to do anything. Mainly food for thoughts. Thanks!

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedStateUpdateTask.java
```diff
@@ -100,6 +101,36 @@ protected ClusterState execute(final ClusterState currentState) {
             }
         }
 
+        checkAndThrowOnError(errors, currentState, reservedStateVersion);
+
+        // Run the postTransform loop for any handlers that need to perform updates to other things than the cluster state
+        for (var handlerName : orderedHandlers) {
+            ReservedClusterStateHandler<?> handler = handlers.get(handlerName);
+            try {
+                // We first try to fetch the existing keys from what we have in the builder if transform saved any keys
```
ywangd, line 110: Nit: the word existing seems a bit confusing since there is a variable called `existingMetadata` and it is not the same thing. Would it be better to say `.. fetch the current keys ...".

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/rolemapping/ReservedRoleMappingAction.java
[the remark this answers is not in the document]
grcevski: Yes, that's correct. All of the keys that the handlers report are then stored as reserved keys in the cluster state.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedStateUpdateTask.java
```diff
@@ -100,6 +101,36 @@ protected ClusterState execute(final ClusterState currentState) {
...
+            } catch (Exception e) {
+                errors.add(format("Error processing %s state change: %s", handler.name(), stackTrace(e)));
+            }
+        }
+
+        // Most error should be caught in the transform phase, however we check if the post transform encountered any
+        checkAndThrowOnError(errors, currentState, reservedStateVersion);
+
+        // remove the last error if we had previously encountered any
+        reservedMetadataBuilder.errorMetadata(null);
```
ywangd, line 124: I understand it is existing code. But I am curious about what other errors we could have here since we already check and throw previously?

server/src/main/java/org/elasticsearch/cluster/metadata/ReservedStateMetadata.java
```diff
@@ -283,6 +283,13 @@ public Builder putHandler(ReservedStateHandlerMetadata handler) {
             return this;
         }
 
+        /**
+         * Returns the current handler metadata stored in the builder
+         */
+        public ReservedStateHandlerMetadata getHandler(String handlerName) {
```
ywangd, line 289: Nit: maybe this is better named `getHandlerMetadata`?

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/rolemapping/ReservedRoleMappingAction.java
```diff
@@ -0,0 +1,141 @@
...
+                public void onResponse(Boolean aBoolean) {}
+
+                @Override
+                public void onFailure(Exception e) {
+                    throw new IllegalStateException("Error deleting role mapping [" + mappingToDelete + "]", e);
+                }
+            });
+        }
+
+        return Collections.unmodifiableSet(entities);
```
ywangd, line 121: Deleting role mapping (as well as updating) is async. So they could fail even when the return statement successfully executed. This means ES could report it successfully updated the file-based-settings while the role mapping persistence could fail? 

I understand that we accept that cluster and index operations are separate and can fail independently. But I think it should be: "When the overall process is reported as failure, there can be part of the update that have been successful."

But **not** "When the overall process is reported as success, there can be part of the update that have failed." Because this would prevent the caller from retrying on failure and leave the system in an inconsistent state indefinitely.

Concretely, I am suggesting that we should handle the async errors (inside listeners) before returning.

ywangd (commented):
I have a new question about how to handle failures in updating and deleting role mapping operations. My thinking is that any failure should cause the overall update to fail. Please see below comment for details.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/rolemapping/ReservedRoleMappingAction.java
grcevski: Not a bad idea, I can allow for metadata to be excluded from validation, so that we can support special metadata keys for these mappings.

server/src/main/java/org/elasticsearch/reservedstate/ReservedClusterStateHandler.java
grcevski: I agree, I'll simplify the method interface. I like your idea of callables, let me explore it a bit to see if it simplifies the whole thing.

server/src/main/java/org/elasticsearch/reservedstate/service/ReservedStateUpdateTask.java
grcevski: So when we fail some file processing, we report the errors in the logs, however I also have a cluster state update task that stores metadata into the cluster state indicating the errors encountered by the most recent processing.

Having this error metadata helps with building external support for verifying if the supplied settings.json file can be applied. Namely, instead of building pipelines to extract log information and look for specific text patters, we can build REST API that extracts the latest file processing information from the cluster state. I don't have this API yet, but it's in plan to be built.

So when we successfully process a file, we nullify the last error state from prior processing, meaning, the latest file processed (with certain version) was saved correctly.

For example:

```
file settings version: 22 -> saved OK, no error metadata, last applied settings version is 22
file settings version: 23 -> error while processing file, last applied settings version is 22, error version: 23, errors:....
file settings version: 24 -> saved OK, cleanup error metadata, last applied settings version is 24, error: null
```

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/rolemapping/ReservedRoleMappingAction.java
grcevski: You are right, this is a bug. I should be waiting for the async actions to complete.

grcevski:
Thanks again for a detailed review Yang, I'll follow-up with changes to address the feedback.

server/src/main/java/org/elasticsearch/cluster/metadata/ReservedStateMetadata.java
grcevski: I'll address this in a followup, it will increase the size of this PR because the method is used in many tests.

x-pack/plugin/security/src/test/java/org/elasticsearch/xpack/security/action/reservedstate/ReservedRoleMappingActionTests.java
```diff
@@ -0,0 +1,215 @@
...
+        prevState = updatedState;
+        updatedState = processJSON(action, prevState, json);
+        assertThat(updatedState.keys(), containsInAnyOrder("everyone_kibana", "everyone_fleet"));
+
+        updatedState = processJSON(action, prevState, emptyJSON);
+        assertThat(updatedState.keys(), empty());
+    }
+
+    @SuppressWarnings("unchecked")
+    public void testPostTransformWaitsOnAsyncActions() throws Exception {
```
grcevski, line 126: I added a test here that ensures we wait properly for any async errors, in both put and delete.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/integration/RoleMappingFileSettingsIT.java
```diff
@@ -0,0 +1,250 @@
...
+        ReservedStateMetadata reservedState = clusterStateResponse.getState()
+            .metadata()
+            .reservedStateMetadata()
+            .get(FileSettingsService.NAMESPACE);
+
+        ReservedStateHandlerMetadata handlerMetadata = reservedState.handlers().get(ReservedRoleMappingAction.NAME);
+
+        assertThat(handlerMetadata.keys(), allOf(notNullValue(), containsInAnyOrder("everyone_kibana", "everyone_fleet")));
+
+        var request = new GetRoleMappingsRequest();
```
grcevski, line 159: This file is more less the same from before, I just added an extra validation step here that ensures we saved everything in the role mappings as well as the reserved state.

server/src/main/java/org/elasticsearch/reservedstate/ReservedClusterStateHandler.java
grcevski: I ended up using your idea for the post transform, looks much nicer. I added a post transform supplier option in the transform state, which is then simply called when the cluster state is saved.

grcevski:

ywangd (commented):
This is looking great. I don't think I'll find anything major. There are places that can use future refactoring. But I think the changes are plenty already for this PR. So we can wait. 

Since there are many changes, I just need some more time to double check and read through the tests. But I'd like to make it clear that I am progressing on it by providing some minor optional feedback.

I also have a question: Will there only be a single file-state-update task to run at a time, i.e. it is not possible that two threads observed file changes and start parallel updating?

ywangd:
I think we could use a few more tests. For example, making sure `TransportPutRoleMappingAction` actually respects the reserved keys. Or maybe this can be tested with `ReservedStateAwareHandledTransportAction`? Or maybe it is already tested somewhere else since it is extraction refactoring? It's not entirely clear to me.

grcevski:
> I also have a question: Will there only be a single file-state-update task to run at a time, i.e. it is not possible that two threads observed file changes and start parallel updating?

File based settings doesn't allow more than one update to run at the same time. So while no synchronization exists in the reserved state service, the file based settings service uses a latch to wait on until all async actions inside the reserved state service to finish, before it checks for new file updates. While an update is going on on certain file contents, the JDK watcher service can queue new file update changes, which will get processed one by one after the previous in-flight update finishes. We have a test for some of this logic in FileSettingsServiceTest.testStopWorksInMiddleOfProcessing and testStopWorksIfProcessingDidntReturnYet.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/rolemapping/ReservedRoleMappingAction.java
```diff
@@ -0,0 +1,147 @@
...
+    public TransformState transform(Object source, TransformState prevState) throws Exception {
+        // We execute the prepare() call to catch any errors in the transform phase.
+        // Since we store the role mappings outside the cluster state, we do the actual save with a
+        // non cluster state transform call.
+        @SuppressWarnings("unchecked")
+        var requests = prepare((List<ExpressionRoleMapping>) source);
+        return new TransformState(prevState.state(), prevState.keys(), ((l) -> nonStateTransform(requests, prevState, l)));
+    }
+
+    private Void nonStateTransform(
```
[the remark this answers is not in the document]
grcevski: 100% I don't remember now why I made it Void, must've been an iteration I had when the consumer was a function.

grcevski:
> I think we could use a few more tests. For example, making sure `TransportPutRoleMappingAction` actually respects the reserved keys. Or maybe this can be tested with `ReservedStateAwareHandledTransportAction`? Or maybe it is already tested somewhere else since it is extraction refactoring? It's not entirely clear to me.

Some of this is tested in RoleMappingFileSettingsIT.assertRoleMappingsSaveOK. At the end the method checks that a REST call cannot update any of the reserved role mappings, but this is a good call-out, I'll add explicit tests for validation of the keys in TransportPutRoleMappingAction, similarly as I've done for TransportPutLifecycleActionTests.

I also have tests that perform the validation on TransportMasterNodeAction (TransportMasterNodeActionTests.testRejectImmutableConflictClusterStateUpdate), but not on the ReservedStateAwareHandledTransportAction which I introduced in the refactor. I'll add tests here too, thanks for bringing it up.

x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/action/rolemapping/ReservedRoleMappingAction.java
```diff
@@ -0,0 +1,145 @@
...
+        for (var request : requests) {
+            // File based defined role mappings are allowed to use MetadataUtils.RESERVED_PREFIX
+            var exception = request.validate(false);
+            if (exception != null) {
+                exceptions.add(exception.getMessage());
+            }
+        }
+
+        if (exceptions.isEmpty() == false) {
+            throw new IllegalStateException(String.join(", ", exceptions));
```
ywangd, line 72: I'd go with
```suggestion
            var illegalStateException = new IllegalStateException("error on validating put role mapping requests");
            exceptions.forEach(illegalStateException::addSuppressed);
```

I am a bit surprised about the choice of IllegalStateException type. Errors here are caused by users which is more of IllegalArgument. But I notice this error type is already used elsewhere. So I am also OK with it.

x-pack/plugin/security/src/internalClusterTest/java/org/elasticsearch/integration/RoleMappingFileSettingsIT.java
```diff
@@ -0,0 +1,338 @@
...
+            var savedClusterState = setupClusterStateListener(internalCluster().getMasterName());
+            writeJSONFile(internalCluster().getMasterName(), testJSON);
+
+            assertRoleMappingsSaveOK(savedClusterState.v1(), savedClusterState.v2());
+        } finally {
+            logger.info("---> cleanup cluster settings...");
+            var savedClusterState = setupClusterStateListenerForCleanup(internalCluster().getMasterName());
+            writeJSONFile(internalCluster().getMasterName(), emptyJSON);
+            boolean awaitSuccessful = savedClusterState.v1().await(20, TimeUnit.SECONDS);
+            assertTrue(awaitSuccessful);
```
ywangd, line 273: Once the empty JSON file is processed, do the role mappings get deleted? If so, I think we should assert the role mappings are now gone and nothing is reserved.

ywangd (approved):
LGTM. Thanks for the iterations. I appreciate your patience and dedication to get this across the line. I learnt a lot in the process.

It would be nice if we could have tests for following two scenarios:
1. Role mappings pass validation but fail to write to index. A easy way to achieve this is to close the security index. 
2. Role mappings succeed, but final cluster state update fails. I am not sure how easy or hard to force cluster state failure at that point.

Happy for them to be follow-ups.

Also my previous experience with Java's WatchService is that it does not work if the file system is NFS. Maybe this is no longer the case? Or maybe this feature does not need work with NFS? Anyhow, even it is an issue, it is *not* part of this PR. I am bringing it up just in case.

grcevski:
> It would be nice if we could have tests for following two scenarios:
> 
> 1. Role mappings pass validation but fail to write to index. A easy way to achieve this is to close the security index.
> 2. Role mappings succeed, but final cluster state update fails. I am not sure how easy or hard to force cluster state failure at that point.

Good idea, I'm going to make a test for 1. I think I can test 2. with some mocked methods, but I guess in that case we are simply testing the flaw in our atomicity across the two stores. Perhaps we can add a test here if we ever fix this situation in the future.

 
> Also my previous experience with Java's WatchService is that it does not work if the file system is NFS. Maybe this is no longer the case? Or maybe this feature does not need work with NFS? Anyhow, even it is an issue, it is _not_ part of this PR. I am bringing it up just in case.

Thanks for this info, I hadn't tested with an NFS file system. I think it's not part of the requirements, but we should definitely document is as a limitation. I'll follow-up with an investigation here.

grcevski:
Thanks for all the help Yang!

PR #89668, opened 2022-08-26
Big arrays sliced from nettey buffers (int)

nik9000:
This teaches `IntArray` about our serialization.
The interesting bit here is that reading slices the reference to the
underlying buffer rather than copying. That reference can be retained as
long as it's needed, holding the underlying buffer open until the
`IntArray` is `close`d.

This should allow aggregations to send dense representations between
nodes with one fewer copy operation.


server/src/main/java/org/elasticsearch/common/bytes/ReleasableIntArray.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public void fill(long fromIndex, long toIndex, int value) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void set(long index, byte[] buf, int offset, int len) {
+        throw new UnsupportedOperationException();
```
nik9000, line 50: All this throwing is kind of annoying, but we're really only going to need the `get` methods. We could make another interface that doesn't have these methods. Maybe that's for another PR though.

nik9000 (commented):
I picked `IntArray` here because `BytesReference` has a `getInt` method. It looks like that method does some bit shifting. We could probably get away with avoiding the bit shifting most of the time, but that seems like a problem for another day.

server/src/main/java/org/elasticsearch/common/bytes/ReleasableIntArray.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public long ramBytesUsed() {
+        return 0; // NOCOMMIT How much? 2 objects? I think we'd double count if we used the usage of the bytes ref
```
nik9000, line 55: The answer to this question sort of depends on what `ramBytesUsed` will be used for.

server/src/main/java/org/elasticsearch/common/bytes/ReleasableIntArray.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public void close() {
+        ref.decRef();
+        ref = null;
```
nik9000, line 61: @not-napoleon liked the `ref = null` thing here. I think I prefer making the thing final and letting it get GCed, but I'm fine either way.

server/src/main/java/org/elasticsearch/common/io/stream/StreamOutput.java
```diff
@@ -872,6 +873,17 @@ public void writeVIntArray(int[] values) throws IOException {
         }
     }
 
+    public void writeBigIntArray(IntArray values) throws IOException {
+        // TODO do we need to specialize this for ReleasableIntArray
+        if (values.size() > Integer.MAX_VALUE / Integer.BYTES) {
+            throw new UnsupportedOperationException(); // NOCOMMIT What do?
+        }
```
nik9000, line 880: We rarely store anything close to this many ints in aggs land. We're not going to do this. That's 2gb of data. 536 millions buckets worth of integers. 268 million buckets of longs or doubles. I suppose one day we'd love to handle this, but I can't imagine our strategy of slamming all of them over the wire at once is really compatible with that now. I'd saw we're fairly safe with a limit like this.

not-napoleon (commented):
For context, this is a step in the larger aggregations memory story we're working on currently.  We are migrating away from large collections of small `InternalAggregation` objects, which each managed a single bucket, in favor of fewer, larger objects which each manage an entire node's result set for a given aggregation.  These objects will be backed by `BigArrays`, which will be built on the data nodes and deserialized on the coordinating nodes.  Something along the lines of this PR will let us avoid copying huge blocks of data when doing that deserialization.

server/src/main/java/org/elasticsearch/common/io/stream/StreamOutput.java
```diff
@@ -872,6 +873,17 @@ public void writeVIntArray(int[] values) throws IOException {
...
+        writeVInt((int) values.size() * Integer.BYTES);
```
nik9000, line 881: We need to think about size here - the arrays are often oversized but we will have a precise size when we're ready to write. If we write the oversized array it could be twice as big.

server/src/main/java/org/elasticsearch/common/bytes/ReleasableBytesReference.java
```diff
@@ -129,6 +130,7 @@ public long ramBytesUsed() {
     @Override
     public StreamInput streamInput() throws IOException {
         assert hasReferences();
+        // NOCOMMIT why doesn't this inc the ref?
```
original-brownbear, line 133: There's no particular reason really and I agree that this might not be optimal. We could make it increment the ref (and decrement on close) and I bet we could use that to clean up quite a bit of code but all the existing code is written under the assumption that all the lifecycle lives with the bytes reference and that's a bigger task to change without introducing leaks I think.

server/src/main/java/org/elasticsearch/common/bytes/ReleasableIntArray.java
original-brownbear: The nulling shouldn't really get us anything here I think outside of making us resilient to double closing. Memory wise it won't help anything with pooled buffers backing the `ref` its bytes will be reused regardless of whether or not `ref` has been GCed.
I somewhat like `final` better because it doesn't introducing a non-final field to an object which might be accessed from multiple threads.

server/src/main/java/org/elasticsearch/common/io/stream/StreamOutput.java
```diff
@@ -872,6 +873,17 @@ public void writeVIntArray(int[] values) throws IOException {
...
+        for (int i = 0; i < values.size(); i++) {
```
original-brownbear, line 882: This seems needlessly expensive when we already have the byte representation for the implementation above. Should we maybe add a `writeTo` to `IntArray` (i.e. make it writable) instead to be able to leverage those bytes?

original-brownbear (commented):
Thanks Nik this is a really cool idea, added a couple comments :)

server/src/main/java/org/elasticsearch/common/bytes/ReleasableBytesReference.java
nik9000: Yeah - I don't want to change this as part of this PR, but it sure confused the hell out of me.

server/src/main/java/org/elasticsearch/common/io/stream/StreamOutput.java
nik9000: Yeah!

nik9000 (commented):
So I've had a bit of a think about this and wonder if we're not better of making a new interface - one that just reads. And to have two implementations of that interface - one that wraps our `IntArray` from `BigArrays` and one that wraps the netty buffer. But, like, that's around the edges. It sounds like folks think this is generally a good idea. I'll try and iterate a bit more.

nik9000:
@original-brownbear could you have another look at this? One thing that's really interesting is endianness. BigArrays always runs in platform-native endianness which is *usually* little endian. `BytesReference#getInt` is big endian. I'd like to keep those two things true. I was thinking we could lock the wire to little endian. That way in the normal case we get to send the data in the native order most of the time. We could add something like `MethodHandles.byteArrayViewVarHandle(int[].class, ByteOrder.LITTLE_ENDIAN);` in `BytesArray`. That *should* let us read those ints while reducing quite quickly most of the time.

I don't think we test on any big endian platforms though.

server/src/main/java/org/elasticsearch/common/util/BigIntArray.java
```diff
@@ -40,6 +42,21 @@ final class BigIntArray extends AbstractBigArray implements IntArray {
...
+            throw new IllegalArgumentException();
+        }
+        int intSize = (int) size;
+        out.writeVInt(intSize * 4);
+        for (int i = 0; i < pages.length - 1; i++) {
+            out.write(pages[i]);
+        }
+        int end = intSize % INT_PAGE_SIZE;
+        out.write(pages[pages.length - 1], 0, end * Integer.BYTES);
+        // NOCOMMIT endian-ness
```
original-brownbear, line 57: ++ to your suggestion of adding a fallback for when the system byte order isn't LE if we have to.

But you do make a good point here, we aren't testing or supporting any BE platforms right now as far as I understand. Couldn't we just add a check to node startup that we're not on a BE platform and deal with the issue that way, saving some complexity?

server/src/main/java/org/elasticsearch/common/util/ReleasableIntArray.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public long size() {
+        return ref.length() / 4;
+    }
+
+    @Override
+    public int get(long index) {
+        if (index > Integer.MAX_VALUE / 4) {
+            throw new UnsupportedOperationException(); // NOCOMMIT Oh god, what do we do here?
```
original-brownbear, line 37: Here in this particular line of code it's probably a non-issue since we only read this from the wire and we'll never run into a case where this kind of index isn't out of bounds. Maybe just throw an out of bounds exception here?

server/src/main/java/org/elasticsearch/common/bytes/AbstractBytesReference.java
```diff
@@ -26,6 +26,11 @@ public int getInt(int index) {
         return (get(index) & 0xFF) << 24 | (get(index + 1) & 0xFF) << 16 | (get(index + 2) & 0xFF) << 8 | get(index + 3) & 0xFF;
     }
 
+    @Override
+    public int getIntLE(int index) {
```
original-brownbear, line 30: We should add an implementation (using var handles as we did for other stuff in `Numbers`) of this to `BytesArray`, otherwise this might be quite slow if it's not getting inlined properly (which it probably won't be).

original-brownbear (commented):
Thanks Nik, I added a few comments but this looks really nice. I commented on endianess inline. The size limitation to 2G is a non-issue at the moment since we don't allow transport messages larger than 2G anyway IMO.

server/src/main/java/org/elasticsearch/common/bytes/AbstractBytesReference.java
nik9000: Yeah. We have stuff in `ByteUtils` for this sort of thing I think. I figured I'd do it in a followup and add a microbenchmark. I could do it now if you'd prefer too.

original-brownbear: Follow-up seems fine by me, especially if it gets a benchmark, thanks! :)

server/src/main/java/org/elasticsearch/common/util/BigIntArray.java
nik9000: I've asked around and we don't support any big endian platforms at all. I think it's enough to fail to start if we are on such a system and to comment about it.

nik9000:
I've brought this locally and fixed somethings up. Is it ready to get in so we can build on it?

server/src/main/java/org/elasticsearch/common/util/BigIntArray.java
```diff
@@ -40,6 +41,26 @@ final class BigIntArray extends AbstractBigArray implements IntArray {
...
+        int intSize = (int) size;
+        out.writeVInt(intSize * 4);
+        for (int i = 0; i < pages.length - 1; i++) {
+            out.write(pages[i]);
+        }
+        int end = intSize % INT_PAGE_SIZE;
+        out.write(pages[pages.length - 1], 0, end * Integer.BYTES);
+    }
+
+    static {
```
original-brownbear, line 58: Nit: maybe put this at the beginning of the class file, it's weird to have the static initializer somewhere in the middle of the file isn't it? :)

server/src/main/java/org/elasticsearch/common/util/BigIntArray.java
```diff
@@ -23,7 +25,6 @@
  * configurable length.
  */
 final class BigIntArray extends AbstractBigArray implements IntArray {
-
```
original-brownbear, line 26: Nit: why remove the empty line here?

server/src/main/java/org/elasticsearch/common/util/ReleasableIntArray.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Override
+    public long ramBytesUsed() {
+        /*
+         * We really have no idea how much heap we're using. We have a pretty
+         * good idea how much of the underlying buffer we're using, but that
+         * is already accounted otherwise.
+         */
+        return 0;
```
original-brownbear, line 70: NIT: I guess you could create a constant from `RamUsageEstimator.shallowSizeOfInstance(ReleasableIntArray.class)` and return that here. It doesn't really matter much, but it's the same number of lines as the above comment and slightly more accurate :)

original-brownbear (approved):
LGTM, thanks Nik!

PR #89674, opened 2022-08-26
support arm for shibboleth-idp

jakelandis:
Work in progress. 

diff from fork: https://github.com/jakelandis/shibboleth-idp-dockerized/compare/7f8b92b4319dc7bfb6bb011fe37198170efaa667...master

To test on arm device

1) update the platform (I will clean up how we build the image)
2) ./gradlew :x-pack:qa:saml-idp-tests:javaRestTest --info

Note - I think we have some outstanding clean up issues with these running containers, so it might fail if not starting from a clean slate. Just run again (or manually clean up the running docker instances). 

Also note ... building from a linux base image instead of the upstream image i forked since we will eventually want to update/upgrade and the upstream image appears to no longer be getting updates.


x-pack/test/idp-fixture/build.gradle
```diff
@@ -34,9 +38,16 @@ tasks.named("preProcessFixture").configure {
...
+  dependsOn "preProcessFixture", "buildShibbolethImage"
 }
 
 tasks.named('composePull').configure {
   enabled = false // this task fails due to docker-compose oddities
 }
+
+task buildShibbolethImage(type: Exec) {
+  workingDir "shibboleth-idp"
+  commandLine 'docker',  'buildx', 'build', '-t', 'elastic/shibboleth-idp:test', '.', '--platform', 'linux/amd64' //change to linux/arm64v8 to test
```
jakelandis, line 50: change here to test arm

mark-vieira:
This worked for me after a changed a couple of things:

1. First I had to pass `linux/arm64/v8` as the `--platform`
2. I had to change the condition on `TARGETARCH` to check for `arm64` instead of `arm64v8`.

I think we don't actually have to worry about (1). Passing `--platform` should actually only be necessary when _overriding_ the default, which is to use the host platform. Second, I think we might have to change the `TARGETARCH` stuff anyway, as that is coming from `buildx` and I don't think we want to rely on that. As I mentioned before, we don't have to worry about multi-arch images here, we'll just build for the host platform and by default, Docker already handles this transparently. The only thing we need to handle explicitly is downloading any architecture-specific binaries within the `Dockerfile`, which in our case is just the JDK. That said, we could simplify that even further by just using a base image with a JDK already installed, like one of the official `openjdk` images, which are already multi-arch. That way we don't deal with architecture _at_ all, and when the image is built locally, it automatically targest the host architecture.

jakelandis:
Thanks Mark ! These pointers make sense and will work on adopting these.

jakelandis:
superseded by: https://github.com/elastic/elasticsearch/pull/91216

PR #89675, opened 2022-08-26
Update running snapshot state value to `STARTED`

asmith-elastic:
The get snapshot status API will currently return a value of `STARTED` for the state of a snapshot that is currently running. The documentation says that the `state` value for a running snapshot is `IN_PROGRESS`. This documentation change will align the docs with the actual result of the get snapshot status API.

### DIFF
--- a/docs/reference/snapshot-restore/apis/get-snapshot-status-api.asciidoc
+++ b/docs/reference/snapshot-restore/apis/get-snapshot-status-api.asciidoc
@@ -169,7 +169,7 @@ Indicates the current snapshot state.
 `FAILED`::
   The snapshot finished with an error and failed to store any data.
 
-`IN_PROGRESS`::
+`STARTED`::
   The snapshot is currently running.
 
 `PARTIAL`::


pxsalehi:
@asmith-elastic Thank you for the PR. This seems to be incorrect in the doc, you are right. This might have been a confusion between the get snapshot status and the [get snapshot API](https://www.elastic.co/guide/en/elasticsearch/reference/current/get-snapshot-api.html). The latter returns a list of `SnapshotInfo`s which could have the `IN_PROGRESS` value. I don't see how the snapshot status api could return `IN_PROGRESS`. In fact, [this test](https://github.com/elastic/elasticsearch/blob/8.4/server/src/internalClusterTest/java/org/elasticsearch/snapshots/SnapshotStatusApisIT.java#L100-L127) shows that the get snapshot status API should return `STARTED`. 

I will backport this to 7.x and also change this on main.

Maybe someone from the team could also confirm this.

pxsalehi:

pxsalehi (approved):

pxsalehi:
@asmith-elastic Was there a particular reason you opened the PR agains the 8.4 branch? Usually, the PR should be against main, and if necessary we use the GH labels to backport it.

asmith-elastic:
@pxsalehi - Thanks for helping out with this! There was no particular reason for making these changes on the 8.4 branch, I'm not entirely certain how I managed to do that. I'm sorry in advance for any inconvenience this may have cased!

Issue #89437, opened 2022-08-17
Enable Circuit Breaker tracking in more parts of the aggregations framework


not-napoleon:
### Description

This meta issue tracks the effort to extend circuit breaker memory tracking beyond the collect phase of aggregations.  There are several existing issues related to this, which do a good job of describing the problem (and are linked below), but we need a place to track the tasks for fixing this.  That's what this issue is for.

## Plan
Currently (8.4) aggregations create an object per collected bucket, called `InternalAggregation`.  These objects are stored in the `QuerySearchResult` which is responsible for serializing them from the data nodes back to the coordinator, and also for de-serializing them on coordinator side.  Managing these objects is quite tricky, and does not provide good places to inject the circuit breaker logic.

Instead, we want to move to a dense representation, which would create one object per aggregator.  These objects would be `Releasable`, and responsible for tracking both the post-collection data node side memory usage and the reduce time coordinating node memory usage.  Obviously this involves a (big) change to the wire format used for `QuerySearchResult`.  Doing this in a backwards compatible way is non trivial.

### Tasks
- [ ] Improve testing for post collection memory access
    - In particular, we need to be able to validate that we don't access something that was "released" at the end of the collect phase.
- [ ] Add in support for a new aggregations response format into `QuerySearchResult`.  We can lay the ground work for this without defining too much about the format itself
- [ ] Prototype a dense representation that addresses the memory concerns.  Initially this can be done with the non-recycling big arrays instance, but long term needs to be wired up to a real circuit breaker.
    - [ ] Works for `Max`
    - [ ] Works for `Range`
    - [ ] Works for `Cardinality`
    - [ ] Works for `Terms`
- [ ] Make sure the prototype dense representation also solves the normalization concerns within terms
- [ ] Reduce phase circuit breaker tests
- [ ] Wire up the reduce side circuit breakers
- [ ] Validate prototype (many reviews)
- [ ] BWC, Mixed Mode, CCS testing, testing, testing
- [ ] convert the rest of the aggs

## Related Issues
- https://github.com/elastic/elasticsearch/issues/59892
- https://github.com/elastic/elasticsearch/issues/88373
- https://github.com/elastic/elasticsearch/issues/88128
- https://github.com/elastic/elasticsearch/issues/72619
- https://github.com/elastic/elasticsearch/issues/77449

PR #89676, opened 2022-08-26
Prototype of dense aggregations object

not-napoleon:
Relates to https://github.com/elastic/elasticsearch/issues/89437

Work In Progress

This prototype explores options for how to track memory held past the collect phase of aggregations.


server/src/main/java/org/elasticsearch/search/aggregations/metrics/CollectedMax.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    public String getWriteableName() {
+        return MaxAggregationBuilder.NAME;
+    }
+
+    @Override
+    public Version getMinimalSupportedVersion() {
+        return null;
```
nik9000, line 48: Probably worth a NOCOMMIT. It has to be whatever version we land. I tend to compile for a while with `./gradlew -p server check -x forbiddenPatterns` or something like that. That way `NOCOMMIT` can block merging but doesn't block `precommit`.

server/src/main/java/org/elasticsearch/common/bytes/ReleasableBytesReference.java
```diff
@@ -231,17 +229,4 @@ public int arrayOffset() {
         return delegate.arrayOffset();
     }
 
-    private static final class RefCountedReleasable extends AbstractRefCounted {
```
not-napoleon, line 234: Promoted to a top level public class

server/src/main/java/org/elasticsearch/search/aggregations/AggregationPhase.java
```diff
@@ -101,20 +102,46 @@ public static void execute(SearchContext context) {
...
         if (context.aggregations().factories().context() != null) {
             // Rollup can end up here with a null context but not null factories.....
             context.aggregations().factories().context().multiBucketConsumer().reset();
         }
+        // decide if we're in new style or old style
+        boolean useNewStyle = true;
         for (Aggregator aggregator : context.aggregations().aggregators()) {
+            useNewStyle = useNewStyle && aggregator.canUseCollectedAggregator();
+        }
+        if (useNewStyle) {
```
not-napoleon, line 114: In a perfect world, this never merges to main.  This is the minimum logic to kind-of support mixed mode operation on a single cluster, but to do that in production we'd want a lot more testing, at a minimum.  I'd rather live in a world where all aggregations are either old style or new style, and we cleanly cut over at a version boundary.

server/src/main/java/org/elasticsearch/search/aggregations/AggregatorBase.java
```diff
@@ -325,6 +325,11 @@ protected final BigArrays bigArrays() {
         return context.bigArrays();
     }
 
+
+    protected final BigArrays bigArraysForResults() {
```
not-napoleon, line 329: I'm not sure we actually need this.  Maybe we just have one BigArrays, and it outlives the context.  Otherwise, what's the use case for data that lives at collect time but doesn't extend to serialization time?

server/src/main/java/org/elasticsearch/search/aggregations/CollectedAggregator.java
```diff
@@ -0,0 +1,153 @@
...
+     * User supplied metadata.  From a framework perspective, this is an opaque pass through that
+     * we just need to return in our response.
+     */
+    protected final Map<String, Object> metadata;
+
+    public BigArrays bigArrays() {
+        return bigArrays;
+    }
+
+    private final BigArrays bigArrays;
```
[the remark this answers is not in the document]
not-napoleon: yeah, the more I think about it, the less I want this class to have a `BigArrays` reference.  I'm going to see if I can get it out.

benchmarks/src/main/java/org/elasticsearch/benchmark/search/aggregations/AggConstructionContentionBenchmark.java
```diff
@@ -252,6 +252,11 @@ public BigArrays bigArrays() {
             return bigArrays;
         }
 
+        @Override
+        public BigArrays bigArraysForResults() {
```
not-napoleon, line 256: I'm not sure this is necessary.  Or, more specifically, what is the use case for big arrays that get cleaned up at the end of the collect phase?  Are there data large enough that we need big arrays for them, but we don't want to pass them onto the reduce?

server/src/main/java/org/elasticsearch/search/aggregations/metrics/MaxAggregator.java
```diff
@@ -127,6 +128,33 @@ public InternalAggregation buildEmptyAggregation() {
...
     }
 
+    @Override
+    public boolean canUseCollectedAggregator() {
+        return true;
+    }
+
+    @Override
+    public CollectedAggregator buildCollectedAggregator(long[] ordsToCollect) {
+        CollectedMax collectedMax = new CollectedMax(name, metadata(), bigArraysForResults(), ordsToCollect.length, formatter);
```
not-napoleon, line 138: It's almost definitely wrong to be creating a new `CollectedAggregator` here. We want there to be exactly one of these per aggregator.

server/src/main/java/org/elasticsearch/common/bytes/RefCountedReleasable.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.elasticsearch.core.AbstractRefCounted;
+import org.elasticsearch.core.Releasable;
+import org.elasticsearch.core.Releasables;
+
+/**
+ * Wrap a {@link Releasable} in a {@link org.elasticsearch.core.RefCounted}.  All the reference counting logic is handled in
+ * {@link AbstractRefCounted}, and close is implemented by closing the wrapped {@linkplain Releasable}
+ */
+public final class RefCountedReleasable extends AbstractRefCounted {
```
not-napoleon, line 19: This was just promoted from an internal class to a top level class.

server/src/main/java/org/elasticsearch/search/aggregations/CollectedAggregator.java
```diff
@@ -0,0 +1,160 @@
...
+    /**
+     * The user supplied name of this aggregation.
+     */
+    protected final String name;
+
+    /**
+     * User supplied metadata.  From a framework perspective, this is an opaque pass through that
+     * we just need to return in our response.
+     */
+    protected final Map<String, Object> metadata;
```
not-napoleon, line 46: Can we just not round trip this?

PR #89677, opened 2022-08-27
Paranoid tests for `meta`

nik9000:
This adds some extra paranoid tests for the `meta` parameter in aggs,
specifically the `filters` agg. These tests are at the REST level so
they provide backwards compatibility tests as well. They make sure that
`meta: {}` and `meta: null` do what we expect - return a `meta: {}` and
return an error.

Relates to #89467


not-napoleon (approved):
LGTM

PR #89680, opened 2022-08-29
Add info of effective roles in denial messages

ywangd:
When an action is denied due to authorization error, the list of assigned roles is shown in the error message. However, it is possible that the effective roles are fewer or more than the assigned list:
* Fewer roles can happen when the role is not defined or the license does not permit it
* More roles can happen when anonymous access is enabled

This PR changes the error message to show the effective roles instead of the assigned roles (whenever possible) to help troubleshooting. In addition, it also reports any missing roles, i.e. roles that are assigned but cannot be found.


x-pack/plugin/security/src/main/java/org/elasticsearch/xpack/security/authz/AuthorizationDenialMessages.java
```diff
@@ -92,6 +103,60 @@ private static String authenticatedUserDescription(Authentication authentication
...
+            sb.append(" with no declared roles");
+        }
+        if (authorizationInfo == null) {
+            return sb.toString();
+        }
+
+        final Set<String> resolvedRoleNames;
+        final Role role = RBACEngine.maybeGetRBACEngineRole(authorizationInfo);
+        if (role == Role.EMPTY) {
+            resolvedRoleNames = Set.of();
```
ywangd, line 126: `Role.EMPTY` does not have a empty list of role names but `["__empty"]`. These code is to take care of this quirk, which I raised a separate discussion to address.

tvernum:
Drive-by-comment:

I think it's unlikely that a typical user would understand what "all resolved" (etc) means. That's not terminology that end users would use.

ywangd:
> I think it's unlikely that a typical user would understand what "all resolved" (etc) means. That's not terminology that end users would use.

The word `resolve` might be more developer oriented. I struggle to find a better verb. But maybe we can use an adjective, e.g. `effective`? So it would be something like:

* "all effective" when every declared role is successfully resolved
* "none effective" when none of the declared role can be resolved
* "ineffective [...]" (or "not effective") lists the unresolved roles
* "additionally effective [...]" lists the extra roles (anonymous roles for now) resolved that are not part of declared roles

I don't particularly like "additionally effective" which sounds a bit awkward than "additionally resolved". I am open to suggestions. Thanks!

slobodanadamovic (approved):
Generally, it look good to me. I like Tim's proposal to point out only roles which were not resolved.

ywangd:
I was trying to be elaborative in the error message to convery more information about what are assigned roles and what are anonymous roles and any of the mismatches. It makes the error message construction as well as the wording choices more complex. I do recognise that the key value of this change is to report effective roles and any unfounded one. Therefore,
I updated the PR following @tvernum's [suggestion](https://github.com/elastic/elasticsearch/pull/89680#issuecomment-1231389453). Please take a look if still interested. Thanks all for the review.

ywangd:
@elasticmachine update branch

PR #89682, opened 2022-08-29
Simplify thread name assertion

idegtiarenko:
Simplify thread name assertion by using variant that permits a test thread


DaveCTurner (approved):
LGTM

PR #89683, opened 2022-08-29
Remove sync allocate implementation

idegtiarenko:


server/src/main/java/org/elasticsearch/cluster/routing/allocation/allocator/ShardsAllocator.java
```diff
@@ -33,22 +33,24 @@ public interface ShardsAllocator {
...
+     * may use the results of this method implementation to decide on allocating shards in the routing table
+     * to the cluster.
+     *
+     * If an implementation of this interface does not support explaining decisions for a single shard through
+     * the cluster explain API, then this method should throw a {@code UnsupportedOperationException}.
      */
-    default void allocate(RoutingAllocation allocation, ActionListener<Void> listener) {
-        allocate(allocation);
-        listener.onResponse(null);
-    }
```
idegtiarenko, line 51: This is the only thing removed. Diff is caused by moving default `execute` method to the bottomo.

DaveCTurner:
I think this is still worth doing, although it'll need fixing up and retargeting at `main`.

PR #89706, opened 2022-08-29
Synthetic _source: support version field type

nik9000:
This adds support for synthetic _source to the `version` field type. It
works very similarly to `keyword` but with an extra decode step.

I modified the decoder to return a `BytesRef` instead of a `String`
because many of the callers seemed to be converting that string directly
into bytes again. Synthetic source would have wanted to do that. As was
the query infrastructure.


romseygeek (approved):
Nice, LGTM.

PR #89708, opened 2022-08-29
Operator/autoscaling

grcevski:
This PR adds support for /_autoscaling/policy for file based settings. The PR uses similar approach in how we handled ILM policies, where we load the state handlers through SPI.


server/src/main/java/org/elasticsearch/cluster/service/ClusterService.java
```diff
@@ -56,6 +57,8 @@ public class ClusterService extends AbstractLifecycleComponent {
 
     private RerouteService rerouteService;
 
+    private AllocationService allocationService;
```
grcevski, line 60: I'm looking for suggestions if this can be made better. Right now the `allocationDeciders` are injected in the TransportActions that autoscaling uses, but I can't use injection since we made a design decision to use SPI for the reserved state handlers.

Essentially, I need a way to get to the allocationDeciders to be able to create/pass the validator in to the transport action transformation methods.

x-pack/plugin/autoscaling/src/internalClusterTest/java/org/elasticsearch/xpack/autoscaling/LocalStateReservedAutoscalingStateHandlerProvider.java
```diff
@@ -0,0 +1,47 @@
...
+import org.elasticsearch.reservedstate.ReservedClusterStateHandler;
+import org.elasticsearch.reservedstate.ReservedClusterStateHandlerProvider;
+
+import java.util.Collection;
+import java.util.Objects;
+
+/**
+ * Autoscaling provider implementation for the {@link ReservedClusterStateHandlerProvider} service interface
+ */
+public class LocalStateReservedAutoscalingStateHandlerProvider implements ReservedClusterStateHandlerProvider {
```
grcevski, line 19: This is a clone of the regular plugin SPI state handler provider for integration test purposes.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/action/ReservedAutoscalingPolicyAction.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    public TransformState transform(Object source, TransformState prevState) throws Exception {
+        var requests = prepare(source);
+        ClusterState state = prevState.state();
+
+        for (var request : requests) {
+            state = TransportPutAutoscalingPolicyAction.putAutoscalingPolicy(
+                state,
+                request,
+                policyValidatorHolder.get(clusterService.getAllocationService().getAllocationDeciders())
```
grcevski, line 79: Perhaps there's a better way to get to the allocationDeciders without injection?

server/src/main/java/org/elasticsearch/cluster/service/ClusterService.java
```diff
@@ -100,6 +103,16 @@ public RerouteService getRerouteService() {
         return rerouteService;
     }
 
+    public void setAllocationService(AllocationService allocationService) {
```
henningandersen, line 106: I think I'd prefer to either:

1. Support injection for the reserved state plugins.
2. Add an explicit AllocationDecidersProvider object that is passed to Plugin.createComponents that can be used to access this.

Putting it here is polluting this class I think with responsibilities that does not concern it.

henningandersen (commented):
I mostly focused on how to provide the allocation deciders and left a comment on that. I'd prefer to defer the rest of the review until that is tackled.

grcevski:
@elasticsearchmachine run elasticsearch-ci/bwc

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/Autoscaling.java
```diff
@@ -115,14 +118,13 @@ public Collection<Object> createComponents(
...
+        Supplier<AllocationDeciders> allocationDecidersSupplier
     ) {
         this.clusterServiceHolder.set(clusterService);
-        return List.of(
-            new AutoscalingCalculateCapacityService.Holder(this),
-            autoscalingLicenseChecker,
-            new AutoscalingNodeInfoService(clusterService, client)
-        );
+        var capacityServiceHolder = new AutoscalingCalculateCapacityService.Holder(this);
+        reservedAutoscalingPolicyAction.set(new ReservedAutoscalingPolicyAction(capacityServiceHolder, allocationDecidersSupplier));
```
henningandersen, line 126: Since we now have the list of deciders directly, could we not as well just set it directly here rather than pass it around in several places? I.e., add:

```
this.allocationDeciders.set(deciders);
```

and remove it from the argument to `ReservedAutoscalingPolicyAction` as well as from being injected into the two transport services that gets it (only to pass it to the plugin).

henningandersen (commented):
I got what I asked for :slightly_smiling_face: , unfortunately it resulted in quite many changed files. Can we make the `allocationDeciders` change in a separate PR first, such that this PR gets easier to review? I added a couple of comments to the `allocationDeciders` change that should be carried over to the new PR as well.

grcevski:
> I got what I asked for 🙂 , unfortunately it resulted in quite many changed files. Can we make the `allocationDeciders` change in a separate PR first, such that this PR gets easier to review? I added a couple of comments to the `allocationDeciders` change that should be carried over to the new PR as well.

Sounds good. I'll split this off in another PR and mark this in draft until that's merged. Thanks again for the review!

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -82,13 +82,11 @@ public FileSettingsService(ClusterService clusterService, ReservedClusterStateSe
         this.operatorSettingsDir = environment.configFile().toAbsolutePath().resolve(OPERATOR_DIRECTORY);
     }
 
-    // package private for testing
-    Path operatorSettingsDir() {
+    public Path operatorSettingsDir() {
```
grcevski, line 85: Exposed for easier test writing.

x-pack/plugin/autoscaling/src/internalClusterTest/java/org/elasticsearch/xpack/autoscaling/LocalStateReservedAutoscalingStateHandlerProvider.java
grcevski, line 19: Mock version of the SPI handler provider so that we can write Java integration tests. It implements equals and hashcode so we deduplicate the plugin, as it can be discovered multiple times in the MockNode because all plugins are loaded by the same classloader.

grcevski:

x-pack/plugin/autoscaling/src/internalClusterTest/java/org/elasticsearch/xpack/autoscaling/LocalStateReservedAutoscalingStateHandlerProvider.java
```diff
@@ -0,0 +1,47 @@
...
+package org.elasticsearch.xpack.autoscaling;
+
+import org.elasticsearch.reservedstate.ReservedClusterStateHandler;
+import org.elasticsearch.reservedstate.ReservedClusterStateHandlerProvider;
+
+import java.util.Collection;
+import java.util.Objects;
+
+/**
+ * Autoscaling provider implementation for the {@link ReservedClusterStateHandlerProvider} service interface
```
henningandersen, line 17: Can you add a comment explaining why we need this test override? It is not clear to me and I'd rather read it than find out.

x-pack/plugin/autoscaling/src/internalClusterTest/java/org/elasticsearch/xpack/autoscaling/LocalStateReservedAutoscalingStateHandlerProvider.java
henningandersen: Should we not fix that in `MockNode` instead? I am not sure I follow the equals/hashCode problem, but sounds like we are relying on `MockNode` using a `Set` to collect these providers?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/Autoscaling.java
```diff
@@ -121,6 +124,7 @@ public Collection<Object> createComponents(
         this.clusterServiceHolder.set(clusterService);
         this.allocationDeciders.set(allocationDeciders);
         var capacityServiceHolder = new AutoscalingCalculateCapacityService.Holder(this);
+        this.reservedAutoscalingPolicyAction.set(new ReservedAutoscalingPolicyAction(capacityServiceHolder));
```
henningandersen, line 127: Can we not simply defer the construction of this until `reservedClusterStateHandlers`? That seems more intuitive to me.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/action/ReservedAutoscalingPolicyAction.java
```diff
@@ -0,0 +1,102 @@
...
+        for (var policy : policies) {
+            validate(policy);
+        }
+
+        return policies;
+    }
+
+    @Override
+    public TransformState transform(Object source, TransformState prevState) throws Exception {
+        var requests = prepare(source);
```
henningandersen, line 66: Can we do the casting here rather than in `prepare`? That avoids `SuppressWarnings` and also seems more appropriate to deal with input validation immediately in this method.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/action/ReservedAutoscalingPolicyAction.java
```diff
@@ -0,0 +1,102 @@
...
+        this.policyValidatorHolder = policyValidatorHolder;
+    }
+
+    @Override
+    public String name() {
+        return NAME;
+    }
+
+    @SuppressWarnings("unchecked")
+    public Collection<PutAutoscalingPolicyAction.Request> prepare(Object input) {
```
henningandersen, line 54: I think this can be private?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/action/ReservedAutoscalingPolicyAction.java
```diff
@@ -0,0 +1,102 @@
...
+    public ReservedAutoscalingPolicyAction(final AutoscalingCalculateCapacityService.Holder policyValidatorHolder) {
+        this.policyValidatorHolder = policyValidatorHolder;
+    }
+
+    @Override
+    public String name() {
+        return NAME;
+    }
+
+    @SuppressWarnings("unchecked")
```
henningandersen, line 53: Can this be moved to the statement?

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/action/ReservedAutoscalingPolicyAction.java
```diff
@@ -0,0 +1,102 @@
...
+
+        for (var policy : policies) {
+            validate(policy);
+        }
+
+        return policies;
+    }
+
+    @Override
+    public TransformState transform(Object source, TransformState prevState) throws Exception {
```
henningandersen, line 65: Could this not get the right type instead? Seems it is T at the interface level and the call in `ReservedStateUpdateTask` can be massaged to fix this. The service does "guarantee" it, since it only passes output of `fromXContent` to this method.

x-pack/plugin/autoscaling/src/test/java/org/elasticsearch/xpack/autoscaling/action/ReservedAutoscalingPolicyTests.java
```diff
@@ -0,0 +1,129 @@
...
+                 "deciders": {
+                   "fixed": {
+                   }
+                 }
+               }
+            }""";
+
+        prevState = updatedState;
+        updatedState = processJSON(action, prevState, json);
+        assertThat(updatedState.keys(), containsInAnyOrder("my_autoscaling_policy", "my_autoscaling_policy_1"));
```
henningandersen, line 103: Can we also validate that the roles and deciders are correct in cluster state?

henningandersen (approved):
LGTM.

Left a number of comments I'd prefer to see included.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/Autoscaling.java
grcevski: I avoided doing that because then I'd have to also keep a local reference to the `AutoscalingCalculateCapacityService.Holder`, I need it to construct the action object.

x-pack/plugin/autoscaling/src/main/java/org/elasticsearch/xpack/autoscaling/action/ReservedAutoscalingPolicyAction.java
grcevski: On the caller side in `ReservedStateUpdateTask` I have a composite object containing the data for all of the handlers. I parse everything first in a map of String key to data type required by the handler, but I can't retain the type information because the data types are in classes provided by plugins and the server doesn't know of them.

PR #89714, opened 2022-08-30
[DOCS] Add simple dev setup instructions back to the README

### LINKED ISSUE #89739: [DOCS] Restore simple dev setup instructions to the Elasticsearch README
### Description

As part of the Onboarding 2.0 effort, we've introduced a set of new getting started tutorials and worked to minimize the number of similar-but-different "getting started" entry points in the docs so the recommended path is the most discoverable path. All well and good, but the cascade of changes over time have kind of left behind the developer-tinkerer audience. 

To show them some ❤️, we want to restore basic dev set-up instructions to the Elasticsearch README, and add a corresponding "Set up a local dev deployment" topic to the Elasticsearch guide.  Note that we want to avoid calling this a "Quick start" or "Getting started" to avoid diverting traffic from the getting started tutorials. 

Once the new topic is in place, we can link to it from the download page.

debadair:
Closes: [89739](https://github.com/elastic/elasticsearch/issues/89739)

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -37,6 +37,11 @@ To try out Elasticsearch on your own machine, we recommend using Docker
 and running both Elasticsearch and Kibana.
 Docker images are available from the https://www.docker.elastic.co[Elastic Docker registry].
 
+NOTE: Starting in Elasticsearch 8.0, security is enabled by default. 
+The first time you start Elasticsearch, TLS encryption is configured automatically, 
+a password is generated for the `elastic` user, 
+and a Kibana enrollment token is created so you can connect Kibana to your secured cluster.
+
 For other installation options, see the
 https://www.elastic.co/guide/en/elasticsearch/reference/current/install-elasticsearch.html[Elasticsearch installation documentation].
 
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -93,7 +93,7 @@ You can interact with Elasticsearch using any client that sends HTTP requests,
 such as the https://www.elastic.co/guide/en/elasticsearch/client/index.html[Elasticsearch
 language clients] and https://curl.se[curl]. 
 Kibana's developer console provides an easy way to experiment and test requests. 
-To access the console, go to **Managment > Dev Tools**.
+To access the console, go to **Management > Dev Tools**.
 
 **Add data**
 
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -30,6 +30,7 @@ https://www.elastic.co/downloads/elasticsearch[elastic.co/downloads/elasticsearc
 
 To try out Elasticsearch on your dev machine, we recommend using Docker
 and running both Elasticsearch and Kibana.
+Docker images are available from the https://www.docker.elastic.co[Elastic Docker registry].
 
 For other installation options, see the
 https://www.elastic.co/guide/en/elasticsearch/reference/current/install-elasticsearch.html[Elasticsearch installation documentation].
@@ -37,29 +38,53 @@ https://www.elastic.co/guide/en/elasticsearch/reference/current/install-elastics
 **Install and run Elasticsearch**
 
 . Install and start https://www.docker.com/products/docker-desktop[Docker
-Desktop].
+Desktop]. Go to **Preferences > Resources > Advanced** and set Memory to at least 4GB.
 
-. Run:
+. Start an Elasticsearch container:
 +
 ----
 docker network create elastic
-docker pull {docker-repo}:{version}
-docker run --name es01-test --net elastic -p 127.0.0.1:9200:9200 -p 127.0.0.1:9300:9300 -e "discovery.type=single-node" {docker-image}
+docker pull docker.elastic.co/elasticsearch/elasticsearch:{version}
+docker run --name es-node1 --net elastic -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" -t docker.elastic.co/elasticsearch/elasticsearch:{version}
 ----
++
+When you start Elasticsearch for the first time, the following security configuration 
+occurs automatically:
++
+* https://www.elastic.co/guide/en/elasticsearch/reference/current/configuring-stack-security.html#stack-security-certificates[Certificates and keys] 
+are generated for the transport and HTTP layers.
+* The Transport Layer Security (TLS) configuration settings are written to
+`elasticsearch.yml`.
+* A password is generated for the `elastic` user.
+* An enrollment token is generated for Kibana.
++
+NOTE: You might need to scroll back a bit in the terminal to view the password 
+and enrollment token.
+
+. Copy the generated password and enrollment token and save them in a secure 
+location. These values are shown only when you start Elasticsearch for the first time.
+You'll use these to enroll Kibana with your Elasticsearch cluster and log in.
 
 **Install and run Kibana**
 
-To easily send requests to Elasticsearch and analyze, visualize, and manage data 
-interactively, install Kibana.
+Kibana enables you to easily send requests to Elasticsearch and analyze, visualize, and manage data interactively.
 
-. In a new terminal session, run:
+. In a new terminal session, start Kibana and connect it to your Elasticsearch container:
 +
 ----
 docker pull docker.elastic.co/kibana/kibana:{version}
-docker run --name kib01-test --net elastic -p 127.0.0.1:5601:5601 -e "ELASTICSEARCH_HOSTS=http://es01-test:9200" docker.elastic.co/kibana/kibana:{version}
+docker run --name kib-01 --net elastic -p 5601:5601 docker.elastic.co/kibana/kibana:{version}
 ----
++
+When you start Kibana, a unique link is output to your terminal.
 
-. To access {kib}, go to http://localhost:5601[http://localhost:5601]
+. To access Kibana, open the generated link in your browser.
+
+  .. Paste the enrollment token that you copied when starting
+  Elasticsearch and click the button to connect your Kibana instance with Elasticsearch.
+
+  .. Log in to Kibana as the `elastic` user with the password that was generated
+  when you started Elasticsearch.
 
 **Send requests to Elasticsarch**
 
@@ -67,7 +92,8 @@ You send data and other requests to Elasticsearch through REST APIs.
 You can interact with Elasticsearch using any client that sends HTTP requests, 
 such as the https://www.elastic.co/guide/en/elasticsearch/client/index.html[Elasticsearch
 language clients] and https://curl.se[curl]. 
-Kibana's developer tools console provides an easy way to experiment and test requests. 
+Kibana's developer console provides an easy way to experiment and test requests. 
+To access the console, go to **Managment > Dev Tools**.
 
 **Add data**
 
@@ -75,7 +101,7 @@ You index data into Elasticsearch by sending JSON objects (documents) through th
 Whether you have structured or unstructured text, numerical data, or geospatial data, 
 Elasticsearch efficiently stores and indexes it in a way that supports fast searches. 
 
-For time series data, such as logs and metrics, you typically add documents to a
+For timestamped data such as logs and metrics, you typically add documents to a
 data stream made up of multiple auto-generated backing indices.
 
 To add a single document to an index, submit an HTTP post request that targets the index. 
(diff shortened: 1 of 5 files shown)


lockewritesdocs (commented):
We need to amend these instructions to account for security-on-by-default. When starting Elasticsearch for the first time, an enrollment token is generated for Kibana. Users need to copy that token and use it to enroll Kibana in their browser. I think that the simplest instructions to include in the README for starting ES + Kibana are the ones from the [Kibana installation docs](https://www.elastic.co/guide/en/kibana/current/docker.html) for Docker. I'm happy to update this PR to reflect those changes; just let me know!

debadair:
I incorporated the changes from the Kibana doc and tested the examples. Once the README content is reviewed & approved, I'll add replicate it for the corresponding topic in the ES guide.

lockewritesdocs:
> Once the README content is reviewed & approved, I'll add replicate it for the corresponding topic in the ES guide.

Just an FYI that we have a page for [installing ES on Docker](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html). The multi-node configuration includes Kibana.

README.asciidoc
```diff
@@ -26,9 +26,150 @@ If you prefer to install and manage Elasticsearch yourself, you can download
...
+
+. To access Kibana, open the generated link in your browser.
+
+  .. Paste the enrollment token that you copied when starting
+  Elasticsearch and click the button to connect your Kibana instance with Elasticsearch.
+
+  .. Log in to Kibana as the `elastic` user with the password that was generated
+  when you started Elasticsearch.
+
+**Send requests to Elasticsarch**
```
lockewritesdocs, line 89:
```suggestion
**Send requests to Elasticsearch**
```

README.asciidoc
```diff
@@ -26,9 +26,150 @@ If you prefer to install and manage Elasticsearch yourself, you can download
...
+**Install and run Elasticsearch**
+
+. Install and start https://www.docker.com/products/docker-desktop[Docker
+Desktop]. Go to **Preferences > Resources > Advanced** and set Memory to at least 4GB.
+
+. Start an Elasticsearch container:
++
+----
+docker network create elastic
+docker pull docker.elastic.co/elasticsearch/elasticsearch:{version}
```
lockewritesdocs, line 47: I don't know that the `{version}` variable will render the current Elasticsearch version in the README. Instead, should we add a note informing users to replace `{version}` with the version that they want to download, such as `8.4.1`?

Same comment in the next command.

README.asciidoc
```diff
@@ -26,9 +26,150 @@ If you prefer to install and manage Elasticsearch yourself, you can download
...
+You'll use these to enroll Kibana with your Elasticsearch cluster and log in.
+
+**Install and run Kibana**
+
+Kibana enables you to easily send requests to Elasticsearch and analyze, visualize, and manage data interactively.
+
+. In a new terminal session, start Kibana and connect it to your Elasticsearch container:
++
+----
+docker pull docker.elastic.co/kibana/kibana:{version}
```
lockewritesdocs, line 75: Same point here about the `{version}` variable.

README.asciidoc
```diff
@@ -26,9 +26,150 @@ If you prefer to install and manage Elasticsearch yourself, you can download
...
+  when you started Elasticsearch.
+
+**Send requests to Elasticsarch**
+
+You send data and other requests to Elasticsearch through REST APIs. 
+You can interact with Elasticsearch using any client that sends HTTP requests, 
+such as the https://www.elastic.co/guide/en/elasticsearch/client/index.html[Elasticsearch
+language clients] and https://curl.se[curl]. 
+Kibana's developer console provides an easy way to experiment and test requests. 
+To access the console, go to **Managment > Dev Tools**.
```
lockewritesdocs, line 96:
```suggestion
To access the console, go to **Management > Dev Tools**.
```

lockewritesdocs (commented):
Suggested a few typo changes and had a question about how to handle the `{version}` variable.

README.asciidoc
debadair: It won't render. I do not know of a way to do substitution that will work when viewed on GH.  A callout is probably the best we can do.

lockewritesdocs: I think that a callout works well here. Something simple like:

>Replace `{version}` with the version of Elasticsearch that you want to install.

Totally at your discretion.

lockewritesdocs (approved):
Left two hyper-nit comments that you can choose to incorporate or ignore. Otherwise, :shipit:

README.asciidoc
```diff
@@ -26,9 +26,157 @@ If you prefer to install and manage Elasticsearch yourself, you can download
 the latest version from 
 https://www.elastic.co/downloads/elasticsearch[elastic.co/downloads/elasticsearch].
 
-For more installation options, see the
-https://www.elastic.co/guide/en/elasticsearch/reference/current/install-elasticsearch.html[Elasticsearch installation
-documentation].
+=== Set up a local dev deployment
```
mark-vieira, line 29: I'd probably title this "Running Elasticsearch locally"

README.asciidoc
```diff
@@ -26,9 +26,157 @@ If you prefer to install and manage Elasticsearch yourself, you can download
...
+
+//// 
+IMPORTANT: This content is replicated in the Elasticsearch guide. 
+If you make changes, you must also update setup/set-up-local-dev-deployment.asciidoc.
+////
+
+To try out Elasticsearch on your dev machine, we recommend using Docker
```
mark-vieira, line 36: Again, I'd use the phrase "local" vs "dev" here as there are implications around the term "development".

README.asciidoc
```diff
@@ -26,9 +26,157 @@ If you prefer to install and manage Elasticsearch yourself, you can download
...
+. Start an Elasticsearch container:
++
+----
+docker network create elastic
+docker pull docker.elastic.co/elasticsearch/elasticsearch:{version} <1>
+docker run --name es-node1 --net elastic -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" -t docker.elastic.co/elasticsearch/elasticsearch:{version}
+----
+<1> Replace {version} with the version of Elasticsearch you want to run, such as 8.4.1.
++
+When you start Elasticsearch for the first time, the following security configuration
```
mark-vieira, line 57: In a local setup do we want to deal with security by default? @tvernum do you have a strong opinion here? I hate use suggesting to people to circumvent the default but this just adds unneeded complexity. Also, as I mentioned in a previous discussion, the fact we default to JSON formatted output in Docker makes sorting through the logs a pain. To me the advantage of using docker for this scenario (ES + Kibana) is to be able to provide a simple Docker compose script.

README.asciidoc
```diff
@@ -26,9 +26,157 @@ If you prefer to install and manage Elasticsearch yourself, you can download
...
+
+. Install and start https://www.docker.com/products/docker-desktop[Docker
+Desktop]. Go to **Preferences > Resources > Advanced** and set Memory to at least 4GB.
+
+. Start an Elasticsearch container:
++
+----
+docker network create elastic
+docker pull docker.elastic.co/elasticsearch/elasticsearch:{version} <1>
+docker run --name es-node1 --net elastic -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" -t docker.elastic.co/elasticsearch/elasticsearch:{version}
```
mark-vieira, line 53: I'd just name the container `elasticsearch` since there's only one.

README.asciidoc
```diff
@@ -26,9 +26,157 @@ If you prefer to install and manage Elasticsearch yourself, you can download
...
+
+**Install and run Kibana**
+
+Kibana enables you to easily send requests to Elasticsearch and analyze, visualize, and manage data interactively.
+
+. In a new terminal session, start Kibana and connect it to your Elasticsearch container:
++
+----
+docker pull docker.elastic.co/kibana/kibana:{version} <1>
+docker run --name kib-01 --net elastic -p 5601:5601 docker.elastic.co/kibana/kibana:{version}
```
mark-vieira, line 82: Similarly to above, let's just call the container `kibana` since we only have one.

README.asciidoc
```diff
@@ -26,9 +26,157 @@ If you prefer to install and manage Elasticsearch yourself, you can download
...
+Desktop]. Go to **Preferences > Resources > Advanced** and set Memory to at least 4GB.
+
+. Start an Elasticsearch container:
++
+----
+docker network create elastic
+docker pull docker.elastic.co/elasticsearch/elasticsearch:{version} <1>
+docker run --name es-node1 --net elastic -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" -t docker.elastic.co/elasticsearch/elasticsearch:{version}
+----
+<1> Replace {version} with the version of Elasticsearch you want to run, such as 8.4.1.
```
mark-vieira, line 55: This is an example of where having `latest` tags for this stuff would be really helpful. We really want to avoid having hard-coded version numbers in the README. Unlike our docs, we can't use variable substitution to ensure this is always the latest release. We'd have to constantly keep this up to date.

mark-vieira (commented):
I'm 50/50 on merging this as-is. Frankly, I think the issue of bootstrapping a local environment is less one of documentation, and more one of _experience_, and there are a number of things we probalby should do to improve that experience with Docker (addressing log output, defaults, latest version tags, etc).

README.asciidoc
debadair: I don't think we should encourage people to dodge security by default. That said, it's probably sufficient here to just say that the password and enrollment token are generated automatically, rather than itemizing all of the config that happens.

Re Docker compose, it doesn't really simplify much in this one node case, and this aligns with the Kibana docs.

README.asciidoc
debadair: Agree it would be really helpful! For now, I deliberately did not hard code the version in the snippet and specified a version as an example in the callout. It's not ideal, but I it makes it less critical to bump the version every release.

debadair:
> I'm 50/50 on merging this as-is. Frankly, I think the issue of bootstrapping a local environment is less one of documentation, and more one of experience, and there are a number of things we probalby should do to improve that experience with Docker (addressing log output, defaults, latest version tags, etc).

Agree that there are things we can do to improve the experience, but I think adding this content addresses *some* of the concerns and doesn't need to wait for those improvements. We can adapt it as things change.

README.asciidoc
mark-vieira: I think I'd avoid having an example version here at all as it'll quickly be out of date and we don't want to encourage folks to use old versions.

mark-vieira (approved):
Thanks, Deb. This looks good from my end.

debadair:
Thanks for the review @mark-vieira!

docs/reference/setup/run-elasticsearch-locally.asciidoc
```diff
@@ -0,0 +1,161 @@
+[[run-Elasticsearch-locally]]
```
debadair, line 1:
```suggestion
[[run-elasticsearch-locally]]
```

README.asciidoc
debadair: @elasticmachine run elasticsearch-ci/docs

debadair:

docs/reference/setup/run-elasticsearch-locally.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+and running both Elasticsearch and Kibana.
+Docker images are available from the https://www.docker.elastic.co[Elastic Docker registry].
+
+NOTE: Starting in Elasticsearch 8.0, security is enabled by default. 
+The first time you start Elasticsearch, TLS encryption is configured automatically, 
+a password is generated for the `elastic` user, 
+and a Kibana enrollment token is created so you can connect Kibana to your secured cluster.
+
+For other installation options, see the
+https://www.elastic.co/guide/en/Elasticsearch/reference/current/install-elasticsearch.html[Elasticsearch installation documentation].
```
debadair, line 26:
```suggestion
https://www.elastic.co/guide/en/elasticsearch/reference/current/install-elasticsearch.html[Elasticsearch installation documentation].
```

README.asciidoc
```diff
@@ -26,9 +26,155 @@ If you prefer to install and manage Elasticsearch yourself, you can download
...
 https://www.elastic.co/downloads/elasticsearch[elastic.co/downloads/elasticsearch].
 
-For more installation options, see the
-https://www.elastic.co/guide/en/elasticsearch/reference/current/install-elasticsearch.html[Elasticsearch installation
-documentation].
+=== Run Elasticsearch locally
+
+//// 
+IMPORTANT: This content is replicated in the Elasticsearch guide. 
+If you make changes, you must also update setup/set-up-local-dev-deployment.asciidoc.
```
debadair, line 33:
```suggestion
If you make changes, you must also update setup/run-elasticsearch-locally.asciidoc.
```

PR #89717, opened 2022-08-30
Add bounded wait for the desired balance computation to avoid followup state update

idegtiarenko:


DaveCTurner:
We discussed this today and decided we were a bit uncomfortable with having the behaviour depend so much on the something nondeterministic like computation time. We might need this eventually but would prefer to hold off until we find a good reason to proceed.

PR #88391, opened 2022-07-08
Add troubleshooting guide for corrupt repository

andreidan:
This adds a troubleshooting guide for the situation when multiple clusters
are writing to the same snapshot repository, resulting in a corrupt repository.

Preview link https://elasticsearch_88391.docs-preview.app.elstc.co/guide/en/elasticsearch/reference/master/add-repository.html

PR #89719, opened 2022-08-30
Re-registering corrupt repository unblocks it

### LINKED ISSUE #89130: Permit no-op reregistration of a snapshot repository if marked corrupted
Today if we detect concurrent modification of a snapshot repository then we mark it with `CORRUPTED_REPO_GEN` which blocks further writes to try and prevent additional damage to its contents. You can remove the marker by calling the register-repository API again, but this API will discard no-op re-registrations without taking any action. You can work around this by unregistering the repository first (which may be impossible if it's still in use for searchable snapshots) or by re-registering it with some trivial setting change (which seems hacky).

We should not consider it a no-op re-registration if the repository is currently marked with `CORRUPTED_REPO_GEN`, because doing this re-registration will re-read the repo contents and clear the corruption marker and therefore is not a no-op.

kingherc:
Fixes #89130


server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
```diff
@@ -215,8 +215,18 @@ public ClusterState execute(ClusterState currentState) {
...
-                                // Previous version is the same as this one no update is needed.
-                                return currentState;
+                                if (repositoryMetadata.generation() == RepositoryData.CORRUPTED_REPO_GEN) {
+                                    // If re-creating a repository with the same settings, reset the corrupted flag (issue #89130).
+                                    // Setting the safe generation to unknown, so that a consistent generation is found.
+                                    repositoryMetadata = new RepositoryMetadata(
+                                        repositoryMetadata,
+                                        RepositoryData.UNKNOWN_REPO_GEN,
+                                        repositoryMetadata.pendingGeneration()
+                                    );
```
idegtiarenko, line 225: NIT: I wonder if it is worth adding `withGeneration` to make it clear what we are changing?

kingherc: Thanks @idegtiarenko ! Done

kingherc:
@elasticmachine please run elasticsearch-ci/cloud-deploy

kingherc:
@elasticmachine please run elasticsearch-ci/cloud-deploy

kingherc:
Added @andreidan as optional reviewer because of the doc changes done in https://github.com/elastic/elasticsearch/pull/88391

server/src/internalClusterTest/java/org/elasticsearch/snapshots/CorruptedBlobStoreRepositoryIT.java
```diff
@@ -56,6 +56,52 @@
...
+        logger.info("--> recreate repository with same settings in order to reset corrupted state");
+        assertAcked(client().admin().cluster().preparePutRepository(repoName).setType("fs").setSettings(settings));
+
+        startDeleteSnapshot(repoName, snapshot).get();
+
+        logger.info("--> make sure snapshot doesn't exist");
+        expectThrows(
+            SnapshotMissingException.class,
+            () -> client().admin().cluster().prepareGetSnapshots(repoName).addSnapshots(snapshot).get()
+        );
```
idegtiarenko, line 102: I think this is not required if above line passes (unless there is something after test checks we manually cleared all repos)

idegtiarenko (approved):
Code changes :+1: 

Should we have somebody to approve docs changes?

UPD: just read the above comment

server/src/internalClusterTest/java/org/elasticsearch/snapshots/CorruptedBlobStoreRepositoryIT.java
kingherc: Hi @idegtiarenko , I copied some of the test's logic from `org.elasticsearch.snapshots.CorruptedBlobStoreRepositoryIT#testConcurrentlyChangeRepositoryContents` which does something similar. I think it's just an added check.

andreidan:
@kingherc Thanks for working on this! Awesome to not have to fully recreate the repo 🚀 

> BTW, in the cloud, I manually tested that editing an existing repo does indeed go into the branch 

I've quickly tested this and I think it _might_ not work if the repository has snapshots (or maybe cloud has a special handling of the `found-snapshots` repo). I'll do some more testing tomorrow to understand if the proposed cloud steps would work in all cases

kingherc:
@andreidan , thanks for replying! I will be of course not merging this until we clarify how it works on the cloud.

> I've quickly tested this and I think it might not work if the repository has snapshots (or maybe cloud has a special handling of the found-snapshots repo).
The integration test of this PR does have a snapshot in it when it becomes corrupt. And re-registering the repository then resets the corrupt flag, and the test continues to actually delete the snapshot and it does so successfully. However, it's of course not on the cloud. So for the moment, I believe it may be something in the cloud or some special handling of `found-snapshots` repo; btw, I'm not an expert yet here. If you can also give me guidance, I can also try testing the cloud scenario. But feel free to do it yourself and let me know of any issues, thanks!

server/src/internalClusterTest/java/org/elasticsearch/snapshots/CorruptedBlobStoreRepositoryIT.java
kingherc: Feel free to unresolve the conversation if it's important

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
```diff
@@ -215,8 +215,17 @@ public ClusterState execute(ClusterState currentState) {
                         final RepositoryMetadata updatedMetadata;
                         if (canUpdateInPlace(newRepositoryMetadata, existing)) {
                             if (repositoryMetadata.settings().equals(newRepositoryMetadata.settings())) {
-                                // Previous version is the same as this one no update is needed.
-                                return currentState;
+                                if (repositoryMetadata.generation() == RepositoryData.CORRUPTED_REPO_GEN) {
+                                    // If recreating a corrupted repository with the same settings, reset the corrupt flag (issue #89130).
+                                    // Setting the safe generation to unknown, so that a consistent generation is found.
+                                    repositoryMetadata = repositoryMetadata.withGeneration(
```
tlrx, line 221: I wonder if we should consider the removal of the corruption marker as a "reset" of the registered repository, like if it was deleted and then recreated. If it got corrupted it is possible that the repository content was modified under us, so we likely want to read the repository content again?

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
```diff
@@ -215,8 +215,17 @@ public ClusterState execute(ClusterState currentState) {
                         final RepositoryMetadata updatedMetadata;
                         if (canUpdateInPlace(newRepositoryMetadata, existing)) {
                             if (repositoryMetadata.settings().equals(newRepositoryMetadata.settings())) {
-                                // Previous version is the same as this one no update is needed.
-                                return currentState;
+                                if (repositoryMetadata.generation() == RepositoryData.CORRUPTED_REPO_GEN) {
```
tlrx, line 218: I think it would be nice to have a log message indicating that we are recreating a repository marked as corrupted (the message should contain the name and the known UUID of the repository)

server/src/internalClusterTest/java/org/elasticsearch/snapshots/CorruptedBlobStoreRepositoryIT.java
```diff
@@ -56,6 +56,52 @@
...
+        assertThat(
+            createSnapshotResponse.getSnapshotInfo().successfulShards(),
+            equalTo(createSnapshotResponse.getSnapshotInfo().totalShards())
+        );
+
+        logger.info("--> move index-N blob to next generation");
+        final RepositoryData repositoryData = getRepositoryData(repoName);
+        Files.move(repo.resolve("index-" + repositoryData.getGenId()), repo.resolve("index-" + (repositoryData.getGenId() + 1)));
+
+        assertRepositoryBlocked(client(), repoName, snapshot);
```
tlrx, line 91: should we verify that the repository is indeed corrupted in cluster state?

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
kingherc: @tlrx , thanks for the comment! For the moment, what I think is happening is that the repository gets re-read to find the safe generation to use. Which sounds like what would happen in the case of deletion + recreation, but I may be wrong? I wonder also if there would be any difference whether searchable snapshots are mounted or not in either case. Feel free to elaborate please.

server/src/internalClusterTest/java/org/elasticsearch/snapshots/CorruptedBlobStoreRepositoryIT.java
kingherc: Good idea, did that.

kingherc:
> I'll do some more testing tomorrow to understand if the proposed cloud steps would work in all cases

Hi @andreidan , did you have the chance to look at this? Thanks!

andreidan:
@kingherc Opened https://github.com/elastic/kibana/issues/139896 for the kibana bug. Let's keep the guide as is for now and before 8.5 I'll amend it in case the bug in kibana is not fixed by then.

Thanks for fixing this ❤️

docs/reference/tab-widgets/troubleshooting/snapshot/corrupt-repository.asciidoc
```diff
@@ -48,36 +48,8 @@ and go to **Stack Management > Snapshot and Restore > Repositories**.
...
-
-. With all the details above noted, next delete the repository. Select the 
-checkbox at the left of the repository and hit the "Remove repository" red button
-at the top left of the page.
-
-. Recreate the repository by clicking the "Register Repository" button
-at the top right corner of the repositories table. 
-+
-[role="screenshot"]
-image::images/register_repo.png[{kib} Console,align="center"]
```
andreidan, line 68: This could potentially also be deleted?

kingherc: Hi @andreidan , thanks! I deleted it

kingherc:
Thanks @andreidan for checking for cloud! I handled your comment. Please feel free to review once more and approve/comment.

kingherc:
Thanks @andreidan for checking for cloud! I handled your comment. Please feel free to review once more and approve.

andreidan (approved):
Troubleshooting guide changes LGTM 
Thanks for working on this ! 🚀

kingherc:
@tlrx feel free to give it another review, and tell me if your review should be considered necessary for merging.

@original-brownbear same for you, feel free to tell me if you would like to review this and if your review should be considered necessary for merging.

Thanks!

kingherc:
@tlrx , @original-brownbear I may go ahead and merge this since there are two approvals already. But feel free to take a quick look in case something is awry.

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
tlrx: > For the moment, what I think is happening is that the repository gets re-read to find the safe generation to use. 

Agree. I was worried about resetting the generation to `UNKNOWN_REPO_GEN` without re-reading the repository UUID along with the pending generation, but if I understand correctly the repository UUID is also read and updated in that case too using `RepositoriesService.updateRepositoryUuidInMetadata()`.

I think we also want to prevent resetting the corruption marker if there are on-going snapshots or deletions? I expect them to quickly fail during finalization if the repository is corrupted but it is probably safer to ensure that when updating the cluster state to remove the corruption marker?

> I wonder also if there would be any difference whether searchable snapshots are mounted or not in either case.

Searchable snapshots use the repository UUID to locate a snapshot so as long as both stay the same it should work.

tlrx (commented):
Sorry for the delay @kingherc - I made more comments.

server/src/internalClusterTest/java/org/elasticsearch/snapshots/CorruptedBlobStoreRepositoryIT.java
```diff
@@ -56,6 +58,52 @@
 
 public class CorruptedBlobStoreRepositoryIT extends AbstractSnapshotIntegTestCase {
 
+    public void testRecreateCorruptedRepositoryUnblocksIt() throws Exception {
+        Path repo = randomRepoPath();
+        final String repoName = "test-repo";
+        Settings.Builder settings = Settings.builder()
+            .put("location", repo)
+            // Don't cache repository data because the test manually modifies the repository data
+            .put(BlobStoreRepository.CACHE_REPOSITORY_DATA.getKey(), false);
```
original-brownbear, line 67: Wouldn't we want to test this with a realistic repository that does cache the repository data?

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
kingherc: Hi @tlrx , thanks for the review!

> I think we also want to prevent resetting the corruption marker if there are on-going snapshots or deletions? I expect them to quickly fail during finalization if the repository is corrupted but it is probably safer to ensure that when updating the cluster state to remove the corruption marker?

I see. Would adding a `ensureRepositoryNotInUse(currentState, request.name());` here (like in line 233 below) be enough? Or would you expect the check somewhere else?

tlrx: Yes, that's what I have in mind yes. We probably want a test for this too.

original-brownbear (commented):
I think the approach here is ok, just one question on the test and +1 to only allowing this when there's no ongoing operations.

server/src/internalClusterTest/java/org/elasticsearch/snapshots/CorruptedBlobStoreRepositoryIT.java
kingherc: Hi @original-brownbear , I re-enabled caching for both this test and the test below (from which I copied it). Please review and feel free to unresolve the conversation if it's not enough.

server/src/main/java/org/elasticsearch/repositories/RepositoriesService.java
kingherc: Hi @tlrx , @original-brownbear , it took me some time to make a test with a concurrent snapshot but I finally did it and now resetting a corrupt repository fails if there are ongoing snapshots! It seems to me like a rare case but you never know indeed :) Feel free to review the new test, and unresolve this conversation if it is not sufficient!

kingherc:
@elasticmachine run elasticsearch-ci/part-1 please

kingherc:
@elasticmachine run elasticsearch-ci/part-1 please

server/src/internalClusterTest/java/org/elasticsearch/snapshots/ConcurrentSnapshotsIT.java
```diff
@@ -116,6 +118,77 @@ public void testLongRunningSnapshotAllowsConcurrentSnapshot() throws Exception {
...
+            final SnapshotException ex = expectThrows(
+                SnapshotException.class,
+                () -> clusterAdmin().prepareCreateSnapshot(repoName, "fast-snapshot2")
+                    .setIndices(indexFast)
+                    .setWaitForCompletion(true)
+                    .execute()
+                    .actionGet()
+            );
+            assertThat(ex.getMessage(), containsString("failed to update snapshot in repository"));
+        } catch (Exception ex) {
```
tlrx, line 158: Can we remove this? Are we expecting something else than a `SnapshotException`?

server/src/internalClusterTest/java/org/elasticsearch/snapshots/ConcurrentSnapshotsIT.java
```diff
@@ -116,6 +118,77 @@ public void testLongRunningSnapshotAllowsConcurrentSnapshot() throws Exception {
...
+        assertEquals(RepositoryData.CORRUPTED_REPO_GEN, getRepositoryMetadata(repoName).generation());
+
+        logger.info("--> recreating the repository in order to reset corrupted state, which should fail due to ongoing snapshot");
+        try {
+            final RepositoryConflictException ex = expectThrows(
+                RepositoryConflictException.class,
+                () -> createRepository(repoName, "mock", Settings.builder().put(repoSettings))
+            );
+            assertThat(ex.getMessage(), containsString("trying to modify or unregister repository that is currently used"));
+        } catch (Exception ex) {
```
tlrx, line 170: Same here, I'm not comfortable in catching all exceptions like this

server/src/internalClusterTest/java/org/elasticsearch/snapshots/ConcurrentSnapshotsIT.java
```diff
@@ -116,6 +118,77 @@ public void testLongRunningSnapshotAllowsConcurrentSnapshot() throws Exception {
...
+                // Inner exceptions: RemoteTransportException > RepositoryException (whose message we check)
+                ex.getCause().getCause().getMessage(),
+                containsString("Could not read repository data because the contents of the repository do not match its expected state")
+            );
+        } catch (Exception ex) {
+            ;
+        }
+
+        logger.info("--> without snapshots in progress, finally recreate repository to reset corrupted state");
+        createRepository(repoName, "mock", Settings.builder().put(repoSettings));
```
tlrx, line 189: Can we check that the repository has indeed generation != corruption marker?

tlrx (commented):
Looks good, I've left some minor comments about tests

kingherc:
@tlrx fixed your comments, feel free to take another look

tlrx (approved):
LGTM, and sorry for all the iterations :)

kingherc:
No problem. I think this covers also @original-brownbear 's comments, and with 3 approvals I will go ahead and merge this. Thanks all!

PR #89729, opened 2022-08-30
[Transform] remove some outdated BWC code

hendrikmuhs:
remove code that provides pre 7.13 transport BWC as these old nodes aren't supported in 8.x


droberts195 (approved):
LGTM

PR #89731, opened 2022-08-30
[DOCS] Document security limitation for terms enum API

### LINKED ISSUE #83632: [Docs update] Document security limitation for _terms_enum
The [_terms_enum](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/search-terms-enum.html) API doesn't return any suggestion when document level security is enabled. This seems to be the intended behavior. Could you please update the [security limitations](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-limitations.html) page to mention this as it's not currently mentioned?

Thanks

abdonpijpelink:
Closes #83632

### DIFF
--- a/x-pack/docs/en/security/limitations.asciidoc
+++ b/x-pack/docs/en/security/limitations.asciidoc
@@ -79,6 +79,8 @@ including the following queries:
 ** `percolate` query
 * If suggesters are specified and document level security is enabled, the specified suggesters are ignored.
 * A search request cannot be profiled if document level security is enabled.
+* The <<search-terms-enum,terms enum API>> does not return terms if document 
+level security is enabled.
 
 [discrete]
 [[alias-limitations]]


javanna (approved):
LGTM, shall we add a link to the DLS docs?

abdonpijpelink:
I did add a link to https://www.elastic.co/guide/en/elasticsearch/reference/master/search-terms-enum.html 
Is there another page you'd like to link to?

javanna:
> Is there another page you'd like to link to?
No, sorry, I got confused on which page we are updating. Do we also need to list this under the terms enum API docs?

abdonpijpelink:
>Do we also need to list this under the terms enum API docs?

That would be very nice. I notice that we don't do that for other similarly impacted APIs like [suggesters](https://www.elastic.co/guide/en/elasticsearch/reference/master/search-suggesters.html) and [profile](https://www.elastic.co/guide/en/elasticsearch/reference/master/search-profile.html) though. Do we want to make this a larger effort to consistently add a note to all these docs?

javanna:
> Do we want to make this a larger effort to consistently add a note to all these docs?

I think it could be a good idea but in a follow-up PR, thanks for checking that!

PR #89735, opened 2022-08-30
Operator/ingest

grcevski:
This PR adds support for /_ingest/pipeline for file based settings. 


server/src/main/java/org/elasticsearch/reservedstate/service/ReservedClusterStateService.java
```diff
@@ -101,17 +113,39 @@ public void process(String namespace, XContentParser parser, Consumer<Exception>
...
-            errorListener.accept(
-                new IllegalStateException("Error processing state change request for " + namespace + ", errors: " + errorState, e)
-            );
+        if (stateChunk.state().isEmpty()) {
+            errorListener.accept(null);
             return;
         }
 
-        process(namespace, stateChunk, errorListener);
+        // After we have parsed the json content, we give each handler an opportunity to augment the data
```
grcevski, line 125: This refactoring was required in ReservedClusterStateService because the Ingest transform operations require NodeInfos, which is an async action. For this purpose, I introduced a preTransform step that runs with an async listener, allowing any handler to do additional fetching of data after we've parsed the JSON.

server/src/main/java/org/elasticsearch/action/ingest/ReservedPipelineAction.java
```diff
@@ -0,0 +1,170 @@
...
+        return NAME;
+    }
+
+    public Collection<PutPipelineRequest> prepare(ReservedPipelinesData pipelines) {
+        var exceptions = new ArrayList<String>();
+        for (var pipeline : pipelines.requests) {
+            try {
+                ingestService.validatePipelineRequest(pipeline, pipelines.nodesInfos);
+            } catch (Exception e) {
+                exceptions.add(e.getMessage());
```
dakrone, line 71: Here we collect validation errors and collapse them into a single exception, but I notice in the SLM operator PR we are piecemeal in that the first exception halts execution. Should we add the collection of multiple failures to the other PR?

grcevski: Ah yeah, sorry, I forgot about that. It's better user experience to collect all errors. I'll fix that for SLM.

server/src/main/java/org/elasticsearch/action/ingest/ReservedPipelineAction.java
```diff
@@ -0,0 +1,170 @@
...
+
+        return new ReservedPipelinesData(result, null);
+    }
+
+    @Override
+    public void preTransform(Object parsedContent, ActionListener<Tuple<String, ?>> listener) {
+        NodesInfoRequest nodesInfoRequest = new NodesInfoRequest();
+        nodesInfoRequest.clear();
+        nodesInfoRequest.addMetric(NodesInfoRequest.Metric.INGEST.metricName());
+        nodeClient.admin().cluster().nodesInfo(nodesInfoRequest, new ActionListener<>() {
```
dakrone, line 154: This makes me uneasy, what happens if the cluster is overloaded and this request times out? I think then we cancel all file-based changes right? (because the error will be thrown from the pre-processing if I understand correctly?)

Should this timeout be changed or configurable?

dakrone (changes requested):
I left a few comments, I'm unsure about the preTransform piece. I understand we need the ingest infos, but I wonder if there is a better way to track it somehow so that validation could not require it. What do you think?

grcevski:
> I left a few comments, I'm unsure about the preTransform piece. I understand we need the ingest infos, but I wonder if there is a better way to track it somehow so that validation could not require it. What do you think?

Yeah good point, let me go back and look at ways we can do without it. It complicates things quite a bit, for such a small thing. I'll address the rest of the feedback based on what I manage to do here.

grcevski:
I pushed an update to see if all tests pass, this seems like a simpler change and more efficient, since we don't expect node configuration to change that much along with file based settings. We keep the node infos cached in the file settings service, and we only refresh them when some nodes have changed.

I need to write extra tests for that logic, make sure we set the flag correctly when nodes are added and removed. I initially thought I only need to track the adds, but removes are needed in case the last ingest capable node is removed.

grcevski:
I reworked how the node infos were being called, it's done on demand and only if nodes have joined/left the cluster. I also just pushed an update with new tests that ensure the nodeInfos are fetched correctly and at the right time.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
```diff
@@ -375,28 +391,57 @@ private WatchKey enableSettingsWatcher(WatchKey previousKey, Path settingsDir) t
...
         ) {
-            stateService.process(NAMESPACE, parser, (e) -> {
-                try {
-                    if (e != null) {
-                        errorHandler.accept(e);
+            ReservedStateChunk parsedState = stateService.parse(NAMESPACE, parser);
+            if (nodeInfosRefreshRequired || nodesInfoResponse == null) {
+                var nodesInfoRequest = NodesInfoRequest.requestWithMetrics(NodesInfoRequest.Metric.INGEST);
+
+                clusterAdminClient().nodesInfo(nodesInfoRequest, new ActionListener<>() {
```
dakrone, line 406: Is `processFileSettings` protected by a lock or AtomicBoolean or anything like that? What happens if the file on disk is replaced hundreds of times every second, will we end up issuing many nodes info requests since we don't have any lock to ensure we only retrieve it once?

(I'm not super close to the internals of the file settings stuff, so it may be that this isn't an issue because we only invoke it once every X seconds, let me know if that's the case)

dakrone (approved):
LGTM, I left one question, but as long as we have some sort of protection I don't think it's a blocker or anything.

server/src/main/java/org/elasticsearch/reservedstate/service/FileSettingsService.java
grcevski: This is a really good question, so we have only one thread that processes the file change events and the return of this call (`processFileSettings`) is a latch, the `CountDownLatch waitForCompletion`. The caller of process file settings waits on this latch, for all async actions to complete, before it decides to pick up another event and to check if the file changed since last time. This effectively makes the processing one at a time, until we have successfully written file state or hit an error we don't try to process again.

The relevant caller code is here: https://github.com/elastic/elasticsearch/pull/89735/files#diff-5a2c0417a5b4bdbcc6873fc853db7a8c531b47536bd41b84e20bb6c7b21ad6d9R334

We also don't trust the events we get from the JDK watch service that the file has changed, we only use them as an indicator that we should check if the file is different. If the file hasn't changed the OS file key or the modified timestamp we skip the event. The reason we can't rely on the JDK watcher is because the series of events that get sent are vastly different between OSs.

grcevski:

grcevski:
Thanks Lee!

PR #89636, opened 2022-08-26
[Doc] Release notes for v8.4.1

ywangd:
Gradle generated release notes for v8.4.1 as part of release process.

PR #89737, opened 2022-08-30
Skip update of breaking changes and migration guide for patch releases

mark-vieira:
As a follow up to [this comment](https://github.com/elastic/elasticsearch/pull/89636#discussion_r957378349) we shouldn't be updating the breaking changes or migration guide for patch releases. These things don't change after the initial release of a new minor (or shouldn't) and since we prune old changelogs after each release, we have no way of rebuilding a consolidated document based on all changes across a minor anyhow.

### DIFF
--- a/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/release/GenerateReleaseNotesTask.java
+++ b/build-tools-internal/src/main/java/org/elasticsearch/gradle/internal/release/GenerateReleaseNotesTask.java
@@ -128,26 +128,29 @@ public class GenerateReleaseNotesTask extends DefaultTask {
             changelogsByVersion.getOrDefault(qualifiedVersion, Set.of())
         );
 
-        LOGGER.info("Generating release highlights...");
-        ReleaseHighlightsGenerator.update(
-            this.releaseHighlightsTemplate.get().getAsFile(),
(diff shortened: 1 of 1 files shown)


lockewritesdocs (approved):
I'm not the best person to be approving code changes, but I trust that you're making the right change here.

mark-vieira:
> I'm not the best person to be approving code changes, but I trust that you're making the right change here.

@lockewritesdocs are my assumptions above correct though? In that the migration and breaking changes guide contents are really based on the iniital minor version, and should change with subsequent patch releases?

lockewritesdocs:
> @lockewritesdocs are my assumptions above correct though? In that the migration and breaking changes guide contents are really based on the iniital minor version, and should change with subsequent patch releases?

In the PR summary, you mention that "we shouldn't be updating the breaking changes or migration guide for patch releases." I agree with this statement. In your previous statement (quoted above), did you mean "...should _not_ change with subsequent patch releases"?

mark-vieira:
> In your previous statement (quoted above), did you mean "...should _not_ change with subsequent patch releases"?

Yes 🤦

lockewritesdocs:
> Yes 🤦

In that case, merge away :shipit:

Issue #76261, opened 2021-08-09
Runtime Fields: Support Text fields with Tokenizer / Analyzer


niemyjski:
The Runtime fields feature is nice but without support for text fields with custom tokenizers / analyzers we can't use it as we support a wide range of types. Will there be support for this and additional field types soon?

PR #89743, opened 2022-08-31
[Draft] Potential ECS mappings in Elasticsearch

ruflin:
This PR is intended to have a conversation about potential mappings for ECS in Elasticsearch. It might not ever get merged into the ES code based but allows to keep a history of the conversation and changes. The core idea of the PR are as following:

* Map all core ECS fields through dynamic mappings on the paths. This will make sure only the fields that are contained in each index will be part of the mapping itself
* All fields except `@timestamp` are runtime fields by default and not indexed
* long and double values are matched with a path to ensure also string values are mapped correct. For example `"7"` in JSON needs to be mapped to long.
* `subobject: false` is set to flatten all the documents that are ingested. This ensures no conflicts can happen on fields like `host` and `host.name`

## Implementation

Even though this PR contains the potential template that could be used for ECS it does not contain any code around the implementation. Implementation could happen in multipel phases:

1. ECS component template for each ECS version exists in Elasticsearch that can easily be used
2. ECS config flag in data stream / mapping to use ECS template. This heavily simplifies the usage

## Discussions

* ECS versions: Should this template be versioned? The current suggestion is yes to prevent any unexpected changes to users. At the same time not versioning it would have the benefits of users get the benefits automatically when upgrading Elasticsearch.
* Fields indexed by default: Which fields should be indexed by default?
* How do we handle keyword_text multifields?
* What about extended ECS fields?

### DIFF
--- a/x-pack/plugin/core/src/main/resources/ecs-template.yml
+++ b/x-pack/plugin/core/src/main/resources/ecs-template.yml
@@ -30,15 +30,17 @@ template:
         ignore_malformed: false
 
       # The data stream properties are required in any index following the data stream naming scheme
-      # Discuss: If there is an ecs: enabled flag it can also be used in indices so this would have to be skipped
-      # These fields might not even be needed in this template
-      data_stream.type:
-        type: constant_keyword
-        #value: logs
-      data_stream.dataset:
-        type: constant_keyword
-      data_stream.namespace:
-        type: constant_keyword
+      # But not all data streams or indices which use ECS are part of the data stream naming scheme.
+      # As an example logstash-* and *beats-*. Because of this, the data_stream.* fields are not part
+      # of this template. It is expected that these are set by another template for the data stream
+      # naming scheme.
+      #data_stream.type:
+      #  type: constant_keyword
+      #  #value: logs
+      #data_stream.dataset:
+      #  type: constant_keyword
+      #data_stream.namespace:
+      #  type: constant_keyword
 
 
     # To ensure no empty mappings are in an index, everything except the properties above are mapped as dynamic templates
--- a/x-pack/plugin/core/src/main/resources/ecs-template.yml
+++ b/x-pack/plugin/core/src/main/resources/ecs-template.yml
@@ -142,7 +142,7 @@ template:
           match_mapping_type: object
           match: location
           runtime:
-            type: geopoint
+            type: geo_point
 
       # Match location fields to geopoint
       - match_location:
--- a/x-pack/plugin/core/src/main/resources/ecs-template.yml
+++ b/x-pack/plugin/core/src/main/resources/ecs-template.yml
@@ -108,8 +108,7 @@ template:
       # Any message field except the top level one is matched as text but not indexed
       # TODO: Does this work? Which one has priority?
       - match_message:
-          match_mapping_type: string
-          match: message
+          path_match: message
           runtime:
             type: match_only_text
 
--- a/x-pack/plugin/core/src/main/resources/ecs-template.yml
+++ b/x-pack/plugin/core/src/main/resources/ecs-template.yml
@@ -1,5 +1,8 @@
 ---
+# This template should serve as a foundation for indices and data streams using ECS.
+
 template:
+
   settings:
     # This is a made up setting which does not exist today. The idea is that
     # any field that shows up after the first 1000 fields is kept in source
@@ -65,36 +68,53 @@ template:
           match_mapping_type: string
           runtime:
             type: keyword
-            # Do we need this for runtime?
-            ignore_above: 1024
+            # As it is a runtime field, ignore_above is not needed
+            #ignore_above: 1024
 
-      # Matching all long fields
+      # Matching all ECS core long fields
       # This matches the following fields: bytes, packets, port, duration, severity, pid
-      # No path matching is needed as it is assumed these fields are already in the json doc a long
-      # Elasticsearch would do this mapping automatically but it is specified here for documentation purpose
-      # DISCUSS: If a field "7" as string comes in, will it still be mapped correctly?
-      - long_as_long:
-          match_mapping_type: long
+      # Path matching on these fields is needed as otherwise if the first value of the field would be for example
+      # "7" as a string in JSON, it would have the wrong default mapping. As it is a runtime field, the mapping could
+      # still be adjusted but not something we should have the user do by default.
+      # TODO: if path_match would be an array, this could be defined much nicer and more compact
+      - bytes:
+          path_match: bytes
+          runtime:
+            type: long
+      - packets:
+          path_match: packets
+          runtime:
+            type: long
+      - port:
+          path_match: port
+          runtime:
+            type: long
+      - duration:
+          path_match: duration
+          runtime:
+            type: long
+      - severity:
+          path_match: severity
+          runtime:
+            type: long
+      - pid:
+          path_match: pid
           runtime:
             type: long
 
-      # Matching all long and double fields
+      # Matching all float and double fields
       # This matches the following fields: event.risk_score
-      # No path matching is needed as it is assumed these fields are already in the json doc a float or double
-      # Elasticsearch would do this mapping automatically but it is specified here for documentation purpose
-      # DISCUSS: If a field "7.7" as string comes in, will it still be mapped correctly?
-      - double_as_double:
-          match_mapping_type: double
+      # Why path_match is used, see discussion for long. For runtime fields, it does not matter if it is a float or a double
+      - risk_score:
+          match_path: risk_score
           runtime:
             type: double
 
       # Matching all boolean fields
       # This matches the following fields: exists
-      # No path matching is needed as it is assumed these fields are already in the json doc as boolean
-      # Elasticsearch would do this mapping automatically but it is specified here for documentation purpose
-      # DISCUSS: If a field "true" as string comes in, will it still be mapped correctly?
-      - strings_as_boolean:
-          match_mapping_type: boolean
+      # Why path_match is used, see discussion for long.
+      - exists:
+          match_path: exists
           runtime:
             type: boolean
 
@@ -115,16 +135,13 @@ template:
             type: match_only_text
 
       # Match all ip fields to ip
-      - match_ip:
-          match_mapping_type: string
-          # DISCUSS: If match could be an array, these mappings could be written much more compact
+      # This matches the following fields: ip, network.forwarded_ip
+      - ip:
           match: ip
           runtime:
             type: ip
 
-      # Match all forwarded_ip fields to ip. This is for network.forwarded_ip
-      - match_ip:
-          match_mapping_type: string
+      - forwarded_ip:
           match: forwarded_ip
           runtime:
             type: ip
@@ -139,15 +156,15 @@ template:
             type: flattened
 
       # Match location fields to geopoint
-      - match_location:
+      - location:
           # TODO: What is the correct matching type here?
           match_mapping_type: object
           match: location
           runtime:
             type: geo_point
 
-      # Match location fields to geopoint
-      - match_location:
+      # Match name fields to text_keyword
+      - match_name:
           # TODO: What is the correct matching type here?
           match_mapping_type: string
           match: name
--- a/x-pack/plugin/core/src/main/resources/ecs-template.yml
+++ b/x-pack/plugin/core/src/main/resources/ecs-template.yml
@@ -1,0 +1,225 @@
+---
+template:
+  settings:
+    # This is a made up setting which does not exist today. The idea is that
+    # any field that shows up after the first 1000 fields is kept in source
+    # but does not show up as runtime for index field
+    #index.mapping.total_fields.soft_limit: 1000
+
+    # Ignores any malformed data. Data will still be persisted in _source
+    index.mapping.ignore_malformed": true
+
+  mappings:
+
+    # Any field that is discovered is mapped as a runtime field
+    dynamic: runtime
+
+    # DISCUSS: What are the pros / cons of having this enabled?
+    date_detection: false
+
+    ### All the properties which are set in each template
+    properties:
+
+      # This ensures all subobjects are flattened and prevents and object / keyword conflicts
+      # As subobject is set to false, no objects like host, error or similar are mapped.
+      subobject: false
+
+      # @timestamp is the only field that MUST exist.
+      "@timestamp":
+        type: date
+        ignore_malformed: false
+
+      # The data stream properties are required in any index following the data stream naming scheme
+      # Discuss: If there is an ecs: enabled flag it can also be used in indices so this would have to be skipped
+      # These fields might not even be needed in this template
+      data_stream.type:
+        type: constant_keyword
+        #value: logs
+      data_stream.dataset:
+        type: constant_keyword
+      data_stream.namespace:
+        type: constant_keyword
+
+
+    # To ensure no empty mappings are in an index, everything except the properties above are mapped as dynamic templates
+    # The dynamic mappings here are a best effort to map ECS core fields: https://github.com/elastic/ecs/blob/8.4/generated/csv/fields.csv
+    # As most fields are keyword, these will be automatically be matched correctly. Only core fields were taken into account.
+    # DISCUSS: Should @timestamp and data_stream.* also be moved under dynamic templates?
+    dynamic_templates:
+
+      ### Index mappings
+
+      # Message field is always index
+      - message:
+          path_match: message
+          mapping:
+            type: match_only_text
+
+      ### Runtime mappings
+
+      # Matching all string fields by default to keyword
+      # TODO: Does first / last match win in the list?
+      - strings_as_keyword:
+          match_mapping_type: string
+          runtime:
+            type: keyword
+            # Do we need this for runtime?
+            ignore_above: 1024
+
+      # Matching all long fields
+      # This matches the following fields: bytes, packets, port, duration, severity, pid
+      # No path matching is needed as it is assumed these fields are already in the json doc a long
+      # Elasticsearch would do this mapping automatically but it is specified here for documentation purpose
+      # DISCUSS: If a field "7" as string comes in, will it still be mapped correctly?
+      - long_as_long:
+          match_mapping_type: long
+          runtime:
+            type: long
+
+      # Matching all long and double fields
+      # This matches the following fields: event.risk_score
+      # No path matching is needed as it is assumed these fields are already in the json doc a float or double
+      # Elasticsearch would do this mapping automatically but it is specified here for documentation purpose
+      # DISCUSS: If a field "7.7" as string comes in, will it still be mapped correctly?
+      - float_as_double:
+          match_mapping_type: float
+          runtime:
+            type: double
+      - double_as_double:
+          match_mapping_type: double
+          runtime:
+            type: double
+
+      # Matching all boolean fields
+      # This matches the following fields: exists
+      # No path matching is needed as it is assumed these fields are already in the json doc as boolean
+      # Elasticsearch would do this mapping automatically but it is specified here for documentation purpose
+      # DISCUSS: If a field "true" as string comes in, will it still be mapped correctly?
+      - strings_as_boolean:
+          match_mapping_type: boolean
+          runtime:
+            type: boolean
+
+      # Matching all string fields by default to keyword
+      # TODO: Does first / last match win in the list?
+      - strings_as_keyword:
+          match_mapping_type: string
+          runtime:
+            type: keyword
+            # Do we need this for runtime?
+            ignore_above: 1024
+
+      # Any message field except the top level one is matched as text but not indexed
+      # TODO: Does this work? Which one has priority?
+      - match_message:
+          match_mapping_type: string
+          match: message
+          runtime:
+            type: match_only_text
+
+      # Match all ip fields to ip
+      - match_ip:
+          match_mapping_type: string
+          # DISCUSS: If match could be an array, these mappings could be written much more compact
+          match: ip
+          runtime:
+            type: ip
+
+      # Match all forwarded_ip fields to ip. This is for network.forwarded_ip
+      - match_ip:
+          match_mapping_type: string
+          match: forwarded_ip
+          runtime:
+            type: ip
+
+      # Match all labels fields. Flattened type instead of keyword is used to prevent too many fields to be created.
+      # DISCUSS: Is this the right choice?
+      - match_labels:
+          match_mapping_type: string
+          # TODO: Does this also match the top level labels?
+          match: labels
+          runtime:
+            type: flattened
+
+      # Match location fields to geopoint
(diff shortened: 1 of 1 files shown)


x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,225 @@
...
+      # - orchestrator.namespace
+      # - orchestrator.id
+      # - orchestrator.resource.id
+      # - orchestrator.resource.name
+
+
+
+# Meta information blocks
+_meta:
+  description: default mappings for the logs index template installed by x-pack
```
ruflin, line 223: Need to change this description

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,225 @@
...
+          runtime:
+            type: long
+
+      # Matching all long and double fields
+      # This matches the following fields: event.risk_score
+      # No path matching is needed as it is assumed these fields are already in the json doc a float or double
+      # Elasticsearch would do this mapping automatically but it is specified here for documentation purpose
+      # DISCUSS: If a field "7.7" as string comes in, will it still be mapped correctly?
+      - float_as_double:
+          match_mapping_type: float
```
felixbarny, line 85: JSON doesn't differ between float and double.

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,225 @@
...
+            type: boolean
+
+      # Matching all string fields by default to keyword
+      # TODO: Does first / last match win in the list?
+      - strings_as_keyword:
+          match_mapping_type: string
+          runtime:
+            type: keyword
+            # Do we need this for runtime?
+            ignore_above: 1024
```
felixbarny, line 110: No, I don't think this would have an effect for runtime fields.

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,225 @@
...
+
+      # Any message field except the top level one is matched as text but not indexed
+      # TODO: Does this work? Which one has priority?
+      - match_message:
+          match_mapping_type: string
+          match: message
```
felixbarny, line 116: If you only want the top-level message field to be indexed, you can use

```suggestion
          path_match: message
```

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,225 @@
...
+            type: ip
+
+      # Match all labels fields. Flattened type instead of keyword is used to prevent too many fields to be created.
+      # DISCUSS: Is this the right choice?
+      - match_labels:
+          match_mapping_type: string
+          # TODO: Does this also match the top level labels?
+          match: labels
+          runtime:
+            type: flattened
```
felixbarny, line 142: Runtime fields don't support flattened.

See https://www.elastic.co/guide/en/elasticsearch/reference/current/runtime-mapping-fields.html for a list of supported field types.

Just relying on dynamic: runtime seems sufficient for labels. But if we want to index them, either flattened or subobjects: false may be a good choice.

Creating too many fields doesn't seem unique to labels, there may also be many top-level fields.

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,225 @@
...
+
+      # Match location fields to geopoint
+      - match_location:
+          # TODO: What is the correct matching type here?
+          match_mapping_type: string
+          match: name
+          runtime:
+            # TODO: This is a field type that does not exist today. What ECS does is that for all name fields, it
+            # has a subfield .text to also do text indexing
+            type: text_keyword
```
felixbarny, line 160: Even if text_keyword will be created, it won't be supported for runtime fields. As runtime fields are not analyzed, there's only one type that's valid for string runtime fields, which is `keyword`.

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,225 @@
...
+          runtime:
+            type: flattened
+
+      # Match location fields to geopoint
+      - match_location:
+          # TODO: What is the correct matching type here?
+          match_mapping_type: object
+          match: location
+          runtime:
+            type: geopoint
```
felixbarny, line 150:
```suggestion
            type: geo_point
```

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,225 @@
...
+          match_mapping_type: string
+          runtime:
+            type: keyword
+            # Do we need this for runtime?
+            ignore_above: 1024
+
+      # Matching all long fields
+      # This matches the following fields: bytes, packets, port, duration, severity, pid
+      # No path matching is needed as it is assumed these fields are already in the json doc a long
+      # Elasticsearch would do this mapping automatically but it is specified here for documentation purpose
```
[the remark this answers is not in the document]
ruflin: The part that is important for me here is line 70 which documents which ECS fields it should match.

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: Cleaning up.

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: I switched over to path_match and also removed match_mapping_type: string as I don't think it is needed in this case.

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: Labels by nature have a lot of fields as these are for example k8s labels and not fields that are directly user created. 

@jsoriano I remember you were involved by the time flattened was introduced but it seems we don't use it for labels today but some other places. Did we just never switch over?

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: What should we do about these text fields? Just not have them by default?

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,221 @@
...
+            # Do we need this for runtime?
+            ignore_above: 1024
+
+      # Any message field except the top level one is matched as text but not indexed
+      # TODO: Does this work? Which one has priority?
+      - match_message:
+          match_mapping_type: string
+          match: message
+          runtime:
+            type: match_only_text
```
ruflin, line 113: @felixbarny Based on https://www.elastic.co/guide/en/elasticsearch/reference/current/runtime-mapping-fields.html it seems this will not work?

x-pack/plugin/core/src/main/resources/ecs-template.yml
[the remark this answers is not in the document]
ruflin: BTW maybe a middle ground that could be found is leave the comments in but comment out the full block for documentation purpose. My assumption is that as soon as this would get loaded into ES / cluster state, comments would be kicked out anyways.

x-pack/plugin/core/src/main/resources/ecs-template.yml
jsoriano: We had this [private] https://github.com/elastic/obs-infraobs-team/issues/461 conversation. The general output was to wait till there is better general support, and evaluate case per case. Now using `subobjects` may be also a good choice, except on places where we are also worried about fields explosion.

x-pack/plugin/core/src/main/resources/ecs-template.yml
felixbarny: Correct, only `keyword` is supported. I'd just leave out this section and rely on `dynamic: runtime`. It is somewhat unfortunate that runtime fields don't support full-text searches, only exact matches.

x-pack/plugin/core/src/main/resources/ecs-template.yml
felixbarny: Currently, our only option is to map them as `keyword`. Therefore, we could remove this block and rely on `dynamic: runtime`.

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,222 @@
+---
+template:
```
felixbarny, line 2: Not really relevant for this POC but FYI: this template will not be automatically loaded. You'll need to adjust `StackTemplateRegistry` to load it. Also, there's currently no support for yaml. Shouldn't be too hard to add support for that, though.

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,222 @@
...
+          runtime:
+            type: keyword
+            # Do we need this for runtime?
+            ignore_above: 1024
+
+      # Matching all long fields
+      # This matches the following fields: bytes, packets, port, duration, severity, pid
+      # No path matching is needed as it is assumed these fields are already in the json doc a long
+      # Elasticsearch would do this mapping automatically but it is specified here for documentation purpose
+      # DISCUSS: If a field "7" as string comes in, will it still be mapped correctly?
```
felixbarny, line 73: No, it would be mapped as a runtime field of type `keyword`. However, if users override the mapping so that the runtime field has a type of `long`, I think ES will coerce the type.

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: I mainly pushed the PR here to the Elasticsearch repo to have the conversation where this might end up. Based on if this becomes a component template or just a config flag, I expect @javanna and team to do the "actual" work :-)

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: That makes me think we should have the path match to have a good out of box experience. I assume the following would do the trick for bytes?

```
      - bytes_match:
          path_match: bytes
          runtime:
            type: long
```

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,222 @@
...
+      # This matches the following fields: event.ingested
+      - match_ingested:
+          match_mapping_type: string
+          match: ingested
+          runtime:
+            type: date
+
+# Keyword fields to discuss if should be indexed or not. Proposed rule for discussion. Quorum needs to be reached
+      # for a field to be indexed. Reason is going from runtime to index is easy, otherway around could be a breaking
+      # It should also be discussed if logs-*, metrics-* and traces-* have exact same default.
```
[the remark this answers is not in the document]
ruflin: I'll start some grouping of the fields below to make the triaging easier.

x-pack/plugin/core/src/main/resources/ecs-template.yml
javanna: How many bytes fields are there? One? Is it always there or often or rarely?

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,222 @@
...
+
+      # Matching all long fields
+      # This matches the following fields: bytes, packets, port, duration, severity, pid
+      # No path matching is needed as it is assumed these fields are already in the json doc a long
+      # Elasticsearch would do this mapping automatically but it is specified here for documentation purpose
+      # DISCUSS: If a field "7" as string comes in, will it still be mapped correctly?
+      - long_as_long:
+          match_mapping_type: long
+          runtime:
+            type: long
```
javanna, line 77: I would also think that this is not needed as it is the default behaviour with dynamic:runtime

javanna: another option to prevent type conflicts would be to have a named based convention instead of type based: list the field names and their mappings, then even if they come in as strings they will be mapped as long. Another question is if these fields are only a few and appear often in documents, is it worth mapping them manually. I know I am going a bit back and forth on this, we need to find a good balance :)

x-pack/plugin/core/src/main/resources/ecs-template.yml
javanna: this is not supported and would make the dynamic template validation fail. ignore_above is to not index big values, but runtime fields don't index anything hence we don't accept ignore_above

x-pack/plugin/core/src/main/resources/ecs-template.yml
javanna: We can prioritize https://github.com/elastic/elasticsearch/issues/76261 if needed.

x-pack/plugin/core/src/main/resources/ecs-template.yml
javanna: if we do want to use flattened, these would be indexed though, not runtime. How often are these labels queried / aggregated on?

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,222 @@
...
+          match_mapping_type: string
+          # TODO: Does this also match the top level labels?
+          match: labels
+          runtime:
+            type: flattened
+
+      # Match location fields to geopoint
+      - match_location:
+          # TODO: What is the correct matching type here?
+          match_mapping_type: object
```
javanna, line 142: nit: when you have name conventions, I think you can safely drop the match_mapping_type

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,222 @@
...
+
+      # Match strings fields to wildcard
+      # This matches the following fields: strings
+      # An example field is registry.data.strings
+      - match_strings:
+          # TODO: What is the correct matching type here?
+          match_mapping_type: string
+          match: strings
+          runtime:
+            type: wildcard
```
javanna, line 165: wildcard is also not supported by runtime fields, this would be keyword already with dynamic:runtime, unless you want it indexed.

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: It is not just about bytes but all these: `bytes, packets, port, duration, severity, pid`. There are quite a few, I would say 20-30 (can do an exact count if it helps). If these are there or not heavily depends on the data. If you have anything related to processes, you will have always a pid, if you have access logs, like you will always have a port etc.

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: I'm by now leaning towards matching based on names. Reason is to ensure `"7"` and `7` are both matched to long. At least my understanding is this required as otherwise `"7"` would not be a long.

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: I'll remove it.

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: Would be great if we could prioritise it as it is kind of add that one of the most common types in ES is not supported by runtime fields. Even if there would be huge limitations compared to index, I would argue this is still good enough.

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: I expect in a k8s use case, filter queries on these could happen frequently, not sure about aggregations. But I would prefer to not have these indexed by default but also allow it in runtime.

Is the following correct: If the field is runtime, keyword vs flattened does not really make a difference. Only when indexed there is a difference between flattened and keyword. So basically as long as it is a runtime, we keep it keyword, only for index we recommend flattened. If yes, this leaves us still with the topic that the number of fields could explode because of lots of labels but this is something that could be overcome by the user based on the setting or with the idea from @felixbarny to have a soft limit for # of fields.

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: Agree, I had it more as a safeguard. What will happen, if a field is `location` and it is keyword instead of an object? As it is a runtime field, ES will just ignore it as conversion fails or is this the script error case?

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: Are there any plans to support wildcard?

A more general thought: There are cases like wildcard or text where runtime field will be less powerful then the indexed variant which I would argue is kind of expected. Would it be possible that instead of the person implementing the template having to think this through, ES would internally do a fallback? Lets say wildcard is internally just treated like a keyword field if used as runtime and if the user sends a query with `*` inside it might return a warning. The reason for the above is that with this we keep the nice property of getting from runtime to index is just moving around the definition instead of having to also select a different type.

x-pack/plugin/core/src/main/resources/ecs-template.yml
felixbarny: > I assume the following would do the trick for bytes?

`dynamic: runtime` should be good enough for all these cases. I don't think we need to (and should) map any of these explicitly.

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,222 @@
...
+
+      # @timestamp is the only field that MUST exist.
+      "@timestamp":
+        type: date
+        ignore_malformed: false
+
+      # The data stream properties are required in any index following the data stream naming scheme
+      # Discuss: If there is an ecs: enabled flag it can also be used in indices so this would have to be skipped
+      # These fields might not even be needed in this template
+      data_stream.type:
```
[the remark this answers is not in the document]
felixbarny: Data streams with multiple values for dataset? I think we should avoid something like that happening at all costs 🤯

x-pack/plugin/core/src/main/resources/ecs-template.yml
felixbarny: > If the field is runtime, keyword vs flattened does not really make a difference.

Runtime fields don't support flattened. Flattened implies a specific way of indexing the fields. If we want to use runtime fields for labels, the best option is to just not map them explicitly and rely on dynamic: runtime.

x-pack/plugin/core/src/main/resources/ecs-template.yml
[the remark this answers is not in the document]
ruflin: DLQ? Alert indices?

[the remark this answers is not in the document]
felixbarny: The dlq will also just have a constant_keyword for the dataset such as `_dlq`. To avoid any conflicts that the original document may have with the DLQ metadata (such as @timestamp, error.message, etc.), the original document should be in a dedicated object, such as `event.original`.

I don't know enough about the altering use case.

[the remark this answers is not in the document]
ruflin: I commented out the fields from this template with a note. These are fields specific to the data stream naming scheme and do not have to be enforced in this template.

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: Moved to fully name based matching

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,222 @@
...
+          mapping:
+            type: match_only_text
+
+      ### Runtime mappings
+
+      # Set the ECS version
+      - match_ecs.version:
+          path_match: ecs.version
+          # DISCUSS: In ECS this is still a keyword field
+          type: version
```
felixbarny, line 68: This is mapped as an indexed field but is under the `### Runtime mappings` section. The version type is currently not supported on runtime fields.

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,222 @@
...
+            type: boolean
+
+      # Match all ip fields to ip
+      # This matches the following fields: ip, network.forwarded_ip
+      - match_ip:
+          match: ip
+          runtime:
+            type: ip
+      - match_forwarded_ip:
+          match: forwarded_ip
```
felixbarny, line 134: If we anticipate more ip fields that end with `_ip`
```suggestion
          match: *_ip
```

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: I don't think the ECS version should be an indexed field as I have not seen lots of query on it. I'll add it as runtime field.

Even in ECS today it is a keyword as by the time it was introduced, version did not exist in Elasticsearch. The simplest is to just map it to keyword as this is being supported by Elasticsearch and is what is in ECS. But is it future proof?

x-pack/plugin/core/src/main/resources/ecs-template.yml
```diff
@@ -0,0 +1,222 @@
...
+            type: keyword
+
+      # Matching all ECS core long fields
+      # This matches the following fields: bytes, packets, port, duration, severity, pid
+      # Path matching on these fields is needed as otherwise if the first value of the field would be for example
+      # "7" as a string in JSON, it would have the wrong default mapping. As it is a runtime field, the mapping could
+      # still be adjusted but not something we should have the user do by default.
+      # TODO: if path_match would be an array, this could be defined much nicer and more compact
+      - match_bytes:
+          path_match: bytes
```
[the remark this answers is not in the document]
ruflin: Yes, I keep mixing up `match` and `patch_match`. The idea is that it should match all `*.bytes`. Same is true for all the other `match_*`

x-pack/plugin/core/src/main/resources/ecs-template.yml
ruflin: I like the idea. Can't see a reason at the moment why something would be called _ip but not be an ip ...

ruflin:
As a first step, this is now implemented in elastic-package with dynamic templates: https://github.com/elastic/integrations/issues/5055 This allows each integration to adopt it when ready. Eventually, it should become the default in Elasticsearch and be available directly from Elasticsearch.

I'm temporarily closing this PR but I expect it to be reopened / fresh started again as soon as https://github.com/elastic/integrations/issues/5055 is completed.

Issue #88884, opened 2022-07-28
[CI] RestClusterInfoActionCancellationIT testGetIndicesCancellation failing


iverase:
**Build scan:**
https://gradle-enterprise.elastic.co/s/yigjrvzadco3g/tests/:qa:smoke-test-http:javaRestTest/org.elasticsearch.http.RestClusterInfoActionCancellationIT/testGetIndicesCancellation

**Reproduction line:**
`gradlew ':qa:smoke-test-http:javaRestTest' --tests "org.elasticsearch.http.RestClusterInfoActionCancellationIT.testGetIndicesCancellation" -Dtests.seed=5F940444C5042635 -Dtests.locale=zh-SG -Dtests.timezone=Pacific/Easter -Druntime.java=18`

**Applicable branches:**
8.3

**Reproduces locally?:**
No

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.http.RestClusterInfoActionCancellationIT&tests.test=testGetIndicesCancellation

**Failure excerpt:**
```
junit.framework.AssertionFailedError: found no cancellable tasks

  at junit.framework.Assert.fail(Assert.java:57)
  at junit.framework.Assert.assertTrue(Assert.java:22)
  at junit.framework.TestCase.assertTrue(TestCase.java:192)
  at org.elasticsearch.test.TaskAssertions.lambda$assertAllCancellableTasksAreCancelled$2(TaskAssertions.java:62)
  at org.elasticsearch.test.ESTestCase.assertBusy(ESTestCase.java:1098)
  at org.elasticsearch.test.ESTestCase.assertBusy(ESTestCase.java:1071)
  at org.elasticsearch.test.TaskAssertions.assertAllCancellableTasksAreCancelled(TaskAssertions.java:47)
  at org.elasticsearch.http.RestClusterInfoActionCancellationIT.runTest(RestClusterInfoActionCancellationIT.java:81)
  at org.elasticsearch.http.RestClusterInfoActionCancellationIT.testGetIndicesCancellation(RestClusterInfoActionCancellationIT.java:48)
  at jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
  at java.lang.reflect.Method.invoke(Method.java:577)
  at com.carrotsearch.randomizedtesting.RandomizedRunner.invoke(RandomizedRunner.java:1758)
[...]

Issue #88201, opened 2022-06-30
[CI] IndicesSegmentsRestCancellationIT testCatSegmentsRestCancellation failing


fcofdez:
**Build scan:**
https://gradle-enterprise.elastic.co/s/2o4ugc7mhz6si/tests/:qa:smoke-test-http:javaRestTest/org.elasticsearch.http.IndicesSegmentsRestCancellationIT/testCatSegmentsRestCancellation

**Reproduction line:**
`./gradlew ':qa:smoke-test-http:javaRestTest' --tests "org.elasticsearch.http.IndicesSegmentsRestCancellationIT.testCatSegmentsRestCancellation" -Dtests.seed=8D552D969E2896B -Dtests.locale=ar-IQ -Dtests.timezone=MIT -Druntime.java=8`

**Applicable branches:**
7.17

**Reproduces locally?:**
Didn't try

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.http.IndicesSegmentsRestCancellationIT&tests.test=testCatSegmentsRestCancellation

[...]

PR #89744, opened 2022-08-31
Increase assertAllCancellableTasksAreCancelled timeout

pxsalehi:
The following two failures happen rarely, but both fail in the same `assertBusy` block. I don't have a clue why, and couldn't reproduce them. Considering the amount of checks in that block, maybe a larger timeout is more suitable. (Also it seems from the test history, it is not uncommon for those tests to take 2-3s, so every few thousand runs hitting the 10s timeout seems likely, IMO!) 
Relates https://github.com/elastic/elasticsearch/issues/88884, https://github.com/elastic/elasticsearch/issues/88201


kingherc (approved):
LGTM, assuming CI passes. Hope it was just the timeout the reason for the failures! :crossed_fingers:

pxsalehi:
Thanks @kingherc. Do you know actually if backporting to 8.3 is necessary here? I have the impression that since main is now 8.5, only the CI for 8.4 will be periodically running, and there is no more 8.3 changes.

kingherc:
@pxsalehi seeing the release schedule, I believe 8.3 may not be necessary. But feel free to check with the team as well to be certain.

pxsalehi:
Yes, I thought the same. I think I'll take out the 8.3. Thanks.

pxsalehi:
@elasticmachine update branch

Issue #89687, opened 2022-08-29
[CI] SearchableSnapshotsIntegTests testCreateAndRestoreSearchableSnapshot failing


valeriy42:
Test 4 times in the past 4 weeks (with different error messages though).

**Build scan:**
https://gradle-enterprise.elastic.co/s/knau5tfnrksy6/tests/:x-pack:plugin:searchable-snapshots:internalClusterTest/org.elasticsearch.xpack.searchablesnapshots.SearchableSnapshotsIntegTests/testCreateAndRestoreSearchableSnapshot

**Reproduction line:**
`./gradlew ':x-pack:plugin:searchable-snapshots:internalClusterTest' --tests "org.elasticsearch.xpack.searchablesnapshots.SearchableSnapshotsIntegTests.testCreateAndRestoreSearchableSnapshot" -Dtests.seed=2BC063F8B6C3C9E0 -Dtests.locale=hu -Dtests.timezone=Pacific/Funafuti -Druntime.java=17`

**Applicable branches:**
main, 7.17

**Reproduces locally?:**
Didn't try

**Failure history:**
https://gradle-enterprise.elastic.co/scans/tests?tests.container=org.elasticsearch.xpack.searchablesnapshots.SearchableSnapshotsIntegTests&tests.test=testCreateAndRestoreSearchableSnapshot

**Failure excerpt:**
```
com.carrotsearch.randomizedtesting.UncaughtExceptionError: Captured an uncaught exception in thread: Thread[id=63, name=elasticsearch[node_s0][masterService#updateTask][T#1], state=RUNNABLE, group=TGRP-SearchableSnapshotsIntegTests]

  at __randomizedtesting.SeedInfo.seed([2BC063F8B6C3C9E0:A1A3148325549853]:0)

  Caused by: java.lang.AssertionError: Batched task executors must marshal response headers to the appropriate task context (e.g. using TaskContext#captureResponseHeaders) or suppress them (e.g. using BatchExecutionContext#dropHeadersContext) and must not leak them to the master service, but executor [org.elasticsearch.action.admin.indices.create.AutoCreateAction$TransportAction$$Lambda$2502/0x00000008013849f8@17e71a3f] leaked the following headers: {Warning=[299 Elasticsearch-8.5.0-SNAPSHOT-06226397fb95c1e72693008ebf929a7066e22c64 "[index.data_path] setting was deprecated in Elasticsearch and will be removed in a future release."]}

[...]

PR #84942, opened 2022-03-14
Increase timeout for waiting for the cleanup of the shared index folder

arteam:
I couldn't reproduce any failures of `FrozenSearchableSnapshotsIntegTests testCreateAndRestorePartialSearchableSnapshot` locally and I couldn't get meaningful logs explaining the failure. My only guess is to increase the wait timeout added in #80341
which tried to tackle the same issue.

Closes #84158

PR #89745, opened 2022-08-31
Fix test create and restore searchable snapshot

idegtiarenko:
Related to: #89687

### DIFF
--- a/x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/BaseSearchableSnapshotsIntegTestCase.java
+++ b/x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/BaseSearchableSnapshotsIntegTestCase.java
@@ -66,6 +66,7 @@ import static org.elasticsearch.xpack.searchablesnapshots.cache.shared.SharedByt
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.greaterThanOrEqualTo;
 import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.not;
 
 @ESIntegTestCase.ClusterScope(supportsDedicatedMasters = false, numClientNodes = 0)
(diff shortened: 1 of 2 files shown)


x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/BaseSearchableSnapshotsIntegTestCase.java
```diff
@@ -246,18 +247,18 @@ protected void assertShardFolders(String indexName, boolean snapshotDirectory) t
...
                 assertThat(
-                    snapshotDirectory ? "Snapshot directory doesn't exist" : "Snapshot directory shouldn't exist",
-                    snapshotDirectory,
-                    not(indexExists)
+                    snapshotDirectory ? "Index file should not exist" : "Index file should exist",
+                    indexExists,
+                    not(snapshotDirectory)
                 );
-                assertTrue("Translog doesn't exist", translogExists);
+                assertThat("Translog should exist", translogExists, is(true));
```
idegtiarenko, line 254: Above rewords assertions messages to remove negation

x-pack/plugin/searchable-snapshots/src/internalClusterTest/java/org/elasticsearch/xpack/searchablesnapshots/SearchableSnapshotsIntegTests.java
```diff
@@ -254,7 +255,7 @@ public void testCreateAndRestoreSearchableSnapshot() throws Exception {
         assertSearchableSnapshotStats(restoredIndexName, cacheEnabled, nonCachedExtensions);
 
         ensureGreen(restoredIndexName);
-        assertBusy(() -> assertShardFolders(restoredIndexName, true));
+        assertBusy(() -> assertShardFolders(restoredIndexName, true), 30, TimeUnit.SECONDS);
```
idegtiarenko, line 258: This is consistent with another change: https://github.com/elastic/elasticsearch/pull/84942

idegtiarenko: It looks like this reproduces only if we close the index (eg index files are kept on the disk) **and** we reuse the same index name for the mounted index. This is likely resulting in files not cleaned in time.

tlrx (approved):
LGTM

